PR #92, opened 2020-09-11
Add the PInvoke smoke test

MichalStrehovsky:
Notable changes from CoreRT repo:

* `DllImport(PInvokeNative)` instead of `DllImport(*)` to prevent diverging from what the current tests too.
* Replaced `#if Windows_NT` with `#if TARGET_WINDOWS` in the CPP file
* Replaced `#if TARGET_WINDOWS` in the CS file with RuntimeInformation check.


safern:
@MichalStrehovsky to make official build and publishing signed bits easier, we're going to add a convention to the repo where the branches are called feature/* and those branches will be the ones that will include official builds. I just created feature branch for this experiment, and retargeted this PR.

jkotas (approved):
Thanks

PR #101, opened 2020-09-14
Fix generic virtual method resolution bug

MichalStrehovsky:
Latest Roslyn builds started optimizing (is it really an optimization?) generic virtual method calls into more specific methods.

```diff
.method public hidebysig virtual instance !!T
        Alloc<.ctor T>() cil managed
{
  // Code size       11 (0xb)
  .maxstack  8
  IL_0000:  newobj     instance void My/AllocViaSecondLevelDerived::.ctor()
- IL_0005:  callvirt   instance !!0 My/AllocViaGVMSecondLevelBase::ActuallyAlloc<!!0>()
+ IL_0005:  callvirt   instance !!0 My/AllocViaSecondLevelDerived::ActuallyAlloc<!!0>()
  IL_000a:  ret
} // end of method AllocViaGVMDerived::Alloc
```

This breaks generic virtual method resolution because the resolution assumes we're looking for the slot defining method.

The fix is to normalize to the handle of the slot defining method during codegen.

I hit this in the DynamicGenerics test, but I'm adding more coverage to the Generics test because the newly added logic that walks base hierarchies wouldn't be covered by the existing test.

This also required a RyuJIT fix, but that one turned out to be a nice simplification.

### DIFF
--- a/src/coreclr/src/jit/importer.cpp
+++ b/src/coreclr/src/jit/importer.cpp
@@ -6034,16 +6034,7 @@ GenTree* Compiler::impImportLdvirtftn(GenTree*                thisPtr,
     // CoreRT generic virtual method
     if ((pCallInfo->sig.sigInst.methInstCount != 0) && IsTargetAbi(CORINFO_CORERT_ABI))
     {
-        GenTree* runtimeMethodHandle = nullptr;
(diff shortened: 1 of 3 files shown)


jkotas (approved):

jkotas:
Thank you for tracing this down!

PR #102, opened 2020-09-15
Avoid Ref.Emit in System.Linq.Expressions

MichalStrehovsky:
We will need a more permanent solution for this, but maybe that can be an issue we track for a potential future RI into dotnet/runtime.


jkotas:
>  maybe that can be an issue we track for a potential future RI into dotnet/runtime.

Could you please open an issue where we link all engineering debt that needs to be solved before potential dotnet/runtime RI and add link to it from the .csproj?

LGTM otherwise.

jkotas (approved):

PR #104, opened 2020-09-15
Enable Linux CI

MichalStrehovsky:
Not expected to succeed since there's some issue with timers not firing on Linux, but I want to see if there's any other surprises.


jkotas:
I guess OSX had more problems. Would you mind submitting a WIP PR that enables OSX as a reminder? It is not high priority to make that work, but I can take a look when I get a chance.

MichalStrehovsky:
> I guess OSX had more problems

Yeah, there's a problem with NuGet restore of the object writer (I guess). The `DefaultAppHostRuntimeIdentifier` property that we use in ILCompiler project to find the path to the file in the NuGet package cache is empty. I tried replacing that with `__DistroRid`, but the path still didn't exist. The property being empty is probably related to the restore problem, but I didn't bother rebooting to macOS for now.

> Would you mind submitting a WIP PR that enables OSX as a reminder?

#114

PR #108, opened 2020-09-16
Use declaring syntax as a baseline for generated syntax for containining types

jkoritzinsky:
Simplify creating the syntax for containing types by using the containing type syntax definition from the user-written source as a baseline.

This way we don't need to support redefining accessibility and the correct type declaration keyword.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/DllImportStub.cs
+++ b/DllImportGenerator/DllImportGenerator/DllImportStub.cs
@@ -119,25 +119,14 @@ namespace Microsoft.Interop
             INamedTypeSymbol currType = method.ContainingType;
             while (!(currType is null))
             {
-                var visibility = currType.DeclaredAccessibility switch
-                {
-                    Accessibility.Public => SyntaxKind.PublicKeyword,
-                    Accessibility.Private => SyntaxKind.PrivateKeyword,
-                    Accessibility.Protected => SyntaxKind.ProtectedKeyword,
-                    Accessibility.Internal => SyntaxKind.InternalKeyword,
-                    _ => throw new NotSupportedException(), // [TODO] Proper error message
-                };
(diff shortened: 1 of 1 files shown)


DllImportGenerator/DllImportGenerator/DllImportStub.cs
```diff
@@ -119,25 +119,14 @@ public class GeneratedDllImportData
...
-                    _ => throw new NotSupportedException(), // [TODO] Proper error message
-                };
-
-                typeDecl = typeDecl.AddModifiers(
-                    Token(visibility),
-                    Token(SyntaxKind.PartialKeyword));
+                // Use the declaring syntax as a basis for this type declaration.
+                // Since we're generating source for the method, we know that the current type
+                // has to be declared in source.
+                TypeDeclarationSyntax typeDecl = (TypeDeclarationSyntax)currType.DeclaringSyntaxReferences[0].GetSyntax();
```
AaronRobinsonMSFT, line 125: > `currType.DeclaringSyntaxReferences[0]`

Is there some invariant we know exists so the indexing is okay?

jkoritzinsky: Yep. I mention it in the comment above, but I can make it more explicit if you like.

If we are generating source for a member, that means that the member is declared in source. If the member is declared in source, then its containing type has to also be declared in source (and so on for the whole sequence of declaring types). If a type is declared in source, then it has at least one declaring syntax reference (allowing us to index for the 0th element).

AaronRobinsonMSFT: Ah. No that is okay. I didn't interpret that as we know we have an entry, but that is indeed what it means. W00t.

elinor-fung (approved):

AaronRobinsonMSFT (approved):

PR #112, opened 2020-09-18
Add Internal.Console

MichalStrehovsky:
This is the same Internal.Console that CoreCLR has, but extended with Unix support.

Is there any appetite to have this, potentially in the shared partition?

As I was debugging the timer/threadpool issue on Linux, I quickly found out that due to timing, printf debugging was the only feasible way to find out what's going on.

### DIFF
--- a/src/coreclr/src/nativeaot/System.Private.CoreLib/src/Internal/Console.cs
+++ b/src/coreclr/src/nativeaot/System.Private.CoreLib/src/Internal/Console.cs
@@ -1,0 +1,41 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Text;
+using Microsoft.Win32.SafeHandles;
+
+namespace Internal
+{
+    //
+    // Simple limited console class for internal printf-style debugging in System.Private.CoreLib
+    // and low-level tests that want to call System.Private.CoreLib directly
+    //
+
+    public static class Console
+    {
+#if TARGET_WINDOWS
+        private static readonly SafeFileHandle _outputHandle =
+            new SafeFileHandle(Interop.Kernel32.GetStdHandle(Interop.Kernel32.HandleTypes.STD_OUTPUT_HANDLE), ownsHandle: false);
+#endif
+
+        public static unsafe void Write(string s)
+        {
+#if TARGET_WINDOWS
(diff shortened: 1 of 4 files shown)


src/coreclr/src/nativeaot/System.Private.CoreLib/src/Internal/Console.cs
```diff
@@ -0,0 +1,41 @@
...
+    //
+    // Simple limited console class for internal printf-style debugging in System.Private.CoreLib
+    // and low-level tests that want to call System.Private.CoreLib directly
+    //
+
+    public static class Console
+    {
+#if TARGET_WINDOWS
+        private static readonly SafeFileHandle _outputHandle =
+            new SafeFileHandle(Interop.Kernel32.GetStdHandle(Interop.Kernel32.HandleTypes.STD_OUTPUT_HANDLE), ownsHandle: false);
```
jkotas, line 19: It may be nice to avoid the SafeHandle, and use `WriteFile_IntPtr.cs` to print the message instead.

src/coreclr/src/nativeaot/System.Private.CoreLib/src/Internal/Console.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    //
+    // Simple limited console class for internal printf-style debugging in System.Private.CoreLib
+    // and low-level tests that want to call System.Private.CoreLib directly
+    //
+
+    public static class Console
+    {
+#if TARGET_WINDOWS
+        private static readonly SafeFileHandle _outputHandle =
```
jkotas, line 18: Nit: Statics should have `s_` prefix

src/libraries/Native/Unix/System.Native/pal_string.c
```diff
@@ -23,3 +25,12 @@ int32_t SystemNative_SNPrintF(char* string, int32_t size, const char* format, ..
     va_end(arguments);
     return result;
 }
+
+int32_t SystemNative_PrintF(const char* format, ...)
```
jkotas, line 29: I think it would be better to expose this as `SystemNative_Log` that just prints string. There is an existing OSX-specific `SystemNative_Log`.

jkotas:
> Is there any appetite to have this, potentially in the shared partition?

I think so.

MichalStrehovsky:
Superseded by https://github.com/dotnet/runtime/pull/42983.

PR #113, opened 2020-09-18
Fix shared library build and add test

MichalStrehovsky:
`/wholearchive:` is the voodoo necessary for linker to include object files from a library that are otherwise not referenced. We need to include the bootstrapper because it has an initializer to initialize runtime that we depend on.

I'm keeping the other workaround (`IlcProcessEntrypoint`) in place because we need it due to the `main`/`WinMain` schism for `OutputType=Exe/WinExe`.

The test is from the CoreRT repo.

### DIFF
--- a/src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.props
+++ b/src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.props
@@ -66,10 +66,12 @@ The .NET Foundation licenses this file to you under the MIT license.
     </ItemGroup>
 
     <ItemGroup>
-      <LinkerArg Include="@(NativeLibrary)" />
-      <LinkerArg Include="-Wl,--whole-archive" />
+      <LinkerArg Include="-Wl,--whole-archive" Condition="'$(TargetOS)' != 'OSX'" />
+      <LinkerArg Include="-Wl,--all_load" Condition="'$(TargetOS)' == 'OSX'" />
       <LinkerArg Include="@(_WholeNativeLibrary)" />
-      <LinkerArg Include="-Wl,--no-whole-archive" />
+      <LinkerArg Include="-Wl,--no-whole-archive" Condition="'$(TargetOS)' != 'OSX'" />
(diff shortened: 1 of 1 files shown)


jkotas:
Would it be better to add unresolved reference to runtime library (e.g. a dummy call in the runtime initialization) and satisfy it in the bootstrapper? It would make it work like it used to work before by accident.

MichalStrehovsky:
> Would it be better to add unresolved reference to runtime library (e.g. a dummy call in the runtime initialization) and satisfy it in the bootstrapper? It would make it work like it used to work before by accident.

I'm indifferent about those options - turns out the `/wholearchive` is a common linker switch for these kinds of problems (also exists as `--whole-archive` on ld and `-all_load` on mac). Adding a dummy reference feels hackier.

src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
```diff
@@ -78,6 +74,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <LinkerArg Condition="'$(OutputType)' == 'Exe'" Include="/SUBSYSTEM:CONSOLE" />
       <LinkerArg Condition="'$(OutputType)' == 'WinExe' or '$(OutputType)' == 'Exe'" Include="/ENTRY:$(IlcProcessEntrypoint)" />
       <LinkerArg Include="/NATVIS:&quot;$(MSBuildThisFileDirectory)CoreRTNatVis.natvis&quot;" />
+      <LinkerArg Include="/wholearchive:$(BootstrapperLibraryName).lib" />
```
jkotas, line 77:
```suggestion
      <LinkerArg Include="/WHOLEARCHIVE:$(BootstrapperLibraryName).lib" />
```
to follow casing convention for link.exe options.

jkotas:
> I'm indifferent about those options

Ok, let's give the wholearchive approach a try.

jkotas (approved):

MichalStrehovsky:
Superseded by #270.

PR #121, opened 2020-09-23
Add support for multifile compilation testing

MichalStrehovsky:
In this mode, the individual framework assemblies get compiled into individual OBJ files, then get packaged as a LIB, and tests get built against that LIB.

Improves test build throughput, but this is not compatible with all tests (notably, generic virtual methods that have crossmodule implications are problematic).

Framework.lib is 2,337,191,886 bytes and includes almost 500 MB of Roslyn object code because apparently we need Microsoft.CodeAnalysis.CSharp for some of the tests. We can probably do better in this respect (we could do tweaks in generic homing of canonical things, for example).

I've confirmed with @sandreenko that the JIT assertion was too narrow for CoreRT. I hit the assertion in FSharp.Core.

### DIFF
--- a/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+++ b/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
@@ -7,12 +7,18 @@ parameters:
 
 steps:
   # Build coreclr native test output
-  - script: $(Build.SourcesDirectory)/src/tests/build$(scriptExt) skipstressdependencies $(buildConfigUpper) ${{ parameters.archType }}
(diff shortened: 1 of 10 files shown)


jkotas:
> Microsoft.CodeAnalysis.CSharp 

Should we skip the tests that use Roslyn?

jkotas (approved):

jkotas:
Nice!

MichalStrehovsky:
> Should we skip the tests that use Roslyn?

Sounds good. Shaved 2 minutes off the shared library build.

PR #127, opened 2020-09-26
Make it possible to build all Pri0 test

MichalStrehovsky:
* Separating out the smoke tests into a smoke test category.
* Creating a nativeaot-specific test build, similar to crossgen/composite builds
* Excluding tests that don't build without CoreCLR's CoreLib
* Baselining the failures.


jkotas (approved):
Thanks

PR #128, opened 2020-09-28
Add pipeline to run Pri-0 CoreCLR tests

MichalStrehovsky:


MichalStrehovsky:
> Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

Well, anyway. I triggered it at https://dev.azure.com/dnceng/public/_build/results?buildId=833511&view=results

I don't know what the devops voodoo to let me define a pipeline that doesn't exist in the branch yet (I have a pipeline against the pull request branch). Once we merge, I'll try this again.

jkotas (commented):
LGTM. I may be good idea for somebody like Santi who understands the intricacies of the CI setup to take a look too.

MichalStrehovsky:
@safern do you have an opinion on this? I think I mirrored how we do the crossgen2 testing over in the runtime repo (just a separate YAML file, without a cron schedule).

Testing is currently structured the way it is (no Helix, just build and run on the build machine) because the Helix machines don't have stuff that is necessary to do AOT compilation (AFAIK Mono's AOT test solutions suffer from the same problem).

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -4,21 +4,26 @@ parameters:
   osGroup: ''
   osSubgroup: ''
   uploadTests: false
+  smokeTestsOnly: 'true'
```
safern, line 7: no need to escape this string.

eng/pipelines/runtimelab/coreclr-tests.yml
```diff
@@ -0,0 +1,46 @@
+trigger: none
+
+pr: none
```
safern, line 3: why don't want it to be triggered by PR? or CI?

eng/pipelines/runtimelab/coreclr-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+    buildConfig: Checked
+    platforms:
+    - Windows_NT_x64
+    jobParameters:
+      timeoutInMinutes: 720
+      testGroup: innerloop
+      buildArgs: -s nativeaot+libs+installer -lc release -rc checked
+      extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+      extraStepsParameters:
+        smokeTestsOnly: 'false'
```
safern, line 26: no need to escape this value.

eng/pipelines/runtimelab/coreclr-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+    buildConfig: Release
+    platforms:
+    - Windows_NT_x64
+    jobParameters:
+      timeoutInMinutes: 720
+      testGroup: innerloop
+      buildArgs: -s nativeaot+libs+installer -c $(_BuildConfig) /p:ArchiveTests=true
+      extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+      extraStepsParameters:
+        smokeTestsOnly: 'false'
```
safern, line 45: no need to escape this.

safern:
> Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

This is because of the `-pr: none` trigger in yml file. Coincidentally we have an issue for some pipelines hitting this on dotnet/runtime. You can see this: https://github.com/dotnet/runtime/issues/32777#issuecomment-701263119 which explains how to enable the ability to run via `/azp run` but not enable the build by default on all PRs.

I wonder, why do we want to make this pipeline optional for this branch and not run it on all PRs for this branch and CI as well?

eng/pipelines/runtimelab/coreclr-tests.yml
MichalStrehovsky: It takes 5 hours to complete right now and it's not of a huge value for most changes. Having it on a similar plan as GCStress runs seems ideal. Thanks to the pointer to the issue discussing `pr: none`. I'll try that out!

safern: Sounds good. Then I would recommend just removing the `pr: none` from here and then setting up the pipeline accordingly in the UI so that it is optional and only triggered by comments.

safern (approved):

MichalStrehovsky:
@jkotas do you have admin right in this repo? somehow feature/NativeAOT-clrtests became a protected branch. I want to delete it because I won't need it anymore.

jkotas:
Done. The branch protection rule is set to `feature/*`, so everything that starts with `feature/` will be protected.

MichalStrehovsky:
> Done. The branch protection rule is set to feature/*, so everything that starts with feature/ will be protected.

Ah, good to know. Thank you!

PR #132, opened 2020-09-30
Ensure Object.Finalize doesn't end up in the vtable

MichalStrehovsky:
I already tried to fix this once in dotnet/corert#8081, but the newly added assert now fired when compiling the System.Runtime tests. This was because somewhere in Moq there's code that does `typeof(object).GetMethod("Finalize")` and the reflection patterns scanner attempted to reflection-enable the method as a regular virtual method. I could patch that place up too, but maybe `VirtualMethodUse` nodes for this are fine to have in the system as long as we ensure that the resulting `VTableSlice` still doesn't have the method in it.

Turns out there's similar code in VTableSliceNode.cs at line 130 (for the eagerly built VTableSlice), so this feels like the right place for the fix.


jkotas (approved):
Thanks

PR #133, opened 2020-09-30
SafeHandle marshalling

jkoritzinsky:
Implement default marshalling for SafeHandles using both CoreCLR and CoreRT's implementations as a baseline.

Introduce a new phase : "LeakSafeUnmarshal" to represent unmarshalling steps that cause leaks if skipped in exceptional cases.

Since we don't have access to a `Compilation` object in `MarshallingGenerators.TryCreate` and plumbing one down there requires passing it through 3 or 4 layers of the API, I've renamed the `MarshallingAttributeInfo` distriminated union to `MarshallingInfo` and extended it to also represent marshalling info that requires a `Compilation` object to calculate it.

I've added some xunit-like execution tests in the Demo project that the Main method just calls directly until we decide how we want to do our execution testing.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
+++ b/DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
@@ -240,7 +240,7 @@ namespace Microsoft.Interop
                     methods);
             }
         
-            static MarshallingInfo CreateTypeBasedMarshallingInfo(ITypeSymbol type, Compilation compilation)
+            static MarshallingInfo? CreateTypeBasedMarshallingInfo(ITypeSymbol type, Compilation compilation)
             {
                 var conversion = compilation.ClassifyCommonConversion(type, compilation.GetTypeByMetadataName(TypeNames.System_Runtime_InteropServices_SafeHandle)!);
                 if (conversion.Exists && 
@@ -250,7 +250,7 @@ namespace Microsoft.Interop
                 {
                     return new SafeHandleMarshallingInfo();
                 }
-                return null!;
+                return null;
             }
         }
 #nullable restore


DllImportGenerator/DllImportGenerator/StubCodeContext.cs
```diff
@@ -46,7 +46,14 @@ public enum Stage
...
             /// </summary>
-            KeepAlive
+            KeepAlive,
+
+            /// <summary>
+            /// Convert native data to managed data
+            /// where native values will leak if we
+            /// fail to unmarshal, but do run cleanup.
+            /// </summary>
+            LeakSafeUnmarshal
```
jkotas, line 56: I think this is getting too hard to understand. Why is this different from Unmarshal and Cleanup?

jkotas, line 56: It may be useful to have a comment that describes the overall structure of the method to make it clear how the stages fit together. Something like: 
```
... Setup ...
try
{
    ... Marshal ...
    fixed (... Pin ...)
    {
        ... Invoke ...
    }
    ... Unmarshal ...
}
finally
{
    ... Cleanup...
}

... KeepAlive ...
```

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
```diff
@@ -72,9 +72,9 @@ public static TypePositionInfo CreateForType(ITypeSymbol type, IEnumerable<Attri
         }
 
 #nullable enable
```
AaronRobinsonMSFT, line 74: Can we remove this now?

DllImportGenerator/DllImportGenerator/StubCodeContext.cs
jkoritzinsky: It needs to be different from Unmarshal because SafeHandle unmarshalling has to happen in the `finally` block instead of in the `try` block (once we add the try-finally to the emitted code). It has to happen in the finally block since the handles will leak if an exception is thrown (copying the behavior from CoreCLR and CoreRT).

Additionally, the SafeHandle unmarshalling conceptually isn't cleanup. We aren't cleaning up native resources here, we're unmarshalling the handle into a SafeHandle-derived object. Since we have the ability to introduce new stages and we don't have to overload the meaning of "cleanup", I chose to introduce a new stage.

AaronRobinsonMSFT: @jkoritzinsky I agree with this argument but I think where this is occurring might be more beneficial than such a specific name. Perhaps `FinallyBlock`? I recall the debate about `GuaranteedCleanup` not being accurate since an exception can throw in a `finally` - perhaps letting that slide here is okay because if an exception is thrown couldn't we consider that an ill-formed `GuaranteedCleanup` section?

This is similar to my feedback about `KeepAlive` which might be easier to call `PostInvoke`. I don't recall the argument for why stating it as `KeepAlive` was appropriate but do recall thinking it was valid.

AaronRobinsonMSFT: This is something we are attempting to push into the design documents: https://github.com/dotnet/runtimelab/blob/DllImportGenerator/DllImportGenerator/designs/Pipeline.md

Which also need to be updated with these new stages.

DllImportGenerator/DllImportGenerator/Marshalling/SafeHandleMarshaller.cs
```diff
@@ -0,0 +1,220 @@
...
+                    else if (info.RefKind == RefKind.Ref)
+                    {
+                        // Decrement refcount on original SafeHandle if we addrefd
+                        yield return IfStatement(
+                            IdentifierName(addRefdIdentifier),
+                            ExpressionStatement(
+                                InvocationExpression(
+                                    MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression,
+                                        IdentifierName(managedIdentifier),
+                                        IdentifierName("DangerousRelease")),
```
AaronRobinsonMSFT, line 184: ~~"DangerousRelease"~~ -> `nameof(SafeHandle.DangerousRelease)`

DllImportGenerator/DllImportGenerator/Marshalling/SafeHandleMarshaller.cs
```diff
@@ -0,0 +1,220 @@
...
+                                                    PredefinedType(Token(SyntaxKind.BoolKeyword)),
+                                                    SingletonSeparatedList(
+                                                        VariableDeclarator(addRefdIdentifier)
+                                                        .WithInitializer(EqualsValueClause(LiteralExpression(SyntaxKind.FalseLiteralExpression))))));
+                    
+                    }
+                    if (info.IsByRef && info.RefKind != RefKind.In)
+                    {
+                        // We create the new handle in the Setup phase
+                        // so we reduce the possible failure points during unmarshalling, where we would
```
jkotas, line 85:
```suggestion
                        // so we eliminate possible failure points during unmarshalling, where we would
```

DllImportGenerator/DllImportGenerator/StubCodeContext.cs
jkoritzinsky: I'd be ok calling it `GuaranteedUnmarshal` and saying that an exception thrown during execution of this section is ill-formed. I'm not a fan of calling this section a cleanup section since it's doing unmarshalling and the Cleanup section is emitted in the same finally block.

IIRC with `KeepAlive` we just kind of left it as-is. I think the reasoning was that we couldn't think of anything else that would go in that section, and that we would rename it once something else fit to go there.

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
jkoritzinsky: I still haven't enabled nullability in the DllImportGenerator project, just in Ancillary.Interop. I'll open an issue to track enabling it and enable it in here in the next PR.

jkotas:
>  an exception can throw in a finally

I think exception thrown during cleanup should be a non-scenario.

jkoritzinsky:
Any more feedback?

DllImportGenerator/DllImportGenerator/Marshalling/SafeHandleMarshaller.cs
```diff
@@ -0,0 +1,230 @@
...
+                            yield return ExpressionStatement(assignHandleToNativeExpression);
+                        }
+                    }
+                    break;
+                case StubCodeContext.Stage.GuaranteedUnmarshal:
+                    StatementSyntax unmarshalStatement = ExpressionStatement(
+                        InvocationExpression(
+                            MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression,
+                                ParseTypeName(TypeNames.System_Runtime_InteropServices_MarshalEx),
+                                IdentifierName("SetHandle")),
```
christianscheuer, line 165: Being a guest here, but happened to see this is the only that doesn't currently use nameof?

jkoritzinsky: The MarshalEx type isn't a type in the .NET Standard 2.0 contract, so we can't use `nameof` for any of its methods. It's currently in Ancillary.Interop. In the future its methods will move into dotnet/runtime in some fashion.

DllImportGenerator/Ancillary.Interop/MarshalEx.cs
```diff
@@ -0,0 +1,29 @@
+
+using System.Reflection;
+
+namespace System.Runtime.InteropServices
+{
+    /// <summary>
+    /// Marshalling helper methods that will likely live in S.R.IS.Marshal
+    /// when we integrate our APIs with dotnet/runtime.
+    /// </summary>
+    public static class MarshalEx
```
AaronRobinsonMSFT, line 10: Can't we just make these extension methods on `Marshal`? That way we can simply add the support in the runtime and it "just works"?

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
```diff
@@ -234,6 +239,19 @@ NativeMarshallingAttributeInfo CreateNativeMarshallingInfo(AttributeData attrDat
...
+            {
+                var conversion = compilation.ClassifyCommonConversion(type, compilation.GetTypeByMetadataName(TypeNames.System_Runtime_InteropServices_SafeHandle)!);
+                if (conversion.Exists && 
+                    conversion.IsImplicit &&
+                    conversion.IsReference &&
+                    !type.IsAbstract)
+                {
+                    return new SafeHandleMarshallingInfo();
+                }
+                return null!;
```
AaronRobinsonMSFT, line 253: This seems wrong. Why not just update the API to be accurate?

AaronRobinsonMSFT (approved):

DllImportGenerator/Ancillary.Interop/MarshalEx.cs
jkoritzinsky: You can't add extension static methods (and as a result you can't add extension methods to static classes). C# doesn't support that.

PR #25, opened 2020-07-23
Adding CI coverage to NativeAOT branch

joperezr:
FYI: @jkotas

PR #136, opened 2020-10-01
Fbsd build

wfurt:
Initial build setup. This may be updated later but the idea would be get Linux and FreeBSD builds from same branch so it is easier to compare supported and experimental platform. I don't know if Windows_NT_x64 is needed for any of the packaging steps but for now, I don't really care about packages as well as FreeBSD changes are more likely regress Linux (and Unix in general) than Windows.


eng/pipelines/runtimelab.yml
```diff
@@ -6,7 +6,7 @@ trigger:
   batch: true
   branches:
     include:
-    # Add Experiment branch here
+    - FreeBSD
```
jkotas, line 9: This should be `feature/FreeBSD`

wfurt: ok. I followed the linked example https://github.com/dotnet/runtimelab/pull/25/files

wfurt: updated

eng/pipelines/runtimelab.yml
```diff
@@ -60,8 +60,8 @@ jobs:
     jobTemplate: /eng/pipelines/common/global-build-job.yml
     buildConfig: Checked
     platforms:
-    - Linux_x64
-    - Windows_NT_x64
+    # - Linux_x64
```
safern, line 63: Just remove the whole section if we don't want to build and test a checked runtime on FreeBSD.

eng/pipelines/runtimelab.yml
```diff
@@ -76,7 +76,7 @@ jobs:
     buildConfig: Release
     platforms:
     - Linux_x64
-    - Windows_NT_x64
+    - FreeBSD
     jobParameters:
```
safern, line 80: Do we want to run tests or just build at the moment?

wfurt: I wish we can run tests but I don't think we have any infrastructure with FreeBSD nodes. For build, we cross-compile on Linux.

eng/pipelines/runtimelab.yml
wfurt: ok, I'll do that. Time may prove me wrong but I don't see value in burning more CPU cycles right now.

eng/pipelines/runtimelab.yml
jkotas: We have added `feature/` prefix to all branches after this example was created. I have merged a PR that enabled the CI for this branch (without the rest of the changes in this PR), and update the example to point to it.

Sorry about the merge conflict that you may get.

eng/pipelines/runtimelab.yml
safern: I see. Can't we add a `FreeBSD` docker image and run tests on that image?

eng/pipelines/runtimelab.yml
```diff
@@ -76,7 +61,7 @@ jobs:
...
       testGroup: innerloop
       buildArgs: -s clr+libs+libs.tests+installer -c $(_BuildConfig) /p:ArchiveTests=true
```
safern, line 67: Since we're not running tests at the moment you need to remove the `extraStepsTemplate` and `extraStepsParameters` section.

eng/pipelines/runtimelab.yml
jkotas: FreeBSD is not Linux. We would need FreeBSD host to run FreeBSD docker image.

safern: Right, that is what @wfurt mentioned offline.

eng/pipelines/runtimelab.yml
```diff
@@ -57,22 +57,27 @@ stages:
...
+        isOfficialBuild: ${{ variables.isOfficialBuild }}
+        testGroup: innerloop
+        extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+        extraStepsParameters:
+          uploadLibrariesTests: ${{ eq(variables.isOfficialBuild, false) }}
+          uploadUnsignedArtifacts: false
+        ${{ if eq(variables.isOfficialBuild, false) }}:
+          buildArgs: -s clr+libs+libs.tests+installer -c $(_BuildConfig) /p:ArchiveTests=true
+        ${{ if eq(variables.isOfficialBuild, true) }}:
+          buildArgs: -s clr+libs -c $(_BuildConfig)
```
safern: Do you need OOB packages? If not you can probably remove the -allConfigurations leg at the bottom, but I'm not sure.

eng/pipelines/runtimelab.yml
```diff
@@ -57,22 +57,27 @@ stages:
...
+      - FreeBSD_x64
+      jobParameters:
+        isOfficialBuild: ${{ variables.isOfficialBuild }}
+        testGroup: innerloop
+        extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+        extraStepsParameters:
+          uploadLibrariesTests: ${{ eq(variables.isOfficialBuild, false) }}
+          uploadUnsignedArtifacts: false
+        ${{ if eq(variables.isOfficialBuild, false) }}:
+          buildArgs: -s clr+libs+libs.tests+installer -c $(_BuildConfig) /p:ArchiveTests=true
```
safern: It seems like you don't need archiveTests=true here since we're not running tests.

eng/pipelines/runtimelab.yml
```diff
@@ -57,22 +57,27 @@ stages:
...
+      buildConfig: Release
+      platforms:
+      - Linux_x64
+      - FreeBSD_x64
+      jobParameters:
+        isOfficialBuild: ${{ variables.isOfficialBuild }}
+        testGroup: innerloop
+        extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+        extraStepsParameters:
+          uploadLibrariesTests: ${{ eq(variables.isOfficialBuild, false) }}
```
safern: set this to false always, since we're not running tests.

eng/pipelines/runtimelab.yml
```diff
@@ -57,22 +57,27 @@ stages:
...
+          uploadUnsignedArtifacts: false
```
safern: You need to change this to be the value of isOfficialBuild as we want to ship the packages produced here for freeBSD right?

```suggestion
          uploadUnsignedArtifacts: ${{ variables.isOfficialBuild }}
```

eng/pipelines/runtimelab.yml
wfurt: I'm not sure either. I think there may be competing goals - to make the build as small and cheap as possible and to have everything similar to "normal" runtime as much as possible. For now I'm trying to get pass as see what I'm missing.

eng/pipelines/runtimelab.yml
wfurt: for this one I'm less sure. In ideal case we would be running tests. I'm sort of hoping that if we build and publish tests, it would be easier to download and run everything outside of Helix. Do you have sense how much space does it take?

safern: Hmm, I have no idea, but we should be able to look at the artifact that we publish with the libraries tests in it.

eng/pipelines/runtimelab.yml
wfurt: updated.

eng/pipelines/runtimelab.yml
akoeplinger: there's a FreeBSD image on the Azure Image Gallery so it should be pretty easy to create a Helix queue off of that.

wfurt: the issues we run into in the past are mostly not technical @akoeplinger.

wfurt:
It seems like this reached somewhat stable point. I'm wondering if this would be good enough to enable the build. I will sort out tests & packaging as time permits e.g. I don't necessarily see this as final version. I may also shout out to community for help once we have basics in place.

eng/pipelines/runtimelab.yml
```diff
@@ -119,108 +101,3 @@ stages:
...
-          isFullMatrix: false
-          isOfficialBuild: false
-          testScope: innerloop
-          liveRuntimeBuildConfig: Release
-          dependsOnTestBuildConfiguration: Release
-          dependsOnTestArchitecture: x64
-          dependsOn:
-          - build_Windows_NT_x64_Release_
-
-- ${{ if eq(variables.isOfficialBuild, true) }}:
```
safern, line 223: You need this for official builds.

safern: And also, it might be worth trying an official build before merging if you're interested on them as well.

wfurt: how do I make official build?

safern: You need to push a branch into this repo: https://dnceng.visualstudio.com/internal/_git/dotnet-runtimelab

Then queue a build, using that branch as the source branch on this pipeline: https://dnceng.visualstudio.com/internal/_build?definitionId=895

eng/pipelines/runtimelab.yml
```diff
@@ -101,3 +101,7 @@ stages:
             isOfficialBuild: true
             librariesBinArtifactName: libraries_bin_official_allconfigurations
 
+- ${{ if eq(variables.isOfficialBuild, true) }}:
+  - template: /eng/pipelines/official/stages/publish.yml
+      parameters:
+            isOfficialBuild: ${{ variables.isOfficialBuild }}
```
safern, line 107: this indentation is off by one tab.

wfurt: I'll take a look. I tried to mimic the original file but it probably got messed up in the copy.

rmszc81:
I have interest on this topic, build .net 5 in freebsd 12.2 or later and even create a ports if possible. How can I help you guys in this mission?

safern (approved):
@wfurt were you able to test the official build?

src/libraries/Directory.Build.props
```diff
@@ -118,6 +118,9 @@
     <!-- We build linux-musl-arm on a ubuntu container, so we can't use the toolset build for alpine runtime. We need to use portable linux RID for our toolset in order to be able to use it. -->
     <ToolRuntimeRID Condition="'$(_runtimeOS)' == 'linux-musl' and $(TargetArchitecture.StartsWith('arm')) and !$(HostArch.StartsWith('arm'))">linux-x64</ToolRuntimeRID>
 
+    <!-- cross-targeting FreeBSD on Linux -->
+    <ToolRuntimeRID Condition="'$(_runtimeOS)' == 'freebsd' and $([System.Runtime.InteropServices.RuntimeInformation]::OSDescription.StartsWith('Linux'))">linux-x64</ToolRuntimeRID>
```
safern:
```suggestion
    <ToolRuntimeRID Condition="'$(TargetOS)' == 'freebsd' and $([System.Runtime.InteropServices.RuntimeInformation]::OSDescription.StartsWith('Linux'))">linux-x64</ToolRuntimeRID>
```

_runtimeOS, will not be `freebsd`.

wfurt: I tried that but my local build is still failing @safern. I need to find more time to debug this.

safern: Ok, ping me offline if you need any help figuring out why.

PR #139, opened 2020-10-01
Fix and optimize SuppressGCTransition

jkotas:


src/coreclr/src/tools/Common/JitInterface/CorInfoImpl.cs
```diff
@@ -1294,9 +1295,17 @@ private void findSig(CORINFO_MODULE_STRUCT_* module, uint sigTOK, CORINFO_CONTEX
 
             Get_CORINFO_SIG_INFO(methodSig, sig);
 
-            if (sig->callConv == CorInfoCallConv.CORINFO_CALLCONV_UNMANAGED)
+            // CORINFO_CALLCONV_UNMANAGED is handled by Get_CORINFO_SIG_INFO
+            Debug.Assert(sig->callConv != CorInfoCallConv.CORINFO_CALLCONV_UNMANAGED);
+
+            // TODO: Replaced this with a public mechanism to mark calli with SuppressGCTransition once it becomes available.
```
MichalStrehovsky, line 1301: Nit:

```suggestion
            // TODO: Replace this with a public mechanism to mark calli with SuppressGCTransition once it becomes available.
```

MichalStrehovsky (approved):
Thank you!

PR #142, opened 2020-10-04
Fix issues in marshallers

MichalStrehovsky:
* Fixes a pretty obvious bug in reverse interface marshalling
* Fixes error case in `SafeHandle` marshaller. The existing code was trying to generate throwing code in a wrong spot, leaving IL that is difficult to compute stack height for with our naive algorithm. `MissingMethodException` is the CLR-compatible behavior (see https://github.com/dotnet/runtime/issues/43016).

### DIFF
--- a/src/coreclr/src/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
+++ b/src/coreclr/src/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
@@ -1686,14 +1686,9 @@ namespace Internal.TypeSystem.Interop
             var ctor = ManagedType.GetParameterlessConstructor();
             if (ctor == null || ((MetadataType)ManagedType).IsAbstract)
             {
-#if READYTORUN
-                // Let the runtime generate the proper MissingMemberException for this.
-                throw new NotSupportedException();
-#else
                 ThrowHelper.ThrowMissingMethodException(ManagedType, ".ctor",
                     new MethodSignature(MethodSignatureFlags.None, genericParameterCount: 0,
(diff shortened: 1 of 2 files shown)


src/coreclr/src/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
```diff
@@ -1690,20 +1690,9 @@ private void AllocSafeHandle(ILCodeStream codeStream)
                 // Let the runtime generate the proper MissingMemberException for this.
                 throw new NotSupportedException();
 #else
-                var emitter = _ilCodeStreams.Emitter;
-
-                MethodSignature ctorSignature = new MethodSignature(0, 0, Context.GetWellKnownType(WellKnownType.Void),
-                      new TypeDesc[] {
-                          Context.GetWellKnownType(WellKnownType.String)
-                      });
-                MethodDesc exceptionCtor = InteropTypes.GetMissingMemberException(Context).GetKnownMethod(".ctor", ctorSignature);
```
jkotas, line 1699: InteropTypes.GetMissingMemberException is no longer used. Delete it?

src/coreclr/src/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
```diff
@@ -1690,20 +1690,9 @@ private void AllocSafeHandle(ILCodeStream codeStream)
                 // Let the runtime generate the proper MissingMemberException for this.
```
jkotas, line 1690: Delete this `#if READYTORUN`?

PR #143, opened 2020-10-05
Mark System.Private assemblies as framework assemblies

MichalStrehovsky:
We mostly need `[assembly: AssemblyMetadata(".NETFrameworkAssembly", "")]` because compiler checks for this in places. Versioning.targets adds a bunch of other stuff too.

### DIFF
--- a/src/coreclr/src/nativeaot/Directory.Build.props
+++ b/src/coreclr/src/nativeaot/Directory.Build.props
@@ -98,4 +98,6 @@
     <AotCommonPath>$(MSBuildThisFileDirectory)Common\src\</AotCommonPath>
     <CompilerCommonPath>$(MSBuildThisFileDirectory)..\tools\Common\</CompilerCommonPath>
   </PropertyGroup>
+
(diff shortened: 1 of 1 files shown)


jkotas:
The CI errors should be fixed by #145

src/coreclr/src/nativeaot/Directory.Build.props
```diff
@@ -98,4 +98,6 @@
     <AotCommonPath>$(MSBuildThisFileDirectory)Common\src\</AotCommonPath>
     <CompilerCommonPath>$(MSBuildThisFileDirectory)..\tools\Common\</CompilerCommonPath>
   </PropertyGroup>
+
+  <Import Condition="'$(MSBuildProjectName)' != 'System.Private.CoreLib'" Project="$(RepositoryEngineeringDir)versioning.targets" />
```
jkotas, line 102: Why do we exclude this for CoreLib?

MichalStrehovsky: CoreCLR CoreLib also doesn't include this target and does these things "manually". Just wanted to stay in sync.

jkotas (approved):

PR #151, opened 2020-10-08
Ship Utf8String experimental package and also don't build all packages from libraries\src\pkg

safern:
Given that the package name contains Experimental, we were marking it as NonShipping, hence not making it to the experimental nuget feed.

cc: @jkotas

### DIFF
--- a/src/libraries/libraries-packages.proj
+++ b/src/libraries/libraries-packages.proj
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="$(PkgDir)*\*.proj" Exclude="$(PkgDir)test\*" />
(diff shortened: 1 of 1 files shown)


jkotas (approved):

safern:
@jkotas should we also stop running all libraries tests and run just the ones that affect the experiment?

jkotas:
I think it is fine to run all libraries tests by default. The experiments can tune it as they wish.

safern:
Yeah I agree on that. My question was towards this experiment. If we should only run the System.Utf8String.Experimental.Tests and probably System.Runtime.*?

jkotas:
I do not have an opinion. It really only matters once there is active development on this experiment.

safern:
cc: @GrabYourPitchforks in case he will be actively developing here and has an opinion.

PR #156, opened 2020-10-08
Publish all packages generated from the experiment branch when in runtimelab

safern:
Still Utf8String package was not published and found the real reason. We were looking for a very specific path, which doesn't exist on runtimelab. 

I will port this to dotnet/runtime for future experiments.

### DIFF
--- a/eng/pipelines/installer/jobs/base-job.yml
+++ b/eng/pipelines/installer/jobs/base-job.yml
@@ -365,7 +365,7 @@ jobs:
             parameters.liveRuntimeBuildConfig,
             parameters.globalBuildSuffix) }}
     - ${{ if eq(parameters.useOfficialAllConfigurations, true) }}:
-      - build_Windows_NT_x64_${{ parameters.liveLibrariesBuildConfig }}_All_Configurations
+      - build_Windows_NT_x64_${{ parameters.liveLibrariesBuildConfig }}_AllConfigurations
 
   steps:
 


src/installer/publish/Directory.Build.targets
```diff
@@ -15,6 +15,10 @@
       $(DownloadDirectory)IntermediateUnsignedArtifacts\Windows_x64\Shipping\a.nupkg
   -->
   <Target Name="FindDownloadedArtifacts">
+    <PropertyGroup>
```
jkotas, line 18: All this should be deleted once there are no experimental packages in dotnet/runtime. Add a comment about it?

jkotas (approved):

src/installer/publish/Directory.Build.targets
safern: Why should it be deleted? The way that the installer publishing infrastructure is done is to publish just the ridless assets from one leg (instead of all legs) because all installer legs produce ridless assets. It publishes that plus libraries_allconfigurations assets.

jkotas: I thought that this is another place where we need to override the check for experimental packages to get them published.

Is this change needed because publishing in dotnet/runtimelab is done differently (in a simpler way) than in dotnet/runtime?

safern: Yeah... this is because `dotnet/runtimelab` just publishes the packages it needs and we don't need to filter anything out.

safern: Hmm, this didn't work. I think I need to adjust how we upload packages instead.

safern:
Ok the last official build with this changes now makes sense and includes the packages I'm expecting we should include.

PR #159, opened 2020-10-09
Fix official build

safern:
The official build is failing with:
```
BUILD: Generating native version headers
  MSBUILD : error MSB1001: Unknown switch.
  Switch: -skipruntime
  
  For switch syntax, type "MSBuild -help"
##[error](NETCORE_ENGINEERING_TELEMETRY=Build) Build failed.
  See log: "F:\workspace\_work\1\s\src\coreclr\..\..\artifacts\log\Release\GenerateVersionHeaders_Windows_NT__x64__Release.binlog"
##[error]BUILD: Error: Failed to generate version headers.
F:\workspace\_work\1\s\src\coreclr\runtime.proj(38,5): error MSB3073: The command ""F:\workspace\_work\1\s\src\coreclr\build-runtime.cmd" -x64 -release -ci -enforcepgo /p:OfficialBuildId=20201008.7 -skipruntime -skipalljits" exited with code 1.
##[error]src\coreclr\runtime.proj(38,5): error MSB3073: (NETCORE_ENGINEERING_TELEMETRY=Build) The command ""F:\workspace\_work\1\s\src\coreclr\build-runtime.cmd" -x64 -release -ci -enforcepgo /p:OfficialBuildId=20201008.7 -skipruntime -skipalljits" exited with code 1.
```

The reason is because the way `build-runtime.cmd`  processes arguments, all MSBuild args need to be at the end, because if they're before a script argument, whatever is after the MSBuild arg, is treated as an unprocessed arg and passed through to MSBuild.

If we hit this point: https://github.com/dotnet/runtimelab/blob/d59fb6e0ab808d87a595db8ab6c0202398f2db47/src/coreclr/build-runtime.cmd#L192-L199

Which is when no matching argument was found, we finish the arg loop and pass whatever was left to process down to MSBuild. 

The reason why we don't hit this in dotnet/runtime, is because we build coreclr by calling `build-runtime.cmd` directly instead of going through the root `build.cmd` which ends up calling `coreclr\runtime.proj`. 

cc: @jkotas @MichalStrehovsky

### DIFF
--- a/src/coreclr/runtime.proj
+++ b/src/coreclr/runtime.proj
@@ -20,13 +20,13 @@
                                    '$(NoPgoOptimize)' != 'true'"
                                    Include="-enforcepgo" />
       <_CoreClrBuildArg Condition="$([MSBuild]::IsOsPlatform(Windows)) and '$(CrossDac)' != ''" Include="-$(CrossDac)dac" />
-      <_CoreClrBuildArg Condition="'$(OfficialBuildId)' != ''" Include="/p:OfficialBuildId=$(OfficialBuildId)" />
       <_CoreClrBuildArg Condition="'$(NoPgoOptimize)' == 'true'" Include="-nopgooptimize" />
       <_CoreClrBuildArg Condition="'$(ClrRuntimeSubset)' != 'true'" Include="-skipruntime" />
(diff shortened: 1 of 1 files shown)


jkotas (approved):
Thanks!

PR #185, opened 2020-10-09
Merge runtime/master into feature/NativeAOT

safern:
cc: @jkotas @MichalStrehovsky

### DIFF
--- a/src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Environment.CoreRT.Unix.cs
+++ b/src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Environment.CoreRT.Unix.cs
@@ -111,7 +111,7 @@ namespace System
             return results;
 
             // Use a local, unsafe function since we cannot use `yield return` inside of an `unsafe` block
-            unsafe bool ParseEntry(IntPtr current, out string key, out string value)
+            static unsafe bool ParseEntry(IntPtr current, out string key, out string value)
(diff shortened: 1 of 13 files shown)


safern:
@jkotas @MichalStrehovsky there was a change in runtime/master to move the targeting TFMs to net6.0 in case you want the ILCompiler to support net5.0.

jkotas:
Thanks for the heads up. I do not think it needs to be addressed in this PR.

safern:
The build is failing, I guess is because of the new compilers and this code is probably not in runtime? the path contains .Corert in it.

```
src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Threading/Thread.CoreRT.Unix.cs(23,84): error CS8902: (NETCORE_ENGINEERING_TELEMETRY=Build) 'Thread.OnThreadExit()' is attributed with 'UnmanagedCallersOnly' and cannot be converted to a delegate type. Obtain a function pointer to this method.
```

jkotas:
I am taking care of the build breaks

safern:
Hmm seems like we need to mark more members as unsafe:

```
src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Threading/Thread.CoreRT.Unix.cs(23,52): error CS0214: (NETCORE_ENGINEERING_TELEMETRY=Build) Pointers and fixed size buffers may only be used in an unsafe context
```

PR #191, opened 2020-10-11
Replace McgIntrinsics with function pointers

jkotas:
Fixes https://github.com/dotnet/corert/issues/86

### DIFF
--- a/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
+++ b/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
@@ -231,7 +231,12 @@ namespace ILCompiler
             {
                 return GetByRefTypeIndex(((ParameterizedType)type).ParameterType);
             }
+            else if (type.IsFunctionPointer)
+            {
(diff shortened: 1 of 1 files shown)


src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Runtime/TypeLoaderExports.cs
```diff
@@ -392,48 +392,34 @@ private static Entry[] ResizeCacheForNewEntryAsNecessary()
     [ReflectionBlocked]
     public delegate IntPtr RuntimeObjectFactory(IntPtr context, IntPtr signature, object contextObject, ref IntPtr auxResult);
 
-    [System.Runtime.InteropServices.McgIntrinsicsAttribute]
-    internal class RawCalliHelper
+    internal unsafe class RawCalliHelper
```
jkotas, line 395: We should be able to replace `RawCalliHelper` with CallConv attribute on the function itself to clean this up further.

MichalStrehovsky: Oh, yes. The C# compiler doesn't care about the names of calling conventions - we can make a new CallConv type and the C# compiler will just respect that. Neat!

MichalStrehovsky (approved):
Thank you!

PR #210, opened 2020-10-12
Fix issue in LayoutClassPtr and Char marshalling

MichalStrehovsky:
Needed a break from triaging Pri0 failures so decided to fix one of the tests. It hit two bugs.


jkotas (approved):
Thanks!

PR #212, opened 2020-10-12
Add Char marshaller

AaronRobinsonMSFT:
`char` Marshaller.
Avoid supporting the built-in semantics for ANSI marshalling of char.
Propagate the `CharSet` down to the marshallers.
Convert the MarshallingGenerators.TryCreate() to a MarshallingGenerators.Create() function throwing an exception.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
+++ b/DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
@@ -150,6 +150,12 @@ namespace Microsoft.Interop
 
                 case { ManagedType: { SpecialType: SpecialType.System_Char }, MarshallingAttributeInfo: null }:
                     return Utf16Char; // [Compat] Default marshalling is UTF-16.
+                case { ManagedType: { SpecialType: SpecialType.System_Char }, MarshallingAttributeInfo: MarshalAsInfo { UnmanagedType: UnmanagedType.I2 } }:
+                    return Utf16Char;
+                case { ManagedType: { SpecialType: SpecialType.System_Char }, MarshallingAttributeInfo: MarshalAsInfo { UnmanagedType: UnmanagedType.U2 } }:
+                    return Utf16Char;
+                case { ManagedType: { SpecialType: SpecialType.System_Char }, MarshallingAttributeInfo: MarshallingInfoStringSupport { CharEncoding: CharEncoding.Utf16 } }:
+                    return Utf16Char;
                 case { ManagedType: { SpecialType: SpecialType.System_Char }, MarshallingAttributeInfo: MarshallingInfoStringSupport { CharEncoding: CharEncoding.Utf8 } }:
                     throw new MarshallingNotSupportedException(info, context) // [Compat] See conversion from CharSet.Ansi to UTF-8.
                     {
@@ -160,8 +166,6 @@ namespace Microsoft.Interop
                     {
                         NotSupportedDetails = Resources.MarshallingCharAsCharSetAutoNotSupported
                     };
-                case { ManagedType: { SpecialType: SpecialType.System_Char }, MarshallingAttributeInfo: MarshallingInfoStringSupport { CharEncoding: CharEncoding.Utf16 } }:
-                    return Utf16Char;
 
                 case { ManagedType: { TypeKind: TypeKind.Delegate }, MarshallingAttributeInfo: null or MarshalAsInfo { UnmanagedType: UnmanagedType.FunctionPtr } }:
                     return Delegate;
--- a/DllImportGenerator/DllImportGenerator/Resources.resx
+++ b/DllImportGenerator/DllImportGenerator/Resources.resx
@@ -160,10 +160,10 @@
     <value>Type '{0}' has a 'GetPinnableReference' method but its native type does not support marshalling in scenarios where pinning is impossible</value>
   </data>
   <data name="MarshallingCharAsCharSetAnsiNotSupported" xml:space="preserve">
-    <value>Marshalling char with CharSet.Ansi is not supported. Instead, manually convert the char type to a single byte and pass to the source-generated P/Invoke.</value>
+    <value>Marshalling char with CharSet.Ansi is not supported. Instead, manually convert the char type to the desired byte representation and pass to the source-generated P/Invoke.</value>
   </data>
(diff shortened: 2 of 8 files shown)


DllImportGenerator/designs/Compatibility.md
```diff
@@ -0,0 +1,15 @@
...
+
+Documentation on compatibility guidance and the current state. The version headings act as a rolling delta between the previous version.
+
+## Version 1
+
+The focus of version 1 is to support `NetCoreApp`. This implies that anything not needed by `NetCoreApp` is subject to change.
+
+### `char` marshaller
+
+The marshalling of `char` as [`CharSet.Ansi`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.charset) will not be supported. The built-in system used the [system default Windows ANSI code page](https://docs.microsoft.com/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte) when on Windows and took the first byte of the UTF-8 encoding on non-Windows platforms. Both approaches are fundamnetally flawed and therefore not supported. If a single-byte character is expected to be marshalled it is left to the caller to convert a .NET `char` into a single `byte` prior to calling the native function.
```
elinor-fung, line 11:
```suggestion
The marshalling of `char` as [`CharSet.Ansi`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.charset) will not be supported. The built-in system used the [system default Windows ANSI code page](https://docs.microsoft.com/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte) when on Windows and took the first byte of the UTF-8 encoding on non-Windows platforms. Both approaches are fundamentally flawed and therefore not supported. If a single-byte character is expected to be marshalled it is left to the caller to convert a .NET `char` into a single `byte` prior to calling the native function.
```

DllImportGenerator/DllImportGenerator/DllImportStub.cs
```diff
@@ -131,18 +131,32 @@ public class GeneratedDllImportData
...
             }
 
+            // Compute the current default string encoding value.
+            var defaultEncoding = CharEncoding.Undefined;
+            if (dllImportData.IsUserDefined.HasFlag(DllImportMember.CharSet))
+            {
+                defaultEncoding = dllImportData.CharSet switch
+                {
+                    CharSet.Unicode => CharEncoding.Utf16,
+                    CharSet.Auto => CharEncoding.Utf16,
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: Shoot. You are right. This will need to be a runtime check. Let me see what I can do here.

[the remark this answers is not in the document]
jkoritzinsky: We may need to say that passing a `char` by value with `CharSet.Auto` isn't allowed, which I think would be reasonable for v1.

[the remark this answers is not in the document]
AaronRobinsonMSFT: Actually, this support isn't needed at this point. `NetCoreApp` doesn't need any `Ansi`, `None`, or `Auto` support so I am going to error out on all non-Unicode scenarios.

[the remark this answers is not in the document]
AaronRobinsonMSFT: I have updated the official issue with the need for this API - see https://github.com/dotnet/runtime/issues/43060.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -96,51 +137,54 @@ public static bool TryCreate(TypePositionInfo info, StubCodeContext context, out
...
+                    {
+                        NotSupportedDetails = Resources.MarshallingCharAsCharSetAnsiNotSupported
+                    };
+                case { ManagedType: { SpecialType: SpecialType.System_Char }, MarshallingAttributeInfo: MarshallingInfoStringSupport { CharEncoding: CharEncoding.PlatformDefined } }:
+                    throw new MarshallingNotSupportedException(info, context) // [Compat] See conversion of CharSet.Auto.
+                    {
+                        NotSupportedDetails = Resources.MarshallingCharAsCharSetAutoNotSupported
+                    };
+                case { ManagedType: { SpecialType: SpecialType.System_Char }, MarshallingAttributeInfo: MarshallingInfoStringSupport { CharEncoding: CharEncoding.Utf16 } }:
+                    return Utf16Char;
```
elinor-fung, line 164: If the `UnmanagedType` is explicitly set, shouldn't that take precedence over the `CharSet` such that `I1`/`U1` -> UTF-8 and `I2`/`U2` -> UTF-16?

elinor-fung: Also, do we care to explicitly not support / error if the `UnmanagedType` is not one of those?

DllImportGenerator/DllImportGenerator/Resources.resx
```diff
@@ -159,6 +159,12 @@
   <data name="GetPinnableReferenceShouldSupportAllocatingMarshallingFallbackMessage" xml:space="preserve">
     <value>Type '{0}' has a 'GetPinnableReference' method but its native type does not support marshalling in scenarios where pinning is impossible</value>
   </data>
+  <data name="MarshallingCharAsCharSetAnsiNotSupported" xml:space="preserve">
+    <value>Marshalling char with CharSet.Ansi is not supported. Instead, manually convert the char type to a single byte and pass to the source-generated P/Invoke.</value>
+  </data>
+  <data name="MarshallingCharAsCharSetAutoNotSupported" xml:space="preserve">
+    <value>Marshalling char with CharSet.Auto is not supported. Instead, manually convert the char type to a single byte and pass to the source-generated P/Invoke.</value>
```
elinor-fung, line 166: `a single byte` -> the desired byte representation? Not sure what exactly that message should be.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
AaronRobinsonMSFT: >Also, do we care to explicitly not support / error if the UnmanagedType is not one of those?

I don't think so. It indicates the `MarshalAs` is invalid for the parameter so I think that works.

elinor-fung (approved):

PR #213, opened 2020-10-12
Enable nullability checks across DllImportGenerator

### LINKED ISSUE #134: Enable nullability checking in DllImportGenerator project.
Today we have nullability partially enabled in the DllImportGenerator code and fully enabled in some of the project files in the experiment. We should update all of the projects to have the nullability annotations and warnings enabled.

jkoritzinsky:
Enable nullable everywhere in the project instead of our current patchwork enable/disable system.

Fixes #134

### DIFF
--- a/DllImportGenerator/DllImportGenerator.IntegrationTests/CharacterTests.cs
+++ b/DllImportGenerator/DllImportGenerator.IntegrationTests/CharacterTests.cs
@@ -1,0 +1,82 @@
+﻿using System.Collections.Generic;
+using System.Runtime.InteropServices;
+
+using Xunit;
+
+namespace DllImportGenerator.IntegrationTests
+{
+    partial class NativeExportsNE
+    {
+        [GeneratedDllImport(nameof(NativeExportsNE), EntryPoint = "unicode_return_as_uint")]
+        public static partial uint ReturnDefaultAsUInt(char input);
+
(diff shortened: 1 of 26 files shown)


DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -161,7 +161,7 @@ private static bool IsGeneratedDllImportAttribute(AttributeSyntax attrSyntaxMayb
 
                 // Found the GeneratedDllImport, but it has an error so report the error.
                 // This is most likely an issue with targeting an incorrect TFM.
-                if (attrData.AttributeClass.TypeKind == TypeKind.Error)
+                if (attrData.AttributeClass?.TypeKind is null or TypeKind.Error)
```
AaronRobinsonMSFT, line 164: I don't see how `AttributeClass` can ever be `null`. Perhaps in this case it makes more sense to `!`

AaronRobinsonMSFT (approved):
I am less than convinced any of this "nullability" logic is worth the effort. It is simply noise and increases metadata for no obvious win.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
jkoritzinsky: So I just confirmed with Fred that AttributeClass can be null in the case of errors when decoding from metadata.

PR #235, opened 2020-10-15
Prevent runtime shutdown while foreground threads keep running

Rattenkrieg:
Port of https://github.com/dotnet/corert/pull/8364
Fixes https://github.com/dotnet/corert/issues/8246

### DIFF
--- a/src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Threading/Thread.CoreRT.Windows.cs
+++ b/src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Threading/Thread.CoreRT.Windows.cs
@@ -133,7 +133,6 @@ namespace System.Threading
             if (currentThread != null)
             {
                 StopThread(currentThread);
-                currentThread.JoinInternal(0);
             }
         }
 


src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Threading/Thread.CoreRT.Windows.cs
```diff
@@ -126,23 +126,15 @@ private bool SetPriorityLive(ThreadPriority priority)
...
-                {
-                    state = _threadState;
-                    if ((state & (int)(ThreadState.Stopped | ThreadState.Aborted)) == 0)
-                    {
-                        SetThreadStateBit(ThreadState.Stopped);
-                        state = _threadState;
-                    }
-                }
+                StopThread(currentThread);
+                currentThread.JoinInternal(0);
```
jkotas, line 136: Why is the JoinInternal needed here?

Rattenkrieg: TLDR: I did it at 2AM 😂 
Probably I just followed the logic we have in Unix. I see that in Unix version wait object is managed and we need to signal it once we got that callback, at the same time on Windows signaling is done with OS handle by kernel mechanisms related to those that invoked our `OnThreadExit`. So yes, completely not needed.

jkotas (approved):
Looks great!

Issue #88, opened 2020-09-09
Define pattern for single-library experiments


ericstj:
Today single library experiments that don't change the runtime/shared-framework don't have a great pattern to follow for build / test.  We should decide how these should behave and create a pattern/template to follow.

ericstj:
@safern 
> Why not make all experiments instead fork from dotnet/runtime -- that would simplify things infrastructure wise.

@ericstj
 > I think that's very undesirable for single-standalone libraries. Building the entire runtime is pretty expensive, cloning and dealing with codeflow is as well. I guess if we aren't going to create a pattern for single libraries we should keep corefxlab around for those. Thoughts @jkotas?

@jkotas
> If there is enough single libraries experiments, I agree it makes sense to have a template optimized for them.

> The pattern for single libraries will need maintenance independent on whether it is in runtimelab or corefxlab. I do not think we would save anything in the long-run by keeping corefxlab alive for single libraries.

> > Building the entire runtime is pretty expensive, cloning and dealing with codeflow is as well.

> Yes and it is a problem for dotnet/runtime too where it affects a lot more people than it would affect in dotnet/runtimelab. Another option may be to just keep improving leaf-library development experience in dotnet/runtime.

> > I see https://github.com/dotnet/runtimelab/tree/DllImportGenerator has no CI nor build support.

> I expect that the DllImportGenerator will eventually need to start adding public runtime APIs to make progress and turn into a dotnet/runtime fork.

ericstj:
> Yes and it is a problem for dotnet/runtime too where it affects a lot more people than it would affect in dotnet/runtimelab. Another option may be to just keep improving leaf-library development experience in dotnet/runtime.

We'll do this regardless.   That's not actually the problem **here**.  The workflow for building a standalone package in dotnet/runtime is already pretty good thanks to @ViktorHofer.  You can clone and build the pkgproj for most packages.  The problem is that we don't have plumbing for YML to do this for PR and official builds.  Nor do we support such flexibility for testing.  I don't think we'd ever care about doing that sort of thing in dotnet/runtime, but we would here: assuming we still agree that this is the place for such experiments.

safern:
I think we can add specific infrastructure running tests and building against LKG runtime and ref assemblies consumed either directly from dotnet/runtime or from the SDK, and turn it on via a an MSBuild flag on that specific branch. 

This wouldn't be that hard I think given that we already build and run against ref and runtime packs built locally.

ericstj:
[...]

PR #238, opened 2020-10-16
 Add standalone template for experiments that don't need runtime features

safern:
Fixes: #88 

This is the template branch for new experiments to use in order to create an experiment that doesn't depend on runtime or framework features.

Make sure to ignore the first commit as that is just pulling all the arcade files into `eng/common`. 

I will add a subsequent PR to enable official builds and to add links within the repo in the documentation I added in the README file. 

If you think the branch name doesn't make sense, please feel free to suggest one, I'm not great and naming and `standalone-template` was the best I could come up with.

FYI: @pgovind @eerhardt @danmosemsft


README.md
```diff
@@ -1,35 +1,31 @@
-# .NET Runtime Lab
+# Standalone Experiments
 
-This repo is for experimentation and exploring new ideas that may or may not make it into the main [dotnet/runtime](https://github.com/dotnet/runtime) repo. [Encouraging .NET Runtime Experiments](https://github.com/dotnet/runtime/issues/35609) describes reasons that motivated creating this repository.
+These branch contains a template for standalone experiments, which means that experiments that are a library, which doesn't depend on runtime changes and doesn't need the overhead of having all the runtime, libraries and installer code, can use this minimal template.
```
jkotas, line 3:
```suggestion
This branch contains a template for standalone experiments, which means that experiments that are a library, which doesn't depend on runtime changes and doesn't need the overhead of having all the runtime, libraries and installer code, can use this minimal template.
```

jkotas (approved):
LGTM. It may be a good idea to get somebody intimately familiar with Arcade, etc. to review the build scripts.

safern:
I believe, the guys I tagged + @ViktorHofer would be very familiar and could provide good feedback for this. Thanks, @jkotas

eng/Build.props
```diff
@@ -0,0 +1,5 @@
+<Project>
+  <ItemGroup>
+    <ProjectToBuild Include="$(RepoRoot)Experiment.sln" />
```
joperezr, line 3: NIT: I think you theoretically don't need this file as Arcade by default will build a solution on the root of the repo if it only finds one. At least that used to be the case.

eng/Versions.props
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <!-- Experiment package versions properties -->
+    <VersionPrefix>1.0.0</VersionPrefix>
+    <!-- File version numbers -->
+    <MajorVersion>1</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <PatchVersion>0</PatchVersion>
+    <PreReleaseVersionLabel>alpha</PreReleaseVersionLabel>
```
joperezr, line 10: We probably will want to avoid version clashes so my suggestion here would be to do the same we do in not standalone experiments and set the label as the experiment name instead.

joperezr: Granted its unlikely to have clashes with package names, but one other reason is to easily tell from which branch was a package produced when just looking at the feed.

global.json
```diff
@@ -0,0 +1,16 @@
+{
+  "sdk": {
+    "version": "5.0.100-rc.2.20479.15",
+    "allowPrerelease": true,
+    "rollForward": "major"
```
joperezr, line 5: Question: we did this for VS experience right?

src/Experiment.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
joperezr, line 4: Should this be 6.0?

eng/pipelines/runtimelab.yml
```diff
@@ -0,0 +1,74 @@
+# Setting batch to true, triggers one build at a time.
+# if there is a push while a build in progress, it will wait,
+# until the running build finishes, and produce a build with all the changes
+# that happened during the last build.
+trigger:
+  batch: true
+  branches:
+    include:
+    - feature/*
```
joperezr, line 9: Technically you don't need this one right?

joperezr (approved):
Left minor comments but looks great otherwise!

jkoritzinsky:
cc:@AaronRobinsonMSFT and @elinor-fung since we own the only experiment currently in this category to my knowledge.

src/Experiment.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
eerhardt, line 1: Should the folder structure be closer to the runtime folder structure?

-src 
    - System.Experiment
        - src /
            - System.Experiment.csproj
            - MyClass.cs
        -tests /

eng/Build.props
ViktorHofer: Correct: https://github.com/dotnet/arcade/blob/4f857c3c19e177cefdd85268151baf7bd8279422/src/Microsoft.DotNet.Arcade.Sdk/tools/Build.proj#L73. I would just remove the whole file.

src/Experiment.csproj
jkoritzinsky: No, we should not have to src directories. If we change it, it should be src/Experiment/Experiment.csproj

eng/testing/runsettings.targets
```diff
@@ -0,0 +1,40 @@
...
+  <Target Name="GenerateRunSettingsFile"
+          Inputs="$(RunSettingsInputFilePath)"
+          Outputs="$(RunSettingsOutputFilePath)">
+
+    <PropertyGroup>
+      <RunSettingsFileContent>$([System.IO.File]::ReadAllText('$(RunSettingsInputFilePath)'))</RunSettingsFileContent>
+      <RunSettingsFileContent>$(RunSettingsFileContent.Replace('$$RESULTSDIRECTORY$$', '$(ArtifactsTestResultsDir)')
+                                                      .Replace('$$SOLUTIONDIRECTORY$$', '$(RepoRoot)')
+                                                      .Replace('$$TARGETPLATFORM$$', '$(TargetPlatform)')
+                                                      .Replace('$$TRXRESULTSNAME$$', '$(ResultsFileName).xml')
```
ViktorHofer, line 24: TRX files should have the `.trx` extension.

ViktorHofer:
Curious, how do we manage changes from dotnet/runtime? Ie the testing infrastructure in this branch will soon be outdated.

tests/Directory.Build.props
```diff
@@ -0,0 +1,16 @@
+<Project>
+  <Import Project="..\Directory.Build.props" />
+
+  <PropertyGroup>
+    <IsPackable>false</IsPackable>
+    <UseVSTestRunner>true</UseVSTestRunner>
```
ViktorHofer, line 6: I believe this should be set in Versions.props instead: https://github.com/mono/linker/blob/e2ce67a42fac9915e764b52fcd15bbbb22576711/eng/Versions.props#L29

src/Experiment.csproj
eerhardt: We should at least establish a structure that supports more than a single library. While some experiments will only need a single library, I'm sure there will be plenty of experiments that will need multiple. (We need multiple in the DataFrame library)

ViktorHofer: Is there a specific reason to not have a tests folder per leaf?

global.json
ViktorHofer: We did this so that you can use a newer SDK when interacting with the repository.

eng/testing/runsettings.targets
```diff
@@ -0,0 +1,40 @@
+<Project InitialTargets="GenerateRunSettingsFile">
+
+  <PropertyGroup>
+    <RunSettingsInputFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsInputFilePath>
+    <RunSettingsAppOutputFilePath>$(OutDir).runsettings</RunSettingsAppOutputFilePath>
+
+    <RunSettingsOutputFilePath>$(RunSettingsAppOutputFilePath)</RunSettingsOutputFilePath>
+
+    <ResultsFileName>$(MSBuildProjectName)_$(TargetFramework)_$(TargetPlatform)</ResultsFileName>
```
ViktorHofer, line 9: Why do you need this? Do test projects in this repo not have a unique intermediate and output directory?

eng/Versions.props
safern: I left it this way since experiments will be in their own branch and the name of the package should tell which experiment it is... in the other kind of experiments since we produce a shared framework it makes sense, but in my opinion here I would just leave it to the experiment owner to use the label they want.

src/Experiment.csproj
pgovind: Or NetCoreAppCurrent?

eng/pipelines/runtimelab.yml
safern: Technically not, but I wanted to leave it like this so that people have to do the minimal set of changes when starting a new experiment.

src/Experiment.csproj
safern: If we wanted to do this I would have to add a workaround for the 5.0 SDK to not complain about a not supported TFM, so that's why I decided to leave with net5.0 for now.

src/Experiment.csproj
safern: Either way I'm fine. I'll change it to support multiple projects per experiment then.

tests/Directory.Build.props
safern: What is the difference?

eng/testing/runsettings.targets
safern: This is because we use arcade's test result directory, so I was following arcade's output convention when running `build.cmd -test`. I wanted to produce the same results file if you run that vs `dotnet run` in order to not have different results file.

But to answer your question we do have different intermediate and bin output per project but arcade puts test results under `artifacts/TestResults/<Configuration>`

tests/Directory.Build.props
ViktorHofer: Just a note that Arcade expects these feature switches to be set in Versions.props.

safern:
> Curious, how do we manage changes from dotnet/runtime? Ie the testing infrastructure in this branch will soon be outdated.

what do you mean with the testing infrastructure?

I just used the runsettings target logic, but we don't use the rest of the dotnet/runtime infrastructure, in fact this doesn't have anything from dotnet/runtime.

Experiments that do need a shared framework are forked off from `runtime-master` branch which is mirrored on every commit that goes into `dotnet/runtime`.

tests/Directory.Build.props
safern: I see, does it set versions based on that?

ViktorHofer:
> I just used the runsettings target logic, but we don't use the rest of the dotnet/runtime infrastructure, in fact this doesn't have anything from dotnet/runtime.

Right, I'm talking about the runsettings.targets and .runsettings files which I'm changing in my VSTest PR in dotnet/runtime. Who would be responsible to update them here as well, ie to benefit from improvements in the VSTest runner?

src/Experiment.csproj
safern: NetCoreAppCurrent doesn't exist on this branch. We could add it though, but given the small number of projects I would imagine a branch would have I don't know if it makes sense.

eng/common/SetupNugetSources.ps1
```diff
@@ -0,0 +1,160 @@
+# This file is a temporary workaround for internal builds to be able to restore from private AzDO feeds.
+# This file should be removed as part of this issue: https://github.com/dotnet/arcade/issues/4080
```
pgovind, line 2: https://github.com/dotnet/arcade/issues/3932 seems to be closed now?

safern:
That would be me or @joperezr I guess... Do you expect this target to change often?

eng/common/SetupNugetSources.ps1
safern: This file comes from arcade. Everything under eng/common is literally a copy of the latest arcade commit.

Should we instead open an issue in arcade asking that?

Directory.Build.props
```diff
@@ -0,0 +1,18 @@
+<Project>
+  <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+
+  <PropertyGroup>
+    <Copyright>$(CopyrightNetFoundation)</Copyright>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
```
pgovind, line 6: Should this also have the property for signed packages? (`SignAssembly` IIRC?)

Directory.Build.props
```diff
@@ -0,0 +1,18 @@
...
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
+    <DebugType>embedded</DebugType>
+    <DebugSymbols>true</DebugSymbols>
```
pgovind, line 9: I thought this was `<IncludeSymbols>`? Maybe they mean the same thing. Just checking :)

Directory.Build.props
```diff
@@ -0,0 +1,18 @@
...
+    <LangVersion>Latest</LangVersion>
+    <IsPackable>true</IsPackable>
```
pgovind, line 11: Not sure if the plan is currently to move all experiments from corefxlab to runtimelab. If it is, this property defaults to false in corefxlab. Otherwise, ignore this comment!

Directory.Build.props
```diff
@@ -0,0 +1,18 @@
...
+
```
pgovind, line 12: Just a note that it might be worth it to check a project's Nuget package after it is built for 
1. Author
2. Project URL

Directory.Build.props
safern: I set this to true here and then test set it to false. In corefxlab we have more projects and we wanted to intentionally enable it because not al projects are packable, here I expect the majority to be, so I think it makes sense to default to true.

pgovind (approved):
Just some minor comments. Looks good to go in for now. We can always make changes as new csprojs get added

eng/testing/runsettings.targets
safern: Arcade sets `.xml` extension to them. I'm doing this to get the same file extension. I'll open an issue in arcade.

ViktorHofer: Yes, see our discussion here: https://github.com/dotnet/arcade/pull/6355#issuecomment-708305947.

safern: I opened: https://github.com/dotnet/arcade/issues/6419

ViktorHofer:
> That would be me or @joperezr I guess... Do you expect this target to change often?

If you use Arcade's test infrastructure why do you need the runsettings file and its target file at all?

eng/common/SetupNugetSources.ps1
pgovind: Nah, if it's internal to arcade, I don't care about it that much given that most folks are not going to see this!

safern:
> If you use Arcade's test infrastructure why do you need the runsettings file and its target file at all?

That is used when people want to use a custom runtime version... so in order to support running with `dotnet test`. Basically I wanted to be able to point vstest to the repo SDK `.dotnet` which contains the runtime that the branch needs... see this in the docs I'm adding: https://github.com/dotnet/runtimelab/blob/a8052dd1c4fb4625d5b684535bdc4f6e42d5030e/README.md#create-your-experiment

> I do need to depend on dotnet/runtime
> Note that if you want to run dotnet test in you test projects you will need to either first run build.cmd/sh or install the runtime version specified by MicrosoftNETCoreAppVersion property in Versions.props in your global dotnet install. If you run build.cmd/sh arcade infrastructure will make sure that the repo dotnet SDK found in <RepoRoot>\.dotnet folder, has this runtime installed. Then during the build of the test projects, we generate a .runsettings file that points to this dotnet SDK.

ViktorHofer:
> That is used when people want to use a custom runtime version... so in order to support running with dotnet test.

You could pass that information in via a CLI switch as Arcade's `Test` target should allow specifying additional CLI options. Or is this to support VS Test Explorer as well? Did you try if that works? I would assume that you need to generate an initial .runsettings file (same as what we do in pretests.proj) that is used, before the test app is built.

safern:
> You could pass that information in via a CLI switch as Arcade's Test target should allow specifying additional CLI options. Or is this to support VS Test Explorer as well? Did you try if that works? I would assume that you need to generate an initial .runsettings file (same as what we do in pretests.proj) that is used, before the test app is built.

Yes, this is also to support `VS Test Explorer` when using a custom runtime version other than the one that comes with the SDK specified in `global.json`. 

Yes I did try this on a clean build by opening the solution... the way I did it was by setting the `GenerateRunSettingsFile` as an `InitialTarget` in test projects, so it will produce it first thing no matter what target is ran.

ViktorHofer:
> Yes I did try this on a clean build by opening the solution... the way I did it was by setting the GenerateRunSettingsFile as an InitialTarget in test projects, so it will produce it first thing no matter what target is ran.

When I added the VS Test Explorer support in dotnet/runtime, this didn't work. VS Test Explorer tried to load the .runsettings file specified in the property immediately (before a target ran) and failed. Keep an eye out for that to be certain that the right .runsettings file and SDK is used.

safern:
> When I added the VS Test Explorer support in dotnet/runtime, this didn't work. VS Test Explorer tried to load the .runsettings file specified in the property immediately (before a target ran) and failed. Keep an eye out for that to be certain that the right .runsettings file and SDK is used.

I'll double check that. Thanks for the heads up.

jkoritzinsky:
So after looking through this from the consuming side, it seems significantly more complicated than what DllImportGenerator (which is the only current standalone project) would need. The inclusion of the platform-matrix.yml style of templating is significantly more complex than what we need for our project. We were planning on writing a simple yml entry-point that's easy to understand without multiple template layers on top of the Arcade ones. We don't need any of the complicated dotnet/runtime pipeline infrastructure for our build.

In fact, a simple `matrix:` object on top of the basic Arcade templates would satisfy our use case completely.

I'm a little concerned that this pipeline was designed without any input from the only standalone project today that might use it (I stumbled on this PR randomly, no one on the Interop team working on the DllImportGenerator project was tagged). I'm also worried that since we might not have any standalone projects using it for a while, it might atrophy and break before anyone adopts it.

I think if the .NET Runtime Infra v-team (preferably someone other than myself) commits to owning this infra and maintaining it in experiments like the DllImportGenerator experiment, I'll be more amenable to this design.

eerhardt:
> which is the only current standalone project

The Microsoft.Data.Analysis package from corefxlab will also be using this template.

safern:
> I'm a little concerned that this pipeline was designed without any input from the only standalone project today that might use it (I stumbled on this PR randomly, no one on the Interop team working on the DllImportGenerator project was tagged). I'm also worried that since we might not have any standalone projects using it for a while, it might atrophy and break before anyone adopts it.

Sorry that I didn't tag you guys. My understanding was that the `DllImportGenerator` was going to evolve into a experiment that needed runtime features, and wasn't going to use this template, that's why I didn't tag you here.

I will reach out offline, and this are valid concerns, so I will reach offline and get consensus on this from the interop team to meet your guys need as well.

jkoritzinsky:
Yes, we'll eventually move to getting runtime features, but that's planned for closer to the end of the experiment. Many of the APIs we need we can initially provide (with less performance) outside of the runtime. We don't want to fork the runtime into our branch (and significantly complicate our workflow) until we absolutely cannot avoid it.

Directory.Build.props
safern: I believe that is on by default with arcade.

Directory.Build.props
safern: Yeah, these are set by arcade. Just doubled checked and it makes sense.

safern:
Thanks for the discussion and reviews. I've addressed/validated all concerns. This is ready to go, could you guys take another look please?

jkoritzinsky (approved):

PR #239, opened 2020-10-16
Fix behavior of interface dispatch on null instances

### LINKED ISSUE #221: NullReferenceException not thrown for nullref in RhpInterfaceDispatch2
Loader/classloader/methodoverriding/regressions/549411/exploit

We tear down the process instead.

MichalStrehovsky:
An interface call with a null `this` needs to throw a `NullReferenceException`. There are two ways to achieve that: either codegen can insert explicit code to trigger a NullRef at the callsite (before calling into the runtime library to do the dispatch), or the runtime library needs to be able to unwind the helper code that tries to resolve the dispatch and blame the callsite. .NET Native did the former, but the latter produces smaller code (no need to insert the checks). The CoreRT runtime doesn't have code to handle the latter. This pull request is adding that.

The fix is relatively straightforward - we mark locations in the helper where AV can occur. If an AV occurs outside managed code, we check whether it's one of the known locations. If so, we remove the helper method frame off the stack and pretend the AV happened in the calling code.

The one-liner fix in CorInfoImpl fixes an unintialized value problem that basically made us roll a dice on whether to do an explicit nullcheck at the callsite.

Fixes #221.

### DIFF
--- a/src/coreclr/src/nativeaot/Runtime/EHHelpers.cpp
+++ b/src/coreclr/src/nativeaot/Runtime/EHHelpers.cpp
@@ -330,32 +330,6 @@ static bool InWriteBarrierHelper(UIntNative faultingIP)
     return false;
 }
 
-static UIntNative UnwindSimpleHelperToCaller(
-#ifdef TARGET_UNIX
-    PAL_LIMITED_CONTEXT * pContext
-#else
-    _CONTEXT * pContext
-#endif
-    )
-{
-#if defined(_DEBUG)
-    UIntNative faultingIP = pContext->GetIp();
-    ASSERT(InWriteBarrierHelper(faultingIP) || InInterfaceDispatchHelper(faultingIP));
-#endif
-#if defined(HOST_AMD64) || defined(HOST_X86)
-    // simulate a ret instruction
-    UIntNative sp = pContext->GetSp();
-    UIntNative adjustedFaultingIP = *(UIntNative *)sp;
-    pContext->SetSp(sp+sizeof(UIntNative)); // pop the stack
-#elif defined(HOST_ARM) || defined(HOST_ARM64)
-    UIntNative adjustedFaultingIP = pContext->GetLr();
-#else
-    UIntNative adjustedFaultingIP = 0; // initializing to make the compiler happy
-    PORTABILITY_ASSERT("UnwindSimpleHelperToCaller");
-#endif
-    return adjustedFaultingIP;
-}
-
 EXTERN_C void* RhpInitialInterfaceDispatch;
 EXTERN_C void* RhpInterfaceDispatchAVLocation1;
 EXTERN_C void* RhpInterfaceDispatchAVLocation2;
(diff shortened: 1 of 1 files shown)


src/coreclr/src/nativeaot/Runtime/EHHelpers.cpp
```diff
@@ -390,6 +452,39 @@ Int32 __stdcall RhpHardwareExceptionHandler(UIntNative faultCode, UIntNative fau
...
+        // If this was an AV and code manager is null, this was an AV in unmanaged code.
+        // Could still be an AV in one of our assembly helpers that we know how to handle.
+        bool inWriteBarrierHelper = InWriteBarrierHelper(faultingIP);
+        bool inInterfaceDispatchHelper = InInterfaceDispatchHelper(faultingIP);
+
+        if (inWriteBarrierHelper || inInterfaceDispatchHelper)
+        {
+            if (faultAddress < NULL_AREA_SIZE)
+            {
+                faultCode = STATUS_REDHAWK_WRITE_BARRIER_NULL_REFERENCE;
```
MichalStrehovsky, line 468: I'm considering to renaming this to something else, but don't have a good name.

src/coreclr/src/nativeaot/Runtime/EHHelpers.cpp
```diff
@@ -356,6 +356,73 @@ static UIntNative UnwindWriteBarrierToCaller(
...
+
+        if (interfaceDispatchAVLocations[i] == faultingIP)
+            return true;
+    }
+#endif // USE_PORTABLE_HELPERS
+
+    return false;
+}
+
+static UIntNative UnwindInterfaceDispatchToCaller(
```
MichalStrehovsky, line 400: This is the exact same code as the write barriers unwinder. I could just rename that one to a common name and call it.

I just wasn't sure whether this code could be different for other arches since here we have an unknown number of parameters. It happens to not be a problem on x64. I don't know about x86.

src/coreclr/src/nativeaot/Runtime/amd64/StubDispatch.asm
```diff
@@ -85,6 +91,11 @@ LEAF_END RhpVTableOffsetDispatch, _TEXT
 ;; Initial dispatch on an interface when we don't have a cache yet.
 LEAF_ENTRY RhpInitialInterfaceDispatch, _TEXT
 ALTERNATE_ENTRY RhpInitialDynamicInterfaceDispatch
+        ;; Trigger an AV if we're dispatching on a null this.
+        ;; The exception handling infrastructure is aware of the fact that this is the first
+        ;; instruction of RhpInitialInterfaceDispatch and uses it to translate an AV here
+        ;; to a NullReferenceException at the callsite.
+        cmp     qword ptr [rcx], 0
```
MichalStrehovsky, line 98: This is technically not necessary since the AV will happen in Runtime.Base, but it's going to throw the Runtime.Base's NullReferenceException that would not be catchable in the original Redhawk model.

src/coreclr/src/nativeaot/Runtime/EHHelpers.cpp
Suchiman: How about one of?
`STATUS_REDHAWK_HELPER_CODE_NULL_REFERENCE`
`STATUS_REDHAWK_NATIVE_HELPER_NULL_REFERENCE`
`STATUS_REDHAWK_UNMANAGED_HELPER_NULL_REFERENCE`
`STATUS_REDHAWK_MANUALLY_MANAGED_NULL_REFERENCE`

src/coreclr/src/nativeaot/Runtime/EHHelpers.cpp
jkotas: x86 is full of problems that none of the other platforms have. I think having the unwinder shared would be a good idea.

jkotas: (If we ever do x86 for completeness, we may want to do it using cdecl callconv to avoid all special casing.)

src/coreclr/src/nativeaot/Runtime/amd64/StubDispatch.S
```diff
@@ -64,6 +69,11 @@ LEAF_END RhpVTableOffsetDispatch, _TEXT
 // Initial dispatch on an interface when we don't have a cache yet.
 LEAF_ENTRY RhpInitialInterfaceDispatch, _TEXT
 ALTERNATE_ENTRY RhpInitialDynamicInterfaceDispatch
+        // Trigger an AV if we're dispatching on a null this.
```
jkotas, line 72: Nit: It may be nice to add these to the ARM64 ones too while you are on it.

src/coreclr/src/nativeaot/Runtime/EHHelpers.cpp
jkotas: Or just `STATUS_REDHAWK_NULL_REFERENCE` ?

MichalStrehovsky: STATUS_REDHAWK_NULL_REFERENCE is already taken and apparently we need to be able to distinguish between these two cases in the frame iterator.

I like STATUS_REDHAWK_UNMANAGED_HELPER_NULL_REFERENCE. Thanks!

jkotas (approved):

Issue #199, opened 2020-10-11
Assertion failed during 'Mark local vars'


MichalStrehovsky:
Assertion failed '(varDsc->lvType != TYP_UNDEF) && (varDsc->lvType != TYP_VOID) && (varDsc->lvType != TYP_UNKNOWN)' in 'FullProof:Test():int' during 'Mark local vars'

JIT/Regression/VS-ia64-JIT/V2.0-Beta2/b309576/b309576

MichalStrehovsky:
This also reproes with CoreCLR. The test in question has a hardcoded `<Optimize>true</Optimize>`. If we remove that, this asserts with corerun too. Looks like a JIT bug.

[jitdump.txt](https://github.com/dotnet/runtimelab/files/5391215/jitdump.txt)

PR #246, opened 2020-10-19
Remove mention of #199

MichalStrehovsky:
I think this is not needed after #211.

There still is a JIT bug or too strong assert, but that's tracked in the runtime repo.


jkotas (approved):

MichalStrehovsky:
```
  Assertion failed 'op1 != nullptr' in 'TestUtil.TestLog:VerifyOutput():int:this' during 'Morph - Global' (IL size 130)

Total failed tests : 79
```

Well that's new...

PR #247, opened 2020-10-19
Add project for LLVM backend (supports Wasm)

yowl:


src/libraries/Directory.Build.props
```diff
@@ -124,10 +124,11 @@
...
     <!-- There are no WebAssembly tools, so use the default ones -->
     <_buildingInOSX>$([MSBuild]::IsOSPlatform('OSX'))</_buildingInOSX>
-    <ToolRuntimeRID Condition="'$(RuntimeOS)' == 'browser' and '$(TargetOS)' == 'windows'">win-x64</ToolRuntimeRID>
+    <_buildingInWindows>$([MSBuild]::IsOSPlatform('windows'))</_buildingInWindows>
+    <ToolRuntimeRID Condition="'$(RuntimeOS)' == 'browser' and $(_buildingInWindows)">win-x64</ToolRuntimeRID>
     <ToolRuntimeRID Condition="'$(RuntimeOS)' == 'browser' and '$(TargetOS)' != 'windows' and $(_buildingInOSX)">osx-x64</ToolRuntimeRID>
-    <ToolRuntimeRID Condition="'$(RuntimeOS)' == 'browser' and '$(TargetOS)' != 'windows' and !$(_buildingInOSX)">linux-x64</ToolRuntimeRID>
-
+<!-- this line seems to cause a failure with _SuppressSdkImports .  Doesn't make sense to me -->
+<!--    <ToolRuntimeRID Condition="'$(RuntimeOS)' == 'browser' and '$(TargetOS)' != 'windows' and !$(_buildingInOSX)" and !$(_buildingInWindows)">linux-x64</ToolRuntimeRID> -->
```
yowl, line 131: I don't understand this.  The condition shouldn't even be met as `buildingInWindows` is true so why it causes an error I don't know.

yowl:
I'm going to release this now for review.  Sorry it's rather long.  If you want to know what each change fixes I can revert it and show the error.  It will publish and run the Samples\HelloWorld (with https://github.com/dotnet/runtimelab/pull/247/files#diff-95c7fc3099f2a4f95cd78e2aae618395f1a7aef4e88e34d30d957f4814bf688e) and that's as far as I've taken it.  Once it goes in I could either:
- Add yml for CI
- Get the old tests passing - HelloWasm/Exceptions/SimpleGenerics

src/libraries/Directory.Build.props
jkotas: I guess it may a problem with evaluation order.

jkotas (approved):

jkotas:
Could you please also update pre-release label as described in https://github.com/dotnet/runtimelab/blob/master/CreateAnExperiment.md so that the packages do not clash?

eng/native/configurecompiler.cmake
```diff
@@ -218,6 +218,14 @@ elseif (CLR_CMAKE_HOST_ARCH_ARM64)
...
   add_definitions(-DHOST_ARM64)
   add_definitions(-DHOST_64BIT)
+elseif (CLR_CMAKE_TARGET_ARCH_WASM)
+  set(ARCH_HOST_NAME wasm)
+  add_definitions(-DHOST_WASM=1)
+  add_definitions(-DFEATURE_64BIT_ALIGNMENT=1)
+  # The -fms-extensions enable the stuff like __if_exists, __declspec(uuid()), etc.
+  add_compile_options(-fms-extensions)
+  add_compile_options(-Wno-invalid-offsetof)
+  add_compile_options(-Wno-tautological-undefined-compare) # this == NULL warning suppression
```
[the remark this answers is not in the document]
yowl: `ms-extensions` I believe is preventing 
```
In file included from E:\GitHub\runtimelab\src\coreclr\src\nativeaot\Runtime\allocheap.cpp:9:
         E:/GitHub/runtimelab/src/coreclr/src/nativeaot/Runtime/./PalRedhawk.h:658:1: error: unknown type name '__forceinline'
         __forceinline
         ^
         In file included from E:\GitHub\runtimelab\src\coreclr\src\nativeaot\Runtime\allocheap.cpp:9:
         In file included from E:/GitHub/runtimelab/src/coreclr/src/nativeaot/Runtime/./PalRedhawk.h:873:
```
`-Wno-invalid-offsetof` is stopping:
```
        In file included from E:\GitHub\runtimelab\src\coreclr\src\nativeaot\Runtime\allocheap.cpp:21:
         E:/GitHub/runtimelab/src/coreclr/src/nativeaot/Runtime/./slist.inl:41:54: warning: offset of on non-standard-layout type 'AllocHeap::BlockListElem' [-Winvalid-offsetof]
             return dac_cast<PTR_PTR_T>(dac_cast<TADDR>(pT) + offsetof(T, m_pNext));
                                                              ^           ~~~~~~~
         E:\GitHub\emsdk\upstream\emscripten\system\include\libc\stddef.h:20:32: note: expanded from macro 'offsetof'
         #define offsetof(type, member) __builtin_offsetof(type, member)
                                        ^                        ~~~~~~
         E:/GitHub/runtimelab/src/coreclr/src/nativeaot/Runtime/./slist.inl:245:14: note: in instantiation of member function 'DefaultSListTraits<AllocHeap::BlockListElem, DoNothingFailFastPolicy>::GetNextPtr' requested here
             *Traits::GetNextPtr(pItem) = *m_ppCur;
                      ^
         E:/GitHub/runtimelab/src/coreclr/src/nativeaot/Runtime/./slist.inl:85:13: note: in instantiation of member function 'SList<AllocHeap::BlockListElem, DefaultSListTraits<AllocHeap::BlockListElem, DoNothingFailFastPolicy>>::Iterator::Insert' requested here
             Begin().Insert(pItem);
                     ^
         E:\GitHub\runtimelab\src\coreclr\src\nativeaot\Runtime\allocheap.cpp:100:17: note: in instantiation of member function 'SList<AllocHeap::BlockListElem, DefaultSListTraits<AllocHeap::BlockListElem, DoNothingFailFastPolicy>>::PushHead' requested here
             m_blockList.PushHead(pBlock);
                         ^
         1 warning generated.
```
`-Wno-tautological-undefined-compare` is for:
```
         E:\GitHub\runtimelab\src\coreclr\src\nativeaot\Runtime\eetype.cpp:24:9: warning: 'this' pointer cannot be null in well-defined C++ code; comparison may be assumed to always evaluate to false [-Wtautological-undefined-compare]
             if (this == NULL)
                 ^~~~    ~~~~
         1 warning generated.
```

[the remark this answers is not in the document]
jkotas: I see. These and warning options are in this file under `CLR_CMAKE_HOST_UNIX`. Would it make sense to change the condition on that block to !MSVC or something similar instead of duplicating them here?

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
```diff
@@ -52,7 +52,6 @@
     <Compile Include="$(MSBuildThisFileDirectory)System\Activator.RuntimeType.cs" Condition="'$(TargetsCoreRT)' != 'true'" />
     <Compile Include="$(MSBuildThisFileDirectory)System\AggregateException.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\AppContext.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)System\AppContext.Browser.cs" Condition="'$(TargetsBrowser)' == 'true'" />
```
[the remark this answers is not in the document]
yowl: Its also here https://github.com/dotnet/runtimelab/blob/82dd290fb057b8c8ae648736ca1c69ed8e44c3cf/src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems#L1871  .  Seems like a duplicate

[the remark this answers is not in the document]
jkotas: I see. Another general goodness change :-)

src/libraries/Directory.Build.props
```diff
@@ -256,4 +256,7 @@
     <CLSCompliant Condition="'$(CLSCompliant)' == ''">true</CLSCompliant>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('windows')) and '$(TargetOS)' == 'Browser'"> <!-- the Unix versions are used without this override -->
```
jkotas, line 259: Is the actual problem a bug in `eng\Configurations.props`?

I think the block in Configurations.props should be something like:
```
    <!-- There are no WebAssembly tools, so use the default ones -->
    <_toolRuntimeRID Condition="'$(_runtimeOS)' == 'browser'">linux-x64</_toolRuntimeRID>
    <_toolRuntimeRID Condition="'$(_runtimeOS)' == 'browser' and $([MSBuild]::IsOSPlatform('WINDOWS'))">win-x64</_toolRuntimeRID>
    <_toolRuntimeRID Condition="'$(_runtimeOS)' == 'browser' and $([MSBuild]::IsOSPlatform('OSX'))">osx-x64</_toolRuntimeRID>
```

eng/Versions.props
```diff
@@ -6,7 +6,7 @@
     <MajorVersion>6</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreReleaseVersionLabel>alpha</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>llvm</PreReleaseVersionLabel>
```
jkotas, line 9: I have realized that this won't be sufficient to avoid collisions given how nuget works and how things are setup currently.

The problem is what `PackageReference Include="Microsoft.DotNet.ILCompiler" Version="6.0.0-*"` package reference is going to pick up the latest `*` that will be `llvm` once this starts publishing.

I think it will be best to use different package name for this branch, like `Microsoft.DotNet.ILCompiler.LLVM`. 

Could you please rename `Microsoft.DotNet.ILCompiler` to `Microsoft.DotNet.ILCompiler.LLVM` under `src\installer\pkg\projects` ? (Both directory names and references in sources.)

There may be a few other places througout the tree that may need similar rename.

src/libraries/Directory.Build.props
yowl: Merging the latest got rid of this

eng/native/build-commons.ps1
```diff
@@ -0,0 +1,21 @@
+if ( -not (Test-Path -LiteralPath $Env:__versionSourceFile))
```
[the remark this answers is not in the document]
yowl: Deleted.  Whatever reason I had has gone.

eng/native/configurecompiler.cmake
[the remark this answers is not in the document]
yowl: Thanks, move as suggested to not duplicate.

eng/native/gen-buildsys.cmd
```diff
@@ -24,12 +24,13 @@ set __Arch=%4
 set __CmakeGenerator=Visual Studio
 set __UseEmcmake=0
 if /i "%__Ninja%" == "1" (
+	echo using Ninja
```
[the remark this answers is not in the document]
yowl: Thanks, missed this one, deleted.

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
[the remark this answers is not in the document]
yowl: merged upstream

src/libraries/Native/build-native.cmd
```diff
@@ -97,7 +97,7 @@ goto :SetupDirs
 echo Commencing build of native components
 echo.
 
-if /i "%__BuildArch%" == "wasm" set __sourceDir=%~dp0..\Unix
+if /i "%__BuildArch%" == "wasm" set __sourceDir=%~dp0\Unix
```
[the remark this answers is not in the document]
yowl: merged upstream

src/libraries/Directory.Build.props
yowl: Thanks, that looks better.  Admittedly I don't know my way around the build files.

eng/Versions.props
yowl: Sure, is the label change now superfluous ?

jkotas: Yes - if you change the package name, it is not necessary to change the label.

eng/Configurations.props
```diff
@@ -123,9 +123,9 @@
     <_toolRuntimeRID Condition="'$(_runtimeOS)' == 'linux-musl' and $(TargetArchitecture.StartsWith('arm')) and !$(_hostArch.StartsWith('arm'))">linux-x64</_toolRuntimeRID>
 
     <!-- There are no WebAssembly tools, so use the default ones -->
-    <_toolRuntimeRID Condition="'$(_runtimeOS)' == 'browser' and '$(TargetOS)' == 'windows'">win-x64</_toolRuntimeRID>
-    <_toolRuntimeRID Condition="'$(_runtimeOS)' == 'browser' and '$(TargetOS)' != 'windows' and $(_buildingInOSX)">osx-x64</_toolRuntimeRID>
-    <_toolRuntimeRID Condition="'$(_runtimeOS)' == 'browser' and '$(TargetOS)' != 'windows' and !$(_buildingInOSX)">linux-x64</_toolRuntimeRID>
+    <_toolRuntimeRID Condition="'$(_runtimeOS)' == 'browser'">linux-x64</_toolRuntimeRID>
```
jkotas, line 126: If it works, it would be nice to submit this to dotnet/runtime. Also, the Android block below should get the same fix.

src/libraries/Native/build-native.cmd
```diff
@@ -158,7 +158,8 @@ set __generatorArgs=
 if [%__Ninja%] == [1] (
     set __generatorArgs=
 ) else if [%__BuildArch%] == [wasm] (
-    set __generatorArgs=-j
+    rem nmake does support -j
```
jkotas, line 161: I think this can also be done in dotnet/runtime

src/libraries/Directory.Build.props
am11: This commit was made before the configurations were even moved to eng/Configurations.props. And it is/was not a bug.

The new thing this PR has introduced is `TargetOS=browser` set from cmd file, we typically do not add TargetOS in cmd/sh for cross-compilation scenarios, where it has specific meaning. If we change cmd/sh files here with new `TargetOS=browser` introducing new meaning, sooner or later that will clash with mono, where we are using common paths in scripts. It is anyway confusing.

If we do not introduce TargetOS=browser inside the script and instead take mono's lead on making 'is browser' decision, you would be able to take benefit of many common work without configuration changes over time in a smooth way.

jkotas: This condition should be using the OS that the build is running on.

TargetOS should mean the OS we are compiling for, everywhere in the build system. If TargetOS means the OS that the build is running in some places, I think we should fix those place.

> mono's lead on making 'is browser' decision

I am not sure what this means.

am11: TargetOS is not `browser` on mono when we are building for wasm today (it has been this way all along). If we intend to change that as a fix, OK.

jkotas: > TargetOS is not browser on mono when we are building for wasm today

Is it really the case? There are number of places under libraries that compare TargetOS with Browser, one example from many: https://github.com/dotnet/runtime/blob/master/src/libraries/System.Diagnostics.StackTrace/tests/System.Diagnostics.StackTrace.Tests.csproj#L16 

If the TargetOS was not set correct for Browser, all these places would be broken.

In any case, I think it is a good idea to keep fixing places that use wrong meaning of TargetOS.

eng/Versions.props
yowl: Maybe I've gone overboard on this, I renamed the targets file and the pkgproj.

jkotas (approved):
LGTM. Thank you!

yowl:
Thanks for all the help.

PR #250, opened 2020-10-20
Add helper class to read trimming dataflow annotations

MichalStrehovsky:
FlowAnnotations.cs corresponds to https://github.com/mono/linker/blob/master/src/linker/Linker.Dataflow/FlowAnnotations.cs.

This is a helper class that makes it possible to query for the new .NET 5 dataflow annotations within the compiler. We also implement warnings/validations when these annotations are incorrect.

This is mostly a copy-paste of FlowAnnotations from linker with tweaks to make it compile - the warning messages and error codes are identical. We want them to be identical to make suppressions work later.

### DIFF
--- a/src/coreclr/src/tools/Common/Compiler/Logger.cs
+++ b/src/coreclr/src/tools/Common/Compiler/Logger.cs
@@ -3,6 +3,8 @@
 
 using System.IO;
 
+using Internal.TypeSystem;
+
 namespace ILCompiler
 {
     // Poor man's logger. We can do better than this.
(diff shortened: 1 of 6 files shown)


src/coreclr/src/tools/Common/Compiler/Logger.cs
```diff
@@ -20,5 +22,18 @@ public Logger(TextWriter writer, bool isVerbose)
             Writer = TextWriter.Synchronized(writer);
             IsVerbose = isVerbose;
         }
+
+        public void LogWarning(string text, int code, TypeSystemEntity origin, int? ilOffset = null, string subcategory = MessageSubCategory.None)
+        {
+            // Temporary implementaton. We'll want to mirror warning infrastructure in the IL linker.
```
jkotas, line 28:
```suggestion
            // Temporary implementation. We'll want to mirror warning infrastructure in the IL linker.
```

jkotas:
> This is mostly a copy-paste of FlowAnnotations from linker with tweaks to make it compile 

Would it be useful to have the verbatim copy of the file in the history with the edits on top of it?

jkotas (approved):
Thanks!

MichalStrehovsky:
> Would it be useful to have the verbatim copy of the file in the history with the edits on top of it?

Verbatim, with weirdo-formatting-rules-of-the-Mono-team removed :). Yup, I can do that.

I was even thinking of checking in the original source (including weirdo formatting) into a ReferenceSource directory, same as we have it for the calling convention converter from CoreCLR so that we can more easily keep this in sync.

PR #251, opened 2020-10-20
Add doc on Native AOT runtime limitations

jkotas:


docs/using-nativeaot/limitations.md
```diff
@@ -0,0 +1,12 @@
...
+The native AOT .NET runtime form factor comes with a number of fundamental limitations and compatibility issues. The reasons 
+behind them are discussed in length in 
+the [.NET Runtime Form Factors](https://github.com/dotnet/designs/blob/main/accepted/2020/form-factors.md) roadmap.
+
+The key limitations include:
+
+- No dynamic loading (e.g. Assembly.LoadFile)
+- No runtime code generation (e.g. System.Reflection.Emit)
+- No managed C++, no built-in COM and WinRT interop support
+- AOT friendly reflection use only or manual annotations to make arbitrary reflection work
```
samsp-msft, line 12: How do you define "AOT friendly reflection"? The Reflection in AOT doc talks about options, but I didn't pickup on which option is the default or how you can change it to one of the others.
I Understand the RD.xml, and I am assuming you support the code annotations we use for Trimming? we should probably list the table of annotations here or in the Reflection in AOT doc.

jkotas: I have replaced this with a link to the more detailed doc about reflection.

> I'd I am assuming you support the code annotations we use for Trimming?

We do not today. I expect that we will get rid of rd.xml format and just replace it with IL link format used for trimming annotations.

samsp-msft: perfect.

PR #261, opened 2020-10-22
Add official build support to standalone templates

safern:
For easier review I recommend disabling white space diffs. 

I tested with: https://dev.azure.com/dnceng/internal/_build/results?buildId=862898&view=logs&j=226748d0-f812-5437-d3f0-2dd291f5666e

FYI: @AaronRobinsonMSFT @eerhardt @pgovind

### DIFF
--- a/eng/pipelines/runtimelab.yml
+++ b/eng/pipelines/runtimelab.yml
@@ -41,6 +41,10 @@ pr:
     - SECURITY.md
     - THIRD-PARTY-NOTICES.TXT
 
+variables:
+  - name: _TeamName
+    value: dotnet-core
+
 stages:
 - stage: build
   displayName: Build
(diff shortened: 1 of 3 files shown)


eng/pipelines/runtimelab.yml
```diff
@@ -49,19 +53,36 @@ stages:
...
+  - template: \eng\common\templates\post-build\post-build.yml
     parameters:
-      osGroup: Linux
-      archType: x64
-      pool:
-        vmImage: 'ubuntu-latest'
+      publishingInfraVersion: 3
+      validateDependsOn:
+        - build
+      enableSourceLinkValidation: true
```
joperezr, line 88: should we also enable validation/publishing of symbols or does that happen automatically?

eng/pipelines/templates/build-job.yml
```diff
@@ -4,50 +4,74 @@ parameters:
...
   runTests: true
 
 jobs:
 
-- template: /eng/common/templates/job/job.yml
+- template: /eng/common/templates/jobs/jobs.yml
   parameters:
-    displayName: ${{ format('{0} {1}', parameters.osGroup, parameters.archType) }}
-    name: ${{ format('{0}_{1}', parameters.osGroup, parameters.archType) }}
-    enableTelemetry: ${{ parameters.isOfficialBuild }}
```
joperezr, line 16: I must be missing it but where were we passing isOfficialBuild before? Looks like we were just not passing it and always defaulting to empty here right?

joperezr (approved):
LGTM thanks @safern!

eng/pipelines/runtimelab.yml
safern: publishing of symbols is already happening. Validation I didn't enable it because I remember there was problems with it due to a delay in the symbol server when you publish new symbols. @riarenas @epananth is that still the case?

eng/pipelines/templates/build-job.yml
safern: I'm passing it down here: https://github.com/dotnet/runtimelab/pull/261/files#diff-d71a1a613ea5de9e905a5c6f641d8c4c79800b54c4667cbcbefdfb2f91a898ddR57

and then the default of the parameter is set to false: https://github.com/dotnet/runtimelab/pull/261/files#diff-6a89a2c86af1832dfc215647fe7dd379b4def9cea440cf14ef179b65eb55e94aR7

joperezr: Yeah, I saw that. I meant before this change, we seem to have been using it to set the enableTelemetry field but we weren't actually setting it on the main yml or importing it unless I'm missing something.

eng/pipelines/runtimelab.yml
riarenas: @safern yes, I believe that's still the case.

eng/pipelines/templates/build-job.yml
safern: Ah I see what you're saying. The arcade templates ignore true if it is in an official build.

PR #265, opened 2020-10-24
Disable reflection invoke for all byref-like arguments

jkotas:
byref-like arguments are leading to asserts/crashes at runtime. It is not straightforward to make them work
well - it would require generics over byref-like types with the current architecture, and it is corner
case that is unlikely to show up in real app.

### DIFF
--- a/src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/InvokeUtils.cs
+++ b/src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/InvokeUtils.cs
@@ -56,6 +56,9 @@ namespace System
 
         internal static object CheckArgument(object srcObject, EETypePtr dstEEType, CheckArgumentSemantics semantics, BinderBundle binderBundle, Func<Type> getExactTypeForCustomBinder = null)
         {
+            // Methods with ByRefLike types in signatures should be filtered out by the compiler
+            Debug.Assert(!dstEEType.IsByRefLike);
+
             if (srcObject == null)
             {
                 // null -> default(T)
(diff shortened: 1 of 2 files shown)


MichalStrehovsky (approved):
Thank you!

src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/InvokeUtils.cs
```diff
@@ -587,16 +590,12 @@ internal static object DynamicInvokeCallTemplate(object thisPtr, IntPtr methodTo
             return null;
         }
 
-        [DebuggerStepThrough]
```
MichalStrehovsky, line 590: I was thinking that we would want to add support for this (and DebuggerGuidedStepThrough) at some point so that we can have a "nice" stepping behavior. I think it can be done with the F00F00 and FEEFEE line numbers. It's Windows-specific, but if the platform allows that, it's nice to use it.

jkotas: Agree. The ones I have deleted here seems to be leftovers from when the code was factored differently. E.g.:

CallDynamicInvokeMethod (marked with `DebuggerGuidedStepThroughAttribute`)
-> `DynamicInvokeParamHelperCore` (no marked)
-> `DynamicInvokeBoxedValuetypeReturn` (no marked)
-> `DynamicInvokeUnboxIntoActualNullable` (I have deleted `DebuggerStepThrough` on it in this PR)

PR #266, opened 2020-10-25
Fix STA apartment handling

jkotas:


src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Threading/WaitHandle.Windows.cs
```diff
@@ -22,23 +22,32 @@ private static unsafe int WaitForMultipleObjectsIgnoringSyncContext(IntPtr* pHan
...
+                //
+                // In the CLR, we use CoWaitForMultipleHandles to pump messages while waiting in an STA.  In that case, we cannot use WAIT_ALL.
+                // That's because the wait would only be satisfied if a message arrives while the handles are signalled.
+                //
+                if (waitAll)
                     throw new NotSupportedException(SR.NotSupported_WaitAllSTAThread);
+
+                // CoWaitForMultipleHandles does not support more than 63 handles. It returns RPC_S_CALLPENDING for more than 63 handles
+                // that is impossible to differentiate from timeout.
+                if (numHandles > 63)
```
jkotas, line 42: This is fixing a test failure that results into fatal unhandled in System.Runtime.Tests run

src/coreclr/src/tools/aot/ILCompiler.Compiler/IL/Stubs/StartupCode/StartupCodeMainMethod.cs
```diff
@@ -92,6 +92,22 @@ public override MethodIL EmitIL()
                 codeStream.Emit(ILOpcode.call, emitter.NewToken(initEntryAssembly));
             }
 
+            // Initialize COM apartment
+            MethodDesc initApartmentState = startup.GetMethod("InitializeApartmentState", null);
```
jkotas, line 96: Fixed this while I was on it.

src/coreclr/src/tools/aot/ILCompiler.Compiler/IL/Stubs/StartupCode/StartupCodeMainMethod.cs
```diff
@@ -92,6 +92,22 @@ public override MethodIL EmitIL()
...
+            if (initApartmentState != null)
+            {
+                if (_mainMethod.HasCustomAttribute("System", "STAThreadAttribute"))
+                {
+                    codeStream.EmitLdc(0 /* ApartmentState.STA */);
+                    codeStream.Emit(ILOpcode.call, emitter.NewToken(initApartmentState));
+                }
+                if (_mainMethod.HasCustomAttribute("System", "MTAThreadAttribute"))
```
jkotas, line 104: This structure is conveniently going to throw exception at runtime if one gets both STAThread and MTAThread attributes specified.

src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Threading/WaitHandle.Windows.cs
MichalStrehovsky: Wouldn't even think to look in this direction. Thanks!

src/coreclr/src/tools/aot/ILCompiler.Compiler/IL/Stubs/StartupCode/StartupCodeMainMethod.cs
```diff
@@ -92,6 +92,22 @@ public override MethodIL EmitIL()
                 codeStream.Emit(ILOpcode.call, emitter.NewToken(initEntryAssembly));
             }
 
+            // Initialize COM apartment
+            MethodDesc initApartmentState = startup.GetMethod("InitializeApartmentState", null);
+            if (initApartmentState != null)
+            {
+                if (_mainMethod.HasCustomAttribute("System", "STAThreadAttribute"))
+                {
+                    codeStream.EmitLdc(0 /* ApartmentState.STA */);
```
MichalStrehovsky, line 101: Nit: We could probably just use the enum directly. The enums are guaranteed to not go out of sync.

MichalStrehovsky (approved):
Thank you!

src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Threading/WaitHandle.Windows.cs
jkotas: It is not fixing the GC crash. It is fixing a different crash that was being hit some of the time.

PR #270, opened 2020-10-26
Fix native library sample

jkotas:


jkotas:
`--require-defined` seems to be a better fit for fixing this than the whole-archive.

src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
```diff
@@ -63,10 +63,6 @@ The .NET Foundation licenses this file to you under the MIT license.
       <NativeLibrary Include="crypt32.lib" />
     </ItemGroup>
 
-    <PropertyGroup>
-      <IlcProcessEntrypoint Condition="$(IlcProcessEntrypoint) == ''">wmainCRTStartup</IlcProcessEntrypoint>
```
jkotas, line 67: I have started by introducing another variable for the SharedLib init entrypoint, but then decided against it as unnecessary abstraction. I think we should wait for a real case why it is useful.

MichalStrehovsky: It would be useful for the @kant2002 inspired project file of my Doom fire effect: https://github.com/MichalStrehovsky/sharpfire/blob/fee39991c22b39d263933d0fd1ec25d585d82e6e/nogui.csproj#L4 (that's why I added option to set your own entrypoint when I was messing with it).

I had to avoid specifying `WinExe` because then the entrypoint gets set up from here and breaks things.

Arguably I never ended up using it because I have better things to do than maintaining my old hacks :).

Two weeks after the hack I learned about [/ALTERNATENAME](https://devblogs.microsoft.com/oldnewthing/20200731-00/?p=104024) that can be used for this too I guess.

I don't particularly care, just wanted to share the history.

MichalStrehovsky (approved):
Thank you!

src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
MichalStrehovsky: Actually we might be able to use /ALTERNATENAME to remove the need for the /ENTRY line.

Something like `/ALTERNATENAME:wWinMainCRTStartup=wmainCRTStartup` should also do the trick and we don't even need to condition it on WinExe.

jkotas: I have left it as is. The ALTERNATENAME trick is neat, but i did not look like an improvement to me.

PR #271, opened 2020-10-26
Disable value task pooling for native AOT

jkotas:
This is experimental feature that adds about 1.7kB in binary footprint per method returning ValueTask.

Ifdef it out for native AOT until we figure out what to do with it. See https://github.com/dotnet/runtime/issues/13633.


MichalStrehovsky (approved):
Thank you!

PR #272, opened 2020-10-26
Array marshalling support

jkoritzinsky:
Implement support for marshalling blittable and non-blittable arrays, with the small size stack allocation optimization for arrays under 256 bytes.

Includes support for both SizeParamIndex and SizeConst with the same semantics as the runtime.

Example codegen:

Pass int array by value:

```csharp
namespace DllImportGenerator.IntegrationTests
{
    partial class NativeExportsNE
    {
        public partial class Arrays
        {
            public static partial int Sum(int[] values, int numValues)
            {
                unsafe
                {
                    int __retVal;
                    //
                    // Invoke
                    //
                    fixed (int *__values_gen_native = &System.Runtime.InteropServices.MemoryMarshal.GetArrayDataReference(values))
                    {
                        __retVal = Sum__PInvoke__(__values_gen_native, numValues);
                    }

                    return __retVal;
                }
            }

            [System.Runtime.InteropServices.DllImportAttribute("NativeExportsNE", EntryPoint = "sum_int_array")]
            extern private static unsafe int Sum__PInvoke__(int *values, int numValues);
        }
    }
}
```

Pass int array by reference:

```csharp
namespace DllImportGenerator.IntegrationTests
{
    partial class NativeExportsNE
    {
        public partial class Arrays
        {
            public static partial void Duplicate(ref int[] values, int numValues)
            {
                unsafe
                {
                    //
                    // Setup
                    //
                    int *__values_gen_native;
                    //
                    // Marshal
                    //
                    int values__bytelen = sizeof(int) * values.Length;
                    __values_gen_native = (int *)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(values__bytelen);
                    values.AsSpan().CopyTo(new System.Span<int>(__values_gen_native, values.Length));
                    //
                    // Invoke
                    //
                    Duplicate__PInvoke__(&__values_gen_native, numValues);
                    //
                    // Unmarshal
                    //
                    if (__values_gen_native != null)
                    {
                        values = new int[checked((int)numValues)];
                        new System.Span<int>(__values_gen_native, values.Length).CopyTo(values);
                    }
                    else
                        values = null;
                    //
                    // Cleanup
                    //
                    System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__values_gen_native);
                }
            }

            [System.Runtime.InteropServices.DllImportAttribute("NativeExportsNE", EntryPoint = "duplicate_int_array")]
            extern private static unsafe void Duplicate__PInvoke__(int **values, int numValues);
        }
    }
}
```

Pass non-blittable array (string array) by reference:

```csharp
namespace DllImportGenerator.IntegrationTests
{
    partial class NativeExportsNE
    {
        public partial class Arrays
        {
            public static partial void ReverseStrings(ref string[] strArray, out int numElements)
            {
                unsafe
                {
                    numElements = default;
                    //
                    // Setup
                    //
                    ushort **__strArray_gen_native;
                    //
                    // Marshal
                    //
                    int strArray__bytelen = sizeof(ushort *) * strArray.Length;
                    __strArray_gen_native = (ushort **)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(strArray__bytelen);
                    for (int i = 0; i < strArray.Length; ++i)
                    {
                        __strArray_gen_native[i] = (ushort *)System.Runtime.InteropServices.Marshal.StringToCoTaskMemUni(strArray[i]);
                    }

                    //
                    // Invoke
                    //
                    fixed (int *__numElements_gen_native = &numElements)
                    {
                        ReverseStrings__PInvoke__(&__strArray_gen_native, __numElements_gen_native);
                    }

                    //
                    // Unmarshal
                    //
                    if (__strArray_gen_native != null)
                    {
                        strArray = new string[checked((int)numElements)];
                        for (int i = 0; i < strArray.Length; ++i)
                        {
                            strArray[i] = __strArray_gen_native[i] == null ? null : new string ((char *)__strArray_gen_native[i]);
                        }
                    }
                    else
                        strArray = null;
                    //
                    // Cleanup
                    //
                    for (int i = 0; i < strArray.Length; ++i)
                    {
                        System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__strArray_gen_native[i]);
                    }

                    System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__strArray_gen_native);
                }
            }

            [System.Runtime.InteropServices.DllImportAttribute("NativeExportsNE", EntryPoint = "reverse_strings")]
            extern private static unsafe void ReverseStrings__PInvoke__(ushort ***strArray, int *numElements);
        }
    }
}
```

Using SizeConst:

```csharp
namespace DllImportGenerator.IntegrationTests
{
    partial class NativeExportsNE
    {
        public partial class Arrays
        {
            public static partial byte[] GetLongBytes(long l)
            {
                unsafe
                {
                    byte[] __retVal;
                    //
                    // Setup
                    //
                    byte *__retVal_gen_native;
                    //
                    // Invoke
                    //
                    __retVal_gen_native = GetLongBytes__PInvoke__(l);
                    //
                    // Unmarshal
                    //
                    if (__retVal_gen_native != null)
                    {
                        __retVal = new byte[8];
                        new System.Span<byte>(__retVal_gen_native, __retVal.Length).CopyTo(__retVal);
                    }
                    else
                        __retVal = null;
                    return __retVal;
                }
            }

            [System.Runtime.InteropServices.DllImportAttribute("NativeExportsNE", EntryPoint = "get_long_bytes")]
            extern private static unsafe byte *GetLongBytes__PInvoke__(long l);
        }
    }
}
```

### DIFF
--- a/DllImportGenerator/DllImportGenerator.IntegrationTests/ArrayTests.cs
+++ b/DllImportGenerator/DllImportGenerator.IntegrationTests/ArrayTests.cs
@@ -107,6 +107,12 @@ namespace DllImportGenerator.IntegrationTests
             var strings = GetStringArray();
             Assert.Equal(strings.Sum(str => str?.Length ?? 0), NativeExportsNE.Arrays.SumStringLengths(strings));
         }
+        
+        [Fact]
+        public void ByValueNullArrayWithElementMarshalling()
+        {
+            Assert.Equal(0, NativeExportsNE.Arrays.SumStringLengths(null));
+        }
 
         [Fact]
         public void ByRefArrayWithElementMarshalling()
@@ -116,6 +122,15 @@ namespace DllImportGenerator.IntegrationTests
             NativeExportsNE.Arrays.ReverseStrings(ref strings, out _);
             
             Assert.Equal((IEnumerable<string>)expectedStrings, strings);
+        }
+
+        [Fact]
+        public void ByRefNullArrayWithElementMarshalling()
+        {
+            string[] strings = null;
+            NativeExportsNE.Arrays.ReverseStrings(ref strings, out _);
+            
+            Assert.Null(strings);
         }
 
         [Fact]
--- a/DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
+++ b/DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
@@ -90,8 +90,7 @@ namespace Microsoft.Interop
                         foreach (var statement in GenerateConditionalAllocationSyntax(
                             info,
                             context,
-                            StackAllocBytesThreshold,
-                            checkForNull: true))
+                            StackAllocBytesThreshold))
                         {
                             yield return statement;
                         }
--- a/DllImportGenerator/DllImportGenerator/Marshalling/ConditionalStackallocMarshallingGenerator.cs
+++ b/DllImportGenerator/DllImportGenerator/Marshalling/ConditionalStackallocMarshallingGenerator.cs
@@ -17,8 +17,7 @@ namespace Microsoft.Interop
         protected IEnumerable<StatementSyntax> GenerateConditionalAllocationSyntax(
             TypePositionInfo info, 
             StubCodeContext context,
-            int stackallocMaxSize,
-            bool checkForNull)
+            int stackallocMaxSize)
         {
             (string managedIdentifier, string nativeIdentifier) = context.GetIdentifiers(info);
             
@@ -43,93 +42,91 @@ namespace Microsoft.Interop
             
             if (!context.CanUseAdditionalTemporaryState || !context.StackSpaceUsable || (info.IsByRef && info.RefKind != RefKind.In))
             {
+                List<StatementSyntax> statements = new List<StatementSyntax>();
                 if (allocationRequiresByteLength)
                 {
-                    yield return byteLenAssignment;
+                    statements.Add(byteLenAssignment);
                 }
-                yield return allocationStatement;
+                statements.Add(allocationStatement);
+                yield return ExpressionStatement(AssignmentExpression(SyntaxKind.SimpleAssignmentExpression,
+                    IdentifierName(nativeIdentifier),
+                    LiteralExpression(SyntaxKind.NullLiteralExpression)));
+                yield return IfStatement(
+                    BinaryExpression(SyntaxKind.NotEqualsExpression,
+                        IdentifierName(managedIdentifier),
+                        LiteralExpression(SyntaxKind.NullLiteralExpression)),
+                    Block(statements));
+                yield break;
             }
-            else
-            {
-                // <allocationMarkerIdentifier> = false;
-                yield return LocalDeclarationStatement(
+            // <allocationMarkerIdentifier> = false;
+            yield return LocalDeclarationStatement(
+                VariableDeclaration(
+                    PredefinedType(Token(SyntaxKind.BoolKeyword)),
+                    SingletonSeparatedList(
+                        VariableDeclarator(allocationMarkerIdentifier)
+                            .WithInitializer(EqualsValueClause(LiteralExpression(SyntaxKind.FalseLiteralExpression))))));
+
+            // Code block for stackalloc if number of bytes is below threshold size
+            var marshalOnStack = Block(
+                // byte* <stackAllocPtr> = stackalloc byte[<byteLen>];
+                LocalDeclarationStatement(
                     VariableDeclaration(
-                        PredefinedType(Token(SyntaxKind.BoolKeyword)),
+                        PointerType(PredefinedType(Token(SyntaxKind.ByteKeyword))),
                         SingletonSeparatedList(
-                            VariableDeclarator(allocationMarkerIdentifier)
-                                .WithInitializer(EqualsValueClause(LiteralExpression(SyntaxKind.FalseLiteralExpression))))));
+                            VariableDeclarator(stackAllocPtrIdentifier)
+                                .WithInitializer(EqualsValueClause(
+                                    StackAllocArrayCreationExpression(
+                                        ArrayType(
+                                            PredefinedType(Token(SyntaxKind.ByteKeyword)),
+                                            SingletonList<ArrayRankSpecifierSyntax>(
+                                                ArrayRankSpecifier(SingletonSeparatedList<ExpressionSyntax>(
+                                                    IdentifierName(byteLenIdentifier))))))))))),
+                GenerateStackallocOnlyValueMarshalling(info, context, Identifier(byteLenIdentifier), Identifier(stackAllocPtrIdentifier)),
+                // <nativeIdentifier> = <stackAllocPtr>;
+                ExpressionStatement(
+                    AssignmentExpression(
+                        SyntaxKind.SimpleAssignmentExpression,
+                        IdentifierName(nativeIdentifier),
+                        CastExpression(
+                            AsNativeType(info),
+                            IdentifierName(stackAllocPtrIdentifier)))));
 
-                yield return byteLenAssignment;
+            //   if (<byteLen> > <StackAllocBytesThreshold>)
+            //   {
+            //       <allocationStatement>;
+            //   }
+            //   else
+            //   {
+            //       byte* <stackAllocPtr> = stackalloc byte[<byteLen>];
+            //       <marshalValueOnStackStatement>;
+            //       <native> = (<nativeType>)<stackAllocPtr>;
+            //   }
+            var allocBlock = IfStatement(
+                BinaryExpression(
+                    SyntaxKind.GreaterThanExpression,
+                    IdentifierName(byteLenIdentifier),
+                    LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(stackallocMaxSize))),
+                Block(
+                    allocationStatement,
+                    ExpressionStatement(
+                        AssignmentExpression(
+                            SyntaxKind.SimpleAssignmentExpression,
+                            IdentifierName(allocationMarkerIdentifier),
+                            LiteralExpression(SyntaxKind.TrueLiteralExpression)))),
+                ElseClause(marshalOnStack));
 
-                // Code block for stackalloc if number of bytes is below threshold size
-                var marshalOnStack = Block(
-                    // byte* <stackAllocPtr> = stackalloc byte[<byteLen>];
-                    LocalDeclarationStatement(
-                        VariableDeclaration(
-                            PointerType(PredefinedType(Token(SyntaxKind.ByteKeyword))),
-                            SingletonSeparatedList(
-                                VariableDeclarator(stackAllocPtrIdentifier)
-                                    .WithInitializer(EqualsValueClause(
-                                        StackAllocArrayCreationExpression(
-                                            ArrayType(
-                                                PredefinedType(Token(SyntaxKind.ByteKeyword)),
-                                                SingletonList<ArrayRankSpecifierSyntax>(
-                                                    ArrayRankSpecifier(SingletonSeparatedList<ExpressionSyntax>(
-                                                        IdentifierName(byteLenIdentifier))))))))))),
-                    GenerateStackallocOnlyValueMarshalling(info, context, Identifier(byteLenIdentifier), Identifier(stackAllocPtrIdentifier)),
-                    // <nativeIdentifier> = <stackAllocPtr>;
+            yield return IfStatement(
+                BinaryExpression(
+                    SyntaxKind.EqualsExpression,
+                    IdentifierName(managedIdentifier),
+                    LiteralExpression(SyntaxKind.NullLiteralExpression)),
+                Block(
                     ExpressionStatement(
                         AssignmentExpression(
                             SyntaxKind.SimpleAssignmentExpression,
                             IdentifierName(nativeIdentifier),
-                            CastExpression(
-                                AsNativeType(info),
-                                IdentifierName(stackAllocPtrIdentifier)))));
-
-                //   if (<byteLen> > <StackAllocBytesThreshold>)
-                //   {
-                //       <allocationStatement>;
-                //   }
-                //   else
-                //   {
-                //       byte* <stackAllocPtr> = stackalloc byte[<byteLen>];
-                //       <marshalValueOnStackStatement>;
-                //       <native> = (<nativeType>)<stackAllocPtr>;
-                //   }
-                var allocBlock = IfStatement(
-                        BinaryExpression(
-                            SyntaxKind.GreaterThanExpression,
-                            IdentifierName(byteLenIdentifier),
-                            LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(stackallocMaxSize))),
-                        Block(
-                            allocationStatement,
-                            ExpressionStatement(
-                                AssignmentExpression(
-                                    SyntaxKind.SimpleAssignmentExpression,
-                                    IdentifierName(allocationMarkerIdentifier),
-                                    LiteralExpression(SyntaxKind.TrueLiteralExpression)))),
-                        ElseClause(marshalOnStack));
-
-                if (checkForNull)
-                {
-                    yield return IfStatement(
-                                BinaryExpression(
-                                    SyntaxKind.EqualsExpression,
-                                    IdentifierName(managedIdentifier),
-                                    LiteralExpression(SyntaxKind.NullLiteralExpression)),
-                                Block(
-                                    ExpressionStatement(
-                                        AssignmentExpression(
-                                            SyntaxKind.SimpleAssignmentExpression,
-                                            IdentifierName(nativeIdentifier),
-                                            LiteralExpression(SyntaxKind.NullLiteralExpression)))),
-                                ElseClause(Block(allocBlock)));
-                }
-                else
-                {
-                    yield return allocBlock;
-                }
-            }
+                            LiteralExpression(SyntaxKind.NullLiteralExpression)))),
+                ElseClause(Block(byteLenAssignment, allocBlock)));
         }
 
         protected StatementSyntax GenerateConditionalAllocationFreeSyntax(
--- a/DllImportGenerator/DllImportGenerator/Marshalling/NonBlittableArrayMarshaller.cs
+++ b/DllImportGenerator/DllImportGenerator/Marshalling/NonBlittableArrayMarshaller.cs
@@ -77,16 +77,20 @@ namespace Microsoft.Interop
                         foreach (var statement in GenerateConditionalAllocationSyntax(
                             info,
                             context,
-                            StackAllocBytesThreshold,
-                            checkForNull: true))
+                            StackAllocBytesThreshold))
                         {
                             yield return statement;
                         }
 
+                        // Iterate through the elements of the array to marshal them
                         string indexerIdentifier = "i";
                         var arraySubContext = new ArrayMarshallingCodeContext(context.CurrentStage, indexerIdentifier, context);
-                        yield return MarshallerHelpers.GetForLoop(managedIdentifer, indexerIdentifier)
-                            .WithStatement(Block(List(_elementMarshaller.Generate(info with { ManagedType = GetElementTypeSymbol(info) }, arraySubContext))));
+                        yield return IfStatement(BinaryExpression(SyntaxKind.NotEqualsExpression,
+                            IdentifierName(managedIdentifer),
+                            LiteralExpression(SyntaxKind.NullLiteralExpression)),
+                            MarshallerHelpers.GetForLoop(managedIdentifer, indexerIdentifier)
+                                .WithStatement(Block(
+                                    List(_elementMarshaller.Generate(info with { ManagedType = GetElementTypeSymbol(info) }, arraySubContext)))));
                     }
                     break;
                 case StubCodeContext.Stage.Unmarshal:
@@ -108,6 +112,7 @@ namespace Microsoft.Interop
                                         ArrayType(GetElementTypeSymbol(info).AsTypeSyntax(),
                                             SingletonList(ArrayRankSpecifier(
                                                 SingletonSeparatedList(_numElementsExpr))))))),
+                                // Iterate through the elements of the native array to unmarshal them
                                 MarshallerHelpers.GetForLoop(managedIdentifer, indexerIdentifier)
                                     .WithStatement(Block(
                                         List(_elementMarshaller.Generate(
@@ -126,8 +131,13 @@ namespace Microsoft.Interop
                         var elementCleanup = List(_elementMarshaller.Generate(info with { ManagedType = GetElementTypeSymbol(info) }, arraySubContext));
                         if (elementCleanup.Count != 0)
                         {
-                            yield return MarshallerHelpers.GetForLoop(managedIdentifer, indexerIdentifier)
-                                .WithStatement(Block(elementCleanup));
+                            // Iterate through the elements of the native array to clean up any unmanaged resources.
+                            yield return IfStatement(
+                                BinaryExpression(SyntaxKind.NotEqualsExpression,
+                                    IdentifierName(managedIdentifer),
+                                    LiteralExpression(SyntaxKind.NullLiteralExpression)),
+                                MarshallerHelpers.GetForLoop(managedIdentifer, indexerIdentifier)
+                                    .WithStatement(Block(elementCleanup)));
                         }
                         yield return GenerateConditionalAllocationFreeSyntax(info, context);
                     }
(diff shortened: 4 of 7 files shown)


DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
```diff
@@ -48,7 +48,31 @@ CharEncoding CharEncoding
...
+    {
+        public MarshalAsInfo(UnmanagedType unmanagedType, CharEncoding charEncoding)
+            :this(unmanagedType, 0, 0, 0, charEncoding)
+        {
+        }
+
+        public MarshallingInfo CreateArraySubTypeMarshalAsInfo()
+        {
+            Debug.Assert(UnmanagedType is UnmanagedType.LPArray or UnmanagedType.ByValArray or UnmanagedType.SafeArray);
+            if (UnmanagedArraySubType == 0)
```
AaronRobinsonMSFT, line 61: Boo. If we are going to use the type system with nullability, we also need to use it properly and accept that `0` isn't valid here. Which means we need a locally named constant with is used to indicate this state.

AaronRobinsonMSFT: The `UnspecifiedData` below is a good example.

DllImportGenerator/DllImportGenerator.IntegrationTests/ArrayTests.cs
```diff
@@ -0,0 +1,129 @@
...
+{
+    partial class NativeExportsNE
+    {
+        public partial class Arrays
+        {
+            [GeneratedDllImport(nameof(NativeExportsNE), EntryPoint = "sum_int_array")]
+            public static partial int Sum(int[] values, int numValues);
+            
+            [GeneratedDllImport(nameof(NativeExportsNE), EntryPoint = "duplicate_int_array")]
+            public static partial void Duplicate([MarshalAs(UnmanagedType.LPArray, SizeParamIndex = 1)] ref int[] values, int numValues);
```
AaronRobinsonMSFT, line 19: I don't see an `in` version.

DllImportGenerator/DllImportGenerator.IntegrationTests/ArrayTests.cs
```diff
@@ -0,0 +1,129 @@
...
+            var expectedStrings = strings.Select(s => ReverseChars(s)).ToArray();
+            NativeExportsNE.Arrays.ReverseStrings(ref strings, out _);
+            
+            Assert.Equal((IEnumerable<string>)expectedStrings, strings);
+        }
+
+        [Fact]
+        public void ConstantSizeArray()
+        {
+            var longVal = 0x12345678L;
```
AaronRobinsonMSFT, line 104: This is only defining 4 bytes, not 8. Is that intentional?

DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
```diff
@@ -33,21 +33,39 @@ public static IEnumerable<object[]> CodeSnippetsToCompile_NoDiagnostics()
...
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<int>() };
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<uint>() };
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<long>() };
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<ulong>() };
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<float>() };
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<double>() };
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<bool>() };
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<IntPtr>() };
+            yield return new[] { CodeSnippets.ArrayParametersAndModifiers<UIntPtr>() };
+            yield return new[] { CodeSnippets.ArrayParameterWithSizeParam<byte>(false) };
```
AaronRobinsonMSFT, line 49: Let's use the argument name with free boolean constants here. Or perhaps we can leverage an existing enum to indicate the byref state?

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -40,6 +40,9 @@ public void Execute(GeneratorExecutionContext context)
             var generatorDiagnostics = new GeneratorDiagnostics(context);
 
             var generatedDllImports = new StringBuilder();
+            generatedDllImports.AppendLine("#pragma warning disable CS8019");
```
AaronRobinsonMSFT, line 43: What's going on here?

DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
```diff
@@ -0,0 +1,232 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Microsoft.Interop
+{
+    internal class BlittableArrayMarshaller : ConditionalStackallocMarshallingGenerator
+    {
+        private const int StackAllocBytesThreshold = 0x200;
```
AaronRobinsonMSFT, line 12: Document why we chose this number. If it was just a "reasonable guess" that is fine, but we need to know why.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallerHelpers.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Microsoft.Interop
+{
+    internal static class MarshallerHelpers
+    {
+        public static ForStatementSyntax GetForLoop(string collectionIdentifier, string indexerIdentifier)
+        {
+            return ForStatement(EmptyStatement())
```
AaronRobinsonMSFT, line 14: C# code example would be helpful here.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -263,5 +275,64 @@ private static IMarshallingGenerator CreateStringMarshaller(TypePositionInfo inf
 
             throw new MarshallingNotSupportedException(info, context);
         }
+
+
```
AaronRobinsonMSFT, line 279: Extra line.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -156,23 +158,33 @@ internal class MarshallingGenerators
...
+                case { MarshallingAttributeInfo: BlittableTypeAttributeInfo }:
                     return Blittable;
 
                 // Marshalling in new model
                 case { MarshallingAttributeInfo: NativeMarshallingAttributeInfo marshalInfo }:
                     return Forwarder;
 
                 // Simple marshalling with new attribute model, only have type name.
-                case { MarshallingAttributeInfo: GeneratedNativeMarshallingAttributeInfo { NativeMarshallingFullyQualifiedTypeName: string name } }:
+                case { MarshallingAttributeInfo: GeneratedNativeMarshallingAttributeInfo(string nativeTypeName) }:
```
AaronRobinsonMSFT, line 172: I don't understand the `string nativeTypeName` here. Why is that there?

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -263,5 +275,64 @@ private static IMarshallingGenerator CreateStringMarshaller(TypePositionInfo inf
 
             throw new MarshallingNotSupportedException(info, context);
         }
+
+        
+        private static ExpressionSyntax GetNumElementsExpressionFromMarshallingInfo(TypePositionInfo info, StubCodeContext context)
+        {
+            ExpressionSyntax numElementsExpression;
+            if (info.MarshallingAttributeInfo is MarshalAsInfo { ArraySizeParamIndex: short index, ArraySizeConst: int constSize })
```
AaronRobinsonMSFT, line 283: I really dislike this deeply nested local declarations. They aren't clearer to me but rather seem to be creating side-effects. Let's just make them local when we end the if branch.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -263,5 +275,64 @@ private static IMarshallingGenerator CreateStringMarshaller(TypePositionInfo inf
...
+                LiteralExpressionSyntax? constSizeExpression = constSize != MarshalAsInfo.UnspecifiedData
+                    ? LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(constSize))
+                    : null;
+                ExpressionSyntax? sizeParamIndexExpression = null;
+                if (index != MarshalAsInfo.UnspecifiedData)
+                {
+                    TypePositionInfo? paramIndexInfo = context.GetTypePositionInfoForManagedIndex(index);
+                    if (paramIndexInfo is null)
+                    {
+                        throw new MarshallingNotSupportedException(info, context);
```
AaronRobinsonMSFT, line 294: Can we make this exception more precise here. I think this really means the caller didn't provide an index correct? If this will be a specific message then okay, otherwise we should provide details.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -263,5 +275,64 @@ private static IMarshallingGenerator CreateStringMarshaller(TypePositionInfo inf
...
+                    }
+                    else if (!paramIndexInfo.ManagedType.IsIntegralType())
+                    {
+                        throw new MarshallingNotSupportedException(info, context);
```
AaronRobinsonMSFT, line 298: Same as above. Is this a generic message?

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -263,5 +275,64 @@ private static IMarshallingGenerator CreateStringMarshaller(TypePositionInfo inf
...
+                        var (managed, native) = context.GetIdentifiers(paramIndexInfo);
+                        string identifier = Create(paramIndexInfo, context).UsesNativeIdentifier(paramIndexInfo, context) ? native : managed;
+                        sizeParamIndexExpression = CastExpression(
+                                PredefinedType(Token(SyntaxKind.IntKeyword)),
+                                IdentifierName(identifier));
+                    }
+                }
+                numElementsExpression = (constSizeExpression, sizeParamIndexExpression) switch
+                {
+                    (null, null) => throw new MarshallingNotSupportedException(info, context),
```
AaronRobinsonMSFT, line 311: Specific message.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -263,5 +275,64 @@ private static IMarshallingGenerator CreateStringMarshaller(TypePositionInfo inf
...
+        {
+            var elementMarshaller = Create(TypePositionInfo.CreateForType(elementType, elementMarshallingInfo), context);
+            ExpressionSyntax numElementsExpression = LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(0));
+            if (info.IsManagedReturnPosition || (info.IsByRef && info.RefKind != RefKind.In))
+            {
+                // In this case, we need a numElementsExpression supplied from metadata, so we'll calculate it here.
+                numElementsExpression = GetNumElementsExpressionFromMarshallingInfo(info, context);
+            }
+            
+            return elementMarshaller == Blittable ? new BlittableArrayMarshaller(numElementsExpression) : new NonBlittableArrayMarshaller(elementMarshaller, numElementsExpression);
```
AaronRobinsonMSFT, line 335: Just for style can we put this on multiple lines?

DllImportGenerator/DllImportGenerator/Marshalling/NonBlittableArrayMarshaller.cs
```diff
@@ -0,0 +1,183 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Microsoft.Interop
+{
+    internal class NonBlittableArrayMarshaller : ConditionalStackallocMarshallingGenerator
+    {
+        private const int StackAllocBytesThreshold = 0x200;
```
AaronRobinsonMSFT, line 12: Constants should have documentation for "why"

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
```diff
@@ -48,7 +48,31 @@ CharEncoding CharEncoding
...
+                return NoMarshallingInfo.Instance;
+            }
+            return new MarshalAsInfo(UnmanagedArraySubType, CharEncoding);
+        }
+
+        /// <summary>
+        /// Helper method to enable cleaner pattern matching for the common case of
+        /// a MarshalAs attribute that just uses the constructor parameter and no additional properties.
+        /// </summary>
+        public void Deconstruct(out UnmanagedType unmanagedType) => unmanagedType = UnmanagedType;
```
AaronRobinsonMSFT, line 72: This doesn't seem to be used anywhere.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
jkoritzinsky: I added a default `using System;` to enable using the `.AsSpan()` extension method on arrays in the blittable array marshalling. This pragma disables the "unused using" warning when the namespace isn't used.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
jkoritzinsky: This statement has the same meaning as the previous version, but instead of using the property deconstruction syntax, it uses the positional syntax that matches the record's constructor (well, technically the auto-generated `Deconstruct` method that matches the primary constructor). So this case matches a `GeneratedNativeMarshallingAttributeInfo` attribute that was created with any `string`. That `string` is then stored in the `nativeTypeName` variable.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
jkoritzinsky: I personally think this way is cleaner, but I can rearrange things and gather these properties in a more traditional C# style.

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
jkoritzinsky: The `Deconstruct` method is  a special method name for pattern matching. In this case, it enables the `x is MarshalAsInfo(UnmanagedType.I4)` syntax instead of the `x is MarshalAsInfo { UnmanagedType: UnmanagedType.I4 }`. The constructor style syntax is much cleaner in my opinion.

DllImportGenerator/DllImportGenerator/StubCodeContext.cs
```diff
@@ -66,6 +66,8 @@ public enum Stage
 
         public abstract bool StackSpaceUsable { get; }
 
+        public abstract bool CanUseAdditionalTemporaryState { get; }
```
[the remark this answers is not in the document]
jkoritzinsky: This bool encapsulates an implicit policy that we had before and are using with the conditional stackalloc and the SafeHandle marshalling. When this is false, we can't introduce additional variables to track state; we can only use the `managedIdentifier` and `nativeIdentifier` variables to store state.

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
jkoritzinsky: Now that we have the two-parameter constructor with the CharSet however, I might modify this method to instead mirror the second constructor with the CharSet parameter.

DllImportGenerator/DllImportGenerator/StubCodeContext.cs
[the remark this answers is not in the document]
elinor-fung: Can we list out the matrix of scenarios / expected values for these properties - in `Pipeline.md` and/or as a comment here? Specifically `PinningSupported`, `StackSpaceUsable`, and `CanUseAdditionalTemporaryState`.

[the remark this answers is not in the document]
jkoritzinsky: I'll update Pipeline.md to include a description of these switches and what they describe as supported.

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
AaronRobinsonMSFT: When/how often is this called? A very common complaint about C++ is implicit constructors and this seems very similar in context. I personally am not fan of the current switch statement we have because it is no longer a jump table but rather a much more complex comparison matrix. @elinor-fung removed the `char` and `string` because it was unnecessarily complex and this seems to add to enable that pattern which is far too dense in my experience What are we gaining with this pattern? I am close to saying this isn't an approach we should be using given it is much clearer to have specialized functions that are easier to maintain than the enormous switch statement.

Is there an argument for this approach?

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
AaronRobinsonMSFT: Let's comment that. Thank you making the code clean. We need to make sure we document why we are disabling anything.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
AaronRobinsonMSFT: I think I understood that actually - yay for being in on the new C# stuff. What I don't see is where `nativeTypeName` is being used, that is my main confusion.

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
jkoritzinsky: The main argument for adding the `Deconstruct` method is to make all of the patterns that were of the form `MarshalAsInfo { UnmanagedType: UnmanagedType.I4 }` to instead be of the form `MarshalAsInfo(UnmanagedType.I4)`. In our marshaller selector in this PR, it's being used 15 times. This syntax also enables the simple MarshalAs checks that just depend on the first parameter to the MarshalAs attribute to look like the application of the MarshalAs attribute, which is a nice touch.

If I move some of the properties (such as the array-specific ones) on the MarshalAs record from being positional (in the constructor) to just regular init properties, I can remove the need for this Deconstruct method by making the two-parameter version be auto-generated. I'll try a few things out to make the Deconstruct method unnecessary.

AaronRobinsonMSFT: Appreciate the effort.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
jkoritzinsky: We could just cast to manually create the span, but I like the `AsSpan` method because it's cleaner and because emitting a generic type in Roslyn is even more verbose than usual. I can change it to instead just create the span instance with the constructor.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
jkoritzinsky: It's not being used today. I originally added `nativeTypeName` when I was starting work on the non-blittable struct marshaller. It will be used when we support generating code for non-blittable structs to enable us to have both source generators working without a dependency between the two (which isn't currently allowed). `nativeTypeName` will be a pre-calculated name for what the struct marshalling source generator will name the native type.

jkoritzinsky: I can remove it and just match on `GeneratedNativeMarshallingAttribute` until we implement support if you want.

AaronRobinsonMSFT: Not necessary. Wanted to make sure we knew why.

jkoritzinsky:
Any more feedback on this PR?

DllImportGenerator/TestAssets/NativeExports/Arrays.cs
```diff
@@ -0,0 +1,114 @@
...
+        [UnmanagedCallersOnly(EntryPoint = "reverse_strings")]
+        public static void ReverseStrings(ushort*** strArray, int* numValues)
+        {
+            List<IntPtr> newStrings = new List<IntPtr>();
+            for (int i = 0; (nint)(*strArray)[i] != 0; i++)
+            {
+                newStrings.Add((IntPtr)Strings.Reverse((*strArray)[i]));
+            }
+            newStrings.Add(IntPtr.Zero);
+            *strArray = (ushort**)Marshal.AllocCoTaskMem(sizeof(ushort*) * newStrings.Count);
```
elinor-fung, line 90: Do we need to free the existing array first?

DllImportGenerator/DllImportGenerator/Marshalling/ConditionalStackallocMarshallingGenerator.cs
```diff
@@ -0,0 +1,220 @@
...
+
+        private static string GetByteLengthIdentifier(string managedIdentifier) => $"{managedIdentifier}__bytelen";
+
+        private static string GetStackAllocIdentifier(string managedIdentifier) => $"{managedIdentifier}__stackptr";
+
+        protected IEnumerable<StatementSyntax> GenerateConditionalAllocationSyntax(
+            TypePositionInfo info, 
+            StubCodeContext context,
+            int stackallocMaxSize,
+            bool checkForNull)
```
elinor-fung, line 21: All current usages seem to set this to true - when would we not want the null check?

DllImportGenerator/DllImportGenerator/Marshalling/NonBlittableArrayMarshaller.cs
```diff
@@ -0,0 +1,188 @@
...
+                        foreach (var statement in GenerateConditionalAllocationSyntax(
+                            info,
+                            context,
+                            StackAllocBytesThreshold,
+                            checkForNull: true))
+                        {
+                            yield return statement;
+                        }
+
+                        string indexerIdentifier = "i";
```
elinor-fung, line 86: Quick comment that this is iterating over and marshalling each element?

DllImportGenerator/DllImportGenerator/Marshalling/NonBlittableArrayMarshaller.cs
```diff
@@ -0,0 +1,188 @@
...
+                        var arraySubContext = new ArrayMarshallingCodeContext(context.CurrentStage, indexerIdentifier, context);
+                        yield return MarshallerHelpers.GetForLoop(managedIdentifer, indexerIdentifier)
```
elinor-fung, line 88: Null check before doing the loop too, since it does `<identifier>.Length`? Also in Cleanup.

elinor-fung: We should also have tests with null array parameter in / out.

DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
```diff
@@ -0,0 +1,238 @@
...
+                SizeOfExpression(GetElementTypeSyntax(info)),
+                MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression,
+                    IdentifierName(context.GetIdentifiers(info).managed),
+                    IdentifierName("Length")
+                ));
+        }
+
+        protected override StatementSyntax GenerateStackallocOnlyValueMarshalling(TypePositionInfo info, StubCodeContext context, SyntaxToken byteLengthIdentifier, SyntaxToken stackAllocPtrIdentifier)
+        {
+            return EmptyStatement();
```
elinor-fung, line 219: Will this actually show up as an empty statement (`;`) in the generated code?

elinor-fung (approved):

DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
jkoritzinsky: Yes it will.

PR #296, opened 2020-11-04
ARM64&Unix: New unwinder (#8345)

jkotas:
Implements a custom unwinder for ARM64 that does not need the generic CFI based libunwind functions. This way we only need to loop just once over all used registers instead of doing the complicated CFI process.This will improve the performances for everything that needs a stack walk.

Additional unwind data are generated and stored as part of the LSDA. 
-4 byte for a relative offset
-3 byte for the CFA
-3 byte for each register that is part of the prolog save
-1 byte end marker

We still keep the CFI data for the debugger. Maybe the could be removed for a build without debug information.

A possible future optimization would be to introduce a compact format for the information used by the most common cases. This could be stored in the relative offset. But compared to the replacement of libunwind the wins would be rather small.

The same can be done for any  other supported architecture but each one will require a custom unwind function

### DIFF
--- a/src/coreclr/src/nativeaot/Runtime/unix/UnixContext.cpp
+++ b/src/coreclr/src/nativeaot/Runtime/unix/UnixContext.cpp
@@ -8,6 +8,7 @@
 #include "daccess.h"
 #include "PalRedhawkCommon.h"
 #include "regdisplay.h"
+#include "ICodeManager.h"
 #include "config.h"
 
 #include <libunwind.h>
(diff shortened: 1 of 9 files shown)


jkotas:
Mop-up integration from dotnet/corert

jkotas:
Wait for the arm64 build to come online

am11:
cc @RalfKornmannEnvision
thanks again for doing this! :)

PR #302, opened 2020-11-04
Implement the non-blittable type marshalling proposal

jkoritzinsky:
Implement the marshaller defined in the StructMarshalling.md design doc.

Additionally, extend the design to support a Value property that returns a byref (currently works for by value and in parameters only).

Example generated code is included below.

<details>
<summary>Native type without Value property</summary>
Struct definition:

```csharp
[NativeMarshalling(typeof(Native))]
struct S
{
    public bool b;
}

struct Native
{
    private int value;
    public Native(S s)
    {
        value = s.b ? 1 : 0;
    }

    public S ToManaged() => new S { b = value != 0 };
}
```

Generated code:

```csharp
partial class Test
{
    public static partial global::S Method(global::S p, in global::S pIn, ref global::S pRef, out global::S pOut)
    {
        unsafe
        {
            global::Native __p_gen_native = default;
            global::Native __pIn_gen_native = default;
            global::Native __pRef_gen_native = default;
            pOut = default;
            global::Native __pOut_gen_native = default;
            global::S __retVal = default;
            global::Native __retVal_gen_native = default;
            //
            // Marshal
            //
            __p_gen_native = new global::Native(p);
            __pIn_gen_native = new global::Native(pIn);
            __pRef_gen_native = new global::Native(pRef);
            //
            // Invoke
            //
            __retVal_gen_native = Method__PInvoke__(__p_gen_native, &__pIn_gen_native, &__pRef_gen_native, &__pOut_gen_native);
            //
            // Unmarshal
            //
            __retVal = __retVal_gen_native.ToManaged();
            pRef = __pRef_gen_native.ToManaged();
            pOut = __pOut_gen_native.ToManaged();
            return __retVal;
        }
    }

    extern private static unsafe global::Native Method__PInvoke__(global::Native p, global::Native*pIn, global::Native*pRef, global::Native*pOut);
}

```

</details>

<details>
<summary>Native type with Value property</summary>
Struct definition:

```csharp
[NativeMarshalling(typeof(Native))]
struct S
{
    public bool b;
}

struct Native
{
    public Native(S s)
    {
        Value = s.b ? 1 : 0;
    }

    public S ToManaged() => new S { b = Value != 0 };

    public int Value { get; set; }
}
```

Generated code:

```csharp
partial class Test
{
    public static partial global::S Method(global::S p, in global::S pIn, ref global::S pRef, out global::S pOut)
    {
        unsafe
        {
            nint __pIn_gen_native = default;
            nint __pRef_gen_native = default;
            pOut = default;
            nint __pOut_gen_native = default;
            global::S __retVal = default;
            nint __retVal_gen_native = default;
            //
            // Setup
            //
            global::Native __retVal_gen_native__marshaler = default;
            global::Native __pIn_gen_native__marshaler = default;
            global::Native __pRef_gen_native__marshaler = default;
            global::Native __pOut_gen_native__marshaler = default;
            //
            // Marshal
            //
            __pIn_gen_native__marshaler = new global::Native(pIn);
            __pIn_gen_native = __pIn_gen_native__marshaler.Value;
            __pRef_gen_native__marshaler = new global::Native(pRef);
            __pRef_gen_native = __pRef_gen_native__marshaler.Value;
            //
            // Invoke
            //
            fixed (void *__p_gen_native = p)
            {
                __retVal_gen_native = Method__PInvoke__((nint)__p_gen_native, &__pIn_gen_native, &__pRef_gen_native, &__pOut_gen_native);
            }

            //
            // Unmarshal
            //
            __retVal_gen_native__marshaler.Value = __retVal_gen_native;
            __retVal = __retVal_gen_native__marshaler.ToManaged();
            __pRef_gen_native__marshaler.Value = __pRef_gen_native;
            pRef = __pRef_gen_native__marshaler.ToManaged();
            __pOut_gen_native__marshaler.Value = __pOut_gen_native;
            pOut = __pOut_gen_native__marshaler.ToManaged();
            return __retVal;
        }
    }

    extern private static unsafe nint Method__PInvoke__(nint p, nint*pIn, nint*pRef, nint*pOut);
}


```

</details>

<details>
<summary>Native type with stackalloc support and no Value property</summary>
Struct definition:

```csharp
[NativeMarshalling(typeof(Native))]
struct S
{
    public bool b;
}

struct Native
{
    private int value;
    public Native(S s, System.Span<byte> b)
    {
        value = s.b ? 1 : 0;
    }

    public S ToManaged() => new S { b = value != 0 };

    public const int StackBufferSize = 1;
}
```

Generated code:

```csharp
partial class Test
{
    public static partial global::S Method(global::S p, in global::S pIn, out global::S pOut)
    {
        unsafe
        {
            global::Native __p_gen_native = default;
            global::Native __pIn_gen_native = default;
            pOut = default;
            global::Native __pOut_gen_native = default;
            global::S __retVal = default;
            global::Native __retVal_gen_native = default;
            //
            // Marshal
            //
            byte *p__stackptr = stackalloc byte[global::Native.StackBufferSize];
            __p_gen_native = new global::Native(p, new System.Span<byte>(p__stackptr, global::Native.StackBufferSize));
            byte *pIn__stackptr = stackalloc byte[global::Native.StackBufferSize];
            __pIn_gen_native = new global::Native(pIn, new System.Span<byte>(pIn__stackptr, global::Native.StackBufferSize));
            //
            // Invoke
            //
            __retVal_gen_native = Method__PInvoke__(__p_gen_native, &__pIn_gen_native, &__pOut_gen_native);
            //
            // Unmarshal
            //
            __retVal = __retVal_gen_native.ToManaged();
            pOut = __pOut_gen_native.ToManaged();
            return __retVal;
        }
    }

    extern private static unsafe global::Native Method__PInvoke__(global::Native p, global::Native*pIn, global::Native*pOut);
}
```

</details>

<details>
<summary>Native type with stackalloc support and Value property</summary>
Struct definition:

```csharp
[NativeMarshalling(typeof(Native))]
struct S
{
    public bool b;
}

struct Native
{
    public Native(S s, System.Span<byte> b)
    {
        Value = s.b ? 1 : 0;
    }

    public S ToManaged() => new S { b = Value != 0 };

    public int Value { get; set; }

    public const int StackBufferSize = 1;
}
```

Generated code:

```csharp
partial class Test
{
    public static partial global::S Method(global::S p, in global::S pIn, out global::S pOut)
    {
        unsafe
        {
            int __p_gen_native = default;
            int __pIn_gen_native = default;
            pOut = default;
            int __pOut_gen_native = default;
            global::S __retVal = default;
            int __retVal_gen_native = default;
            //
            // Setup
            //
            global::Native __retVal_gen_native__marshaler = default;
            global::Native __p_gen_native__marshaler = default;
            global::Native __pIn_gen_native__marshaler = default;
            global::Native __pOut_gen_native__marshaler = default;
            //
            // Marshal
            //
            byte *p__stackptr = stackalloc byte[global::Native.StackBufferSize];
            __p_gen_native__marshaler = new global::Native(p, new System.Span<byte>(p__stackptr, global::Native.StackBufferSize));
            __p_gen_native = __p_gen_native__marshaler.Value;
            byte *pIn__stackptr = stackalloc byte[global::Native.StackBufferSize];
            __pIn_gen_native__marshaler = new global::Native(pIn, new System.Span<byte>(pIn__stackptr, global::Native.StackBufferSize));
            __pIn_gen_native = __pIn_gen_native__marshaler.Value;
            //
            // Invoke
            //
            __retVal_gen_native = Method__PInvoke__(__p_gen_native, &__pIn_gen_native, &__pOut_gen_native);
            //
            // Unmarshal
            //
            __retVal_gen_native__marshaler.Value = __retVal_gen_native;
            __retVal = __retVal_gen_native__marshaler.ToManaged();
            __pOut_gen_native__marshaler.Value = __pOut_gen_native;
            pOut = __pOut_gen_native__marshaler.ToManaged();
            return __retVal;
        }
    }

    extern private static unsafe int Method__PInvoke__(int p, int *pIn, int *pOut);
}

```

</details>

<details>
<summary>Native type with Managed GetPinnableReference optimization</summary>

Struct definition:

```csharp
[NativeMarshalling(typeof(Native))]
class S
{
    public int i;

    public ref int GetPinnableReference() => ref i;
}

unsafe struct Native
{
    private int* ptr;
    public Native(S s)
    {
        ptr = (int*)Marshal.AllocHGlobal(sizeof(int));
        *ptr = s.i;
    }

    public S ToManaged() => new S { i = *ptr };

    public nint Value
    {
        get => (nint)ptr;
        set => ptr = (int*)value;
    }
}
```

Generated code:

```csharp
partial class Test
{
    public static partial global::S Method(global::S p, in global::S pIn, ref global::S pRef, out global::S pOut)
    {
        unsafe
        {
            nint __pIn_gen_native = default;
            nint __pRef_gen_native = default;
            pOut = default;
            nint __pOut_gen_native = default;
            global::S __retVal = default;
            nint __retVal_gen_native = default;
            //
            // Setup
            //
            global::Native __retVal_gen_native__marshaler = default;
            global::Native __pIn_gen_native__marshaler = default;
            global::Native __pRef_gen_native__marshaler = default;
            global::Native __pOut_gen_native__marshaler = default;
            //
            // Marshal
            //
            __pIn_gen_native__marshaler = new global::Native(pIn);
            __pIn_gen_native = __pIn_gen_native__marshaler.Value;
            __pRef_gen_native__marshaler = new global::Native(pRef);
            __pRef_gen_native = __pRef_gen_native__marshaler.Value;
            //
            // Invoke
            //
            fixed (void *__p_gen_native = p)
            {
                __retVal_gen_native = Method__PInvoke__((nint)__p_gen_native, &__pIn_gen_native, &__pRef_gen_native, &__pOut_gen_native);
            }

            //
            // Unmarshal
            //
            __retVal_gen_native__marshaler.Value = __retVal_gen_native;
            __retVal = __retVal_gen_native__marshaler.ToManaged();
            __pRef_gen_native__marshaler.Value = __pRef_gen_native;
            pRef = __pRef_gen_native__marshaler.ToManaged();
            __pOut_gen_native__marshaler.Value = __pOut_gen_native;
            pOut = __pOut_gen_native__marshaler.ToManaged();
            return __retVal;
        }
    }

    extern private static unsafe nint Method__PInvoke__(nint p, nint*pIn, nint*pRef, nint*pOut);
}

```
</details>

<details>
<summary>Native type with GetPinnableReference method</summary>
Struct definition:

```csharp
class S
{
    public byte c;
}

unsafe ref struct Native
{
    private byte* ptr;
    private Span<byte> stackBuffer;

    public Native(S s) : this()
    {
        ptr = (byte*)Marshal.AllocCoTaskMem(sizeof(byte));
        *ptr = s.c;
    }

    public Native(S s, Span<byte> buffer) : this()
    {
        stackBuffer = buffer;
        stackBuffer[0] = s.c;
    }

    public ref byte GetPinnableReference() => ref (ptr != null ? ref *ptr : ref stackBuffer.GetPinnableReference());

    public S ToManaged()
    {
        return new S { c = *ptr };
    }

    public byte* Value
    {
        get => ptr != null ? ptr : throw new InvalidOperationException();
        set => ptr = value;
    }

    public void FreeNative()
    {
        if (ptr != null)
        {
            Marshal.FreeCoTaskMem((IntPtr)ptr);
        }
    }

    public const int StackBufferSize = 1;
}
```

Generated code:

```csharp
partial class Test
{
    public static partial void Method(global::S s, in global::S sIn)
    {
        unsafe
        {
            //
            // Setup
            //
            global::Native __s_gen_native__marshaler = default;
            global::Native __sIn_gen_native__marshaler = default;
            try
            {
                //
                // Marshal
                //
                byte *s__stackptr = stackalloc byte[global::Native.StackBufferSize];
                __s_gen_native__marshaler = new global::Native(s, new System.Span<byte>(s__stackptr, global::Native.StackBufferSize));
                byte *sIn__stackptr = stackalloc byte[global::Native.StackBufferSize];
                __sIn_gen_native__marshaler = new global::Native(sIn, new System.Span<byte>(sIn__stackptr, global::Native.StackBufferSize));
                //
                // Invoke
                //
                fixed (byte *__s_gen_native = &__s_gen_native__marshaler.GetPinnableReference())
                {
                    fixed (byte *__sIn_gen_native = &__sIn_gen_native__marshaler.GetPinnableReference())
                    {
                        Method__PInvoke__(__s_gen_native, &__sIn_gen_native);
                    }
                }
            }
            finally
            {
                //
                // Cleanup
                //
                __s_gen_native__marshaler.FreeNative();
                __sIn_gen_native__marshaler.FreeNative();
            }
        }
    }

    extern private static unsafe void Method__PInvoke__(byte *s, byte **sIn);
}
```

</details>

### DIFF
--- a/DllImportGenerator/DllImportGenerator.IntegrationTests/CustomMarshallingTests.cs
+++ b/DllImportGenerator/DllImportGenerator.IntegrationTests/CustomMarshallingTests.cs
@@ -37,7 +37,7 @@ namespace DllImportGenerator.IntegrationTests
         public static partial int ReturnStringLength([MarshalUsing(typeof(Utf16StringMarshaler))] string s);
     }
 
-    public class NonBlittableStructTests
+    public class CustomMarshallingTests
     {
         [Fact]
         public void NonBlittableStructWithFree()
--- a/DllImportGenerator/TestAssets/NativeExports/CustomMarshalling.cs
+++ b/DllImportGenerator/TestAssets/NativeExports/CustomMarshalling.cs
@@ -4,7 +4,7 @@ using SharedTypes;
 
 namespace NativeExports
 {
-    public static unsafe class NonBlittableStructs
+    public static unsafe class CustomMarshalling
     {
         [UnmanagedCallersOnly(EntryPoint = "stringcontainer_deepduplicate")]
         [DNNE.C99DeclCode("struct string_container { char* str1; char* str2; };")]


DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
```diff
@@ -446,5 +460,285 @@ partial class Test
...
 
         public static string ArrayParameterWithNestedMarshalInfo<T>(UnmanagedType nestedMarshalType) => ArrayParameterWithNestedMarshalInfo(typeof(T).ToString(), nestedMarshalType);
+        public static string CustomStructMarshallingParametersAndModifiers = BasicParametersAndModifiers("S") + @"
+[NativeMarshalling(typeof(Native))]
+struct S
+{
+    public bool b;
+}
+
+struct Native
```
jkotas, line 470: It would be nice to try to establish naming convention for these and use it here. `Native` is very generic and it does not scale to have more than one in the given namespace.

DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
```diff
@@ -446,5 +460,285 @@ partial class Test
...
+    public int Value { get; set; }
+}
+";
+        public static string CustomStructMarshallingPinnableParametersAndModifiers = BasicParametersAndModifiers("S") + @"
+[NativeMarshalling(typeof(Native))]
+class S
+{
+    public int i;
+
+    public ref int GetPinnableReference() => ref i;
```
jkotas, line 593: I would expect `GetPinnableReference` to be on `Native` for this case. The marshaler type should be in full control for how the value is marshalled.

DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
```diff
@@ -446,5 +460,285 @@ partial class Test
...
+struct Native
+{
+    public Native(S s, System.Span<byte> b)
+    {
+        Value = s.b ? 1 : 0;
+    }
+
+    public S ToManaged() => new S { b = Value != 0 };
+
+    public int Value { get; set; }
```
jkotas, line 563: > Native type with Value property

To reduce number of locals, I may be nice to pass the Value in directly for the simple case, e.g.:

```
__retVal_gen_native = Method__PInvoke__(__p_gen_native__marshaler.Value, ...);
```

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
```diff
@@ -0,0 +1,240 @@
...
+                            (_marshallingMethods & SupportedMarshallingMethods.ManagedToNativeStackalloc) != 0;
+
+                        List<ArgumentSyntax> arguments = new List<ArgumentSyntax>
+                        {
+                            Argument(IdentifierName(managedIdentifier))
+                        };
+
+                        if (scenarioSupportsStackalloc && (!info.IsByRef || info.RefKind == RefKind.In))
+                        {
+                            // stackalloc byte[<_nativeLocalType>.StackBufferSize]
```
jkotas, line 137: We should set the no-zero init flag on the method - if it is not set globally on the assembly already.

AaronRobinsonMSFT: @jkotas Is the intent here avoid the cost of zero initialing the stack space?

jkotas: Yup

DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
jkoritzinsky: The `GetPinnableReference` method on a struct type is only used when we're using marshalling defined by the type author (via the `[NativeMarshalling]` attribute instead of `[MarshalUsing]`). Allowing the `GetPinnableReference` on the user type enables people to write types that can be marshalled exclusively via pinning, like how `string` and blittable arrays support being marshalled by only using a pinning statement.

If we require the construction of the marshaler struct first, then we'd either stack allocate and copy when we don't need to (if the user provides a stackalloc constructor), heap allocate and copy when we don't need to, require the user to allocate on the first access of the `Value` property (basically assuming that the `Value` property is only accessed when a copy is needed), or a new constructor would need to be introduced that doesn't do any allocations.

I don't like the idea of forcing users to delay allocations until the `Value` property access since it seems non-intuitive and like a hole in the design.

I am thinking about allowing `GetPinnableReference` on the marshaler type instead of allowing a ref return `Value` property though. That will likely be a cleaner solution for that problem.

DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
jkoritzinsky: I agree that we need to establish a naming convention for these types, but I don't think we need to do that immediately. We'll need to do it when we decide we want to support generating the native types for structs.

DllImportGenerator/designs/StructMarshalling.md
```diff
@@ -61,6 +61,27 @@ The analyzer will report an error if neither the construtor nor the ToManaged me
 
 If the native type `TNative` also has a public `Value` property, then the value of the `Value` property will be passed to native code instead of the `TNative` value itself. As a result, the type `TNative` will be allowed to be non-blittable and the type of the `Value` property will be required to be blittable. If the `Value` property is settable, then when marshalling in the native-to-managed direction, a default value of `TNative` will have its `Value` property set to the native value. If `Value` does not have a setter, then marshalling from native to managed is not supported.
 
+If the `Value` property has a `ref` or `readonly ref` return type, then the marshalling type is only supported in scenarios where pinning is available. See below for an example of a managed type with a native type with a `Value` property.
```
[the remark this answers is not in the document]
jkoritzinsky: Yes that's currently the case. I'm not a fan of this, so I think I'm going to try an alternative approach where instead of supporting a ref return `Value` property, we support a `GetPinnableReference()` method on the marshaller type. That should remove this restriction and make it possible to support all the scenarios we want with one marshaller type.

jkoritzinsky:
So I've encountered a slight hiccup with the design we have for enabling users to use the stackalloc optimization.

Given the following example managed and native type:

```csharp
using System;

class S
{
    public byte c;
}

unsafe ref struct Native
{
    private byte* ptr;
    private Span<byte> stackBuffer;

    public Native(S s) : this()
    {
        ptr = (byte*)Marshal.AllocCoTaskMem(sizeof(byte));
        *ptr = s.c;
    }

    public Native(S s, Span<byte> buffer) : this()
    {
        stackBuffer = buffer;
        stackBuffer[0] = s.c;
    }

    public ref byte GetPinnableReference() => ref (ptr != null ? ref *ptr : ref stackBuffer.GetPinnableReference());

    public S ToManaged()
    {
        return new S { c = *ptr };
    }

    public byte* Value
    {
        get => ptr != null ? ptr : throw new InvalidOperationException();
        set => ptr = value;
    }

    public void FreeNative()
    {
        if (ptr != null)
        {
            Marshal.FreeCoTaskMem((IntPtr)ptr);
        }
    }

    public const int StackBufferSize = 1;
}
```

The following code is generated:

```csharp

partial class Test
{
    public static void Method(global::S s, in global::S sIn)
    {
        unsafe
        {
            //
            // Setup
            //
            global::Native __s_gen_native__marshaler = default;
            global::Native __sIn_gen_native__marshaler = default;
            //
            // Marshal
            //
            __s_gen_native__marshaler = new global::Native(s, stackalloc byte[global::Native.StackBufferSize]);
            __sIn_gen_native__marshaler = new global::Native(sIn, stackalloc byte[global::Native.StackBufferSize]);
            //
            // Invoke
            //
            fixed (byte *__s_gen_native = &__s_gen_native__marshaler.GetPinnableReference())
            {
                fixed (byte *__sIn_gen_native = &__sIn_gen_native__marshaler.GetPinnableReference())
                {
                    Method__PInvoke__(__s_gen_native, &__sIn_gen_native);
                }
            }

            //
            // Cleanup
            //
            __s_gen_native__marshaler.FreeNative();
            __sIn_gen_native__marshaler.FreeNative();
        }
    }

    extern private static unsafe void Method__PInvoke__(byte *s, byte **sIn);
}

```

This causes a compile error because `__s_gen_native__marshaler` is allowed to escape the method based on its declaration, but the stackalloc'd Span cannot. The pattern I've seen to fix this is to create an instance of the struct at declaration time with a 0-length stackalloc (application in our scenario shown below), but I'm not a big fan of this workaround, so I'm looking for alternative ideas on how to solve this problem.

```diff
- global::Native __s_gen_native__marshaler = default;
+ global::Native __s_gen_native__marshaler = new global::Native(default, stackalloc byte[0]);
```

jkotas:
Assign it directly?
```
global::Native __s_gen_native__marshaler = new global::Native(s, stackalloc byte[global::Native.StackBufferSize]);
```

jkoritzinsky:
We can do that today, but once we start emitting the `try` and `finally` blocks to actually support cleanup in exceptional scenarios, that approach breaks down.

We can't just declare the variable in the try block, because then we won't be able to do cleanup. We can't declare the variable outside the try block and only assign it in the try block because then in the finally it would be considered unassigned and the lifetime error would still happen. And we can't just assign it outside of the try block because then we'd risk leaking resources if another struct that marshals via stack allocation (and as such is assigned outside of the try block) fails and throws an exception.

elinor-fung:
Raw pointer for the `stackalloc` and explicitly construct the Span over it (I think I've seen this in parts of the runtime libraries)?
```C#
byte* __s_gen_native__stackPtr = stackalloc byte[global::Native.StackBufferSize];
__s_gen_native__marshaler = new global::Native(s, new Span<byte>(__s_gen_native__stackPtr, global::Native.StackBufferSize));
```

stephentoub:
> The pattern I've seen to fix this is to create an instance of the struct at declaration time with a 0-length stackalloc (application in our scenario shown below), but I'm not a big fan of this workaround

That's what the language folks recommend until a better solution is available. 
CC: @jaredpar

jkoritzinsky:
> Raw pointer for the `stackalloc` and explicitly construct the Span over it (I think I've seen this in parts of the runtime libraries)?
> 
> ```cs
> byte* __s_gen_native__stackPtr = stackalloc byte[global::Native.StackBufferSize];
> __s_gen_native__marshaler = new global::Native(s, new Span<byte>(__s_gen_native__stackPtr, global::Native.StackBufferSize));
> ```

That should work and doesn’t look that bad! I’ll try that.

DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
jkotas: > The GetPinnableReference method on a struct type is only used when we're using marshalling defined by the type author (via the [NativeMarshalling] attribute instead of [MarshalUsing])

Ok, I was hoping that the rules for marshalers defined by the type author will be exactly same as the rules for MarshalUsing. Is the only reason for this difference to save the author of marshaler from implementing a trivial forwarder for `GetPinnableRefence`, or is there more to it?

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
```diff
@@ -101,16 +101,7 @@ public IEnumerable<StatementSyntax> Generate(TypePositionInfo info, StubCodeCont
...
-                    if (!_marshalerTypePinnable || info.IsByRef && info.RefKind != RefKind.In)
-                    {
-                        yield return LocalDeclarationStatement(
-                        VariableDeclaration(
-                            _nativeTypeSyntax,
-                            SingletonSeparatedList(
-                                VariableDeclarator(nativeIdentifier))));
-                    }
-                    
+                case StubCodeContext.Stage.Setup:
```
AaronRobinsonMSFT, line 104: Oops. Would this even build? What are the semantics when this is removed?

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
```diff
@@ -262,7 +253,7 @@ public bool UsesNativeIdentifier(TypePositionInfo info, StubCodeContext context)
             }
             if (context.PinningSupported)
             {
-                if (!info.IsByRef && (_marshallingMethods & SupportedMarshallingMethods.Pinning) == 0)
+                if (!info.IsByRef && (_marshallingMethods & SupportedMarshallingMethods.Pinning) != 0)
```
AaronRobinsonMSFT, line 256: I assume a test should have failed here. If not, we should consider why not and if a class of bugs are missed by our testing.

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
jkoritzinsky: Yes. In Elinor's work, we've moved the `nativeIdentifier` to be automatically declared when `UsesNativeIdentifier` is true. This ensures that the implementation of `UsesNativeIdentifier` is accurate (and found the bug that I fixed in the other commit I pushed).

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
jkoritzinsky: Yes, when I was doing local testing after consuming Elinor's change that makes automatically declaring `nativeIdentifier` conditional on `UsesNativeIdentifier` returning true, I saw multiple unit test failures as well as failures building the integration tests due to this mistake.

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
AaronRobinsonMSFT: A non-cased statement (i.e. ` if (_useValueProperty) ...`) is permitted in a `switch` statement is the main tenor of my question. What does it mean?

elinor-fung: Handled offline - no issue here. Aaron's diff was confusing.

AaronRobinsonMSFT: @jkoritzinsky Sorry.

elinor-fung:
Could you also update the example generated code in the description to reflect the latest?

jkoritzinsky:
Updated the generated code examples.

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
```diff
@@ -0,0 +1,268 @@
...
+                                            StackAllocArrayCreationExpression(
+                                                    ArrayType(
+                                                        PredefinedType(Token(SyntaxKind.ByteKeyword)),
+                                                        SingletonList(ArrayRankSpecifier(SingletonSeparatedList<ExpressionSyntax>(
+                                                            MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression,
+                                                                _nativeLocalTypeSyntax,
+                                                                IdentifierName(ManualTypeMarshallingHelper.StackBufferSizeFieldName))
+                                                        ))))))))));
+
+                            // stackalloc byte[<_nativeLocalType>.StackBufferSize]
```
elinor-fung, line 149: Update comment

DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
```diff
@@ -516,5 +530,305 @@ partial class Test
...
+    public bool b;
+}
+
+[StructLayout(LayoutKind.Sequential)]
+struct Native
+{
+    private int i;
+    public S ToManaged() => new S { b = i != 0 };
+}
+";
```
elinor-fung, line 832: Do we have `MarshalUsing` covered in any of these snippets for unit tests?

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -374,5 +374,59 @@ private static IMarshallingGenerator CreateArrayMarshaller(TypePositionInfo info
...
+                    NotSupportedDetails = string.Format(Resources.CustomTypeMarshallingNativeToManagedUnsupported, marshalInfo.NativeMarshallingType.ToDisplayString())
+                };
+            }
+            // The marshalling method for this type doesn't support marshalling from managed to native by value,
+            // but our scenario requires marshalling from managed to native by value.
+            // Pinning is required for the stackalloc marshalling to enable users to safely pass the stackalloc Span's byref
+            // to native if we ever start using a conditional stackalloc method and cannot guarantee that the Span we provide
+            // the user with is backed by stack allocated memory.
+            else if (!info.IsByRef &&
+                (marshalInfo.MarshallingMethods & SupportedMarshallingMethods.ManagedToNative) == 0 &&
```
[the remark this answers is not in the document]
jkoritzinsky: I thought you liked it the other way. I'm fine with either, I just want to be consistent.

DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
jkoritzinsky: The reason we allow the `GetPinnableReference` member on the managed type is it enables developers to author a type that has all of the optimizations that we have for marshalling a `string` to a `LPWSTR` with only one marshaller type. The `GetPinnableReference` method on the managed type is used for the pinning scenario (by value in), the `Span<byte>` constructor is used for the stack allocation scenario (by ref in), and the constructor that only takes the managed object is the fallback heap allocation scenario (by ref in/out).

If we didn't allow the `GetPinnableReference` method on the managed type to participate, then we'd have to determine when to prefer the non-stackalloc constructor over the stackalloc constructor (so we don't stack allocate), and the marshaller type author would have to delay allocating and marshalling their data to native until they know that the `GetPinnableReference` method will not be used, so it would have to done be in the `Value` property get accessor if we wanted to enable a single marshaller type to cover all scenarios. And in the current design, we only support associating one marshaler at type author time, any others need to be associated at type usage time via MarshalUsing.

Enabling the marshaller to use the managed `GetPinnableReference` method makes it very simple to describe the scenarios where each constructor and method on the native value type is called and what the author should put in each method body.

MarshalUsing does not enable the managed GetPinnableReference optimization because the marshaller referenced by a MarshalUsing likely marshals the data differently than a marshaller written by the type author. If a type author wants to define marshalling for the type, they can apply the `[NativeMarshalling]` attribute to the type; `[MarshalUsing]` is primarily for scenarios where the marshaler cannot be applied to the type because it lives in another assembly or already has a different default marshalling scheme.

jkotas: Ok, make sense. Thank you for the explanation.

DllImportGenerator/designs/StructMarshalling.md
```diff
@@ -61,7 +61,7 @@ The analyzer will report an error if neither the construtor nor the ToManaged me
 
 If the native type `TNative` also has a public `Value` property, then the value of the `Value` property will be passed to native code instead of the `TNative` value itself. As a result, the type `TNative` will be allowed to be non-blittable and the type of the `Value` property will be required to be blittable. If the `Value` property is settable, then when marshalling in the native-to-managed direction, a default value of `TNative` will have its `Value` property set to the native value. If `Value` does not have a setter, then marshalling from native to managed is not supported.
 
-If the `Value` property has a `ref` or `readonly ref` return type, then the marshalling type is only supported in scenarios where pinning is available. See below for an example of a managed type with a native type with a `Value` property.
+A `ref` or `ref readonly` typed `Value` property is unsupported. If a ref-return is required, the type author can supply a `GetPinnableReference` method on the native type. If a `GetPinnableReference` method is supplied, then the `Value` property must have a pointer-sized primitive type.
```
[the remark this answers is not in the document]
jkoritzinsky: Added

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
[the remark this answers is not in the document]
jkoritzinsky: So currently in the repo we're using operators at the end of the line. Do we want to switch to at the beginning of the line?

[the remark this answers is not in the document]
jkoritzinsky: Actually, we're using a mix. Files are consistent within themselves, but our style is split 50/50 right now.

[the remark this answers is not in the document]
AaronRobinsonMSFT: I would prefer at the beginning of the line.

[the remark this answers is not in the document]
jkoritzinsky: Converted to beginning of the line.

jkoritzinsky:
Does anyone have any more comments/requested changes?

DllImportGenerator/DllImportGenerator.IntegrationTests/NonBlittableStructTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        public static partial IntWrapper DoubleIntRef(IntWrapper pInt);
+
+        [GeneratedDllImport(nameof(NativeExportsNE), EntryPoint = "reverse_replace_ref_ushort")]
+        public static partial void ReverseReplaceString([MarshalUsing(typeof(Utf16StringMarshaler))] ref string s);
+
+        [GeneratedDllImport(nameof(NativeExportsNE), EntryPoint = "return_length_ushort")]
+        public static partial int ReturnStringLength([MarshalUsing(typeof(Utf16StringMarshaler))] string s);
+    }
+
+    public class NonBlittableStructTests
```
elinor-fung, line 40: Since this is testing `MarshalUsing` (not just non-blittable structs), should it be `CustomNativeType`, `CustomMarshalling` or something similar?

DllImportGenerator/DllImportGenerator.UnitTests/ManualTypeMarshallingAnalyzerTests.cs
```diff
@@ -466,6 +466,102 @@ public Native(S s) : this()
...
+    public Native(S s) : this()
+    {
+        value = s;
+    }
+
+    public ref byte Value { get => ref value.GetPinnableReference(); }
+}";
+
+            await VerifyCS.VerifyAnalyzerAsync(source,
+                VerifyCS.Diagnostic(NativeTypeMustBePointerSizedRule).WithSpan(22, 5, 22, 71).WithArguments("ref byte", "S"),
```
elinor-fung, line 529: The analyzer testing APIs support [marking up the source text](https://github.com/dotnet/roslyn-sdk/blob/506767dc02d71f40b179f3e6b1e97b5d0f975b00/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/TestFileMarkupParser.cs) with expected diagnostics. For the tests in roslyn-analyzers, the general guideline is to use the markup if possible.

If doing argument validation, you still need to create/pass in the expected diagnostics, but I do find it simpler to at least not have to deal with the span line/column numbers. `GeneratedDllImportAnalyzerTests` has some tests that use the markup just for the location and combine it with the argument validation.

Not advocating for switching them all in this change - just for future reference.

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
```diff
@@ -0,0 +1,272 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+#nullable enable
+
+namespace Microsoft.Interop
+{
+    class CustomNativeTypeMarshaler : IMarshallingGenerator
```
elinor-fung, line 13: We have been going with a double `l` for marshaller/marshalling (file name and variable suffix too).

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaler.cs
```diff
@@ -0,0 +1,272 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+#nullable enable
```
elinor-fung, line 9: Is this still needed? I thought we had it enabled for all of the `DllImportGenerator` project.

elinor-fung (approved):
Yay!

AaronRobinsonMSFT:
LGTM. I did a quick run through again. I don't see anything that is concerning.

DllImportGenerator/DllImportGenerator.UnitTests/ManualTypeMarshallingAnalyzerTests.cs
jkoritzinsky: I'll do this in a follow up PR

PR #308, opened 2020-11-05
Set up CI/CD pipeline based on the standalone template

jkoritzinsky:
This PR merges in the standalone template that @safern designed and adapts it to enable building DllImportGenerator.

Commits after 3e1b2ad have the changes to the template to integrate into our build; all commits before it are from the template.

There are now `build.cmd` and `build.sh` scripts that are the frontend of the Arcade build in the repo root.

The following commands will be useful from the command line:

- `build.cmd` to build the repo
- `build.cmd -test` to build the repo and run unit tests
- `build.cmd -integrationtest` to build the repo and run integration tests
- `build.cmd -test -integrationtest` to build the repo and run all tests

You'll need to run `git clean -xdf` after pulling down a branch with this work since it changes the output directories to live under `artifacts`, so some files in the old `obj` directories cause confilcts.

Once you've run `build.cmd` from the root of the repo once, you can develop, build, and test the project entirely from within VS or VSCode. The only time you need to run the root build script is when we update the runtime that we run tests on (useful for utilizing runtime bug fixes to improve our tests).

I'm going to request that we do not Squash Merge or Rebase Merge this PR. I want to preserve the history of the standalone template (and I don't want git history to show that I own it).

### DIFF
--- a/DllImportGenerator/TestAssets/NativeExports/NativeExports.csproj
+++ b/DllImportGenerator/TestAssets/NativeExports/NativeExports.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="DNNE" Version="1.0.15" />
+    <PackageReference Include="DNNE" Version="1.0.16" />
   </ItemGroup>
 
   <ItemGroup>


jkoritzinsky:
I may need some help with getting Linux and Mac onboard. Looks like there's something up with the clang support in DNNE that isn't hooked up quite right, but I might be wrong.

global.json
```diff
@@ -0,0 +1,16 @@
+{
```
AaronRobinsonMSFT, line 1: Who updates this file?

AaronRobinsonMSFT (approved):
This looks fine to me. I want to understand ownership here. What files do we own and what files does the engineering team own?

AaronRobinsonMSFT:
>I may need some help with getting Linux and Mac onboard. Looks like there's something up with the clang support in DNNE that isn't hooked up quite right, but I might be wrong.

@jkoritzinsky DNNE assumes that `clang` is on the path. It doesn't try to discover anything.

global.json
```diff
@@ -11,6 +11,6 @@
       }
```
safern, line 11: Do you still need the runtimes section?

jkoritzinsky: Yes, we need the runtimes section to get some bug fixes in the hosting APIs for components. If we didn't need these bug fixes (or if they were in .NET 5), we wouldn't have the section.

global.json
safern: You can add an arcade subscription and then it would be auto update.

safern:
> Do the names in the branch-specific runtimelab.yml files need updating as well?

I added `feature/*` into the branch triggers so that people no longer have to do that.

AaronRobinsonMSFT:
:shipit:

PR #310, opened 2020-11-06
More examples of use within a native library; typos: #elf->#elif, struct population

darderik:
I did a pr a while ago that didn't respect type size consistency between c# and c. I made use only of IntPtr in order to respect it.

Library handle is now defined at the beginning of the program, and casted implicitly if used with LoadLibraryA.
Let me know if there's something that is not needed at all or something else that could be of use.

### DIFF
--- a/samples/NativeLibrary/LoadLibrary.c
+++ b/samples/NativeLibrary/LoadLibrary.c
@@ -37,19 +37,19 @@ static void** handle;
 
 int main()
 {
-    #ifdef _WIN32
-        HINSTANCE hdl = LoadLibraryA(PathToLibrary);
-    #else
-        void *hdl = dlopen(PathToLibrary, RTLD_LAZY);
-    #endif
-    handle = &hdl;
-
     // Check if the library file exists
     if (access(PathToLibrary, F_OK) == -1)
     {
         puts("Couldn't find library at the specified path");
         return 0;
     }
+    
+    #ifdef _WIN32
+        HINSTANCE hdl = LoadLibraryA(PathToLibrary);
+    #else
+        void *hdl = dlopen(PathToLibrary, RTLD_LAZY);
+    #endif
+    handle = &hdl;
 
     // Sum two integers
     int sum = callSumFunc(2, 8);


samples/NativeLibrary/Class1.cs
```diff
@@ -51,5 +52,39 @@ public static IntPtr sumstring(IntPtr first, IntPtr second)
...
+        struct myStruct
+        {
+            public IntPtr name;
+            public int value;
+        }
+
+        [UnmanagedCallersOnly(EntryPoint = "popstruct")]
+        public static void PopulateStruct(IntPtr structurePtr)
+        {
+            myStruct structure = Marshal.PtrToStructure<myStruct>(structurePtr);
```
jkotas, line 84: For blittable struct, it is more appropriate to just use C# pointers. 

```csharp
public static void PopulateStruct(myStruct* structurePtr)
{
    structurePtr->name = Marshal.StringToHGlobalAnsi(Guid.NewGuid().ToString());
    structurePtr->value = new Random().Next();
}
```

samples/NativeLibrary/LoadLibrary.c
```diff
@@ -6,7 +6,7 @@
 //Set this value accordingly to your workspace settings
 #if defined(_WIN32)
 #define PathToLibrary "bin\\Debug\\net5.0\\win-x64\\native\\NativeLibrary.dll"
-#elf defined(__APPLE__)
+#elif defined(__APPLE__)
```
jkotas, line 9: Thanks for fixing this typo!

samples/NativeLibrary/LoadLibrary.c
```diff
@@ -93,3 +108,35 @@ char *callSumStringFunc(char *path, char *funcName, char *firstString, char *sec
...
+    // Declare a typedef
+    typedef void (*myFunc)();
+
+    // Import Symbol named funcName
+    myFunc MyImport = (myFunc)symLoad(*handlePtr, funcName);
+
+    // The C# function will return a pointer
+    MyImport(structure);
+
+    // CoreRT libraries do not support unloading
```
jkotas, line 140: There comments are not relevant anymore now that these methods do not call LoadLibrary.

samples/NativeLibrary/Class1.cs
```diff
@@ -51,5 +52,39 @@ public static IntPtr sumstring(IntPtr first, IntPtr second)
             // Return pointer
             return sumPointer;
         }
+
+        [UnmanagedCallersOnly(EntryPoint = "mergestrings")]
+        public static void MergeStrings(IntPtr oput, IntPtr strArr, int count)
```
jkotas, line 57: This example is using unsecure pattern. The buffer size needs to be passed together with the buffer pointer, and the function needs to handle the situation where the buffer passed in is not large enough.

It would make the sample a lot more complex. I am not sure whether having this sample is worth it.

samples/NativeLibrary/LoadLibrary.c
```diff
@@ -40,29 +42,49 @@ int main()
...
+    printf("struct.name:\n%s\n", newStruct->name);
+    printf("struct.value:\n%d\n", newStruct->value);
+
+    // Free strings
     free(sumstring);
+    free(mergedString);
 }
 
-int callSumFunc(char *path, char *funcName, int firstInt, int secondInt)
+int callSumFunc(char *funcName, int firstInt, int secondInt, void **handlePtr)
```
jkotas, line 84: I think that the real-world would be more likely structured as:

- The signature of the function exposed to the rest of the program would be just `int callSumFunc(int firstInt, int secondInt)`
- HINSTANCE and the function pointers would be cached in static variables

samples/NativeLibrary/Class1.cs
darderik: Definitely better! I just put the method in unsafe context so i can use pointers

samples/NativeLibrary/Class1.cs
darderik: Sorry about that, completely missed that 🤦‍♂️ 

I thought it may have been of use to show how the native library could be used to manipulate existing memory allocation where strings are stored. 
I'm personally using this for creating extensions for a game server (that calls a function passing as parameters a memory location with data and a memory location where 'output'  has to be stored) . 
In case i'll just remove it

PR #316, opened 2020-11-08
Add support for embedded debug symbols

MichalStrehovsky:
Realized that this is missing because the repro project embeds its debug symbols.


MichalStrehovsky:
(Warning: CI is not triggering)

jkotas (approved):
Thank you!

PR #320, opened 2020-11-09
Fix merge conflict in runtimelab.yml

MichalStrehovsky:


MichalStrehovsky:
Sigh. We also picked up a regression somewhere around codegen I guess. I'm investigating.

eng/pipelines/runtimelab.yml
```diff
@@ -84,151 +84,16 @@ stages:
       buildConfig: Release
       platforms:
       - Linux_x64
```
jkotas, line 86: Thanks for fixing it up. I have resolved the first conflict in the file, but forgot to look for more...

jkotas (approved):

MichalStrehovsky:
> 2020-11-09T16:02:02.6131666Z F:\workspace\_work\1\s\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\nativeaot\build\Microsoft.NETCore.Native.Unix.props(98,5): error : Platform linker ('clang') not found. Try installing clang or the appropriate package for your platform to resolve the problem. 

Well okay, I'll just merge this for now because we're producing broken NuGet packages.

Looks like the Windows_NT rename went too rename-happy and also renamed the well-known MSBuild `$(OS)` property that is defined to return Windows_NT by the SDK. I don't feel like digging that details of that out of the 300 file change tonight, but we definitely think we're building on Unix.

Issue #321, opened 2020-11-09
GT_RET_EXPR in fat transform


MichalStrehovsky:
NO_WAY("Cloning of GT_RET_EXPR node not supported");

Hitting an issue in fat transform when running with optimizations on unoptimized IL of this method:

https://github.com/dotnet/runtimelab/blob/cd82da18bdfa0390378682c0811251fca5d75752/src/coreclr/src/nativeaot/Runtime.Base/src/System/Runtime/DispatchResolve.cs#L20-L23

(`DynamicTypeSlotDispatchResolve` is a function pointer, hence this is a `calli`)

We end up with a following tree for the calli:

```
STMT00040 (IL   ???...  ???)
               [000151] -ACXG-------              *  ASG       long
               [000150] D------N----              +--*  LCL_VAR   long   V20 tmp5
               [000149] --CXG-------              \--*  CALL ind  long
               [000145] ------------ arg0            +--*  LCL_VAR   long   V19 tmp4
               [000146] --C--------- arg1            +--*  RET_EXPR  long  (inl return from call [000142])
               [000147] ------------ arg2            +--*  LCL_VAR   ushort V02 arg2
               [000148] ------------ calli tgt       \--*  LCL_VAR   long   V06 loc3
```

The inl return is a call to the IntPtr casting operator - note the C# source has two casts like this, but one of them is spilled into a temporary, so the tree has a load from the temporary instead of the RET_EXPR.

`FatPointerCallTransformer::CreateThen` (also `CreateElse`, I assume), is trying to clone this tree and hitting the `NO_WAY("Cloning of GT_RET_EXPR node not supported");`. I assume we need to spill arguments into temps first?

This problem is new after #191 - before that, the calli was in a separate method and the casts happened elsewhere.

Thanks for C# 9.0 introducing direct support for function pointers, this can be hit in customer code too.

MichalStrehovsky:
`GuardedDevirtualizationTransformer` uses `gtCloneCandidateCall` instead of `gtCloneExpr` which apparently is the thing that prevents issues with GT_RET_EXPR. Not sure what the "appropriate other fixup" will be tho, so maybe let's leave this to the JIT team after the holidays...

```
            // We can't safely clone calls that have GT_RET_EXPRs via gtCloneExpr.
            // You must use gtCloneCandidateCall for these calls (and then do appropriate other fixup)
```

MichalStrehovsky:
@sandreenko is this something you could have a look at?

[...]

PR #322, opened 2020-11-09
Improve type information in function pointers

MichalStrehovsky:
Works around #321, but coincidentally also improves readabiliy by getting rid of unnecessary casting.


MichalStrehovsky:
(CI didn't trigger. Do not merge yet.)

src/coreclr/src/nativeaot/Runtime.Base/src/System/Runtime/DispatchResolve.cs
```diff
@@ -19,7 +19,7 @@ internal static unsafe class DispatchResolve
             // Use the dynamic module resolver if it's present
             if ((dynamicModule != null) && (dynamicModule->DynamicTypeSlotDispatchResolve != null))
```
jkotas, line 20: Maybe we can also cache `DynamicTypeSlotDispatchResolve` in a local to avoid calling the property twice?

jkotas (approved):

PR #324, opened 2020-11-09
Rename Windows_NT to windows for TargetOS

jkotas:


MichalStrehovsky (approved):
I don't know much about the TargetOS thing - I'm just worried about how this integrates E2E - we don't have testing for that (e.g. if I use ILCompiler package with the 3.1 SDK). I guess we'll find out.

jkotas:
TargetOS is not a .NET SDK thing.

src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -20,8 +20,9 @@ The .NET Foundation licenses this file to you under the MIT license.
     <NativeOutputPath Condition="'$(NativeOutputPath)' == ''">$(OutputPath)native\</NativeOutputPath>
     <NativeCompilationDuringPublish Condition="'$(NativeCompilationDuringPublish)' == ''">true</NativeCompilationDuringPublish>
     <IlcBuildTasksPath Condition="'$(IlcBuildTasksPath)' == ''">$(MSBuildThisFileDirectory)..\tools\netstandard\ILCompiler.Build.Tasks.dll</IlcBuildTasksPath>
+    <TargetOS Condition="'$([MSBuild]::IsOSPlatform(Windows))' == 'true'">windows</TargetOS>
     <TargetOS Condition="'$([MSBuild]::IsOSPlatform(OSX))' == 'true'">OSX</TargetOS>
-    <TargetOS Condition="'$(TargetOS)' == ''">$(OS)</TargetOS>
+    <TargetOS Condition="'$(TargetOS)' == ''">$(TargetOS)</TargetOS>
```
MichalStrehovsky, line 25: I thought TargetOS is an SDK thing based on this line, but on a closer look, this doesn't look right.

jkotas: https://github.com/dotnet/runtimelab/pull/326

PR #325, opened 2020-11-09
Maybe fix the CI

MichalStrehovsky:


src/tests/run.proj
```diff
@@ -408,7 +408,7 @@ namespace $([System.String]::Copy($(Category)).Replace(".","_").Replace("\","").
   <Target Name="NativeAotCompileFramework" AfterTargets="CopyDependencyToCoreRoot" Condition="'$(NativeAotMultimodule)' == 'true'">
     <MSBuild Projects="$(CORE_ROOT)/nativeaot/build/BuildFrameworkNativeObjects.proj"
       Targets="CreateLib"
-      Properties="FrameworkLibPath=$(CORE_ROOT)/nativeaot/sdk;FrameworkObjPath=$(BaseIntermediateOutputPath)/NativeAOTFX;IlcPath=$(CORE_ROOT)/nativeaot" />
+      Properties="FrameworkLibPath=$(CORE_ROOT)/nativeaot/sdk;FrameworkObjPath=$(BaseIntermediateOutputPath)/NativeAOTFX;IlcPath=$(CORE_ROOT)/nativeaot;TargetOS=$(OS)" />
```
jkotas, line 411: I think this will likely break OSX. TargetOS is not the same namespace as OS.

I have renamed TargetOS in BuildIntegration to match what it means in the rest of the repo: https://github.com/dotnet/runtimelab/pull/324 .

MichalStrehovsky: It wouldn't break things much because we only test multifile on Windows but your change looks better (modulo the "windows" with lowercase "w" but that tasteless choice was made in the runtime repo).

PR #330, opened 2020-11-09
Make JsonSerializableAttribute applicable to modules, not types

layomia:
Resolves https://github.com/dotnet/runtimelab/projects/1#card-48358025.

### DIFF
--- a/src/libraries/System.Text.Json/System.Text.Json.SourceGeneration.Tests/TestClasses.cs
+++ b/src/libraries/System.Text.Json/System.Text.Json.SourceGeneration.Tests/TestClasses.cs
@@ -1,13 +1,22 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿using System.Collections.Generic;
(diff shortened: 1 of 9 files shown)


src/libraries/System.Text.Json/System.Text.Json.SourceGeneration.Tests/TestClasses.cs
```diff
@@ -1,13 +1,22 @@
...
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿using System.Collections.Generic;
 using System.Text.Json.Serialization;
 
+using TestsNamespace = System.Text.Json.SourceGeneration.Tests;
+
+[module: JsonSerializable(typeof(TestsNamespace.RepeatedTypes.Location))]
```
steveharter, line 6: Any reason why not "assembly:" instead?

layomia: Per offline discussion there's really no reason to since the natural application in C# is `assembly`, I'll make a PR changing this.

PR #333, opened 2020-11-10
Harmonize pointer type handling

MichalStrehovsky:
CoreCLR treats pointer types as UIntPtr in a couple places.


jkotas:
This is failing in non-obvious ways. Would it be better to make the change in getClassAttribs in dotnet/runtime first? We have better test coverage for crossgen2.

jkotas (approved):

MichalStrehovsky:
This will come as dotnet/runtime#44465.

PR #345, opened 2020-11-12
Fix s390x frame unwinding logic

nealef:
Previously s390x had relied on the use of a backchain field in the save area (enabled by a GCC flag). This has not been standard practice in >10 years. This fix eliminates that requirement and corrects the CFI being generated. Without this fix running the runtime on s390x would periodically crash due to incorrect unwinding.


akoeplinger:
This was merged as part of https://github.com/dotnet/runtimelab/commit/4a24095db02e9a3d4b50cfc6a3e428d726f601f8.

PR #347, opened 2020-11-12
Hook up dataflow analysis into compilation

MichalStrehovsky:
With this, we can analyze reflection use using the dataflow analysis attributes introduced in .NET 5 (`DynamicallyAccessedMembersAttribute`).

We introduce a new node type (`DataflowAnalyzedMethodNode`) to make sure every method only gets analyzed once. The dependencies discovered by dataflow analysis are the depednencies of this node.

The node is injected into the graph by the metadata manager if we're running in the mode that collects metadata. Codegen backends (and scanners) need to call into this whenever they see a write to a field, or a method call. With RyuJIT, we don't have such granularity, so we also look at field loads and delegate creation sequences. The metadata manager only injects the dataflow analysis dependency if it sees the called method/written field are "interesting". This way we don't have to analyze many method bodies.

### DIFF
--- a/src/tests/nativeaot/SmokeTests/Dataflow/Dataflow.csproj
+++ b/src/tests/nativeaot/SmokeTests/Dataflow/Dataflow.csproj
@@ -4,6 +4,9 @@
     <CLRTestKind>BuildAndRun</CLRTestKind>
     <CLRTestPriority>0</CLRTestPriority>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+    <!-- We don't run the scanner in optimized builds -->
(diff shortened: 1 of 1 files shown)


jkotas (approved):

jkotas:
The CI has some failing tests. Also, every tests has set of new warnings/errors like `EXEC : warning : Method `string..ctor(char*)` will always throw because: Invalid IL or CLR metadata in 'Void System.String..ctor(Char*)`

MichalStrehovsky:
> EXEC : warning : Method string..ctor(char*) will always throw because: Invalid IL or CLR metadata in 'Void System.String..ctor(Char*)

Heh, I'll take a look. I guess we need to move the handling of "string constructors are magic" to some common place so that this can see that too.

PR #350, opened 2020-11-12
Fix official build by shortening prerelease label length

joperezr:
cc: @layomia @safern 

Official builds are failing due to NuGet requiring prerelease label length to be < 20 chars. This change will make it so that's the case.


layomia (approved):
Thanks!

joperezr:
Given this is a small change only in prerelease label, I'll go ahead and merge to unblock official builds without waiting for CI to finish.

PR #352, opened 2020-11-13
Add note about prerelease label length

safern:
A couple of experiments just hit this, JsonCodeGen and ManagedQuic.
cc: @jkotas @joperezr


joperezr (approved):
Thanks for adding it @safern

jkotas (approved):
Thanks!

Issue #200, opened 2020-10-11
Disallow reflection setting readonly static fields


MichalStrehovsky:
reflection/SetValue/TrySetReadonlyStaticField
reflection/SetValue/TrySetReadonlyStaticField2

MichalStrehovsky:
Add `IsInitOnly` flag to the FieldTableFlags enum:

https://github.com/dotnet/runtimelab/blob/beef8b0fad689f8d7a7f7b2ec0f72263698e658e/src/coreclr/src/tools/Common/Internal/Runtime/MappingTableFlags.cs#L40-L54

Next, emit the flag from the AOT compiler - we're computing the value for the flags around here:

https://github.com/dotnet/runtimelab/blob/beef8b0fad689f8d7a7f7b2ec0f72263698e658e/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ReflectionFieldMapNode.cs#L86-L93

Add a line that sets the flag based on `field.IsInitOnly`.

Now we just need to respect that flag at runtime. The code that interprets it is here:

https://github.com/dotnet/runtimelab/blob/beef8b0fad689f8d7a7f7b2ec0f72263698e658e/src/coreclr/src/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/ExecutionEnvironmentImplementation.MappingTables.cs#L1376-L1380

Here we're passing `fieldBase` to the accessors. `fieldBase` is a stripped version of `fieldAccessMetadata.Flags` - change the code to pass `fieldAccessMetadata.Flags` instead so that the accessors get the full flags, not just the lowest two bits. Note that the code expect the rest of the flags to be dropped, so check for places that do `==` with these values - they'll have to do the `& FieldTableFlags.StorageClass ` mask first

Then update the `StaticFieldAccessor` class to throw if we're trying to set a field that has the IsInitOnly flag set.

rohithv1997:
@MichalStrehovsky - I would like to take this up if no one else is assigned on this

MichalStrehovsky:
@rohithv1997 assigned to you! Thank you!

rohithv1997:
Hi @MichalStrehovsky ,
Can you let me know how to change the status of this bug?

MichalStrehovsky:
> Can you let me know how to change the status of this bug?

[...]

PR #353, opened 2020-11-14
Fixed https://github.com/dotnet/runtimelab/issues/200

rohithv1997:
Fixed https://github.com/dotnet/runtimelab/issues/200

Approach :
At RegularStaticFieldsAccessor, created a private readonly property for FieldBaseFlags.
Exposed an expression body member StrippedFieldBase that masks with storage flag.
At StaticFieldAccessor, if field is marked as InitOnly, throw an InvalidOperationException

### DIFF
--- a/src/tests/issues.targets
+++ b/src/tests/issues.targets
@@ -1368,6 +1368,9 @@
         <ExcludeList Include="$(XunitTestBinBase)/reflection/Modifiers/modifiers/*">
             <Issue>https://github.com/dotnet/runtimelab/issues/201</Issue>
         </ExcludeList>
+        <ExcludeList Include="$(XunitTestBinBase)/reflection/SetValue/TrySetReadonlyStaticField/*">
+            <Issue>https://github.com/dotnet/runtimelab/issues/200</Issue>
+        </ExcludeList>
         <ExcludeList Include="$(XunitTestBinBase)/tracing/**/*">
             <Issue>EventPipe</Issue>
         </ExcludeList>


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=353) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=353) before we can accept your contribution.<br/><br/>:x: rohithv1997 [sign now](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=353)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/runtimelab?pullRequest=353) it.</sub>

src/coreclr/src/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/FieldAccessors/StaticFieldAccessor.cs
```diff
@@ -51,6 +56,11 @@ public sealed override void SetFieldDirect(TypedReference typedReference, object
             {
                 RuntimeAugments.EnsureClassConstructorRun(_cctorContext);
             }
+            if (IsFieldInitOnly)
```
MichalStrehovsky, line 59: Could you move this check to `WritableStaticFieldAccessor.SetFieldBypassCctor` (same for the Direct version.

We want to make sure that:

```csharp
typeof(Foo).GetField("F").SetValue(null, "hello");

struct Foo
{
    public static readonly int F = 12;
}
```

Will keep throwing the "System.ArgumentException: 'Object of type 'System.String' cannot be converted to type 'System.Int32'.'" exception (i.e. check the value being set first).

src/coreclr/src/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/FieldAccessors/StaticFieldAccessor.cs
```diff
@@ -51,6 +56,11 @@ public sealed override void SetFieldDirect(TypedReference typedReference, object
...
+            {
+                Debug.Fail("Cannot set field with InitOnly flag set.");
+                throw new InvalidOperationException();
```
MichalStrehovsky, line 62: Could you make this a FieldAccessException? It's what CoreCLR throws. Also, no need for the Debug.Fail - this is not breaking internal invariants - it's simply invalid use of the API that we handle.

Let's put "Cannot set initonly static field after its owning type is initialized.'" as the message.

Could you also put the message into a resource string? LiteralFieldAccessor.cs has a sample - you basically need to add it to Strings.resx of the project.

src/coreclr/src/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/FieldAccessors/RegularStaticFieldAccessor.cs
```diff
@@ -12,12 +12,14 @@ protected RegularStaticFieldAccessor(IntPtr cctorContext, IntPtr staticsBase, in
             : base(cctorContext, fieldTypeHandle)
         {
             StaticsBase = staticsBase;
-            FieldBase = fieldBase;
+            _fieldBase = fieldBase;
```
MichalStrehovsky, line 15:
```suggestion
            _fieldFlags = fieldFlags;
```

Might be better because it's no longer just the information about the base of the field.

src/coreclr/src/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/FieldAccessors/RegularStaticFieldAccessor.cs
```diff
@@ -12,12 +12,14 @@ protected RegularStaticFieldAccessor(IntPtr cctorContext, IntPtr staticsBase, in
...
             FieldOffset = fieldOffset;
         }
 
         protected IntPtr StaticsBase { get; }
-        protected FieldTableFlags FieldBase { get; }
+        private readonly FieldTableFlags _fieldBase;
         protected int FieldOffset { get; }
+        protected FieldTableFlags StrippedFieldBase => _fieldBase & FieldTableFlags.StorageClass;
```
MichalStrehovsky, line 22:
```suggestion
        protected FieldTableFlags FieldBase => _fieldBase & FieldTableFlags.StorageClass;
```

I would call it this since these are the flags that correspond to the actual base.

src/coreclr/src/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/FieldAccessors/RegularStaticFieldAccessor.cs
```diff
@@ -12,12 +12,14 @@ protected RegularStaticFieldAccessor(IntPtr cctorContext, IntPtr staticsBase, in
...
+        protected sealed override bool IsFieldInitOnly => _fieldBase == FieldTableFlags.IsInitOnly;
```
MichalStrehovsky, line 23:
```suggestion
        protected sealed override bool IsFieldInitOnly => (_fieldBase & FieldTableFlags.IsInitOnly) == FieldTableFlags.IsInitOnly;
```

MichalStrehovsky (commented):
A couple things, but looks great otherwise. Thanks!

rohithv1997:
Hi @MichalStrehovsky ,
I have implemented the review changes in the commit https://github.com/rohithv1997/runtimelab/commit/4d3b9a89a336f351fc6be1012499cdcc71b02383

src/coreclr/src/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/FieldAccessors/StaticFieldAccessor.cs
```diff
@@ -66,5 +66,6 @@ public sealed override int Offset
         protected abstract object GetFieldBypassCctor();
         protected abstract void SetFieldBypassCctor(object value, BinderBundle binderBundle);
         protected abstract void SetFieldDirectBypassCctor(object value);
+        protected abstract bool IsFieldInitOnly { get; }
```
MichalStrehovsky, line 69: Looks like we don't need this here - maybe introduce the abstract method on the WritableStaticFieldAccessor so that we don't need to make changes in this file and in the literal accessor?

src/coreclr/src/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/FieldAccessors/WritableStaticFieldAccessor.cs
```diff
@@ -16,14 +16,26 @@ protected WritableStaticFieldAccessor(IntPtr cctorContext, RuntimeTypeHandle fie
 
         protected abstract override object GetFieldBypassCctor();
 
+        protected abstract override bool IsFieldInitOnly { get; }
+
         protected sealed override void SetFieldBypassCctor(object value, BinderBundle binderBundle)
         {
+            if (IsFieldInitOnly)
```
MichalStrehovsky, line 23: Could you move the CheckArgument call before this line?

(Same for the Direct version below)

MichalStrehovsky (commented):
Looks great otherwise. Thank you!

rohithv1997:
Hi @MichalStrehovsky,
I have implemented the feedback changes in the commit https://github.com/rohithv1997/runtimelab/commit/a635ad998027d2d9b05d2842b5478ab476ea754b.
Thank you!

MichalStrehovsky (approved):
Looks good, thank you! I kicked off a full validation pass to see if the tests now work (some failures are expected right now, so don't worry if it's not green - I'll take a look).

rohithv1997:
Thank you for the guidance, @MichalStrehovsky !

MichalStrehovsky:
Had to leave one of the tests disabled because we apparently should allow setting the value while the class constructor is still executing. Not sure I care about that detail much, so this is good to merge. Thank you!

PR #358, opened 2020-11-16
Add dataflow from custom attribute instances

MichalStrehovsky:
This is to allow flowing values from custom attribute instances into annotated places on the custom attribute type (basically, we pretend an applied custom attribute is a method call/field set operation for interesting callees/storage locations).

### DIFF
--- a/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
@@ -13,6 +13,9 @@ using Internal.TypeSystem;
 using BindingFlags = System.Reflection.BindingFlags;
 using NodeFactory = ILCompiler.DependencyAnalysis.NodeFactory;
 using DependencyList = ILCompiler.DependencyAnalysisFramework.DependencyNodeCore<ILCompiler.DependencyAnalysis.NodeFactory>.DependencyList;
+using CustomAttributeValue = System.Reflection.Metadata.CustomAttributeValue<Internal.TypeSystem.TypeDesc>;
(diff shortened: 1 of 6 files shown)


src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -97,63 +100,116 @@ public static DependencyList ScanAndProcessReturnValue(NodeFactory factory, Flow
...
+                                result = scanner._dependencies;
+                            }
+                            else
+                            {
+                                result.AddRange(scanner._dependencies);
+                            }
+                        }
+                        finally
+                        {
+                            reflectionContext.Dispose();
```
jkotas, line 188: This makes it look like that `ReflectionPatternContext` has some resources that have to be reclaimed, but it is actually just a debug check.

Instead of finally block and Dispose, would it be better to have to debug-only method to do the debug-only check?

jkotas: (Ignore this if this is just mirroring ILLinker structure.)

jkotas (approved):

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
MichalStrehovsky: Yeah, it's from IL linker. The `ReflectionPatternContext` pre-dates the new dataflow analysis (it was introduced for the peephole scanner). It deals with the fact that one of the places that uses it has many exits from the method and it's a chore to manually call it in each of those spots.

`IDisposable` sounds good in theory (could be used with `using`), but then we have to pass the thing as a `ref` and C# forbids passing things we're `using` like that, leaving us with manual `finally`.

PR #362, opened 2020-11-18
Do not force interface calls as virtual

MichalStrehovsky:
This logic was removed from CoreCLR in .NET 3.0 timeframe and is going to be a problem for default interface methods.


jkotas (approved):
Thank you!

PR #364, opened 2020-11-18
Cache result of MightHaveDispatchMap

MichalStrehovsky:
Noticed this by accident.

We call this from multiple places within EEType construction, some of which are executed multiple times (especially the occurrence within `ComputeOptionalEETypeFields` that is called way more often than it really should be). We should just call once and cache the result. Computing it is expensive because it requires resolving all interface methods on the type to their implementation.

Saves about 2-3% of wallclock time within the AOT compiler.

### DIFF
--- a/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/CanonicalEETypeNode.cs
+++ b/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/CanonicalEETypeNode.cs
@@ -44,7 +44,7 @@ namespace ILCompiler.DependencyAnalysis
 
             DefType closestDefType = _type.GetClosestDefType();
 
-            if (_mightHaveInterfaceDispatchMap)
+            if (MightHaveInterfaceDispatchMap(factory))
                 dependencyList.Add(factory.InterfaceDispatchMap(_type), "Canonical interface dispatch map");
 
             dependencyList.Add(factory.VTable(closestDefType), "VTable");
--- a/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
+++ b/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
@@ -79,7 +79,7 @@ namespace ILCompiler.DependencyAnalysis
 
             DefType closestDefType = _type.GetClosestDefType();
 
-            if (_mightHaveInterfaceDispatchMap)
+            if (MightHaveInterfaceDispatchMap(factory))
             {
                 dependencyList.Add(factory.InterfaceDispatchMap(_type), "Interface dispatch map");
             }
--- a/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
+++ b/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
@@ -60,7 +60,7 @@ namespace ILCompiler.DependencyAnalysis
         protected readonly TypeDesc _type;
         internal readonly EETypeOptionalFieldsBuilder _optionalFieldsBuilder = new EETypeOptionalFieldsBuilder();
         internal readonly EETypeOptionalFieldsNode _optionalFieldsNode;
-        protected readonly bool _mightHaveInterfaceDispatchMap;
+        protected bool? _mightHaveInterfaceDispatchMap;
 
         public EETypeNode(NodeFactory factory, TypeDesc type)
         {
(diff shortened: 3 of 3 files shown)


jkotas (approved):
Thank you!

MichalStrehovsky:
Sigh, the CI is all red because in optimized builds we sometimes create constructed EEType nodes for things that were not constructed before, just to check whether the node is marked - it isn't marked. I guess I'll need to make this `bool` a `bool?` and initialize lazily first time we need it.

I've already switched to other things, so this will wait here for a bit...

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/CanonicalEETypeNode.cs
```diff
@@ -44,7 +44,7 @@ protected override DependencyList ComputeNonRelocationBasedDependencies(NodeFact
 
             DefType closestDefType = _type.GetClosestDefType();
 
-            if (InterfaceDispatchMapNode.MightHaveInterfaceDispatchMap(_type, factory))
+            if (_mightHaveInterfaceDispatchMap)
```
MichalStrehovsky, line 47:
```suggestion
            if (MightHaveInterfaceDispatchMap(factory))
```

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
```diff
@@ -79,7 +79,7 @@ protected override DependencyList ComputeNonRelocationBasedDependencies(NodeFact
 
             DefType closestDefType = _type.GetClosestDefType();
 
-            if (InterfaceDispatchMapNode.MightHaveInterfaceDispatchMap(_type, factory))
+            if (_mightHaveInterfaceDispatchMap)
```
MichalStrehovsky, line 82:
```suggestion
            if (MightHaveInterfaceDispatchMap(factory))
```

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -57,9 +57,10 @@ namespace ILCompiler.DependencyAnalysis
     /// </summary>
     public partial class EETypeNode : ObjectNode, IExportableSymbolNode, IEETypeNode, ISymbolDefinitionNode, ISymbolNodeWithLinkage
     {
-        protected TypeDesc _type;
-        internal EETypeOptionalFieldsBuilder _optionalFieldsBuilder = new EETypeOptionalFieldsBuilder();
-        internal EETypeOptionalFieldsNode _optionalFieldsNode;
+        protected readonly TypeDesc _type;
+        internal readonly EETypeOptionalFieldsBuilder _optionalFieldsBuilder = new EETypeOptionalFieldsBuilder();
+        internal readonly EETypeOptionalFieldsNode _optionalFieldsNode;
+        protected readonly bool _mightHaveInterfaceDispatchMap;
```
MichalStrehovsky, line 63:
```suggestion
        protected bool? _mightHaveInterfaceDispatchMap;
```

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -72,6 +73,8 @@ public EETypeNode(NodeFactory factory, TypeDesc type)
             _type = type;
             _optionalFieldsNode = new EETypeOptionalFieldsNode(this);
 
+            _mightHaveInterfaceDispatchMap = EmitVirtualSlotsAndInterfaces && InterfaceDispatchMapNode.MightHaveInterfaceDispatchMap(type, factory);
+
```
MichalStrehovsky, line 77:
```suggestion
```

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -882,7 +885,7 @@ private void OutputGenericInstantiationDetails(NodeFactory factory, ref ObjectDa
         /// </summary>
         protected internal virtual void ComputeOptionalEETypeFields(NodeFactory factory, bool relocsOnly)
         {
-            if (!relocsOnly && EmitVirtualSlotsAndInterfaces && InterfaceDispatchMapNode.MightHaveInterfaceDispatchMap(_type, factory))
+            if (!relocsOnly && _mightHaveInterfaceDispatchMap)
```
MichalStrehovsky, line 888:
```suggestion
            if (!relocsOnly && MightHaveInterfaceDispatchMap(factory))
```

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -72,6 +73,8 @@ public EETypeNode(NodeFactory factory, TypeDesc type)
             _type = type;
             _optionalFieldsNode = new EETypeOptionalFieldsNode(this);
 
+            _mightHaveInterfaceDispatchMap = EmitVirtualSlotsAndInterfaces && InterfaceDispatchMapNode.MightHaveInterfaceDispatchMap(type, factory);
+
             factory.TypeSystemContext.EnsureLoadableType(type);
         }
```
MichalStrehovsky, line 79:
```suggestion
        }
        
        protected bool MightHaveInterfaceDispatchMap(NodeFactory factory)
        {
            if (!_mightHaveInterfaceDispatchMap.HasValue)
            {
                _mightHaveInterfaceDispatchMap = EmitVirtualSlotsAndInterfaces && InterfaceDispatchMapNode.MightHaveInterfaceDispatchMap(_type, factory);
            }
            
            return _mightHaveInterfaceDispatchMap.Value;
        }
```

PR #365, opened 2020-11-18
Speed up the AOT compiler by 10%

MichalStrehovsky:
A ~6.6 second AOT compilation goes down to ~5.9 seconds..


MichalStrehovsky:
Cc @dotnet/crossgen-contrib - I assume there might be similar wins to be had on crossgen2, but I don't have the cycles to spend measuring it.

MichalStrehovsky:
I filed https://github.com/dotnet/runtime/issues/44890 for crossgen2 so that it's not lost.

PR #368, opened 2020-11-19
Initial default interface method support

MichalStrehovsky:
Support for default interface methods in the compiler and runtime.

It felt like the most straightforward approach is to just give interface types a dispatch map and populate vtables of interface method with the default implementation. I didn't realize they get a full vtable with a bunch of null slots. This might be something we could potentially optimize, but it's orthogonal.

There's plenty of things missing still, but it's a start and tests are starting to pass.

### DIFF
--- a/src/tests/issues.targets
+++ b/src/tests/issues.targets
@@ -1448,6 +1448,7 @@
         <ExcludeList Include="$(XunitTestBinBase)/Regressions\coreclr\16355\variance\*" />
         <ExcludeList Include="$(XunitTestBinBase)/Regressions\coreclr\16775\sharedinterfacemethod\*" />
         <ExcludeList Include="$(XunitTestBinBase)/Regressions\coreclr\20452\twopassvariance\*" />
+        <ExcludeList Include="$(XunitTestBinBase)/Regressions\coreclr\22021\consumer\*" />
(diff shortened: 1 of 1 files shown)


jkotas (approved):
Thank you!

jkotas:
What is the type loader behavior when it hits a type with DIMs? It would be nice if it was a clean crash, and not a bad AV somewhere.

jkotas:
cc @davidwrighton

MichalStrehovsky:
> What is the type loader behavior when it hits a type with DIMs? It would be nice if it was a clean crash, and not a bad AV somewhere.

The type loader is not involved right now since I'm not attempting to solve the "we need an instantiating stub for shared methods on generic interfaces". Right now we would just end up calling a shared method without its generic context.

I pushed out an improvement that treats these as a reabstraction instead (so we never end up placing the canonical method without a generic context into the vtable).

I also added a couple very basic tests so that we can smoke test this. We should definitely do a Pri-0 test pass before merging.

MichalStrehovsky:
The StaticInterfaceMembers failure was also present in the https://github.com/dotnet/runtimelab/runs/1405225153 run so it's unrelated. I'll take a look when we clean up the Pri-0 runs again. I couldn't get it to repro in single file, so it's probably a multifile specific wart.

PR #369, opened 2020-11-20
Print linker logging only in verbose mode

jkotas:


MichalStrehovsky (approved):
Thanks. I was mildly annoyed by these as well, but not enough to do something about it. They served as a reminder for "we really should stop generating metadata for DebuggerDisplayAttribute", because most of the hits for this come from debugger display classes because of how DebuggerDisplay is annotated.

PR #371, opened 2020-11-20
Fix publishing intermediate artifacts.

jkoritzinsky:
One of the parameter names changed, so runtimelab.yml needs to respond to it.


jkotas (approved):
Thank you!

PR #372, opened 2020-11-20
Add support for dataflow analysis to generic parameters

MichalStrehovsky:
This will ensure that if e.g. a T in a generic type is annotated as "keep all methods on all the T's that this type is instantiated with", the compiler will do that.

What's still missing is validation that the generic annotations are consistent (e.g. if I have `Foo<[DynamicallyAccessed(PublicMethods] T>` and then I do `class Other<TOther> { Blah() => new Foo<TOther>()`, I will get a warning because TOther is not annotated the same as the T in Foo and we potentially broke the promise.

I think we can do this by hooking into the generic lookup tracking logic, but I don't want to add that now because I haven't figured out how to test that (we need to test a warning gets emitted).

### DIFF
--- a/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/FlowAnnotations.cs
+++ b/src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/FlowAnnotations.cs
@@ -540,7 +540,7 @@ namespace ILCompiler.Dataflow
             private readonly FieldAnnotation[] _annotatedFields;
             private readonly DynamicallyAccessedMemberTypes[] _genericParameterAnnotations;
 
-            public bool IsDefault => _annotatedFields == null && _annotatedFields == null && _genericParameterAnnotations == null;
+            public bool IsDefault => _annotatedMethods == null && _annotatedFields == null && _genericParameterAnnotations == null;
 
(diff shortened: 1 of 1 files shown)


src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/FlowAnnotations.cs
```diff
@@ -533,6 +540,8 @@ private class TypeAnnotations
             private readonly FieldAnnotation[] _annotatedFields;
             private readonly DynamicallyAccessedMemberTypes[] _genericParameterAnnotations;
 
+            public bool IsDefault => _annotatedFields == null && _annotatedFields == null && _genericParameterAnnotations == null;
```
jkotas, line 543: Typo: This is testing `_annotatedFields` twice.

jkotas (approved):
Thank you!

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/FlowAnnotations.cs
MichalStrehovsky: Oops, good catch!

PR #374, opened 2020-11-21
Merge from dotnet/runtime

jkotas:


jkotas:
@sandreenko Many tests are failing in this integration with:
```
/Users/runner/work/1/s/src/coreclr/src/jit/assertionprop.cpp:1628
2020-11-21T08:09:55.0171590Z           Assertion failed '(lvaTable[assertion->op1.lcl.lclNum].lvType != TYP_REF) || (assertion->op2.u1.iconVal == 0)' in 
```
I suspect that it is caused by https://github.com/dotnet/runtime/pull/44398. Could you please take a look?

sandreenko:
Sure, could I just follow what CI is doing for a repro or is there a faster way to debug this repo?

jkotas:
I have extracted a simple repro at:
```
xcopy /s \\jkotas9\drops\runtimelab-374 .
runme.bat
```

Note that the JIT .dll is renamed to `clrjitilc.dll`, so if you build a private you need to replace the clrjitilc.dll with it.

jkotas:
If you would prefer to just build locally, here is the workflow that I use (based on https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/docs/workflow/building/coreclr/nativeaot.md):

- `build nativeaot+libs -lc release -rc checked`
- open `src\coreclr\src\tools\aot\ilc.sln` in VS
- set configuration to checked x64
- in the project debug tab, set Application Arguments to `@C:\runtimelab\artifacts\bin\repro\x64\Checked\compile-with-Release-libs.rsp` (or wherever the project is on your machine) and also check "Enable native debugging"
- Append `-O` to `C:\runtimelab\artifacts\bin\repro\x64\Checked\compile-with-Release-libs.rsp`
- Change Program.cs in repro project to:
```csharp
using System;
using System.Threading;

new ThreadLocal<int>().ToString();
```
F5

You should see the assert

PR #377, opened 2020-11-22
Apply SkipLocalsInitAttribute to generated stubs.

AaronRobinsonMSFT:


DllImportGenerator/DllImportGenerator/DllImportStub.cs
```diff
@@ -202,6 +206,22 @@ public class GeneratedDllImportData
             var stubGenerator = new StubCodeGenerator(method, dllImportData, paramsTypeInfo, retTypeInfo, diagnostics);
             var (code, dllImport) = stubGenerator.GenerateSyntax();
 
+            var additionalAttrs = new List<AttributeListSyntax>();
+
+            // Define additional attributes for the stub definition.
+            if (env.TargetFrameworkVersion >= new Version(5, 0))
+            {
+                additionalAttrs.Add(
+                    AttributeList(
```
jkoritzinsky, line 215: Any particular reason other than style that we're going to emit the additional attributes using the  `[Attr1] [Attr2]` syntax instead of the `[Attr1, Attr2]` syntax?

AaronRobinsonMSFT: I don't think we are. This was simply the easiest way to express the scenario here. I can't create a syntax list with no elements - that is invalid C#. We are only exposing a single additional attribute right now and if we have another unconditional then we can group them I personally don't like the the attribute lists as I think it is a confusion about information density and "fewer characters to type is better".

All that said, if you have a suggestion on how to express this that doesn't introduce the nullable nonsense I am all for it.

jkoritzinsky: The only way I can think of requires a nullable and I agree with your rationale around readability for more complicated attributes (which I could see us adding in the future), so I'm ok keeping this how it is.

jkoritzinsky (approved):

PR #379, opened 2020-11-23
Support arrays of types with simple custom marshalling

jkoritzinsky:
Add support for marshalling arrays of types with simple custom marshalling (no Value property) and gracefully fail when an array of a type with complex custom marshalling (Value property) is introduced.

To do so, we create a new `ArrayMarshallingInfo` subtype of `MarshallingInfo` for the default array marshalling case and pre-calculate the element type's marshalling information instead of calculating it in MarshallingGenerator.cs

<summary>Snippet example (BoolStruct is defined as in the SharedTypes library)</summary>
<details>
User-Provided:

```csharp
            [GeneratedDllImport(nameof(NativeExportsNE), EntryPoint = "and_all_members")]
            [return:MarshalAs(UnmanagedType.U1)]
            public static partial bool AndAllMembers(BoolStruct[] pArray, int length);
```

Generated:

```csharp

namespace DllImportGenerator.IntegrationTests
{
    partial class NativeExportsNE
    {
        public partial class Arrays
        {
            public static partial bool AndAllMembers(global::SharedTypes.BoolStruct[] pArray, int length)
            {
                unsafe
                {
                    global::SharedTypes.BoolStructNative*__pArray_gen_native = default;
                    bool __retVal = default;
                    byte __retVal_gen_native = default;
                    //
                    // Setup
                    //
                    bool pArray__allocated = false;
                    try
                    {
                        //
                        // Marshal
                        //
                        if (pArray == null)
                        {
                            __pArray_gen_native = null;
                        }
                        else
                        {
                            int pArray__bytelen = sizeof(global::SharedTypes.BoolStructNative) * pArray.Length;
                            if (pArray__bytelen > 512)
                            {
                                __pArray_gen_native = (global::SharedTypes.BoolStructNative*)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(pArray__bytelen);
                                pArray__allocated = true;
                            }
                            else
                            {
                                byte *pArray__stackptr = stackalloc byte[pArray__bytelen];
                                ;
                                __pArray_gen_native = (global::SharedTypes.BoolStructNative*)pArray__stackptr;
                            }
                        }

                        if (pArray != null)
                            for (int __i = 0; __i < pArray.Length; ++__i)
                            {
                                __pArray_gen_native[__i] = new global::SharedTypes.BoolStructNative(pArray[__i]);
                            }

                        //
                        // Invoke
                        //
                        __retVal_gen_native = AndAllMembers__PInvoke__(__pArray_gen_native, length);
                        //
                        // Unmarshal
                        //
                        __retVal = __retVal_gen_native != 0;
                    }
                    finally
                    {
                        //
                        // Cleanup
                        //
                        if (pArray__allocated)
                        {
                            System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__pArray_gen_native);
                        }
                    }

                    return __retVal;
                }
            }

            [System.Runtime.InteropServices.DllImportAttribute("NativeExportsNE", EntryPoint = "and_all_members")]
            extern private static unsafe byte AndAllMembers__PInvoke__(global::SharedTypes.BoolStructNative*pArray, int length);
        }
    }
}
```
</details>

<summary>By Value, In, Ref, Out, Return Value sample</summary>
<details>
User provided:

```csharp

partial class Test
{
    [GeneratedDllImport("DoesNotExist")]
    [return:MarshalAs(UnmanagedType.LPArray, SizeConst=10)]
    public static partial IntStructWrapper[] Method(
        IntStructWrapper[] p,
        in IntStructWrapper[] pIn,
        int pRefSize,
        [MarshalAs(UnmanagedType.LPArray, SizeParamIndex=2)] ref IntStructWrapper[] pRef,
        [MarshalAs(UnmanagedType.LPArray, SizeParamIndex=5, SizeConst=4)] out IntStructWrapper[] pOut,
        out int pOutSize
        );
}
[NativeMarshalling(typeof(IntStructWrapperNative))]
public struct IntStructWrapper
{
    public int Value;
}

public struct IntStructWrapperNative
{
    private int value;

    public IntStructWrapperNative(IntStructWrapper managed)
    {
        value = managed.Value;
    }

    public IntStructWrapper ToManaged() => new IntStructWrapper { Value = value };
}

```

Generated:

```csharp
partial class Test
{
    public static partial global::IntStructWrapper[] Method(global::IntStructWrapper[] p, in global::IntStructWrapper[] pIn, int pRefSize, ref global::IntStructWrapper[] pRef, out global::IntStructWrapper[] pOut, out int pOutSize)
    {
        unsafe
        {
            global::IntStructWrapperNative*__p_gen_native = default;
            global::IntStructWrapperNative*__pIn_gen_native = default;
            global::IntStructWrapperNative*__pRef_gen_native = default;
            pOut = default;
            global::IntStructWrapperNative*__pOut_gen_native = default;
            pOutSize = default;
            global::IntStructWrapper[] __retVal = default;
            global::IntStructWrapperNative*__retVal_gen_native = default;
            //
            // Setup
            //
            bool p__allocated = false;
            bool pIn__allocated = false;
            try
            {
                //
                // Marshal
                //
                if (p == null)
                {
                    __p_gen_native = null;
                }
                else
                {
                    int p__bytelen = sizeof(global::IntStructWrapperNative) * p.Length;
                    if (p__bytelen > 512)
                    {
                        __p_gen_native = (global::IntStructWrapperNative*)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(p__bytelen);
                        p__allocated = true;
                    }
                    else
                    {
                        byte *p__stackptr = stackalloc byte[p__bytelen];
                        ;
                        __p_gen_native = (global::IntStructWrapperNative*)p__stackptr;
                    }
                }

                if (p != null)
                {
                    new System.Span<global::IntStructWrapperNative>((global::IntStructWrapperNative*)__p_gen_native, p.Length).Clear();
                    for (int __i = 0; __i < p.Length; ++__i)
                    {
                        __p_gen_native[__i] = new global::IntStructWrapperNative(p[__i]);
                    }
                }

                if (pIn == null)
                {
                    __pIn_gen_native = null;
                }
                else
                {
                    int pIn__bytelen = sizeof(global::IntStructWrapperNative) * pIn.Length;
                    if (pIn__bytelen > 512)
                    {
                        __pIn_gen_native = (global::IntStructWrapperNative*)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(pIn__bytelen);
                        pIn__allocated = true;
                    }
                    else
                    {
                        byte *pIn__stackptr = stackalloc byte[pIn__bytelen];
                        ;
                        __pIn_gen_native = (global::IntStructWrapperNative*)pIn__stackptr;
                    }
                }

                if (pIn != null)
                {
                    new System.Span<global::IntStructWrapperNative>((global::IntStructWrapperNative*)__pIn_gen_native, pIn.Length).Clear();
                    for (int __i = 0; __i < pIn.Length; ++__i)
                    {
                        __pIn_gen_native[__i] = new global::IntStructWrapperNative(pIn[__i]);
                    }
                }

                __pRef_gen_native = null;
                if (pRef != null)
                {
                    int pRef__bytelen = sizeof(global::IntStructWrapperNative) * pRef.Length;
                    __pRef_gen_native = (global::IntStructWrapperNative*)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(pRef__bytelen);
                }

                if (pRef != null)
                {
                    new System.Span<global::IntStructWrapperNative>((global::IntStructWrapperNative*)__pRef_gen_native, pRef.Length).Clear();
                    for (int __i = 0; __i < pRef.Length; ++__i)
                    {
                        __pRef_gen_native[__i] = new global::IntStructWrapperNative(pRef[__i]);
                    }
                }

                //
                // Invoke
                //
                fixed (int *__pOutSize_gen_native = &pOutSize)
                {
                    __retVal_gen_native = Method__PInvoke__(__p_gen_native, &__pIn_gen_native, pRefSize, &__pRef_gen_native, &__pOut_gen_native, __pOutSize_gen_native);
                }

                //
                // Unmarshal
                //
                if (__retVal_gen_native != null)
                {
                    __retVal = new global::IntStructWrapper[10];
                    for (int __i = 0; __i < __retVal.Length; ++__i)
                    {
                        __retVal[__i] = __retVal_gen_native[__i].ToManaged();
                    }
                }
                else
                    __retVal = null;
                if (__pRef_gen_native != null)
                {
                    pRef = new global::IntStructWrapper[checked((int)pRefSize)];
                    for (int __i = 0; __i < pRef.Length; ++__i)
                    {
                        pRef[__i] = __pRef_gen_native[__i].ToManaged();
                    }
                }
                else
                    pRef = null;
                if (__pOut_gen_native != null)
                {
                    pOut = new global::IntStructWrapper[checked(4 + (int)pOutSize)];
                    for (int __i = 0; __i < pOut.Length; ++__i)
                    {
                        pOut[__i] = __pOut_gen_native[__i].ToManaged();
                    }
                }
                else
                    pOut = null;
            }
            finally
            {
                //
                // Cleanup
                //
                if (p__allocated)
                {
                    System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__p_gen_native);
                }

                if (pIn__allocated)
                {
                    System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__pIn_gen_native);
                }

                System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__pRef_gen_native);
                System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__pOut_gen_native);
            }

            return __retVal;
        }
    }

    extern private static unsafe global::IntStructWrapperNative*Method__PInvoke__(global::IntStructWrapperNative*p, global::IntStructWrapperNative**pIn, int pRefSize, global::IntStructWrapperNative**pRef, global::IntStructWrapperNative**pOut, int *pOutSize);
}
```
</details>

### DIFF
--- a/DllImportGenerator/designs/Compatibility.md
+++ b/DllImportGenerator/designs/Compatibility.md
@@ -54,6 +54,8 @@ Specifying array-specific marshalling members on the `MarshalAsAttribute` such a
 
 Only single-dimensional arrays are supported for source generated marshalling.
 
+Jagged arrays (arrays of arrays) are technically unsupported as was the case in the built-in marshalling system, but currently are not explicitly blocked by the source generator since they are not blocked at an architectural level, which was the case in the built-in system.
+
 In the source-generated marshalling, arrays will be allocated on the stack (instead of through [`AllocCoTaskMem`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.marshal.alloccotaskmem)) if they are passed by value or by read-only reference if they contain at most 256 bytes of data. The built-in system does not support this optimization for arrays.
 
 ### `LCIDConversion` support


DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
```diff
@@ -159,11 +159,13 @@ public static IEnumerable<object[]> CodeSnippetsToCompile_NoDiagnostics()
             yield return new[] { CodeSnippets.CustomStructMarshallingPinnableParametersAndModifiers };
             yield return new[] { CodeSnippets.CustomStructMarshallingNativeTypePinnable };
             yield return new[] { CodeSnippets.CustomStructMarshallingMarshalUsingParametersAndModifiers };
+            yield return new[] { CodeSnippets.ArrayMarshallingWithCustomStructElement };
         }
 
         public static IEnumerable<object[]> CodeSnippetsToCompile_WithDiagnostics()
         {
             yield return new[] { CodeSnippets.AllSupportedDllImportNamedArguments };
+            yield return new[] { CodeSnippets.ArrayMarshallingWithCustomStructElementWithValueProperty };
```
AaronRobinsonMSFT, line 168: This should go into the non-supported tests. The above snippets are for those we planned on enabling.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -182,15 +182,15 @@ internal class MarshallingGenerators
                     }
                     return SafeHandle;
 
-                case { ManagedType: IArrayTypeSymbol { IsSZArray: true, ElementType: ITypeSymbol elementType }, MarshallingAttributeInfo: NoMarshallingInfo }:
-                    return CreateArrayMarshaller(info, context, elementType, NoMarshallingInfo.Instance);
+                case { ManagedType: IArrayTypeSymbol { IsSZArray: true, ElementType: ITypeSymbol elementType }, MarshallingAttributeInfo: ArrayMarshallingInfo(MarshallingInfo elementMarshallingInfo) }:
```
AaronRobinsonMSFT, line 185: Let's collapse this case and make a local function. I really dislike this compound statements in switch cases. They are too dense and I fear we don't fully appreciate all the cases we are or aren't supporting.

AaronRobinsonMSFT: I think removing the `MarshallingAttributeInfo` and handling that in a sub function is sufficient.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -379,6 +379,14 @@ private static IMarshallingGenerator CreateArrayMarshaller(TypePositionInfo info
 
         private static IMarshallingGenerator CreateCustomNativeTypeMarshaller(TypePositionInfo info, StubCodeContext context, NativeMarshallingAttributeInfo marshalInfo)
         {
+            if (marshalInfo.ValuePropertyType is not null && !context.CanUseAdditionalTemporaryState)
```
AaronRobinsonMSFT, line 382: This is a great branch and the message is clear. I do have a request on `CanUseAdditionalTemporaryState`. I see in the `StubContext` it is marked `abstract` but it isn't obvious as to why it would be true or false. Can you add a bit of commentary in `StubContext` on why one would or wouldn't support that. Usage here is clear but I don't fully understand all the cases when we would chose to set this to be true or false.

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
```diff
@@ -208,15 +208,29 @@ static MarshalAsInfo CreateMarshalAsInfo(AttributeData attrData, DefaultMarshall
                     }
                 }
 
-                return isArrayType
-                    ? new ArrayMarshalAsInfo(
+                if (isArrayType)
```
AaronRobinsonMSFT, line 211: Seems like we are at the end of the method here (i.e. last logical conclusions). Let's flip this predicate and return here in the non-array case and remove the nesting for the more complicated follow-up to handle arrays.

DllImportGenerator/TestAssets/NativeExports/Arrays.cs
```diff
@@ -120,5 +121,18 @@ public static void Append(int** values, int numOriginalValues, int newValue)
             newArray[numOriginalValues] = newValue;
             *values = newArray;
         }
+
+        [UnmanagedCallersOnly(EntryPoint = "and_all_members")]
+        [DNNE.C99DeclCode("struct bool_struct;")]
```
AaronRobinsonMSFT, line 126: Up to you, but support exists for defining a header file and then supplying include paths - see [`DnneAdditionalIncludeDirectories`](https://github.com/AaronRobinsonMSFT/DNNE/blob/master/src/msbuild/DNNE.props).

DllImportGenerator/TestAssets/NativeExports/Arrays.cs
```diff
@@ -120,5 +121,18 @@ public static void Append(int** values, int numOriginalValues, int newValue)
...
+        public static byte AndAllMembers([DNNE.C99Type("struct bool_struct*")] BoolStructNative* pArray, int length)
```
AaronRobinsonMSFT, line 127: Absolutely thrilled with the fact we can test/demonstrate both sides of our custom marshalling in a single test!

DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
jkoritzinsky: I was thinking that we do plan on enabling this, just not in the immediate term. This is for only the near-term enabling in v1 scenario, then I'll move this over to the non-supported tests.

AaronRobinsonMSFT: Yep. I think the calculus has changed how much we are going to be able to enable in v1. I would move it for now.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
jkoritzinsky: Personally I prefer putting all of the cases for metadata that we support in the case statement, otherwise we start depending on ordering of case statements. For example, there was a bug where we didn't support putting a MarshalUsing attribute on a `string` parameter since the case statement for "is this a string" didn't check metadata and the "is this any type marked with MarshalUsing" was after the string check. But I can move the metadata check into the `CreateArrayMarshaller` function if you'd prefer.

AaronRobinsonMSFT: Okay. Let's continue with this pattern then.

>But I can move the metadata check into the CreateArrayMarshaller function if you'd prefer.

Please and thank you.

jkotas:
Could you please share code-snippets that this generates?

jkoritzinsky:
I've added the example from the integration tests into the PR description. The only part that's "custom marshalling"-specific is the body of the for loop. The rest is the same array code as in the main branch.

jkotas:
What is it going to do if the marshaller has `FreeNative` method?

jkoritzinsky:
If the marshaller has a FreeNative method, the code-gen will have another `for` loop that calls `FreeNative` in the finally block right before the `FreeCoTaskMem` if block. The loop would look like the following:

```csharp
                        if (pArray != null)
                            for (int __i = 0; __i < pArray.Length; ++__i)
                            {
                                __pArray_gen_native[__i].FreeNative(pArray[__i]);
                            }
```

jkotas:
Does this need to account for error handling? What if the initialization of the array failed half-way through?

jkoritzinsky:
If the initialization failed halfway through, the native structs should be default initialized. The implementor of the native struct should handle this scenario.

I may have to add handling for that case.

AaronRobinsonMSFT:
>Does this need to account for error handling? What if the initialization of the array failed half-way through?

I thought we were going to assume non-exception cases only at least custom marshalling. In this case that would seem to imply that if the custom marshalling throws, we can either let them handle the possible clean _or_ determine which of the elements were initialized and only free them. @jkotas are you implying the latter?

jkoritzinsky:
I've added in some basic handling for this case similar to the old system where we zero-init the memory before array marshalling in case we only partially marshal. I also fixed a bug around marshalling arrays of pointers since I found it along the way.

jkotas:
> we zero-init the memory before array marshalling in case we only partially marshal

This works, but it may be more efficient to keep track of how many elements of the array were marshalled successfully and cleanup only those.

jkotas:
Also, what does would the code look for `[in,out] ref BoolStruct[] pArray`?

jkotas:
> we can either let them handle the possible clean or determine which of the elements were initialized and only free them

Agree with that.

jkoritzinsky:
I've added a more comprehensive sample to the PR description that covers the `ref` case.

I prefer to use the mem-clear method initially since that's what the old system used and it still leaves open the option for switching methods in the future.

jkotas:
> `for (int __i = 0; __i < pRef.Length; ++__i)`

This is prone to buffer overruns. `pRef` can change between the time the buffer is allocate and the marshalling happens.

jkotas:
> sizeof(global::IntStructWrapperNative) * pRef.Length

This is prone to overflow. It should `be checked(...)`

jkoritzinsky:
> > `for (int __i = 0; __i < pRef.Length; ++__i)`
> 
> This is prone to buffer overruns. `pRef` can change between the time the buffer is allocate and the marshalling happens.
> 
> The length should be cached to avoid this problem.

Which loop is this referring to? There's multiple loops with this condition (one in marshal, one in unmarshal).

Since both of these comments are relating to general array marshalling, not marshalling of arrays of custom marshaled types, I'm going to put out a separate PR for these fixes.

jkotas:
The problem is in the loops that operate on arrays passed in via `ref`. The user code can be changing the value on other thread.

jkotas:
```
                        if (pArray == null)
                        {
                            __pArray_gen_native = null;
                        }
                        else
                        {
```

This can be just `if (pArray != null)`. `__pArray_gen_native` is initialized to null from the prolog.

jkoritzinsky:
@jkotas I've opened #384 with the array marshalling improvements.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -205,11 +195,15 @@ internal class MarshallingGenerators
                 case { MarshallingAttributeInfo: GeneratedNativeMarshallingAttributeInfo(string nativeTypeName) }:
                     return Forwarder;
 
+                // Cases that just match on type must come after the checks for the new marshalling model.
```
AaronRobinsonMSFT, line 198: >new marshalling model

Not sure I follow this statement.

jkoritzinsky: I've clarified the comment.

jkoritzinsky:
Up to date stub example with BoolStruct (taken from the integration tests):

```csharp
namespace DllImportGenerator.IntegrationTests
{
    partial class NativeExportsNE
    {
        public partial class Arrays
        {
            [System.Runtime.CompilerServices.SkipLocalsInitAttribute]
            public static partial bool AndAllMembers(global::SharedTypes.BoolStruct[] pArray, int length)
            {
                unsafe
                {
                    global::SharedTypes.BoolStructNative*__pArray_gen_native = default;
                    bool __retVal = default;
                    byte __retVal_gen_native = default;
                    //
                    // Setup
                    //
                    bool pArray__allocated = false;
                    try
                    {
                        //
                        // Marshal
                        //
                        if (pArray != null)
                        {
                            int pArray__bytelen = checked(sizeof(global::SharedTypes.BoolStructNative) * pArray.Length);
                            if (pArray__bytelen > 512)
                            {
                                __pArray_gen_native = (global::SharedTypes.BoolStructNative*)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(pArray__bytelen);
                                pArray__allocated = true;
                            }
                            else
                            {
                                byte *pArray__stackptr = stackalloc byte[pArray__bytelen];
                                ;
                                __pArray_gen_native = (global::SharedTypes.BoolStructNative*)pArray__stackptr;
                            }
                        }

                        if (pArray != null)
                        {
                            new System.Span<global::SharedTypes.BoolStructNative>((global::SharedTypes.BoolStructNative*)__pArray_gen_native, pArray.Length).Clear();
                            for (int __i = 0; __i < pArray.Length; ++__i)
                            {
                                __pArray_gen_native[__i] = new global::SharedTypes.BoolStructNative(pArray[__i]);
                            }
                        }

                        //
                        // Invoke
                        //
                        __retVal_gen_native = AndAllMembers__PInvoke__(__pArray_gen_native, length);
                        //
                        // Unmarshal
                        //
                        __retVal = __retVal_gen_native != 0;
                    }
                    finally
                    {
                        //
                        // Cleanup
                        //
                        if (pArray__allocated)
                        {
                            System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__pArray_gen_native);
                        }
                    }

                    return __retVal;
                }
            }

            [System.Runtime.InteropServices.DllImportAttribute("Microsoft.Interop.Tests.NativeExportsNE", EntryPoint = "and_all_members")]
            extern private static unsafe byte AndAllMembers__PInvoke__(global::SharedTypes.BoolStructNative*pArray, int length);
        }
    }
}
```

jkoritzinsky:
Here's a more comprehensive one taken from the unit tests:

```csharp
partial class Test
{
    [System.Runtime.CompilerServices.SkipLocalsInitAttribute]
    public static partial global::IntStructWrapper[] Method(global::IntStructWrapper[] p, in global::IntStructWrapper[] pIn, int pRefSize, ref global::IntStructWrapper[] pRef, out global::IntStructWrapper[] pOut, out int pOutSize)
    {
        unsafe
        {
            global::IntStructWrapperNative*__p_gen_native = default;
            global::IntStructWrapperNative*__pIn_gen_native = default;
            global::IntStructWrapperNative*__pRef_gen_native = default;
            pOut = default;
            global::IntStructWrapperNative*__pOut_gen_native = default;
            pOutSize = default;
            global::IntStructWrapper[] __retVal = default;
            global::IntStructWrapperNative*__retVal_gen_native = default;
            //
            // Setup
            //
            bool p__allocated = false;
            bool pIn__allocated = false;
            global::IntStructWrapper[] pIn_local = pIn;
            global::IntStructWrapper[] pRef_local = pRef;
            global::IntStructWrapper[] pOut_local = pOut;
            try
            {
                //
                // Marshal
                //
                if (p != null)
                {
                    int p__bytelen = checked(sizeof(global::IntStructWrapperNative) * p.Length);
                    if (p__bytelen > 512)
                    {
                        __p_gen_native = (global::IntStructWrapperNative*)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(p__bytelen);
                        p__allocated = true;
                    }
                    else
                    {
                        byte *p__stackptr = stackalloc byte[p__bytelen];
                        ;
                        __p_gen_native = (global::IntStructWrapperNative*)p__stackptr;
                    }
                }

                if (p != null)
                {
                    new System.Span<global::IntStructWrapperNative>((global::IntStructWrapperNative*)__p_gen_native, p.Length).Clear();
                    for (int __i = 0; __i < p.Length; ++__i)
                    {
                        __p_gen_native[__i] = new global::IntStructWrapperNative(p[__i]);
                    }
                }

                if (pIn_local != null)
                {
                    int pIn__bytelen = checked(sizeof(global::IntStructWrapperNative) * pIn_local.Length);
                    if (pIn__bytelen > 512)
                    {
                        __pIn_gen_native = (global::IntStructWrapperNative*)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(pIn__bytelen);
                        pIn__allocated = true;
                    }
                    else
                    {
                        byte *pIn__stackptr = stackalloc byte[pIn__bytelen];
                        ;
                        __pIn_gen_native = (global::IntStructWrapperNative*)pIn__stackptr;
                    }
                }

                if (pIn_local != null)
                {
                    new System.Span<global::IntStructWrapperNative>((global::IntStructWrapperNative*)__pIn_gen_native, pIn.Length).Clear();
                    for (int __i = 0; __i < pIn.Length; ++__i)
                    {
                        __pIn_gen_native[__i] = new global::IntStructWrapperNative(pIn_local[__i]);
                    }
                }

                __pRef_gen_native = null;
                if (pRef_local != null)
                {
                    int pRef__bytelen = checked(sizeof(global::IntStructWrapperNative) * pRef_local.Length);
                    __pRef_gen_native = (global::IntStructWrapperNative*)System.Runtime.InteropServices.Marshal.AllocCoTaskMem(pRef__bytelen);
                }

                if (pRef_local != null)
                {
                    new System.Span<global::IntStructWrapperNative>((global::IntStructWrapperNative*)__pRef_gen_native, pRef.Length).Clear();
                    for (int __i = 0; __i < pRef.Length; ++__i)
                    {
                        __pRef_gen_native[__i] = new global::IntStructWrapperNative(pRef_local[__i]);
                    }
                }

                //
                // Invoke
                //
                fixed (int *__pOutSize_gen_native = &pOutSize)
                    __retVal_gen_native = Method__PInvoke__(__p_gen_native, &__pIn_gen_native, pRefSize, &__pRef_gen_native, &__pOut_gen_native, __pOutSize_gen_native);
                //
                // Unmarshal
                //
                if (__retVal_gen_native != null)
                {
                    __retVal = new global::IntStructWrapper[10];
                    for (int __i = 0; __i < __retVal.Length; ++__i)
                    {
                        __retVal[__i] = __retVal_gen_native[__i].ToManaged();
                    }
                }
                else
                    __retVal = null;
                if (__pRef_gen_native != null)
                {
                    pRef_local = new global::IntStructWrapper[checked((int)pRefSize)];
                    for (int __i = 0; __i < pRef_local.Length; ++__i)
                    {
                        pRef_local[__i] = __pRef_gen_native[__i].ToManaged();
                    }
                }
                else
                    pRef_local = null;
                pRef = pRef_local;
                if (__pOut_gen_native != null)
                {
                    pOut_local = new global::IntStructWrapper[checked(4 + (int)pOutSize)];
                    for (int __i = 0; __i < pOut_local.Length; ++__i)
                    {
                        pOut_local[__i] = __pOut_gen_native[__i].ToManaged();
                    }
                }
                else
                    pOut_local = null;
                pOut = pOut_local;
            }
            finally
            {
                //
                // Cleanup
                //
                if (p__allocated)
                {
                    System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__p_gen_native);
                }

                if (pIn__allocated)
                {
                    System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__pIn_gen_native);
                }

                System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__pRef_gen_native);
                System.Runtime.InteropServices.Marshal.FreeCoTaskMem((System.IntPtr)__pOut_gen_native);
            }

            return __retVal;
        }
    }

    extern private static unsafe global::IntStructWrapperNative*Method__PInvoke__(global::IntStructWrapperNative*p, global::IntStructWrapperNative**pIn, int pRefSize, global::IntStructWrapperNative**pRef, global::IntStructWrapperNative**pOut, int *pOutSize);
}

```

AaronRobinsonMSFT (approved):
Let's add some notes about arrays to the compat doc - multi-dimensional not supported and support for nested array types?

jkoritzinsky:
The compat doc already mentions multidimensional arrays (it says only single dimensional arrays are supported). I've added a little blurb about current support for arrays of arrays.

PR #380, opened 2020-11-23
Add handling for [In, Out] attributes.

jkoritzinsky:
Implement support for [Out] and [In, Out] on by-value arrays. Disallow usage of [In] or [Out] attributes on by reference parameters or on by value parameters where they have no effect.

Integration tests depend on #379

### DIFF
--- a/DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
+++ b/DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
@@ -180,5 +180,68 @@ namespace DllImportGenerator.UnitTests
             var newCompDiags = newComp.GetDiagnostics();
             Assert.Empty(newCompDiags);
         }
+        
+        public static IEnumerable<object[]> CodeSnippetsToCompileWithForwarder()
+        {
+            yield return new[] { CodeSnippets.UserDefinedEntryPoint };
+            yield return new[] { CodeSnippets.AllSupportedDllImportNamedArguments };
+            
+            // Parameter / return types (supported in DllImportGenerator)
+            yield return new[] { CodeSnippets.BasicParametersAndModifiers<byte>() };
+            // Parameter / return types (not supported in DllImportGenerator)
+            yield return new[] { CodeSnippets.BasicParametersAndModifiers<string>() };
+        }
+
+        [Theory]
+        [MemberData(nameof(CodeSnippetsToCompileWithForwarder))]
+        public async Task ValidateSnippetsWithForwarder(string source)
+        {
+            Compilation comp = await TestUtils.CreateCompilation(source);
+            TestUtils.AssertPreSourceGeneratorCompilation(comp);
+
+            var newComp = TestUtils.RunGenerators(
+                comp,
+                new DllImportGeneratorOptionsProvider(useMarshalType: false, generateForwarders: true),
+                out var generatorDiags,
+                new Microsoft.Interop.DllImportGenerator());
+
+            Assert.Empty(generatorDiags);
+
+            var newCompDiags = newComp.GetDiagnostics();
+            Assert.Empty(newCompDiags);
+        }
+
+        public static IEnumerable<object[]> CodeSnippetsToCompileWithMarshalType()
+        {
+            // SetLastError
+            yield return new[] { CodeSnippets.AllSupportedDllImportNamedArguments };
+            
+            // SafeHandle
+            yield return new[] { CodeSnippets.BasicParametersAndModifiers("Microsoft.Win32.SafeHandles.SafeFileHandle") };
+        }
+
+        [Theory]
+        [MemberData(nameof(CodeSnippetsToCompileWithMarshalType))]
+        public async Task ValidateSnippetsWithMarshalType(string source)
+        {
+            Compilation comp = await TestUtils.CreateCompilation(source);
+            TestUtils.AssertPreSourceGeneratorCompilation(comp);
+
+            var newComp = TestUtils.RunGenerators(
+                comp,
+                new DllImportGeneratorOptionsProvider(useMarshalType: true, generateForwarders: false),
+                out var generatorDiags,
+                new Microsoft.Interop.DllImportGenerator());
+
+            Assert.Empty(generatorDiags);
+
+            var newCompDiags = newComp.GetDiagnostics();
+
+            Assert.All(newCompDiags, diag =>
+            {
+                Assert.Equal("CS0117", diag.Id);
+                Assert.StartsWith("'Marshal' does not contain a definition for ", diag.GetMessage());
+            });
+        }
     }
 }
--- a/DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
+++ b/DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
@@ -4,6 +4,7 @@ using System.Runtime.InteropServices;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
 using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
 
 namespace Microsoft.Interop
@@ -118,7 +119,6 @@ namespace Microsoft.Interop
         public static readonly Forwarder Forwarder = new Forwarder();
         public static readonly BlittableMarshaller Blittable = new BlittableMarshaller();
         public static readonly DelegateMarshaller Delegate = new DelegateMarshaller();
-        public static readonly SafeHandleMarshaller SafeHandle = new SafeHandleMarshaller();
         public static readonly HResultExceptionMarshaller HResultException = new HResultExceptionMarshaller();
 
         /// <summary>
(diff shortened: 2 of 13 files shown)


DllImportGenerator/DllImportGenerator.UnitTests/CompileFails.cs
```diff
@@ -39,6 +39,23 @@ public static IEnumerable<object[]> CodeSnippetsToCompile()
...
+            yield return new object[] { CodeSnippets.ByValueParameterWithModifier("in int", "In"), 1, 0 };
+            yield return new object[] { CodeSnippets.ByValueParameterWithModifier("ref int", "In"), 1, 0 };
+            yield return new object[] { CodeSnippets.ByValueParameterWithModifier("ref int", "In, Out"), 1, 0 };
+            yield return new object[] { CodeSnippets.ByValueParameterWithModifier("out int", "Out"), 1, 0 };
+
+            // By value non-array with [In, Out] attributes
+            yield return new object[] { CodeSnippets.ByValueParameterWithModifier<byte>("In"), 1, 0 };
+            yield return new object[] { CodeSnippets.ByValueParameterWithModifier<byte>("Out"), 1, 0 };
+            yield return new object[] { CodeSnippets.ByValueParameterWithModifier<byte>("In, Out"), 1, 0 };
+
```
AaronRobinsonMSFT, line 57: Extra line.

DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
```diff
@@ -91,6 +91,11 @@ public static IEnumerable<object[]> CodeSnippetsToCompile_NoDiagnostics()
             yield return new[] { CodeSnippets.ArrayParameterWithNestedMarshalInfo<string>(UnmanagedType.LPUTF8Str) };
             yield return new[] { CodeSnippets.ArrayParameterWithNestedMarshalInfo<string>(UnmanagedType.LPStr) };
 
+            // [In, Out] attributes
+            // By value non-blittable array
+            yield return new[] { CodeSnippets.ByValueParameterWithModifier<bool[]>("Out") };
+            yield return new[] { CodeSnippets.ByValueParameterWithModifier<bool[]>("In, Out") };
```
AaronRobinsonMSFT, line 97: Seems like we are missing non-blittable with just `In`. I think you mentioned this previously as an issue with detection, is that true?

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -60,6 +60,8 @@ internal interface IMarshallingGenerator
         /// <see cref="StubCodeContext.CurrentStage" /> of <paramref name="context"/> may not be valid.
         /// </remarks>
         bool UsesNativeIdentifier(TypePositionInfo info, StubCodeContext context);
+
+        bool SupportsByValueMarshalKind(ByValueContentsMarshalKind marshalKind, StubCodeContext context);
```
AaronRobinsonMSFT, line 64: Missing doc.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -113,7 +115,7 @@ internal class MarshallingGenerators
         public static readonly HResultExceptionMarshaller HResultException = new HResultExceptionMarshaller();
 
         /// <summary>
-        /// Create an <see cref="IMarshallingGenerator"/> instance to marshalling the supplied type.
+        /// Create an <see cref="IMarshallingGenerator"/> instance to marshalling the supplied type in the given position.
```
AaronRobinsonMSFT, line 118: ~~to marshalling~~ -> "for marshalling" or "to marshal"

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -122,6 +124,47 @@ internal class MarshallingGenerators
             TypePositionInfo info,
             StubCodeContext context)
         {
+            IMarshallingGenerator generator = CreateCore(info, context);
+            ValidateByValueMarshalKind(context, info, generator);
```
AaronRobinsonMSFT, line 128: Thoughts on making this a gate? I am suggesting that we convert this to a pipeline style function call so it becomes `return ValidateByValueMarshalKind(context, info, generator);`. Which means `ValidateByValueMarshalKind` would return the `IMarshallingGenerator` instance.

AaronRobinsonMSFT: May I assume it is just easier to to do all these checks here as opposed to pushing these checks deeper into the create pipeline?

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
```diff
@@ -278,6 +310,28 @@ static bool TryCreateTypeBasedMarshallingInfo(ITypeSymbol type, Compilation comp
             }
         }
 
+        private static ByValueContentsMarshalKind GetByValueContentsMarshalKind(IEnumerable<AttributeData> attributes, Compilation compilation)
+        {
+            INamedTypeSymbol outAttributeType = compilation.GetTypeByMetadataName(TypeNames.System_Runtime_InteropServices_OutAttribute)!;
```
AaronRobinsonMSFT, line 315: Can we return early if either of these are `null`? I'm not arguing for it, just making sure I understand these could be null. If not, are they never `null` since they are defined in SPCL or the like?

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
```diff
@@ -16,6 +16,35 @@ namespace Microsoft.Interop
...
     );
 
+    /// <summary>
+    /// Describes how to marshal the contents of a value in comparison to the value itself.
+    /// Only makes sense for array-like types. For example, an "out" array doesn't change the
+    /// pointer to the array value, but it marshals the contents of the native array back to the
+    /// contents of the managed array.
+    /// </summary>
+    [Flags]
+    enum ByValueContentsMarshalKind
```
AaronRobinsonMSFT, line 26: Mark `internal`.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
jkoritzinsky: I can change this into a gate. I'm fine with that and it will be a little cleaner.

It's much easier doing the checks here than in every single marshaller, especially since we share marshaller instances for the simple marshallers.

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
jkoritzinsky: These types are defined in CoreLib, so they'll never be null.

DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
jkoritzinsky: Non-blittable with just `In` is unsupported. We only support `[Out]` and `[In, Out]` since `[In]` on its own doesn't do anything. Looks like I forgot to add a test for this in CompileFails though, so I'll be sure to add that. This is not an issue with detection; that issue I was seeing was due to a bug.

jkoritzinsky:
This is finally ready for another review pass.

DllImportGenerator/DllImportGenerator/Marshalling/NonBlittableArrayMarshaller.cs
```diff
@@ -135,13 +145,46 @@ public override IEnumerable<StatementSyntax> Generate(TypePositionInfo info, Stu
...
+                                    List(_elementMarshaller.Generate(
+                                        info with { ManagedType = GetElementTypeSymbol(info), RefKind = elementRefKind },
+                                        arraySubContext))))));
                     }
                     break;
                 case StubCodeContext.Stage.Unmarshal:
-                    if (info.IsManagedReturnPosition || (info.IsByRef && info.RefKind != RefKind.In))
+                    if (info.IsManagedReturnPosition
+                        || (info.IsByRef && info.RefKind != RefKind.In)
+                        || (info.ByValueContentsMarshalKind & ByValueContentsMarshalKind.Out) != 0)
```
AaronRobinsonMSFT, line 156: Can't we use the `HasFlag()` function for all of these checks with `ByValueContentsMarshalKind`?

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
```diff
@@ -16,6 +16,35 @@ namespace Microsoft.Interop
         CharEncoding CharEncoding
     );
 
+    /// <summary>
+    /// Describes how to marshal the contents of a value in comparison to the value itself.
+    /// Only makes sense for array-like types. For example, an "out" array doesn't change the
```
AaronRobinsonMSFT, line 21: >Only makes sense for array-like types.

I'm assuming we have a nice assert or even a Diagnostic warning for the user that one is using `In`/`Out` on a type that we think is silly?

AaronRobinsonMSFT (approved):

DllImportGenerator/DllImportGenerator/TypePositionInfo.cs
jkoritzinsky: Yes we emit a diagnostic warning or error if someone applies `In` or `Out` in a case where the attribute has no effect.

PR #383, opened 2020-11-25
Surface the trimming warnings

MichalStrehovsky:
This fully lights up the dynamic code dataflow analysis warnings. The warnings are disabled by default and controlled by the same MSBuild property as IL linker uses.

I took the MessageContainer and MessageOrigin concepts from the linker. I really don't have opinion on them but I guess if they work there, we can use them here too (this is done in two commits - first commit is linker sources verbatim, the second commit are tweaks).

This also affects crossgen2, so crossgen2 can now also emit warnings and have suppressions. Currently it doesn't use any of that, but we compile the code.

We are using the linker's error code namespace: https://github.com/mono/linker/blob/master/docs/error-codes.md

### DIFF
--- a/src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -34,6 +34,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <IlcGenerateCompleteTypeMetadata Condition="$(IlcGenerateCompleteTypeMetadata) == ''">true</IlcGenerateCompleteTypeMetadata>
     <IlcGenerateStackTraceData Condition="$(IlcGenerateStackTraceData) == ''">true</IlcGenerateStackTraceData>
     <IlcScanReflection Condition="$(IlcScanReflection) == ''">true</IlcScanReflection>
+    <SuppressTrimAnalysisWarnings Condition="$(SuppressTrimAnalysisWarnings) == ''">true</SuppressTrimAnalysisWarnings>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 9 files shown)


src/coreclr/src/tools/aot/ILCompiler/Program.cs
```diff
@@ -864,6 +867,21 @@ private static bool DumpReproArguments(CodeGenerationFailedException ex)
             return false;
         }
 
+        private static IEnumerable<int> ProcessWarningCodes(IEnumerable<string> warningCodes)
+        {
+            foreach (string value in warningCodes)
+            {
+                string[] values = value.Split(new char[] { ',', ';', ' ' }, StringSplitOptions.RemoveEmptyEntries);
```
jkotas, line 874:
```suggestion
                string[] values = value.Split(",; ", StringSplitOptions.RemoveEmptyEntries);
```

src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -34,6 +34,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <IlcGenerateCompleteTypeMetadata Condition="$(IlcGenerateCompleteTypeMetadata) == ''">true</IlcGenerateCompleteTypeMetadata>
     <IlcGenerateStackTraceData Condition="$(IlcGenerateStackTraceData) == ''">true</IlcGenerateStackTraceData>
     <IlcScanReflection Condition="$(IlcScanReflection) == ''">true</IlcScanReflection>
+    <SuppressTrimAnalysisWarnings Condition=" '$(SuppressTrimAnalysisWarnings)' == '' ">true</SuppressTrimAnalysisWarnings>
```
jkotas, line 37:
```suggestion
    <SuppressTrimAnalysisWarnings Condition="$(SuppressTrimAnalysisWarnings) == ''">true</SuppressTrimAnalysisWarnings>
```
to have the style consistent with the conditions above?

src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -96,6 +97,30 @@ The .NET Foundation licenses this file to you under the MIT license.
     <AutoInitializedAssemblies Include="System.Private.DeveloperExperience.Console" Condition="$(IlcDisableUnhandledExceptionExperience) != 'true'"  />
   </ItemGroup>
 
+  <!-- Keep the list of default warning suppressions in sync with Microsoft.NET.ILLink.targets -->
+  <PropertyGroup Condition=" '$(SuppressTrimAnalysisWarnings)' == 'true' ">
```
jkotas, line 101:
```suggestion
  <PropertyGroup Condition="$(SuppressTrimAnalysisWarnings) == 'true'">
```

jkotas (approved):
Thanks!

MichalStrehovsky:
Heh, every single piece of feedback was to code imported from linker/sdk repo :). I should do less mindless copying.

src/coreclr/src/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -34,6 +34,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <IlcGenerateCompleteTypeMetadata Condition="$(IlcGenerateCompleteTypeMetadata) == ''">true</IlcGenerateCompleteTypeMetadata>
     <IlcGenerateStackTraceData Condition="$(IlcGenerateStackTraceData) == ''">true</IlcGenerateStackTraceData>
     <IlcScanReflection Condition="$(IlcScanReflection) == ''">true</IlcScanReflection>
+    <SuppressTrimAnalysisWarnings Condition="$(SuppressTrimAnalysisWarnings) == ''">true</SuppressTrimAnalysisWarnings>
```
jkotas, line 37: @MichalStrehovsky This does not seem to be working. I am seeing a pile of warnings when building samples/HelloWorld using the current package.

MichalStrehovsky: Sigh, I should have retested after applying the code review feedback. #397

jkotas:
> This also affects crossgen2, so crossgen2 can now also emit warnings and have suppressions. Currently it doesn't use any of that, but we compile the code.

Do you think it would be a good idea to push this change to dotnet/runtime? If yes, could you please take care of it?

MichalStrehovsky:
> Do you think it would be a good idea to push this change to dotnet/runtime? If yes, could you please take care of it?

I can do that once we iron out the bugs like #406 in it.

It should be just about copying back the updated Logger.cs, and the newly added MessageContainer.cs and MessageOrigin.cs files (plus the updated crossgen2 csproj). I verified crossgen2 is building fine in this repo. There isn't anything else to it since I'm not switching logging to the new plan yet.

PR #386, opened 2020-11-25
Fix comments in assembler files

AntonLapounov:
#239 renamed `UnwindWriteBarrierToCaller` to `UnwindSimpleHelperToCaller` without updating assembler files.  Also fix comments regarding the return address on ARM/ARM64.


jkotas (approved):
Thank you!

PR #390, opened 2020-11-26
CI: Run debug configuration instead of checked

MichalStrehovsky:
The tests match the build configuration of the runtime. We don't have explicit `_r` and `_d` csproj for the smoke tests - this means that we always run with optimizations enabled.

We need coverage for the smoke tests with optimizations off - so we either add extra CSPROJ files a la JIT tests that have _r and _d, or just run a debug build in the CI.

### DIFF
--- a/eng/pipelines/runtimelab.yml
+++ b/eng/pipelines/runtimelab.yml
@@ -76,12 +76,12 @@ stages:
           extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
 
   #
-  # Build with Release config and Release runtimeConfiguration
+  # Build with Release config and Checked runtimeConfiguration
   #
   - template: /eng/pipelines/common/platform-matrix.yml
     parameters:
       jobTemplate: /eng/pipelines/common/global-build-job.yml
-      buildConfig: Release
+      buildConfig: Checked
       platforms:
       - Linux_x64
       - windows_x64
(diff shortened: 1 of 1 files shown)


jkotas (approved):

eng/pipelines/runtimelab.yml
```diff
@@ -58,21 +58,21 @@ stages:
...
         - windows_x64
         - OSX_x64
         jobParameters:
           timeoutInMinutes: 90
           testGroup: innerloop
-          buildArgs: -s nativeaot+libs+nativeaot.packages -lc release -rc checked
+          buildArgs: -s nativeaot+libs+nativeaot.packages -lc release -rc debug
           extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
 
   #
```
jkotas, line 78: Would it make sense this other run to be Release config with Checked runtime configuration?

jkotas: I think it is useful to run RyuJIT with both optimizations and asserts enables. It caches most of the JIT bugs.

MichalStrehovsky: Yep, that makes sense. There's an assertion failure - I'm fixing that in another pull request.

We don't run with assertions in CHK builds because dotnet/runtime#43017 was still not merged :/

PR #395, opened 2020-11-26
Remove duplicate include for AppContext.Browser.cs

yowl:
This PR to NativeAOT as this is already removed in dotnet/runtime, but the latest merge did not remove it here.

image


jkotas (approved):
Thank you!

PR #397, opened 2020-11-27
Fix processing of warning suppressions

MichalStrehovsky:


jkotas (approved):
Thanks! I lived under impression that the two overloads are equivalent.

Issue #106, opened 2020-09-16
Unification tech debt


MichalStrehovsky:
Tech debt that we need to pay to RI into dotnet/runtime:

- [ ] Build a System.Linq.Expressions that works on both full AOT and JIT-based runtimes (debt introduced in #102)
- [ ] Nullable annotations in the CoreLib (nullability warnings in the libraries partition disabled in src/libraries/Directory.Build.props in #70)
- [x] Re-enable IL linker warnings disabled in https://github.com/dotnet/runtimelab/commit/00210dc0ef42561d267879dbeaa17b3710ce5489
- [ ] #140 undid dotnet/runtime#41966. We should find a way to do what we need while not regressing size of CoreCLR-based packages (unless we want to use the same package).
- [ ] System.IO.Compression.Native build options from #461
- [ ] RequiresDynamicCodeAttribute needs to go through API review and we need to annotate ref assemblies (validation for this is currently [disabled](https://github.com/dotnet/runtimelab/blob/58f5ffbf8130889ffb41f27be8892ae706488b14/eng/ApiCompatExcludeAttributes.txt#L3))

kant2002:
I think "Nullable annotations in the CoreLib" is handled somehow.

jkotas:
The nullability warnings are disabled in https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/coreclr/nativeaot/Directory.Build.props#L33-L37 now.

kant2002:
thanks.

MichalStrehovsky:
Integration to dotnet/runtime is done.

PR #402, opened 2020-11-29
Enforce IL linker warnings while building CoreLib

jkotas:
Contributes to #106


src/coreclr/src/nativeaot/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreRT.cs
```diff
@@ -284,6 +284,10 @@ public static void PrepareDelegate(Delegate d)
                 throw new ArgumentNullException(nameof(d));
         }
 
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2059:UnrecognizedReflectionPattern",
```
jkotas, line 287: I have removed these by accident when resolving merge conflict

MichalStrehovsky (approved):
Thanks! Forgot we were suppressing these!

Issue #396, opened 2020-11-27
Drive ILC Trim Analysis warnings for "Hello world" to zero


jkotas:
We have several dozens ILC Trim Analysis warnings even for simplest `Console.WriteLine("Hello world")` type of programs.

Some of them are inherited from dotnet/runtime. We should mirror the approach used to address those in dotnet/runtime.
Some of them are unique to NativeAOT and required annotations in NativeAOT specific code.

MichalStrehovsky:
This is done, Hello World has 0 warnings.

PR #406, opened 2020-11-30
Fix UnconditionalSuppressMessage parsing

MichalStrehovsky:
We weren't processing the suppressions due to a off-by-one bug.'

Contributes to #396.


jkotas (approved):
Thanks!

PR #407, opened 2020-11-30
Add test coverage for GetUninitializedObject

MichalStrehovsky:
Just to make sure we don't regress this if we ever try to optimize things more. The trimming warning suppressions assume this.


jkotas (approved):
Thank you!

PR #408, opened 2020-11-30
Annotate a couple reflection APIs

MichalStrehovsky:
There isn't a good way to annotate these - we can only suppress and hope the public entrypoints to these are properly annotated.

Contributes to #396.


jkotas (approved):
LGTM

PR #412, opened 2020-12-01
Undo workarounds

jkotas:


MichalStrehovsky (approved):
Thanks for cleaning it up!

dotnet.cmd
```diff
@@ -19,9 +19,4 @@ set DOTNET_MULTILEVEL_LOOKUP=0
 :: Disable first run since we want to control all package sources
 set DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
 
-:: Temporarily make this dotnet more permanent
-:: We need this for Native AOT CI testing until ilc becomes a selfcontained app.
-set DOTNET_ROLL_FORWARD=Major
-set DOTNET_ROOT=%dotnetPath%
```
yowl, line 25: Didn't this used to set the msbuild `$(DotnetRoot)` for https://github.com/dotnet/runtimelab/blob/f8cac7528ef2e113d1b80f44f9677c972c186e84/src/tests/Common/tests.targets#L56

jkotas: I do not think there is any path how setting of `DOTNET_ROOT` influences setting of `DotnetRoot`.

Is there a problem that you are seeing with this change?

Somebody on Discord reported a problem with `DotnetRoot` not getting set correctly. I am not sure how it gets triggered.

yowl: thanks.  sorry, just jumping to the wrong conclusions, my problem (as on gitter) is not related to this change.

PR #418, opened 2020-12-03
Merge from NativeAOT

yowl:
This is a merge from the NativeAOT branch.  The commits below should confirm that there are no other changes.


yowl:
Thanks

PR #420, opened 2020-12-03
Extend CoreRT README with a workaround for older tfms

adamsitnik:
@jkotas this is what helped me when working on https://github.com/dotnet/BenchmarkDotNet/pull/1606 but I expect that older tfms might not be supported and this PR might get rejected.


jkotas (approved):
Thanks

PR #422, opened 2020-12-04
Add support for feature switches

MichalStrehovsky:
This is still a work in progress, I just want to get an opinion on it.

* Parses the ILLinkTrim annotations
* Rewrites method bodies to remove unreachable branches.

The removal of unreachable branches is done by nopping out the unreachable basic block. We put a `br $-2` at the beginning of the nop block to make it clear to the importer that it doesn't need to bother.

The logic mirrors what linker does, in a way. We don't try to model the evaluation stack and instead just assume there will be known instructions in front of the branch. It has problems. For example if the known instruction is a call that takes parameters.

Linker will also try to remove the conditional branch itself, but I avoided that so that we don't need to recompute branch targets, EH blocks, or debug info.

In general, the linker implementation makes it very hard to reason about what is going to be constpropped. I wish we went with the original design that simply eliminated entire bodies, like ConditionalAttribute does...

This also only does a subset of what linker can constprop. Linker tries to constprop a bunch of stuff, a lot of it is [quite questionable](https://github.com/mono/linker/blob/b24d0e663d5bf4772b0896e2bc3dab3391fae744/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs#L718-L747), like treating signed and unsigned comparisons the same.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ProcessLinkerXmlBase.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ProcessLinkerXmlBase.cs
@@ -315,9 +315,6 @@ namespace ILCompiler
 
             public override void AppendName(StringBuilder sb, FunctionPointerType type)
             {
-                // WTF is this CECIL?!
-                //signature.Append(function.Name);
-
                 sb.Append(" ");
                 AppendName(sb, type.Signature.ReturnType);
                 sb.Append(" *");
--- a/src/coreclr/tools/aot/ILCompiler/Program.cs
+++ b/src/coreclr/tools/aot/ILCompiler/Program.cs
@@ -186,7 +186,7 @@ namespace ILCompiler
                 syntax.DefineOption("methodbodyfolding", ref _methodBodyFolding, "Fold identical method bodies");
                 syntax.DefineOptionList("initassembly", ref _initAssemblies, "Assembly(ies) with a library initializer");
                 syntax.DefineOptionList("appcontextswitch", ref _appContextSwitches, "System.AppContext switches to set");
-                syntax.DefineOptionList("feature", ref _featureSwitches, "Feature switches to apply (format: 'Feature.Name=[true|false]'");
+                syntax.DefineOptionList("feature", ref _featureSwitches, "Feature switches to apply (format: 'Namespace.Name=[true|false]'");
                 syntax.DefineOptionList("runtimeopt", ref _runtimeOptions, "Runtime options to set");
                 syntax.DefineOptionList("removefeature", ref _removedFeatures, "Framework features to remove");
                 syntax.DefineOption("singlethreaded", ref _singleThreaded, "Run compilation on a single thread");


jkotas:
> It has problems. For example if the known instruction is a call that takes parameters

I am not following. Could you please elaborate?

MichalStrehovsky:
ldc.i4.0
lcd.i4.0
call bool Foo(bool)
Beq

Assume Foo has a substitution.

When trying to figure out the beq, we will bail because Foo takes parameters and we don't know how to skip over those without modelling the stack.

The removal would work if the parameters to beq are fed in the opposite order.

jkotas:
The substitutions for methods with arguments are used in BCL already. I think it means that we will need to bite the bullet and do the stack modeling. Do you see any other viable options?

MichalStrehovsky:
Il linker is not modeling the stack either so it probably ends up on the "good" side of the comparison.

I don't want to spend more energy on this than was spent in the linker because this whole thing is a big back.

jkotas:
I agree. I was under impression that the linker is modelling the stack... .

src/coreclr/src/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
```diff
@@ -0,0 +1,905 @@
...
+
+        public MethodIL GetMethodILWithInlinedSubstitutions(MethodIL method)
+        {
+            //Debug.Assert(method.GetMethodILDefinition() == method);
+
+            ILExceptionRegion[] ehRegions = method.GetExceptionRegions();
+            byte[] methodBytes = method.GetILBytes();
+            OpcodeFlags[] flags = new OpcodeFlags[methodBytes.Length];
+
+            // TODO comment how it works
```
jkotas, line 80: ... and where it does not work.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Compilation.cs
```diff
@@ -37,6 +37,8 @@ public abstract class Compilation : ICompilation
         private readonly AssemblyGetExecutingAssemblyMethodThunkCache _assemblyGetExecutingAssemblyMethodThunks;
         private readonly MethodBaseGetCurrentMethodThunkCache _methodBaseGetCurrentMethodThunks;
 
+        private readonly FeatureSwitchManager mgr;
```
MichalStrehovsky, line 40: This file is still not done yet - I just want it to run the mark&sweep on all IL in all tests to weed out bugs.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ProcessLinkerXmlBase.cs
```diff
@@ -0,0 +1,380 @@
...
+
+            sb.Append(')');
+            return sb.ToString();
+        }
+
+        private bool ShouldProcessElement()
+        {
+            return true;
+
+            /*string feature = _reader.GetAttribute("feature");
```
MichalStrehovsky, line 257: Note to self: uncomment - I just want to enable all feature switches so that it does things.

MichalStrehovsky:
I want to rebase the one commit I factored out into #444 but otherwise this is ready to review.

What's missing is the functionality to only do IL rewriting if we detect it's needed. That might end up being a bigger change if we don't want to do it with exceptions so it's probably better to do it separately.

src/coreclr/tools/aot/ILCompiler/Program.cs
```diff
@@ -184,6 +186,7 @@ private ArgumentSyntax ParseCommandLine(string[] args)
                 syntax.DefineOption("methodbodyfolding", ref _methodBodyFolding, "Fold identical method bodies");
                 syntax.DefineOptionList("initassembly", ref _initAssemblies, "Assembly(ies) with a library initializer");
                 syntax.DefineOptionList("appcontextswitch", ref _appContextSwitches, "System.AppContext switches to set");
+                syntax.DefineOptionList("feature", ref _featureSwitches, "Feature switches to apply (format: 'Feature.Name=[true|false]'");
```
jkotas, line 189:
```suggestion
                syntax.DefineOptionList("feature", ref _featureSwitches, "Feature switches to apply (format: 'Namespace.Name=[true|false]'");
```

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ProcessLinkerXmlBase.cs
```diff
@@ -0,0 +1,384 @@
...
+            {
+                AppendName(sb, type.ParameterType);
+                sb.Append('*');
+            }
+
+            public override void AppendName(StringBuilder sb, FunctionPointerType type)
+            {
+                // WTF is this CECIL?!
+                //signature.Append(function.Name);
+
```
jkotas, line 320:
```suggestion
```

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
```diff
@@ -0,0 +1,837 @@
...
+                    if (erase)
+                        newBody[position] = (byte)ILOpCode.Nop;
+                    position++;
+                    if (!erase)
+                        endOfUsefulCode = position;
+                }
+                while (position < newBody.Length && flags[position] == 0);
+            }
+
+            // RyuJIT doesn't like when there's unreachable garbage at the end of the method.
```
jkotas, line 416: Where is the code that chokes on this? I would expect that RyuJIT should handle fine.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
```diff
@@ -0,0 +1,837 @@
...
+                return substitution.EmitIL(method);
+            }
+
+            // BEGIN TEMPORARY WORKAROUND
+            //
+            // The following lines should just be:
+            // return _nestedILProvider.GetMethodIL(method);
+            // But we want to allow this to be used as a general-purpose IL provider.
+            //
+            // Rewriting all IL has compilation throughput hit we don't want.
```
jkotas, line 80: How much throughput hit does this change have? (Trying to see whether we need to bother with rewriting only once we detect there is something to rewrite.)

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
MichalStrehovsky: IIRC it was this one:

https://github.com/dotnet/runtime/blob/f513349a68a2f0e05189ca2fe2972b1da306995e/src/coreclr/jit/flowgraph.cpp#L6021-L6024

RyuJIT seems to generate basic blocks also for junk bytes not referenced from any live code and this code chokes on it. It was a block of nops following a `br $-2`.

MichalStrehovsky: I'll double check and capture an ngendump

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
MichalStrehovsky: I'll measure. It's possible it's pretty cheap. We run this in the multithreaded phase.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
MichalStrehovsky: Yeah, it's that one:

```
****** START compiling System.SR:Format(System.String,System.Object):System.String (MethodHash=b9e74bbf)
Generating code for Windows x64
OPTIONS: compCodeOpt = BLENDED_CODE
OPTIONS: compDbgCode = true
OPTIONS: compDbgInfo = true
OPTIONS: compDbgEnC  = false
OPTIONS: compProcedureSplitting   = false
OPTIONS: compProcedureSplittingEH = false
OPTIONS: Jit invoked for ngen
IL to import:
IL_0000  28 01 01 00 06    call         0x6000101
IL_0005  2c 19             brfalse.s    25 (IL_0020)
IL_0007  72 ad 01 00 70    ldstr        0x700001AD
IL_000c  18                ldc.i4.2
IL_000d  8d 06 00 00 01    newarr       0x1000006
IL_0012  25                dup
IL_0013  16                ldc.i4.0
IL_0014  02                ldarg.0
IL_0015  a2                stelem.ref
IL_0016  25                dup
IL_0017  17                ldc.i4.1
IL_0018  03                ldarg.1
IL_0019  a2                stelem.ref
IL_001a  28 6d 00 00 0a    call         0xA00006D
IL_001f  2a                ret
IL_0020  2b fe             br.s         -2 (IL_0020)
IL_0022  00                nop
IL_0023  00                nop
IL_0024  00                nop
IL_0025  00                nop
IL_0026  00                nop
IL_0027  00                nop

lvaSetClass: setting class for V00 to (4000000000420430) System.String
Arg #0    passed in register(s) rcx

lvaSetClass: setting class for V01 to (4000000000420260) System.Object
Arg #1    passed in register(s) rdx

lvaGrabTemp returning 2 (V02 tmp0) (a long lifetime temp) called for OutgoingArgSpace.
; Initial local variable assignments
;
;  V00 arg0              ref  class-hnd
;  V01 arg1              ref  class-hnd
;  V02 OutArgs        lclBlk <na>  "OutgoingArgSpace"
*************** In compInitDebuggingInfo() for System.SR:Format(System.String,System.Object):System.String
getVars() returned cVars = 0, extendOthers = true
info.compVarScopesCount = 2
        VarNum  LVNum         Name      Beg     End
 0:     00h     00h       V00 arg0      000h    028h
 1:     01h     01h       V01 arg1      000h    028h
New Basic Block BB01 [0000] created.
New scratch BB01
Debuggable code - Add new BB01 [0000] to perform initialization of variables
info.compStmtOffsetsCount    = 0
info.compStmtOffsetsImplicit = 0007h ( STACK_EMPTY NOP CALL_SITE )
*************** In fgFindBasicBlocks() for System.SR:Format(System.String,System.Object):System.String
weight= 79 : state  40 [ call ]
weight= 27 : state  44 [ brfalse.s ]
weight= 66 : state 102 [ ldstr ]
weight= 34 : state  25 [ ldc.i4.2 ]
weight=152 : state 118 [ newarr ]
weight= 11 : state  38 [ dup ]
weight= 15 : state  23 [ ldc.i4.0 ]
weight= 10 : state   3 [ ldarg.0 ]
weight= 94 : state 139 [ stelem.ref ]
weight= 11 : state  38 [ dup ]
weight= 28 : state  24 [ ldc.i4.1 ]
weight= 16 : state   4 [ ldarg.1 ]
weight= 94 : state 139 [ stelem.ref ]
weight= 79 : state  40 [ call ]
weight= 19 : state  42 [ ret ]
weight= 44 : state  43 [ br.s ]
weight= 65 : state   2 [ noshow ]
weight= 65 : state   2 [ noshow ]
weight= 65 : state   2 [ noshow ]
weight= 65 : state   2 [ noshow ]
weight= 65 : state   2 [ noshow ]
weight= 65 : state   2 [ noshow ]
Jump targets:
  IL_0020
New Basic Block BB02 [0001] created.
BB02 [000..007)
New Basic Block BB03 [0002] created.
BB03 [007..020)
New Basic Block BB04 [0003] created.
BB04 [020..022)
```

MichalStrehovsky:
#456 is going to fix the multifile test failure.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
MichalStrehovsky: I'm getting pretty noisy results (+- 2%) and this is basically within the noise)

PR #423, opened 2020-12-05
Update prefix of packages and published assemblies

AaronRobinsonMSFT:
Update all packages and published assemblies to have an official `Microsoft` prefix. Placed under `Microsoft.Interop.` to be more specific.

/cc @jkoritzinsky @elinor-fung @safern @mmitche


jkoritzinsky (approved):

safern (commented):
Thanks for updating so quickly.

PR #437, opened 2020-12-07
Merge from dotnet/runtime

jkotas:


jkotas:
@MichalStrehovsky Could you please review the last commit?

MichalStrehovsky:
I'm a bit surprised we need to call `FindMethodOnTypeWithMatchingTypicalMethod` in the first place, but this change looks good (wonder what the result of resolution is that we need that).

We could probably put the new if check under `#if !READYTORUN` for clarity. Crossgen2 doesn't model the generic interfaces on arrays at all.

PR #444, opened 2020-12-11
Do not require ILProvider in UsageBasedMetadataManager

MichalStrehovsky:
Extracting this commit out of #422 for better reviewability.

WASM will need adjustment similar to what is done in the scanner and the RyuJIT-based codegen (we call into CodeBasedDependencyAlgorithm.AddDependenciesDueToMethodCodePresence at a different spot where we still have the `MethodIL`) - cc @yowl.

Instead of requiring UsageBasedMetadataManager to have its own ILProvider, we provide a `MethodIL` when calling into the manager.

The advantages are twofold:

* We can play tricks and substitute the `MethodIL` when some other components finds out it's necessary to bypass the one from `ILProvider`.
* We now call into the logic while still compiling the method - which means the IL scanning that `MetadataManager` sometimes has to do happens during the multithreaded phase. This should help throughput.

The `ILProvider` used by `FlowAnnotations` doesn't participate in this scheme but it's unlikely it would ever matter - it's only used to scan bodies of autogenerated property getters/setters to find the underlying field. I don't expect we would ever need to mess with that IL.

### DIFF
--- a/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
+++ b/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
@@ -303,6 +303,8 @@ namespace Internal.JitInterface
                 _methodCodeNode.Fixups.Add(node);
             }
 #else
+            MethodIL methodIL = (MethodIL)HandleToObject((IntPtr)_methodScope);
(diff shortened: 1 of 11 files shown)


jkotas (approved):
LGTM. Thanks!

PR #411, opened 2020-11-30
NativeAOT-LLVM: Enable CI for Windows/Debug

yowl:
This PR enables CI for the LLVM backend, initially by running just the Generics test for Windows build of Wasm in Debug configuration.

PR #448, opened 2020-12-11
NativeAot-LLVM: Fix EHInfo RelocType 

yowl:
The RelocType used to swith on the Target OS, which used to be Windows in CoreRT, but is now WebAssembly, so just fix the type to absolute.

Draft as waiting for CI PR, #411


jkotas (approved):
LGTM

PR #451, opened 2020-12-12
Normalize unspecified CharSet mangling to CharSet.Ansi

### LINKED ISSUE #450: [ NativeAot ] Sizes are much smaller than there were originally
After upgrading all sizes in an app don't match non-coreRt version.

Here is a screen for empty Avalonia app :

изображение_2020-12-12_094730

App from the publish folder is on the right side

Version of packages:

`Avalonia`: 0.10.0-rc1
`microsoft.dotnet.ilcompiler` 6.0.0-alpha.1.20611.2

jkotas:
Fixes #450

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/NodeFactory.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/NodeFactory.cs
@@ -2,18 +2,17 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.Collections.Generic;
 using System.Collections.Concurrent;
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/NodeFactory.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Concurrent;
+using System.Runtime.InteropServices;
```
MichalStrehovsky, line 7: Do we need this?

MichalStrehovsky (approved):
Thank you!

PR #456, opened 2020-12-14
Fix Checked multifile builds

MichalStrehovsky:
The leaf (test) assemblies are correctly compiled with optimizations turned on in checked builds, but the framework wasn't. This resulted in fun-to-debug differences in generic dictionary shapes between the code compiled as part of the framework and dictionary lookups generated by app code. Optimizations can remove the need for some slots in certain cases.


jkotas (approved):
I can imagine this was fun to debug

MichalStrehovsky:
> I can imagine this was fun to debug

We need an option to dump the dictionary (and maybe even vtable) layouts the compiler came up with. It would probably be easier to troubleshoot. I'll put it in my when-I-have-time wishlist.

PR #458, opened 2020-12-15
Implement support for configurable generator options.

jkoritzinsky:
Implement support for a "forwarder" mode and implement an option to emit references to the Marshal type when the MarshalEx type is requested to support perf testing and integration with dotnet/runtime respectively.

Uses https://github.com/dotnet/roslyn/blob/master/docs/features/source-generators.cookbook.md#consume-msbuild-properties-and-metadata as the basis for the feature.

The `DllImportGenerator_UseMarshalType` MSBuild property changes usages of the `MarshalEx` type to be the `Marshal` type.

The `DllImportGenerator_GenerateForwarders` MSBuild property causes the generator to emit functions that forward to built-in P/Invoke stubs instead of handling the marshalling in the generated code.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/Microsoft.Interop.DllImportGenerator.props
+++ b/DllImportGenerator/DllImportGenerator/Microsoft.Interop.DllImportGenerator.props
@@ -1,6 +1,19 @@
 ﻿<Project>
+  <!--
+    Define all of the configuration options supported for the DllImportGenerator.
+    To use, set an MSBuild property with the name of the option to `true`.
+    See OptionsHelper.cs for more information on usage.
+  -->
   <ItemGroup>
+      <!--
+        Use the System.Runtime.InteropServices.Marshal type instead of
+        the System.Runtime.InteropServices.MarshalEx type when emitting code.
+      -->
       <CompilerVisibleProperty Include="DllImportGenerator_UseMarshalType" />
+      <!--
+        Generate a stub that forwards to a runtime implemented P/Invoke stub instead
+        of generating a stub that handles all of the marshalling.
+      -->
       <CompilerVisibleProperty Include="DllImportGenerator_GenerateForwarders" />
   </ItemGroup>
 </Project>


DllImportGenerator/DllImportGenerator/Microsoft.Interop.DllImportGenerator.props
```diff
@@ -0,0 +1,6 @@
+﻿<Project>
+  <ItemGroup>
+      <CompilerVisibleProperty Include="DllImportGenerator_UseMarshalType" />
```
AaronRobinsonMSFT, line 3: Let's document what these mean here.

AaronRobinsonMSFT: And reference where one can find more information about usage or adding more.

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
```diff
@@ -174,7 +177,7 @@ public override (string managed, string native) GetIdentifiers(TypePositionInfo
                 AppendVariableDeclations(setupStatements, retMarshaller.TypeInfo, retMarshaller.Generator);
             }
 
-            if (this.dllImportData.SetLastError)
+            if (this.dllImportData.SetLastError && !options.GenerateForwarders())
```
AaronRobinsonMSFT, line 180: Comment on this logic. Why does this matter? Also, I like to see `this.` so it is clean this isn't an argument.

AaronRobinsonMSFT: Same below.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -303,7 +305,7 @@ private static IMarshallingGenerator CreateStringMarshaller(TypePositionInfo inf
             throw new MarshallingNotSupportedException(info, context);
         }
         
-        private static ExpressionSyntax GetNumElementsExpressionFromMarshallingInfo(TypePositionInfo info, StubCodeContext context)
+        private static ExpressionSyntax GetNumElementsExpressionFromMarshallingInfo(TypePositionInfo info, StubCodeContext context, AnalyzerConfigOptions options)
```
AaronRobinsonMSFT, line 308: Help me defend not exposing the `AnalyzerConfigOptions` on `StubCodeContext`. I am mostly there but want to hear another argument.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -287,7 +288,7 @@ private static bool IsGeneratedDllImportAttribute(AttributeSyntax attrSyntaxMayb
 
                 Debug.Assert(expSyntaxMaybe is not null);
 
-                if (PassThroughToDllImportAttribute(namedArg.Key))
+                if (generateForwarders || PassThroughToDllImportAttribute(namedArg.Key))
```
AaronRobinsonMSFT, line 291: Spell out the fact that "pass through" and "forwarder" are the same concept here.

DllImportGenerator/DllImportGenerator/DllImportStub.cs
```diff
@@ -203,7 +205,7 @@ public class GeneratedDllImportData
             }
 
             // Generate stub code
-            var stubGenerator = new StubCodeGenerator(method, dllImportData, paramsTypeInfo, retTypeInfo, diagnostics);
+            var stubGenerator = new StubCodeGenerator(method, dllImportData, paramsTypeInfo, retTypeInfo, diagnostics, env.Options);
```
AaronRobinsonMSFT, line 208: Do we want to expose the `StubEnvironment` on the `StubCodeContext`? This is another option rather than adding the analyzer options to the context.

DllImportGenerator/DllImportGenerator.UnitTests/TestUtils.cs
```diff
@@ -89,9 +94,23 @@ public static Compilation RunGenerators(Compilation comp, out ImmutableArray<Dia
...
+        /// <param name="diagnostics">Resulting diagnostics</param>
+        /// <param name="generators">Source generator instances</param>
+        /// <returns>The resulting compilation</returns>
+        public static Compilation RunGenerators(Compilation comp, out ImmutableArray<Diagnostic> diagnostics, ISourceGenerator[] generators, AnalyzerConfigOptionsProvider? options = null)
+        {
+            CreateDriver(comp, generators, options).RunGeneratorsAndUpdateCompilation(comp, out var d, out diagnostics);
+            return d;
+        }
+
+        private static GeneratorDriver CreateDriver(Compilation c, ISourceGenerator[] generators, AnalyzerConfigOptionsProvider? options = null)
```
AaronRobinsonMSFT, line 110: Remove default `= null`.  I dislike these typically but very much dislike them when they are `private`.

DllImportGenerator/DllImportGenerator.UnitTests/DllImportGeneratorOptionsProvider.cs
```diff
@@ -0,0 +1,66 @@
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace DllImportGenerator.UnitTests
+{
+    internal class DllImportGeneratorOptionsProvider : AnalyzerConfigOptionsProvider
```
AaronRobinsonMSFT, line 7: Please document these classes - just the class. What are they for?

DllImportGenerator/DllImportGenerator.UnitTests/DllImportGeneratorOptionsProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            return EmptyOptions.Instance;
+        }
+
+        public override AnalyzerConfigOptions GetOptions(AdditionalText textFile)
+        {
+            return EmptyOptions.Instance;
+        }
+
+        class GlobalGeneratorOptions : AnalyzerConfigOptions
```
AaronRobinsonMSFT, line 26: mark classes `private`.

DllImportGenerator/DllImportGenerator.UnitTests/DllImportGeneratorOptionsProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            private readonly bool _useMarshalType = false;
```
AaronRobinsonMSFT, line 28: I think we are keeping with the `this.` syntax rather than the `_`.

DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
```diff
@@ -175,5 +175,75 @@ public async Task ValidateSnippets(string source)
...
+                    new Microsoft.Interop.DllImportGenerator()
+                },
+                new DllImportGeneratorOptionsProvider(useMarshalType: false, generateForwarders: true));
+
+            Assert.Empty(generatorDiags);
+
+            var newCompDiags = newComp.GetDiagnostics();
+            Assert.Empty(newCompDiags);
+        }
+        public static IEnumerable<object[]> CodeSnippetsToCompileWithMarshalType()
```
AaronRobinsonMSFT, line 211: New line above.

DllImportGenerator/DllImportGenerator.UnitTests/Compiles.cs
```diff
@@ -175,5 +175,75 @@ public async Task ValidateSnippets(string source)
...
+
+            var newCompDiags = newComp.GetDiagnostics();
+
+            Assert.All(newCompDiags, diag =>
+            {
+                Assert.Equal("CS0117", diag.Id);
+                Assert.StartsWith("'Marshal' does not contain a definition for ", diag.GetMessage());
+            });
+        }
+
```
AaronRobinsonMSFT, line 246: 2 extra lines.

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
jkoritzinsky: We don't want to generate our own SetLastError logic if we're having the runtime handle the SetLastError logic.

I'll add a comment.

DllImportGenerator/DllImportGenerator.UnitTests/DllImportGeneratorOptionsProvider.cs
jkoritzinsky: Our code base is very much a mix depending on who wrote each part. Parts that I wrote have the `_` prefix and parts that you and Elinor wrote have the `this.` prefix. Personally I prefer not using either prefix and I really don't like the `this.` prefix, it feels particularly clunky to me. I adjusted to the `_` prefix because people wanted a prefix and dotnet/runtime uses the `_` prefix.

After the holiday break we should set up an .editorconfig file so we can enforce the style rules we decide on as part of the build.

AaronRobinsonMSFT: yep that makes sense. @elinor-fung and I are used to the old StyleCop rules and follow them because it makes code reviews easier for us. I agree we should have the conversation.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
jkoritzinsky: Here's an argument for not putting it on the context:

The options are part of the configuration of the generators. Technically, the context is not only part of the configuration of the generators, but the execution as well. Since the options don't change over time but the context object might, we should initialize the marshallers with the options at selection time instead of at generation time.

Additionally, in the future the `UseMarshalType` option will become a no-op when we ship, so we won't need to pass down the options to the specific generators. At which point, it wouldn't make sense for the context to carry the options with it.

AaronRobinsonMSFT: Done.

DllImportGenerator/DllImportGenerator.UnitTests/DllImportGeneratorOptionsProvider.cs
jkoritzinsky: I'm used to the default C# style options in VS, so that's what I default to. Once we have the editorconfig, Roslyn will help out with enforcing our style, so I'm willing to be convinced to change styles. We can also install the Rosyln StyleCop package to enforce the rules if we go that route.

DllImportGenerator/DllImportGenerator/OptionsHelper.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Interop
+{
+    public static class OptionsHelper
+    {
+        public const string UseMarshalTypeOption = "build_property.DllImportGenerator_UseMarshalType";
```
AaronRobinsonMSFT, line 12: If this can be `private` please mark them as such.

AaronRobinsonMSFT (approved):

DllImportGenerator/DllImportGenerator/OptionsHelper.cs
jkoritzinsky: These are public so the tests can access them.

PR #459, opened 2020-12-15
Populate list of Windows APIs that are safe to direct call

jkotas:


src/coreclr/nativeaot/BuildIntegration/WindowsAPIs.txt
```diff
@@ -1,21 +1,2336 @@
-# Windows APIs that must be direct calls for the runtime to be able to startup
-# TODO: Expand this to list all APIs available in supported Windows versions
+# List of ubiquitous Windows APIs that safe to emit direct calls for
```
jkotas, line 1:
```suggestion
# List of ubiquitous Windows APIs that are safe to emit direct calls for
```

MichalStrehovsky (approved):
Nice!

PR #473, opened 2020-12-16
Fix logging in the absence of symbols

MichalStrehovsky:
There was a commented out block that I forgot to enable.

Enabling it required taking the `DisplayName` functionality out of the `Dataflow` namespace.


jkotas (approved):
Thank you!

PR #480, opened 2020-12-18
Disable BinaryFormatter by default

MichalStrehovsky:
https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md

For now, it can be turned back on by opting in explicitly, but per the above doc we reserve the right to just hardcode it as always disabled ("This hardcoded behavior extends to all other novel application types that might be introduced after the .NET 5 timeframe, including mobile apps and other restricted environment configurations.").

Gets rid of almost half of trimming warnings on a hello world app. Saves around 300 kB on a hello world app.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -24,7 +24,6 @@ The .NET Foundation licenses this file to you under the MIT license.
     <TargetOS Condition="'$([MSBuild]::IsOSPlatform(OSX))' == 'true'">OSX</TargetOS>
     <TargetOS Condition="'$(TargetOS)' == ''">$(OS)</TargetOS>
     <NativeDebugSymbols Condition="$(DebugSymbols) == 'true' or ($(DebugType) != 'none' and $(DebugType) != '')">true</NativeDebugSymbols>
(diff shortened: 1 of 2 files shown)


jkotas:
Add a paragraph about this default to somewhere in the docs, e.g. to https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/docs/using-nativeaot/optimizing.md ?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -24,6 +24,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <TargetOS Condition="'$([MSBuild]::IsOSPlatform(OSX))' == 'true'">OSX</TargetOS>
     <TargetOS Condition="'$(TargetOS)' == ''">$(OS)</TargetOS>
     <NativeDebugSymbols Condition="$(DebugSymbols) == 'true' or ($(DebugType) != 'none' and $(DebugType) != '')">true</NativeDebugSymbols>
+    <EnableUnsafeBinaryFormatterSerialization Condition="'$(EnableUnsafeBinaryFormatterSerialization)' == ''">false</EnableUnsafeBinaryFormatterSerialization>
```
jkotas, line 27: Would it make more sense to have it in the next PropertyGroup block that has defaults for all sorts of feature switches?

MichalStrehovsky: I made a separate PropertyGroup for this. The next block has default settings that improve compatibility.

This default does the opposite.

PR #481, opened 2020-12-18
NativeAot-LLVM: Consistency in test host OS checking

yowl:
Following from https://github.com/dotnet/runtime/pull/46105, this PR updates the remaining test for checking the host OS to be in sync with other places.


yowl:
I'm going to push these into runtime, so maybe wait to see if there is feedback there, before oking this.

yowl:
Going to close this as dont think its needed anymore.  Certainly these changes are not in the current code.

PR #484, opened 2020-12-19
NativeAOT-LLVM split compilation of bitcode and final link

yowl:
This PR reinserts the step from CoreRT for separate compilation of LLVM bitcode to WebAssembly object file before final linking.  Fixes the bitcode extension (to `.bc`) and the switch for the debug compilation settings.  Changes some of the MSBuild property names/values from Wasm to LLVM as that feels more appropriate.


jkotas (approved):
LGTM. Thanks!

PR #488, opened 2020-12-20
Substitute IsDynamicCodeSupported/Compiled as always false

MichalStrehovsky:
This lets the compiler eliminate branches that depend on this value. This lets the compiler eliminate various Ref.Emit-based optimizations and leave us with "normal reflection" fallbacks.

This e.g. removes 5 dataflow-related warnings when System.Text.RegularExpressions is used since those warnings apply to reflection use connected with System.Reflection.Emit. Saw these in a customer app that I was analyzing.


jkotas (approved):
Thank you!

Issue #485, opened 2020-12-19
AccessViolation with PLINQ


queequac:
One of my dependencies uses PLINQ internally.
Depending on the actual data and query, I get the following error:

`Access Violation: Attempted to read or write protected memory. This is often an indication that other memory is corrupt. The application will be terminated since this platform does not support throwing an AccessViolationException.
`

That's the stacktrace:
```
Text.exe!System_Linq_Parallel_System_Linq_Parallel_MergeExecutor_1<System___Canon>__Execute<Int32>() Line 91	Unknown
Test.exe!System_Linq_Parallel_System_Linq_Parallel_PartitionedStreamMerger_1<System___Canon>__Receive<Int32>() Line 60	Unknown
Text.exe!System_Linq_Parallel_System_Linq_Parallel_UnaryQueryOperator_2_UnaryQueryOperatorResults_ChildResultsRecipient<System___Canon__System___Canon>__Receive<Int32>() Line 162	Unknown
Text.exe!System_Linq_Parallel_System_Linq_Parallel_ScanQueryOperator_1_ScanEnumerableQueryOperatorResults<System___Canon>__GivePartitionedStream() Line 136	Unknown
Test.exe!System_Linq_Parallel_System_Linq_Parallel_UnaryQueryOperator_2_UnaryQueryOperatorResults<System___Canon__System___Canon>__GivePartitionedStream() Line 132	Unknown
Test.exe!System_Linq_Parallel_System_Linq_Parallel_QueryOperator_1<System___Canon>__GetOpenedEnumerator() Line 165	Unknown
Test.exe!System_Linq_Parallel_System_Linq_Parallel_ForAllOperator_1<System___Canon>__RunSynchronously() Line 64	Unknown
```

Running without Native-AOT, everything is okay. Also noteworthy, the issue does not show up when compiling with .NET Native for UWP.

jkotas:
Any chance you can share the repro; or at least a crash dump with symbols?

queequac:
I will do some more investigations.
Unfortunately I cannot repro the issue easily, because it depends a lot on the actual data loaded into that library ... Data I cannot share.

Let's see if I can create something similar with dummy data.

queequac:
Here you go: [ReproTest](https://1drv.ms/u/s!AqT0PAFUIw04h_xIPtBoJWkIPKcrdw?e=l2e3M9)
[...]

PR #489, opened 2020-12-21
Don't allow generic inlining if context differs

MichalStrehovsky:
The `contextFromMethodBeingCompiled` optimization doesn't apply if we're changing the generic context from "method dictionary" to "the type".

Not sure if this can lead to other problems, but at minimum it causes a problem for the generic dictionary of the generic method computed by the scanner - we don't expect "foreign entries" in the method dictionary.

Contributes to #485.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
@@ -1159,6 +1159,8 @@ namespace Internal.JitInterface
                     pResolvedToken.tokenContext == contextFromMethodBeingCompiled() &&
                     constrainedType == null &&
                     exactType == MethodBeingCompiled.OwningType &&
+                    // But don't allow inlining into generic methods since the generic context won't be the same.
+                    // The scanner won't be able to predict such inlinig. See https://github.com/dotnet/runtimelab/pull/489
                     !MethodBeingCompiled.HasInstantiation)
                 {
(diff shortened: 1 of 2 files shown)


src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
```diff
@@ -1158,7 +1158,8 @@ private void getCallInfo(ref CORINFO_RESOLVED_TOKEN pResolvedToken, CORINFO_RESO
                 if (pResult->exactContextNeedsRuntimeLookup &&
                     pResolvedToken.tokenContext == contextFromMethodBeingCompiled() &&
                     constrainedType == null &&
-                    exactType == MethodBeingCompiled.OwningType)
+                    exactType == MethodBeingCompiled.OwningType &&
+                    !MethodBeingCompiled.HasInstantiation)
```
jkotas, line 1162: It may be worth a comment that this condition is required to make dictionaries pre-computed by the scanner work.

jkotas: (The condition is not present and needed in the R2R version of this method and it is not obvious why.)

jkotas (approved):

MichalStrehovsky:
> Should we also change the assert in #485 (comment) to throw fatal compiler error?

Yup, done. I don't know why I did it this way. I guess I copied it from the VTable slot assignment logic, but that one actually uses the -1 in some places. Might be nice to fix it there too at some point.

Issue #452, opened 2020-12-12
A lot of time spent in GVMDependenciesNode.SearchDynamicDependencies


jkotas:
Repro: 
1. Clone https://github.com/jkotas/AvaloniaCoreRTDemo/tree/update-to-0.10.0-rc1
2. dotnet publish -c Release -r win-x64

Result: The native compilation takes many time longer compared to previous Avalonia version (close to 10 minutes on my notebook)

The problem seems to be a lot of time spent under SearchDynamicDependencies:

Name | Inc %
-- | --
| ILCompiler.Compiler!GVMDependenciesNode.SearchDynamicDependencies | 74.0
| - ILCompiler.TypeSystem!MetadataVirtualMethodAlgorithm.ResolveInterfaceMethodToVirtualMethodOnType | 52.7
| - ILCompiler.TypeSystem!MetadataVirtualMethodAlgorithm.FindVirtualFunctionTargetMethodOnObjectType | 20.5

jkotas:
@FoggyFinder Have you noticed much longer compilation times with 10.0-rc1 in your project as well?

jkotas:
@MichalStrehovsky Do you believe that we need a new cache to fix this, or should the existing caches be enough?

FoggyFinder:
> @FoggyFinder Have you noticed much longer compilation times with 10.0-rc1 in your project as well?

Yes. But that was expected. I saw somewhere a message that something heavy was added to 0.10-x version of Avalonia. I don't remember details though.

MichalStrehovsky:
We don't cache the results of virtual method resolution at all and the algorithm we have is pretty slow.

There have been [several attempts](https://github.com/dotnet/corert/pull/3872) to fix GVM analysis, but generic virtual methods are just too annoying - this is the _one thing_ I envy C++ - no templated virtual methods there.

It's possible this is just hitting some bad pattern. I don't know much about this part of the codebase. David and Fadi were working in this area. I'll have to ramp up on this for default interface methods at some point.

MichalStrehovsky:
https://github.com/AvaloniaUI/Avalonia/pull/3255 is the source of bad compilation throughput in more recent Avalonia. It introduced a number of generic interface methods on interfaces that are fundamental to Avalonia and implemented by dozens of classes, some of them even generic. We are spending most of the time figuring out what are all the possible method bodies and data structures we need to generate.

[...]

PR #493, opened 2020-12-22
Generic virtual method analysis tweaks

MichalStrehovsky:
Three tweaks:

* Do not go into the SearchDynamicDependencies codepath if the method cannot be overriden. I believe the SearchDynamicDependencies wouldn't come up with anything for these cases, so we might as well skip that. This mostly helps generic virtual method use in LINQ because those classes are sealed. Doesn't help much in Avalonia.
* Don't do a while loop over the inheritance chain in SearchDynamicDependencies. I believe the dependencies on less derived methods will be injected as part of those GVMDependencies nodes so this feels unnecessary.
* Don't mark EETypes as interesting for GVM analysis if they don't provide an implementation of an interface method that is _new_ in this inheritance hierarchy.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
@@ -59,7 +59,19 @@ namespace ILCompiler.DependencyAnalysis
                                 MethodDesc genericDefinition = method.GetMethodDefinition();
                                 MethodDesc slotDecl = _type.ResolveInterfaceMethodTarget(genericDefinition);
                                 if (slotDecl != null)
-                                    return true;
+                                {
+                                    // If the type doesn't introduce this interface method implementation (i.e. the same implementation
+                                    // already exists in the base type), do not consider this type interesting for GVM analysis just yet.
+                                    //
+                                    // We need to limit the number of types that are interesting for GVM analysis at all costs since
+                                    // these all will be looked at for every unique generic virtual method call in the program.
+                                    // Having a long list of interesting types affects the compilation throughput heavily.
+                                    if (slotDecl.OwningType == _type ||
+                                        _type.BaseType.ResolveInterfaceMethodTarget(genericDefinition) != slotDecl)
+                                    {
+                                        return true;
+                                    }
+                                }
                             }
                         }
                     }


MichalStrehovsky:
So this drops the compilation time of the Avalonia demo app in #452 from ~12 minutes to:
* ~9 minutes after the first commit
* 2 minutes and 40 seconds after the second commit

Hopefully I didn't break anything,

MichalStrehovsky:
> Azure Pipelines failed to run 1 pipeline(s).

Thanks for nothing, Azure pipelines. Great failure logs!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
```diff
@@ -114,10 +121,10 @@ public override bool HasDynamicDependencies
 
         public override IEnumerable<CombinedDependencyListEntry> SearchDynamicDependencies(List<DependencyNodeCore<NodeFactory>> markedNodes, int firstNode, NodeFactory factory)
         {
-            Debug.Assert(_method.IsVirtual && _method.HasInstantiation);
```
MichalStrehovsky, line 117: Deleting this because we assert this in the ctor.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
```diff
@@ -130,13 +137,11 @@ public override IEnumerable<CombinedDependencyListEntry> SearchDynamicDependenci
                 if (!potentialOverrideType.IsDefType)
                     continue;
 
-                Debug.Assert(!potentialOverrideType.IsRuntimeDeterminedSubtype);
```
MichalStrehovsky, line 133: Deleting this because IsRuntimeDeterminedSubtype is expensive and getting one here would mean we already have a bad EEType in the system. EETypeNode asserts this is not the case.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
```diff
@@ -186,23 +191,10 @@ public override IEnumerable<CombinedDependencyListEntry> SearchDynamicDependenci
                             continue;
                     }
 
-                    overrideTypeCur = potentialOverrideType;
```
MichalStrehovsky, line 189: Couldn't figure out when this loop would have a meaningful impact. We could consider not checking in this part of the change and keep the loop, but I don't like to have code around that nobody understands and we have no test coverage for (there's apparently no failing test with this deleted even though we test generic virtual methods in the Generics test and in the DynamicGenerics test).

jkotas (approved):
Thank you!

PR #494, opened 2020-12-22
Enable build on s390x

nealef:
- Add s390x as a ProcessorArchitecture & enumerated architecture


akoeplinger (approved):

akoeplinger:
Let's go with #495

PR #498, opened 2020-12-24
NOP out bytes in unreachable basic blocks

MichalStrehovsky:
Fixes a regression in:

JIT\Methodical\eh\deadcode\deadoponerrorinfunclet_r

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
@@ -225,13 +225,14 @@ namespace ILCompiler
                         if (reader.HasNext)
                         {
                             // If the bytes following this basic block are not reachable from anywhere,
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
```diff
@@ -221,7 +221,20 @@ public MethodIL GetMethodILWithInlinedSubstitutions(MethodIL method)
                     }
 
                     if ((flags[offset] & OpcodeFlags.EndBasicBlock) != 0)
+                    {
+                        if (reader.HasNext)
+                        {
+                            // If the bytes following this basic block are not reachable from anywhere,
+                            // the flags would make it look like the current instruction is just veeery long
```
jkotas, line 228:
```suggestion
                            // the flags would make it look like the current instruction is just very long
```
I have read this as "veery" that is actually a bird name.

jkotas (approved):
LGTM

PR #499, opened 2020-12-24
Fix matching of direct call modules with suffix

### LINKED ISSUE #497: Has issue from dotnet/corert issue 8241 and dotnet/runtimelab pull 98 resurfaced?
The original post from [dotnet/corert#8241](https://github.com/dotnet/corert/issues/8241) mentioned a problem loading `System.Security.Cryptography.Native.Apple`. Reading through the thread, I was under the impression the problem was fixed by [dotnet/runtimelab#98](https://github.com/dotnet/runtimelab/pull/98), but I am still encountering the following error when running my example app:

```
Unhandled Exception: System.AggregateException: One or more errors occurred. (The SSL connection could not be established, see inner exception.)
 ---> System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception.
 ---> System.Security.Authentication.AuthenticationException: Authentication failed, see inner exception.
 ---> System.DllNotFoundException: Unable to load DLL 'libSystem.Security.Cryptography.Native.Apple.dylib': The specified module could not be found.
```

I'm not sure how much of the below is helpful, but here is info on my setup:

**.net: 5.0.101**
**macOS: 10.15.7**
**Microsoft.DotNet.ILCompiler: 6.0.0-alpha.1.20622.2**
#### .csproj
```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="6.0.0-*" />
  </ItemGroup>

</Project>
```
#### code
```csharp
using System;
using System.Net.Http;
using System.Threading.Tasks;

namespace Test
{
    class Program
    {
        private static readonly HttpClient client = new HttpClient();

        {
            SendRequest().Wait();
        }

        private static async Task SendRequest()
        {
            var response = await client.GetStringAsync("https://jsonplaceholder.typicode.com/todos/1");
            Console.WriteLine(response);
            return;
        }
    }
}
```

jkotas:
Fixes #497

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler/ConfigurablePInvokePolicy.cs
+++ b/src/coreclr/tools/aot/ILCompiler/ConfigurablePInvokePolicy.cs
@@ -84,15 +84,17 @@ namespace ILCompiler
 
             if (_target.IsWindows)
             {
-                if (name.EndsWith(".dll", StringComparison.OrdinalIgnoreCase))
-                    yield return name.Substring(0, name.Length - 4);
+                string suffix = ".dll";
+
+                if (name.EndsWith(suffix, StringComparison.OrdinalIgnoreCase))
+                    yield return name.Substring(0, name.Length - suffix.Length);
             }
             else
             {
                 string suffix = _target.IsOSX ? ".dylib" : ".so";
 
                 if (name.EndsWith(suffix, StringComparison.Ordinal))
-                    yield return name.Substring(0, suffix.Length);
+                    yield return name.Substring(0, name.Length - suffix.Length);
             }
         }
 


MichalStrehovsky (approved):

src/coreclr/tools/aot/ILCompiler/ConfigurablePInvokePolicy.cs
```diff
@@ -84,15 +84,17 @@ private IEnumerable<string> ModuleNameVariations(string name)
...
+                string suffix = ".dll";
+
+                if (name.EndsWith(suffix, StringComparison.OrdinalIgnoreCase))
+                    yield return name.Substring(0, name.Length - suffix.Length);
             }
             else
             {
                 string suffix = _target.IsOSX ? ".dylib" : ".so";
 
                 if (name.EndsWith(suffix, StringComparison.Ordinal))
```
hez2010, line 96: macOS is not case-sensitive so I think the suffix comparasion logic should be the same with Windows for OSX.

jkotas: CoreCLR uses case-sensitive comparisons for these as well, e.g.  https://github.com/dotnet/runtime/blob/e98fb61d8b4bb6687ccddead861d8b140751647b/src/coreclr/vm/dllimport.cpp#L6062

since there is a lof

PR #502, opened 2020-12-24
NativeAOT-LLVM: Merge nativeaot

yowl:
Merge latest from NativeAOT.  One modification which is the additional `AND NOT CLR_CMAKE_TARGET_ARCH_WASM` as this code moved from `configurecompiler.cmake` and that's what the logic was there.

https://github.com/yowl/runtimelab/blob/1e0622671fb81111d3bbd630b8edfccbe3015cb5/eng/native/configuretools.cmake#L66


jkotas:
Good to merge?

PR #505, opened 2020-12-26
NativeAOT: fix for Stloc ldloc pattern feature switch branch elimination 

yowl:
This PR fixes a couple of small problems with the IL detection for the pattern
```
  IL_0048:  stloc.s     4
  IL_004A:  ldloc.s     4
  IL_004C:  brfalse.s   IL_005B
```
Added a test which is a bit crude, which generates this pattern in debug mode.  If there's something better then let me know.

@MichalStrehovsky


MichalStrehovsky (approved):
Thank you!

PR #506, opened 2020-12-28
Generate new debug info for methods with substituted code

MichalStrehovsky:
Methods with substituted code weren't getting any debug information. Fixing that.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
@@ -446,12 +446,13 @@ namespace ILCompiler
             // in generating debug information for NOPs) - generate new debug information by filtering
             // out the sequence points associated with nopped out instructions.
             MethodDebugInformation debugInfo = method.GetDebugInfo();
-            if (debugInfo != null)
+            IEnumerable<ILSequencePoint> oldSequencePoints = debugInfo?.GetSequencePoints();
+            if (oldSequencePoints != null)
             {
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
```diff
@@ -441,7 +442,25 @@ public MethodIL GetMethodILWithInlinedSubstitutions(MethodIL method)
...
+            // Existing debug information might not match new instruction boundaries (plus there's little point
+            // in generating debug information for NOPs) - generate new debug information by filtering
+            // out the sequence points associated with nopped out instructions.
+            MethodDebugInformation debugInfo = method.GetDebugInfo();
+            if (debugInfo != null)
+            {
+                ArrayBuilder<ILSequencePoint> sequencePoints = new ArrayBuilder<ILSequencePoint>();
+                foreach (var sequencePoint in debugInfo.GetSequencePoints())
+                {
+                    if ((flags[sequencePoint.Offset] & OpcodeFlags.Mark) != 0)
```
jkotas, line 454: It is not unusual to see bad PDBs. Should we silently ignore out-of-range sequencePoints here?

jkotas (approved):

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
MichalStrehovsky: Yep, fixed!

PR #507, opened 2020-12-28
Fix custom attribute metadata processing

MichalStrehovsky:
This is a preparatory step to add support for IL Linker's conditional custom attributes: https://github.com/mono/linker/blob/e68a1c586b5b5174e21e3f0d4b00322c547e4c3d/docs/data-formats.md#conditional-custom-attributes

We were already capable of filtering what custom attributes get generated, but this was mostly to filter out bad custom attributes, or custom attributes that refer to reflection-blocked types. This logic was hardcoded in two places (one that computes general dependencies, the other that produces the actual metadata).

I'm centralizing the logic into the "general dependencies" location. It will be trivial to extend this to add the additional filtering.


jkotas (approved):
LGTM

PR #510, opened 2020-12-29
Respect ILLink annotations to remove custom attributes

MichalStrehovsky:
Removes a couple attribute when support for managed debugging and/or EventSource is disabled.


jkotas:
Do we also want the trimming done by `src\mono\netcore\System.Private.CoreLib\src\ILLink\ILLink.LinkAttributes.xml` ?

jkotas (approved):
Thank you!

MichalStrehovsky:
> Do we also want the trimming done by `src\mono\netcore\System.Private.CoreLib\src\ILLink\ILLink.LinkAttributes.xml` ?

I'm still thinking about the right course of action for that. Those attributes are removed unconditionally and the user experience around that is pretty bad:

* By default you don't get them, which is great if you don't care about them.
* If you do want them though (because e.g. you care about nullable annotations at runtime), you need to first find out what the problem is (linker stripped it without a warnings). Once you find that out, you'll probably need to ask around because you would never figure out how to fix the problem you have.

To beat the default that says "kill them" in CoreLib, you need to add a `<type fullname="TheAttribute" />` entry to your ILLinkTrim file and that one will beat the attribute removal in CoreLib (basically if you specify you want to keep the type, linker will also keep the attributes of that type). I don't emulate this rather nonsensical behavior in the Native AOT compiler.

I think these all should be conditioned on feature switches and moved to the shared ILLinkTrim file. Some IL Linker devs believe that the above workaround is sufficient and I got tired of fighting that.

PR #512, opened 2020-12-30
Remove Object and UIntPtr as valid cast from pointer

### LINKED ISSUE #189: typeof(object).IsAssignableFrom(typeof(byte*)) should be false
JIT/Intrinsics/TypeIntrinsics_r

ThomasGoulet73:
Removes Object and UIntPtr as valid cast from pointer.

Fixes the following code which returned true and now returns false:
```
typeof(object).IsAssignableFrom(typeof(byte*))
typeof(UIntPtr).IsAssignableFrom(typeof(byte*))
```

Fixes #189

### DIFF
--- a/src/tests/issues.targets
+++ b/src/tests/issues.targets
@@ -1169,12 +1169,6 @@
         <ExcludeList Include="$(XunitTestBinBase)/JIT/HardwareIntrinsics/General/Vector256_1/Vector256_1_ro/*">
             <Issue>https://github.com/dotnet/runtimelab/issues/184</Issue>
         </ExcludeList>
-        <ExcludeList Include="$(XunitTestBinBase)/JIT/Intrinsics/TypeIntrinsics_r/*">
-            <Issue>https://github.com/dotnet/runtimelab/issues/189</Issue>
-        </ExcludeList>
-        <ExcludeList Include="$(XunitTestBinBase)/JIT/Intrinsics/TypeIntrinsics_ro/*">
-            <Issue>https://github.com/dotnet/runtimelab/issues/189</Issue>
-        </ExcludeList>
         <ExcludeList Include="$(XunitTestBinBase)/JIT/jit64/opt/cse/hugeSimpleExpr1/*">
             <Issue>https://github.com/dotnet/runtimelab/issues/190</Issue>
         </ExcludeList>


MichalStrehovsky (approved):
Thank you!

I spent some time research why this code was there in the first place, since this was written by Ati and that code is usually _very well_ researched.

Seems like .NET Core did a breaking change in this behavior. The code being deleted is compatible with desktop .NET Framework and that's why it was there.

MichalStrehovsky:
Ah, could you also delete these lines:

https://github.com/dotnet/runtimelab/blob/1eda1978dfe2eccb6718786b13b3c9dbad2e54e7/src/tests/issues.targets#L1172-L1177

So that we run the tests in the CI.

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=512) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=512) before we can accept your contribution.<br/><br/>:x: ThomasGoulet73 [sign now](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=512)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/runtimelab?pullRequest=512) it.</sub>

ThomasGoulet73:
I wasn't able to test locally, I kept getting errors while building with this command: `src\tests\build.cmd nativeaot Release`. I got the following error: `The CoreCLR artifacts path does not exist 'XXX\Runtimeab\artifacts\bin\coreclr\windows.x64.Release\'`. The weird thing is that the path XXX\Runtimeab\artifacts\bin\coreclr\windows.x64.**Debug**\ even though I built using release. I'll try to get my local build working if the CI fails.

MichalStrehovsky:
If the Debug directory exists, the runtime must have been built with Debug configuration.

`build.cmd nativeaot+libs+nativeaot.packages -c Release` will build everything as Release. `build.cmd nativeaot+libs+nativeaot.packages -rc Debug -lc Release` will build the runtime (runtime config) as Debug, and libraries (library config) as release. You can combine these in various ways. The runtime test scripts always assume the libraries were built in Release config and the config that is passed to the script is the Runtime config.

We really need a more user friendly approach to the development inner loop. I rarely run the test scripts locally because it's a bit annoying.

PR #514, opened 2020-12-30
Optimize variant interface implementation generation

MichalStrehovsky:
Saves 106 kB on a Hello World app.

Previously when the compiler saw we're generating a type that implements variant interfaces, we would throw hands in the air and generate the implementations of all variant interface methods on the type since we had no idea which ones would be actually needed.

This has drawbacks, especially around array types that implement a lot of variant (or variant-like) interfaces.

This commit introduces tracking of individual variant interface methods being used. It lets us avoid generating implementations of variant interface methods that are not statically seen as necessary.

We track this by adding conditional dependencies on a new `VariantInterfaceMethodUseNode` instead of the old logic that just generated hard `VirtualMethodUse` dependencies.

We try to further optimize when a virtual method call or interface method implementation is actually considered variant. E.g. an implementation or a call to `IEnumerable<MyStruct>` is not actually variant since we can't cast it to any other `IEnumerable<T>`.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
@@ -288,7 +288,7 @@ namespace ILCompiler.DependencyAnalysis
                             yield return new CombinedDependencyListEntry(factory.VirtualMethodUse(implMethod), factory.VirtualMethodUse(interfaceMethod), "Interface method");
 
                             // If any of the implemented interfaces have variance, calls against compatible interface methods
-                            // could result in interface methods of this type being used (e.g. IEnumberable<object>.GetEnumerator()
+                            // could result in interface methods of this type being used (e.g. IEnumerable<object>.GetEnumerator()
                             // can dispatch to an implementation of IEnumerable<string>.GetEnumerator()).
                             if (isVariantInterfaceImpl)
                             {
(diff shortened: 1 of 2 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/VariantInterfaceMethodUseNode.cs
```diff
@@ -0,0 +1,155 @@
...
+            _decl = decl;
+        }
+
+        public static bool IsVariantInterfaceImplementation(NodeFactory factory, TypeDesc providingType, DefType implementedInterface)
+        {
+            Debug.Assert(implementedInterface.IsInterface);
+            Debug.Assert(!implementedInterface.IsGenericDefinition);
+
+            // If any of the implemented interfaces have variance, calls against compatible interface methods
+            // could result in interface methods of this type being used (e.g. IEnumberable<object>.GetEnumerator()
```
jkotas, line 53:
```suggestion
            // could result in interface methods of this type being used (e.g. IEnumerable<object>.GetEnumerator()
```

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -284,6 +286,16 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
                         if (implMethod != null)
                         {
                             yield return new CombinedDependencyListEntry(factory.VirtualMethodUse(implMethod), factory.VirtualMethodUse(interfaceMethod), "Interface method");
+
+                            // If any of the implemented interfaces have variance, calls against compatible interface methods
+                            // could result in interface methods of this type being used (e.g. IEnumberable<object>.GetEnumerator()
```
jkotas, line 291:
```suggestion
                            // could result in interface methods of this type being used (e.g. IEnumerable<object>.GetEnumerator()
```

jkotas (approved):

jkotas:
Nice!

MichalStrehovsky:
Sigh, the Pri-0 runs have a lot of failure. I bet I messed up the IL rewriter again. I'll take a look tomorrow.

PR #515, opened 2020-12-30
Implement ILLinkTrim.xml-based resource filtering

MichalStrehovsky:
Adds support for the `resource` element in ILLinkTrim files. This is the facility IL Linker (in cooperation with framework ILLinkTrim files) uses to remove framework resource strings.

This replaces the existing mechansm in ILCompiler that was exposed with `--removefeature:FrameworkStrings`.

### DIFF
--- a/src/coreclr/nativeaot/Directory.Build.targets
+++ b/src/coreclr/nativeaot/Directory.Build.targets
@@ -3,5 +3,11 @@
   <!-- resources.targets need to be imported before the Arcade SDK. -->
   <Import Project="$(RepositoryEngineeringDir)resources.targets" />
 
+  <PropertyGroup>
+    <!-- resources.targets will "helpfully" opt us into this in debug builds, but we don't really want
+         a debug/checked/release difference on this. It makes resx strings removal just harder to test. -->
+    <GenerateResxSourceIncludeDefaultValues>false</GenerateResxSourceIncludeDefaultValues>
+  </PropertyGroup>
+  
   <Import Project="..\Directory.Build.targets" />
 </Project>


src/coreclr/nativeaot/System.Private.CoreLib/System.Private.CoreLib.csproj
```diff
@@ -3,14 +3,12 @@
   <!-- Note that various places in SPCL depend on this resource name i.e. TplEventSource -->
   <ItemGroup Label="Embedded Resources">
     <EmbeddedResource Include="src\Resources\Strings.resx">
+      <LogicalName>System.Private.CoreLib.Strings.resources</LogicalName>
```
MichalStrehovsky, line 6: The default logical name included `.src.` for some reason (it basically looked like a relative path to the csproj). I couldn't reverse engineer why.

The libraries partition has a bunch of convention-based rules in `eng/resources.targets` that we don't include because `src/coreclr` doesn't follow the convention. eng/illink.targets basically expects that and Mono's and CoreCLR's CoreLib.csproj files were basically hacked to make it work without including `resources.targets`.

jkotas: Does the same problem exists for all other `src/coreclr/nativeaot/System.*` assemblies?

jkotas: Ok, I guess the other assemblies have slightly different directory structure that makes it work.

jkotas, line 6:
```suggestion
      <ManifestResourceName>System.Private.CoreLib.Strings</ManifestResourceName>
```
Setting ManifestResourceName should cut off the problem closer to where it gets created.

MichalStrehovsky: I'm looking into moving the CoreLib project into src like they do in the libraries partition. The existing structure is inspired by CoreCLR's CoreLib, but it's inconsistent.

The test failures are because we can't find CoreLib's resource strings because of this hack :(

MichalStrehovsky: OK, I think I managed to make the current state an improvement instead of adding a hack.

jkotas (approved):
Thank you!

PR #516, opened 2020-12-31
Fully skip uninteresting instruction

MichalStrehovsky:


MichalStrehovsky:
Sigh, this is still failing.

We remove the leave.s at offset 0x11 because it's unreachable and replace it with two nops. Then RyuJIT barfs because as part of the business where it looks at unreachable basic blocks (for whatever reasons), it finds out the unreachable basic block falls through without `leave`.

```
  .try IL_0006 to IL_0013 fault handler IL_0013 to IL_0026
  IL_0000:  ldc.i4      3
  IL_0005:  stloc.0
  IL_0006:  ldstr       "message"
  IL_000B:  newobj      instance void [System.Private.CoreLib]System.Exception::.ctor(string)
  IL_0010:  throw
  IL_0011:  leave.s     IL_0063
  IL_0013:  ldstr       "in fault block."
  IL_0018:  call        void [System.Console]System.Console::WriteLine(string)
  IL_001D:  ldloc.0
  IL_001E:  ldc.i4      1
  IL_0023:  sub
  IL_0024:  stloc.0
  IL_0025:  endfinally
```

jkotas (approved):

MichalStrehovsky:
The tests are still failing. I haven't looked at them, but I just realized that I fixed the one repro case, but if the unreachable block is longer than 2 bytes, the same problem is going to repeat again (the NOPs after the `br $-2` are going to upset RyuJIT).

This might be a dead end. We either need to fix RyuJIT to stop looking at junk unreachable bytes, or actually reformat the IL so that we don't have NOPs. I'm about to cry.

MichalStrehovsky:
I'll still merge this because it reduces Pri0 failure noise by more than a half and I want to run Pri0 on the other change, but it will need to be redone.

jkotas:
> (the NOPs after the br $-2 are going to upset RyuJIT

Can this be fixed by placing the `br $-2` at the end of the unreachable run?

MichalStrehovsky:
> Can this be fixed by placing the `br $-2` at the end of the unreachable run?

Ah. That might just work! Thanks. I'll try it tomorrow.

PR #517, opened 2021-01-02
Ifdef out dump debugging support

MichalStrehovsky:
Saves 40 kB on a Hello World app.

This is the Windows-specific, .NET Native specific, dump debugging support. It serializes exception information into a format the managed dump debugger understands.

We don't have the minidump auxiliary provider to collect this data for Native AOT, nor a debugger to interpret the data.

Crash dumps have a "native" experience with no managed niceties.


jkotas (approved):
Thanks!

PR #518, opened 2021-01-02
Speed up reliable array copy

MichalStrehovsky:
We were calling `Array.SetValue`, but that one does unnecessary checks for widening. At this point we know the arrays are of the same type.

Also makes Hello World 31 kB smaller because we can get rid of the widening infrastructure.


jkotas (approved):
Thanks!

PR #519, opened 2021-01-02
Make typeof a tiny bit faster

MichalStrehovsky:
`Type.GetTypeFromHandle` does a check for EEType being null, but this helper is called for the ldtoken/GetTypeFromHandle sequence where the handle will not end up being null.


jkotas (approved):
Thanks!

PR #521, opened 2021-01-02
Merge from dotnet/runtime

jkotas:


Directory.Build.props
```diff
@@ -228,6 +228,8 @@
     <!-- We don't want Private packages to be shipped to NuGet.org -->
     <IsShippingPackage Condition="($(MSBuildProjectName.Contains('Private')) or $(IsExperimentalAssembly)) and '$(MSBuildProjectExtension)' == '.pkgproj'">false</IsShippingPackage>
     <!-- A package isn't generated (in traversal builds) if in servicing or in runtimelab. Intended to be overridden at project level. -->
+    <!-- Disable building of libraries packages in this branch -->
+    <GeneratePackage>false</GeneratePackage>
```
MichalStrehovsky, line 232: @jkotas What is this working around? It seems like it disabled generation of Microsoft.DotNet.ILCompiler as well.

jkotas: #528 should fix it.

dotnet/runtimelab publishing is not fully independent from dotnet/runtime publishing. Building and publishing all packages may result into collisions, and it is also wasteful.

PR #523, opened 2021-01-03
Mangle FunctionPointerType instances

wjk:
When I was test-compiling a WPF app using NativeAOT, it crashed when trying to mangle a function pointer type. I have thus introduced a case to handle `FunctionPointerType`. With this patch applied, ILC gets much further into the compilation process before crashing. Let me know if you want me to mangle it to something different. Thanks!

### DIFF
--- a/src/coreclr/tools/Common/Compiler/CoreRTNameMangler.cs
+++ b/src/coreclr/tools/Common/Compiler/CoreRTNameMangler.cs
@@ -334,6 +334,11 @@ namespace ILCompiler
                     {
                         mangledName = GetPrefixMangledTypeName((IPrefixMangledType)type).ToString();
                     }
+                    else if (type is FunctionPointerType pointerType)
+                    {
+                        if (_mangleForCplusPlus) throw new NotImplementedException("Cannot mangle function pointer types when emitting C++");
+                        mangledName = "<fptr>" + pointerType.Signature.ToString();
+                    }
                     else
                     {
                         // This is a type definition. Since we didn't fall in the `is EcmaType` case above,


src/coreclr/tools/Common/Compiler/CoreRTNameMangler.cs
```diff
@@ -334,6 +334,11 @@ private string ComputeMangledTypeName(TypeDesc type)
                     {
                         mangledName = GetPrefixMangledTypeName((IPrefixMangledType)type).ToString();
                     }
+                    else if (type is FunctionPointerType pointerType)
+                    {
+                        if (_mangleForCplusPlus) throw new NotImplementedException("Cannot mangle function pointer types when emitting C++");
+                        mangledName = "<fptr>" + pointerType.Signature.ToString();
```
MichalStrehovsky, line 340: `ToString` is not unique enough to guarantee we will not conflict with a different type - this will need to be more complex.

I don't expect we'll need mangling for function pointers until https://github.com/dotnet/runtime/issues/11354 is resolved over in mainstream runtimes.

We end up needing one in latest WPF because the compiler ends up creating a generic instance over a function pointer type internally, but that's a different bug. I've submitted #525 to fix that.

MichalStrehovsky (commented):

If you would like to continue working on the mangling, we can do that, but #525 is going to fix the immediate issue in WPF.

Note that last time I looked into WPF with CoreRT/NativeAOT, it had COM in an unavoidable (without recompiling WPF) code path at startup.

wjk:
Thanks @MichalStrehovsky. I'll close this issue in favor of #525, since that implementation is much more correct. (Additionally, since the dotnet/wpf repo is not accepting external contributions, I have created a fork and am now working on that. The plan is to make my fork of WPF accessible via NuGet. Therefore, I apply whatever changes are needed to fix COM support.)

PR #525, opened 2021-01-03
Check for function pointers in SupportsSignature more thoroughly

MichalStrehovsky:
Repro case hit a pointer to a function pointer in a signature that this check wasn't covering. Reflection-invoking things that have function pointers in them is a TODO.


jkotas (approved):
Thanks!

PR #526, opened 2021-01-04
Place endless loop at the end of unreachable basic blocks

MichalStrehovsky:
RyuJIT's importer looks at unreachable basic blocks (and even considers completely unreachable bytes as basic blocks). It enforces various rules on those bytes, such as they cannot fall through to the next basic block if they were part of an EH region.

Having a string of NOPs terminated by an endless loop should be a fine replacement for anything. Endless loop followed by a string of NOPs doesn't work because of the above.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
@@ -276,6 +276,12 @@ namespace ILCompiler
                                 offsetsToVisit.Push(ehRegion.FilterOffset);
 
                             offsetsToVisit.Push(ehRegion.HandlerOffset);
+
+                            // RyuJIT is going to look at this basic block even though it's unreachable.
+                            // Consider it visible so that we replace the tail with an endless loop.
+                            int handlerEnd = ehRegion.HandlerOffset + ehRegion.HandlerLength;
+                            if (handlerEnd < flags.Length)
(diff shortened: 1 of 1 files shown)


jkotas (approved):
LGTM. Thanks!

MichalStrehovsky:
Cool cool, down to 5 failures, none of them related to dead basic block elimination:

```
F:\workspace\_work\1\s\artifacts\tests\coreclr\windows.x64.Release\JIT\Regression\VS-ia64-JIT\V1.2-M01\b10827\b10827\b10827.cmd (15 minutes 0 seconds)
F:\workspace\_work\1\s\artifacts\tests\coreclr\windows.x64.Release\JIT\Intrinsics\TypeIntrinsics_ro\TypeIntrinsics_ro.cmd (27 seconds)
F:\workspace\_work\1\s\artifacts\tests\coreclr\windows.x64.Release\JIT\SIMD\Vector3Interop_ro\Vector3Interop_ro.cmd (15 seconds)
F:\workspace\_work\1\s\artifacts\tests\coreclr\windows.x64.Release\JIT\SIMD\Vector3Interop_r\Vector3Interop_r.cmd (14 seconds)
F:\workspace\_work\1\s\artifacts\tests\coreclr\windows.x64.Release\baseservices\exceptions\stackoverflow\stackoverflowtester\stackoverflowtester.cmd (11 seconds)
```

One is a timeout, two are probably the same interop/type layout bug, one is friendly stack overflow we don't have.

And the TypeIntrinsic test is using the `dynamic` keyword for reasons that escape me.

PR #529, opened 2021-01-05
Place exception handling info in a separate section

MichalStrehovsky:
Addressing the TODO in the ObjectWriter.

Does two things:
* Improves locality (GC info is more tightly packed together and the rarely needed exception handling information is out of the way)
* Gives opportunity for constant blob folding - byte-identical EH info blobs can be shared in theory. In practice I'm not seeing savings.

This will need tweaks in crossgen2 since it touched shared files. I was hoping we would get something out of the blob folding. I'm not sure how much this is worth it.


jkotas (approved):
LGTM. Thanks!

PR #530, opened 2021-01-05
Synchronize Win32 library list with the .targets file

MichalStrehovsky:
The repro project was linking against a different set of Win32 import libraries than Microsoft.DotNet.ILCompiler. Hit the missing crypt32.lib.


jkotas (approved):

jkotas:
FrameworkStrings test failed with timeout.

Issue #446, opened 2020-12-11
Delete --removefeature ILC command line option


MichalStrehovsky:
We should be able to replace this with the "official" feature switches: #422

This item tracks:
* Creating an XML descriptor for the existing features (some are already covered by what's in dotnet/runtime).
* Deleting the handling of --removefeature and related code from the compiler

MichalStrehovsky:
One thing that is left before we can delete this command line option is `--removefeature:Comparers`. I can't decide about the right course of action.

The current mechanism is a bit inconsistent. It rewrites the fallback method body, which means that if `Comparer/EqualityComparer<Foo>.Default` is statically known, the compiler will still do the method body rewriting and produce a "good" comparer (one that respects whether `Foo` implements `IComparable<T>`/`IEquatable<T>`). If we don't know the answer statically (because we're e.g. in shared code), it will do the fallback and the fallback with `--removefeature:Comparers` will ignore the generic interfaces.

We could make this fully consistent (never respect the generic interfaces), but this code also interacts with RyuJIT (RyuJIT knows what concrete type `get_Default` is supposed to return), so it can't just be done with ILLinkTrim files. Being fully consistent would also means we get boxing for valuetypes, which is pretty bad.

This would push me towards just won't fixing this and dropping the feature.

But on the other hand, a reflection-free Hello World is currently 910 kB (with `--removefeature:Comparers`). Once we drop that argument, the size jumps to 1,370 kB because of all the type loader junk to load the appropriate comparer types.

jkotas:
> --removefeature:Comparers. I can't decide about the right course of action.

If we care enough, I think we can preserve the current inconsistent behavior with low cost by ILLinkTrim files and a property to tell that we are in the reflection-free mode. Something like: `ReflectionFreeMode ? new ObjectComparer<T>() : ... the type loader path ...`.

Also, is there a replacement XmlDownloadNonFileStream?

MichalStrehovsky:
Fixed in #797.

PR #532, opened 2021-01-05
Add substitutions to allow disabling serialization guard

MichalStrehovsky:
Contributes to #446.

I could try my luck pushing this into dotnet/runtime, but I already had a discussion on it with the security people and I have better things to do (https://github.com/dotnet/runtime/issues/29212).

### DIFF
--- a/src/libraries/System.Diagnostics.Process/src/ILLink/ILLink.Substitutions.xml
+++ b/src/libraries/System.Diagnostics.Process/src/ILLink/ILLink.Substitutions.xml
@@ -1,0 +1,7 @@
+<linker>
+  <assembly fullname="System.Diagnostics.Process">
+    <type fullname="System.Runtime.Serialization.SerializationGuard">
+      <method signature="System.Void ThrowIfDeserializationInProgress(System.String,System.Int32&amp;)" body="stub" feature="System.Runtime.Serialization.SerializationGuard.IsSupported" featurevalue="false" />
+    </type>
+  </assembly>
+</linker>
--- a/src/libraries/System.Diagnostics.Process/src/System.Diagnostics.Process.csproj
+++ b/src/libraries/System.Diagnostics.Process/src/System.Diagnostics.Process.csproj
@@ -7,6 +7,13 @@
     <!-- Suppress unused field warnings when using PlatformNotSupportedException stubs -->
     <NoWarn Condition=" '$(TargetsFreeBSD)' == 'true' or '$(TargetsUnknownUnix)' == 'true' ">$(NoWarn);0649</NoWarn>
   </PropertyGroup>
+  <!-- ILLinker settings -->
+  <PropertyGroup>
+    <ILLinkDirectory>$(MSBuildThisFileDirectory)ILLink\</ILLinkDirectory>
+  </PropertyGroup>
+  <ItemGroup>
+    <ILLinkSubstitutionsXmls Include="$(ILLinkDirectory)ILLink.Substitutions.xml" />
+  </ItemGroup>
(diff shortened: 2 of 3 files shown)


src/libraries/System.Diagnostics.Process/src/ILLink/ILLink.Substitutions.xml
```diff
@@ -0,0 +1,7 @@
+<linker>
+  <assembly fullname="System.Diagnostics.Process">
+    <type fullname="System.Runtime.Serialization.SerializationGuard">
+      <method signature="System.Void ThrowIfDeserializationInProgress(System.String,System.Int32&amp;)" body="stub" feature="System.Runtime.Serialization.SerializationGuard.IsSupported" featurevalue="false" />
```
jkotas, line 4: Can all substitutions for this be in CoreLib instead of being spread around?

jkotas: Ah, this is to trim the reflection glue.

jkotas:
I think it may be better to remove this feature switch completely and just stick with what we have in dotnet/runtime. My original reason for introducing this feature switch was to workaround for the stale CoreFX that was not matching CoreLib.

jkotas:
If we want to avoid the reflection, we can change System.Diagnostics.Process to build against CoreLib. It should be fine to push that change into dotnet/runtime.

MichalStrehovsky:
Yeah, one thing is the reflection. The other thing is dragging AsyncLocal and related infrastructure just because someone uses e.g. a FileStream. I guess I would need to measure how much that costs.

MichalStrehovsky:
We don't really need this right now. It can be a bugfix later. I'm not going to block deleting `--removefeature` on this.

jkotas:
If somebody adds this `ILLink.Substitutions.xml` file locally to their app, can they still strip this if they really want to (both in shipping .NET Core and in nativeaot)?

MichalStrehovsky:
> If somebody adds this `ILLink.Substitutions.xml` file locally to their app, can they still strip this if they really want to (both in shipping .NET Core and in nativeaot)?

Yeah, that should work. But it means they're hardcoding implementation details of the framework into their app and I'm not a huge fan. It's an okay workaround though.

NativeAOT currently doesn't support passing ILLink.Substitutions.xml on the command line. I guess we should add that but then we also need to make the parser more robust and uncomment all the places in it that are supposed to generate warnings but are not hooked up.

jkotas:
> they're hardcoding implementation details of the framework into their app

Yeah, I agree. It may be useful to keep track of the non-productized framework substitutions that help with trimming a bit of size here and there (with appropriate disclaimer) to see what they can add up to over time.

> NativeAOT currently doesn't support passing ILLink.Substitutions.xml on the command line

How is it different from just embedding it into your main app assembly?

MichalStrehovsky:
> How is it different from just embedding it into your main app assembly?

The embedded XML can only substituted things within that assembly. Otherwise we could end up in a situation where we load an assembly and find out that code we already compiled long time ago should have been substituted with something else. IL Linker is removing support for this as well (it's allowed right now because ILLinkTrim processing happens as a prepass before mark&sweep, but will be removed because IL Linker is also moving towards the lazy design to help perf and fix longstanding issues such as `Type.GetType("Foo, SomeAssemblyThatWeDidntHearAboutYet")` not being supportable with IL Linker's existing design).

I guess the entrypoint assembly is special in the sense that we could look at it earlier but I don't know how that would work when e.g. building a DLL.

PR #533, opened 2021-01-06
Remove more code from the type loader

MichalStrehovsky:
* Universal shared code specific codepath
* Diagnostic codepath


jkotas (approved):
Thank you!

PR #540, opened 2021-01-08
Switch comparers to the ILLinkTrim plan

MichalStrehovsky:
Part of the planned deletion of `--removefeature`.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.xml
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.xml
@@ -3,10 +3,18 @@
     <method signature="System.Boolean get_IsDynamicCodeCompiled()" body="stub" value="false" />
     <method signature="System.Boolean get_IsDynamicCodeSupported()" body="stub" value="false" />
   </type>
+
   <type fullname="System.Collections.Generic.Comparer`1" feature="System.Collections.Generic.DefaultComparers" featurevalue="false">
     <method signature="System.Boolean get_SupportsGenericIComparableInterfaces()" body="stub" />
   </type>
   <type fullname="System.Collections.Generic.EqualityComparer`1" feature="System.Collections.Generic.DefaultComparers" featurevalue="false">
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Collections/Generic/Comparer.CoreRT.cs
```diff
@@ -14,13 +14,22 @@ public abstract partial class Comparer<T> : IComparer, IComparer<T>
...
             // The compiler will overwrite the Create method with optimized
             // instantiation-specific implementation.
             // This body serves as a fallback when instantiation-specific implementation is unavailable.
-            Interlocked.CompareExchange(ref s_default, Unsafe.As<Comparer<T>>(ComparerHelpers.GetComparer(typeof(T).TypeHandle)), null);
+            // If that happens, the compiler ensures we generate data structures to make the fallback work
+            // when this method is compiled.
+            Interlocked.CompareExchange(ref s_default,
+                SupportsGenericIComparableInterfaces
+                ? Unsafe.As<Comparer<T>>(ComparerHelpers.GetComparer(typeof(T).TypeHandle))
+                : new ObjectComparer<T>(),
```
jkotas, line 31: Could you please double check that there are no unreachable code or structures left around when `SupportsGenericIComparableInterfaces` is true? 

Maybe we should also do the trimming when `SupportsGenericIComparableInterfaces` is true.

jkotas (approved):
Thank you!

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Collections/Generic/Comparer.CoreRT.cs
MichalStrehovsky: Good call. We were generating a bunch of useless `ObjectEqualityComparer<Foo>` data structures because the scanner is computing generic dictionaries and scanner isn't inlining the `SupportsGenericIComparableInterfaces`. Fixed.

MichalStrehovsky:
Timeout in DynamicGenerics

PR #541, opened 2021-01-08
Do not root .cctor of constructed types

MichalStrehovsky:
The only reason why we root it is because of the `RunClassConstructor` API. This API is marked trimming unfriendly but has intrinsic recognition to work without warnings when used with known `System.Type` instances (it also works without warnings if the `System.Type` was annotated).

The rooting is purely defensive. We could do more defensive rootings elsewhere, but it's death by a thousand cuts so we usually don't. IL Linker won't root it on its own either.


jkotas (approved):
LGTM. Thanks!

PR #547, opened 2021-01-11
Align reflection settings with IL Linker

MichalStrehovsky:
Opening as a draft. The compiler changes should be what we need (they allow per-assembly specification of illinker's "copy", "link" and "copyused").

The end goal is to support the documented flow that works with ILLinker, but without ILLinker: https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options

The msbuild changes are tricky.

* We documented PrepareForILLink as the extension point to specify per-assembly settings, but I don't see a clear place for us to hook into that. ILLinker's targets file gets unconditionally included into SDK :(.
* The per-assembly IsTrimmable msbuild metadata that we set for the framework is getting lost because we're picking up the framework that ILCompiler ships with instead of the one from the SDK.

I'm still thinking about the right ways to approach this, but opening this if anyone has thoughts (@jkotas :)).

### DIFF
--- a/docs/using-nativeaot/optimizing.md
+++ b/docs/using-nativeaot/optimizing.md
@@ -26,7 +26,7 @@ The Native AOT compiler supports the [documented options](https://docs.microsoft
 
 By default, the compiler tries to maximize compatibility with existing .NET code at the expense of compilation speed and size of the output executable. This allows people to use their existing code that worked well in a fully dynamic mode without hitting issues caused by trimming. To read more about reflection, see the [Reflection in AOT mode](reflection-in-aot-mode.md) document.
 
-🛈 Native AOT difference: the `TrimMode` of framework assemblies is set to `link` by default. To compile entire framework assemblies, use `TrimmerRootAssembly` to root the select assemblies. It's not recommended to root the entire framework.
+🛈 Native AOT difference: the `TrimMode` of framework assemblies is set to `link` by default. To compile entire framework assemblies, use `TrimmerRootAssembly` to root the selected assemblies. It's not recommended to root the entire framework.
 
 To enable more aggressive removal of unreferenced code, set the `<TrimMode>` property to `link`.
 
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -102,8 +102,8 @@ The .NET Foundation licenses this file to you under the MIT license.
     <PrivateSdkAssemblies Include="$(IlcPath)\sdk\*.dll" />
   </ItemGroup>
   <ItemGroup>
-    <!-- Exclude System.IO.Compression.Native.dll for now https://github.com/dotnet/corert/issues/5496 -->
-    <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\System.IO.Compression.Native.dll" />
+    <!-- Exclude System.IO.Compression.Native.dll -->
+    <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\*.Native.dll" />
   </ItemGroup>
 
   <ItemGroup>
@@ -139,8 +139,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <PrivateSdkAssemblies Include="$(IlcPath)\sdk\*.dll" />
     </ItemGroup>
     <ItemGroup>
-      <!-- Exclude System.IO.Compression.Native.dll for now https://github.com/dotnet/corert/issues/5496 -->
-      <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\System.IO.Compression.Native.dll" />
+      <!-- Exclude System.IO.Compression.Native.dll -->
+      <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\*.Native.dll" />
     </ItemGroup>
 
     <ItemGroup>


jkotas:
I think it would be ok to not have support for the fine-grained msbuild properties for now, and instead focus on enabling the app global linker xml file(s). The per-app xml file is equivalent and more powerful.

Supporting the fine-grained msbuild properties will likely need msbuild task. We have msbuild task to figure out the framework, so maybe it can be added to that. Just a thought - can the msbuild task generate the xml file so that we do not need to invent command line options for all of this?

MichalStrehovsky:
> The per-app xml file is equivalent and more powerful.

The XML unfortunately doesn't have means to say "if this assembly is used, root the entire assembly", which is how the framework is rooted in the IL Linking defaults in the SDK. It's also what we considered "RootAllApplicationAssemblies" in NativeAOT (we only rooted the used ones, which is different from the default in the SDK that roots them unconditionally).

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -147,10 +132,15 @@ The .NET Foundation licenses this file to you under the MIT license.
...
     <!-- If running from a package these values need to be set again with the resolved IlcPath -->
     <ItemGroup>
-      <PrivateSdkAssemblies Include="$(IlcPath)\sdk\*.dll" />
+      <PrivateSdkAssemblies Include="$(IlcPath)\sdk\*.dll">
+        <IsTrimmable>true</IsTrimmable>
+      </PrivateSdkAssemblies>
     </ItemGroup>
     <ItemGroup>
-      <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" />
+      <!-- Exclude System.IO.Compression.Native.dll for now https://github.com/dotnet/corert/issues/5496 -->
```
MichalStrehovsky, line 140: We have the `FrameworkAssemblies` definition duplicated for reasons I don't remember anymore. This occurrence didn't have the workaround and it kills the compiler with `TrimMode=copy`

MichalStrehovsky (commented):
Well, it seems to work. The ILLinker's target that we're pulling in is mostly harmless. It brings in the suppressions which is nice.

The side effect of this:

* Hello World is now 33 MB because we match .NET 5 defaults (keep entire assemblies if used). Adding `<TrimMode>link</TrimMode>` to the csproj fixes that. (We could make that our default.)
* I assume it will no longer work withe .NET Core 3.1 SDK. (We could detect+reject that).

@jkotas what do you think?

jkotas:
> Hello World is now 33 MB because we match .NET 5 defaults

Can we change the defaults when you reference the nativeaot package (similar to how we change the defaults for the binary serialization)?

jkotas:
> I assume it will no longer work withe .NET Core 3.1 SDK. (We could detect+reject that).

We should update https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/samples/prerequisites.md . Also, check with the virtual team whether .NET 5 SDK depependency is ok.

MichalStrehovsky:
>Can we change the defaults when you reference the nativeaot package (similar to how we change the defaults for the binary serialization)?

So I made this do TrimMode=link on framework assemblies. The rest respects whatever TrimMode the user specified. I made our default `copyused` if not specified. The SDK would choose `copy`.

> We should update https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/samples/prerequisites.md . Also, check with the virtual team whether .NET 5 SDK depependency is ok.

I checked with the teams we work with directly. One is in the process of updating to .NET 5 (it was supposed to happen yesterday Seattle time). The other team is already on .NET 5.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -28,13 +28,12 @@ The .NET Foundation licenses this file to you under the MIT license.
 
   <!-- Set up the defaults for the compatibility mode -->
   <PropertyGroup>
-    <_BuildingInCompatibleMode Condition="$(RootAllApplicationAssemblies) == '' and $(IlcGenerateCompleteTypeMetadata) == '' and $(IlcGenerateStackTraceData) == '' and $(IlcDisableReflection) == ''">true</_BuildingInCompatibleMode>
+    <_BuildingInCompatibleMode Condition="$(TrimMode) == '' and $(IlcGenerateStackTraceData) == '' and $(IlcDisableReflection) == ''">true</_BuildingInCompatibleMode>
 
-    <RootAllApplicationAssemblies Condition="$(RootAllApplicationAssemblies) == ''">true</RootAllApplicationAssemblies>
-    <IlcGenerateCompleteTypeMetadata Condition="$(IlcGenerateCompleteTypeMetadata) == ''">true</IlcGenerateCompleteTypeMetadata>
```
MichalStrehovsky, line 34: This one is orthogonal, but illink doesn't have such option to begin with. The option still valuable to troubleshoot trimming issues, but it shouldn't be on by default. I added a blurb to the docs.

docs/using-nativeaot/optimizing.md
```diff
@@ -16,14 +16,24 @@ under the `<Project>` node of your project file.
...
+The Native AOT compiler supports the [documented options](https://docs.microsoft.com/en-us/dotnet/core/deploying/trim-self-contained) for removing unused code (trimming). By default, the compiler tries to very conservatively remove some of the unused code.
+
+🛈 Native AOT difference: The documented `PublishTrimmed` property is implied to be `true` when Native AOT is active.
+
+By default, the compiler tries to maximize compatibility with existing .NET code at the expense of compilation speed and size of the output executable. This allows people to use their existing code that worked well in a fully dynamic mode without hitting issues caused by trimming. To read more about reflection, see the [Reflection in AOT mode](reflection-in-aot-mode.md) document.
+
+🛈 Native AOT difference: the `TrimMode` of framework assemblies is set to `link` by default. To compile entire framework assemblies, use `TrimmerRootAssembly` to root the select assemblies. It's not recommended to root the entire framework.
```
jkotas, line 29:
```suggestion
🛈 Native AOT difference: the `TrimMode` of framework assemblies is set to `link` by default. To compile entire framework assemblies, use `TrimmerRootAssembly` to root the selected assemblies. It's not recommended to root the entire framework.
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -150,7 +139,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <PrivateSdkAssemblies Include="$(IlcPath)\sdk\*.dll" />
     </ItemGroup>
     <ItemGroup>
-      <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" />
+      <!-- Exclude System.IO.Compression.Native.dll for now https://github.com/dotnet/corert/issues/5496 -->
```
jkotas, line 142:
```suggestion
      <!-- Exclude System.IO.Compression.Native.dll -->
```
I do not think that the link to that issue is relevant. That issue is fixed, just not marked as such because of the corert repo is archived.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -150,7 +139,8 @@ The .NET Foundation licenses this file to you under the MIT license.
...
+      <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\System.IO.Compression.Native.dll" />
```
jkotas, line 143:
```suggestion
      <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\*.Native.dll" />
```
Future proofing

MichalStrehovsky: This was a copypaste from a different spot in this file. I fixed both.

PR #551, opened 2021-01-12
NativeAOT-LLVM: set GeneratePackage based on package prefix

yowl:
This PR reenables the generation of the nuget packages by matching on the compiler package prefix.  If you think this is useful for other experiments I can push it to NativeAOT ?


jkotas (approved):
Thank you!

jkotas:
> If you think this is useful for other experiments I can push it to NativeAOT ?

I think it would be useful.

PR #553, opened 2021-01-13
Copy ObjWriter from CoreRT repo

AntonLapounov:
Unmodified copy of ObjWriter files from CoreRT repo, which is not yet connected to the build script.


jkotas (approved):
Thanks!

PR #555, opened 2021-01-13
Annotate System.Private.Reflection.Core

MichalStrehovsky:
I've analyzed all of nativeaot SDK libraries for trimming. Except for things that are already known in dotnet/runtime, we don't have warnings anymore after this.

(For future reference - to analyze trim safety of a library, one just need to compile the library in the `--multifile` mode and enable `--scanreflection`. This makes sure all code in the library is looked at.)

Contributes to (or fixes?) #396.


jkotas (approved):
Thank you!

MichalStrehovsky:
Timed out in BasicThreading on Linux/release.

PR #552, opened 2021-01-12
Fix reflection-enabling properties on arrays

MichalStrehovsky:
Arrays are not `EcmaType`.

Fixes #550.

PR #559, opened 2021-01-13
Fix GetEventsOntypeHierarchy

MichalStrehovsky:
Same problem as #552 in properties. Found by accident, but should have done the due dilligence in #552.


jkotas (approved):

MichalStrehovsky:
Timeouts in BasicThreading and Interfaces tests on Linux.

From the logs it appears we timeout during compilation so maybe the machine just gets too busy. I'll try to come up with some extra logging or we can just bump up the 15 minute timeout.

PR #560, opened 2021-01-13
Fix Windows ARM64 build

AntonLapounov:


src/coreclr/nativeaot/Runtime/windows/CoffNativeCodeManager.cpp
```diff
@@ -554,11 +559,41 @@ bool CoffNativeCodeManager::UnwindStackFrame(MethodInfo *    pMethodInfo,
     pRegisterSet->pIP = PTR_PCODE(pRegisterSet->SP - sizeof(TADDR));
 
     memcpy(pRegisterSet->Xmm, &context.Xmm6, sizeof(pRegisterSet->Xmm));
-#endif // TARGET_X86
+#elif defined(TARGET_ARM64)
+    for (int i = 8; i < 16; i++)
+    {
+        context.V[i].Low = pRegisterSet->D[i - 8];
+        context.V[i].High = 0;
```
AntonLapounov, line 566: Do we need to copy NEON registers here? We do copy XMM6-XMM15 for x64.

src/coreclr/nativeaot/Runtime/windows/CoffNativeCodeManager.cpp
```diff
@@ -554,11 +559,41 @@ bool CoffNativeCodeManager::UnwindStackFrame(MethodInfo *    pMethodInfo,
...
+                    (PRUNTIME_FUNCTION)pNativeMethodInfo->runtimeFunction,
+                    &context,
+                    &HandlerData,
+                    &EstablisherFrame,
+                    &contextPointers);
+
+    pRegisterSet->SP = context.Sp;
+    pRegisterSet->IP = context.Pc;
+
+    pRegisterSet->pIP = PTR_PCODE(pRegisterSet->SP - sizeof(TADDR));
```
jkotas, line 587: How do we know that `pIP` is `SP - sizeof(TADDR)`? I do not think that this is a valid assumption on ARM64.

src/coreclr/nativeaot/Runtime/windows/CoffNativeCodeManager.cpp
jkotas: We only need to track non-volatile registers (https://docs.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions?view=msvc-160) here.

src/coreclr/nativeaot/Runtime/windows/CoffNativeCodeManager.cpp
```diff
@@ -554,11 +559,41 @@ bool CoffNativeCodeManager::UnwindStackFrame(MethodInfo *    pMethodInfo,
     pRegisterSet->pIP = PTR_PCODE(pRegisterSet->SP - sizeof(TADDR));
 
     memcpy(pRegisterSet->Xmm, &context.Xmm6, sizeof(pRegisterSet->Xmm));
-#endif // TARGET_X86
+#elif defined(TARGET_ARM64)
+    for (int i = 8; i < 16; i++)
+    {
+        context.V[i].Low = pRegisterSet->D[i - 8];
+        context.V[i].High = 0;
+    }
```
jkotas, line 567:
```suggestion
    for (int i = 8; i < 16; i++)
        context.V[i].Low = pRegisterSet->D[i - 8];
```

jkotas:
LGTM otherwise.

src/coreclr/nativeaot/Runtime/windows/CoffNativeCodeManager.cpp
AntonLapounov: Good catch, I guess it should be `pRegisterSet->pIP = pRegisterSet->pLR`.

src/coreclr/nativeaot/Runtime/windows/CoffNativeCodeManager.cpp
AntonLapounov: I meant that if we use this function for GC stack walking only and we never save addresses to `D` registers, we could probably skip even non-volatile `D8`-`D15` ones.

jkotas: This is also used for exception handling where we need to restore the non-volatile registers.

jkotas (approved):
Thanks!

PR #565, opened 2021-01-15
Fix System.IO.Compression build race condition on Windows

jkotas:
The export lib and the static lib are overwriting each other


jkotas:
https://gitter.im/dotnet/corert?at=6001706f252c0a6ded10531f

MichalStrehovsky (approved):

PR #566, opened 2021-01-16
NativeAOT: Match project prefix to allow experiment suffixes

yowl:
This PR is the same as https://github.com/dotnet/runtimelab/pull/551 and allows experiments to have different project name suffixes and still generate packages.


jkotas (approved):
Thanks!

PR #577, opened 2021-01-22
Delete RuntimeTypeHandle.GetValueInternal

MichalStrehovsky:
GetValueInternal may or may not go away from RyuJIT at some point.

We originally used this to implement EETypePtrOf, but the IL-based implementation had problems with generic inlining so we replaced it in https://github.com/dotnet/corert/pull/3814.

The fallback IL was probably still used for CppCodegen at that time so it wasn't deleted.


jkotas (approved):
LGTM. Thank you!

PR #583, opened 2021-01-23
NativeAOT-LLVM: Merge nativeaot

yowl:
Merge NativeAOT and address https://github.com/dotnet/runtimelab/pull/577 

Draft as not started the `RuntimeTypeHandle.GetValueInternal` changes and something has broken the build as getting

```
C:\Users\ScottWaye\.nuget\packages\microsoft.dotnet.apicompat\6.0.0-beta.21062.10\build\Microsoft.DotNet.ApiCompat.targets(82,5): error : MembersMustExist : Member 'public void System.Threading.Thread.UnsafeStart()' does not exist in the implementation but it does exist in the contract. [E:\GitHub\runtimelab\src\libraries\System.Threading.Thread\src\System.Threading.Thread.csproj]
```

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -2566,23 +2566,20 @@ namespace Internal.IL
                     }
                     break;
                 case "EETypePtrOf":
-                    if (_mangledName == "S_P_CoreLib_System_Reflection_RuntimeAssemblyName__GetHashCode")
-                    {
-
-                    }
                     if (metadataType.Namespace == "System" && metadataType.Name == "EETypePtr" && method.Instantiation.Length == 1)
                     {
+                        LLVMValueRef eeTypePtrRef;
                         if (runtimeDeterminedMethod.IsRuntimeDeterminedExactMethod)
                         {
-                            var genericTypeSymbol = CallGenericHelper(ReadyToRunHelperId.TypeHandle, runtimeDeterminedMethod.Instantiation[0]);
-                            PushExpression(StackValueKind.Int32, "eeTypePtr", genericTypeSymbol, GetWellKnownType(WellKnownType.IntPtr));
+                            eeTypePtrRef = CallGenericHelper(ReadyToRunHelperId.TypeHandle, runtimeDeterminedMethod.Instantiation[0]);
                         }
                         else
                         {
-                            var genericTypeSymbol = _compilation.NodeFactory.ConstructedTypeSymbol(method.Instantiation[0]);
-
-                            PushExpression(StackValueKind.Int32, "eeTypePtr", GetEETypePointerForTypeDesc(genericTypeSymbol.Type, true), GetWellKnownType(WellKnownType.IntPtr));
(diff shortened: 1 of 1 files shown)


yowl:
@joktas can you guide me a bit on this, please?  As you can see it fails on the ApiCompat task.
```
C:\Users\ScottWaye\.nuget\packages\microsoft.dotnet.apicompat\6.0.0-beta.21062.10\build\Microsoft.DotNet.ApiCompat.targets(82,5): 
error : MembersMustExist : Member 'public void System.Threading.Thread.UnsafeStart()' does not exist in the implementation but it does exist in the contract. 
[E:\GitHub\runtimelab\src\libraries\System.Threading.Thread\src\System.Threading.Thread.csproj]
```
  As I understand this makes sure that the built libraries conform with the API surface.  It fails on `System.Threading.Thread` which is the one where the new `UnsafeStart` method is added.  The `ValidateApiCompatForSrc ` task runs  with the `rsp` input
```
"E:\GitHub\runtimelab\artifacts\bin\System.Threading.Thread\ref\net6.0-Release\System.Threading.Thread.dll" 
--contract-depends "E:\GitHub\runtimelab\artifacts\bin\System.Threading.Thread\ref\net6.0-Release\,E:\GitHub\runtimelab\artifacts\bin\microsoft.netcore.app.ref\ref\net6.0\,E:\GitHub\runtimelab\artifacts\bin\coreclr\Browser.wasm.Debug\aotsdk\," 
--exclude-attributes "E:\GitHub\runtimelab\eng\DefaultGenApiDocIds.txt,E:\GitHub\runtimelab\eng\ApiCompatExcludeAttributes.txt" 
--enforce-optional-rules 
--impl-dirs "E:\GitHub\runtimelab\artifacts\obj\System.Threading.Thread\net6.0-Release\,E:\GitHub\runtimelab\artifacts\bin\coreclr\Browser.wasm.Debug\aotsdk\,E:\GitHub\runtimelab\artifacts\bin\microsoft.netcore.app.ref\ref\net6.0\,"
```
In the dll in the `ref` folder, this method is present:
image

So that's the contract generated from the source https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/src/libraries/System.Threading.Thread/ref/System.Threading.Thread.cs I think.  The implementation type forwards all classes from a generated file `artifacts\obj\System.Threading.Thread\net6.0-Release\\System.Threading.Thread.Forwards.cs`  e.g.
```
[assembly: System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Threading.Thread))]
```
And the csc build includes `reference:E:\GitHub\runtimelab\artifacts\bin\coreclr\Browser.wasm.Debug\aotsdk\System.Private.CoreLib.dll`  so I read that as it pulling the `Thread` class from `System.Private.CoreLib.dll`  However I've got something wrong as if I look at `System.Private.CoreLib.dll` I see 
image

Where is my logic wrong?

Thanks

jkotas:
You need to change `UnsafeStart()` in src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Thread.CoreRT.Wasm.cs to `public`

yowl:
Doh

yowl:
Looks like I've squashed the merge commits again, I'll redo that.

PR #584, opened 2021-01-24
Make System.Text.Json work in NativeAOT

MichalStrehovsky:
Use the non-Emit codepath if `IsDynamicCodeSupported` is false.


jkotas (approved):
Thanks!

Issue #578, opened 2021-01-22
CodeGenerationFailedException using Protobuf.Net


proficien1t444:
Hi!
1) Install protobuf.net
2) Create a simple data class and serialize it 
3) Try to compile it using native aot

` Unhandled exception. ILCompiler.CodeGenerationFailedException: Code generation failed for method 'uint8[]..ctor(int32)'
   ---> System.InvalidOperationException: Operation is not valid due to the current state of the object.
     at Internal.IL.Stubs.ArrayMethodILEmitter.EmitIL()
     at Internal.IL.Stubs.ArrayMethodILEmitter.EmitIL(ArrayMethod arrayMethod)
     at Internal.IL.CoreRTILProvider.GetMethodIL(MethodDesc method)
     at ILCompiler.RemovingILProvider.GetMethodIL(MethodDesc method)
     at ILCompiler.FeatureSwitchManager.GetMethodIL(MethodDesc method)
     at ILCompiler.Compilation.CombinedILProvider.GetMethodIL(MethodDesc method)
     at ILCompiler.Compilation.ILCache.CreateValueFromKey(MethodDesc key)
     at Internal.TypeSystem.LockFreeReaderHashtable'2.CreateValueAndEnsureValueIsInTable(TKey key)
     at ILCompiler.Compilation.GetMethodIL(MethodDesc method)
     at Internal.IL.ILImporter..ctor(ILScanner compilation, MethodDesc method, MethodIL methodIL)
     at ILCompiler.ILScanner.CompileSingleMethod(ScannedMethodNode methodCodeNodeNeedingCode)
     --- End of inner exception stack trace ---
     at ILCompiler.ILScanner.CompileSingleMethod(ScannedMethodNode methodCodeNodeNeedingCode)
     at ILCompiler.ILScanner.<CompileMultiThreaded>b__5_0(Object m)
     at System.Threading.QueueUserWorkItemCallbackDefaultContext.Execute()
     at System.Threading.ThreadPoolWorkQueue.Dispatch()
     at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback()`

jkotas:
@MichalStrehovsky Does this ctor need to be handled in the ILScanner?

MichalStrehovsky:
> @MichalStrehovsky Does this ctor need to be handled in the ILScanner?

Yeah, looks like it. I'll take a look.

I don't know what Protobuf is doing there, but I don't really want to know.

[...]

PR #586, opened 2021-01-25
Fix handling of SzArray constructors

MichalStrehovsky:
* The change in the scanner fixes the immediate problem reported in #578. We should just the the multi-dim array path for all array constructors.
* I also fixed the jagged array variant while I was at it (the comment describes what the problem was).

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
@@ -1266,14 +1266,14 @@ namespace Internal.JitInterface
                 pResult->kind = CORINFO_CALL_KIND.CORINFO_CALL;
 
                 TypeDesc owningType = targetMethod.OwningType;
-                if (targetMethod.IsConstructor && owningType.IsString)
+                if (owningType.IsString && targetMethod.IsConstructor)
                 {
                     // Calling a string constructor doesn't call the actual constructor.
                     pResult->codePointerOrStubLookup.constLookup = CreateConstLookupToSymbol(
                         _compilation.NodeFactory.StringAllocator(targetMethod)
                         );
                 }
-                else if (targetMethod.IsConstructor && owningType.IsArray)
+                else if (owningType.IsArray && targetMethod.IsConstructor)
                 {
                     // Constructors on arrays are special and don't actually have entrypoints.
                     // That would be fine by itself and wouldn't need special casing. But


src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
```diff
@@ -1263,32 +1263,45 @@ private void getCallInfo(ref CORINFO_RESOLVED_TOKEN pResolvedToken, CORINFO_RESO
...
                 }
 
-                MethodDesc concreteMethod = targetMethod;
-                targetMethod = targetMethod.GetCanonMethodTarget(CanonicalFormKind.Specific);
-
                 pResult->kind = CORINFO_CALL_KIND.CORINFO_CALL;
 
-                if (targetMethod.IsConstructor && targetMethod.OwningType.IsString)
+                TypeDesc owningType = targetMethod.OwningType;
+                if (targetMethod.IsConstructor && owningType.IsString)
```
jkotas, line 1269: IsConstructor is more expensive than IsString. It would be nice to reorder the checks.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
```diff
@@ -1263,32 +1263,45 @@ private void getCallInfo(ref CORINFO_RESOLVED_TOKEN pResolvedToken, CORINFO_RESO
...
                 {
                     // Calling a string constructor doesn't call the actual constructor.
                     pResult->codePointerOrStubLookup.constLookup = CreateConstLookupToSymbol(
                         _compilation.NodeFactory.StringAllocator(targetMethod)
                         );
                 }
+                else if (targetMethod.IsConstructor && owningType.IsArray)
```
jkotas, line 1276: Dtto

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
MichalStrehovsky:
```suggestion
                if (owningType.IsString && targetMethod.IsConstructor)
```

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
MichalStrehovsky:
```suggestion
                else if (owningType.IsArray && targetMethod.IsConstructor)
```

PR #587, opened 2021-01-25
Fix build break in crossgen2

MichalStrehovsky:
Seems like I made these files build in crossgen2 when I originally added them. They got hooked up properly in the runtime repo and now we got duplicates.


jkotas (approved):
Thanks!

Issue #218, opened 2020-10-13
Move ObjWriter from CoreRT repo


jkotas:
Move https://github.com/dotnet/corert/tree/master/src/Native/ObjWriter over and hook it to the build

AntonLapounov:
Fixed by #591.

PR #591, opened 2021-01-26
Support cross-architecture compilation in ILC

### LINKED ISSUE #63: Infrastructure cleanup
- [ ] Remove old ObjectWriter feed from NuGet.config

AntonLapounov:
• Build `clr.alljits` subset by default and publish resulting JIT libraries with ILC.
• Choose JIT library in ILC dynamically, like we do in crossgen2.
• In official builds, build `objwriter` library by requesting `native.objwriter` subset and publish it with ILC.
• If `native.objwriter` subset was not requested, extract `objwriter` library from Microsoft.DotNet.ILCompiler package.
• Remove extra backslashes in paths.

Fixes #63 and #218.

### DIFF
--- a/src/coreclr/tools/aot/ObjWriter/build.cmd
+++ b/src/coreclr/tools/aot/ObjWriter/build.cmd
@@ -13,7 +13,7 @@ git apply "%~dp0llvm.patch" || exit /b 1
 
 :: Copy ObjWriter files
 robocopy /mir "%~dp0\" tools\ObjWriter
-if %ErrorLevel% geq 8 exit /b %ErrorLevel%
+if %ErrorLevel% geq 8 exit /b 1
 
 :: Configure and build
 if not exist build mkdir build


eng/pipelines/runtimelab.yml
```diff
@@ -103,7 +103,7 @@ stages:
         isOfficialBuild: ${{ variables.isOfficialBuild }}
         timeoutInMinutes: 90
         testGroup: innerloop
-        buildArgs: -s nativeaot+libs+nativeaot.packages -c release
+        buildArgs: -s nativeaot.objwriter+nativeaot+libs+nativeaot.packages -c release
```
jkotas, line 106: We should watch next couple official builds and verify that they succeed.

src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
```diff
@@ -117,4 +125,21 @@
...
     <Compile Remove="repro\*" />
   </ItemGroup>
+
+  <PropertyGroup>
+    <TargetOSComponent>unix</TargetOSComponent>
+    <TargetOSComponent Condition="'$(TargetOS)' == 'windows'">win</TargetOSComponent>
+    <TargetSpec>$(TargetOSComponent)_$(TargetArchitecture)_$(TargetArchitecture)</TargetSpec>
+  </PropertyGroup>
+
+  <!-- On Windows we can re-use PGO-optimized clrjit.dll produced by the build for AOT compilation.
```
jkotas, line 135: I am not sure whether it is worth the trouble that it may cause (e.g. with symbols), but since you have done it we can give it a try.

jkotas (approved):
LGTM. Thank you!

src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
AntonLapounov: This logic is copied from crossgen2.csproj.

PR #592, opened 2021-01-26
Fix 'tools' casing for src/coreclr/Tools/aot/ObjWriter

AntonLapounov:
PR #556 did not use the correct casing.


jkotas (approved):
Thank you!

PR #593, opened 2021-01-26
Make XmlSerializer work with NativeAOT

MichalStrehovsky:
Force the reflection-only code paths that don't Reflection.Emit.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/FeatureSwitchManager.cs
@@ -876,7 +876,7 @@ namespace ILCompiler
 
             private object TryCreateSubstitution(TypeDesc type, string value)
             {
-                switch (type.Category)
+                switch (type.UnderlyingType.Category)
                 {
                     case TypeFlags.Int32:
                         if (string.IsNullOrEmpty(value))


src/libraries/System.Private.Xml/src/System/Xml/Serialization/XmlSerializer.cs
```diff
@@ -199,6 +199,12 @@ public XmlSerializer(Type type, string? defaultNamespace)
                 _primitiveType = type;
                 return;
             }
+
+            if (Mode == SerializationMode.ReflectionOnly)
+            {
+                return;
```
jkotas, line 205: Do we need to also execute `GetTopLevelMapping` at the very bottom of the method in ReflectionOnly mode?

src/libraries/System.Private.Xml/src/System/Xml/Serialization/XmlSerializer.cs
```diff
@@ -199,6 +199,12 @@ public XmlSerializer(Type type, string? defaultNamespace)
...
+            }
+
             _tempAssembly = s_cache[defaultNamespace, type];
             if (_tempAssembly == null)
```
jkotas, line 209: Do we need to fix ` XmlSerializer(XmlTypeMapping xmlTypeMapping)` constructor in the same way?

src/libraries/System.Private.Xml/src/System/Xml/Serialization/XmlSerializer.cs
MichalStrehovsky: No, I think this was all ifdeffed out:

https://github.com/dotnet/corefx/blob/e31a2b7f3bdf985f56798df530daa4c06fd80376/src/System.Private.Xml/src/System/Xml/Serialization/XmlSerializer.cs#L266-L308

src/libraries/System.Private.Xml/src/System/Xml/Serialization/XmlSerializer.cs
MichalStrehovsky: Welp, I haven't done a full audit of what were all of the deleted ifdefs doing. This is more of an emergency fix to see what else the partner team is going to hit. We could file an issue to do a full audit.

jkotas (approved):

src/libraries/System.Private.Xml/src/System/Xml/Serialization/XmlSerializer.cs
MichalStrehovsky: https://github.com/dotnet/runtimelab/issues/596

PR #594, opened 2021-01-26
Replace old ReflectionMethodBodyScanner with the illink-based one

MichalStrehovsky:
This required beefing up the illink-based one a little:

* Attempt to instantiate generic things. This matches the behavior of the old scanner that was able to make `typeof(SomeType<>).MakeGenericType(otherType).GetMethod("Foo")` work for reference-type `otherType`. This is a heuristic. `MakeGenericType` just needs to warn as being unsafe.
* Recognize `Enum.GetValues` and `Marshal.SizeOf`. As a bonus, this hooks into the flow analysis infra, so we can potentially figure out more use cases. As another bonus, there's now a warning.
* Detect when Type.GetType resolved things through a forwarder and add the forwarder metadata.

Lost features:
* We no longer do the optimization for `typeof(T) == typeof(Foo)` that was trying to avoid seeing `Foo` as constructed (with a fully populated vtable). This can be added later; I'm leaving the compiler infra inplace.
* Ability to run the scanner to only scan for interop pattern.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
@@ -911,6 +911,11 @@ namespace ILCompiler.Dataflow
                     //
                     case IntrinsicId.Enum_GetValues:
                         {
+                            // Enum.GetValues returns System.Array, but it's the array of the enum type under the hood
+                            // and people depend on this undocumented detail (could have returned enum of the underlying
+                            // type instead).
+                            //
+                            // At least until we have shared enum code, this needs extra handling to get it right.
                             foreach (var value in methodParams[0].UniqueValues())
                             {
                                 if (value is SystemTypeValue systemTypeValue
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UsageBasedMetadataManager.cs
```diff
@@ -492,7 +481,7 @@ public override void GetDependenciesDueToAccess(ref DependencyList dependencies,
         public override void GetDependenciesDueToAccess(ref DependencyList dependencies, NodeFactory factory, MethodIL methodIL, MethodDesc calledMethod)
         {
             bool scanReflection = (_generationOptions & UsageBasedMetadataGenerationOptions.ReflectionILScanning) != 0;
-            if (scanReflection && FlowAnnotations.RequiresDataflowAnalysis(calledMethod))
+            if (scanReflection && Dataflow.ReflectionMethodBodyScanner.RequiresReflectionMethodBodyScannerForCallSite(FlowAnnotations, calledMethod))
```
MichalStrehovsky, line 484: This was an ooops. We were running the analysis less often than we should (we need to run it not just when calling into an annotated method, but also if the method is an intrinsic. Most, but not all, intrinsics have annotations.

jkotas (approved):
Thank you!

PR #598, opened 2021-01-27
Update Environment.ProcessorCount - port dotnet/runtime#45943

jkotas:
Also delete some dead code

### DIFF
--- a/src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
+++ b/src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
@@ -39,14 +39,6 @@
 #include <lwp.h>
 #endif
 
-#if HAVE_SYSCONF
-// <unistd.h> already included above
-#elif HAVE_SYSCTL
-#include <sys/sysctl.h>
-#else
-#error Either sysctl or sysconf is required for GetSystemInfo.
-#endif
-
 #if HAVE_SYS_VMPARAM_H
 #include <sys/vmparam.h>
 #endif  // HAVE_SYS_VMPARAM_H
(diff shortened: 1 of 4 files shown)


jkotas:
I have verified that `Environment.ProcessorCount` takes process affinity into account with this change.

src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
```diff
@@ -126,10 +126,6 @@ static const int tccMilliSecondsToMicroSeconds = 1000;
 static const int tccMilliSecondsToNanoSeconds = 1000000;
 static const int tccMicroSecondsToNanoSeconds = 1000;
 
-static uint32_t g_cNumProcs = 0;
```
MichalStrehovsky, line 129: I think we can also delete a couple SYSCONF/SYSCTL related defines and ifdefs at the beginning of this file.

Maybe the can also be deleted from config.h.in and configure.cmake?

src/coreclr/nativeaot/Runtime/windows/PalRedhawkMinWin.cpp
```diff
@@ -35,20 +35,6 @@ uint32_t PalEventWrite(REGHANDLE arg1, const EVENT_DESCRIPTOR * arg2, uint32_t a
 // Index for the fiber local storage of the attached thread pointer
 static uint32_t g_flsIndex = FLS_OUT_OF_INDEXES;
 
-GCSystemInfo g_RhSystemInfo;
-
-bool InitializeSystemInfo()
-{
-    SYSTEM_INFO systemInfo;
```
MichalStrehovsky, line 42: Can we delete the SYSTEM_INFO struct from PalRedhawk.h?

MichalStrehovsky (approved):
Thank you!

PR #600, opened 2021-01-27
Fix remaining spots that Ref.Emit in XmlSerializer

### LINKED ISSUE #596: Audit S.X.S.XmlSerializer for Ref.Emitting code
We should audit places in historic versions of XmlSerializer sources that had FEATURE_SERIALIZATION_UAPAOT defines and friends.

Those might need to be restored so that we don't try to do Ref.Emit crap at runtime.

https://github.com/dotnet/runtimelab/pull/593

I didn't find the place where the defines were deleted, but the UAPAOT branch of corefx will have the "known good" version of the serializer compiling for UAPAOT.

MichalStrehovsky:
These were ifdeffed out for UAPAOT in the past. The rest of the UAPAOT ifdefs seem to be dealing with with the pregenerated serialization code that .NET Native creates and are not relevant for NativeAOT.

Fixes #596.


jkotas (approved):
Thank you!

PR #601, opened 2021-01-27
Add check for .NET 5 SDK, update NativeLibrary sample instructions

### LINKED ISSUE #589: NativeLibrary sample issue
I downloaded the NativeLibrary sample project from Github and tried publishing it without any modifications but failed. The command used and the error  (NETSDK1102) resulting from it are as follows. Visual studio pro version 16.8.4 and Microsoft.DotNet.ILCompiler version 6.0.0-preview.1.21075.1 was used for publishing. Please advice.

C:\Program Files (x86)\Microsoft Visual Studio 12.0\Common7\Tools\Shortcuts>dotnet publish /p:NativeLib=Shared -r win-x64 -c release "C:\NativeLibrary\NativeLibrary.csproj"
Microsoft (R) Build Engine version 16.8.3+39993bd9d for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored C:\NativeLibrary\NativeLibrary.csproj (in 2.15 sec).
  NativeLibrary -> C:\NativeLibrary\bin\release\net5.0\win-x64\NativeLibrary.dll
C:\Program Files\dotnet\sdk\5.0.102\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.ILLink.targets(143,5): error NETSDK1102: Optimizing assemblies for size is not supported for the selected publish configuration. Please ensure that you are publishing a self-contained app. [C:\NativeLibrary\NativeLibrary.csproj]

### LINKED ISSUE #590: [ NativeAOT ] error MSB4057: The target "PrepareForILLink" does not exist in the project
Unable to build a solution anymore.

I've found PR #547 where in one comment was mentioned that "_I assume it will no longer work withe .NET Core 3.1 SDK._"

So there is no way to use CoreRt/NativeAOT with .Net Core 3.1, right? If so 

* shouldn't be the error message more clear ?
* shouldn't be the section [Know issues](https://github.com/dotnet/runtimelab/tree/feature/NativeAOT/samples/HelloWorld#know-issues) (I suppose there is a typo `Know**n**`) updated / removed in such case ?

Otherwise the Q is how to fix it ?

jkotas:
Fixes #590
Fixes #589

### DIFF
--- a/samples/NativeLibrary/README.md
+++ b/samples/NativeLibrary/README.md
@@ -9,13 +9,13 @@ Create a .NET Core class library project using `dotnet new classlib -o NativeLib
 ## Building static libraries
 
 ```bash
-> dotnet publish /p:NativeLib=Static -r <RID> -c <Configuration>
+> dotnet publish /p:NativeLib=Static /p:SelfContained=true -r <RID> -c <Configuration>
 ```
 
 where `<Configuration>` is your project configuration (such as Debug or Release) and `<RID>` is the runtime identifier (one of win-x64, linux-x64, osx-x64). For example, if you want to publish a release configuration of your library for a 64-bit version of Windows the command would look like:
 
 ```bash
-> dotnet publish /p:NativeLib=Static -r win-x64 -c release
+> dotnet publish /p:NativeLib=Static /p:SelfContained=true -r win-x64 -c release
 ```
 
 The above command will drop a static library (Windows `.lib`, OSX/Linux `.a`) in `./bin/[configuration]/netstandard2.0/[RID]/publish/` folder and will have the same name as the folder in which your source file is present.
@@ -23,7 +23,7 @@ The above command will drop a static library (Windows `.lib`, OSX/Linux `.a`) in
 ## Building shared libraries
 
 ```bash
-> dotnet publish /p:NativeLib=Shared -r <RID> -c <Configuration>
+> dotnet publish /p:NativeLib=Shared /p:SelfContained=true -r <RID> -c <Configuration>
 ```
 
 The above command will drop a shared library (Windows `.dll`, OSX `.dylib`, Linux `.so`) in `./bin/[configuration]/netstandard2.0/[RID]/publish/` folder and will have the same name as the folder in which your source file is present.
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
@@ -48,6 +48,7 @@
   <Target Name="_ComputeAssembliesToCompileToNative" DependsOnTargets="$(IlcDynamicBuildPropertyDependencies)">
 
     <!-- Fail with descriptive error message for common mistake. -->
+    <Error Condition="$([MSBuild]::VersionLessThan('$(NETCoreSdkVersion)', '5.0.0'))" Text=".NET SDK 5+ is required for native compilation." />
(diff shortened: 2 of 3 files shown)


jkotas:
Error message with .NET Core SDK 3.1 + netcoreapp3.1:

C:\runtimelab\samples\HelloWorld>dotnet publish -c Release -r win-x64
Microsoft (R) Build Engine version 16.7.2+b60ddb6f4 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored C:\runtimelab\samples\HelloWorld\HelloWorld.csproj (in 6.57 sec).
  HelloWorld -> C:\runtimelab\samples\HelloWorld\bin\Release\netcoreapp3.1\win-x64\HelloWorld.dll
C:\runtimelab\samples\HelloWorld\pkg\microsoft.dotnet.ilcompiler\6.0.0-preview.2.21076.5\build\Microsoft.NETCore.Native.Publish.targets(50,5): error : .NET SDK 5+ is required for native compilation. [C:\runtimelab\samples\HelloWorld\HelloWorld.csproj]

samples/NativeLibrary/README.md
```diff
@@ -9,21 +9,21 @@ Create a .NET Core class library project using `dotnet new classlib -o NativeLib
 ## Building static libraries
 
 ```bash
-> dotnet publish /p:NativeLib=Static -r <RID> -c <Configuration>
+> dotnet publish /p:NativeLib=Static /p:SelfContained=true -r <RID> -c <Configuration>
```
MichalStrehovsky, line 12: I was trying to figure out why this is not a problem when publishing an app as opposed to a library. It seems like it hits some defaults that set it to true. Could we just set this unconditionally in the targets instead? Would it hurt something else?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
```diff
@@ -48,6 +48,7 @@
   <Target Name="_ComputeAssembliesToCompileToNative" DependsOnTargets="$(IlcDynamicBuildPropertyDependencies)">
 
     <!-- Fail with descriptive error message for common mistake. -->
+    <Error Condition="$([MSBuild]::VersionLessThan('$(NETCoreSdkVersion)', '5.0.0'))" Text=".NET SDK 5+ is required for native compilation." />
```
MichalStrehovsky, line 51: Ah. VersionLessThan - nice trick.

MichalStrehovsky (approved):
Thank you!

samples/NativeLibrary/README.md
jkotas: The original problem is the error check here: https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L153

So we are going to suppress this error check by setting `SelfContained` to `true`. The problem with `SelfContained` property is that it controls NuGet restore behavior. We cannot set it for NuGet restore time since our package is not downloaded yet. And when `SelfContained` is not set at restore time and later gets set for build time, you will see errors like: `C:\Program Files\dotnet\sdk\5.0.102\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(387,5): error NETSDK1112: The runtime pack for Microsoft.NETCore.App.Runtime.win-x64 was not downloaded. Try running a NuGet restore with the RuntimeIdentifier 'win-x64'.`. It leads to never ending sequence of hacks.

I think the proper fix for this would be to fix the error check in Microsoft.NET.ILLink.targets that's causing the problem in the first place.

MichalStrehovsky: Thanks for the explanation!

PR #605, opened 2021-01-28
Add detection of AOT unfriendly APIs

MichalStrehovsky:
* Adds an attribute to mark AOT unfriendly APIs. We'll want to make that a public attribute eventually.
* Attributes a couple things in the corelib
* Recognition of the attribute

It works the same as the single-file- and trim- attributes (suppresses warnings within the method annotated as such, warns if an unannotated method calls into such API).

This is not ready for merging yet - I'll need to suppress/propagate the attribute in a couple spots.

Just seeking early feedback.


jkotas:
Looks good to me so far.

MichalStrehovsky:
I ended up undoing the annotation on Array.CreateInstance because it's pretty viral and the API is used in places that were quite a surprise to me (System.DefaultBinder?). I don't know what to make of that yet.

jkotas (approved):

PR #606, opened 2021-01-28
NativeAOT-LLVM: Merge from NativeAOT and make EETypePtr intrinsic

yowl:
This PR replaces https://github.com/dotnet/runtimelab/pull/583 as that fast-forwarded the commits, losing the history.  Makes `EETypePtrOf<T>` an intrinsic as per #577


yowl:
this is a mess :-(

PR #612, opened 2021-01-31
Switch coredistools to the dotnet/runtime plan

### LINKED ISSUE #341: Update CoreDisTools
ILCompiler uses CoreDisTools to generate disassembly. We should switch to the plan from dotnet/runtime#44490.

MichalStrehovsky:
Fixes #341.


jkotas (approved):
Thanks!

PR #613, opened 2021-01-31
Delete support for reading optional fields from C

MichalStrehovsky:
This is a lot of code that we only need because of a couple asserts.


jkotas (approved):
LGTM. Thanks!

PR #616, opened 2021-02-01
Delete CommonRuntimeTypes

### LINKED ISSUE #520: Delete CommonRuntimeTypes
src\coreclr\nativeaot uses a `CommonRuntimeTypes` class to access commonly used `typeof` objects. Caches like this are not great. This one also has a negative size impact since accessing any member will trigger the class constructor that will do all the `typeof` operations to populate the cache.

```csharp
using System;
using System.Diagnostics;
using System.Runtime.CompilerServices;

class CommonRuntimeTypes
{
    public static Type Array { get; } = typeof(Array);
    public static Type Int32 { get; } = typeof(Int32);
    public static Type Double { get; } = typeof(Double);
    public static Type Single { get; } = typeof(Single);
    public static Type Void { get; } = typeof(void);
}

class test
{
    public static Type Benchmark1() => typeof(void);

    public static Type Benchmark2() => CommonRuntimeTypes.Void;

    public static void Main()
    {
        var sw = Stopwatch.StartNew();
        for (int i = 0; i < 1000000; i++)
        {
            Benchmark2();
        }
        Console.WriteLine(sw.ElapsedMilliseconds);
        sw = Stopwatch.StartNew();
        for (int i = 0; i < 1000000; i++)
        {
            Benchmark2();
        }
        Console.WriteLine(sw.ElapsedMilliseconds);
        sw = Stopwatch.StartNew();
        for (int i = 0; i < 1000000; i++)
        {
            Benchmark2();
        }
        Console.WriteLine(sw.ElapsedMilliseconds);

        sw = Stopwatch.StartNew();
        for (int i = 0; i < 1000000; i++)
        {
            Benchmark1();
        }
        Console.WriteLine(sw.ElapsedMilliseconds);
        sw = Stopwatch.StartNew();
        for (int i = 0; i < 1000000; i++)
        {
            Benchmark1();
        }
        Console.WriteLine(sw.ElapsedMilliseconds);
        sw = Stopwatch.StartNew();
        for (int i = 0; i < 1000000; i++)
        {
            Benchmark1();
        }
        Console.WriteLine(sw.ElapsedMilliseconds);
    }
}
```

On CoreCLR, this all prints ~3.

On NativeAOT, this prints ~8 for no static static cache case and ~2 for the static cache (interestingly, accessing statics seems to be faster than CoreCLR).

If we really really care, we can make direct `typeof` even faster, but I don't think the places that use `CommonRuntimeTypes` are _that_ perf sensitive.

kant2002:
Closes #520


jkotas (approved):
Thank you!

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Reflection/Augments/ReflectionAugments.cs
```diff
@@ -82,13 +82,13 @@ public static TypeCode GetRuntimeTypeCode(Type type)
                     break;
             }
 
-            if (type.Equals(CommonRuntimeTypes.String))
+            if (type.Equals(typeof(string)))
```
jkotas, line 85:
```suggestion
            if (type == typeof(string))
```
When we know that `type` is runtime type, it is more efficient to just use `==` instead of type.Equals.

jkotas: Ok, I see that the other places are using `object.ReferenceEquals(type, typeof(string))`. It works too - it may be better to use that for consistency.

Both `==` and `object.ReferenceEquals` produce same code with optimization on in this case.

jkotas:
I think we can merge this and address the additional feedback separately. @kant2002 Are you interested in submitting a PR for that?

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Reflection/Augments/ReflectionAugments.cs
jkotas: Opened #621

kant2002:
Yes. I forgot to fix that yesterday. @jkotas I think about inspect codebase for `==` or `ReferenceEquals` to be really consistent. I like `==` more, but consistency is priority I think. If you have definitive answer then let me know, I can quickly do that if not, it will take some time.

Issue #618, opened 2021-02-01
Async state machines don't need Equals/GetHashCode implementations


MichalStrehovsky:
We currently have code to inject support code for valuetype Equals/GetHashCode (the `__GetFieldHelper` method).

Looks like async state machines are getting those overrides. Maybe they shouldn't because it's just a waste of disk space.

Side effect of the `__GetFieldHelper` override is that we consider all fields on the struct to be boxed/constructed because we're doing LDTOKEN of them - which is also annoying. Should look into that too. We're possibly wasting quite some space.

MichalStrehovsky:
>Side effect of the __GetFieldHelper override is that we consider all fields on the struct to be boxed/constructed because we're doing LDTOKEN of them - which is also annoying. Should look into that too. We're possibly wasting quite some space.

#624 fixes that overhead.

This issue still tracks whether we can avoid generating the override in the first place.

MichalStrehovsky:
Fixed in https://github.com/dotnet/runtime/pull/83369.

PR #620, opened 2021-02-01
Limit impact of `typeof()` in a custom attribute

MichalStrehovsky:
When a type is referenced by a typeof in a custom attribute, don't immediately create a constructed EEType for it.

This is a heuristic - we don't know what the type will be used for. The field/property/constructor parameter that is initialized with the `typeof` should ideally have a dataflow annotation that describes the intent - the dataflow infrastructure will then make sure we have all that's needed.

I hit this while I was working on #619 - the stubbed out method had a `[AsyncStateMachine(typeof(<GetNonFileStreamAsync>d__2))]` attribute on it. The code I'm making conditional was making a constructed type for the state machine. Then #618 created a perfect storm that brought a constructed `HttpClient` object back into the picture and rendered #619 completely ineffective by itself.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/PayForPlayExperience/MissingMetadataExceptionCreator.cs
+++ b/src/coreclr/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/PayForPlayExperience/MissingMetadataExceptionCreator.cs
@@ -224,7 +224,37 @@ namespace Internal.Reflection.Execution.PayForPlayExperience
             }
             else
             {
-                return null;
+                // First, see if Type.Name is available. If Type.Name is available, then we can be reasonably confident that it is safe to call Type.FullName.
+                // We'll still wrap the call in a try-catch as a failsafe.
+                string s = type.InternalNameIfAvailable;
+                if (s == null)
+                    return null;
+
+                try
+                {
+                    s = type.FullName;
+                }
+                catch (MissingMetadataException)
+                {
+                }
+
+                // Insert commas so that CreateConstructedGenericTypeStringIfAvailable can fill the blanks.
+                // This is not strictly correct for types nested under generic types, but at this point we're doing
+                // best effort within reason.
+                if (type.IsGenericTypeDefinition)
+                {
+                    s += "[";
+                    int genericArgCount = type.GetGenericArguments().Length;
+                    while (genericArgCount-- > 0)
+                    {
+                        genericParameterOffsets.Add(s.Length);
+                        if (genericArgCount > 0)
+                            s = s + ",";
+                    }
+                    s += "]";
+                }
+
+                return s;
             }
         }
 
--- a/src/tests/nativeaot/SmokeTests/Reflection/Reflection.cs
+++ b/src/tests/nativeaot/SmokeTests/Reflection/Reflection.cs
@@ -638,6 +638,8 @@ internal class ReflectionTest
 
         struct SecondNeverUsedType { }
 
+        struct ThirdNeverUsedType { }
+
         class Gen<T> { }
 
         class TypeAttribute : Attribute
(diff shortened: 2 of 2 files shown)


jkotas:
> the dataflow infrastructure will then make sure we have all that's needed.

How is the dataflow infrastructure going to do this?

MichalStrehovsky:
```csharp
class FooAttribute : Attribute
{
    [DynamicallAccessedMembers(Constructors)]
    public Type Foo;
}

[Foo(Foo = typeof(OtherClass))]
class SomeClass
{
}

class OtherClass
{
}
```

Dataflow analysis considers this an assignment of OtherClass into Foo. It will make sure we keep all constructors on OtherClass. This also keeps the owning type. Also makes it possible to Activator.CreateInstance whatever we get out of Foo without warnings.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/CustomAttributeBasedDependencyAlgorithm.cs
```diff
@@ -252,11 +252,17 @@ private static bool AddDependenciesFromCustomAttributeArgument(DependencyList de
...
-            // Theoretically, we only need the EEType if this is a constructed type, but let's go for the full
-            // EEType for the sake of consistency.
-            dependencies.Add(factory.MaximallyConstructableType(typeofType), "Custom attribute blob");
+            // Grab the metadata nodes that will be necessary to represent the typeof in the metadata blob
+            TypeMetadataNode.GetMetadataDependencies(ref dependencies, factory, typeofType, "Custom attribute blob");
+
+            if (typeofType.IsArray
+                || (typeofType.HasInstantiation && !typeofType.IsGenericDefinition))
+            {
+                // We go for the entire EEType because the reflection stack at runtime will need an EEType anyway
```
jkotas, line 261: Is there a good place in the reflection stack that shows the dependency on EEType and that can be linked from here?

It looks odd to me that the reflection stack does not need EETypes for regular types, but it does need them for arrays and generics.

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/CustomAttributeBasedDependencyAlgorithm.cs
MichalStrehovsky: IIRC is this one for arrays:

https://github.com/dotnet/runtimelab/blob/2c625048711145c22780dffd8f6a6635aa245c4b/src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/General/TypeUnifier.cs#L278-L289

And this one for generic types:

https://github.com/dotnet/runtimelab/blob/2c625048711145c22780dffd8f6a6635aa245c4b/src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/General/TypeUnifier.cs#L426-L443

(The `// We only permit` part)

If we just go and delete those lines things will probably still work.

The intent of that code was to blame `MakeGenericType`/`MakeArrayType` for reflection failures at runtime instead of having these APIs succeed and getting an exception once we actually want to do something with the type (call static method, activate, etc.).

I think this behavior makes sense for user code because this API is the problem.

It's annoying for framework code because e.g. without the special casing in the compiler we will get an exception when trying to `GetCustomAttributeData` for the attribute.

MichalStrehovsky: On a second thought, maybe the compiler doesn't need to do this special casing here. This situation is the same as what we would get when e.g. querying fields:

```csharp
using System;

ClassWithField.Klass = new ClassWithField() { MyGen = null };

typeof(Gen<>).ToString();

foreach (var f in typeof(ClassWithField).GetFields(System.Reflection.BindingFlags.Instance | System.Reflection.BindingFlags.Public))
{
    Console.WriteLine(f.FieldType.FullName);
}

class ClassWithField
{
    public static ClassWithField Klass;

    public Gen<int> MyGen;
}

class Gen<T> { }
```

This will fail with:

```
Unhandled Exception: System.Reflection.MissingMetadataException: 'Gen<System.Int32>' is missing metadata. For more information, please visit http://go.microsoft.com/fwlink/?LinkID=392859
   at System.Reflection.Runtime.TypeInfos.RuntimeConstructedGenericTypeInfo.ConstructedGenericTypeTable.Factory(RuntimeConstructedGenericTypeInfo.UnificationKey) + 0x24b
   at System.Collections.Concurrent.ConcurrentUnifierWKeyed`2.GetOrAdd(K) + 0x23f
   at System.Reflection.Runtime.TypeInfos.RuntimeConstructedGenericTypeInfo.GetRuntimeConstructedGenericTypeInfo(RuntimeTypeInfo, RuntimeTypeInfo[], RuntimeTypeHandle) + 0x6f
   at System.Reflection.Runtime.TypeInfos.RuntimeConstructedGenericTypeInfo.GetRuntimeConstructedGenericTypeInfo(RuntimeTypeInfo, RuntimeTypeInfo[]) + 0x56
   at System.Reflection.Runtime.General.TypeUnifier.GetConstructedGenericType(RuntimeTypeInfo, RuntimeTypeInfo[]) + 0x2a
   at System.Reflection.Runtime.General.TypeResolver.TryResolveTypeSignature(TypeSpecificationHandle, MetadataReader, TypeContext, Exception&) + 0x945
   at System.Reflection.Runtime.General.TypeResolver.TryResolve(Handle, MetadataReader, TypeContext, Exception&) + 0x160
   at System.Reflection.Runtime.General.TypeResolver.Resolve(Handle, MetadataReader, TypeContext) + 0x61
   at System.Reflection.Runtime.FieldInfos.NativeFormat.NativeFormatRuntimeFieldInfo.get_FieldRuntimeType() + 0x93
   at System.Reflection.Runtime.FieldInfos.RuntimeFieldInfo.get_FieldType() + 0x70
   at <Program>$.<Main>$(String[]) + 0x130
```

The compiler doesn't try to go out of its way to make sure we can access the field type either. Maybe that's fine?

jkotas: I do not have opinion, except that consistency and less special cases is good.

MichalStrehovsky: Okay, let's see if we can get away without it. We can always restore that behavior.

MichalStrehovsky:
Of course I wrote a test for this exact scenario.

Looks like the exception message leaves things to be desired, so I need to fix that first to make this change acceptable.

PR #624, opened 2021-02-02
Limit the overhead of ValueType.Equals/GetHashCode support

MichalStrehovsky:
See https://github.com/dotnet/corert/pull/5436 for overview.

The `__GetFieldHelper` method we inject into a lot of valuetypes doesn't have to specialize to concrete reference types and to concrete enum types of the individual fields. This gives the AOT compiler an opportunity to eliminate EETypes associated with these fields.

The former saves about 24 kB on a moderately-sized app. The latter saves 7 kB.

We also don't need the override on System.Double/Single. That one saves less than 512 bytes, but it's just good hygiene.


jkotas (approved):
Thanks!

PR #628, opened 2021-02-03
Move TypePreinit construction into try/catch

MichalStrehovsky:
The constructor might end up throwing a `TypeSystemException` too and it did in a particularly broken app.


jkotas (approved):
Thanks!

PR #629, opened 2021-02-03
Optimize Thread.CurrentThread.ManagedThreadId

MichalStrehovsky:
We were not reporting an intrinsic which was preventing this optimization from kicking in.


jkotas (approved):
Thanks!

PR #633, opened 2021-02-04
Allow specify custom entry

kant2002:
Use case for this would be custom runtime libraries, like UEFI for example

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
@@ -19,7 +19,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <CppLibCreator>lib</CppLibCreator>
     <FullRuntimeName>Runtime</FullRuntimeName>
     <FullRuntimeName Condition="'$(ServerGarbageCollection)' != ''">Runtime.ServerGC</FullRuntimeName>
-    <ApplicationEntryName Condition="'$(ApplicationEntryName)' == ''">wmainCRTStartup</ApplicationEntryName>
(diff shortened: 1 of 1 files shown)


kant2002:
I actually need more then this. I finally understand that I have two entries in the link RSP file.
```
/SUBSYSTEM:EFI_APPLICATION
...
/SUBSYSTEM:CONSOLE
```

kant2002:
Build failures seems to be unrelated.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
```diff
@@ -19,6 +19,9 @@ The .NET Foundation licenses this file to you under the MIT license.
     <CppLibCreator>lib</CppLibCreator>
     <FullRuntimeName>Runtime</FullRuntimeName>
     <FullRuntimeName Condition="'$(ServerGarbageCollection)' != ''">Runtime.ServerGC</FullRuntimeName>
+    <ApplicationEntryName Condition="'$(ApplicationEntryName)' == ''">wmainCRTStartup</ApplicationEntryName>
```
MichalStrehovsky, line 22: Could you name these `SubSystem` and `EntryPointSymbol`? It's how they're named in Visual C++ projects. In the absence of other reasons, it feels better to stick with naming that already exists.

kant2002: `EntryPointSymbol` make better sense to me. Reason why I prepend `Linker` to `SubSystem` is because I was not sure that this is not enough generic name, that would not be used in other applications.

MichalStrehovsky (approved):

PR #636, opened 2021-02-04
Ifdef EventSource validation that looks at IL bodies

MichalStrehovsky:
This was ifdeffed under ES_BUILD_PN that disappeared.

https://github.com/dotnet/corert/blob/24f1edc938af9954eac9a3385958787ca64956d5/src/System.Private.CoreLib/shared/System/Diagnostics/Tracing/EventSource.cs#L3707


jkotas (approved):
LGTM.

I am wondering whether it would make sense to ifdef out the entire validation.

PR #637, opened 2021-02-04
Build assembler helpers for Windows ARM64

AntonLapounov:
It seems WriteBarriers helpers will not work correctly if `WRITE_BARRIER_CHECK` is enabled. I have not fixed that.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
+++ b/src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
@@ -32,11 +32,11 @@ INVALIDGCVALUE  EQU 0xCCCCCCCD
 
     MACRO
         ;; On entry:
-        ;;  $destReg: location to be updated
-        ;;  $refReg: objectref to be stored
+        ;;  $destReg: location to be updated (cannot be x9,x12)
+        ;;  $refReg: objectref to be stored (cannot be x9,x12)
         ;;
         ;; On exit:
-        ;;  x9,x10: trashed
+        ;;  x9,x12: trashed
         ;;  other registers are preserved
         ;;
         UPDATE_GC_SHADOW $destReg, $refReg
@@ -48,7 +48,7 @@ INVALIDGCVALUE  EQU 0xCCCCCCCD
 
         ;; Save $destReg since we're about to modify it (and we need the original value both within the macro and
         ;; once we exit the macro).
-        mov     x10, $destReg
+        mov     x12, $destReg
 
         ;; Transform $destReg into the equivalent address in the shadow heap.
         adrp    x9, g_lowest_address
@@ -73,7 +73,7 @@ INVALIDGCVALUE  EQU 0xCCCCCCCD
         dmb     ish
 
         ;; Now check that the real heap location still contains the value we just wrote into the shadow heap.
-        mov     x9, x10
+        mov     x9, x12
         ldr     x9, [x9]
         cmp     x9, $refReg
         beq     %ft0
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
```diff
@@ -317,4 +319,74 @@ ExchangeRetry
...
+;;   x10 : trashed if WRITE_BARRIER_CHECK
+;;   x14 : incremented by 8 to implement JIT_ByRefWriteBarrier contract
+;;
+    LEAF_ENTRY RhpCheckedAssignRefArm64, _TEXT
+    ALTERNATE_ENTRY RhpCheckedAssignRefAVLocation
+    ALTERNATE_ENTRY RhpCheckedAssignRefX1AVLocation
+
+        stlr    x15, [x14]
+
+        ;; TODO: This trashes x14
```
jkotas, line 360: Is this TODO relevant? Do want to file issue on it to make sure we do not forget about it?

src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
```diff
@@ -317,4 +319,74 @@ ExchangeRetry
...
+;;
+;; On exit:
+;;   x9  : trashed
+;;   x10 : trashed if WRITE_BARRIER_CHECK
+;;   x13 : incremented by 8
+;;   x14 : incremented by 8
+;;
+    LEAF_ENTRY RhpByRefAssignRefArm64, _TEXT
+        ldr x15, [x13]
+        str x15, [x14]
```
jkotas, line 381: Does this need to be `stlr` ?

jkotas (approved):
Thank you!

src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
AntonLapounov: I am going to fix the two `TODO` issues regarding trashing `x14` in the helpers I am adding. The Linux version avoids this issue by passing the register to trash to `INSERT_CHECKED_WRITE_BARRIER_CORE` macro.

I also placed two `TODO` comments regarding trashing `x10` if `WRITE_BARRIER_CHECK` is enabled in the existing helpers, `RhpCheckedLockCmpXchg` and `RhpCheckedXchg`. On closer inspection, there is no bug because `x10` is 'trashed' with `x0`'s value. I'll add a comment on that.

src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
AntonLapounov: Probably yes, however, I see no `stlr`/`dmb` in Linux ARM64 and LinuxARM32 versions:
https://github.com/dotnet/corert/commit/39c6ceb72d3764e48504ac26ec926e7b648c4752#diff-5f1553cb1c44e19848424918ad7a351a00fce13fd08216e3d866cc41f1e0bf0fR360
https://github.com/dotnet/corert/pull/4235/commits/e08c6b80f650f280414b22843125ccde07b1c29e#diff-99ef768b3a97ff13d34b55ad20908bb94982d34ff51f8b2a91efbd4dc010e6c0R289

jkotas: It may be a bug

AntonLapounov: I will use `stlr` for Windows ARM64.  I filed #639 for Linux ARM32 and ARM64.

PR #640, opened 2021-02-05
Merge from runtime-master and update yml

danmoseley:


danmosemsft:
Per @safern advice, merging as CI won't run with so many files.

PR #643, opened 2021-02-05
Strip "likely unreachable" instance method bodies

MichalStrehovsky:
***Saves almost 5% in size on a `dotnet new webapi` (with `<TrimMode>link</TrimMode>` but no other customizations).***

Instance method bodies that can only be called with a null this (since we never allocated their owning type) are likely unreachable and can be stripped. C# would call them all through callvirt which would nullref at the callsite. Calling a method with a null this leads to general unhappiness anyway.

Example scenario:

```csharp
object foo = ...;
if (foo is Foo f) // this condition is never actually true, but we don't eval that
    f.DoSomethingExpensive();
```

What this does:

* New node type that we give out instead of the real method body when someone is asking for a non-valuetype instance methods entrypoint.
* The new node works similar to `ConstructedEEType`/`NecessaryEEType` - it goes out of the way and acts as the real method body at object emission phase if the real body became part of the dependency graph.
* The fake body just tail calls into a throw helper so that's the experience someone would have if they actually call something with a null this.
* Also had to do something about inlining - since scanner also does this optimization, we had to prevent compilation phase from inlining the unreachable code. This is done by exporting this knowledge out of the scanner.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ILScanNodeFactory.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ILScanNodeFactory.cs
@@ -41,7 +41,7 @@ namespace ILCompiler.DependencyAnalysis
             if (CompilationModuleGroup.ContainsMethodBody(method, false))
             {
                 // We might be able to optimize the method body away if the owning type was never seen as allocated.
-                if (method.NotCallableWithoutOwningEEType())
+                if (method.NotCallableWithoutOwningEEType() && CompilationModuleGroup.AllowInstanceMethodOptimization(method))
                     return new TentativeInstanceMethodNode(new ScannedMethodNode(method));
 
                 return new ScannedMethodNode(method);
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilationModuleGroup.Aot.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilationModuleGroup.Aot.cs
@@ -73,5 +73,10 @@ namespace ILCompiler
         /// accessed through the target platform's import mechanism (ie, Import Address Table on Windows)
         /// </summary>
         public abstract bool CanHaveReferenceThroughImportTable { get; }
+
+        /// <summary>
+        /// If true, instance methods will only be generated once their owning type is created.
+        /// </summary>
+        public abstract bool AllowInstanceMethodOptimization(MethodDesc method);
     }
 }
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/MultiFileCompilationModuleGroup.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/MultiFileCompilationModuleGroup.cs
@@ -106,7 +106,7 @@ namespace ILCompiler
             {
                 return false;
             }
-        } 
+        }
     }
 
     /// <summary>
@@ -133,6 +133,16 @@ namespace ILCompiler
         {
             return (type.HasInstantiation || type.IsArray) && ShouldProduceFullVTable(type) && 
                    type.ConvertToCanonForm(CanonicalFormKind.Specific).IsCanonicalSubtype(CanonicalFormKind.Any);
+        }
+
+        public override bool AllowInstanceMethodOptimization(MethodDesc method)
+        {
+            // Both the instance methods and the owning type are homed in a single compilation group
+            // so if we're able to generate the body, we would also generate the owning type here
+            // and nowhere else.
+            Debug.Assert(ContainsMethodBody(method, unboxingStub: false));
+            TypeDesc owningType = method.OwningType;
+            return owningType.IsDefType && !owningType.HasInstantiation && !method.HasInstantiation;
         }
     }
 }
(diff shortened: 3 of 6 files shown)


src/coreclr/tools/Common/TypeSystem/Interop/IL/PInvokeDelegateWrapper.cs
```diff
@@ -83,6 +83,14 @@ public override MetadataType MetadataBaseType
             }
         }
 
+        public override DefType BaseType
```
MichalStrehovsky, line 86: This was a bug that I found because the methods on this type started to be stripped. Ooops.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ILScanNodeFactory.cs
```diff
@@ -40,6 +40,12 @@ protected override IMethodNode CreateMethodEntrypointNode(MethodDesc method)
 
             if (CompilationModuleGroup.ContainsMethodBody(method, false))
             {
+                // If this is an instance method on a non-valuetype, don't add direct reference to
+                // the method body just yet. We might be able to optimize the method body
+                // away if the owning type was never seen as allocated.
+                if (!method.Signature.IsStatic && !method.OwningType.IsValueType && !method.IsCanonicalMethod(CanonicalFormKind.Any))
```
MichalStrehovsky, line 46: We should be able to enable this on canonical methods too, but it's a couple dozens of lines of code so I scoped it out for now to keep the pull request short.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeInstanceMethodNode.cs
```diff
@@ -0,0 +1,118 @@
...
+        public TentativeInstanceMethodNode(IMethodBodyNode methodNode)
+        {
+            Debug.Assert(!methodNode.Method.Signature.IsStatic);
+            Debug.Assert(!methodNode.Method.OwningType.IsValueType);
+            _methodNode = methodNode;
+        }
+
+        private ISymbolNode GetTarget(NodeFactory factory)
+        {
+            // TODO: Make a new entrypoint?
```
jkotas, line 42: Nit: Fix this TODO?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs
```diff
@@ -437,5 +442,43 @@ public override bool IsEffectivelySealed(MethodDesc method)
...
+                        _constructedTypes.Add(type);
+                    }
+                }
+            }
+
+            public bool CanInline(MethodDesc caller, MethodDesc callee)
+            {
+                if (_baseGroup.CanInline(caller, callee))
+                {
+                    /// Since the scanner doesn't look at instance methods whose owning type
```
jkotas, line 469: This is not a doc block. Use double-slash instead of triple-slash

MichalStrehovsky: I never understood how these doc blocks within a method work - do they do anything.

They enable VS to understand the `<see cref=...>`. Otherwise it's just a text string that is going to start bitrotting the moment it's checked in.

MichalStrehovsky:
> runtimelab (Build Build windows x64 Checked)

Oh cool. So apparently multidimensional arrays are in the category of "type loader can create an EEType for any of these at runtime". I guess SzArrays of pointers are in the same category.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs
jkotas: C# compiler warning "CS1587: XML comment is not placed on a valid language element" suggests that this is not meant to work. We do not have this warning enabled for the compiler, but it is enabled in most of the tree.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/Target_ARM/ARMTentativeInstanceMethodNode.cs
```diff
@@ -0,0 +1,24 @@
...
+using ILCompiler.DependencyAnalysis.ARM;
+
+namespace ILCompiler.DependencyAnalysis
+{
+    public partial class TentativeInstanceMethodNode
+    {
+        protected override void EmitCode(NodeFactory factory, ref ARMEmitter encoder, bool relocsOnly)
+        {
+            ISymbolNode target = GetTarget(factory);
+            if (!target.RepresentsIndirectionCell)
```
jkotas, line 13: Move this into `EmitJMP` so that it does not have to be duplicated?

jkotas (approved):
Nice!

yowl:
I guess wasm is going to need a `WasmTentativeMethodNode`  with something for 
```
        protected override void EmitCode(NodeFactory factory, ref ARM64Emitter encoder, bool relocsOnly)
        {
            encoder.EmitJMP(GetTarget(factory));
        }
```
This is the tail call throw I suppose?

MichalStrehovsky:
> I guess wasm is going to need a WasmTentativeMethodNode with something for

It might be best to treat these the same as the ReadyToRunGenericHelper nodes instead of introducing a new node (they have the same problem and will have the same "cleaner" solution eventually).

Yes, the generated body should just tailcall to whatever is the target. I assume WASM will be offended by the signature mismatch, so this will have to be an actual call and the thunk generated for this will need to have parameters that match the pretended method.

Issue #313, opened 2020-11-07
LLVM as a back-end for RyuJIT


EgorBo:
Since https://github.com/dotnet/corert/issues/8230 thread was archived recently, I'm creating a new one here to continue the discussion.

I'm trying to build a minimal dynamically-loaded extension for RyuJIT to generate LLVM IR (and compile it for tier1 (JIT) or use in AOT scenarios) from GenTrees (HIR/LIR) in order to benefit from RyuJIT's C# specific optimizations and SSA transformation for locals. Something like this:

![image](https://user-images.githubusercontent.com/523221/98451774-fcae0100-2159-11eb-925b-4022eadafc8a.png)
_(my local version is a bit more complicated but it's still just a very high-level prototype and is a mess)_

Some of the challenges (too early to make a deep analysis):
- Setup LLVM dependency (currently, I just use a locally built one)
- Make RyuJIT's headers portable, basically, provide an external API, stabilize layout (e.g. a lot of fields here and there depend on different condition symbols)
- Implement all operators, GC/EH, etc - [How to draw an owl](https://i.kym-cdn.com/photos/images/original/000/572/078/d6d.jpg)

Since there are [LLILC](https://github.com/dotnet/llilc) and https://github.com/dotnet/runtimelab/pull/247 efforts I wonder what can be shared/merged. Maybe it makes sense to bind all RyuJIT's structures/API to C# (e.g. via `CppAst.NET`) and continue IR -> LLVM IR transformation there using LLVMSharp (C# bindings for LLVM C-api). Maybe it's better to join https://github.com/dotnet/runtimelab/pull/247 and just port RyuJIT's optimizations one-by-one there.

Initially, I just wanted to make an extension for RyuJIT to replace tier1 with LLVM-JIT for HPC# code (a GC-free subset of C# in Unity's Burst for high-performance stuff) to generate vectorized loops, etc where RyuJIT currently doesn't perform well. But I see that there is a quite a demand for AOT. So maybe you have some thoughts regarding the direction it is better to move to to make it at least theoretically useful?

/cc @jkotas @MichalStrehovsky @yowl @AndyAyersMS

AndyAyersMS:
Right, EH and GC are still enormous hurdles.

I think you may be able to get pretty far with EH, perhaps only compromising on currently somewhat rare cases -- like filters that have side effects -- that require a true understanding of the order in which things can happen.

[...]

PR #647, opened 2021-02-06
NativeAOT-LLVM: RyuJIT IR -> LLVM for improved performance

yowl:
Some mucking around to try to get something that can be played with.  Copied a lot of clrjit macros and code from xarch/amd64 and deleted/changed to get compilation to succeed and run at least as far as compiling the first method.  This is my attempt to tackle https://github.com/dotnet/runtimelab/issues/313 and may be total garbage.  The clrjit does nothing at the moment exception `assert(false)` at the appropriate phase.  In theory it should be possible to catch this and revert to CIL -> LLVM allowing tests to run at an early stage of playing while adding IR support.


jkotas:
cc @EgorBo

src/coreclr/inc/corinfoinstructionset.h
```diff
@@ -108,7 +108,42 @@ enum CORINFO_InstructionSet
...
 #endif // TARGET_X86
-
+#if defined(TARGET_WASM32) || defined(TARGET_WASM64)
+    //InstructionSet_X86Base = 1,
+    //InstructionSet_SSE = 2,
+    //InstructionSet_SSE2 = 3,
+    //InstructionSet_SSE3 = 4,
+    //InstructionSet_SSSE3 = 5,
+    //InstructionSet_SSE41 = 6,
+    InstructionSet_SSE42 = 7,
```
jkotas, line 118: It does not sound right that you have to define SSE2 and AVX2 for wasm

src/coreclr/jit/emit.h
```diff
@@ -576,6 +576,10 @@ class emitter
 #elif defined(TARGET_ARM64)
```
jkotas, line 576: `CodeGen` and `emitter` are late phases. They should be all replaced by the LLVM part. Is it possible to ifdef them out?

yowl: Do you think I should be able #if out the Lowering class?  https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/coreclr/jit/lower.h

yowl: Have removed those + lsra + lower.  There's a few places that require `#ifdefs` to allow compilation to succeed.  In some places  where I'm not sure what's best I've left the wasm build throwing an `assert`.

src/coreclr/inc/corinfoinstructionset.h
yowl: They were getting used for the SIMD check https://github.com/dotnet/runtimelab/blob/fdc65fccbe815a9b81ef2ddfa6ed27e0e7292210/src/coreclr/jit/compiler.h#L8038.  Wasm can support SIMD, for c++ emscripten handles things.  for LLVM I've not looked what might be necessary, but it would be good if the clrjit optimisation for SIMD that occurs before lowering, if any, was still in I suppose.  For now this method is an `assert` for Wasm and these instructions in `corinfoinstructionset.h` are removed.

src/coreclr/jit/lsra.h
```diff
@@ -735,7 +735,7 @@ class LinearScan : public LinearScanInterface
     // Hence the "SmallFPSet" has 5 elements.
     CLANG_FORMAT_COMMENT_ANCHOR;
 
-#if defined(TARGET_AMD64)
+#if defined(TARGET_AMD64) || defined(TARGET_WASM)
```
jkotas, line 738: There changes should not be needed now that the whole file is ifdefed out

src/coreclr/jit/emit.h
jkotas: I think `inst*` can be remove too.

src/coreclr/jit/stacklevelsetter.cpp
```diff
@@ -19,8 +19,12 @@ StackLevelSetter::StackLevelSetter(Compiler* compiler)
     , throwHelperBlocksUsed(comp->fgUseThrowHelperBlocks() && comp->compUsesThrowHelper)
 #endif // !FEATURE_FIXED_OUT_ARGS
 {
+#ifndef TARGET_WASM
     // The constructor reads this value to skip iterations that could set it if it is already set.
     compiler->codeGen->resetWritePhaseForFramePointerRequired();
+#else
+    assert(false); // Wasm - TODO can this be ignored?
```
jkotas, line 26: It should be possible to exclude the whole StackLevelSetter

src/coreclr/inc/corinfoinstructionset.h
jkotas: I would expect Wasm to have its own wasm-specific names for SIMD levels, and not to reuse the x86/x64 ones. ARM does not reuse the x86/x64 ones either.

yowl:
@jkotas  For the wasm build, I need to build the clrjit for the host platform, initially win64.  The gen-buildsys script at the moment uses `emcmake` to configure the CMake system for emscripten, but I don't want that for clrjit.  One thought I have is to just run the build twice, run build-runtime.cmd twice I suppose, once for wasm and once for the host, possibly disabling as much as possible on the host build so that just the jitinterface and clrjit is build.  The other option which might be possible, it to use `__BuildCrossArchNative` but I'm not sure if this is really the intended use for that?

jkotas:
I am not sure which one of these options will turn out better. I think either one can work fine.

src/coreclr/inc/corinfoinstructionset.h
tannergooding: I agree. https://github.com/WebAssembly/simd isn't finalized yet, but when it is I'd expect we would look at introducing a new `System.Runtime.Intrinsics.Wasm` namespace and corresponding ISAs.

yowl: Ok, for now I've just deleted this change.  https://github.com/dotnet/runtimelab/pull/647/files#diff-5bc0c93aea7b72e9e37d8742e7a59ffb882ff206b746494f2e1f484bc3ef47c1

src/coreclr/jit/lsra.h
yowl: Removed.

src/coreclr/jit/emit.h
yowl: Thanks, removed.

src/coreclr/jit/stacklevelsetter.cpp
yowl: Thanks, yes, removed also.

yowl:
At 
https://github.com/dotnet/runtimelab/blob/fab115791ea49b0d59a0da84f52e0ecd6073cc77/src/coreclr/inc/crosscomp.h#L385
It defines `T_CONTEXT`.  For windows this comes from `winnt.h` in the SDK, but for Wasm what should I do here?  It looks like its mostly registers, so would an empty 
```
typedef struct DECLSPEC_ALIGN(16) DECLSPEC_NOINITALL _CONTEXT {
} CONTEXT, *PCONTEXT;
```
Be somewhere to start?

jkotas:
Yes, a dummy definitions like this sound like a good way to start. You can omit `DECLSPEC_ALIGN(16)`.

This whole header should not be really included by the JIT (there are only a very few things - if any - needed by the JIT from this header), but refactoring to fix that may be too involved.

yowl:
I don't think I'm going to be able to do this incrementally as I don't see a way around the two versions of LLVM.  One is coming from LLVMSharp, i think built with /MD as it has `MSVCP140.dll` as a dependency.  Then there is the .lib linked from clrjit with is built with `/MTd` .

EgorBo:
@yowl may I ask why do you need LLVMSharp in the first place? Isn't simpler to do everything directly in C++ with the LLVM C++ API (a lot of APIs don't exist in the C API and the LLVMSharp, I guess, is basically bindings for the C API) - or you just already have a lot of C# code you want to re-use?

yowl:
I just wanted to not have to do everything in one go.  The ILToLLVMImporter code can generate enough LLVM to compile and run a fair chunk of input IL.  I thought it might be less daunting a task with some more easy, early attainable goals, if the C++ side didn't have to match it, and could start by picking off the easier bits of IL.  So the C++ generated those methods and the ILToLLVMImporter did the rest.  Seems like that won't be as easy as I'd hoped and as you say, it will be easier to just forget ILToLLVMImporter straight away and do it all in C++.  I imagine some bits can still just be ported,

EgorBo:
> and do it all in C++
Still, you can do everything in C# if you want - you only need a sort of CppAST.NET/CppSharp tool to generate C# structs for JIT objects to bind 🙂

yowl:
I guess, just to please my C++ frailties is probably not a good reason though :-)

@jkotas There is a `jitStartup` that is marked as `DLLEXPORT`.  

https://github.com/dotnet/runtimelab/blob/848031fd01a8cde20097729008ab872560f6d13f/src/coreclr/jit/ee_il_dll.cpp#L42

I'd like something equivalent for the end of compilation to write out the LLVM bitcode.  `jitShutdown` is not a DLLEXPORT, 

https://github.com/dotnet/runtimelab/blob/848031fd01a8cde20097729008ab872560f6d13f/src/coreclr/jit/ee_il_dll.cpp#L42

Any suggestions?

jkotas:
A clean way to do it would be to add a new method on the JIT/EE interface, but that will touch a lot of files and it will be a constant source of merge conflicts. I think you can just add a new export for now to do what you need.

yowl:
Thanks, that works fine.  I have a reverse requirement now though I think, to get the mangled / C name for a node, e.g the method name.  I guess I can add some `[UnmanagedCallersOnly]` method somewhere so that the clrjit can ask the `CoreRTNameMangler` for the name?

jkotas:
Yes, that would work as interim solution

yowl:
I've added some loops (copying from @EgorBo ) to process the blocks and statements, but while I have blocks, the statement list is empty and I never get to the `visitNode` (at least for the first method I come to https://github.com/yowl/runtimelab/blob/c26328bb346002867b497e62c69a4cd63ce1e1a0/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/__Finalizer.cs#L20 - doing void(void) functions first.  

https://github.com/yowl/runtimelab/blob/c26328bb346002867b497e62c69a4cd63ce1e1a0/src/coreclr/jit/llvm.cpp#L63-L105

Have I got my hook positioned too early or too late to access the `bbStmtList`?

https://github.com/yowl/runtimelab/blob/c26328bb346002867b497e62c69a4cd63ce1e1a0/src/coreclr/jit/compiler.cpp#L5156

EgorBo:
> I've added some loops (copying from @EgorBo ) to process the blocks and statements, but while I have blocks, the statement list is empty and I never get to the `visitNode` (at least for the first method I come to https://github.com/yowl/runtimelab/blob/c26328bb346002867b497e62c69a4cd63ce1e1a0/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/__Finalizer.cs#L20 - doing void(void) functions first.
> 
> https://github.com/yowl/runtimelab/blob/c26328bb346002867b497e62c69a4cd63ce1e1a0/src/coreclr/jit/llvm.cpp#L63-L105
> 
> Have I got my hook positioned too early or too late to access the `bbStmtList`?
> 
> https://github.com/yowl/runtimelab/blob/c26328bb346002867b497e62c69a4cd63ce1e1a0/src/coreclr/jit/compiler.cpp#L5156

Rationalization is basically a phase where IR is converted into LIR, my demo called LLVM right before the rationalization (thus, I used IR). If you want to do it after (honestly, I've no idea when it's better to do it) see the LIR scheme https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/jit/ryujit-tutorial.md#linear-ir-lir

EgorBo:
@yowl so in your branch it's possible to build the jit and call LLVM stuff in the same lib? I mean is LLVM Context created without problems? If so - it's awesome!

yowl (commented):
> @yowl so in your branch it's possible to build the jit and call LLVM stuff in the same lib? I mean is LLVM Context created without problems? 

That bit at least is working so far.  I had to disable a few warnings about the removal of implicit constructors and destructors to get the LLVM headers to play nice with clrjit which may come back to bite.  Top of the stack here from one of LLVM's libs

```
	clrjit_browser_wasm32_x64.dll!llvm::BasicBlock::Create(llvm::LLVMContext & Context, const llvm::Twine & Name, llvm::Function * Parent, llvm::BasicBlock * InsertBefore) Line 103	C++
 	clrjit_browser_wasm32_x64.dll!Llvm::Compile(Compiler * pCompiler) Line 92	C++
 	clrjit_browser_wasm32_x64.dll!DoLlvmPhase(Compiler * pCompiler) Line 4449	C++
 	clrjit_browser_wasm32_x64.dll!Compiler::compCompile(void * * methodCodePtr, unsigned int * methodCodeSize, JitFlags * compileFlags) Line 5273	C++
 	clrjit_browser_wasm32_x64.dll!Compiler::compCompileHelper(CORINFO_MODULE_STRUCT_ * classPtr, ICorJitInfo * compHnd, CORINFO_METHOD_INFO * methodInfo, void * * methodCodePtr, unsigned int * methodCodeSize, JitFlags * compileFlags) Line 6490	C++
 	clrjit_browser_wasm32_x64.dll!`Compiler::compCompile'::`61'::__Body::Run(Compiler::compCompile::__l2::__JITParam * __JITpParam) Line 5757	C++
 	clrjit_browser_wasm32_x64.dll!Compiler::compCompile(CORINFO_MODULE_STRUCT_ * classPtr, void * * methodCodePtr, unsigned int * methodCodeSize, JitFlags * compileFlags) Line 5761	C++
 	clrjit_browser_wasm32_x64.dll!``jitNativeCode'::`8'::__Body::Run'::`6'::__Body::Run(jitNativeCode::__l8::__Body::Run::__l5::__JITParam * __JITpParam) Line 7137	C++
 	clrjit_browser_wasm32_x64.dll!`jitNativeCode'::`8'::__Body::Run(jitNativeCode::__l2::__JITParam * __JITpParam) Line 7140	C++
 	clrjit_browser_wasm32_x64.dll!jitNativeCode(CORINFO_METHOD_STRUCT_ * methodHnd, CORINFO_MODULE_STRUCT_ * classPtr, ICorJitInfo * compHnd, CORINFO_METHOD_INFO * methodInfo, void * * methodCodePtr, unsigned int * methodCodeSize, JitFlags * compileFlags, void * inlineInfoPtr) Line 7164	C++
 	clrjit_browser_wasm32_x64.dll!CILJit::compileMethod(ICorJitInfo * compHnd, CORINFO_METHOD_INFO * methodInfo, unsigned int flags, unsigned char * * entryAddress, unsigned int * nativeSizeOfCode) Line 279	C++
 	jitinterface_x64.dll!JitCompileMethod(CorInfoExceptionClass * * ppException, ICorJitCompiler * pJit, void * thisHandle, void * * callbacks, CORINFO_METHOD_INFO * methodInfo, unsigned int flags, unsigned char * * entryAddress, unsigned int * nativeSizeOfCode) Line 36	C++
 	[Managed to Native Transition]	
 	ILCompiler.RyuJit.dll!Internal.JitInterface.CorInfoImpl.CompileMethodInternal(ILCompiler.DependencyAnalysis.IMethodNode methodCodeNodeNeedingCode, Internal.IL.MethodIL methodIL) Line 179	C#
 	ILCompiler.RyuJit.dll!Internal.JitInterface.CorInfoImpl.CompileMethod(ILCompiler.DependencyAnalysis.IMethodCodeNode methodCodeNodeNeedingCode, Internal.IL.MethodIL methodIL) Line 72	C#
 	ILCompiler.LLVM.dll!ILCompiler.LLVMCodegenCompilation.CompileSingleMethod(Internal.JitInterface.CorInfoImpl corInfo, ILCompiler.DependencyAnalysis.LLVMMethodCodeNode methodCodeNodeNeedingCode) Line 132	C#
 	ILCompiler.LLVM.dll!ILCompiler.LLVMCodegenCompilation.CompileSingleThreaded(System.Collections.Generic.List<ILCompiler.DependencyAnalysis.LLVMMethodCodeNode> methodsToCompile) Line 117	C#
 	ILCompiler.LLVM.dll!ILCompiler.LLVMCodegenCompilation.ComputeDependencyNodeDependencies(System.Collections.Generic.List<ILCompiler.DependencyAnalysisFramework.DependencyNodeCore<ILCompiler.DependencyAnalysis.NodeFactory>> obj) Line 95	C#
 	ILCompiler.DependencyAnalysisFramework.dll!ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer<ILCompiler.DependencyAnalysisFramework.NoLogStrategy<ILCompiler.DependencyAnalysis.NodeFactory>, ILCompiler.DependencyAnalysis.NodeFactory>.ComputeDependencies(System.Collections.Generic.List<ILCompiler.DependencyAnalysisFramework.DependencyNodeCore<ILCompiler.DependencyAnalysis.NodeFactory>> deferredStaticDependencies) Line 174	C#
 	ILCompiler.DependencyAnalysisFramework.dll!ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer<ILCompiler.DependencyAnalysisFramework.NoLogStrategy<ILCompiler.DependencyAnalysis.NodeFactory>, ILCompiler.DependencyAnalysis.NodeFactory>.ComputeMarkedNodes() Line 302	C#
 	ILCompiler.LLVM.dll!ILCompiler.LLVMCodegenCompilation.CompileInternal(string outputFile, ILCompiler.ObjectDumper dumper) Line 63	C#
 	ILCompiler.Compiler.dll!ILCompiler.Compilation.ILCompiler.ICompilation.Compile(string outputFile, ILCompiler.ObjectDumper dumper) Line 478	C#
 	ilc.dll!ILCompiler.Program.Run(string[] args) Line 779	C#
```

jkotas:
> Rationalization is basically a phase where IR is converted into LIR, my demo called LLVM right before the rationalization (thus, I used IR). 

@AndyAyersMS Do you think it is better to create LLVM IR from RyuJIT IR before or after rationalization?

src/coreclr/jit/compiler.cpp
```diff
@@ -7670,9 +7774,17 @@ void CompTimeSummaryInfo::AddInfo(CompTimeInfo& info, bool includePhases)
 
         // Update the totals and maxima.
         m_total.m_byteCodeBytes += info.m_byteCodeBytes;
+#ifdef TARGET_WASM
+        m_maximum.m_byteCodeBytes = std::max(m_maximum.m_byteCodeBytes, info.m_byteCodeBytes);
```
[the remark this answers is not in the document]
jkotas: clr/jit has been growing its own std under `src\coreclr\inc\clr_std` and `src\coreclr\inc\clr_std` to solve this sort of problems. I think the best solution would be to add min/max to that.

AndyAyersMS:
>  Do you think it is better to create LLVM IR from RyuJIT IR before or after rationalization?

I would say after rationalization. Linear IR is easier to deal with than trees, and some the transformations rationalize does should produce IR that is simpler to reason about.

yowl:
Going to record here what the size numbers are for the debug build of HelloWasm before this change.  I don't know if smaller is necessarily going to be better once this is done performance wise, but the size will be interesting nevertheless.

```
.bc 49,410,632 bytes
.wasm 25,517,511 bytes
.txt 238,535,841bytes
```

yowl:
First method compiles (and runs) although admittedly it's probably not called when using the repro project.  Method is not the most exciting https://github.com/dotnet/runtimelab/blob/def5b33ae2de9b9ff11003f62b10584cd59cb3a1/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/GCStress.cs#L14 with `FEATURE_GC_STRESS` undefined.  
Stats for compiling just this one method:
```
RyuJIT compilation results, total methods 17098 RyuJit Methods 1 % 0.0058
```

LLVM Modules is

```
; ModuleID = 'netscripten-clrjit'
source_filename = "netscripten-clrjit"
target datalayout = "e-m:e-p:32:32-i64:64-n32:64-S128"
target triple = "wasm32-unknown-emscripten"

define internal void @S_P_CoreLib_System_Runtime_GCStress__Initialize() {
  call void @llvm.donothing()
  ret void
}

; Function Attrs: nounwind readnone willreturn
declare void @llvm.donothing() #0

attributes #0 = { nounwind readnone willreturn }
```
Have to compile and link manually for now but then
image

A small start at least.

yowl:
@jkotas Any thoughts on what is going to be acceptable for getting the LLVM side of things into CI?  It's going to need the headers and .libs (or equivalents for *nix).  
A couple of ideas from opposite ends of the spectrum:

1. Download and build LLVM (at least LLVMCore/LLVMBitWriter/Dwarf..?) - Going to add a fair bit to the CI time.
2. Create our own nuget with a preconfigured build of headers and the libs

jkotas:
For this dotnet/runtimelab branch, I think it is fine for the CI to take longer.

src/coreclr/jit/compiler.cpp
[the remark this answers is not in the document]
yowl: I'm adding min/max to the `std` namespace.  For places like
https://github.com/dotnet/runtimelab/blob/684559a317c74725128f107ae70321dcb7d61f18/src/coreclr/inc/contract.inl#L359
Should I change those, adding 
```
add_compile_definitions(NOMINMAX)
```
to the `CMakeLists.txt` so it applies for all projects or I might be able to getaway with leaving them and wrapping `std::min` into `(std::min)` so they can coexist?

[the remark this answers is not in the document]
jkotas: I do not have an opinion. I am not able to tell what will turn out better.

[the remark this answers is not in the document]
yowl: This is probably the smallest change I can do to remove these `#ifs` https://github.com/dotnet/runtimelab/pull/647/commits/3cf6e3814061c1d94d27306f834bc8a54ca6cd8c
I added a stripped down copy of std::min/max here 
https://github.com/dotnet/runtimelab/pull/647/files#diff-4be6621f6c0e29394673c6cf43f31cfd000d545e289417c0c62de0fe6de2827f  
I deleted some of the signature parts compared to the version that is in `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.28.29333\include\utility` - removed `_NODISCARD` and the `noexcept`s

I added 

https://github.com/dotnet/runtimelab/pull/647/commits/3cf6e3814061c1d94d27306f834bc8a54ca6cd8c#diff-ca6f589b7abc874942a222dbbbdf2f647cefc33e7a43ab57ed30ebb09df858dd

just to `compiler.cpp` so other files can continue to use `min` and `max`.  And that's about it.

yowl:
@jkotas  This has gotten pretty long in number of commits, especially the attempts to get the CI working, which I tried to squash, but must have failed.  Do you want me to squash the whole lot?

Its ready for review now that the nuget package is working and the hard coded method is removed (it does 9 methods currently, which is all the `nop;ret void` that it encounters).   I put in a bit of a temporary fudge in `src\coreclr\nativeaot\BuildIntegration\Microsoft.NETCore.Native.targets` to handle the fact that while coverage is brought to 100% there will be 2 native object files.

From here I was going to suggest that coverage of LIR is done in smaller PRs, each one adding to the % of method coverage.

src/coreclr/jit/morph.cpp
```diff
@@ -8621,8 +8624,13 @@ void Compiler::fgMorphTailCallViaJitHelper(GenTreeCall* call)
     assert(ppArg != nullptr);
     assert(*ppArg == nullptr);
 
+#ifndef TARGET_WASM
     unsigned nOldStkArgsWords =
         (compArgSize - (codeGen->intRegState.rsCalleeRegArgCount * REGSIZE_BYTES)) / REGSIZE_BYTES;
+#else
+    unsigned nOldStkArgsWords = 0;
+    assert(false); // TODO: Wasm: what to do here?
```
jkotas, line 8632: This whole method is x86-specific. It should be fine to put the whole method into x86 ifdef.

src/coreclr/jit/regalloc.cpp
```diff
@@ -385,7 +387,11 @@ void Compiler::raMarkStkVars()
                                   // stack frame
 
     NOT_STK:;
+#ifndef TARGET_WASM
         varDsc->lvFramePointerBased = codeGen->isFramePointerUsed();
+#else
+        varDsc->lvFramePointerBased = false; // TODO Wasm sensible default?
```
jkotas, line 393: Can this whole file be ifdefed out for wasm?

src/coreclr/jit/stacklevelsetter.cpp
```diff
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+#ifndef TARGET_WASM
```
jkotas, line 4: It may be easier to just exclude the .cpp files in CMakeLists.txt instead of ifdefing out each one end-to-end. (The end-to-end ifdef is probably a fine solution for header files.)

src/coreclr/jit/targetwasm.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include "target.h"
+
+const char*            Target::g_tgtCPUName           = "wasm";
+const Target::ArgOrder Target::g_tgtArgOrder          = ARG_ORDER_R2L;
+const Target::ArgOrder Target::g_tgtUnmanagedArgOrder = ARG_ORDER_R2L;
+
+// clang-format off
+#ifdef UNIX_AMD64_ABI
+const regNumber intArgRegs [] = { REG_EDI, REG_ESI, REG_EDX, REG_ECX, REG_R8, REG_R9 };
```
jkotas, line 21: Is this still needed? I would expect this to needed once `codegen*` is excluded.

src/coreclr/jit/typelist.h
```diff
@@ -3,7 +3,11 @@
 
 #define GCS EA_GCREF
 #define BRS EA_BYREF
+#if defined(TARGET_WASM)
+#define PS TARGET_POINTER_SIZE
```
jkotas, line 7: This can be `TARGET_POINTER_SIZE` unconditionally

src/coreclr/jit/unwindwasm.cpp
```diff
@@ -0,0 +1,426 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+/*XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```
jkotas, line 4: This should be only called from codegen. None of it should be needed (or be reachable).

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
```diff
@@ -173,6 +175,74 @@ public ILImporter(LLVMCodegenCompilation compilation, MethodDesc method, MethodI
             _builder = Context.CreateBuilder();
         }
 
+//         [DllImport(JitSupportLibrary)]
```
jkotas, line 178: Delete?

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/RyuJitLLVMCompilation.cs
```diff
@@ -0,0 +1,200 @@
+// // Licensed to the .NET Foundation under one or more agreements.
+// // The .NET Foundation licenses this file to you under the MIT license.
+//
```
jkotas, line 3: Delete? The whole file is commented out.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public unsafe sealed partial class CorInfoImpl
+    {
+        [ThreadStatic]
+        private static CorInfoImpl _thisStatic;
+
+        [UnmanagedCallersOnly]
+        public static byte* getMangledMethodName(IntPtr thisHandle, CORINFO_METHOD_STRUCT_* ftn)
+        {
+            //var _this = GetThis(thisHandle); // TODO: this doesn't work, but how does it cope anyway with this being moved by the GC?
```
jkotas, line 17: > but how does it cope anyway with this being moved by the GC?

The IntPtr points to stack location where `CorInfoImpl` reference is stored. 

JIT/EE interface assumes that the JIT does not store the this pointer in a static.  `registerLlvmCallbacks` stores the this pointer in a static. I suspect that it is why it does not work.

jkotas (commented):
Thank you for pushing this forward!

src/coreclr/jit/codegencommon.cpp
```diff
@@ -30,6 +30,7 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 #include "patchpointinfo.h"
 
 /*****************************************************************************/
+#endif //!TARGET_WASM
 
 const BYTE genTypeSizes[] = {
```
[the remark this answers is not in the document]
yowl: Thanks, 
If I've got this right at https://github.com/dotnet/runtimelab/pull/647/commits
/76fb9b5a3ffc78b4164526e4a4f706b4fac3cbe9 then I'll do the runtime PR

src/coreclr/jit/morph.cpp
yowl: There's one problem: https://github.com/dotnet/runtimelab/blob/9b5c306f8983f90036a0f2a49c7c0d2393260c36/src/coreclr/jit/morph.cpp#L7749  What should I do with that reference?

SingleAccretion: >  What should I do with that reference?

This path is unreachable on anything but `x86` (see `fgCanTailCallViaJitHelper`), and is the place that calls `fgMorphTailCallViaJitHelper`. I'd think `fgMorphTailCallViaJitHelper` can be left as-is (and the compiler should eliminate it anyway)?

jkotas: Right, `fgMorphTailCallViaJitHelper` is reachable on x86 only, and the compiler should be eliminating it today. It just not obvious at all from cursory look at the code. If the whole method was under ifdef, it would be much more obvious.

SingleAccretion: Makes sense. The complication is of course that `fgCanTailCallViaJitHelper` can return `false` for methods that use `stackalloc`, which means the `else` path is used on `x86` too.

Perhaps something like the following would work (optionally `ifdef`ing out both `fgCanTailCallViaJitHelper` and `fgMorphTailCallViaJitHelper` too):
```C++
#ifdef TARGET_X86
if (fgCanTailCallViaJitHelper())
{
    tailCallViaJitHelper = true;
}
#endif

if (!tailCallViaJitHelper)
{
    // The former "else" block
}
```

src/coreclr/jit/stacklevelsetter.cpp
yowl: I think that's a good idea and will consolidate the removal of files that are not required.

viv718918 (approved):

src/coreclr/inc/switches.h
```diff
@@ -65,6 +65,8 @@
     #define USE_UPPER_ADDRESS       0
 #endif // !HOST_UNIX
 
+#elif defined(TARGET_WASM)
+    #define USE_UPPER_ADDRESS       0 // TODO : what's this?
```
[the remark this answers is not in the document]
yowl: thanks, have left it in, but updated the comment

src/coreclr/jit/compiler.cpp
```diff
@@ -24,6 +24,18 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 #include "patchpointinfo.h"
 #include "jitstd/algorithm.h"
 
+#undef min
+#undef max
+
+#if defined(TARGET_WASM)
+#include "llvm.h"
+#else
+// TODO: how to get different exports.def for the different clrjits?
```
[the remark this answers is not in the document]
yowl: I handled the exports.def in the CMakeLists.txt so deleted this comment.

src/coreclr/jit/compiler.cpp
```diff
@@ -4360,6 +4445,16 @@ void Compiler::EndPhase(Phases phase)
     mostRecentlyActivePhase = phase;
 }
 
+#if defined(TARGET_WASM)
+inline void DoLlvmPhase(Compiler* pCompiler)
+{
+    Llvm* llvm = new Llvm();
+    llvm->Compile(pCompiler);
+    delete llvm;
+    //assert(false);
```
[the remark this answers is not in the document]
yowl: Done

src/coreclr/jit/typelist.h
yowl: ok, push to runtime?

src/coreclr/jit/unwindwasm.cpp
yowl: deleted

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
yowl: yes, done, thanks

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/RyuJitLLVMCompilation.cs
yowl: Done.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
yowl: Thanks for the explanation, have done something similar so this now works without the static.

src/coreclr/jit/compiler.cpp
```diff
@@ -5056,6 +5155,11 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
     Rationalizer rat(this); // PHASE_RATIONALIZE
     rat.Run();
 
+#if defined(TARGET_WASM)
+    // TODO:after rat, but better before?
```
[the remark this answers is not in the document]
yowl: No.  Deleted the comment.

src/coreclr/jit/compiler.h
```diff
@@ -11255,10 +11293,14 @@ const instruction INS_SQRT = INS_fsqrt;
 
 /*****************************************************************************/
 
+#ifndef TARGET_WASM
```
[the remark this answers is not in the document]
yowl: Have tidied a lot of these up, following the advice below to do most of it in the CMakeLists.txt.

src/coreclr/jit/importer.cpp
```diff
@@ -19774,7 +19774,11 @@ void Compiler::impInlineInitVars(InlineInfo* pInlineInfo)
                         return;
                     }
                 }
+#ifndef TARGET_WASM
                 else if (genTypeSize(sigType) < EA_PTRSIZE)
+#else
+                else if (genTypeSize(sigType) < TARGET_POINTER_SIZE) // TODO: is this sensible in the abscence of EA_PTRSIZE?
```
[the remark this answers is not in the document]
yowl: will do, thanks.

src/coreclr/jit/morph.cpp
yowl: have followed the advice here.

yowl:
Thanks @jkotas @SingleAccretion for the help.

src/coreclr/jit/typelist.h
jkotas: Yes, that may be nice.

jkotas (approved):
Thank you!

PR #649, opened 2021-02-06
fixed failing Regex unit tests 

veanes:
Some original Regex unit tests failed because of the new DFA option used to be out of range.
Updated related unit tests for SRM and also disabled Unicode category BDD initialization as it has an upfront cost that is almost never justified.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
@@ -134,11 +134,14 @@ namespace System.Text.RegularExpressions
         /// </summary>
         private static SRM.Regex InitializeSRM(RegexNode rootNode, RegexOptions options)
         {
-            // TBD: this could potentially be supported quite easily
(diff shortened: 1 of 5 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
```diff
@@ -134,11 +134,14 @@ private void Init(string pattern, RegexOptions options, TimeSpan matchTimeout, C
         /// </summary>
         private static SRM.Regex InitializeSRM(RegexNode rootNode, RegexOptions options)
         {
-            // TBD: this could potentially be supported quite easily
+            // TBD: this could potentially be supported quite easily but is not of priority
```
stephentoub, line 137: Yeah, we also excluded RTL from a bunch of the optimizations we did in .NET 5, as it's such a rarely used option it wasn't worth it.

stephentoub (approved):

veanes:
Merging with regex unit test fix related to new DFA option.

PR #651, opened 2021-02-07
Trigger precise static cctors from PInvokes

### LINKED ISSUE #648: [NativeAOT] Segmentation fault calling SHA256.HashData on linux
Calling SHA1/SHA256 hash methods (HashData, ComputeHash) lead to segmentation fault on Debian 10.

Code:
`var buffer = SHA256.HashData(new byte[] {0});
Console.WriteLine(buffer.Length);`

gdb output:
`Reading symbols from Showcase...done.
(gdb) run
Starting program: /mnt/d/Development/Workshop/Showcase/bin/Release/net5.0/linux-x64/native/Showcase
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffd7548700 (LWP 5524)]

Thread 1 "Showcase" received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()`

jkotas:
Fixes #648

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
@@ -1716,6 +1716,11 @@ namespace Internal.JitInterface
             if (method.IsRawPInvoke())
                 return false;
 
+            // Stub is required to trigger precise static constructor
+            TypeDesc owningType = method.OwningType;
+            if (_compilation.HasLazyStaticConstructor(owningType) && !((MetadataType)owningType).IsBeforeFieldInit)
+                return true;
+
             // We could have given back the PInvoke stub IL to the JIT and let it inline it, without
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
```diff
@@ -1716,6 +1716,11 @@ private bool pInvokeMarshalingRequired(CORINFO_METHOD_STRUCT_* handle, CORINFO_S
             if (method.IsRawPInvoke())
                 return false;
 
+            // Stub is required to trigger precise static constructor
+            TypeDesc owningType = method.OwningType;
+            if (_compilation.HasLazyStaticConstructor(owningType) && !((MetadataType)owningType).IsBeforeFieldInit)
```
jkotas, line 1721: Adding this check into `IsPInvokeStubRequired` would require passing Compilation into new places, so I have added it here instead.

MichalStrehovsky (approved):
Thank you!

PR #658, opened 2021-02-08
Optimize away unused virtual implementations on abstract classes

MichalStrehovsky:
This change lets us optimize away virtual method bodies on abstract classes that are overriden in all derived classes.

E.g.

```csharp
class Base
{
    public virtual void DoSomething() => CallExpensiveMethod();
}

class Derived : Base
{
    public override void DoSomething() => SomeOtherMethod();
}
```

We shouldn't need to generate `Base.DoSomething` above since nobody calls that.

The optimization reuses the tentative instance methods introduced last week - the first commit just extracts that into a reusable component. The second commit makes us consider references to virtual method on abstract classes tentative, until e.g. a derived class that uses the base implementation makes them not tentative anymore.

We would probably get more bang if we could this to non-abstract classes too, but it's pretty difficult to discern "constructed EEType that is only constructed because it's a base of an allocated type". The biggest complication is reflection.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilationModuleGroup.Aot.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilationModuleGroup.Aot.cs
@@ -78,5 +78,11 @@ namespace ILCompiler
         /// If true, instance methods will only be generated once their owning type is created.
         /// </summary>
         public abstract bool AllowInstanceMethodOptimization(MethodDesc method);
+
+        /// <summary>
+        /// If true, virtual methods on abstract types will only be generated once a non-abstract derived
+        /// type that doesn't override the virtual method is created.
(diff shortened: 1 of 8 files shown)


jkotas:
I would be fine with this. If we have the capability, we can also use it where possible - even it means just 0.6% for what you have measured.

jkotas (approved):
Thank you!

MichalStrehovsky:
Who disabled "rebase and merge" and for what reason? Sigh. Merge commit it is.

jkotas:
I was asked to disable it to match dotnet/runtime.

MichalStrehovsky:
> I was asked to disable it to match dotnet/runtime.

Sigh. I remember that issue in the runtime repo and it didn't make sense to me at that time and it doesn't make sense to me today after re-reading it either. Yay for consistency.

PR #663, opened 2021-02-09
ObjWriter fixes for Windows ARM64

AntonLapounov:
- Translate `IMAGE_REL_BASED_RELPTR32` relocations to `IMAGE_REL_ARM64_REL32`.
- Fix emission of `.pdata` records.
- Minor fix for inconsistent indentation introduced during Linux ARM64 work.

### DIFF
--- a/src/coreclr/tools/aot/ObjWriter/llvm.patch
+++ b/src/coreclr/tools/aot/ObjWriter/llvm.patch
@@ -1,3 +1,15 @@
+diff --git a/include/llvm/BinaryFormat/COFF.h b/include/llvm/BinaryFormat/COFF.h
+index b395db6eaa8..74820a8f79f 100644
+--- a/include/llvm/BinaryFormat/COFF.h
++++ b/include/llvm/BinaryFormat/COFF.h
+@@ -395,6 +395,7 @@ enum RelocationTypesARM64 {
+   IMAGE_REL_ARM64_ADDR64 = 0x000E,
+   IMAGE_REL_ARM64_BRANCH19 = 0x000F,
+   IMAGE_REL_ARM64_BRANCH14 = 0x0010,
++  IMAGE_REL_ARM64_REL32 = 0x0011,
+ };
+ 
+ enum COMDATType {
 diff --git a/include/llvm/MC/MCObjectStreamer.h b/include/llvm/MC/MCObjectStreamer.h
 index 7c1189e46ab..d1d77c97311 100644
 --- a/include/llvm/MC/MCObjectStreamer.h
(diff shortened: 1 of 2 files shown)


src/coreclr/tools/aot/ObjWriter/llvm.patch
```diff
@@ -1,3 +1,15 @@
+diff --git a/include/llvm/BinaryFormat/COFF.h b/include/llvm/BinaryFormat/COFF.h
+index b395db6eaa8..74820a8f79f 100644
+--- a/include/llvm/BinaryFormat/COFF.h
++++ b/include/llvm/BinaryFormat/COFF.h
+@@ -395,6 +395,7 @@ enum RelocationTypesARM64 {
+   IMAGE_REL_ARM64_ADDR64 = 0x000E,
+   IMAGE_REL_ARM64_BRANCH19 = 0x000F,
+   IMAGE_REL_ARM64_BRANCH14 = 0x0010,
++  IMAGE_REL_ARM64_REL32 = 0x0011,
```
jkotas, line 9: Have you looked at how hard it would be to update to newer LLVM that includes this change?

src/coreclr/tools/aot/ObjWriter/objwriter.cpp
```diff
@@ -495,15 +491,23 @@ void ObjectWriter::EmitWinFrameInfo(const char *FunctionName, int StartOffset,
...
 
-  // end Offset
-  const MCExpr *EndOfs = MCConstantExpr::create(EndOffset, *OutContext);
-  Streamer->EmitValue(MCBinaryExpr::createAdd(BaseRefRel, EndOfs, *OutContext),
-                      4);
+  if (Arch == Triple::x86 || Arch == Triple::x86_64) {
+    // end Offset
+    const MCExpr *EndOfs = MCConstantExpr::create(EndOffset, *OutContext);
+    Streamer->EmitValue(
+        MCBinaryExpr::createAdd(BaseRefRel, EndOfs, *OutContext), 4);
```
jkotas, line 509: What is this space for on x86/x64?

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ObjWriter/llvm.patch
AntonLapounov: The latest [Arch64WinCOFFObjectWriter.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AArch64/MCTargetDesc/AArch64WinCOFFObjectWriter.cpp) in LLVM still does not generate `IMAGE_REL_ARM64_REL32` relocations, so I would have to change that file anyway.

The version of LLVM we currently use takes 0.78 GiB for the repository + 1.6 GiB for build artifacts.  The latest LLVM takes 2.5 GiB for  the repository + 2.3 GiB for build artifacts. It takes significantly longer to clone and build. Most importantly, incremental build of LLVM with MSBuild and `-DLLVM_OPTIMIZED_TABLEGEN=1` is broken and rebuilds almost everything (35 minutes on my machine). Need to follow up on that.

src/coreclr/tools/aot/ObjWriter/objwriter.cpp
AntonLapounov: Here we emit an `IMAGE_RUNTIME_FUNCTION_ENTRY` record. For x86/x64 it contains `BeginAddress`, `EndAddress`, and `union { UnwindInfoAddress; UnwindData; }`. For ARM/ARM64 it contains just `BeginAddress` and `UnwindData`.

PR #666, opened 2021-02-10
Add sample how to create WebAPI

kant2002:
Fortunately this is mostly documenting how to add NativeAOT to existing project. Default Web Application template works out of the box.

### DIFF
--- a/samples/WebApi/README.md
+++ b/samples/WebApi/README.md
@@ -1,0 +1,63 @@
+# Building a WebAPI app with NativeAOT
+
+NativeAOT is an AOT-optimized .NET Core runtime. This document will guide you through compiling a .NET Core Web API application with NativeAOT.
+
+## Create your app
+Open a new shell/command prompt window and run the following commands.
+```bash
+> dotnet new webapi -o myApp
+> cd myApp
+```
+
+## Add NativeAOT to your project
+To use NativeAOT with your project, you need to add a reference to the ILCompiler NuGet package that contains the NativeAOT ahead of time compiler and runtime.
+For the compiler to work, it first needs to be added to your project.
+
+In your shell/command prompt navigate to the root directory of your project and run the command:
+
+```bash
+> dotnet new nuget
+```
+
+This will add a nuget.config file to your application. Open the file and in the ``<packageSources> `` element under ``<clear/>`` add the following:
+
+```xml
+<add key="dotnet-experimental" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json" />
+<add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+```
+
+Once you've added the package source, add a reference to the compiler by running the following command:
+
+```bash
+> dotnet add package Microsoft.DotNet.ILCompiler -v 6.0.0-*
+```
+
+## Restore and Publish your app
+
+Once the package has been successfully added it's time to compile and publish your app! In the shell/command prompt window, run the following command:
+
+```bash
+> dotnet publish -r <RID> -c <Configuration>
+```
+
+where `<Configuration>` is your project configuration (such as Debug or Release) and `<RID>` is the runtime identifier (one of win-x64, linux-x64, osx-x64). For example, if you want to publish a release configuration of your app for a 64-bit version of Windows the command would look like:
+
+```bash
+> dotnet publish -r win-x64 -c release
+```
+
+Once completed, you can find the native executable in the root folder of your project under `/bin/x64/<Configuration>/net5.0/publish/`
+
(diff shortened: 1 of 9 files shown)


samples/WebApi/README.md
```diff
@@ -0,0 +1,63 @@
...
+> dotnet publish -r win-x64 -c release
+```
+
+Once completed, you can find the native executable in the root folder of your project under `/bin/x64/<Configuration>/net5.0/publish/`
+
+Once you run application, it would be available at default https://localhost:5001/weatherforecast address. Check it out!
+
+## Using reflection
+
+More advanced use cases might hit issues due to reflection. For information about the topic see [MonoGame sample](../MonoGame/README.md) or [Reflection in AOT mode](../../docs/using-nativeaot/reflection-in-aot-mode.md).
```
jkotas, line 55: These links are not valid

samples/WebApi/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Try it out!
+
+If you are running macOS, make sure you have [libuv](https://github.com/libuv/libuv) installed, as ASP.NET is built on top of libuv. You can use [homebrew](https://brew.sh/) to get it (`brew install libuv`).
```
jkotas, line 59: This is no longer the case

jkotas:
Since this works out of the box, these steps are verbatim copy of https://github.com/dotnet/runtimelab/tree/feature/NativeAOT/samples/HelloWorld with console replaced with webapi. Do we really need to have it checked in into the repo?

jkotas:
E.g. Would it be enough to just have one sentence mention of webapi template at the end of https://github.com/dotnet/runtimelab/tree/feature/NativeAOT/samples/HelloWorld  ?

ryancheung:
Could I ask whether SharpDX works with NativeAOT for now? I tried MonoGame WindowsDX project with NativeAOT and it crashes while D3D graphics device initializing.

kant2002:
I think about it two fold.
In general I agree that probably as a sample it has nothing to show. Maybe that PR can be abandoned or left hanging open.
Second opinion is that NativeAOT is lacking some clarification what can and what cannot be done using it in terms of end-user technology. Most likely developer willing to try NativeAOT will think about migrate existing project to try if it has any benefits for them. That mean they looking for answers
- Is EF Core supported by NativeAOT?
- Is MVC Core supported by NativeAOT?
- Is Active Directory support working in NativeAOT?
- Is SharpDX working in NAtive AOT etc.

I do not really know proper way to answer these questions which can be suitable for this project (given it's technical nature) so I use this PR as a way to discuss these things + show some examples.

jkotas:
>  MonoGame WindowsDX project with NativeAOT and it crashes while D3D graphics device initializing.

#668 is fixing this crash. The game starts up with this fix, but crashes in WinForms on missing COM interop.

MichalStrehovsky:
There's the limitations document listed here: https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/docs/using-nativeaot/limitations.md. We could add some popular libraries, but the list is going to be never ending.

We're slowly working on adding analysis so that when you compile your app with NativeAOT, you'll get clear warnings when something might not work (or won't work for sure).

There are multiple aspects to that:

* Trimming friendliness - this is mostly done and on par with analysis from IL Linker - you can turn it on by adding `<SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>` to you project.
* AOT friendlines - partially done and enabled by default. It will warn for things like MakeGenericType
* Built-in COM interop usage - this is done in IL Linker and well understood (IDispatch has problems with trimming and the problems are not limited to IDispatch). We should port it to ILC - #670
* Single file friendliness - this is done in the IL linker repo and well understood - we should port it to ILC - #669

ryancheung:
After #668 applied I can confirm MonoGame WindowsDX game works with NativeAOT now with rooting all assemblies.

hez2010:
I don't think this works without a `rd.xml`, you may encounter HTTP 404 error while navigating to any endpoints defined in Controllers.
ASP.NET Core reflects the assembly to find controllers and add endpoints at startup, you need to add all your controllers to `rd.xml` to make sure they won't be trimmed.

> Is EF Core supported by NativeAOT?

No if without https://github.com/dotnet/efcore/pull/24211.
> Is MVC Core supported by NativeAOT?

Views are not going to work.
```

kant2002:
When I say I want to have repository with samples which explain what's possible and what's not. I mean resource similar in spirit like this one: https://github.com/zhangchaoza/corert-samples where different technologies evaluated separately and provided common snippets how somebody can start playing with technology.

jkotas:
I would be fine with linking to repos like https://github.com/zhangchaoza/corert-samples  or https://github.com/teobugslayer/AvaloniaCoreRTDemo from here.

PR #668, opened 2021-02-10
Fix Marshal Com Helpers

jkotas:


jkotas:
I have somehow missed these when cleaning up the function pointer intrinsics.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
```diff
@@ -23,8 +23,7 @@ public static unsafe int AddRef(IntPtr pUnk)
             if (pUnk == IntPtr.Zero)
                 throw new ArgumentNullException(nameof(pUnk));
 
-            return CalliIntrinsics.StdCall__AddRef(((__com_IUnknown*)(void*)pUnk)->pVtable->
-                pfnAddRef, pUnk);
+            return ((delegate* unmanaged<IntPtr, int>)(*(*(void***)pUnk + 1 /* IUnknown.AddRef slot */)))(pUnk);
```
jkotas, line 26: Copy & paste from CoreCLR file

MichalStrehovsky: Yes, I remember this line being popular on twitter.

MichalStrehovsky (approved):
Weird that they didn't have McgIntrinsicsAttribute.

jkotas:
It had McgIntrinsicsAttribute via partial class.

MichalStrehovsky:
> It had McgIntrinsicsAttribute via partial class.

Ah. No wonder this was missed. Good riddance.

PR #671, opened 2021-02-10
Generate debug info for all types that can be allocated

MichalStrehovsky:
We were only generating debug information for types that show up as locals or for which we generated methods. While this covered a lot of things, it didn't cover a pretty common scenario of types with shared method bodies.

I'm also deleting the `ISymbolNodeWithDebugInfo` abstraction. It was added for .NET Native and even there it doesn't look like it was implemented (the only implementation of the interface I could find was the `NullTypeIndexDebugInfo` I'm deleting).

We might want to add something like this in the future, but for now I'm just violating the layering and special casing `ConstructedEETypeNode` in the object writer.


jkotas (approved):
Thank you!

PR #672, opened 2021-02-10
Set AllowCustomResourceTypes to false

MichalStrehovsky:
Gets rid of a trimming warning. The 6.0 SDK will enable this (dotnet/sdk#15702), so this is just a shim for NativeAOT with the 5.0 SDK.


jkotas (approved):

joperezr (approved):

jkotas:
> The 6.0 SDK will enable this (dotnet/sdk#15702)

@MichalStrehovsky I just tried this with .NET 6.0 SDK and `SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>`. It is not getting automatically enabled for NativeAOT. Do we need more adjustments here?

MichalStrehovsky:
> @MichalStrehovsky I just tried this with .NET 6.0 SDK and SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>. It is not getting automatically enabled for NativeAOT. Do we need more adjustments here?

Looks like we need to set `PublishTrimmed`. There's a conversation on what is it exactly that we want to be set (https://github.com/mono/linker/issues/2085). The quick fix would be to set `PublishTrimmed` to true in our `.props` but maybe lets wait for the conversation to resolve.

Setting PublishTrimmed seems to work though.

PR #676, opened 2021-02-11
Fix ARM64 assembler helpers

AntonLapounov:
- Reuse Linux implementation of `RhpGcPoll` and `RhpGcPollRare` for Windows.
- Fix the parameter register of `RhpInterfaceDispatchSlow` for Windows.
- Fix the parameter register of `RhpVTableOffsetDispatch` for Windows and Linux. I think this function is not used at present.
- Move `RhpInitialDynamicInterfaceDispatch` label to coincide with `RhpInitialInterfaceDispatch` instead of `RhpInterfaceDispatchSlow`. Otherwise, checking `this` for `null` added in #239 will not work as expected. @MichalStrehovsky
- Minor style and comment fixes.
- Fix `IMAGE_REL_ARM64_REL32` relocations, which were off by 4 bytes.  It seems the latest LLVM code is missing this adjustment as well.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/arm64/GcProbe.asm
+++ b/src/coreclr/nativeaot/Runtime/arm64/GcProbe.asm
@@ -5,6 +5,7 @@
 
     TEXTAREA
 
+    EXTERN      RhpGcPoll2
     EXTERN      g_fGcStressStarted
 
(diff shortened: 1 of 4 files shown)


jkotas (approved):
Thank you!

PR #677, opened 2021-02-11
Make ThrowInstanceBodyRemoved/ThrowBodyRemoved helpers optional

MichalStrehovsky:
Fixes https://github.com/MichalStrehovsky/zerosharp/issues/9.

I would be fine also with not taking this and making the helpers mandatory. I don't have a strong opinion either way. This makes it easier to experiment with the compiler, but also it's just a two-liner to provide a stub in the core module.


jkotas (approved):
LGTM. Thanks!

PR #678, opened 2021-02-11
Annotate a couple more marshal APIs

### LINKED ISSUE #634: Marshal.GetDelegateForFunctionPointer should be AOT unsafe
We should mark the version that takes System.Type as AOT unfriendly.

But: dataflow analysis should figure out the simple cases (the parameter is a `typeof` that we just see).

Basically do the same thing as we do for `Marshal.SizeOf`.

MichalStrehovsky:
And add intrinsic handling. Fixes #634.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
@@ -100,7 +100,7 @@ namespace System.Runtime.InteropServices
             }
         }
 
-        [RequiresDynamicCode("Marshalling code for the object might not be available")]
+        [RequiresDynamicCode("Marshalling code for the object might not be available. Use the DestroyStructure<T> overload instead.")]
         public static unsafe void DestroyStructure(IntPtr ptr, Type structuretype)
         {
             if (ptr == IntPtr.Zero)
@@ -137,7 +137,7 @@ namespace System.Runtime.InteropServices
             }
         }
 
-        [RequiresDynamicCode("Marshalling code for the object might not be available")]
+        [RequiresDynamicCode("Marshalling code for the object might not be available. Use the StructureToPtr<T> overload instead.")]
         public static unsafe void StructureToPtr(object structure, IntPtr ptr, bool fDeleteOld)
         {
             if (structure == null)
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
@@ -104,7 +104,7 @@ namespace System.Runtime.InteropServices
             return string.CreateStringFromEncoding((byte*)ptr, byteLen, Encoding.UTF8);
         }
 
-        [RequiresDynamicCode("Marshalling code for the object might not be available")]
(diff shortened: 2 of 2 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -330,6 +330,9 @@ enum IntrinsicId
             Expression_New,
             Enum_GetValues,
             Marshal_SizeOf,
+            Marshal_PtrToStructure,
```
jkotas, line 333: `StructureToPtr<T>` as well?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -330,6 +330,9 @@ enum IntrinsicId
...
+            Marshal_DestroyStructure,
+            Marshal_GetDelegateForFunctionPointer,
```
jkotas, line 335: `Marshal.GetFunctionPointerForDelegate` as well?

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
MichalStrehovsky: The generic versions can be analyzed without dataflow and we do that: https://github.com/dotnet/corert/blob/24f1edc938af9954eac9a3385958787ca64956d5/src/ILCompiler.Compiler/src/Compiler/UsageBasedInteropStubManager.cs#L88-L121

There's no nongeneric variant that takes a Type, so there's no way to make this with with the current dataflow analysis (we don't track object types).

jkotas: Ah ok.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
```diff
@@ -136,6 +137,7 @@ public static unsafe void DestroyStructure(IntPtr ptr, Type structuretype)
             }
         }
 
+        [RequiresDynamicCode("Marshalling code for the object might not be available")]
```
jkotas, line 140: Should we include suggestion for how to fix where possible? E.g. "Use Marshal.StructurePtr<T> instead."

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
MichalStrehovsky: Ah. Didn't see it in Intellisense and apparently we obsoleted and editorbrowsable(never)-hidden it.

Seems like the other non-generic version is obsolete. But people still use it. Doesn't hurt to add support for it.

jkotas: I think it has the same problem as `StructureToPtr`. It takes object, not the type. So we can just emit warning for it.

PR #680, opened 2021-02-12
making SRM.Regex constructor threadsafe

veanes:
Build of: **runtimelab (Build Libraries Test Run Release coreclr windows x64 Release)** caused test failures due to thread safety failures. I suspect this is because the constructor of SRM.Regex was not threadsafe. Adding this fix (lock in the constructor) should fix this.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/Regex.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/Regex.cs
@@ -27,16 +27,22 @@ namespace System.Text.RegularExpressions.SRM
             if (root.info.CanBeNullable)
                 throw new NotSupportedException(_DFA_incompatible_with + "pattern allowing 0-length match");
 
-            var partition = root.ComputeMinterms();
-            if (partition.Length > 64)
+            // make sure no two threads can create an SRM regex at the same time
+            // this may cause thread safety issues deep down in multiple places
+            // in particular, BDDs or use of the BDD algebra is not threadsafe in ComputeMinterms
+            lock (this)
             {
-                //more than 64 bits needed to represent a set
-                matcher = new SymbolicRegexBV(root, solver, converter.srBuilder, partition, options);
-            }
-            else
-            {
-                //enough to use 64 bits
-                matcher = new SymbolicRegexUInt64(root, solver, converter.srBuilder, partition, options);
+                var partition = root.ComputeMinterms();
+                if (partition.Length > 64)
+                {
+                    //more than 64 bits needed to represent a set
+                    matcher = new SymbolicRegexBV(root, solver, converter.srBuilder, partition, options);
+                }
+                else
+                {
+                    //enough to use 64 bits
+                    matcher = new SymbolicRegexUInt64(root, solver, converter.srBuilder, partition, options);
+                }
             }
         }
 


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/Regex.cs
```diff
@@ -27,16 +27,22 @@ public Regex(RegexNode rootNode, System.Text.RegularExpressions.RegexOptions opt
             if (root.info.CanBeNullable)
                 throw new NotSupportedException(_DFA_incompatible_with + "pattern allowing 0-length match");
 
-            var partition = root.ComputeMinterms();
-            if (partition.Length > 64)
+            // make sure no two threads can create an SRM regex at the same time
+            // this may cause thread safety issues deep down in multiple places
+            // in particular, BDDs or use of the BDD algebra is not threadsafe in ComputeMinterms
+            lock (this)
```
jkotas, line 33: There is no way for two threads to run constructor in parallel. I do not think this lock is going to fix anything.

stephentoub: In particular, if the problem here is that ComputeMinterms or SymboloicRegexBV/UInt64 are accessing/mutating global state, having each Regex instance synchronize on their own instance is going to be an expensive nop.  Either this needs to lock on global object that all constructors synchronize with, or wherever those accesses are happening "deep down in multiple places" need to be made safe.

jkotas:
@veanes Our prefered workflow is to push the changes into private fork (e.g. https://github.com/veanes/runtimelab) and submit the PR from there. It keeps the list of branches in the main repo clean. Could you please do it next time? 

https://github.com/dotnet/runtime/blob/master/docs/issues-pr-management.md#what-is-dotnetruntimes-branch-policy

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/Regex.cs
veanes: I need to investigate this further, need to have a unit test to replicate the error behavior. Sorry about not pushing into private fork, this was an unintended mistake, also still learning about the overall policies. What it looked like from the error message seemed that the root cause was in ComputeMinterms, this method is only called during the construction of the regex, which is why I thought the lock there would protect.

veanes: By "deep down in multiple places" I meant that BDD algebra is inherently sequential because it internalizes the nodes representing objects through a shared cache in such a way that two equivalent BDDs are actually identical as objects. If two threads were to be allowed to modify those caches simultaneously (even in a thread-safe manner) during some logical operations, I believe it might break functional correctness of the BDD operations (such as AND, OR, NOT) at least as it is now. One way would be to relax on the property that "equivalent BDDs are actually identical" and allow that algebra to be potentially useful in a multithreaded environment. But, right now it is only used the constructor.

stephentoub: > What it looked like from the error message seemed that the root cause was in ComputeMinterms, this method is only called during the construction of the regex, which is why I thought the lock there would protect.

It might, but it needs to lock on something shared across everyone trying to protect the same state.  As currently written, two different threads could be creating two different instances at the same time, and each of them is going to lock on their own instance, two completely different locks.

veanes: Oh, yes, you are absolutely right, silly of me. Indeed, the lock should be on SRM.Regex.solver (not 'this' that makes no sense there).

veanes: I think the best solution is if the BDD algebra itself would be thread-safe. Essentially, most operations that need to modify its internal state (essentially the Boolean operation cache's) would be protected by locks. This would also make it more generally useful in other potential contexts.

veanes:
Also, I will make sure to only submit PR from my private fork, as I said this was an unintended mistake. The fix will be inside the BDD algebra itself. I'll upgrade its operations to be thread safe.

PR #683, opened 2021-02-12
Emit message for RequiresXYZCodeAttribute

MichalStrehovsky:


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/DiagnosticUtilities.cs
```diff
@@ -43,5 +43,41 @@ internal static string GetGenericParameterDeclaringMemberDisplayName(GenericPara
...
+                return null;
+
+            var decoded = ecmaMethod.GetDecodedCustomAttribute("System.Diagnostics.CodeAnalysis", "RequiresUnreferencedCodeAttribute");
+            if (decoded == null)
+                return null;
+
+            var decodedValue = decoded.Value;
+
+            if (decodedValue.FixedArguments.Length != 0)
+                return (string)decodedValue.FixedArguments[decodedValue.FixedArguments.Length - 1].Value;
```
jkotas, line 60: Is there a prior art for using the last argument instead of the first argument as the message?

It does not matter today since the attribute constructor is not overloaded. If we were to overload the attribute constructor, the message is likely going to stay as the first argument and the additional arguments will go after it.

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/DiagnosticUtilities.cs
MichalStrehovsky: Ah. That's copy-paste from how we parse RuntimeImport. That's definitely a pattern that wouldn't pass the API review board.

PR #685, opened 2021-02-12
Early out in VEH for debugger exceptions

### LINKED ISSUE #673: Unhandled exception at 0x00007FFDF663BD5C (KernelBase.dll) in ConsoleApp.exe: 0x00001001
Hello,
when I call an exported function from a C# console application, as soon as I try to step into the exported function I get the exception in the title after a couple of seconds. Before the exception occurs sometimes I can also see the variables value. If I step over and remove the breakpoint (as showed in the video below) I don't get any exception and the program execution flow continue without any problem.
It doesn't matter if I directly pinvoke the function or use a delegate.
`Unhandled exception at 0x00007FFDF663BD5C (KernelBase.dll) in ConsoleApp.exe: 0x00001001.
The Common Language Runtime cannot stop at this exception. Common causes include: incorrect COM interop marshalling and memory corruption. To investigate further use native-only debugging.`

https://user-images.githubusercontent.com/63934725/107553464-15083e80-6bd5-11eb-9c14-98b2c10e69ab.mp4

1.  I suppose that this exception is voluntarily raised from your staff because basically I am trying to debug a library that now is completely native. If I execute the function without stepping through code, can I assume that it will be safe as seems to be from the video or I can expect bad things?
2. How it is possible to debug the C# code? The doc gives some tips and I can debug the provided "LoadLibrary.c", but this is unmanaged and will be very hard if I can't step into C#. Is there a workaround or at least a way to debug portions of C# code?
3. Should I have used `<DirectPInvoke Include="NativeLibrary!add" />` ??
Thanks

Edit: Strangely, from a c++ console app, the same code debug fine or at least the debug output does not shows anything wrong:

https://user-images.githubusercontent.com/63934725/107573332-b26f6c80-6bed-11eb-8334-eb425544aa01.mp4

At this point I am wondering what is the right way to go and if the exceptions I'm getting from a C# console app are due to something wrong from my side.

jkotas:
Fixes #673

### DIFF
--- a/src/coreclr/nativeaot/Runtime/EHHelpers.cpp
+++ b/src/coreclr/nativeaot/Runtime/EHHelpers.cpp
@@ -451,10 +451,17 @@ int32_t __stdcall RhpHardwareExceptionHandler(uintptr_t faultCode, uintptr_t fau
 
 int32_t __stdcall RhpVectoredExceptionHandler(PEXCEPTION_POINTERS pExPtrs)
 {
+    uintptr_t faultCode = pExPtrs->ExceptionRecord->ExceptionCode;
+
+    // Do not interfere with debugger exceptions
+    if (faultCode == STATUS_BREAKPOINT || faultCode == STATUS_SINGLE_STEP)
+    {
+        return EXCEPTION_CONTINUE_SEARCH;
+    }
+
     uintptr_t faultingIP = pExPtrs->ContextRecord->GetIp();
 
     ICodeManager * pCodeManager = GetRuntimeInstance()->FindCodeManagerByAddress((PTR_VOID)faultingIP);
-    uintptr_t faultCode = pExPtrs->ExceptionRecord->ExceptionCode;
     bool translateToManagedException = false;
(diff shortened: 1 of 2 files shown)


src/coreclr/nativeaot/Runtime/EHHelpers.cpp
```diff
@@ -451,10 +451,17 @@ int32_t __stdcall RhpHardwareExceptionHandler(uintptr_t faultCode, uintptr_t fau
 
 int32_t __stdcall RhpVectoredExceptionHandler(PEXCEPTION_POINTERS pExPtrs)
 {
+    uintptr_t faultCode = pExPtrs->ExceptionRecord->ExceptionCode;
+
+    // Do not interfere with debugger exceptions
+    if (faultCode == STATUS_BREAKPOINT || faultCode == STATUS_SINGLE_STEP)
```
jkotas, line 457: CoreCLR VEH has similar logic.

MichalStrehovsky (approved):
Thank you!

PR #689, opened 2021-02-15
Fix emitting CodeView information for locals

AntonLapounov:
* Emit correct CodeView register numbers for ARM64. Support x86 and ARM as well.
* Stop emitting uninitialized `Hdr`'s fields to the PDB file.

Note that ordering of special ARM64 registers is different in `cordebuginfo.h` and `cvconst.h` from DIA SDK:

| Header file | ARM64 register order |
| -- | -- |
| cordebuginfo.h | FP, LR, SP, PC |
| cvconst.h | FP, LR, SP, **ZR**, PC |

I am not sure we can insert the `ZR` register into the `RegNum` enumeration in `cordebuginfo.h` without creating a potential mess between different .NET versions. For instance, the debugger uses the hard-coded value of `REGNUM_AMBIENT_SP`, which would change after the insertion:
https://github.com/dotnet/runtimelab/blob/3d3b11a3e3060583e5af00c6f835c64a72d5cf10/src/coreclr/debug/inc/dbgipcevents.h#L1878

For that reason I handle ARM64's `PC` register separately.  Of course, `PC` is very unlikely to be used for locals, so we might stop at `SP` instead.

### DIFF
--- a/src/coreclr/inc/cvconst.h
+++ b/src/coreclr/inc/cvconst.h
@@ -1580,10 +1580,12 @@ typedef enum CV_HREG_e {
     CV_ARM64_LR     =  80,
     CV_ARM64_SP     =  81,
     CV_ARM64_ZR     =  82,
+    CV_ARM64_PC     =  83,
 
-    // statue register
+    // status registers
 
     CV_ARM64_NZCV   =  90,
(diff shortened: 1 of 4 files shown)


jkotas (approved):
Thank you!

PR #690, opened 2021-02-15
Silence some marshalling warnings

MichalStrehovsky:
* Call into PInvokeMarshal from the generated marshallers. These are all analyzed at compile time and safe. Calling directly into implementation sidesteps the warnings and provides a mild performance boost. Only the `GetFunctionPointerForDelegate` part is actually needed. The other was done for consistency.
* Silence warnings on generic things. I didn't see this before because these are generic and they weren't instantiated anywhere I was looking at.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
@@ -149,7 +149,7 @@ namespace System.Runtime.InteropServices
             Type t = typeof(T);
             if (t.IsGenericType)
             {
-                throw new ArgumentException(SR.Argument_NeedNonGenericType, nameof(t));
+                throw new ArgumentException(SR.Argument_NeedNonGenericType, nameof(T));
             }
 
             return SizeOfHelper(t, throwIfNotMarshalable: true);
@@ -607,7 +607,7 @@ namespace System.Runtime.InteropServices
             Type structureType = typeof(T);
             if (structureType.IsGenericType)
             {
-                throw new ArgumentException(SR.Argument_NeedNonGenericType, nameof(structureType));
+                throw new ArgumentException(SR.Argument_NeedNonGenericType, nameof(T));
             }
 
             return (T)PtrToStructureHelper(ptr, structureType);
@@ -933,9 +933,9 @@ namespace System.Runtime.InteropServices
                 throw new ArgumentException(SR.Argument_NeedNonGenericType, nameof(TDelegate));
             }
 
-            // COMPAT: This block of code isn't entirely correct.
-            // Users passing in typeof(MulticastDelegate) as 't' skip this check
-            // since Delegate is a base type of MulticastDelegate.
+            // For backward compatibility, we allow lookup of existing delegate to
+            // function pointer mappings using abstract MulticastDelegate type. We will check
+            // for the non-abstract delegate type later if no existing mapping is found.
             Type? c = t.BaseType;
             if (c != typeof(Delegate) && c != typeof(MulticastDelegate))
             {
@@ -956,6 +956,8 @@ namespace System.Runtime.InteropServices
             return GetFunctionPointerForDelegateInternal(d);
         }
 
+        [UnconditionalSuppressMessage("AotAnalysis", "IL9700:AotUnfriendlyApi",
+            Justification = "AOT compilers can see the T.")]
         public static IntPtr GetFunctionPointerForDelegate<TDelegate>(TDelegate d) where TDelegate : notnull
         {
             return GetFunctionPointerForDelegate((Delegate)(object)d);


src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -582,6 +584,8 @@ public static void PtrToStructure<T>(IntPtr ptr, [DisallowNull] T structure)
             PtrToStructure(ptr, (object)structure!);
```
jkotas, line 584: Do we need `UnconditionalSuppressMessage` on `void PtrToStructure<T>(IntPtr ptr, [DisallowNull] T structure)` too?

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -582,6 +584,8 @@ public static void PtrToStructure<T>(IntPtr ptr, [DisallowNull] T structure)
...
         }
 
+        [UnconditionalSuppressMessage("AotAnalysis", "IL9700:AotUnfriendlyApi",
+            Justification = "AOT compilers can see the T.")]
```
jkotas, line 588: I am wondering whether it would be better to call `PtrToStructureHelper` instead of suppressing the warning. Some of the argument checks would need to be duplicated, but we would be also able to omit some.

jkotas: (similar to other places like `GetDelegateForFunctionPointer`)

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -891,7 +922,27 @@ public static Delegate GetDelegateForFunctionPointer(IntPtr ptr, Type t)
...
+            Type? c = t.BaseType;
+            if (c != typeof(Delegate) && c != typeof(MulticastDelegate))
+            {
+                throw new ArgumentException(SR.Arg_MustBeDelegate, nameof(TDelegate));
+            }
+
+            return (TDelegate)(object)GetDelegateForFunctionPointerInternal(ptr, t);
         }
 
         [RequiresDynamicCode("Marshalling code for the delegate might not be available. Use the GetFunctionPointerForDelegate<TDelegate> overload instead.")]
```
jkotas, line 948: Do we also need to update `public static IntPtr GetFunctionPointerForDelegate<TDelegate>(TDelegate d) where TDelegate : notnull` to not produce warning?

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
```diff
@@ -891,7 +922,27 @@ public static Delegate GetDelegateForFunctionPointer(IntPtr ptr, Type t)
...
+            }
+
+            // COMPAT: This block of code isn't entirely correct.
+            // Users passing in typeof(MulticastDelegate) as 't' skip this check
+            // since Delegate is a base type of MulticastDelegate.
+            Type? c = t.BaseType;
+            if (c != typeof(Delegate) && c != typeof(MulticastDelegate))
+            {
+                throw new ArgumentException(SR.Arg_MustBeDelegate, nameof(TDelegate));
+            }
```
jkotas, line 943: This odd check should not be needed here at all.

The `MulticastDelegate` corner case needs to be handled on the unmanaged side. It is handled correctly for CoreCLR already.

`Marshal.GetDelegateForFunctionPointer<MulticastDelegate>(Marshal.AllocHGlobal(1))` should throw `Type must derive from Delegate.`. It does so in CoreCLR already even with this check deleted, but not in CoreRT.

jkotas: (We can deal with this separately.)

jkotas (approved):
Thank you!

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
jkotas, line 948: Nit: I just tried this and noticed that there is a double-dot printed in the warning message `Use the GetFunctionPointerForDelegate<TDelegate> overload instead..`. Should we stop appending `.` to the message in ReflecitonMethodBodyScanner.cs?

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
jkotas: This check is actually needed. I am updating the comment and adding a test in dotnet/runtime

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
MichalStrehovsky: I think we'll never be consistent with this. It might be better to just fix this up when the message is generated (append the dot only if it's not there).

PR #692, opened 2021-02-15
Make DataContractSerializer more AOT friendly

MichalStrehovsky:
Only go down the `MakeGenericMethod` path if we can get by without JITting.

Also factoring out the KeyValue part out of the `MakeGenericMethod` path. That looks like general goodness.

### DIFF
--- a/src/libraries/System.Private.DataContractSerialization/src/System/Runtime/Serialization/AccessorBuilder.cs
+++ b/src/libraries/System.Private.DataContractSerialization/src/System/Runtime/Serialization/AccessorBuilder.cs
@@ -60,15 +60,13 @@ namespace System.Runtime.Serialization
                 if (RuntimeFeature.IsDynamicCodeSupported || (!declaringType.IsValueType && !propertyType.IsValueType))
                 {
                     var createGetterGeneric = s_createGetterInternal.MakeGenericMethod(declaringType, propertyType).CreateDelegate<Func<PropertyInfo, Getter>>();
-                    Getter accessor = createGetterGeneric(propInfo);
-                    return accessor;
+                    return createGetterGeneric(propInfo);
                 }
                 else
                 {
                     return (obj) =>
                     {
-                        var value = propInfo.GetValue(obj);
-                        return value;
+                        return propInfo.GetValue(obj);
                     };
                 }
             }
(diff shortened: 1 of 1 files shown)


src/libraries/System.Private.DataContractSerialization/src/System/Runtime/Serialization/AccessorBuilder.cs
```diff
@@ -33,9 +34,43 @@ public static Getter CreateGetter(MemberInfo memberInfo)
...
+                    }
+                }
+
+                // If either of the arguments to MakeGenericMethod is a valuetype, this is going to cause JITting.
+                // Only JIT if dynamic code is supported.
+                if (RuntimeFeature.IsDynamicCodeSupported || (!declaringType.IsValueType && !propertyType.IsValueType))
+                {
+                    var createGetterGeneric = s_createGetterInternal.MakeGenericMethod(declaringType, propertyType).CreateDelegate<Func<PropertyInfo, Getter>>();
+                    Getter accessor = createGetterGeneric(propInfo);
+                    return accessor;
```
jkotas, line 64:
```suggestion
                    return createGetterGeneric(propInfo);
```

src/libraries/System.Private.DataContractSerialization/src/System/Runtime/Serialization/AccessorBuilder.cs
```diff
@@ -58,9 +93,42 @@ public static Setter CreateSetter(MemberInfo memberInfo)
...
+                        }
+                    }
+
+                    // If either of the arguments to MakeGenericMethod is a valuetype, this is going to cause JITting.
+                    // Only JIT if dynamic code is supported.
+                    if (RuntimeFeature.IsDynamicCodeSupported || (!declaringType.IsValueType && !propertyType.IsValueType))
+                    {
+                        var buildSetAccessorGeneric = s_createSetterInternal.MakeGenericMethod(propInfo.DeclaringType!, propInfo.PropertyType).CreateDelegate<Func<PropertyInfo, Setter>>();
+                        Setter accessor = buildSetAccessorGeneric(propInfo);
+                        return accessor;
```
jkotas, line 123:
```suggestion
                        var createSetterGeneric = s_createSetterInternal.MakeGenericMethod(propInfo.DeclaringType!, propInfo.PropertyType).CreateDelegate<Func<PropertyInfo, Setter>>();
                        return createSetterGeneric(propInfo);
```

jkotas (approved):
LGTM. Thank you!

src/libraries/System.Private.DataContractSerialization/src/System/Runtime/Serialization/AccessorBuilder.cs
MichalStrehovsky: These all match the pattern that is used in this file. Granted, I touch a lot of the file, but `GetMakeNewInstanceFunc` still uses this pattern. I think the CoreFX guidelines say to keep existing pattern. But it's true the reasons for this pattern disappeared when debuggers started showing "The method returned X".

PR #693, opened 2021-02-15
Speed up reflection invoke

MichalStrehovsky:
CoreRT/RyuJIT doesn't have the `typeof(X).TypeHandle` optimization that .NET Native had so this was a bit more expensive than it should be (6% of `PropertyInvoke.GetValue` wallclock time was being spent here).


jkotas (approved):
Thank you!

PR #694, opened 2021-02-16
Delete RhpCheckCctor

MichalStrehovsky:
This is unused. We might want to introduce something like this if we ever change how statics are accessed (at which point we can just restore this if needed), but even then we're probably better off with codegen inlining the fast path and at that point it can call into a managed helper to do the slow path.


jkotas (approved):
Thanks!

PR #697, opened 2021-02-16
Delete __ComObject

MichalStrehovsky:
Even if were to hook this up into the pluggable COM support, the base type will definitely not be `System.__ComObject`.

I'm also putting all of this under FEATURE_COMINTEROP.


jkotas (approved):
LGTM. Thanks!

PR #701, opened 2021-02-17
Append dot to warning messages conditionally

jkotas:

PR #703, opened 2021-02-17
Refresh code shared with IL Linker

MichalStrehovsky:
This will have a conflict with #701. I'll deal with that.


jkotas (approved):
Thanks!

PR #704, opened 2021-02-17
Fix BDDs

veanes:
Rewrote the whole BDD algebra to be thread safe. 
This addresses (should fix) the related unit test failures in cloud build.
Some important changes:

All changes to shared dictionaries now use a lock on the BDD algebra to prevent multiple threads changing them, while multithreaded reading from the dictionaries remains ok without locks, which will happen in the long run during matching as most predicates and operations among them have been cached and need not be computed. 

Removed the BDD algebra owner field from BDDs. The owner does now not matter for the semantics to be correct since internalization by caching is now used only for efficiency (as opposed to correctness before). Also made True and False fixed static BDDs. 

Introduced several key changes/simplifications into the core algorithm structure. In particular, introduced a shared BDD operation algo and added XOR to decide equivalence, as BDDs no longer need to be identical when equivalent, so equivalence checking that is used e.g. in the minterm algorithm now uses (a XOR b) == False to decide if a and b are equivalent. Earlier it was just pointer equality a == b, but it is not feasible in general unless all BDDs are cached and internalized indefinitely, which would be infeasible in the long run, essentially this is a tradeoff between efficiency and flexibility. 

I am considering to abandon the specialized algebras BVAlgebra/BV64Algebra in favor of just using the BDD algebra in the matcher now. This could be discussed/evaluated there are pros and cons. (There could also be a combined algebra that uses BDDS only for non-ASCII)

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -48,7 +48,6 @@ namespace System.Text.RegularExpressions.SRM
         private BDD(int ordinal)
         {
             Ordinal = ordinal;
-            //let the ordinal also be the hashcode
             hashcode = (ordinal, 0, 0).GetHashCode();
         }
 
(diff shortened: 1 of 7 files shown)


olsaarik:
As a general comment about abandoning BVAlgebra/BV64Algebra, it feels likely to me that this would make sense to do. We mostly don't expect derivative creation to dominate running time, except with small amounts of input, in which case mintermization is already contributing a lot to running time.

olsaarik (approved):

veanes:
This merge will fix the could build process for regexsrm. 
We will revisit use/avoidance of  BVAlgebra/BV64Algebra later.

PR #706, opened 2021-02-18
Build the PDB for ObjWriter

MichalStrehovsky:
In a rather confusing choice of defaults, CMake decided that "Release" builds should get no debugging symbols. Build RelWithDebInfo to get the PDBs.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
+++ b/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
@@ -33,7 +33,7 @@
   <PropertyGroup>
     <ILCompilerVersion>6.0.0-preview.1.21075.1</ILCompilerVersion>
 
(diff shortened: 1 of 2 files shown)


jkotas (approved):
Thanks!

AntonLapounov:
@MichalStrehovsky For future fixes please keep in sync Windows and Linux build scripts:
https://github.com/dotnet/runtimelab/blob/3d3b11a3e3060583e5af00c6f835c64a72d5cf10/src/coreclr/tools/aot/ObjWriter/build.sh#L6
I am working on changes allowing cross-compilation of `ObjWriter`, so I can include this fix for you.

The reason I did not change it to `RelWithDebInfo` myself was significant (41%) DLL size increase: from 18.2 MiB to 25.6 MiB.  I am wondering what effect on performance that may have.

AntonLapounov:
At present we use the default CMake configuration, which passes `/incremental` to the linker for `RelWithDebInfo` builds. That inserts padding for code and data and disables `/OPT:REF,ICF`, significantly increasing the DLL size. I think we should go back to `Release` builds and use the same set of compiler and linker switches that we use for building the runtime, e.g., include `eng\native\configurecompiler.cmake`.

MichalStrehovsky:
Yeah, if LLVM build can survive configurecompiler.cmake that would probably be the best.

PR #709, opened 2021-02-19
Add debug header for diagnostic tools

davmason:
This change adds the concept of a "Debug Header" which allows diagnostic tooling to get information about the runtime types and layout without being tightly coupled to the runtime like the DAC is for coreclr. It is a lightweight self describing metadata format focused on types and their members, globals, and build defines.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/DebugHeader.cpp
+++ b/src/coreclr/nativeaot/Runtime/DebugHeader.cpp
@@ -46,10 +46,7 @@ static GlobalValueEntry s_GlobalEntries[GlobalEntriesArraySize];
 // This structure is part of a in-memory serialization format that is used by diagnostic tools to
 // reason about the runtime. As a contract with our diagnostic tools it must be kept up-to-date
 // by changing the MajorVersion when breaking changes occur. If you are changing the runtime then
-// you are responsible for understanding what changes are breaking changes. You can do this by
-// reading the specification (Documentation\design-docs\diagnostics\ProcessMemoryFormatSpec.md)
-// to understand what promises the runtime makes to diagnostic tools. Any change that would make that
-// document become inaccurate is a breaking change.
+// you are responsible for understanding what changes are breaking changes.
 //
 // If you do want to make a breaking change please coordinate with diagnostics team as breaking changes
 // require debugger side components to be updated, and then the new versions will need to be distributed
@@ -73,8 +70,8 @@ struct DotNetRuntimeDebugHeader
     // earlier debuggers should treat the module as if it had no .Net runtime at all.
     // If the cookie is valid a debugger is safe to assume the Major/Minor version fields
     // will follow, but any contents beyond that depends on the version values.
-    // The cookie value is currently set to 0x4E 0x41 0x44 0x48 (NADH in ascii)
-    const uint8_t Cookie[4] = { 0x4E, 0x41, 0x44, 0x48 };
+    // The cookie value is currently set to 0x4E 0x41 0x44 0x48 (DNDH in ascii)
+    const uint8_t Cookie[4] = { 0x44, 0x4E, 0x44, 0x48 };
 
     // This counter can be incremented to indicate breaking changes
     // This field must be encoded little endian, regardless of the typical endianess of
@@ -103,20 +100,20 @@ struct DotNetRuntimeDebugHeader
     // the same pointer size and endianess encoding unless otherwise specified.
 
     // A pointer to an array describing important types and their offsets
-    DebugTypeEntry (*DebugTypeEntries)[DebugTypeEntriesArraySize] = nullptr;
+    DebugTypeEntry (* volatile DebugTypeEntries)[DebugTypeEntriesArraySize] = nullptr;
 
     // A pointer to an array that contains pointers to important globals
-    GlobalValueEntry (*GlobalEntries)[GlobalEntriesArraySize] = nullptr;
+    GlobalValueEntry (* volatile GlobalEntries)[GlobalEntriesArraySize] = nullptr;
 };
 
 extern "C" DotNetRuntimeDebugHeader DotNetRuntimeDebugHeader = {};
 
-#define MAKE_DEBUG_ENTRY(TypeName, FieldName, Value)                                                                        \
-    do                                                                                                                      \
-    {                                                                                                                       \
-        (*DotNetRuntimeDebugHeader.DebugTypeEntries)[currentDebugPos] = { #TypeName, #FieldName, Value, 0  };             \
-        ++currentDebugPos;                                                                                                  \
-        ASSERT(currentDebugPos <= DebugTypeEntriesArraySize);                                 \
+#define MAKE_DEBUG_ENTRY(TypeName, FieldName, Value)                             \
+    do                                                                           \
+    {                                                                            \
+        s_DebugEntries[currentDebugPos] = { #TypeName, #FieldName, Value, 0  };  \
+        ++currentDebugPos;                                                       \
+        ASSERT(currentDebugPos <= DebugTypeEntriesArraySize);                    \
     } while(0)
 
 #define MAKE_DEBUG_FIELD_ENTRY(TypeName, FieldName) MAKE_DEBUG_ENTRY(TypeName, FieldName, offsetof(TypeName, FieldName))
@@ -125,13 +122,13 @@ extern "C" DotNetRuntimeDebugHeader DotNetRuntimeDebugHeader = {};
 
 #define MAKE_SIZE_ENTRY(TypeName) MAKE_DEBUG_ENTRY(TypeName, SIZEOF, sizeof(TypeName))
 
-#define MAKE_GLOBAL_ENTRY(Name)                                                                                             \
-    do                                                                                                                      \
-    {                                                                                                                       \
-        (*DotNetRuntimeDebugHeader.GlobalEntries)[currentGlobalPos] = { #Name, Name };                                    \
-        ++currentGlobalPos;                                                                                                 \
-        ASSERT(currentGlobalPos <= GlobalEntriesArraySize)                                    \
-    } while(0)                                                                                                              \
+#define MAKE_GLOBAL_ENTRY(Name)                                                   \
+    do                                                                            \
+    {                                                                             \
+        s_GlobalEntries[currentGlobalPos] = { #Name, Name };                      \
+        ++currentGlobalPos;                                                       \
+        ASSERT(currentGlobalPos <= GlobalEntriesArraySize)                        \
+    } while(0)                                                                    \
 
 extern "C" void PopulateDebugHeaders()
 {
@@ -140,9 +137,6 @@ extern "C" void PopulateDebugHeaders()
 
     ZeroMemory(s_DebugEntries, DebugTypeEntriesArraySize);
     ZeroMemory(s_GlobalEntries, GlobalEntriesArraySize);
-
-    DotNetRuntimeDebugHeader.DebugTypeEntries = &s_DebugEntries;
-    DotNetRuntimeDebugHeader.GlobalEntries = &s_GlobalEntries;
 
     MAKE_SIZE_ENTRY(GcDacVars);
     MAKE_DEBUG_FIELD_ENTRY(GcDacVars, major_version_number);
@@ -268,12 +262,15 @@ extern "C" void PopulateDebugHeaders()
     HANDLE moduleBaseAddress = PalGetModuleHandleFromPointer((void *)&PopulateDebugHeaders);
     MAKE_GLOBAL_ENTRY(moduleBaseAddress);
 
-    static_assert(EEType::Flags::EETypeKindMask         == 0x0003, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
-    static_assert(EEType::Flags::RelatedTypeViaIATFlag  == 0x0004, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
-    static_assert(EEType::Flags::HasFinalizerFlag       == 0x0010, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
-    static_assert(EEType::Flags::HasPointersFlag        == 0x0020, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
-    static_assert(EEType::Flags::GenericVarianceFlag    == 0x0080, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
-    static_assert(EEType::Flags::IsGenericFlag          == 0x0400, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
-    static_assert(EEType::Flags::ElementTypeMask        == 0xf800, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
-    static_assert(EEType::Flags::ElementTypeShift       == 11,     "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
+    DotNetRuntimeDebugHeader.DebugTypeEntries = &s_DebugEntries;
+    DotNetRuntimeDebugHeader.GlobalEntries = &s_GlobalEntries;
+
+    static_assert(EEType::Flags::EETypeKindMask         == 0x0003, "The debugging data contract has a hard coded dependency on this value of EEType::Flags. If you change this value you must bump major_version_number.");
+    static_assert(EEType::Flags::RelatedTypeViaIATFlag  == 0x0004, "The debugging data contract has a hard coded dependency on this value of EEType::Flags. If you change this value you must bump major_version_number.");
+    static_assert(EEType::Flags::HasFinalizerFlag       == 0x0010, "The debugging data contract has a hard coded dependency on this value of EEType::Flags. If you change this value you must bump major_version_number.");
+    static_assert(EEType::Flags::HasPointersFlag        == 0x0020, "The debugging data contract has a hard coded dependency on this value of EEType::Flags. If you change this value you must bump major_version_number.");
+    static_assert(EEType::Flags::GenericVarianceFlag    == 0x0080, "The debugging data contract has a hard coded dependency on this value of EEType::Flags. If you change this value you must bump major_version_number.");
+    static_assert(EEType::Flags::IsGenericFlag          == 0x0400, "The debugging data contract has a hard coded dependency on this value of EEType::Flags. If you change this value you must bump major_version_number.");
+    static_assert(EEType::Flags::ElementTypeMask        == 0xf800, "The debugging data contract has a hard coded dependency on this value of EEType::Flags. If you change this value you must bump major_version_number.");
+    static_assert(EEType::Flags::ElementTypeShift       == 11,     "The debugging data contract has a hard coded dependency on this value of EEType::Flags. If you change this value you must bump major_version_number.");
 }


src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+    DefineEntry *DefinesList = nullptr;
+};
+
+extern "C" NativeAOTRuntimeDebugHeader g_NativeAOTRuntimeDebugHeader = {};
+
+#define MAKE_DEBUG_ENTRY(TypeName, FieldName, Value)                                                                        \
+    do                                                                                                                      \
+    {                                                                                                                       \
+        currentType = new (nothrow) DebugTypeEntry{ previousType, #TypeName, #FieldName, Value };                             \
```
MichalStrehovsky, line 115: I'm wondering if we could pre-initialize most of these in the data region. The extra `new`s are probably not measurably expensive, but if we can avoid doing doing extra work at startup, it's better to avoid it.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,275 @@
...
+    MAKE_DEBUG_FIELD_ENTRY(ThreadBuffer, m_pThreadStressLog);
+
+    // EEThreadID is forward declared and not available
+    MAKE_DEBUG_ENTRY(EEThreadID, SIZEOF, sizeof(void*));
+    MAKE_DEBUG_ENTRY(EEThreadID, m_FiberPtrId, 0);
+
+    MAKE_SIZE_ENTRY(EEType);
+    MAKE_DEBUG_FIELD_ENTRY(EEType, m_uBaseSize);
+    MAKE_DEBUG_FIELD_ENTRY(EEType, m_usComponentSize);
+    MAKE_DEBUG_FIELD_ENTRY(EEType, m_usFlags);    MAKE_DEBUG_ENTRY(EEType, m_pBaseType, offsetof(EEType, m_RelatedType) + offsetof(EEType::RelatedTypeUnion, m_pBaseType));
```
MichalStrehovsky, line 198: It would be good to capture somewhere which flags the debugger is going to depend on. We have been changing these in the past.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
jkotas: +1 I was about to write the same feedback.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
davmason: The flags are used for !verifyheap, the logic is copied from [EEType::Validate](https://github.com/dotnet/runtimelab/blob/99dbab06575ab15fac0fbfd53df1ecd863914fbe/src/coreclr/nativeaot/Runtime/eetype.cpp#L19) for object validation. 

I assumed that the flags were stable, I can add some static asserts though that check that they change and instruct to bump the debug header version if they are different.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
davmason: I originally intended to have it be all at compile time. The data is just offsets and build defines so it is available at  compile time.  I spent some time a while back trying to get it to work with constexpr functions, but the main issue I kept running in to was how to get a dynamically sized array that works with constexpr. I suppose I could have a statically sized array and just have to keep it in sync with the number of entries.

I can take a look at it again and see if I can come up with something. If there's any prior art I could follow please let me know.

davmason: I wasn't able to come up with a way to embed them at compile time like I would like, but I at least got rid of the allocations. Now there is an array and at runtime the values are assigned to the array without allocation.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
davmason: Asserts are added now

jkotas:
Could you please take a look at the build breaks?

davmason:
Build breaks are fixed and I tested the SOS changed on Linux as well.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,294 @@
...
+    
+    // Reserved - Currently it only serves as alignment padding for the pointers which
+    // follow but future usage will be considered a back-compatible change.
+    const uint32_t ReservedPadding3 = 0;
+
+    // An array that contains pointers to important globals
+    GlobalValueEntry GlobalEntries[GLOBALS_ARRAY_SIZE];
+};
+
+extern "C" NativeAOTRuntimeDebugHeader g_NativeAOTRuntimeDebugHeader = {};
```
jkotas, line 118: Would it make sense to have non-NativeAOT specific name for this? E.g. `DotNetRuntimeDebugHeader`.

I would love to see this unified accross all runtime variants eventually.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,294 @@
...
+    MAKE_DEBUG_FIELD_ENTRY(StressMsg, fmtOffsCArgs);
+    MAKE_DEBUG_FIELD_ENTRY(StressMsg, facility);
+    MAKE_DEBUG_FIELD_ENTRY(StressMsg, timeStamp);
+    MAKE_DEBUG_FIELD_ENTRY(StressMsg, args);
+
+    MAKE_SIZE_ENTRY(Object);
+    MAKE_DEBUG_FIELD_ENTRY(Object, m_pEEType);
+
+    MAKE_SIZE_ENTRY(Array);
+    MAKE_DEBUG_FIELD_ENTRY(Array, m_Length);
```
jkotas, line 264: I do not think we need to have the System.Object and array layout configurable. They are pretty much set in stone.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    MAKE_SIZE_ENTRY(ThreadBuffer);
+    MAKE_DEBUG_FIELD_ENTRY(ThreadBuffer, m_pNext);
+    MAKE_DEBUG_FIELD_ENTRY(ThreadBuffer, m_rgbAllocContextBuffer);
+    MAKE_DEBUG_FIELD_ENTRY(ThreadBuffer, m_threadId);
+    MAKE_DEBUG_FIELD_ENTRY(ThreadBuffer, m_pThreadStressLog);
+
+    // EEThreadID is forward declared and not available
+    MAKE_DEBUG_ENTRY(EEThreadID, SIZEOF, sizeof(void*));
+    MAKE_DEBUG_ENTRY(EEThreadID, m_FiberPtrId, 0);
```
jkotas, line 206: Is this right for non-Windows?

jkotas (approved):
LGTM otherwise. Thanks!

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,294 @@
...
+    // The length of the DebugTypeEntries array
+    #define DEBUG_TYPES_ARRAY_SIZE 101
+    const uint32_t DebugTypeEntriesArraySize = DEBUG_TYPES_ARRAY_SIZE;
+
+    // Reserved - Currently it only serves as alignment padding for the pointers which
+    // follow but future usage will be considered a back-compatible change.
+    const uint32_t ReservedPadding2 = 0;
+
+    // An array describing important types and their offsets
+    DebugTypeEntry DebugTypeEntries[DEBUG_TYPES_ARRAY_SIZE];
```
MichalStrehovsky, line 104: I'm wondering whether it would be useful to have a little bit more extensibility in this structure. The previous design had some pointer-typed fields that allowed to do things with it.

I was thinking that if we're ever in a situation where we need to capture information in this structure that is not known at the time of compiling the native runtime bits, but becomes known when we do the AOT compilation of the managed code, we could have the AOT compiler generate the necessary part of the data structure. But with DebugHeader in this shape, there's no way to append those records to this structure (we would have to reformat it at startup).

If instead of storing these entries inline in the header, we put a pointer here to the inline array (still stored statically), and if instead of storing the count here, we would have a convention that the array ends with a single all-null entry, this would allow enough extensibility that we can append to this list with the platform linker (i.e. the compiler can generate additional entries and the linker would lay it out in the appropriate spot so that it shows up continuously in the image).

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,294 @@
...
+    static_assert(EEType::Flags::HasFinalizerFlag       == 0x0010, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
+    static_assert(EEType::Flags::HasPointersFlag        == 0x0020, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
+    static_assert(EEType::Flags::GenericVarianceFlag    == 0x0080, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
+    static_assert(EEType::Flags::OptionalFieldsFlag     == 0x0100, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
+    static_assert(EEType::Flags::IsGenericFlag          == 0x0400, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
+    static_assert(EEType::Flags::ElementTypeMask        == 0xf800, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
```
MichalStrehovsky, line 292: Do we really have a dependency on `IsDynamicType` and `OptionalFieldsFlag`?

src/coreclr/nativeaot/Bootstrap/main.cpp
```diff
@@ -164,6 +166,9 @@ static int InitializeRuntime()
 
     InitializeModules(osModule, __modules_a, (int)((__modules_z - __modules_a)), (void **)&c_classlibFunctions, _countof(c_classlibFunctions));
 
+    // Populate the values needed for debugging
```
MichalStrehovsky, line 169: Wondering if we could just call this from `RhInitialize`.

MichalStrehovsky (approved):
Thank you!

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
jkotas: `g_gcDacGlobals` in the current list demonstrates similar problem as the AOT compiler generated records.
 
I think that this flat list may over time evolve into a list of individually versioned contracts that are defined locally by each component (StressLog, Threading, GC, AOT compiler/typesystem, ...).

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
davmason: I also would like to see this approach unified across all the runtimes, I was originally thinking that they would have their own contracts but there's no reason not to have the same name for the type of the contract itself.

I would like to keep the export name as "g_NativeAOTRuntimeDebugHeader" though so we can differentiate runtimes by just checking the export name. So then coreclr can have g_CoreCLRRuntimeDebugHeader, etc.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
davmason: I thought it was. In coreclr\vm\util.hpp EEThreadId just has the one member of m_FiberPtrId.

It doesn't really matter for this work anyways, none of the commands need to access EEThreadID internals. I just need to read from ThreadBuffer and can delete these entries

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
davmason: No, I was just on auto pilot here

src/coreclr/nativeaot/Bootstrap/main.cpp
davmason: Sure!

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
jkotas: We like to talk about NativeAOT project as a flavor of CoreCLR runtime. The place we want to get to is there is a continuum of runtime configurations to choose from, not distinct runtimes.

davmason: Is the concern over API surface area? For a tool like SOS which is inspecting runtime level data structures we will have to be aware which runtime flavor we are running against for things like whether to look up the EEType in the symbols file, or to use CLR metadata to get type information.

I'm happy to change it to g_DotNetRuntimeDebugHeader, but then I would have to either differentiate on the Cookie or some define we put in the contract. That doesn't feel meaningfully different to me other than the API surface.

jkotas: The places that use this should differentiate on contract type, not based on runtime type.

For example, stresslog or GC contract is same between all runtimes, no need to differentiate today.

For type system contract, there are multiple choices - e.g. try JIT-based type system contract first, and if it is not available try no-JIT AOT-based contract second.

src/coreclr/nativeaot/Runtime/ObjectLayout.h
```diff
@@ -37,9 +37,13 @@ typedef DPTR(class EEType) PTR_EEType;
 class MethodTable;
 
 //-------------------------------------------------------------------------------------------------
+
+extern "C" void PopulateDebugHeaders();
```
MichalStrehovsky, line 41: It doesn't look like the changes to this file are necessary anymore.

src/coreclr/nativeaot/Runtime/threadstore.cpp
```diff
@@ -542,4 +542,5 @@ Boolean ThreadStore::GetExceptionsForCurrentThread(Array* pOutputArray, int32_t*
     *pWrittenCountOut = countWritten;
     return Boolean_false;
 }
+
```
MichalStrehovsky, line 545: Nit: unnecessary whitespace change

MichalStrehovsky (approved):
Thank you!

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -281,6 +281,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Include="-o &quot;$(NativeBinary)&quot;" Condition="'$(TargetOS)' != 'windows'" />
       <CustomLinkerArg Include="/OUT:&quot;$(NativeBinary)&quot;" Condition="'$(TargetOS)' == 'windows'" />
       <CustomLinkerArg Include="/DEF:&quot;$(ExportsFile)&quot;" Condition="'$(TargetOS)' == 'windows' and $(ExportsFile) != ''" />
+      <CustomLinkerArg Include="/EXPORT:g_DotNetRuntimeDebugHeader,DATA" Condition="'$(OS)' == 'Windows_NT'" />
```
MichalStrehovsky, line 284: Btw, would the debugger be able to find this symbol in the PDB? Do we need to export it?

If it's exported, a lot of people are going to see it so it might be "nicer" to go with Jan's original suggestion of naming it `DotNetRuntimeDebugHeader` without the `g_`. If we're expecting to keep this as our debugging future, we might as well make it not look like an accidentally exported implementation detail. It's more of a nit of course.

davmason: If it wasn't exported I think the debugger could find it, but there are other tools outside of a debugger that we would want to support, such as the dotnet-dump tool. It's a lot easier to find as an export than having to be able to load a symbol file. I removed the "g_" from the name.

src/coreclr/nativeaot/Runtime/threadstore.cpp
davmason: Removed!

src/coreclr/nativeaot/Runtime/ObjectLayout.h
davmason: Removed those too

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,279 @@
...
+struct DotNetRuntimeDebugHeader
+{
+    // The cookie serves as a sanity check against process corruption or being requested
+    // to treat some other non-.Net module as though it did contain the coreRT runtime.
+    // It can also be changed if we want to make a breaking change so drastic that
+    // earlier debuggers should treat the module as if it had no .Net runtime at all.
+    // If the cookie is valid a debugger is safe to assume the Major/Minor version fields
+    // will follow, but any contents beyond that depends on the version values.
+    // The cookie value is currently set to 0x4E 0x41 0x44 0x48 (NADH in ascii)
+    const uint8_t Cookie[4] = { 0x4E, 0x41, 0x44, 0x48 };
```
noahfalk, line 77: Nit: Does NADH stand for 'Native AOT Debug Header'? If we changed the name elsewhere maybe this should change too for naming consistency.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+    void *g_stressLog = &StressLog::theLog;
+    MAKE_GLOBAL_ENTRY(g_stressLog);
+
+    // Some DAC functions need to know the module base address, easiest way is with
+    // the HANDLE to our module which is the base address.
+    HANDLE moduleBaseAddress = PalGetModuleHandleFromPointer((void *)&PopulateDebugHeaders);
+    MAKE_GLOBAL_ENTRY(moduleBaseAddress);
+
+    static_assert(EEType::Flags::EETypeKindMask         == 0x0003, "SOS has a hard coded dependency on the values of EEType::Flags. If you change these values you must bump major_version_number.");
```
noahfalk, line 271: Nit: I would replace 'SOS' with a more generic 'Debugging data contracts'. SOS may be the first user of this but it probably won't be the last.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,279 @@
...
+// This size should be one bigger than the number of entries since a null entry
+// signifies the end of the array.
+static constexpr size_t GlobalEntriesArraySize = 6;
+static GlobalValueEntry s_GlobalEntries[GlobalEntriesArraySize];
+
+// This structure is part of a in-memory serialization format that is used by diagnostic tools to
+// reason about the runtime. As a contract with our diagnostic tools it must be kept up-to-date
+// by changing the MajorVersion when breaking changes occur. If you are changing the runtime then
+// you are responsible for understanding what changes are breaking changes. You can do this by
+// reading the specification (Documentation\design-docs\diagnostics\ProcessMemoryFormatSpec.md)
```
noahfalk, line 50: Lacking the doc we should probably remove this part of the comment : )

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -0,0 +1,279 @@
...
+extern "C" void PopulateDebugHeaders()
+{
+    size_t currentDebugPos = 0;
+    size_t currentGlobalPos = 0;
+
+    ZeroMemory(s_DebugEntries, DebugTypeEntriesArraySize);
+    ZeroMemory(s_GlobalEntries, GlobalEntriesArraySize);
+
+    DotNetRuntimeDebugHeader.DebugTypeEntries = &s_DebugEntries;
+    DotNetRuntimeDebugHeader.GlobalEntries = &s_GlobalEntries;
```
noahfalk, line 145: We could get rare race conditions if we publish the pointer to the table before it is initialized. Defining these pointers as volatile and setting them after the table is initialized should eliminate that possibility.

noahfalk (approved):
A few small things but overall LGTM!

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
davmason: That is definitely what it stands for 😄. I can change it to DNDH for DotNetDebugHeader.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
davmason: Sure, I can make that change

davmason:
Thanks everybody for reviewing. At this point I believe all feedback is accounted for and I am planning on merging later tonight.

PR #723, opened 2021-02-20
ObjWriter fixes

MichalStrehovsky:
* Don't use x86 NOP to pad non-text sections. LLVM asserts if we try to do this in e.g. the `.bss` section
* Grab type index only after the `IsStatic` check. The `IsStatic` check can put a new thing in the `DwarfTypes` table and the `Index` will no longer be what we think it is. `static_cast` fun ensues (`static_cast` should be a punishable offense).


jkotas (approved):
Thanks!

PR #725, opened 2021-02-21
Fix debug build of ObjWriter

### LINKED ISSUE #708: Cannot build Debug version of ObjWriter
The Debug build of ObjWriter fails with llvm-tblgen.exe complaining that it doesn't understand the `-gen-attrs` parameter.

I spent 3 minutes debugging it, but command line argument parsing is done using some C++ bs instead of `if` checks that I didn't have the nerves to untangle.

Restoring the `  -DLLVM_OPTIMIZED_TABLEGEN=1 ^` line in build.cmd fixed the problem for me.

Cc @AntonLapounov in case you saw this.

MichalStrehovsky:
Backporting an LLVM fix.

Fixes #708.

### DIFF
--- a/src/coreclr/tools/aot/ObjWriter/llvm.patch
+++ b/src/coreclr/tools/aot/ObjWriter/llvm.patch
@@ -48,6 +48,53 @@ index 5390e794242..5b258e76787 100644
    /// Functionally identical to EmitBytes. When emitting textual assembly, this
    /// method uses .byte directives instead of .ascii or .asciz for readability.
    virtual void EmitBinaryData(StringRef Data);
+diff --git a/include/llvm/Support/ManagedStatic.h b/include/llvm/Support/ManagedStatic.h
+index b4bf3210cc7..f7f7cbf781a 100644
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ObjWriter/llvm.patch
```diff
@@ -1,5 +1,5 @@
 diff --git a/include/llvm/BinaryFormat/COFF.h b/include/llvm/BinaryFormat/COFF.h
-index b395db6eaa8..74820a8f79f 100644
```
MichalStrehovsky, line 2: I don't know why these changed. I ran `git diff --output c:\git\runtime\src\coreclr\tools\aot\ObjWriter\llvm.patch`

kant2002: you may want to try `--abbrev=11`

MichalStrehovsky: Whoa, magic! Thanks!

AntonLapounov (approved):

PR #726, opened 2021-02-21
Cross reference NativeAOT related docs

kant2002:
I will help reader find other NativeAOT pieces faster


docs/using-nativeaot/README.md
```diff
@@ -1,5 +1,6 @@
 # Using Native AOT
 
+- [Building](../workflow/building/coreclr/nativeaot.md)
```
jkotas, line 3: People will get to this doc by clicking on "how to debug, optimize and troubleshoot applications published using the native AOT toolchain" in the readme.

How to build the repo is not relevant for that.

kant2002:
Because NativeAot is folder, I found that location more often. And original
location with build instructions I sometimes even cannot find at all.
That’s actually the reason why I decide to provide hints to other locations
with related information

jkotas:
> Because NativeAot is folder, I found that location more often.

The file you are referencing has nativeaot in the name as well...

We have intentional split between "how to build the repo" and "how to use the toolchain" documentation. When we had these types of documentations intermixed in the past, we had seen number of cases where people started going through the pain of building the repo to try it out, but all they actually needed to do was to just add the package reference.

kant2002:
Now I understand why I see that gap. I may have to think of there is a way
how his gap can be crossed without increasing support costs for you guys.

jkotas:
Closing this for now

PR #729, opened 2021-02-22
Cleanup any leftover native artifacts from tests

MichalStrehovsky:
Matches what we do on Windows. Makes sure we don't re-use existing native bits (incremental compile in MSBuild would make AOT compilation a no-op).


jkotas (approved):
LGTM. Thanks!

PR #730, opened 2021-02-22
Hide symbols instead of making them local

MichalStrehovsky:
This article made me understand what we need to do here: https://www.technovelty.org/code/why-symbol-visibility-is-good.html

We made symbols local by default in https://github.com/dotnet/corert/pull/7249 to fix problems building shared libraries. What we actually should have done was to keep them global, but make them hidden.


MichalStrehovsky:
Oh, right. The Debug/Checked builds in the CI don't build the object writer. I guess I could make this change backwards compatible, but maybe it's not worth the hassle of splitting this.

jkotas (approved):
Thanks

PR #731, opened 2021-02-22
Fix building packages

MichalStrehovsky:
We now have "an official" way to do this.

Cc @yowl - similar change might be needed in the LLVM branch.


MichalStrehovsky:
Multifile failures are expected because the build is broken until we can refresh objwriter. Can't refresh objwriter without fixing packaging.

PR #735, opened 2021-02-23
Add information about state machine MoveNext methods

MichalStrehovsky:
This exposes the debug records necessary to identify state machine MoveNext methods. Only implementing this for the portable PDB reader. It didn't look straightforward to do for unmanaged PDBs but those are Windows only anyway.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
@@ -41,6 +41,121 @@ namespace ILCompiler
             }
         }
 
+        public uint GetStateMachineThisVariableTypeIndex(TypeDesc type)
+        {
+            // If the state machine is a valuetype, the this parameter will be a byref
+            if (type.IsByRef)
+            {
+                type = type.GetParameterType();
+                Debug.Assert(type.IsValueType);
+            }
+
+            type = DebuggerCanonicalize(type);
+
+            lock (_lock)
+            {
+                if (!_knownStateMachineThisTypes.TryGetValue(type, out uint typeIndex))
+                {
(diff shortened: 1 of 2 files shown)


MichalStrehovsky:
Decided to also push out the commit that adds demangling. Not sure if I'll be able to do the deluxe version that actually shows these as locals since the C++ EE might not be involved after all. Still checking.

The current experience is:

image

MichalStrehovsky:
The debug information for the state machine method itself is left intact:

image

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
```diff
@@ -41,6 +41,110 @@ public uint GetVariableTypeIndex(TypeDesc type)
...
+
+                    foreach (var fieldDesc in defType.GetFields())
+                    {
+                        if (fieldDesc.IsStatic)
+                            continue;
+
+                        string fieldNameEmit = fieldDesc.Name;
+                        if (fieldNameEmit.Length > 0 && fieldNameEmit[0] == '<')
+                        {
+                            if (TryGetGeneratedNameKind(fieldNameEmit, out char kind))
```
jkotas, line 74: I assume that this logic is copied from somewhere, or at least inspired by some code in Roslyn (?). Add a comment with a link for where this came from?

jkotas (approved):
Nice!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
MichalStrehovsky: Ah yes. I had such comment but it looks like it fell victim to when I was updating this. Fixed.

PR #738, opened 2021-02-24
Enable ARM64 builds in CI

AntonLapounov:
* Enable Windows and Linux ARM64 builds in CI and disable tests for them.
* Fix a race condition for `Runtime` and `Runtime.ServerGC` libraries, which preprocessed and compiled assembler helpers to the same directory during Windows ARM64 builds. (You might see intermittent `Cannot open compiler generated file` errors.) Also remove dead code at the end of the `nativeaot/Runtime/Full/CMakeLists.txt` file.
* Enable cross-building Windows ARM64-hosted `objwriter.dll`. To make it work, I had to introduce architecture-specific subdirectories under `artifacts\llvm\build`. Building ARM64-hosted `objwriter.dll` requires to build an x64-hosted `llvm-tblgen.exe` first.
* Make incremental builds of `objwriter` much faster. Before the build, save the current diff of the LLVM repo to a file, so you do not accidentally lose your changes. If the current diff is the same as the `llvm.patch`, skip discarding the current changes and reapplying the patch.
* The length of the hash in `llvm.patch` was different for deep and shallow clones of LLVM. Use the `--full-index` option to always use the full length. (In _very_ recent Git versions (>= 2.29, Q4 2020) it is possible to use the `--abbrev[=<n>]` option to control the hash length in diff-patch output format.)
* Enable publishing Windows ARM64-hosted `objwriter.dll` in `ILCompiler.csproj`. We will have to slightly update that file again once there is an official Windows ARM64 package.
* Switching to `RelWithDebInfo` builds in #706 increased the size of the `objwriter.dll` by about 40% due to CMake passing the `/incremental` option to the linker, which inserts padding for code and data and disables `/OPT:REF,ICF` optimizations. `RelWithDebInfo` builds also disable some compiler optimizations by default. In this PR I switch back to `Release` builds and introduce a custom CMake toolchain file to alter default C/CXX/linker flags to enable debug information.
* Fix `comparison between ‘enum ICorDebugInfo::RegNum’ and ‘enum Arm64/etc.::ICorDebugInfo::RegNum’` warnings in `objwriter.cpp` on Linux. Technically we should cast to the proper `enum` type; however, that would not help readability, so I cast to `int` to suppress the warnings.
* Quick and dirty fix for a Windows ARM64 build break in LLVM repo (`lib/Support/Windows/Signals.inc`). That stack dumping code would require real changes to work on Windows ARM64; hopefully we will not need it to work.
* Fix minor Linux ARM64 build breaks in `gcinfodecoder.cpp`.
* Unfortunately I was unable to cross-build Linux ARM64-hosted `libobjwriter.so` with Docker images currently used by CI. In particular, LLVM complained on missing `Threads` package and `std::atomic` support. For that reason I did not include changes required for cross-building `objwriter` on Linux.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/Full/CMakeLists.txt
+++ b/src/coreclr/nativeaot/Runtime/Full/CMakeLists.txt
@@ -8,27 +8,23 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
 add_definitions(-DFEATURE_RX_THUNKS)
 
-# For Windows ARM/ARM64 .asm files in this list are replaced with the corresponding .obj files
-set(RUNTIME_SOURCES_ARCH_ASM_RESOLVED "${RUNTIME_SOURCES_ARCH_ASM}")
-
 if (CLR_CMAKE_TARGET_WIN32)
   if (CLR_CMAKE_HOST_ARCH_ARM OR CLR_CMAKE_HOST_ARCH_ARM64)
     # Needed to include AsmOffsets.inc
     include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})
     preprocess_files(RUNTIME_SOURCES_ARCH_ASM ${RUNTIME_SOURCES_ARCH_ASM})
-    set(RUNTIME_SOURCES_ARCH_ASM_RESOLVED "${RUNTIME_SOURCES_ARCH_ASM}")
     if (CMAKE_GENERATOR MATCHES "Visual Studio")
-      compile_asm(TARGET Runtime ASM_FILES ${RUNTIME_SOURCES_ARCH_ASM} OUTPUT_OBJECTS RUNTIME_SOURCES_ARCH_ASM_RESOLVED)
+      # Replaces .asm files in RUNTIME_SOURCES_ARCH_ASM with the corresponding .obj files
+      compile_asm(TARGET Runtime ASM_FILES ${RUNTIME_SOURCES_ARCH_ASM} OUTPUT_OBJECTS RUNTIME_SOURCES_ARCH_ASM)
     endif()
   endif()
 endif (CLR_CMAKE_TARGET_WIN32)
 
-add_library(Runtime STATIC ${COMMON_RUNTIME_SOURCES} ${FULL_RUNTIME_SOURCES} ${RUNTIME_SOURCES_ARCH_ASM_RESOLVED})
+add_library(Runtime STATIC ${COMMON_RUNTIME_SOURCES} ${FULL_RUNTIME_SOURCES} ${RUNTIME_SOURCES_ARCH_ASM})
 
-add_library(Runtime.ServerGC STATIC ${COMMON_RUNTIME_SOURCES} ${FULL_RUNTIME_SOURCES} ${RUNTIME_SOURCES_ARCH_ASM_RESOLVED} ${SERVER_GC_SOURCES})
+add_library(Runtime.ServerGC STATIC ${COMMON_RUNTIME_SOURCES} ${FULL_RUNTIME_SOURCES} ${RUNTIME_SOURCES_ARCH_ASM} ${SERVER_GC_SOURCES})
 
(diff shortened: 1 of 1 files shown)


MichalStrehovsky:
> This version of D:\workspace\_work\1\s\artifacts\tests\coreclr\windows.arm64.Release\Tests\Core_Root\nativeaot\tools\ilc.exe is not compatible with the version of Windows you're running. Check your computer's system information and then contact the software publisher.
D:\workspace\_work\1\s\artifacts\tests\coreclr\windows.arm64.Release\Tests\Core_Root\nativeaot\build\Microsoft.NETCore.Native.targets(247,5): error MSB3073: The command ""D:\workspace\_work\1\s\artifacts\tests\coreclr\windows.arm64.Release\Tests\Core_Root/nativeaot\tools\ilc" @"D:\workspace\_work\1\s\artifacts\obj\/NativeAOTFX\CommandLine.ilc.rsp"" exited with code 216. [D:\workspace\_work\1\s\artifacts\tests\coreclr\windows.arm64.Release\Tests\Core_Root\nativeaot\build\BuildFrameworkNativeObjects.proj]

We could disable the multimodule testing on ARM64 (it's the mode where each managed assembly gets compiled to individual OBJ files and we use platform linker to merge them together). This mode is not a shipping scenario and we just test it so that it doesn't completely bitrot away. The testing currently only happens on Windows anyway.

The multimodule testing is selected in eng\pipelines\runtimelab\runtimelab-post-build-steps.yml

src/coreclr/vm/gcinfodecoder.cpp
```diff
@@ -1706,7 +1706,7 @@ void GcInfoDecoder::ReportRegisterToGC( // ARM64
     pCallBack(hCallBack, pObjRef, gcFlags DAC_ARG(DacSlotLocation(regNum, 0, false)));
 }
 
-#ifdef TARGET_UNIX
+#if defined(TARGET_UNIX) && !defined(FEATURE_REDHAWK)
```
jkotas, line 1709: Nit: Could you please make the same change on ARM version of this function too? (It would be needed if we ever enabled ARM.)

src/coreclr/nativeaot/Runtime/Full/CMakeLists.txt
```diff
@@ -57,14 +61,17 @@ add_custom_command(
...
+  DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/AsmOffsets.inc" "${RUNTIME_SOURCES_ARCH_ASM_RESOLVED}"
 )
 
-# The AsmOffsets.inc will be built as a dependency of the Runtime
-add_dependencies(Runtime AsmOffsets)
-add_dependencies(Runtime.ServerGC AsmOffsets)
+add_dependencies(Runtime RuntimeAsmHelpers)
+add_dependencies(Runtime.ServerGC RuntimeAsmHelpers)
+
+# REVIEW: COMPILE_DEFINITIONS_LIST is not defined here. Should we remove the following code?
```
jkotas, line 74: Does adding defines for the ASM work without this, or is it completely broken right now?

Either way, it sounds good to me to clean it up.

jkotas (approved):
Thank you!

src/coreclr/nativeaot/Runtime/Full/CMakeLists.txt
AntonLapounov: We set all required preprocessor definitions here:
https://github.com/dotnet/runtimelab/blob/248949a1ecdfa34c89c2832b0c273c849e46ab55/src/coreclr/nativeaot/Runtime/CMakeLists.txt#L233

The code in this file is missing
```CMake
get_directory_property(COMPILE_DEFINITIONS_LIST COMPILE_DEFINITIONS)
```
However, even if we add that line, the code would still do nothing, because it does not handle complex conditions that we use, e.g., `$<$<OR:$<CONFIG:DEBUG>,$<CONFIG:CHECKED>>:DEBUG;_DEBUG;...>`.  I am going to remove it.

PR #740, opened 2021-02-24
Allow StartupCodeHelpers to be missing

MichalStrehovsky:
Use the new `GetOptionalHelperType` to grab the type and propagate the null.

I was using the minimal mode of the compiler to experiment with debug records and I don't like looking at the extra type.


jkotas (approved):
LGTM. Thanks!

Issue #739, opened 2021-02-24
[NativeAOT] Would it be possible to make the recent debug header export optional?


JoshuaWierenga:
I am currently working on a program with /SUBSYSTEM:EFI_APPLICATION and the recent debug header commit adds a new /EXPORT argument to the linker but that is not supported on this subsystem. I did not just pull request a fix since perhaps there is an existing solution since the export is already conditional. is there a different OS variable for cases like this? In that case, does that change the build process since changing OS already requires DisableUnsupportedError to avoid the cross-compilation error which feels overkill and might just lead to more issues?

MichalStrehovsky:
Would it help if we lump it together with the rest of debugging support and make it also conditional on NativeDebugSymbols? Then you could opt out by disabling debugging support.

https://github.com/dotnet/runtimelab/blob/168e3047c487140d4e04795d618b1c0defc33c19/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L26

JoshuaWierenga:
That would probably work. There is no debugging support currently anyway
and while I believe uefi does support it, it would use it's own system that
most likely does not use any of the dotnet debug symbols.

MichalStrehovsky:
Fixed in #743

JoshuaWierenga:
Thanks

PR #743, opened 2021-02-25
Optional Debug Headers (#739)

JoshuaWierenga:
Fix for building code for platforms without linker /EXPORT support. Tried just changing OS target to use the existing condition but that lead to errors regarding cross compilation so @MichalStrehovsky suggested reusing NativeDebugSymbols to selectively disable the debug headers.


MichalStrehovsky:
@davmason Any objection to making this conditional on generating debug info in general?

MichalStrehovsky (approved):

davmason:
What do we gain by making it debug only?

Additionally I would expect there is a use case for SOS on release builds. Things like !dumpheap, !verifyheap and !dumpasync shouldn’t have any degraded functionality on release builds.

What happens when a NativeAOT compiled app crashes in production? Without the headers you wouldn’t be able to run SOS commands to diagnose it.

davmason:
Wait, this isn’t about release or debug is it? It’s just about whether we are including symbols in the build. Without symbols the SOS commands won’t do anything useful. 

As long as it still is included with release builds when symbols are present then it’s ok by me.

MichalStrehovsky:
> Wait, this isn’t about release or debug is it? It’s just about whether we are including symbols in the build. Without symbols the SOS commands won’t do anything useful.

Yup, that's what this is about. Joshua was looking for a way to opt out when targeting UEFI boot applications (#739) and this seemed like the most natural way to accomplish this (I really don't want to add UEFI special casing into our targets).

davmason (approved):

PR #744, opened 2021-02-25
Speed up debug var info generation

MichalStrehovsky:
* Translation went through several `Dictionary<,>` for no good reason. Got rid of those and directly indexing into arrays instead.
* The representation within the compiler was not very memory efficient so compressed things a little.
* Moved the actual translation of the debug data from managed terms to native terms (that expands the data) to the object writing phase.

Improves wallclock throughput by about 1%. Improves total allocated GC bytes by about 1%.

One disadvantage of this is that we were doing the translation of managed to native debug data in the multithreaded phase and now we're doing it in the single threaded object writing phase. But eventually, I would like us to make object writing multithreaded as well. Debug info generation can be independent of the rest of object writing and could happen on a separate thread.

I would like to also change the ObjWriter interface to use the newly introduced struct instead of NativeVarInfo so that we don't need the extra translation, but that's an interface breaking change I would prefer to do separately.


jkotas (approved):
Thank you!

PR #746, opened 2021-02-25
Print more useful information on GVM resolution failure

MichalStrehovsky:
Looks like we have multiple fail points for this, each with different diagnosability. Not sure why and it doesn't look like a puzzle I would want to solve right now. Matching the error format with what we have in TypeLoaderEnvironment.GVMResolution.cs. This one will use the LastResortString instead of raw EEType address. Might have more useful info.

Hitting this is a compiler bug in any case, but the message can help with workarounds.


jkotas (approved):

jkotas:
> Hitting this is a compiler bug in any case

Do we want to mention it in the message, so that the bug gets reported and we have a chance to fix it?

Issue #733, opened 2021-02-23
Issues while using efcore with NativeAOT


hez2010:
After efcore fixed infinite recursive generics, I hit this exception while trying to use efcore daily build with SQLite using NativeAOT:

```
      System.InvalidOperationException: Operation is not valid due to the current state of the object.
         at System.SharedTypeExtensions.GetRequiredRuntimeMethod(Type, String, Type[]) + 0x57
         at Microsoft.EntityFrameworkCore.Sqlite.Query.Internal.SqliteDateTimeAddTranslator..ctor(ISqlExpressionFactory) + 0xd2
         at Microsoft.EntityFrameworkCore.Sqlite.Query.Internal.SqliteMethodCallTranslatorProvider..ctor(RelationalMethodCallTranslatorProviderDependencies) + 0xb0
```

A simple repro project:

[efaottest.zip](https://github.com/dotnet/runtimelab/files/6026897/efaottest.zip)

MichalStrehovsky:
Looking at the code, we probably need RD.XML to keep all methods on DateTime:

https://github.com/dotnet/efcore/blob/da00fb69d615fa22a83dfee2077ad31b7bd15823/src/EFCore.Sqlite.Core/Query/Internal/SqliteDateTimeAddTranslator.cs#L33-L38

Hopefully these places will be fixed in EF as they're working on https://github.com/dotnet/efcore/issues/21894.

If this reflection wasn't wrapped in the GetRequiredRuntimeMethod helper, the compiler would actually figure this out on it's own (`typeof(DateTime).GetMethod("Foo", ...)` is analyzed by the compiler).

MichalStrehovsky:
You can add `<SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>` to a PropertyGroup of your CSPROJ and the compiler will warn whenever there's reflection that can't be analyzed. I expect you'll see a warning for `GetRequiredRuntimeMethod`. It produces a lot of noise though.

hez2010:
Thanks @MichalStrehovsky .
After some investigation, I hit this but don't know how to deal it with `rd.xml`:

```
Generic virtual method pointer lookup failure.

Declaring type handle: EEType:0x00007FF70B898D40
Target type handle: EEType:0x00007FF70B3632D8
Method name: CreateGeneric
Instantiation:
  Argument 00000000: EEType:0x00007FF70B3845A0
[...]

PR #749, opened 2021-02-26
Fix how root provider resolve generic methods

hez2010:
- There may be multiple candidate generic methods available, ilc shouldn't throw exceptions when it meets the first unmatched method.
- To improve diagnosability I also append generic parameters to failure output.
- Matching methods containing signature variables

Contributes to #733

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler/RdXmlRootProvider.cs
+++ b/src/coreclr/tools/aot/ILCompiler/RdXmlRootProvider.cs
@@ -153,97 +153,17 @@ namespace ILCompiler
                 }
             }
 
-            static bool InstantiationMatches(TypeDesc type, TypeDesc parameter, List<TypeDesc> instArgs)
-            {
-                if (!type.ContainsSignatureVariables())
-                {
-                    return type == parameter;
-                }
-
-                switch (type.Category)
-                {
-                    case TypeFlags.Array:
-                    case TypeFlags.SzArray:
-                    case TypeFlags.ByRef:
-                    case TypeFlags.Pointer:
-                        if (parameter is not ParameterizedType)
-                        {
-                            return false;
-                        }
-
-                        ParameterizedType parameterizedType = (ParameterizedType)type;
-                        ParameterizedType parameterizedParam = (ParameterizedType)parameter;
-                        return InstantiationMatches(parameterizedType.ParameterType, parameterizedParam.ParameterType, instArgs);
-
-                    case TypeFlags.FunctionPointer:
-                        if (parameter is not FunctionPointerType)
-                        {
-                            return false;
-                        }
-
-                        MethodSignature pointerSignatureType = ((FunctionPointerType)type).Signature;
-                        MethodSignature pointerSignatureParam = ((FunctionPointerType)parameter).Signature;
-
-                        for (int i = 0; i < pointerSignatureType.Length; i++)
-                        {
-                            if (!InstantiationMatches(pointerSignatureType[i], pointerSignatureParam[i], instArgs))
-                            {
-                                return false;
-                            }
-                        }
-
-                        return InstantiationMatches(pointerSignatureType.ReturnType, pointerSignatureParam.ReturnType, instArgs);
-
-                    case TypeFlags.SignatureMethodVariable:
-                        SignatureMethodVariable sigMethodVar = (SignatureMethodVariable)type;
-                        return sigMethodVar.Index < instArgs.Count && instArgs[sigMethodVar.Index] == parameter;
-
-                    case TypeFlags.SignatureTypeVariable:
-                        SignatureTypeVariable sigTypeVar = (SignatureTypeVariable)type;
-                        return sigTypeVar.Index < instArgs.Count && instArgs[sigTypeVar.Index] == parameter;
-
-                    case TypeFlags.GenericParameter:
-                        throw new ArgumentException();
-
-                    default:
-                        if (parameter.Instantiation.Length != type.Instantiation.Length)
-                        {
-                            return false;
-                        }
-
-                        for (int i = 0; i < type.Instantiation.Length; i++)
-                        {
-                            if (type.Instantiation[i].ContainsSignatureVariables())
-                            {
-                                if (!InstantiationMatches(type.Instantiation[i], parameter.Instantiation[i], instArgs))
-                                {
-                                    return false;
-                                }
-                            }
-                            else if (type.Instantiation[i] != parameter.Instantiation[i])
-                            {
-                                return false;
-                            }
-                        }
-
-                        return true;
-                }
-            }
-
-            static bool SignatureMatches(MethodDesc method, List<TypeDesc> parameter, List<TypeDesc> instArgs)
+            static bool SignatureMatches(MethodDesc method, List<TypeDesc> parameter)
             {
                 if (parameter.Count != method.Signature.Length)
-                {
                     return false;
-                }
 
                 for (int i = 0; i < method.Signature.Length; i++)
                 {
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler/RdXmlRootProvider.cs
```diff
@@ -160,7 +160,7 @@ static bool SignatureMatches(MethodDesc method, List<TypeDesc> parameter)
 
                 for (int i = 0; i < method.Signature.Length; i++)
                 {
-                    if (method.Signature[i] != parameter[i])
+                    if (!method.Signature[i].IsSignatureVariable && method.Signature[i] != parameter[i])
```
MichalStrehovsky:
```suggestion
                    if (!method.Signature[i].ContainsSignatureVariables() && method.Signature[i] != parameter[i])
```

This will also catch things like `void Method(List<T> parameter)` as opposed to just `void Method(T parameter)`.

hez2010: I've update the code to match it with instArgs.

hez2010: Done in 9410939

hez2010:
I think it's better to match the signature variable, but it seems that the correct approach is not as simple as the implementation in 2eef27e.

hez2010:
@MichalStrehovsky I've added support to match all parameters, and I've tested it with below code:

```csharp
using System;

namespace DimRepro
{
    unsafe interface ITest<T>
    {
        void Bar(string baz) => Console.WriteLine(baz);
        void Bar(T what, string baz) => Console.WriteLine(what.ToString() + baz);
        void Bar<U>(U what, string baz, List<T> how) => Console.WriteLine(what.ToString() + how.ToString() + baz);
        void Bar<U>(U[] what, string baz, List<T> how) => Console.WriteLine(what.ToString() + how.ToString() + baz);
        void Bar<U>(List<T> what, string baz, List<U> how) => Console.WriteLine(what.ToString() + how.ToString() + baz);
        void Bar<U>(List<U[]> what, string baz, List<T> how) => Console.WriteLine(what.ToString() + how.ToString() + baz);
        void Bar<U>(Dictionary<U, T> what, string baz, List<T> how) => Console.WriteLine(what.ToString() + how.ToString() + baz);
        void Bar<U>(Dictionary<T, U> what, string baz, List<T> how) => Console.WriteLine(what.ToString() + how.ToString() + baz);
    }

    class Test<T> : ITest<T> { }

    class Program
    {
        {
            ITest<float> x = new Test<float>();
            x.Bar("test");
            x.Bar(5, "test");
            x.Bar(5, "hello", new List<float>());
            x.Bar(new[] { 5 }, "hello", new List<float>());
            x.Bar(new List<float> { 5f }, "hello", new List<string>());
            x.Bar(new List<int[]> { new[] { 5 } }, "hello", new List<float>());
            x.Bar(new Dictionary<int, float>(), "hello", new List<float>());
            x.Bar(new Dictionary<float, int>(), "hello", new List<float>());
        }
    }
}
```

rd.xml:
```xml
<Directives>
    <Application>
        <Assembly Name="DimRepro">
            <Type Name="DimRepro.ITest`1[[System.Single,System.Private.CoreLib]]" Dynamic="Required All">
                <Method Name="Bar" Dynamic="Required">
                    <Parameter Name="System.Single,System.Private.CoreLib" />
                    <Parameter Name="System.String,System.Private.CoreLib" />
                </Method>
                <Method Name="Bar" Dynamic="Required">
                    <GenericArgument Name="System.Int32,System.Private.CoreLib" />
                    <Parameter Name="System.Int32,System.Private.CoreLib" />
                    <Parameter Name="System.String,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.List`1[[System.Single,System.Private.CoreLib]],System.Private.CoreLib" />
                </Method>
                <Method Name="Bar" Dynamic="Required">
                    <GenericArgument Name="System.Int32,System.Private.CoreLib" />
                    <Parameter Name="System.Int32[],System.Private.CoreLib" />
                    <Parameter Name="System.String,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.List`1[[System.Single,System.Private.CoreLib]],System.Private.CoreLib" />
                </Method>
                <Method Name="Bar" Dynamic="Required">
                    <GenericArgument Name="System.Int32,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.List`1[[System.Single,System.Private.CoreLib]],System.Private.CoreLib" />
                    <Parameter Name="System.String,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.List`1[[System.Int32,System.Private.CoreLib]],System.Private.CoreLib" />
                </Method>
                <Method Name="Bar" Dynamic="Required">
                    <GenericArgument Name="System.Int32,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.List`1[[System.Int32[],System.Private.CoreLib]],System.Private.CoreLib" />
                    <Parameter Name="System.String,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.List`1[[System.Single,System.Private.CoreLib]],System.Private.CoreLib" />
                </Method>
                <Method Name="Bar" Dynamic="Required">
                    <GenericArgument Name="System.Int32,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.Dictionary`2[[System.Int32,System.Private.CoreLib],[System.Single,System.Private.CoreLib]],System.Private.CoreLib" />
                    <Parameter Name="System.String,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.List`1[[System.Single,System.Private.CoreLib]],System.Private.CoreLib" />
                </Method>
                <Method Name="Bar" Dynamic="Required">
                    <GenericArgument Name="System.Int32,System.Private.CoreLib" />
                    <Parameter Name="System.Int32*,System.Private.CoreLib" />
                    <Parameter Name="System.String,System.Private.CoreLib" />
                    <Parameter Name="System.Collections.Generic.List`1[[System.Single,System.Private.CoreLib]],System.Private.CoreLib" />
                </Method>
            </Type>
            <Type Name="DimRepro.Test`1[[System.Single,System.Private.CoreLib]]" Dynamic="Required All" />
        </Assembly>

    </Application>
</Directives>
```

src/coreclr/tools/aot/ILCompiler/RdXmlRootProvider.cs
```diff
@@ -153,14 +153,93 @@ private void ProcessMethodDirective(IRootingServiceProvider rootProvider, Module
...
             {
                 if (parameter.Count != method.Signature.Length)
+                {
                     return false;
+                }
 
                 for (int i = 0; i < method.Signature.Length; i++)
                 {
-                    if (method.Signature[i] != parameter[i])
+                    if (!InstantiationMatches(method.Signature[i], parameter[i], instArgs))
```
MichalStrehovsky, line 242: Good trick with just having this match the instantiated siganture.

Wondering if we would get the same effect if we just move the `SignatureMatches` check after the `MakeInstantiatedMethod` - we would be comparing the already instantiated signature that way.

hez2010: I'll give it a try.

hez2010: I think they're the same.
Done in 4ea4b2f.

hez2010:
I think this PR is ready.

MichalStrehovsky (approved):
Thank you!

PR #750, opened 2021-02-26
Disable tiered compilation of ILCompiler

MichalStrehovsky:
Tiered compilation is detrimental to both:

* raw throughput (disabling it leads to 3-5% improvement in AOT compilation wallclock time), and
* getting stable throughput numbers (throughput numbers are noisy within 2-3%, but well above the numbers with TC off)

Cc @dotnet/crossgen-contrib - these things usually map to crossgen2 as well; might be worth trying. But it's also possible that this depends on the size of input. NativeAOT compiles more stuff and maps more closely to composite crossgen2 compilation sizes - not the "normal" crossgen2 compilations.


jkotas (approved):
Nice!

PR #751, opened 2021-02-26
More efficient debug loc info generation

MichalStrehovsky:
This puts locinfo generation on the same plan as varinfo - only capture the native information in JitInterface and translate to managed terms at a later time.

Drops the number of total GC allocations by 10%. The throughput impact is mostly a wash since we no longer do the mapping in the multithreaded phase. This should be fixable, maybe even while we're still using the LLVM-based ObjWriter.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/DependencyAnalysis/MethodCodeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/DependencyAnalysis/MethodCodeNode.cs
@@ -247,9 +247,7 @@ namespace ILCompiler.DependencyAnalysis
                     int offset = sequencePoint.Offset;
                     if (offset >= sequencePoints.Length)
                     {
-                        int newLength = sequencePoints.Length;
-                        while (newLength <= offset)
-                            newLength *= 2;
+                        int newLength = Math.Max(2 * sequencePoints.Length, sequencePoint.Offset + 1);
                         Array.Resize(ref sequencePoints, newLength);
                     }
                     sequencePoints[offset] = (sequencePoint.Document, sequencePoint.LineNumber);


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ObjectWriter.cs
```diff
@@ -496,7 +483,7 @@ public void BuildDebugLocInfoMap(ObjectNode node)
             INodeWithDebugInfo debugNode = node as INodeWithDebugInfo;
             if (debugNode != null)
             {
-                DebugLocInfo[] locs = debugNode.DebugLocInfos;
+                IEnumerable<NativeSequencePoint> locs = debugNode.GetNativeSequencePoints();
                 if (locs != null)
```
jkotas, line 487: Null check no longer needed?

src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/DependencyAnalysis/MethodCodeNode.cs
```diff
@@ -236,6 +237,53 @@ public void InitializeNonRelocationDependencies(DependencyList dependencies)
...
+            var sequencePoints = new (string Document, int LineNumber)[_debugLocInfos.Length * 4 /* chosen empirically */];
+            try
+            {
+                foreach (var sequencePoint in _debugInfo.GetSequencePoints())
+                {
+                    int offset = sequencePoint.Offset;
+                    if (offset >= sequencePoints.Length)
+                    {
+                        int newLength = sequencePoints.Length;
+                        while (newLength <= offset)
```
jkotas, line 251: `Math.Max(2 * sequencePoints.Length, sequencePoint.Offset + 1)` instead of loop?

jkotas (approved):
Thank you!

PR #755, opened 2021-02-26
NativeAOT-LLVM: Enable dead code elimination via TentativeMethodNode in LLVM generation

yowl:
This is the LLVM side for https://github.com/dotnet/runtimelab/pull/643 to enable dead code elimination for the LLVM target.  I had a bit of a problem with the throw help dependency below.

Enabled the test for the Wasm target in CI.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeMethodNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeMethodNode.cs
@@ -40,8 +40,6 @@ namespace ILCompiler.DependencyAnalysis
         {
             if (factory.Target.Architecture == TargetArchitecture.Wasm32)
             {
-                DependencyList dependencyList = base.ComputeNonRelocationBasedDependencies(factory) ?? new DependencyList();
-                dependencyList.Add(GetTarget(factory), "WebAssembly tentative node throw helper");
                 return new ObjectData(null, new Relocation[] { new Relocation(RelocType.IMAGE_REL_BASED_HIGHLOW, 0, GetTarget(factory)) }, 0, null);
             }
             return base.GetData(factory, relocsOnly);


src/coreclr/tools/Common/Compiler/DependencyAnalysis/AssemblyStubNode.cs
```diff
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-
```
MichalStrehovsky, line 5: Nit: unnecessary whitespace change

src/coreclr/tools/Common/Compiler/DependencyAnalysis/AssemblyStubNode.cs
```diff
@@ -68,6 +67,9 @@ public override ObjectData GetData(NodeFactory factory, bool relocsOnly)
                     arm64Emitter.Builder.AddSymbol(this);
                     return arm64Emitter.Builder.ToObjectData();
 
+                case TargetArchitecture.Wasm32:
+                    return new ObjectData(null, null, 0, null);
```
MichalStrehovsky, line 71: Might be better to add a new virtual method for this. This is mostly so that the TentativeMethod node can override it to return `new ObjectData(null, new Relocation[] { the target goes here }...);`.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeMethodNode.cs
```diff
@@ -87,5 +87,13 @@ public override bool RepresentsIndirectionCell
         public override int ClassCode => 0x562912;
 
         public override bool IsShareable => ((ObjectNode)_methodNode).IsShareable;
+
+        protected override DependencyList ComputeNonRelocationBasedDependencies(NodeFactory factory)
+        {
+            Debug.Assert(factory.Target.Architecture == TargetArchitecture.Wasm32); // only Web Assembly target needs to add the ThrowHelpers.ThrowBodyRemoved dependency like this.
```
MichalStrehovsky, line 93: This assert is going to hit outside WASM because we can ComputeNonRelocation everywhere.

Would be better to do the virtual method I suggested in the comment on AssemblyStubNode.cs.

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/LLVMCodegenNodeFactory.cs
```diff
@@ -47,6 +47,10 @@ protected override IMethodNode CreateMethodEntrypointNode(MethodDesc method)
             }
             if (CompilationModuleGroup.ContainsMethodBody(method, false))
             {
+                // // We might be able to optimize the method body away if the owning type was never seen as allocated.
```
MichalStrehovsky, line 50:
```suggestion
                // We might be able to optimize the method body away if the owning type was never seen as allocated.
```

yowl:
Hhave made a virtual method which I hope is close to what you were requesting.  Let me know if you'd rather it was done differently.

MichalStrehovsky:
Now that I look at it, maybe we don't even need a new virtual method. We can just override `public override ObjectData GetData(NodeFactory factory, bool relocsOnly)` in the tentative node and do pretty much exactly what we're doing in the new virtual.

yowl:
I don't know why the generics test is failing here.   Locally its ok, so upgrading emscripten in case that is it.

image

src/coreclr/tools/Common/Compiler/DependencyAnalysis/AssemblyStubNode.cs
```diff
@@ -68,6 +66,9 @@ public override ObjectData GetData(NodeFactory factory, bool relocsOnly)
                     arm64Emitter.Builder.AddSymbol(this);
                     return arm64Emitter.Builder.ToObjectData();
 
+                case TargetArchitecture.Wasm32:
```
MichalStrehovsky, line 69: Would this lead to infinite recursion if this is actually reached? Maybe we can undo all changes tot his file.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeMethodNode.cs
```diff
@@ -36,6 +36,17 @@ protected virtual ISymbolNode GetTarget(NodeFactory factory)
             return helper == null ? RealBody : factory.MethodEntrypoint(helper);
         }
 
+        public override ObjectData GetData(NodeFactory factory, bool relocsOnly)
+        {
+            if (factory.Target.Architecture == TargetArchitecture.Wasm32)
+            {
+                DependencyList dependencyList = base.ComputeNonRelocationBasedDependencies(factory) ?? new DependencyList();
+                dependencyList.Add(GetTarget(factory), "WebAssembly tentative node throw helper");
```
MichalStrehovsky, line 44:
```suggestion
```

MichalStrehovsky (approved):
Thank you!

yowl:
Thanks for the help, I made a bit of a meal of this one.

PR #760, opened 2021-03-01
SRM Regex serialization support

veanes:
Implemented serialization and deserialization for SRM.Regex that is based on a custom Base64 base serialization utility. The serializer does not depend on any built-in serialization functionality (that is mostly non-CLS compliant) and works fully in the visible ASCII-text range. Is safe in any environment and is independent of line endings (e.g. if \n gets replaced by \n\r). 

Turned on serialization in all tests that are run in DEBUG mode, so that every created regex goes through a serialize/deserialize roundtrip. 

Fixed related bugs and added unit test for testing found corner cases.


veanes:
I added the last commit that exposes the SRM/DFA serialization through the ISerializable interface of Regex. It would be good if @danmoseley  or @stephentoub  take a look at that, if @olsaarik  agrees with the rest.
I also added a unit test to show how this can be used from the outside (e.g. by CredScan implementers). Personally, I think that this is an easier route to  expose this feature, rather than through a compiled assembly, although exposing it through the "DFA" in the SerializationInfo as a string may seem a bit unorthodox.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
```diff
@@ -185,11 +185,37 @@ protected internal static void ValidateMatchTimeout(TimeSpan matchTimeout)
             }
         }
 
-        protected Regex(SerializationInfo info, StreamingContext context) =>
-            throw new PlatformNotSupportedException();
+        protected Regex(SerializationInfo info, StreamingContext context)
+        {
+            SerializationInfoEnumerator infoEnum = info.GetEnumerator();
+            SRM.Regex srmregex = null;
+            while (infoEnum.MoveNext())
```
danmoseley, line 192: In this context perhaps the code can assume that the first entry in the SerializationInfo is the SRM blob, or it's not there at all.

But I'm not sure about using the ISerialization mechanism anyway (don't feel strongly)

veanes: Right, then it does not really matter what the name of the key is.
The code would then be if (infoEnum.MoveNext()) .. use the Current.Value (if it is a string) throw exception. I'll update that.

veanes: Regarding Dan's questions below. 

CredScan would use the deserialization and store them separately, as it does currently, for around 150 regexes or so for a fast deployment from scratch (in some key scenarios) to avoid an upfront cost  -- when the regexes have not been cached or loaded yet -- so the runtime is not up and running running yet.

However, the above represents a rather unusual scenario, but covers what they need/or have needed so far. (The drawback is that the particular serialization format is exposed.)
It would make **total sense** to also cache the serialization for a given (regex,options) pair for SRM so that if a DFA regex has already been seen it would just be deserialized -- as part of the existing regex caching system (modified a bit for the srm case).

However, all that said, we need to reevaluate the improved startup cost. I eliminated several bottlenecks (that had to do with use of earlier BSTs, perhaps minterms was not the only culprit there). But I still expect the cost to be considerable compared to deserialization that should be just nanosecs even for a reasonably large regex.

Regarding cost calculations. I think there are two scenarios. 1) Seeing what the initial "one off" cost is for, say say setting up all the 150 or so CredScan regexes (to pick a concrete example), i.e., the initial serialization cost compared to just deserializing them. 2) But also caching the DFA regex object and reusing it when possible (this is TBD): essentially caching
(pattern,options) -> srmregex.serialize(), when DFA in options, an emasuring what impact this has overall when reusing same regexes.

I can evaluate (1) and report on that. (I have problems compiling in release mode though from the VS solution). I was also considering putting the credscan regexes into a unit test, BUT: **CredScan team does not want these regexes to be public**. 
To doublecheck before I do that: I assume the test framework code is not public!!!

veanes:
A new test error that show up is:
System.Text.RegularExpressions.Tests.RegexConstructorTests.Serialization_ThrowsNotSupported has failed.
I'll take a look at these unit tests when I arrive in Sweden (Im right now in airplane).

jkotas:
> exposes the SRM/DFA serialization through the ISerializable interface

We have published roadmap for obsoleting all BinaryFormatter related types that includes `ISerializable` interface: https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md. Exposing new features via ISerializable interface does not sound like a good idea.

stephentoub:
I don't think we should change the meaning of `pattern`.  If we want to have a way for deserializing a compiled string representation into a Regex instance we should have a separate method for that.

However, I also don't think we should be exposing new API for this.  We already have the notion of saving compiled implementations, via Regex.CompileToAssembly.  While that's not currently implemented in .NET Core (because we currently lack the ability to save ref emit'd assemblies), we are working on a Regex source generator that will achieve the same thing.  I would rather we just leverage that.
- If you use the Regex ctor with RegexOptions.DFA, the relevant state is computed on the fly and cached along with the rest of the Regex state.
- If you use the Regex ctor with RegexOptions.DFA | RegexOptions.Compiled, the relevant state is computed on the fly and cached along with the rest of the Regex state, and potentially a more optimized implementation that utilizes ref emit is employed.
- If you use the Regex source generator, then it spits out the precomputed state into the generated regex code (in whatever form makes sense, maybe as a string, maybe as a byte[], etc.).

veanes:
Ok. Regex.CompileToAssembly is then the way to go. I'll then get rid of the last commit (a695134) also.
> but I suspect there would be ways to optimize the DFA implementation further by generating C# code tailored to that one DFA.
This is true (we could generated a DFA as C# code, we've done it before actually).

jkotas:
> Regex.CompileToAssembly is then the way to go

The source generators are the way to go. The existing Regex.CompileToAssembly method is not going to be used by the regex source generator.

danmoseley:
It’s not a goal to write a source generator at this point. Just a convenient hook to read/write state in order to make some perf measurements. We aren’t committing to any shape.

jkotas:
You can just add methods that do what you need, and access them via reflection.

veanes:
The methods we need to access for perf measurements to start with, are essentially  saving SRM regex as string and parsing it back from a string. These are currently the methods SRM.Regex.Serialize and SRM.Regex.Deserialize -- I mean in this PR, once it gets approved and I can merge it. **Just to be clear**, these methods themselves have **nothing to do with ISerializable** or any of the related types. They could also be e.g. SRM.Regex.ToString and SRM.Regex.Parse.

The issue right now as Dan said is to expose those conveniently, I don't want to make too much fuss about it, but I want to move on. Right now I am unsure what to do. I guess, we can use reflection to access the methods but then it is not clear how easy it will be to set up the actual Regex object through reflection, but more importantly what the impact will be wrt performace and whether using reflection in this process might defeat the purpose.

stephentoub:
> add an internal static method to the Regex class that constructs it from an srm_string that can be accessed through reflection. I guess calling this method through reflection should not imply any significant overhead wrt performance measurements.

This is what Jan was suggesting.

stephentoub:
That said, I don't think this support is meaningful to the primary investigations at hand.  As I understand it, its only impact is on startup time, in that the Regex constructor will take longer to compute all that state.  That's fine right now.  From a perf standpoint, the most important things we're interested in are a) steady state throughput of invoking methods on a Regex and b) steady state memory consumption from such processing, and neither of those are going to be impacted by this; I'd expect only tests of the Regex ctor to be impacted here.

Short term if you'd like to add such methods to access via reflection, that's fine, go for it.

Longer term, I think the answer is that such methods don't exist.  If you construct a Regex via its constructor, we compute the state on the fly.  If you use the Regex source generator, the DFA is emitted in whatever precomputed / optimized / C# form that makes processing even more efficient and solves the startup overhead issue.

veanes:
I agree. The short-term goal was to see what the cost benefit is, some of the changes may also have helped improve the startup time (in particular the change from decision trees to decision diagrams and use of MTBDDs).

olsaarik (approved):

PR #761, opened 2021-03-01
Implement DynamicallyAccessedMembers.All

MichalStrehovsky:
This is not a union of all `DynamicallyAccessedMembers` in IL Linker.


jkotas (approved):
Thanks!

Issue #346, opened 2020-11-12
Add support for DynamicDependencyAttribute


MichalStrehovsky:
This was smuggled in with the dataflow annotation attributes (DynamicallyAccessedMembersAttribute) mostly because Xamarin team really insisted they need it. I don't see much point in this attribute because it caters to embedding scenarios (a method should not be removed because it's called from native code that cannot be analyzed during treeshaking).

So far all uses of this attribute within the framework are either about dependencies from native code (our own internal embedding scenario), or to work around linker design flaws (`Type.GetType("Foo, BarAssembly")` is not capable to bring BarAssembly into the closure unless it's already there, or got added with DynamicDependency).

Native AOT doesn't have these problems.

MichalStrehovsky:
Fixed in #774.

PR #766, opened 2021-03-03
Add handling of DynamicDependencyAttribute

MichalStrehovsky:
This attribute has mostly been used to work around illinker design issues, but a few usages cropped up that need this for correctness (there are trimming warning suppressions that rely on this attribute). I'm not a big fan because we never seem to be able to get suppressions right on the first try and rely on code reviews to catch bugs.

I had to copy over the Roslyn document-format parser from illink repo since this is the format that is used in `DynamicDependencyAttribute`. It will be useful for other things too.

The first commit is straight copy form the linker repo. Second commit tweaks it to compile.

The rooting is not hooked up yet because this rooting is very similar to the rooting in flow analysis, but we need to make that shareable. This pull request is already big enough.

Contributes to #346.


jkotas (approved):
LGTM. Thanks

PR #772, opened 2021-03-05
Build ObjWriter for Linux ARM64

AntonLapounov:
Use the `gen-buildsys.sh` script to properly configure cross-compilation.  `clang-9` reported the warning below, so fix `ObjectWriter::GetCVRegNum` one more time by changing its argument's type to be `unsigned` rather than `ICorDebugInfo::RegNum`, which is not properly defined.  By the way, the warning message seems a bit misleading, because the compiler still emits the comparison even when optimizations are enabled, see this [bug report](https://bugs.llvm.org/show_bug.cgi?id=16154).
```
/runtime/src/coreclr/tools/aot/ObjWriter/objwriter.cpp:628:16: warning: result of comparison of constant 16 with expression of type 'ICorDebugInfo::RegNum' is always true [-Wtautological-constant-out-of-range-compare]
    if (RegNum < sizeof(CVRegMapAmd64) / sizeof(CVRegMapAmd64[0])) {
        ~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/coreclr/tools/aot/ObjWriter/build.sh
+++ b/src/coreclr/tools/aot/ObjWriter/build.sh
@@ -6,10 +6,29 @@ RepoRoot="$2"
 BuildArch="$3"
 TargetArch="$4"
 BuildType="${5:-Release}"
+CompilerId="$(echo "$6" | tr "[:upper:]" "[:lower:]")"
 
 # Check that we have enough arguments
 if [ $# -lt 4 ]; then
-    echo "Usage: $(basename $0) ArtifactsDir RepoRoot BuildArch TargetArch [BuildType]"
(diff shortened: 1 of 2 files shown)


jkotas (approved):
Thanks!

AntonLapounov:
In an [earlier run](https://dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_apis/build/builds/1024636/logs/108) OSX x64 Release build job could not finish building JIT in 87 minutes.  In the [last run](https://dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_apis/build/builds/1024636/logs/120) it built JIT in 15 minutes.  I am wondering if there was something wrong with the machine.

PR #773, opened 2021-03-05
Add workaround for DuplicatePublishOutput check

### LINKED ISSUE #771: NETSDK1148 when using NativeAOT
Please see the attached zipped binlog. It complains that I have two files named `apphost.exe` when, in fact, I have one copy of `apphost.exe`; the other is named `x2copy.exe`, and is my all-native executable product as output by NativeAOT. This did not happen before, but I cannot say when it started, as I only run NativeAOT when cutting a release, which is not very often. I am using NativeAOT version 6.0.0-preview.3.21153.3, as downloaded automatically by NuGet when running the Restore task. Any pointers would be appreciated. Thanks!

[Here is the zipped binlog](https://github.com/dotnet/runtimelab/files/6087806/msbuild.binlog.zip).

jkotas:
Fixes #771


MichalStrehovsky (approved):
Thank you!

PR #774, opened 2021-03-05
Finish DynamicDependencyAttribute rooting

MichalStrehovsky:
* Move reflection rooting logic into a shareable location and make it more general purpose
* Call into the new logic from DynamicDependencyAttribute processor
* Test


jkotas (approved):
Thank you!

PR #775, opened 2021-03-05
Add Alpine Linux support

MichalStrehovsky:
Wanted to see how much it would take to add support for this over lunch... looks like this "Just Works" because all the infrastructural grunt work was already done for CoreCLR and the runtime doesn't seem to take any dependencies on glibc weirdnesses.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
+++ b/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
@@ -12,7 +12,6 @@
     <OutputPath>$(RuntimeBinDir)ilc</OutputPath>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
-    <RuntimeIdentifiers>linux-x64;linux-musl-x64;win-x64;osx-x64</RuntimeIdentifiers>
     <Configurations>Debug;Release;Checked</Configurations>
     <UseAppHost>true</UseAppHost>
     <ServerGarbageCollection>true</ServerGarbageCollection>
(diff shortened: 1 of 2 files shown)


MichalStrehovsky:
> Unable to find package runtime.linux-musl-x64.Microsoft.DotNet.ILCompiler.

Ah, okay, this has annoying bootstrapping problems because the code that prevents injecting this dependency relies on someone running `build.sh nativeaot.objwriter` _before_ trying to `build.sh nativeaot.tools`. I forgot I hit this locally too. I already context-switched to something else so this will have to wait.

jkotas:
> annoying bootstrapping problems

We hit the same problem for arm64. I have pushed a temporary change to get around this.

jkotas (approved):
Thanks!

src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
```diff
@@ -23,6 +22,7 @@
     <ILCompilerRuntimeIdentifier Condition="'$(TargetsWindows)' == 'true'">win-$(TargetArchitecture)</ILCompilerRuntimeIdentifier>
     <ILCompilerRuntimeIdentifier Condition="'$(TargetsLinux)' == 'true'">linux-$(TargetArchitecture)</ILCompilerRuntimeIdentifier>
     <ILCompilerRuntimeIdentifier Condition="'$(TargetsOSX)' == 'true'">osx-$(TargetArchitecture)</ILCompilerRuntimeIdentifier>
+    <ILCompilerRuntimeIdentifier Condition="'$(__DistroRid)' == 'linux-musl-x64'">linux-musl-x64</ILCompilerRuntimeIdentifier>
```
AntonLapounov, line 25: It seems `ILCompilerRuntimeIdentifier` always equals to `__DistroRid`.  We can remove the duplicate variable then.

jkotas: Good idea. #777

PR #778, opened 2021-03-07
Make 'dotnet publish' work on ARM64

AntonLapounov:
* Make `dotnet publish -r win-arm64` work on Windows ARM64 by using the correct link environment.
* Make `dotnet publish -r linux-arm64` work on Linux ARM64 by fixing `undefined reference` link errors.

Targeting ARM64 still requires passing `-p:DisableUnsupportedError=true` until we have some test coverage.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/findvcvarsall.bat
+++ b/src/coreclr/nativeaot/BuildIntegration/findvcvarsall.bat
@@ -1,6 +1,11 @@
 @ECHO OFF
 SETLOCAL
 
+IF "%~1"=="" (
+    ECHO Usage: %~nx0 ^<arch^>
+    GOTO :ERROR
+)
+
 SET vswherePath=%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe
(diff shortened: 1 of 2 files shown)


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
```diff
@@ -79,10 +79,6 @@ The .NET Foundation licenses this file to you under the MIT license.
       <LinkerArg Include="/NATVIS:&quot;$(MSBuildThisFileDirectory)CoreRTNatVis.natvis&quot;" />
     </ItemGroup>
 
-    <ItemGroup>
-      <!-- TODO <LinkerArg Include="/MACHINE:X64" /> -->
```
AntonLapounov, line 83: The linker infers the machine type from the `.obj` file we generated, which goes first on the command line.

jkotas (approved):
Thanks

AntonLapounov:
For `dotnet publish` we have `PlatformTarget` instead of `TargetArchitecture`.  I made changes to set `PlatformTarget` in generated `nativebuild.proj` files.

PR #779, opened 2021-03-07
Speed up reflection invoke by 32%

MichalStrehovsky:
This was a size on disk optimization, but in reality it saves only two movs in the invoke thunks but slows down invoke considerably.

Threadstatic access is not as fast in NativeAOT as it was in .NET Native. It's still plenty fast, but could be faster.

Also cleaning up some cruft.


jkotas (approved):
Nice!

PR #786, opened 2021-03-08
Avoid double cast for root converter when writing to UTF-8 bytes

layomia:


layomia:
(Will skip CI for small change. Need these bits ASAP via the nuget package for local perf testing)

PR #787, opened 2021-03-08
Enable Linux-x64-musl build and tests

jkotas:


MichalStrehovsky:
Thanks for looking into these! I assumed if Hello World works, everything will work.

MichalStrehovsky (approved):
Awesome, thank you!

PR #789, opened 2021-03-09
Fix generation of DWARF debug info for `void` and `void*`

MichalStrehovsky:
Fixes a spurious `error: need to add support for DW_TAG_base_type 'void' encoded with DW_ATE = 0x0, bit_size = 0` error message from lldb.

The problem was that `DW_ATE_address` is actually used to represent `void*`, not `void` and we didn't fill out the fields that debugger wants (the size of the type).

DWARF doesn't have a well-known `void` type. It has `DW_TAG_unspecified_type` that is meant to be used to represent `void` (per http://eagercon.com/dwarf/issues/minutes-001017.htm), but lldb [can only handle it for `nullptr_t`](https://github.com/llvm/llvm-project/blob/3b47bd32f9df4a57db98db5f35e680c7bd9fde3e/lldb/source/Plugins/SymbolFile/DWARF/DWARFASTParserClang.cpp#L617-L632). Clang represents `void*` as a simple `DW_TAG_pointer_type` with no element type.

Because of how ObjWriter API surface is structured, we need some way to get to a TypeIndex of `void` so I'm implementing `void` as `DW_TAG_unspecified_type`, but also adding a kludge to emit a elementless `DW_TAG_pointer_type` for `void*` instead of a pointer to void. This means we will end up emitting an unreferenced definition of `void`. Probably fine.

### DIFF
--- a/src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
+++ b/src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
@@ -171,16 +171,16 @@ void DwarfPrimitiveTypeInfo::DumpTypeInfo(MCObjectStreamer *Streamer, UserDefine
 // DwarfVoidTypeInfo
 
 void DwarfVoidTypeInfo::DumpStrings(MCObjectStreamer* Streamer) {
-    Streamer->EmitBytes(StringRef("void"));
-    Streamer->EmitIntValue(0, 1);
+  Streamer->EmitBytes(StringRef("void"));
+  Streamer->EmitIntValue(0, 1);
 }
 
 void DwarfVoidTypeInfo::DumpTypeInfo(MCObjectStreamer* Streamer, UserDefinedDwarfTypesBuilder* TypeBuilder) {
-    // Abbrev Number
-    Streamer->EmitULEB128IntValue(DwarfAbbrev::VoidType);
+  // Abbrev Number
+  Streamer->EmitULEB128IntValue(DwarfAbbrev::VoidType);
 
-    // DW_AT_name
-    EmitSectionOffset(Streamer, StrSymbol, 4);
+  // DW_AT_name
+  EmitSectionOffset(Streamer, StrSymbol, 4);
 }
 
 // DwarfEnumerator


src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
```diff
@@ -712,8 +737,23 @@ unsigned UserDefinedDwarfTypesBuilder::GetArrayTypeIndex(
...
+  // Creating a pointer to what DWARF considers Void type (DW_TAG_unspecified_type -
+  // per http://eagercon.com/dwarf/issues/minutes-001017.htm) leads to unhappines
+  // since debuggers don't really know how to handle that. The Clang symbol parser
+  // in LLDB only handles DW_TAG_unspecified_type if it's named
+  // "nullptr_t" or "decltype(nullptr)".
+  //
+  // We resort to this kludge to generate the exact same debug info for void* that
+  // clang would generate (pointer type with no element type specified).
+  if (PointerDescriptor.ElementType == VoidTypeIndex)
+      DwarfTypes.push_back(make_unique<DwarfVoidPtrTypeInfo>());
```
jkotas, line 753: Nit: Indentation

jkotas (approved):
LGTM. Thank you!

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
jkotas: (multiple places - 2 vs 4 spaces)

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
```diff
@@ -510,6 +524,17 @@ void DwarfPointerTypeInfo::DumpTypeInfo(MCObjectStreamer *Streamer, UserDefinedD
   Streamer->EmitIntValue(TypeDesc.Is64Bit ? 8 : 4, 1);
 }
 
+// DwarfVoidPtrTypeInfo
+
+void DwarfVoidPtrTypeInfo::DumpStrings(MCObjectStreamer* Streamer) {
+    // nothing to dump
```
MichalStrehovsky, line 530:
```suggestion
  // nothing to dump
```

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
```diff
@@ -510,6 +524,17 @@ void DwarfPointerTypeInfo::DumpTypeInfo(MCObjectStreamer *Streamer, UserDefinedD
...
+}
+
+void DwarfVoidPtrTypeInfo::DumpTypeInfo(MCObjectStreamer* Streamer, UserDefinedDwarfTypesBuilder* TypeBuilder) {
+    // Abbrev Number
+    Streamer->EmitULEB128IntValue(DwarfAbbrev::VoidPointerType);
```
MichalStrehovsky, line 535:
```suggestion
  // Abbrev Number
  Streamer->EmitULEB128IntValue(DwarfAbbrev::VoidPointerType);
```

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
MichalStrehovsky, line 753:
```suggestion
    DwarfTypes.push_back(make_unique<DwarfVoidPtrTypeInfo>());
```

MichalStrehovsky: Oh god, I was fighting VS on this everywhere. Need to find the right .editorconfig voodoo so that we can check that in.

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
```diff
@@ -169,6 +168,21 @@ void DwarfPrimitiveTypeInfo::DumpTypeInfo(MCObjectStreamer *Streamer, UserDefine
   Streamer->EmitIntValue(TD.ByteSize, 1);
 }
 
+// DwarfVoidTypeInfo
+
+void DwarfVoidTypeInfo::DumpStrings(MCObjectStreamer* Streamer) {
+    Streamer->EmitBytes(StringRef("void"));
+    Streamer->EmitIntValue(0, 1);
```
MichalStrehovsky, line 175:
```suggestion
  Streamer->EmitBytes(StringRef("void"));
  Streamer->EmitIntValue(0, 1);
```

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
```diff
@@ -169,6 +168,21 @@ void DwarfPrimitiveTypeInfo::DumpTypeInfo(MCObjectStreamer *Streamer, UserDefine
...
+}
+
+void DwarfVoidTypeInfo::DumpTypeInfo(MCObjectStreamer* Streamer, UserDefinedDwarfTypesBuilder* TypeBuilder) {
+    // Abbrev Number
+    Streamer->EmitULEB128IntValue(DwarfAbbrev::VoidType);
+
+    // DW_AT_name
+    EmitSectionOffset(Streamer, StrSymbol, 4);
```
MichalStrehovsky, line 183:
```suggestion
  // Abbrev Number
  Streamer->EmitULEB128IntValue(DwarfAbbrev::VoidType);

  // DW_AT_name
  EmitSectionOffset(Streamer, StrSymbol, 4);
```

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfTypeBuilder.cpp
AntonLapounov: You can set `Default formatting style = LLVM` in Tools->Options:
https://devblogs.microsoft.com/cppblog/clangformat-support-in-visual-studio-2017-15-7-preview-1/

MichalStrehovsky: The annoying part about that is that that's a global setting and I would need to turn it off depending on what project I'm working with.

PR #792, opened 2021-03-10
Add editorconfig for objwriter

MichalStrehovsky:
Just enough to not have to fight IntelliSense on every newline.

### DIFF
--- a/src/coreclr/tools/aot/ObjWriter/.editorconfig
+++ b/src/coreclr/tools/aot/ObjWriter/.editorconfig
@@ -1,0 +1,3 @@
+[*.{cpp,h}]
+
+indent_size = 2


AntonLapounov:
Why don't we enforce LLVM formatting rules by copying the `artifacts\llvm\.clang-format` file to the `ObjWriter` directory following [this article](https://devblogs.microsoft.com/cppblog/clangformat-support-in-visual-studio-2017-15-7-preview-1/)?

MichalStrehovsky:
You mean copy that out as part of the build and add it to gitignore so that it doesn't pollute the workspace, or just take the file and check that in, along with a THIRD-PARTY-NOTICES.txt entry for the included file? Frankly, neither option sound like something I would want to sink my time into.

jkotas:
> I have clang-format option turned off in VS

Same here.

AntonLapounov:
> take the file and check that in, along with a THIRD-PARTY-NOTICES.txt entry for the included file

The whole content of the file is the single line `BasedOnStyle: LLVM`, which will never change.

> Also I have clang-format option turned off in VS because the first (and last) time I tried it

The required setting ("Enable ClangFormat support") is ON by default.  There is another setting "Default formatting style = Visual Studio" setting, which we do not need to change.

> it found the clangformat file in the RyuJIT codebase and reformatted a bunch of stuff

Out of curiosity, why do we have the `.clang-format` file in `src/coreclr/jit` if we do not want to use it?

MichalStrehovsky:
> Out of curiosity, why do we have the .clang-format file in src/coreclr/jit if we do not want to use it?

RyuJIT team enforces the coding style in the CI with that file; they have some custom tooling for that. Fighting with the CI is the rite of passage everyone making a change to RyuJIT has to do. I assume it very much depends on how exactly you invoke clang-format and which version you use because as I wrote above, last time I tried it in VS, it very much disagreed with how the already checked in code should look like and reformatting with VS would fail the CI.

It's [a lot of fun](https://github.com/dotnet/coreclr/pull/8014#issuecomment-258885307).

PR #794, opened 2021-03-11
Document minimal VS installation for samples

AntonLapounov:
Reduce prerequisites to build and run samples on Windows.  Alleviates MichalStrehovsky/SeeSharpSnake#2.  I am keeping "CoreRT" references for consistency with other documentation.

Tested on Windows ARM64 with 1 GiB of RAM.

### DIFF
--- a/samples/prerequisites.md
+++ b/samples/prerequisites.md
@@ -20,11 +20,11 @@ Then launch the bootstrapper passing the installation path and the two required 
 ```cmd
 vs_buildtools.exe --installPath C:\VS2019 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 Microsoft.VisualStudio.Component.Windows10SDK.19041 --passive --norestart --nocache
 ```
-Alternatively you may launch the bootstrapper without any options and use Visual Studio Installer UI to enable "C++ x64/x86 build tools" and "Windows 10 SDK" individual components.
+Alternatively you may launch the bootstrapper without any options and use Visual Studio Installer UI to enable C++ x64/x86 build tools and Windows 10 SDK individual components.
 
 Notes:
-- You may skip the `Windows10SDK` component if you already have Windows 10 SDK installed on your machine.
-- To target Windows ARM64, you need to add the `Microsoft.VisualStudio.Component.VC.Tools.ARM64` ("C++ ARM64 build tools") component instead.
+- You may skip the `Windows10SDK.19041` component if you already have Windows 10 SDK installed on your machine.
+- To target Windows ARM64, you need to add the `Microsoft.VisualStudio.Component.VC.Tools.ARM64` (C++ ARM64 build tools) component instead.
 - The `--installPath` option affects Build Tools installation only. Visual Studio Installer is always installed into
 the `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer` directory.
 


samples/prerequisites.md
```diff
@@ -1,23 +1,51 @@
...
+edition installation required. First download the [bootstrapper](https://aka.ms/vs/16/release/vs_buildtools.exe) executable.
+On Windows 10 version 1803 or later you may use the `curl` tool:
+
+```cmd
+curl -L https://aka.ms/vs/16/release/vs_buildtools.exe -o vs_buildtools.exe
+```
+
+Then launch the bootstrapper passing the installation path and the two required components (requires elevation):
+```cmd
+vs_buildtools.exe --installPath C:\VS2019 --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 Microsoft.VisualStudio.Component.Windows10SDK.19041 --passive --norestart --nocache --lang en-US
```
jkotas, line 21: I have just run it and it did not install anything into `c:VS2019`. Does this modify your existing VS installation if there is one already?

samples/prerequisites.md
```diff
@@ -1,23 +1,51 @@
...
-The following pre-requisites need to be installed for building .NET 5 projects with CoreRT:
+The following pre-requisites need to be installed for building .NET 5 projects with CoreRT.
 
 # Windows
 
-* Install [Visual Studio 2019](https://visualstudio.microsoft.com/vs/community/), including Visual C++ support.
+## Install [Visual Studio 2019](https://visualstudio.microsoft.com/downloads/) with Visual C++ Build Tools
+
+The free [Community](https://visualstudio.microsoft.com/vs/community/) edition or higher is recommended.
+If you wish to save disk space and do not need Visual Studio IDE, here is the minimal [Build Tools](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019)
```
jkotas, line 11: These steps feel a lot more complicated compared to launch Visual Studio Installer and check "C++ suport". I think we should keep this simple option as the preferred option. We can have these disk-space saving steps as "## Advanced Alternative: Minimum Visual C++ Build Tools installation" or something.

samples/prerequisites.md
```diff
@@ -1,23 +1,51 @@
...
+Alternatively you may launch the bootstrapper without any options and use Visual Studio Installer UI to enable "C++ x64/x86 build tools" and "Windows 10 SDK" individual components.
+
+Notes:
+- You may skip the `Windows10SDK` component if you already have Windows 10 SDK installed on your machine.
+- To target Windows ARM64, you need to add the `Microsoft.VisualStudio.Component.VC.Tools.ARM64` ("C++ ARM64 build tools") component instead.
+- The `--installPath` option affects Build Tools installation only. Visual Studio Installer is always installed into
+the `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer` directory.
+
+## Install Visual C++ 2019 Redistributable Package
+If you have no `vcruntime140.dll` library installed in your `%SystemRoot%\System32` directory, you also need to install Visual C++ 2019 Redistributable package, which the compiler depends on:
```
jkotas, line 32: Which compiler is this talking about?

samples/prerequisites.md
AntonLapounov: Need to remove `--lang en-US` (it is used for creating a layout only). Thanks for catching this.

samples/prerequisites.md
AntonLapounov: `objwriter.dll` has dependency on it. IIRC, this missing dependency was reported by one of early adopters in the past.

jkotas: It would be nice to just copy it over to the package so that we do not have to have this step here.

jkotas: (Or link it statically.)

samples/prerequisites.md
AntonLapounov: Makes sense.

jkotas (approved):
Thank you!

samples/prerequisites.md
AntonLapounov: I have created issue #795 to track that.

MichalStrehovsky:
Nice!

AntonLapounov:
It is noteworthy that with this minimal setup (~1.5 GB for VS plus ~2GB for SDK) one can successfully build not only NativeAOT samples but also do the normal build of NativeAOT and Runtime repos if we add `-products *` in a couple of `vswhere` commands: https://github.com/dotnet/runtimelab/blob/3553284bea7b650874c8dcc3f78d6be085a047bb/eng/native/init-compiler-and-cmake.cmd#L37 Without that addition `vswhere` ignores Build Tools installations.

jkotas:
Sounds like something that may be worth sending a PR to dotnet/runtime for?

PR #797, opened 2021-03-11
Delete RemovingILProvider

jkotas:


jkotas:
There does not seem to be anything left that needs switching to the new feature plan.

MichalStrehovsky (approved):
The serialization guard thing was the one leftover. It was added as part of https://github.com/dotnet/corert/pull/7909. We can restore that when it's needed later.

PR #801, opened 2021-03-12
Fix runsettings generation

safern:
When the build is a cross targeting build and using `dotnet test` it generates the .runsettings file to the wrong locations.

Arcade already sets `OutputPath` for us so it is better to use that rather than `OutDir` which might not be set when the targets file is imported.

### DIFF
--- a/eng/testing/runsettings.targets
+++ b/eng/testing/runsettings.targets
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <RunSettingsInputFilePath>$(MSBuildThisFileDirectory).runsettings</RunSettingsInputFilePath>
-    <RunSettingsAppOutputFilePath>$(OutDir).runsettings</RunSettingsAppOutputFilePath>
+    <RunSettingsAppOutputFilePath>$(OutputPath).runsettings</RunSettingsAppOutputFilePath>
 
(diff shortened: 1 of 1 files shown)


ericstj (approved):
Looks like an OK change.  It’s unfortunate we need to do this, especially in an initial target.  There is no official extension point for providing runsettings content?

safern:
I believe the only other way is via command line arguments.

This is also how dotnet/runtime does it for vstest settings.

Also .runsettings is used by the test explorer in VS so this is the way to have a consistent story in between command line and VS.

cc: @ViktorHofer

safern:
I'll push this commit to the standalone-template branch and will put up a PR to other standalone experiments.

ViktorHofer:
> There is no official extension point for providing runsettings content?

As Santiago mentioned, only via CLI arguments. The .runsettings approach allows using VS Test Explorer and `dotnet test` without overriding the `VSTest` target with a custom host.

The InitialTarget (or any target that runs before VS is opened, i.e. part of build) is only required by VS as there is no hot-reload functionality for runsettings files built in today. If VS, VSCode and VS4M would support that, the InitialTarget wouldn't be necessary anymore and the runsettings file could just be generated as part of the test assembly's build.

PR #804, opened 2021-03-12
Fix heap corruption in StelemRef on negative index

MichalStrehovsky:
StelemRef with a negative index would result in a fun heap corruption. The `IList_Generic_ItemSet_NegativeIndex_ThrowsException` test in System.Runtime.Tests was leaving us with a corrupted heap.

### DIFF
--- a/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/TypeCast.cs
+++ b/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/TypeCast.cs
@@ -725,7 +725,7 @@ namespace System.Runtime
             // This is supported only on arrays
             Debug.Assert(array.EEType->IsArray, "first argument must be an array");
 
-            if (index >= array.Length)
+            if ((uint)index >= (uint)array.Length)
             {
                 throw array.EEType->GetClasslibException(ExceptionIDs.IndexOutOfRange);
             }


jkotas (approved):
Thanks!

jkotas (approved):
Thanks!

AntonLapounov:
There are other places in the code with like that.  I am wondering if there is a good analyzer to find them.
https://github.com/dotnet/runtimelab/blob/fddede139c15cffcf73e8f73bfec1b1dda3bff18/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/CallInterceptor.cs#L119-L125 where https://github.com/dotnet/runtimelab/blob/fddede139c15cffcf73e8f73bfec1b1dda3bff18/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/CallInterceptor.cs#L66

jkotas:
@AntonLapounov Would you mind submitting a fix for the ones you have noticed?

MichalStrehovsky:
> SetByRef

FWIW, that appears to be dead code (even dead in the CoreRT repo that contains all of the universal shared generic support code).

AntonLapounov:
I think this one is not dead. `resourceInfo.Index` may be negative and incorrectly pass the check. #816
https://github.com/dotnet/runtimelab/blob/88dd09557a2890ce604f9a87517cfb1f1f9aeb56/src/coreclr/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/ExecutionEnvironmentImplementation.ManifestResources.cs#L87-L93

PR #805, opened 2021-03-12
Run API compat

ericstj:


src/System.Web.Compatibility/src/ApiCompatBaseline.txt
```diff
@@ -0,0 +1,34 @@
...
+TypesMustExist : Type 'System.Web.Security.MembershipPasswordFormat' does not exist in the implementation but it does exist in the contract.
+TypesMustExist : Type 'System.Web.Security.MembershipProvider' does not exist in the implementation but it does exist in the contract.
+TypesMustExist : Type 'System.Web.Security.MembershipProviderCollection' does not exist in the implementation but it does exist in the contract.
+TypesMustExist : Type 'System.Web.Security.MembershipUser' does not exist in the implementation but it does exist in the contract.
+TypesMustExist : Type 'System.Web.Security.MembershipUserCollection' does not exist in the implementation but it does exist in the contract.
+TypesMustExist : Type 'System.Web.Security.MembershipValidatePasswordEventHandler' does not exist in the implementation but it does exist in the contract.
+TypesMustExist : Type 'System.Web.Security.RoleProvider' does not exist in the implementation but it does exist in the contract.
+TypesMustExist : Type 'System.Web.Security.ValidatePasswordEventArgs' does not exist in the implementation but it does exist in the contract.
+
+// Don't care about these missing attributes
```
safern, line 14: What is this API Compat running against as the contract? Should you instead use an exclude attributes file?

ericstj: .NETFramework reference assembly.  I didn't bother with a list of attributes since it'll never change and this works just as well.

I'm adding API compat this here to make sure that when folks disable some types from the PNSE we make sure they are fully implemented

safern: Ah I see. Makes sense.

safern (approved):

PR #806, opened 2021-03-12
Simplify metadata classes

layomia:


layomia:
Skipping CI for small change.

PR #816, opened 2021-03-15
Remove CallInterceptor and harden some range asserts

AntonLapounov:


src/coreclr/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/ExecutionEnvironmentImplementation.ManifestResources.cs
```diff
@@ -82,17 +82,12 @@ private unsafe Stream ReadResourceFromBlob(ResourceInfo resourceInfo)
                 throw new BadImageFormatException();
             }
 
-            checked
+            if (resourceInfo.Index < 0 || resourceInfo.Length < 0 || (uint)(resourceInfo.Index + resourceInfo.Length) > cbBlob)
```
MichalStrehovsky, line 85: Since you're on this - could you make this an assert? This info comes from the executable image we're currently running - it's not user provided. (Same for the if check above this.)

MichalStrehovsky (approved):
Thank you!

src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/CallInterceptor.cs
```diff
@@ -111,28 +111,13 @@ public unsafe void SetVar<T>(int index, T value)
             Unsafe.Write<T>((void*)address, value);
         }
 
-        /// <summary>
```
jkotas, line 114: I think the whole file is unused code.

AntonLapounov: Great observation. Removed the file.

src/coreclr/nativeaot/System.Private.Reflection.Execution/src/Internal/Reflection/Execution/ExecutionEnvironmentImplementation.ManifestResources.cs
AntonLapounov: Converted to an assert but kept the previous `if` unchanged.

PR #820, opened 2021-03-16
Generate a suppressible warning if compilation fails

MichalStrehovsky:
IL1005 is a [preexisting error code](https://github.com/mono/linker/blob/main/docs/error-codes.md#il1005-error-processing-method-method-in-assembly-assembly) that seems fitting.

I extracted COM into a separate code but I'm not firm on that one.


jkotas (approved):
Thank you

PR #821, opened 2021-03-17
Merge from dotnet/runtime:main

jkotas:


jkotas:
@MichalStrehovsky Could you please take a look at this build break?

MichalStrehovsky:
Hm, this is still failing in ObjWriter build. I could have a look but my local git repo is in a messy state right now because I'm looking at System.Runtime tests. And I don't have another local clone because I foolishly created a 200 GB partition for BootCamp 5 years ago and I've been having disk space trouble for the last 2 years as a result. I have a new machine on order, finally (that will probably fix other things too).

jkotas:
@AntonLapounov Could you please take a look at the ObjWriter build break?  I have changed it to use your new script (https://github.com/dotnet/runtimelab/pull/821/commits/e2ffe45afc83795a71a4f8d4d19c546e08054153#diff-8bb555c9ef0d91e66546d2fa904a310f4d22f129e191dff6b12e4fbdfe65e321L49) but there are some more changes needed to make it work.

src/coreclr/tools/aot/ObjWriter/build.cmd
```diff
@@ -46,10 +46,7 @@ git apply "%ScriptDir%llvm.patch" || goto Error
 
 :PatchApplied
 
-:: Set CMakePath by evaluating the output from set-cmake-path.ps1
-call "%RepoRoot%src\coreclr\setup_vs_tools.cmd" || goto Error
-for /f "delims=" %%a in ('powershell -NoProfile -ExecutionPolicy ByPass "& ""%RepoRoot%eng\native\set-cmake-path.ps1"""') do %%a
-echo Using CMake at "%CMakePath%"
+call "%RepoRoot%eng\native\init-vs-env.cmd" || goto Error
```
AntonLapounov, line 49: I think we should keep the `for /f` and `echo` lines. When `init-vs-env.cmd` is called without arguments, it does not set `CMakePath`.

AntonLapounov:
I am not sure how to remove `698b161` (right before my fix) from the list of commits.

jkotas:
Fixed. Thanks!

PR #825, opened 2021-03-17
Implement better GetExceptionForHRInternal

MichalStrehovsky:
This is taken from Mono.

We have an implementation in .NET Native that seems to be more complete but requires porting more code. We can probably get by with Mono's "good enough".

We could alternatively block tests that test this on COM support. There's a bunch of them. Not sure what's the point of this API without COM.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
@@ -187,9 +187,6 @@ namespace System.Runtime.InteropServices
         {
             switch (errorCode)
             {
-                case HResults.S_OK:
-                case HResults.S_FALSE:
-                    return null;
                 case HResults.COR_E_AMBIGUOUSMATCH:
                     return new System.Reflection.AmbiguousMatchException();
                 case HResults.COR_E_APPLICATION:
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
```diff
@@ -185,10 +185,163 @@ public static unsafe void StructureToPtr(object structure, IntPtr ptr, bool fDel
 
         internal static Exception GetExceptionForHRInternal(int errorCode, IntPtr errorInfo)
         {
-            return new COMException()
+            switch (errorCode)
             {
-                HResult = errorCode
-            };
+                case HResults.S_OK:
```
jkotas, line 190: The success cases are checked by the caller.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
```diff
@@ -185,10 +185,163 @@ public static unsafe void StructureToPtr(object structure, IntPtr ptr, bool fDel
...
+                    return new System.UnauthorizedAccessException();
+                //case HResults.COR_E_UNSUPPORTEDFORMAT:
+                case HResults.COR_E_VERIFICATION:
+                    return new System.Security.VerificationException();
+                //case HResults.E_INVALIDARG:
+                case HResults.E_NOTIMPL:
+                    return new System.NotImplementedException();
+                //case HResults.E_POINTER:
+                case HResults.RO_E_CLOSED:
+                    return new System.ObjectDisposedException("");
```
jkotas, line 340: Is this sufficient to make all tests pass? I see that Mono has the mapping for a bunch more HRESULTs, and there are tests validating it (e.g. src\libraries\System.Runtime\tests\System\IO\FileLoadException.InteropTests.cs).

It may be nice to move the Mono impl into the shared partition.

jkotas:
> Not sure what's the point of this API without COM.

It is sometimes used to convert Windows errors codes into Exceptions. I think one possible middle ground would be to handle all native Windows error codes, like `E_POINTER`; but not the managed-specific ones like `COR_E_VERIFICATION`.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
MichalStrehovsky: Yeah, I didn't rerun the tests with this. I foolishly deleted the entries for which we didn't have the HRESULT definition. We also need to copy the HRESULT definitions and keep those all.

MichalStrehovsky:
>  I think one possible middle ground would be to handle all native Windows error codes, like E_POINTER; but not the managed-specific ones like COR_E_VERIFICATION.

Yeah, but we have tests for that junk so then we would need to block those tests.

jkotas (approved):
It would be really nice to refactor the Mono impl in dotnet/runtime and move it under libraries so that it can be shared instead of this...

MichalStrehovsky:
> It would be really nice to refactor the Mono impl in dotnet/runtime and move it under libraries so that it can be shared instead of this...

Filed an issue on it; I'll get to it later. My local clone of runtimelab is in a messy state because I have temporary workarounds and fixes for various things as I'm digging through 2,800 System.Runtime test failures (down to 240). This particular bucket is causing a lot of noise so I want to get over with. Switching branches is super painful and I need to make forward progress on this before I lose my mind.

PR #828, opened 2021-03-17
Add support for default parameters in Delegate.DynamicInvoke

MichalStrehovsky:
I've been postponing this - in .NET Native this was implemented by the compiler emitting a special encoding for parameters with default values that we would interpret at runtime. I think the position was "Delegate.DynamicInvoke is not reflection".

This position is indefensible in the presence of `CustomConstantAttribute` that .NET Native apparently didn't handle because we have no test coverage for it in dotnet/runtime (filed an issue on that).

So implementing this through reflection. This likely results in size-on-disk savings since delegates no longer have GC threadstatic bases and a vtable slot is gone.


jkotas (approved):
Thanks

PR #829, opened 2021-03-17
Relax signed/unsigned checks in Array.Copy

MichalStrehovsky:
Port of https://github.com/dotnet/coreclr/pull/25209.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Array.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Array.CoreRT.cs
@@ -365,9 +365,17 @@ namespace System
                 }
                 else if (sourceElementEEType.IsPrimitive && destinationElementEEType.IsPrimitive)
                 {
-                    // The only case remaining is that primitive types could have a widening conversion between the source element type and the destination
-                    // If a widening conversion does not exist we are going to throw an ArrayTypeMismatchException from it.
-                    CopyImplPrimitiveTypeWithWidening(sourceArray, sourceIndex, destinationArray, destinationIndex, length, reliable);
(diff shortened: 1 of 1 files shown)


jkotas (approved):
Thanks!

MichalStrehovsky:
Sigh, I'm still digging through like 280 failures in System.Runtime and this is now failing a test because I allowed long/ulong/IntPtr/UIntPtr to be interchangeable (on x64), but CoreCLR still disallows it.

@jkotas Should we allow that on CoreCLR side? I think it passes the litmus test of:

```csharp
int[] dest = DoTheMagicTrick();
int[] src = new int[] { 7 };
// I would never expect this to throw
Array.Copy(src, dest, 1);
```

<details><summary>CLICK ME TO SEE MAGIC</summary>

```csharp
static int[] DoTheMagicTrick() => (int[])(object)new IntPtr[] { IntPtr.Zero };
```
</details>

MichalStrehovsky:
Huh, when did CoreCLR break the desktop compatible behavior for this? @jkotas do you know if that was intentional?

I now get

System.InvalidCastException: 'Unable to cast object of type 'System.IntPtr[]' to type 'System.Int64[]'.'

So this also has a different fix on the NativeAOT side then.

jkotas:
I do not think `int[]/long[]` <-> `IntPtr[]` casting was ever allowed, in either .NET Framework nor .NET Core. What is the desktop compabible behavior you have in mind?

jkotas:
As ok, I see what you mean. I do not know where it changed, but I do not think it makes sense to fix it to behave same as .NET Framework. I think `int[]/long[]` <-> `IntPtr[]` is better disallowed.

MichalStrehovsky:
I think it got broken in the pull request I'm mentioning above: dotnet/coreclr#25209.

I'll disable this casting on NativeAOT too then.

PR #842, opened 2021-03-22
S390x fixes

uweigand:
* Do not append "Mono" to Microsoft.NETCore.App.Runtime package names

* Handle s390x architecture in CheckArchitectureTests.cs
  (Fixes System.Runtime.InteropServices.RuntimeInformation tests)

* Add supported s390x distros to runtimeGroups.props
  (Fixes System.Data.Common tests)

### DIFF
--- a/src/libraries/Microsoft.NETCore.Platforms/pkg/runtime.compatibility.json
+++ b/src/libraries/Microsoft.NETCore.Platforms/pkg/runtime.compatibility.json
@@ -31,6 +31,18 @@
     "any",
     "base"
   ],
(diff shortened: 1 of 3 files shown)


uweigand:
FYI @akoeplinger - With this patch and the changes already in runtime (once the network patch is merged), all remaining test failures on s390x should be gone, except for the crypto tests (I'll post another patch for those).

akoeplinger (approved):
Looks good but we're actually discussing adding the `.Mono` to all runtime packs based on the mono runtime rather than the mix we have today so this might change again in the future.

PR #843, opened 2021-03-22
RFC: S390x openssl workaround

uweigand:
Set key and IV simultaneously in AesGcm.Unix.cs
* Work around problems in OpenSSL implementations on some targets
  (Similar to existing workaround in AesCcm.Unix.cs)

Do not use EvpCipherReset for CFB ciphers
* Work around problems in OpenSSL implementation on some targets
  (Similar to existing workaround in AppleCCCryptor.cs)

### DIFF
--- a/src/libraries/Native/Unix/System.Security.Cryptography.Native/pal_evp_cipher.c
+++ b/src/libraries/Native/Unix/System.Security.Cryptography.Native/pal_evp_cipher.c
@@ -4,9 +4,55 @@
 #include "pal_evp_cipher.h"
 
 #include <assert.h>
+#include <string.h>
 
 #define SUCCESS 1
 #define KEEP_CURRENT_DIRECTION -1
+
+// With some versions of OpenSSL on some IBM Z models, when
+// hardware-accelerated cryptography routines are enabled,
+// calls to EVP_CipherInit_ex may have no effect unless a key
+// is provided.  To work around this problem, we pass any key
+// that was provided earlier to all subsequent EVP_CipherInit_ex
+// calls as well.  This requires saving that key in an extra
+// buffer, which we stash into the EVP_CIPHER_CTX using the
+// "app_data" mechanism.
+#ifdef __s390x__
+#define CIPHERINIT_REQUIRES_KEY
+#endif
+
+#ifdef CIPHERINIT_REQUIRES_KEY
+static uint8_t*
+SavedKey_Allocate(EVP_CIPHER_CTX* ctx)
+{
+    size_t keyLength = (size_t)EVP_CIPHER_CTX_key_length(ctx);
+    uint8_t* keyBuf = calloc(1, keyLength);
+    if (keyBuf)
+    {
+        EVP_CIPHER_CTX_set_app_data(ctx, keyBuf);
+    }
+    return keyBuf;
+}
+
+static void
+SavedKey_Free(EVP_CIPHER_CTX* ctx)
+{
+    size_t keyLength = (size_t)EVP_CIPHER_CTX_key_length(ctx);
+    uint8_t* keyBuf = EVP_CIPHER_CTX_get_app_data(ctx);
+    if (keyBuf)
+    {
+        EVP_CIPHER_CTX_set_app_data(ctx, NULL);
+        OPENSSL_cleanse(keyBuf, keyLength);
+        free(keyBuf);
+    }
+}
+
+static uint8_t*
+SavedKey_Lookup(EVP_CIPHER_CTX* ctx)
+{
+    return EVP_CIPHER_CTX_get_app_data(ctx);
+}
+#endif
 
 EVP_CIPHER_CTX*
 CryptoNative_EvpCipherCreate2(const EVP_CIPHER* type, uint8_t* key, int32_t keyLength, int32_t effectiveKeyLength, unsigned char* iv, int32_t enc)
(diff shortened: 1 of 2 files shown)


uweigand:
FYI @nealef @akoeplinger @directhex @steveisok - I'm not sure what the best way to handle this is.

We have a few failing cryptography test cases on s390x.  The problem is that the way the dotnet code base uses the OpenSSL EVP interface exposes issues/bugs in the underlying implementation.  This implementation is platform-specific, and the specific conditions when those bugs are triggered vary per platform.

Note that dotnet already today has some workarounds for those issues, in cases where they trigger on Linux/x86 or Apple platforms.  However, on Linux/s390x the issues trigger in a few different cases, where there are no workarounds in dotnet.

This patch adds workarounds in those cases, similar to the existing ones.  It fixes all remaining test case failures on s390x.

Do you think this is the correct approach?  If so, should be included in the s390x feature branch only, or should it go to the main runtime branch?

If this is not the correct approach, should we instead disable the test cases?  Or else disable the platform-specific crypto exploitation code in OpenSSL that causes the issues (e.g. by exporting OPENSSL_s390xcap=z900 when running the tests)?

akoeplinger:
I think it depends on whether you think the conditions these tests exercise are important for real-world usage or not. If not then I'd simply disable the tests.

If we do need to kepp the workarounds then we should make them specific to s390x so other platforms aren't impacted.

uweigand:
> I think it depends on whether you think the conditions these tests exercise are important for real-world usage or not. If not then I'd simply disable the tests.

Yes, this does look important for real-world usage.  Pretty much any use of GCM would simply fail, and many uses of CFB as well. We are planning to fix this in the underlying OpenSSL implementation, so at some point those workarounds would no longer be needed, but it'll take time until that change made its way into all the distros out there ...
 
> If we do need to kepp the workarounds then we should make them specific to s390x so other platforms aren't impacted.

The workaround doesn't really have any impact (on either performance or functionality), except for making the code logic a little bit more complicated.   Adding special coding for platforms wouldn't help with the latter, only make it even worse, I think.

akoeplinger:
I don't really have enough context to make a good decision here. @bartonjs what do you think?

bartonjs:
Since the native shim is a different compiled output (by virtue of being for a different architecture), can the fixes be made there, instead (where they'd already have some preprocessor conditionals available)?  e.g. try using `EVP_CIPHER_CTX_original_iv(ctx)` as part of the implementation for reset.

We're happy with NOT saving the keys as managed fields, so if at all possible we don't want these changes.

uweigand:
> Since the native shim is a different compiled output (by virtue of being for a different architecture), can the fixes be made there, instead (where they'd already have some preprocessor conditionals available)? e.g. try using `EVP_CIPHER_CTX_original_iv(ctx)` as part of the implementation for reset.

I can try if I can make it work just inside the shim.  The problem is that with the current implementation of the s390x accelerated OpenSSL functions, we need to pass in the key again to perform a reset, and -as opposed to the IV- there is no interface to retrieve the key, so it has to be stored _somewhere_.

If we have to do this in the shim layer, I guess we'd need to allocate extra memory there to save the key to be re-used.  It might be possible to stash this in the context via `EVP_CIPHER_CTX_set_app_data` -- I'll have a try and see if this works.

> We're happy with NOT saving the keys as managed fields, so if at all possible we don't want these changes.

Agreed that saving the key is not ideal.  We're also working on a fix to the OpenSSL code to make these workarounds no longer needed, but it'll be a while until this has reached all the existing Linux distros.

uweigand:
> If we have to do this in the shim layer, I guess we'd need to allocate extra memory there to save the key to be re-used. It might be possible to stash this in the context via `EVP_CIPHER_CTX_set_app_data` -- I'll have a try and see if this works.

I've just updated the PR to use this approach in the shim layer instead of the workarounds in managed code.  This also fixes the problems I was seeing on IBM Z.

bartonjs (approved):
Looks reasonable, and well-contained, to me.

uweigand:
Just FYI: we're working in parallel on changing OpenSSL so this workaround should no longer be necessary in the future.   See the discussion here:  https://github.com/openssl/openssl/pull/14900

src/libraries/Native/Unix/System.Security.Cryptography.Native/pal_evp_cipher.c
```diff
@@ -4,10 +4,56 @@
...
 
+// With some versions of OpenSSL on some IBM Z models, when
+// hardware-accelerated cryptography routines are enabled,
+// calls to EVP_CipherInit_ex may have no effect unless a key
+// is provided.  To work around this problem, we pass any key
+// that was provided earlier to all subsequent EVP_CipherInit_ex
+// calls as well.  This requires saving that key in an extra
+// buffer, which we stash into the EVP_CIPHER_CTX using the
+// "app_data" mechanism.
+#ifdef __s390x__
```
akoeplinger, line 20: nit: I'd use `if defined(TARGET_S390X)` that we're using elsewhere rather than the compiler define.

uweigand:
> Just FYI: we're working in parallel on changing OpenSSL so this workaround should no longer be necessary in the future. See the discussion here: [openssl/openssl#14900](https://github.com/openssl/openssl/pull/14900)

Another update: the fix has been merged into OpenSSL mainline and 1.1.1.  Backport to Ubuntu is in progress and being tracked here: https://bugs.launchpad.net/ubuntu/+source/openssl/+bug/1931994

akoeplinger:
@uweigand awesome! do you think we still need this patch then?

uweigand:
> @uweigand awesome! do you think we still need this patch then?

Hopefully we won't need the patch, but we'll have to watch how quickly the distro backports proceed.

uweigand:
> > @uweigand awesome! do you think we still need this patch then?
> 
> Hopefully we won't need the patch, but we'll have to watch how quickly the distro backports proceed.

Ubuntu has now released updated openssl packages including the fix for 18.04, 20.04 and later.  The fix has also landed in RHEL8.5 and later.  At this point, we don't need this workaround any more.  Closing this PR.

PR #845, opened 2021-03-23
Add support for profile guided optimizations

### LINKED ISSUE #569: Consume PGO information in the compiler
We should be able to hook up the PGO consumption parts of the crossgen2 compiler into the NativeAOT compiler pretty cheaply. It's the same codebase after all.

The PGO data for the framework trained on workloads that we care about (e.g. ASP.NET worloads) is readily available as part of the build because it's the PGO data we ship the product with. PGO data for things like CoreLib should "mostly match". We won't have PGO data for System.Private assemblies or the managed parts of the NativeAOT runtime, but that might not matter that much.

Eventually, we can add support for PGO data collection too, but that can wait.

MichalStrehovsky:
Hooks up into PGO support in crossgen2 - the workflow is to generate profiling information with CoreCLR and consume it with the NativeAOT compiler.

Eventually, we should also add PGO data collection, but that's more than 2 hours of work.

I'm also adding the PGO data we use to build .NET 6 ReadyToRun assemblies to the package. This should be representative of some interesting workloads.

Fixes #569.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -107,6 +107,9 @@ The .NET Foundation licenses this file to you under the MIT license.
     <!-- Exclude System.IO.Compression.Native.dll -->
     <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\*.Native.dll" />
   </ItemGroup>
+  <ItemGroup Condition="'$(IlcPgoOptimize)' == 'true'">
+    <MibcFile Include="$(IlcPath)\mibc\*.mibc" />
+  </ItemGroup>
 
   <ItemGroup>
     <DefaultFrameworkAssemblies Include="@(FrameworkAssemblies)" />
@@ -143,6 +146,9 @@ The .NET Foundation licenses this file to you under the MIT license.
     <ItemGroup>
       <!-- Exclude System.IO.Compression.Native.dll -->
       <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\*.Native.dll" />
+    </ItemGroup>
+    <ItemGroup Condition="'$(IlcPgoOptimize)' == 'true'">
+      <MibcFile Include="$(IlcPath)\mibc\*.mibc" />
     </ItemGroup>
 
     <ItemGroup>
@@ -199,6 +205,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <IlcArg Include="@(IlcCompileInput)" />
       <IlcArg Include="-o:$(NativeIntermediateOutputPath)%(ManagedBinary.Filename)$(IlcOutputFileExt)" />
       <IlcArg Include="@(IlcReference->'-r:%(Identity)')" />
+      <IlcArg Include="@(MibcFile->'--mibc:%(Identity)')" />
       <IlcArg Condition="$(IlcGenerateMetadataLog) == 'true'" Include="--metadatalog:$(NativeIntermediateOutputPath)%(ManagedBinary.Filename).metadata.csv" />
       <IlcArg Condition="$(IlcMultiModule) == 'true'" Include="--multifile" />
       <IlcArg Condition="$(Optimize) == 'true'" Include="-O" />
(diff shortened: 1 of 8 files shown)


src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
```diff
@@ -1504,9 +1515,94 @@ private unsafe HRESULT allocPgoInstrumentationBySchema(CORINFO_METHOD_STRUCT_* f
...
         }
 
-        private HRESULT getPgoInstrumentationResults(CORINFO_METHOD_STRUCT_* ftnHnd, ref PgoInstrumentationSchema* pSchema, ref uint pCountSchemaItems, byte** pInstrumentationData)
+        private PgoSchemaElem[] getPgoInstrumentationResults(MethodDesc method)
         {
-            throw new NotImplementedException("getPgoInstrumentationResults");
+            return _compilation.ProfileData[method]?.SchemaData;
+        }
+
+        private HRESULT getPgoInstrumentationResults(CORINFO_METHOD_STRUCT_* ftnHnd, ref PgoInstrumentationSchema* pSchema, ref uint countSchemaItems, byte** pInstrumentationData)
```
MichalStrehovsky, line 1523: I'm moving all of this to Common in the runtime repo - we can delete all of this later.

src/installer/pkg/projects/Microsoft.DotNet.ILCompiler/Microsoft.DotNet.ILCompiler.pkgproj
```diff
@@ -26,6 +26,9 @@
       <File Include="$(CoreCLRAotSdkDir)*">
         <TargetPath>sdk</TargetPath>
       </File>
+      <File Include="$(MibcOptimizationDataDir)/$(TargetOS)/$(TargetArchitecture)/**/*.mibc">
```
MichalStrehovsky, line 29: These compress to 1.2 MB so it's not that terrible to include them.

jkotas (approved):
Nice!

jkotas:
Is there an example of code that this improves?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -107,6 +107,9 @@ The .NET Foundation licenses this file to you under the MIT license.
     <!-- Exclude System.IO.Compression.Native.dll -->
     <FrameworkAssemblies Include="$(IlcPath)\framework\*.dll" Exclude="$(IlcPath)\framework\*.Native.dll" />
   </ItemGroup>
+  <ItemGroup Condition="'$(IlcPgoOptimize)' == 'true'">
```
jkotas, line 110: Do we want this on by default?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -107,6 +107,9 @@ The .NET Foundation licenses this file to you under the MIT license.
...
+    <MibcFile Include="$(IlcPath)\mibc\*.mibc" />
```
jkotas, line 111: Do we plan to ship framework *.mibc files for crossgen2 as well? Or is crossgen2 going to read the mibc data attached to .dlls? I am wondering whether this should be on the same plan.

MichalStrehovsky: Crossgen2 embeds this data into ReadyToRun formats - we don't currently ship with the readytorun compiled framework because it's just a bunch of useless junk.

Yes, eventually, we'll probably want to parse this out of the R2R headers.

MichalStrehovsky:
> Is there an example of code that this improves?

I don't have anything readily available - I've been itching to write a jit-diff for NativeAOT.

I'm planning to get this enabled in the techempower runs and see what it does.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
MichalStrehovsky: > I've seen this asserting the JIT

(There's something wrong with the PGO data I assume because the assert is in the form of `assert(haveBlockCounts != haveEdgeCounts);`)

PR #847, opened 2021-03-23
Switch attribute equality to dotnet/runtime plan

MichalStrehovsky:
.NET Native injected methods into these to make equality not reflection based. dotnet/runtime just walks fields using reflection.

It's about the equal amount of pain to deal with either one. The reflection based implementation should work for the mainstream framework scenarios. It will become more tricky if we want to make reflection metadata more pay for play in the future. The IL Linker suppressions on this implementation assume we need field metadata to operate. That's not the case in NativeAOT.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/BlockedInternalsBlockingPolicy.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/BlockedInternalsBlockingPolicy.cs
@@ -281,12 +281,12 @@ namespace ILCompiler
                     return false;
                 }
 
-                // Exempt fields on custom attributes from blocking. We use a heuristic instead of
-                // unconditionally walking the base class hierarchy as a perf optimization.
-                // Reflection blocking will be active for framework assemblies and those follow the
-                // attribute naming guidelines.
-                if (owningType.Name.EndsWith("Attribute")
-                    && AttributeType != null
+                // Exempt fields on custom attributes from blocking.
+                // Attribute.Equals and Attribute.GetHashCode depends on being able to
+                // walk all fields on custom attributes using reflection.
+                // We're opening this hole in hopes that the fields won't have any "interesting"
+                // types (that would be themselves blocked). Doing that could be problematic.
+                if (AttributeType != null
                     && owningType.CanCastTo(AttributeType))
                 {
                     return false;


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/BlockedInternalsBlockingPolicy.cs
```diff
@@ -278,6 +280,18 @@ public override bool IsBlocked(FieldDesc field)
                 {
                     return false;
                 }
+
+                // Exempt fields on custom attributes from blocking. We use a heuristic instead of
+                // unconditionally walking the base class hierarchy as a perf optimization.
+                // Reflection blocking will be active for framework assemblies and those follow the
+                // attribute naming guidelines.
+                if (owningType.Name.EndsWith("Attribute")
```
jkotas, line 288:
```suggestion
                if (owningType.Name.EndsWith("Attribute", StringComparison.Ordinal)
```

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/BlockedInternalsBlockingPolicy.cs
```diff
@@ -278,6 +280,18 @@ public override bool IsBlocked(FieldDesc field)
                 {
                     return false;
                 }
+
+                // Exempt fields on custom attributes from blocking. We use a heuristic instead of
+                // unconditionally walking the base class hierarchy as a perf optimization.
```
jkotas, line 285: Is it really faster on averate to compare the names than just do `CanCastTo` ?

jkotas (approved):

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/BlockedInternalsBlockingPolicy.cs
```diff
@@ -278,6 +280,18 @@ public override bool IsBlocked(FieldDesc field)
...
+                // Reflection blocking will be active for framework assemblies and those follow the
+                // attribute naming guidelines.
+                if (owningType.Name.EndsWith("Attribute")
+                    && AttributeType != null
+                    && owningType.CanCastTo(AttributeType))
```
MichalStrehovsky, line 290:
```suggestion
                // Exempt fields on custom attributes from blocking.
                // Attribute.Equals and Attribute.GetHashCode depends on being able to
                // walk all fields on custom attributes using reflection.
                // We're opening this hole in hopes that the fields won't have any "interesting"
                // types (that would be themselves blocked). Doing that could be problematic.
                if (AttributeType != null
                    && owningType.CanCastTo(AttributeType))
```

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/BlockedInternalsBlockingPolicy.cs
MichalStrehovsky: Yeah, it might be premature. Just worried about a scenario where a generic type has a lot of fields, each of which is a deeply nested type.

This will currently only trigger in internals-reflection-blocked assemblies, which somewhat limits the fallout to framework only.

PR #849, opened 2021-03-24
Add support for Activator.CreateInstance of structs with ctors

MichalStrehovsky:
Detect the cases when:

* Struct has no parameterless ctor
* Struct has a private parameterless ctor
* Struct has a public parameterless ctor

I had to switch the marker method to be static because apparently the instance methods optimization was kicking in for the owning type and reduced the constructor method body into a throw helper that was folded with everything else.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
@@ -28,14 +28,12 @@ namespace System
         // This method is the public surface area. It wraps the CreateInstance intrinsic with the appropriate try/catch
         // block so that the correct exceptions are generated. Also, it handles the cases where the T type doesn't have
         // a default constructor.
+        //
+        // This method is intrinsic. The compiler might replace it with more efficient implementation.
         [DebuggerGuidedStepThrough]
+        [Intrinsic]
         public static unsafe T CreateInstance<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]T>()
         {
-            if (RuntimeHelpers.IsValueTypeWithNoParameterlessConstructor<T>())
-            {
-                return default;
-            }
-
             // Grab the pointer to the default constructor of the type. If T doesn't have a default
             // constructor, the intrinsic returns a marker pointer that we check for.
             IntPtr defaultConstructor = DefaultConstructorOf<T>();
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
@@ -29,47 +29,56 @@ namespace System
         // block so that the correct exceptions are generated. Also, it handles the cases where the T type doesn't have
         // a default constructor.
         [DebuggerGuidedStepThrough]
-        public static T CreateInstance<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]T>()
+        public static unsafe T CreateInstance<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]T>()
         {
-            if (!RuntimeHelpers.IsReference<T>())
+            if (RuntimeHelpers.IsValueTypeWithNoParameterlessConstructor<T>())
             {
-                // Early out for valuetypes since we don't support default constructors anyway for now.
-                // This lets codegens that expand IsReference<T> optimize away the rest of this code.
                 return default;
             }
-            else
+
+            // Grab the pointer to the default constructor of the type. If T doesn't have a default
+            // constructor, the intrinsic returns a marker pointer that we check for.
+            IntPtr defaultConstructor = DefaultConstructorOf<T>();
+
+            // Check if we got the marker back.
+            //
+            // TODO: might want to disambiguate the different cases for abstract class, interface, etc.
+            if (defaultConstructor == (IntPtr)(delegate*<Guid>)&MissingConstructorMethod)
+                throw new MissingMethodException(SR.Format(SR.MissingConstructor_Name, typeof(T)));
+
+            T t;
+            try
             {
-                // Grab the pointer to the default constructor of the type. If T doesn't have a default
-                // constructor, the intrinsic returns a marker pointer that we check for.
-                IntPtr defaultConstructor = DefaultConstructorOf<T>();
-
-                // Check if we got the marker back.
-                //
-                // TODO: might want to disambiguate the different cases for abstract class, interface, etc.
-                if (defaultConstructor == DefaultConstructorOf<ClassWithMissingConstructor>())
-                    throw new MissingMethodException(SR.Format(SR.MissingConstructor_Name, typeof(T)));
-
-                // Grab a pointer to the optimized allocator for the type and call it.
-                // TODO: we need RyuJIT to respect that RawCalliHelper doesn't do fat pointer transform
-                // IntPtr allocator = AllocatorOf<T>();
-                // T t = RawCalliHelper.Call<T>(allocator, EETypePtr.EETypePtrOf<T>().RawValue);
-                T t = (T)RuntimeImports.RhNewObject(EETypePtr.EETypePtrOf<T>());
-
-                try
+                // Call the default constructor on the allocated instance.
+                if (RuntimeHelpers.IsReference<T>())
                 {
-                    // Call the default constructor on the allocated instance.
+                    // Grab a pointer to the optimized allocator for the type and call it.
+                    IntPtr allocator = AllocatorOf<T>();
+                    t = RawCalliHelper.Call<T>(allocator, EETypePtr.EETypePtrOf<T>().RawValue);
                     RawCalliHelper.Call(defaultConstructor, t);
 
                     // Debugger goo so that stepping in works. Only affects debug info generation.
                     // The call gets optimized away.
                     DebugAnnotations.PreviousCallContainsDebuggerStepInCode();
-
-                    return t;
                 }
-                catch (Exception e)
+                else
                 {
-                    throw new TargetInvocationException(e);
+                    t = default;
+                    if (defaultConstructor != (IntPtr)(delegate*<Guid>)&ValueTypeWithNoConstructorMethod)
+                    {
+                        RawCalliHelper.Call(defaultConstructor, ref Unsafe.As<T, byte>(ref t));
+
+                        // Debugger goo so that stepping in works. Only affects debug info generation.
+                        // The call gets optimized away.
+                        DebugAnnotations.PreviousCallContainsDebuggerStepInCode();
+                    }
                 }
+
+                return t;
+            }
+            catch (Exception e)
+            {
+                throw new TargetInvocationException(e);
             }
         }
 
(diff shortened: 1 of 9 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
```diff
@@ -29,47 +29,56 @@ public static partial class Activator
         // block so that the correct exceptions are generated. Also, it handles the cases where the T type doesn't have
         // a default constructor.
         [DebuggerGuidedStepThrough]
-        public static T CreateInstance<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]T>()
+        public static unsafe T CreateInstance<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]T>()
         {
-            if (!RuntimeHelpers.IsReference<T>())
+            if (RuntimeHelpers.IsValueTypeWithNoParameterlessConstructor<T>())
```
jkotas, line 34: Do I read this right that this path is just an optimization, but otherwise redundant with `ValueTypeWithNoConstructorMethod` marker method?

Comment may be useful. It is not obvious that `IsValueTypeWithNoParameterlessConstructor` can lie.

src/coreclr/tools/Common/Compiler/TypeExtensions.cs
```diff
@@ -46,6 +46,25 @@ public static DefType GetClosestDefType(this TypeDesc type)
...
 
+        public static MethodDesc GetConstructorForCreateInstanceIntrinsic(this TypeDesc type)
+        {
+            MethodDesc ctor = type.GetDefaultConstructor();
+            if (ctor == null)
+            {
+                MetadataType activatorType = type.Context.SystemModule.GetKnownType("System", "Activator");
+                if (type.IsValueType && type.GetParameterlessConstructor() == null)
+                {
+                    ctor = activatorType.GetKnownMethod("ValueTypeWithNoConstructorMethod", null);
```
jkotas, line 57: This looks too tightly coupled with corelib for it to live in in this file. Is there a better place for it? E.g. Common/TypeSystem/IL/Stubs/RuntimeHelpersIntrinsics.cs 

Also, this looks too special to be an extension method.

jkotas (approved):
Thank you!

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
MichalStrehovsky: Yeah, I put it there so that it keeps working with universal shared code.

Thinking about this, maybe I should put the struct without constructors on the same plan as CoreCLR (just rewrite the entire method body of `CreateInstance`). The added helper is not very general purpose.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
```diff
@@ -28,48 +28,55 @@ public static partial class Activator
...
                     DebugAnnotations.PreviousCallContainsDebuggerStepInCode();
-
-                    return t;
                 }
-                catch (Exception e)
+                else
                 {
-                    throw new TargetInvocationException(e);
+                    t = default;
+                    if (defaultConstructor != (IntPtr)(delegate*<Guid>)&ValueTypeWithNoConstructorMethod)
```
jkotas, line 65: We can have just an empty method that does nothing for this case to avoid the Guid hack. The check can then be just remove here.

src/coreclr/tools/Common/TypeSystem/IL/CoreRTILProvider.cs
```diff
@@ -114,6 +114,28 @@ private MethodIL TryGetPerInstantiationIntrinsicMethodIL(MethodDesc method)
...
             switch (owningType.Name)
             {
+                case "Activator":
+                    {
+                        TypeSystemContext context = owningType.Context;
+                        if (methodName == "CreateInstance" && method.Signature.Length == 0 && method.HasInstantiation
+                            && method.Instantiation[0] is TypeDesc activatedType
+                            && activatedType != context.UniversalCanonType
+                            && activatedType.IsValueType
+                            && activatedType.GetParameterlessConstructor() == null)
```
jkotas, line 124: The R2R version of this method calls `GetDefaultConstructor`. Does it have a bug?

src/coreclr/tools/Common/TypeSystem/IL/CoreRTILProvider.cs
```diff
@@ -114,6 +114,28 @@ private MethodIL TryGetPerInstantiationIntrinsicMethodIL(MethodDesc method)
...
+                        {
+                            ILEmitter emit = new ILEmitter();
+                            ILCodeStream codeStream = emit.NewCodeStream();
+
+                            var t = emit.NewLocal(context.GetSignatureVariable(0, method: true));
+                            codeStream.EmitLdLoca(t);
+                            codeStream.Emit(ILOpcode.initobj, emit.NewToken(context.GetSignatureVariable(0, method: true)));
```
jkotas, line 131: I guess crossgen2 would choke on this, and that's why you choose to not have this in the ILProvider base type?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
MichalStrehovsky: Are there no concerns about the calling convention of the method?

I wanted to do that, but I can't express a parameterless constructor in a struct in C# yet and right now there's a mismatch.

src/coreclr/tools/Common/TypeSystem/IL/CoreRTILProvider.cs
MichalStrehovsky: Looks like it. dotnet/runtime#50228

src/coreclr/tools/Common/TypeSystem/IL/CoreRTILProvider.cs
MichalStrehovsky: Yep. The strategy is the same as crossgen2 but crossgen2 can't make up new tokens. I expect using the exact same strategy as crossgen2 (returning a different `EcmaMethodIL`) would lead to problems. The crossgen2 strategy is already leaking out in places and it's annoying (https://github.com/dotnet/runtime/blob/bcc7e0e0ce1758322015675356efc121b211e312/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs#L1238).

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs
jkotas: Ok, you are right it would not be an obvious improvement.

yowl:
@MichalStrehovsky I know this is a few months old now, but just getting round to merging it to NativeAot-LLVM.  I'm going to need to expand https://github.com/dotnet/runtimelab/blob/2fe0c20be81952fdfcc5b39b138ca98c759eb3b6/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Activator.CoreRT.cs#L93-L97 I suppose ?

yowl:
I've gone with 
```
                case "AllocatorOf":
                    if (metadataType.Namespace == "System" && metadataType.Name == "Activator" && method.Instantiation.Length == 1)
                    {
                        if (runtimeDeterminedMethod.IsRuntimeDeterminedExactMethod)
                        {
                            MetadataType helperType = _compilation.TypeSystemContext.SystemModule.GetKnownType("System.Runtime", RuntimeExport);
                            MethodDesc helperMethod = helperType.GetKnownMethod("RhNewObject", null);
                            PushExpression(StackValueKind.Int32, "RhNewObjectPtr", LLVMFunctionForMethod(helperMethod, helperMethod, null, false, null, helperMethod, out bool _, out LLVMValueRef _, out LLVMValueRef _), GetWellKnownType(WellKnownType.IntPtr));
                        }
                        else
                        {
                            // a value type <T> ?  TODO: write a test
                        }

                        return true;
                    }
                    break;
```
Which for the runtime determined types seems to work.

MichalStrehovsky:
It would be better to do this through `ReadyToRunHelperId.ObjectAllocator` like we do in the IL scanner and in the RyuJIT-based codegen. The new object helper will be different if the object is e.g. finalizable (See logic in JitHelper.GetNewObjectHelperForType).

yowl:
Thanks, will do.  But I have a question, this will go from `RawCalliHelper` into, for example, `RhpNewFast` via the generic dict lookup.  How do I know that this is a managed -> unmanaged call.  Do I need to generate a function that handles that transition and calling convention difference for each ExternSymbolNode (if a function)?

MichalStrehovsky:
This is not a p/invoke - it shouldn't need to do a transition.

yowl:
Thanks.  I've still got a problem with the methods returned from https://github.com/dotnet/runtimelab/blob/6f776969b674f5cf688de8443bc23ba747a1d5e2/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/JitHelper.cs#L309  The `IntPtr` ultimately created from the string is passed to https://github.com/dotnet/runtimelab/blob/6f776969b674f5cf688de8443bc23ba747a1d5e2/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/TypeLoaderExports.cs#L395 where the `IntPtr` is cast to a `delegate*` and called.  That is assumed to be a managed call in LLVM as per any other delegate and the shadow stack arg added as the first parameter.  `RhpNewFast` does not have that so the there is an argument mismatch.  I can either add that argument to these helpers for LLVM/Wasm in https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/coreclr/nativeaot/Runtime/portable.cpp (with the corresponding problem of adding that dummy parameter when not called via the helper), or maybe create a thunk for each one using a new node type and return the thunk name out of `GetNewObjectHelperForType`?

Is the thunk the better option, or is there something better I've missed?

yowl:
Or maybe better is to intercept the writing of the indirectionNodes and handle it all there.  I think I'll try that as it should only require changes to LLVM files.

MichalStrehovsky:
The reason why this is even possible on native architectures (do a managed calli into something written in C) is that we keep the managed calling convention very close to the native calling convention.

Now, I don't know much about WASM, but I understand C/C++ also has the concept of a shadow stack on WASM. That shadow stack is not part of the signature, which is why we have a mismatch here.

Would it be a big disadvantage if we used the same approach on the managed side? Could we even use the C/Emscripten shadow stack?

If not, another approach I was thinking about is that we could indicate this in the calling convention. RawCalliHelper is already special cased in the compiler: https://github.com/dotnet/corert/blob/c6af4cfc8b625851b91823d9be746c4f7abdc667/src/ILCompiler.Compiler/src/Compiler/Compilation.cs#L412-L421. We don't want to do a fat call for it because the alignment might indicate this is a fat call, but it never is. We could similarly encode the information that we don't want to do a fat call and also the target is in C. Roslyn recently introduced an extensible way to specify calling conventions for calli (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/function-pointers#function-pointers-1). We basically only need to add new types to System.Runtime.CompilerServices with a CallConv prefix and they can be used to indicate calling convention in C# (`delegate* unmanaged[Stdcall, SuppressGCTransition] <int, int>;` sample in the doc). We could have `delegate* unmanaged[RawCall, CManaged]` to indicate that. I've been meaning to switch RawCalliHelper to that scheme but didn't have time for it yet. It's not ideal, because there's a certain degree of fluidity in the helpers (we've been moving them from C to C# sometimes), but it might be workable too.

yowl:
@MichalStrehovsky "but I understand C/C++ also has the concept of a shadow stack" <- is this based on some of the comments in the mono code base?

MichalStrehovsky:
No, it's from my shallow knowledge of WASM. E.g. this paper calls it "unmanaged stack" https://www.usenix.org/system/files/sec20-lehmann.pdf.

yowl:
You are right I think. There are some globals (or thread local if threads are enabled I think) for this "unmanaged stack" which work in a similar fashion to the shadow stack that LLVM creates.  By using globals, the need to pass around the shadow stack is removed, so I think this could work.  I think it would be good to test the performance implications as we'd have to put things on this "unmanaged stack" manually in some cases.  E.g. If we passed an object reference, it would be an `i8*` in LLVM and an `i32` in Wasm so would go in Wasm's managed stack and not be visible to the conservative GC.  Also the GC stack scan would now cross more than just what we think should be tracked, anything the clang compiler decides need to go in the unmanaged stack would also get traversed.  We could test which performs better.

Its a fair bit of work, the use of the shadow stack predates my participation in the LLVM generation, so I'm going to go a bit slopey shoulders on this and suggest we put it off.  Unless you have objections to the workaround I have at https://github.com/yowl/runtimelab/blob/7580ab1e845cb96b772aaa5abff3b85ded380309/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs#L418-L452 ?

MichalStrehovsky:
> Its a fair bit of work, the use of the shadow stack predates my participation in the LLVM generation, so I'm going to go a bit slopey shoulders on this and suggest we put it off. Unless you have objections to the workaround I have at https://github.com/yowl/runtimelab/blob/7580ab1e845cb96b772aaa5abff3b85ded380309/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs#L418-L452 ?

Yep, I assumed it would be fair a bit of work, I just want to throw it out there as a possible solution that maps to how we solve it elsewhere (and why this is not a problem elsewhere). The workaround will work. I don't expect user code to ever do this (call unmanaged code as if it was managed code), so it's limited to code under our control.

yowl:
Well that workaround didn't last long - https://github.com/dotnet/runtimelab/pull/1371

MichalStrehovsky:
> Well that workaround didn't last long - #1371

I hope it's not too difficult to root cause. I'll try to cc you whenever this changes in the future.

PR #852, opened 2021-03-25
Put failing tests behind ActiveIssue

MichalStrehovsky:


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Enum.CoreRT.cs
```diff
@@ -222,7 +222,7 @@ public static string[] GetNames(Type enumType)
                 return enumType.GetEnumNames();
 
             if (!enumType.IsEnum)
-                throw new ArgumentException(SR.Arg_MustBeEnum);
+                throw new ArgumentException(SR.Arg_MustBeEnum, nameof(enumType));
```
MichalStrehovsky, line 225: Oh well. My local repo is a mess so I would prefer this is just taken for the ride.

src/libraries/System.Runtime/tests/System/Type/TypeTests.cs
```diff
@@ -527,6 +527,7 @@ public void GetTypeByName_Invalid(string typeName, Type expectedException, bool
         }
 
         [Fact]
+        [ActiveIssue("https://github.com/dotnet/runtimelab/issues/155" /* NativeAot */)]
         public void GetTypeByName_InvokeViaReflection_Success()
```
jkotas, line 531: What is the problem with this one?

jkotas (approved):
Thanks!

src/libraries/System.Runtime/tests/System/Type/TypeTests.cs
MichalStrehovsky: I didn't debug it, but the type is not assembly qualified so we're supposed to look in the assembly that is calling the API. I assume CoreCLR knows the calling assembly even when reflection is involved. I figure this was needed for CAS, but for native AOT I assume we can put this in the "never going to happen" bucket.

PR #856, opened 2021-03-25
Implement Exception.get_TargetSite

MichalStrehovsky:
Seemed easy enough.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Exception.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Exception.CoreRT.cs
@@ -22,7 +22,7 @@ namespace System
                 if (!HasBeenThrown)
                     return null;
 
-                return new StackFrame(_corDbgStackTrace[_idxFirstFreeStackTraceEntry - 1], needFileInfo: false).GetMethod();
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Exception.CoreRT.cs
```diff
@@ -15,9 +15,16 @@ namespace System
...
-        // seems more useful than throwing a PlatformNotSupportedException here.
-        public MethodBase TargetSite => null;
+        public MethodBase TargetSite
+        {
+            get
+            {
+                if (!HasBeenThrown)
+                    return null;
+
+                return new StackFrame(_corDbgStackTrace[_idxFirstFreeStackTraceEntry - 1], needFileInfo: false).GetMethod();
```
jkotas, line 25:
```suggestion
                return new StackFrame(_corDbgStackTrace[0], needFileInfo: false).GetMethod();
```
?

MichalStrehovsky: Yea. I don't know what I was thinking.

Also submitted https://github.com/dotnet/runtime/pull/50275 so that we have test coverage for the bug I almost introduced.

jkotas (approved):

PR #857, opened 2021-03-25
Allow open delegates to nullables

MichalStrehovsky:
Port of https://github.com/dotnet/runtime/pull/42837.

Won't win any perf prizes, but will do for now. I don't think the amount of work required for good support will ever meet the bar.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/MethodInfos/CustomMethodInvoker.cs
+++ b/src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/MethodInfos/CustomMethodInvoker.cs
@@ -53,8 +53,30 @@ namespace System.Reflection.Runtime.MethodInfos
         {
             if (_thisType.IsConstructedGenericType && _thisType.GetGenericTypeDefinition() == typeof(Nullable<>))
             {
-                // Desktop compat: MethodInfos to Nullable<T> methods cannot be turned into delegates.
-                throw new ArgumentException(SR.Arg_DlgtTargMeth);
+                if (isOpen)
+                {
+                    return DynamicDelegateAugments.CreateObjectArrayDelegate(Type.GetTypeFromHandle(delegateType),
+                        (args) =>
+                        {
+                            object[] arguments;
+                            if (args.Length > 1)
+                            {
+                                arguments = new object[args.Length - 1];
+                                Array.Copy(args, 1, arguments, 0, args.Length - 1);
+                            }
+                            else
+                            {
+                                arguments = Array.Empty<object>();
+                            }
+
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/MethodInfos/CustomMethodInvoker.cs
```diff
@@ -53,8 +53,30 @@ public sealed override Delegate CreateDelegate(RuntimeTypeHandle delegateType, o
         {
             if (_thisType.IsConstructedGenericType && _thisType.GetGenericTypeDefinition() == typeof(Nullable<>))
             {
-                // Desktop compat: MethodInfos to Nullable<T> methods cannot be turned into delegates.
-                throw new ArgumentException(SR.Arg_DlgtTargMeth);
+                if (isOpen)
+                {
+                    return DynamicDelegateAugments.CreateObjectArrayDelegate(Type.GetTypeFromHandle(delegateType),
```
jkotas, line 58: This won't work for methods that take refs, byref-like types, etc. Is that right?

jkotas:
> Won't win any perf prizes, but will do for now.

I agree that it makes the test pass, but I am not sure whether partial slow implementation like this is an improvement. It may be better to just keep it broken and wait until somebody actually needs it.

src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/MethodInfos/CustomMethodInvoker.cs
MichalStrehovsky: Yes, but Nullable doesn't have any of those. This only works for the following methods:

https://github.com/dotnet/corert/blob/24f1edc938af9954eac9a3385958787ca64956d5/src/System.Private.Reflection.Core/src/System/Reflection/Runtime/MethodInfos/CustomMethodMapper.Nullable.cs#L28-L96

MichalStrehovsky:
>  It may be better to just keep it broken and wait until somebody actually needs it.

That will just mean we'll have a bug open and an extra ActiveIssue that we will have roll around from repo to repo. It felt like adding localized code that we will never have to touch will save us time taking care and feeding a bug forever.

src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/MethodInfos/CustomMethodInvoker.cs
jkotas: ah ok, I thought that it is for more than that.

jkotas (approved):

PR #858, opened 2021-03-25
Throw better exception for byref-like invokes

MichalStrehovsky:
Before we assume code is missing, check whether any of the parameters is byref-like.


jkotas (approved):

src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/MethodInfos/RuntimeNamedMethodInfo.cs
```diff
@@ -305,7 +305,36 @@ private RuntimeTypeInfo[] GenericTypeParameters
...
+            if (invoker == null)
+            {
+                // If we have byref-like types in the signature, the reason we couldn't find an invoker
+                // is that.
+                bool hasByRefLikeParameter = false;
+                foreach (ParameterInfo parameter in GetParametersNoCopy())
+                {
+                    if (Unwrap(parameter.ParameterType).IsByRefLike)
+                    {
+                        hasByRefLikeParameter = true;
```
Suchiman, line 318:
```suggestion
                        hasByRefLikeParameter = true;
                        break;
```

MichalStrehovsky: Heh, I had it there and deleted it to save a line of code. This is already in the super slow path.

PR #859, opened 2021-03-25
More suppressions

MichalStrehovsky:
I'm not sure whether this is the best approach. Up for discussion.


jkotas (approved):
I think that this is fine.

PR #869, opened 2021-03-27
Add opt out for eager constructed type validation

MichalStrehovsky:
xunit Assert.Equals has a ridiculous implementation that checks whether the compared things implement `IEquatable` using reflection. The check is done by `MakeGenericType` the interface and checking assignability. Instead of writing RD.XML for all valuetypes used with Assert.Equals, just add an opt out from the eager validation. The created type will not be usable and reflection APIs will not have a great failure mode (e.g. `Array.CreateInstance` will complain about inability to get a `TypeHandle`), but it's a painless way to get xunit off the ground.


jkotas (approved):
Thanks

PR #870, opened 2021-03-27
Allocate GC statics on the pinned object heap

### LINKED ISSUE #841: Allocate GC statics on the pinned object heap
See discussion in https://github.com/dotnet/runtime/pull/49576. The GC static bases in NativeAOT don't suffer from the fragmentation concerns because we initialize them at process start and they live until the process finishes. We could avoid an indirection when accessing them.

MichalStrehovsky:
This allows us to get rid of a level of indirection when accessing GC static fields.

Fixes #841.

### DIFF
--- a/src/coreclr/nativeaot/Common/src/Internal/Runtime/CompilerHelpers/StartupCodeHelpers.cs
+++ b/src/coreclr/nativeaot/Common/src/Internal/Runtime/CompilerHelpers/StartupCodeHelpers.cs
@@ -6,6 +6,8 @@ using System.Runtime;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
+using Internal.Runtime.CompilerServices;
+
 using Debug = Internal.Runtime.CompilerHelpers.StartupDebug;
 
 namespace Internal.Runtime.CompilerHelpers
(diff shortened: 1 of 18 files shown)


MichalStrehovsky:
Cc @yowl - this is removing an indirection when accessing GC static bases - I assume the WASM side will be a one-liner delete, like in the R2R helper nodes in this pull request.

src/coreclr/nativeaot/Common/src/Internal/Runtime/CompilerHelpers/StartupCodeHelpers.cs
```diff
@@ -27,17 +29,25 @@ public static partial class StartupCodeHelpers
         /// </summary>
         private static int s_moduleCount;
 
+        /// <summary>
+        /// GC handle of an array with s_moduleCount elements, each representing and array of GC static bases of the types in the module.
```
jkotas, line 33:
```suggestion
        /// GC handle of an array with s_moduleCount elements, each representing an array of GC static bases of the types in the module.
```

src/coreclr/nativeaot/Runtime/GCHelpers.cpp
```diff
@@ -354,71 +356,52 @@ static Array* AllocateNewArrayImpl(Thread* pThread, EEType* pArrayEEType, uint32
...
-    {
-        max_object_size = (INT64_MAX - 7 - min_obj_size);
-    }
-    else
-#endif // HOST_64BIT
-    {
-        max_object_size = (INT32_MAX - 7 - min_obj_size);
-    }
-
-    if (size >= max_object_size)
```
jkotas, line 369: Is this checked somewhere else?

src/coreclr/nativeaot/Runtime/GCHelpers.cpp
```diff
@@ -354,71 +356,52 @@ static Array* AllocateNewArrayImpl(Thread* pThread, EEType* pArrayEEType, uint32
...
-    const int MaxArrayLength = 0x7FEFFFFF;
-    const int MaxByteArrayLength = 0x7FFFFFC7;
+    pArray->set_EEType(pArrayEEType);
+    pArray->InitArrayLength(numElements);
 
-    // Impose limits on maximum array length in each dimension to allow efficient
-    // implementation of advanced range check elimination in future. We have to allow
-    // higher limit for array of bytes (or one byte structs) for backward compatibility.
-    // Keep in sync with Array.MaxArrayLength in BCL.
-    if (size > MaxByteArrayLength /* note: comparing allocation size with element count */)
```
jkotas, line 381: Checked somewhere else?

Suchiman:
Does the removed indirection mean that statics can be debug evaluated now?

src/coreclr/nativeaot/Runtime/GCHelpers.cpp
MichalStrehovsky: Yup, it's checked in the newly factored out GcAllocInternal. We had all of this duplicated between RhpGcAlloc and AllocateNewArrayImpl. Instead of adding a third copy for AllocateNewObjectImpl, I'm funneling it all to the common helper. But please double check, I would hate to mess this up.

src/coreclr/nativeaot/Runtime/GCHelpers.cpp
MichalStrehovsky: This was duplicated as above.

MichalStrehovsky:
> Does the removed indirection mean that statics can be debug evaluated now?

I think it would also be possible before although this still has one extra indirection compared to statics in C++ (statics in C++ are simply allocated in the data section of the executable - here they're on the GC heap and symbols can't refer to the GC heap directly).

I think we're not generating the debug info in the format the debugger expects. I haven't look what VC++ generates.

jkotas (approved):
Nice!

PR #872, opened 2021-03-28
Remove support for binary serialization reflection analysis

MichalStrehovsky:
IL Linker is also removing this support in https://github.com/mono/linker/pull/1925.

We could keep the reflection blocking exemption, but there's a workaround: just turn off reflection blocking.


jkotas (approved):
Good!

PR #873, opened 2021-03-28
Fix cmake scripts to work with new VS dogfood

jkotas:
The default layout have changed and so the hardcoding *.dir for the location of .pdb for static libraries does not work anymore


src/coreclr/nativeaot/Directory.Build.props
```diff
@@ -45,6 +45,9 @@
     <DisableImplicitConfigurationDefines>true</DisableImplicitConfigurationDefines>
 
     <DocumentationFile>$(OutputPath)$(MSBuildProjectName).xml</DocumentationFile>
+
+    <!-- Send .deps.json to intermediate directory to avoid polluting published package -->
```
jkotas, line 49: Unrelated fix that I made along the way

MichalStrehovsky (approved):
Thank you!

PR #875, opened 2021-03-29
Delete RD.XML warning

MichalStrehovsky:
Libraries tests build with WarnAsError and just setting that property to false still doesn't unbreak the build. We could make this warning suppressible, but it needs an error code and that's just too much hassle.

We've been doing breaking changes and this is just going to be one of them at some point.


jkotas (approved):
Thanks

Issue #155, opened 2020-10-08
Limitations that we're not planning to address right now


MichalStrehovsky:
- [ ] Runtime doesn't wrap non-exception throws into a System.Exception
- [ ] Limited support for multi-dim arrays with non-zero lower bounds
- [ ] Type equivalence

tomrus88:
> * COM and SAFEARRAY marshalling
Does it mean WinForms and WPF won't be supported? Because I think both of those require COM support.

MichalStrehovsky:
We'll need to build tools that convert COM interop into something AOT friendlier before the AOT compiler ever sees it - similar to how [CsWinRT](https://github.com/microsoft/CsWinRT) does it for WinRT.

There's been a lot of discussions about this e.g. here https://github.com/dotnet/corert/issues/4219.

darkguy2008:
Sorry to bother, but what about `Assembly.Load`, `Assembly.LoadFrom` or `AssemblyName.GetAssemblyName` ? Since NativeAOT/CoreRT doesn't support Entity Framework, my idea was to put all the EF stuff in a secondary library and then call it with the Assembly functions or Reflection, but it seems it's not even possible. Such a deal-breaker.

However, investigating I found out that @tonerdo https://github.com/dotnet/corert/issues/5011#issuecomment-408715555 did have some success getting `Assembly.Load` to work with some kind of experimental code/libs. 

Is there any way to see this implemented in the short/mid term anytime soon? What can we do (or needs to be done) for this to work?

jkotas:
That would require ability to load managed plugins. We do not have plans to do that as part of this project. There is a lot of discussion on the reasons behind that in https://github.com/dotnet/runtimelab/issues/248.

hez2010:
Is `GetEntryAssembly().Location` the same? I hit this which returns a null instead of the entry assembly location.

MichalStrehovsky:
No managed assembly exists after compilation. Native EXE != assembly. This is the same as when publishing single file in .NET 5.

PR #876, opened 2021-03-29
Block strings intern test on #155

### LINKED ISSUE #823: Interned strings don't have same identity as string literals
Failing the System.Tests.StringTests.InternTest System.Runtime test.

We could probably fix this by walking the frozen objects region, looking for all strings in it and adding them to the intern table on first `String.Intern` call.

MichalStrehovsky:
Fixes #823.


jkotas (approved):
Thanks

PR #679, opened 2021-02-11
Reenable most tests on s390x

akoeplinger:

PR #884, opened 2021-03-30
Merge upstream runtime main to feature/s390x

akoeplinger:
/cc @uweigand @nealef


uweigand:
Thanks for doing the merge @akoeplinger !  With this version, it should be possible to remove all test exclusions except for the crypto ones (unless https://github.com/dotnet/runtimelab/pull/843 is also merged).

akoeplinger:
Interestingly we have a few System.Drawing.Common tests fail with this merge. Maybe something changed with the libgdiplus version on the bots?

akoeplinger:
Will merge this and disable System.Drawing.Common in https://github.com/dotnet/runtimelab/pull/679 and reenable the rest.

uweigand:
Hmm, I've seen the System.Drawing.Common failures when Mono is also installed on the machine; they seem to go away when it is uninstalled.  Not sure if this happened on your build machines ...

In any case, I agree we should disable the test for now; I'll have a look and see if I can fix it.

akoeplinger:
Ah ok, that makes sense. The System.Drawing.Common tests check whether libgdiplus is available and skips tests otherwise. I assume somebody installed mono (which includes libgdiplus) on the bots.

uweigand:
Huh, interesting.  Does this mean dotnet relies on (parts of) mono being available for some of its functionality, then?

uweigand:
Hi @akoeplinger I've now fixed all failures in the System.Drawing.Common tests.  However, the only bug in dotnet code is this one:
https://github.com/dotnet/runtime/pull/50499

The majority of the failures are actually caused by endian bugs in libgdiplus itself.  I've submitted pull requests here:
https://github.com/mono/libgdiplus/pull/697
https://github.com/mono/libgdiplus/pull/698

In addition, while not necessary to make the dotnet tests pass, when running the libgdiplus test suite I also found these:
https://github.com/mono/libgdiplus/pull/699
https://github.com/mono/libgdiplus/pull/700

uweigand:
Unfortunately this means that even when the libgdiplus fixes are merged upstream, we still need to keep the dotnet test disabled until the fixes are available in Linux distros (at least in the distro on the test machine).

akoeplinger:
@uweigand thanks, I've now merged the PRs :)

uweigand:
Thanks @akoeplinger !

PR #885, opened 2021-03-30
More fixes

MichalStrehovsky:
Enough to get all System.Collections and System.Threading tests pass.


src/libraries/Common/tests/TestUtilities/System/PlatformDetection.cs
```diff
@@ -112,6 +112,7 @@ public static bool IsDrawingSupported
 
         public static bool IsAssemblyLoadingSupported => !IsNativeAot;
         public static bool IsMethodBodySupported => !IsNativeAot;
+        public static bool IsDebuggerTypeProxyAttributeSupported => !IsNativeAot;
```
MichalStrehovsky, line 115: The aspect that is not supported is MakeGenericType of the proxy for the collection element types. We have no code to ensure that's possible.

jkotas (approved):

PR #887, opened 2021-03-30
Delete support for .NET Native import mechanism

MichalStrehovsky:
This is the "MrtImport" mechanism used to support SharedLibrary.dll in .NET Native. The only compilation models that are currently available in Native AOT is single file (single OBJ and multiple OBJ files).

The multi-obj files mode ensures the compiler infrastructure for this is kept, should we ever want to restore this.


jkotas (approved):
Thanks

PR #888, opened 2021-03-30
Delete debug info generation inherited from .NET Native

MichalStrehovsky:
All of this requires debugger support that is specific to .NET Native UWP. This code is unused in Native AOT.

A potentially reusable part of this is the CodeView debug information encoder - but we currently use an equivalent piece of code in ObjWriter. It might become relevant as we move away from LLVM ObjWriter in the future.


jkotas (approved):
Thanks

PR #890, opened 2021-03-31
NativeAOT-LLVM: Add call helper for gc/nongc statics node/RhpAssignRef for STOREIND

yowl:
This PR adds support for a few more LIR instructions:

- Calls to the helpers for GC and non GC Static bases
- Add for pointers and ints -  LLVM GEPs
- CNSINT for TYP_INT, only 32bit not sure how to identify others yet, and for nullptrs TYP_REF when IconVal == 0
- STOREIND for storing a reference type, calls RhpAssignRef.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -3,6 +3,7 @@
 
 #ifdef TARGET_WASM
 #include <string.h>
+#include "alloc.h"
 #include "compiler.h"
 #include "block.h"
 #include "gentree.h"
@@ -13,6 +14,7 @@
 #include "llvm/IR/Function.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/Bitcode/BitcodeWriter.h"
+#include <unordered_map>
 
 using llvm::Function;
 using llvm::FunctionType;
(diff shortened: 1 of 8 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -87,59 +139,156 @@ void EmitDoNothingCall(llvm::IRBuilder<>& builder)
...
+    // dont think RhpAssignRef will ever reverse PInvoke, so probably ok not to store the shadow stack here
+    builder.CreateCall(getOrCreateRhpAssignRef(), ArrayRef<Value*>{address, toStore});
+}
+
+Value* visitNode(llvm::IRBuilder<> &builder, GenTree* node)
 {
     switch (node->OperGet())
     {
+        case GT_ADD:
+            return buildAdd(builder, node, getTreeIdValue(reinterpret_cast<GenTreeOp*>(node)->gtOp1), getTreeIdValue(reinterpret_cast<GenTreeOp*>(node)->gtOp2));
```
SingleAccretion, line 220: Why not `node->AsOp()` instead of the `reinterpret_cast<GenTreeOp*>`?

yowl: Thanks, fixed as suggested.

src/coreclr/jit/llvm.cpp
```diff
@@ -87,59 +139,156 @@ void EmitDoNothingCall(llvm::IRBuilder<>& builder)
...
+        if (gtCall->gtCallMethHnd == _compiler->eeFindHelper(CORINFO_HELP_READYTORUN_STATIC_BASE))
+        {
+            const char* symbolName = (*_getMangledSymbolName)(_thisPtr, gtCall->gtEntryPoint.handle);
+            Function* llvmFunc = _module->getFunction(symbolName);
+            if (llvmFunc == nullptr)
+            {
+                llvmFunc = Function::Create(FunctionType::get(Type::getInt8PtrTy(_llvmContext), ArrayRef<Type*>(Type::getInt8PtrTy(_llvmContext)), false), Function::ExternalLinkage, 0U, symbolName, _module); // TODO: ExternalLinkage forced as linked from old module
+            }
+            // replacement for _info.compCompHnd->recordRelocation(nullptr, gtCall->gtEntryPoint.handle, IMAGE_REL_BASED_REL32);
+            (*_addCodeReloc)(_thisPtr, gtCall->gtEntryPoint.handle);
```
jkotas, line 166: I am not able to figure out what `addCodeReloc` does or why it is needed. Could you please shed some light on it for me?

src/coreclr/jit/llvm.cpp
```diff
@@ -87,59 +139,156 @@ void EmitDoNothingCall(llvm::IRBuilder<>& builder)
...
+    Value* address = getTreeIdValue(node->AsOp()->gtOp1);
+    Value* toStore = getTreeIdValue(node->AsOp()->gtOp2);
+    // TODO: delete this temporary check for supported stores
+    if (!address->getType()->isPointerTy() || !toStore->getType()->isPointerTy())
+    {
+        //RhpAssignRef only
+        failFunctionCompilation();
+    }
+
+    // dont think RhpAssignRef will ever reverse PInvoke, so probably ok not to store the shadow stack here
```
jkotas, line 211: Yes, that's correct.

yowl:
What I'm trying to do is say to Ilc "The helper is used externally so make sure you add it as a dependency, along with the static bases".  If this line is removed, then in the clrjit module the declaration is created 

```
declare i8* @__GetGCStaticBase_HelloWasm_Program(i8*)
```
But there is nothing in the Ilc module, and at link time:

```
warning: undefined symbol: __GetGCStaticBase_HelloWasm_Program (referenced by top-level compiled C/C++ code)
warning: undefined symbol: __GetNonGCStaticBase_HelloWasm_NonBeforeFieldInitTest (referenced by top-level compiled C/C++ code)
```
(with a missing function error at runtime)

If the helper is used, then the static base node, e.g. `@__NonGCStaticBase_HelloWasm_NonBeforeFieldInitTest___SYMBOL` also is required, and the helper may be the only use of the static base node so I add those as dependencies also.  

The helper is written in `LLVMObjectWriter` and at that point it is too late to add the dependencies.  For non wasm this happens I think through `recordRelocation`  and 

https://github.com/dotnet/runtimelab/blob/bd215e1ff87ed33bc5645d2d572088e2a16c3965/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs#L3310

and later in the same function

https://github.com/dotnet/runtimelab/blob/bd215e1ff87ed33bc5645d2d572088e2a16c3965/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs#L3360

jkotas:
> The helper is used externally so make sure you add it as a dependency

Maybe a better name would be addDependency, but then it gets added to the relocs array so it would not actually help.

LGTM. Thanks for the explanation.

PR #892, opened 2021-03-31
Skip test on s390x that runs into OOM killer

akoeplinger:


akoeplinger:
@uweigand looks like this fixes the OOM, no idea why it doesn't happen on other architectures but it might give you some pointers to investigate.

uweigand:
I see.  Well, this test by construction touches 6.5 GB of data in memory.   Given that the s390x machine currently only has 4GB memory + 4GB swap, it is easy to see how the test can bring it to its limits.   I'm assuming the Intel machines where you're running the test simply have more memory available.

Hopefully next week we'll be able to bump up the available memory on the s390x machine as well, and then this problem will go away.

akoeplinger:
Hm yeah, maybe this test should be improved to not try to allocate so much data on Linux/Android where we have the OOM killer. The test tries to catch the .NET OutOfMemoryException but that's probably not enough in these cases.

uweigand:
Hi @akoeplinger , now that the helix workers have been upgraded, I think this patch can be reverted again.

PR #893, opened 2021-03-31
Delete .NET Native funceval support

MichalStrehovsky:
* We don't have a debugger to take advantage of this right now.
* Depends on System.Private.Debug that was not open sourced.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/Augments/RuntimeAugments.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/Augments/RuntimeAugments.cs
@@ -1075,11 +1075,6 @@ namespace Internal.Runtime.Augments
             return typeHandle.LastResortToString;
         }
 
-        public static void RhpSendCustomEventToDebugger(IntPtr payload, int length)
-        {
-            RuntimeImports.RhpSendCustomEventToDebugger(payload, length);
-        }
-
         public static IntPtr RhHandleAlloc(object value, GCHandleType type)
         {
             return RuntimeImports.RhHandleAlloc(value, type);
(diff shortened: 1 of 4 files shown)


src/coreclr/nativeaot/Runtime/threadstore.cpp
```diff
@@ -325,7 +303,7 @@ void ThreadStore::WaitForSuspendComplete()
 
 void ThreadStore::InitiateThreadAbort(Thread* targetThread, Object * threadAbortException, bool doRudeAbort)
 {
-    SuspendAllThreads(/* waitForGCEvent = */ false, /* fireDebugEvent = */ false);
+    SuspendAllThreads(/* waitForGCEvent = */ false);
     // TODO: consider enabling multiple thread aborts running in parallel on different threads
```
jkotas, line 307: Everything related to threadabort is also only for FuncEval.

jkotas (approved):
LGTM. Thanks

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/Augments/RuntimeAugments.cs
```diff
@@ -1080,12 +1080,6 @@ public static void RhpSendCustomEventToDebugger(IntPtr payload, int length)
             RuntimeImports.RhpSendCustomEventToDebugger(payload, length);
```
jkotas, line 1080: RhpSendCustomEventToDebugger is also only for FuncEval

src/coreclr/nativeaot/Runtime/threadstore.cpp
MichalStrehovsky: Looks like that would require undoing https://github.com/dotnet/corert/pull/4100/commits/a61fac9dce2b1873a61cbe70489d8c0ecdad3ad4. It's a bit convoluted, so maybe a separate pull request at another time.

PR #895, opened 2021-03-31
Sync files shared with ILLinker

MichalStrehovsky:


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReferenceSource/ValueNode.cs
```diff
@@ -77,7 +80,7 @@ public ValueNode ()
 		/// applied so that each 'unique value' can be considered on its own without regard to the structure that led to
 		/// it.
 		/// </summary>
-		public UniqueValueCollection UniqueValues {
+		public UniqueValueCollection UniqueValuesInternal {
```
jkotas, line 83: Just curious - why was this renamed to `UniqueValuesInternal` ?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/MethodBodyScanner.cs
```diff
@@ -762,17 +759,29 @@ Stack<StackSlot> currentStack
             }
         }
 
-        private static void ScanLdtoken(MethodIL methodBody, TypeDesc type, Stack<StackSlot> currentStack)
+        private static void ScanLdtoken(MethodIL methodBody, object operand, Stack<StackSlot> currentStack)
         {
-            if (type.IsGenericParameter)
+            if (operand is TypeDesc type)
```
jkotas, line 764: Nit: You can use the new shine pattern matching in cases like this:

switch (operand)
{
    case TypeDesc type:
    ...
    case MethodDesc method:
    ...
    default:
    ....
}
```

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReferenceSource/ValueNode.cs
MichalStrehovsky: I wanted to ask about that one, but I'm trying to hold back on questions that don't matter in the end so that there's not so much feedback. But I'm curious too. @vitek-karas?

vitek-karas: Because there's also an extension method called `UniqueValues` which is public. All external callers should use the method, because it can handle `null` as the input value (it's an extension method, so `null` `this` is possible and handled). This property is only so that it's possible to implement the method.

This caused bugs in the data flow code since there were couple of places where by accident we called the property instead of the method, and thus would end up with a NullRef if the value happened to be null (which can happen in some special cases).

The name is not ideal - but better than before.

PR #903, opened 2021-04-02
Speed up ObjectWriter a bit

MichalStrehovsky:
We were computing sorted sets for stuff that should be just directly addressable.

Improves E2E compilation wallclock time by about 1%. Also reduces GC allocations by about 2%.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ObjectWriter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ObjectWriter.cs
@@ -941,9 +941,9 @@ namespace ILCompiler.DependencyAnalysis
         {
             int neededInterruptionsBytes = nodeContents.Data.Length + 1;
             if (_byteInterruptionOffsets == null || _byteInterruptionOffsets.Length < neededInterruptionsBytes)
-                _byteInterruptionOffsets = new bool[nodeContents.Data.Length + 1];
+                _byteInterruptionOffsets = new bool[neededInterruptionsBytes];
             else
-                Array.Fill(_byteInterruptionOffsets, false, 0, neededInterruptionsBytes);
+                Array.Clear(_byteInterruptionOffsets, 0, neededInterruptionsBytes);
 
             foreach (var reloc in nodeContents.Relocs)
             {


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ObjectWriter.cs
```diff
@@ -1010,6 +1007,7 @@ public static void EmitObject(string objectFilePath, IEnumerable<DependencyNode>
                     objectWriter.SetSection(section);
                     objectWriter.EmitAlignment(nodeContents.Alignment);
 
+                    objectWriter._byteInterruptionOffsets = new bool[nodeContents.Data.Length + 1];
```
jkotas, line 1010: Can we reallocate the array only if it is not big enough, and just zero-fill it otherwise?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ObjectWriter.cs
```diff
@@ -1010,6 +1007,12 @@ public static void EmitObject(string objectFilePath, IEnumerable<DependencyNode>
                     objectWriter.SetSection(section);
                     objectWriter.EmitAlignment(nodeContents.Alignment);
 
+                    int neededInterruptionsBytes = nodeContents.Data.Length + 1;
+                    if (objectWriter._byteInterruptionOffsets == null || objectWriter._byteInterruptionOffsets.Length < neededInterruptionsBytes)
+                        objectWriter._byteInterruptionOffsets = new bool[nodeContents.Data.Length + 1];
+                    else
+                        Array.Fill(objectWriter._byteInterruptionOffsets, false);
```
jkotas, line 1014: This can be pretty wasteful if there is a large node followed by a lot of small nodes.

MichalStrehovsky: Yeah, I already realized because it got pretty slow. We have a 400 kB node in the metadata blob :)

jkotas, line 1014: It may be nice to move this into `ResetByteRunInterruptionOffsets` method

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ObjectWriter.cs
```diff
@@ -937,13 +937,17 @@ private ObjectNodeSection GetSharedSection(ObjectNodeSection section, string key
...
 
-        public void ResetByteRunInterruptionOffsets(Relocation[] relocs)
+        public void ResetByteRunInterruptionOffsets(ObjectData nodeContents)
         {
-            _byteInterruptionOffsets.Clear();
+            int neededInterruptionsBytes = nodeContents.Data.Length + 1;
+            if (_byteInterruptionOffsets == null || _byteInterruptionOffsets.Length < neededInterruptionsBytes)
+                _byteInterruptionOffsets = new bool[nodeContents.Data.Length + 1];
+            else
+                Array.Fill(_byteInterruptionOffsets, false, 0, neededInterruptionsBytes);
```
jkotas, line 946:
```suggestion
                Array.Clear(_byteInterruptionOffsets, 0, neededInterruptionsBytes);
```

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ObjectWriter.cs
```diff
@@ -937,13 +937,17 @@ private ObjectNodeSection GetSharedSection(ObjectNodeSection section, string key
             return new ObjectNodeSection(standardSectionPrefix + section.Name, section.Type, key);
         }
 
-        public void ResetByteRunInterruptionOffsets(Relocation[] relocs)
+        public void ResetByteRunInterruptionOffsets(ObjectData nodeContents)
         {
-            _byteInterruptionOffsets.Clear();
+            int neededInterruptionsBytes = nodeContents.Data.Length + 1;
+            if (_byteInterruptionOffsets == null || _byteInterruptionOffsets.Length < neededInterruptionsBytes)
+                _byteInterruptionOffsets = new bool[nodeContents.Data.Length + 1];
```
jkotas, line 944:
```suggestion
                _byteInterruptionOffsets = new bool[neededInterruptionsBytes];
```

jkotas (approved):
Thank you!

PR #904, opened 2021-04-02
Handle nullable fields in ValueType.GetHashCode

MichalStrehovsky:
Noticed this as I was deleting the .NET Native codepath. We would crash when calling `GetHashCode` on a struct that doesn't override `GetHashCode` and has a nullable field whose value is null.


jkotas (approved):
Thanks

PR #911, opened 2021-04-04
Use static CRT for ObjWriter, fix parallelism factor

### LINKED ISSUE #795: Remove ILCompiler's external dependency on Visual C++ CRT
`Objwriter.dll` fails to load if there is no `vcruntime140.dll` installed on the machine.

AntonLapounov:
Use the static CRT for `objwriter.dll` (fixes #795).  That increases the ARM64 library size by about 1.2%.

The `-j 10` CMake argument has been copy-pasted without giving a thought and it is probably not a good setting for our CI machines.  On my quad-core Windows machine with HT enabled, parallelism factor 5 gave the best release build time (times for factors 4 to 10 were pretty close though); however, passing `-m` to MSBuild was even better and that matches what we do for building native components in the Runtime repo.

### DIFF
--- a/samples/prerequisites.md
+++ b/samples/prerequisites.md
@@ -28,10 +28,6 @@ Notes:
 - The `--installPath` option affects Build Tools installation only. Visual Studio Installer is always installed into
 the `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer` directory.
 
-If you have no `vcruntime140.dll` library installed in your `%SystemRoot%\System32` directory, you also need to install Visual C++ 2019 Redistributable package, which ILCompiler depends on:
-- [vc_redist.x64.exe](https://aka.ms/vs/16/release/vc_redist.x64.exe) (for x64 machine)
-- [vc_redist.arm64.exe](https://aka.ms/vs/16/release/vc_redist.arm64.exe) (for ARM64 machine)
-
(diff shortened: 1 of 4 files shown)


AntonLapounov:
While this is not very accurate, I compared build times for PR 905 and this PR spent between `Build files have been written to: .../llvm/build/...` and `[100%] Built target objwriter` (Linux/macOS) or `objwriter.vcxproj -> ...\objwriter.dll` (Windows) markers:

Build | PR 905 | This PR | Change
-- | --: | --: | --:
Linux arm64 Release | 19:18 | 21:35 | +11.8%
Linux_musl x64 Release | 30:40 | 29:05 | −5.2%
Linux x64 Release | 29:27 | 24:19 | −17.4%
OSX x64 Release | 31:16 | 21:02 | −32.7%
Windows x64 Release | 25:20 | 23:49 | −6.0%
Windows arm64 Release | 26:10 | 29:28 | +12.6%

That shows 7.6% geomean improvement in objwriter build time.  I looked at details of Windows arm64 Release build and it seems the machine in this PR was slower in general, e.g., CMake configuration phase for LLVM, which is non-parallel, took 2.2 times longer comparing to PR 905.

MichalStrehovsky (approved):
Thank you!

PR #918, opened 2021-04-05
Fix casting

kant2002:
On the method `[linq2db]LinqToDB.Linq.Expressions+<>c.<.cctor>b__88_838()`
During attempt to compile accessor for property
`[S.P.CoreLib]System.Nullable``1<float64>.get_Value()`

I receive following error `An unhandled exception of type 'System.InvalidCastException' occurred in System.Private.CoreLib.dll
Unable to cast object of type 'Internal.TypeSystem.InstantiatedType' to type 'Internal.TypeSystem.Ecma.EcmaType'.`

Offending code looks approximately this way.
```
class Sql {
	public static double? Truncate(double? value)
	{
		if (value.HasValue)
		{
			return Math.Truncate(value.Value);
		}
		return null;
	}
}

public static LambdaExpression L<T1, TR>(Expression<Func<T1, TR>> func)
{
	return func;
}

internal LambdaExpression SomeFunction()
{
	return L((double p) => Sql.Truncate(p).Value);
}
```

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/EcmaExtensions.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/EcmaExtensions.cs
@@ -67,7 +67,7 @@ namespace ILCompiler.Dataflow
         public static PropertyPseudoDesc GetPropertyForAccessor(this MethodDesc accessor)
         {
             var ecmaAccessor = (EcmaMethod)accessor.GetTypicalMethodDefinition();
-            var type = (EcmaType)accessor.OwningType;
+            var type = (EcmaType)accessor.OwningType.GetTypeDefinition();
             var reader = type.MetadataReader;
             var module = type.EcmaModule;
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/EcmaExtensions.cs
```diff
@@ -67,7 +67,7 @@ public static PropertyPseudoDesc GetProperty(this MetadataType mdType, string na
         public static PropertyPseudoDesc GetPropertyForAccessor(this MethodDesc accessor)
         {
             var ecmaAccessor = (EcmaMethod)accessor.GetTypicalMethodDefinition();
-            var type = (EcmaType)accessor.OwningType;
+            var type = (EcmaType)accessor.OwningType.GetTypeDefinition();
```
MichalStrehovsky:
```suggestion
            var type = (EcmaType)ecmaAccessor.OwningType;
```

MichalStrehovsky (approved):
Thank you!

PR #919, opened 2021-04-05
Enable cross-architecture publishing

AntonLapounov:
Enable cross-architecture publishing (e.g., win-x64 -> win-arm64) provided the project file has a reference to the ILCompiler package for the _host_ architecture. For example,
```xml
<PackageReference Include="Microsoft.DotNet.ILCompiler;runtime.win-x64.Microsoft.DotNet.ILCompiler" Version="6.0.0-preview.4.21203.1" />
```
would allow one to run `dotnet publish -r win-arm64` on a win-x64 host.

This is implemented by calling the `ilc` executable from the host architecture's ILCompiler package instead of the target architecture's one. This approach is rough, but does not require any packaging changes.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -208,8 +208,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <IlcArg Include="-o:$(NativeIntermediateOutputPath)%(ManagedBinary.Filename)$(IlcOutputFileExt)" />
       <IlcArg Include="@(IlcReference->'-r:%(Identity)')" />
       <IlcArg Include="@(MibcFile->'--mibc:%(Identity)')" />
-      <IlcArg Include="--targetarch:$(PlatformTarget)" />
       <IlcArg Condition="$(IlcGenerateMetadataLog) == 'true'" Include="--metadatalog:$(NativeIntermediateOutputPath)%(ManagedBinary.Filename).metadata.csv" />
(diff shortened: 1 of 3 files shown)


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -207,6 +209,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <IlcArg Include="@(IlcReference->'-r:%(Identity)')" />
       <IlcArg Include="@(MibcFile->'--mibc:%(Identity)')" />
       <IlcArg Condition="$(IlcGenerateMetadataLog) == 'true'" Include="--metadatalog:$(NativeIntermediateOutputPath)%(ManagedBinary.Filename).metadata.csv" />
+      <IlcArg Condition="$(PlatformTarget) != ''" Include="--targetarch:$(PlatformTarget)" />
```
MichalStrehovsky, line 212: If this is just for the multifile build of the coreclr test tree, maybe we could just pass a PlatformTarget here:

https://github.com/dotnet/runtimelab/blob/ee45ce1bbeb58d982956120723b41eb13455bf08/src/tests/run.proj#L430

MichalStrehovsky (approved):
Thank you! Nice trick!

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
AntonLapounov: Thank you for the hint. We should add a check that `PlatformTarget` is set and then we can remove this `Condition`. Going to do it the next time.

Issue #882, opened 2021-03-30
Options how library can speify their own RD.xml


kant2002:
My understanding that rd.xml right now is application level directive. Is is possible/valuable to be able provide alternative to rd.xml via attributes which can be picked up by ILC ? Somewhat related to https://github.com/dotnet/corert/issues/5001.
I personally start thinking about that when hit [issue with BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet/issues/1680) and playing with Dappers.PerformanceTests

jkotas:
cc @MichalStrehovsky

MichalStrehovsky:
The RD.XML file format is not great, but we don't have anything better. The "way forward" is the ILLink descriptors format, which is not great either, but already in use in the mainstream runtimes, which means we're stuck with it for the time being. We don't parse the rooting part yet in ILC. It also cannot express instantiated generics, why is why I haven't even bothered with it. I'm reluctant to allow RD.XML in more places.

Can you work around with code. E.g.

```csharp
if (string.Empty.Length > 0)
{
    Type t = Type.GetType("Foo, FooAssembly");
    t.GetMethods();
}
```

This will keep all public methods on Foo in FooAssembly. It's a misuse of the reflection dataflow analysis that the compiler does, but it works pretty well.

We don't model MakeGenericType/MakeGenericMethod. I would want to add that at some point.

MichalStrehovsky:
https://github.com/mono/linker/blob/main/docs/data-formats.md#descriptor-format

This can be embedded in the assembly as a managed resource. ILC currently doesn't parse this though. Also, there's no way to express generic instantiations. One would think that Unity (who is using this as well and has similar problems for full AOT in their IL2CPP) would have already contributed such extension, but they use this format + an additional text file for generics to specify what needs to be kept/generated, which is just lovely.

kant2002:
Do you want to support ILLink format in ILC in theory?

MichalStrehovsky:
> Do you want to support ILLink format in ILC in theory?

It's already used (in the embedded form) in e.g. ASP.NET, so supporting it eventually is pretty much mandatory.

MichalStrehovsky:
No plans for RD.XML experiences.

PR #924, opened 2021-04-07
Basic support for ILLink.Descriptors.xml

kant2002:
- No support for embedding
- Same command line argument -x as for ILLink
- No wildcards in typenames
- No nested type translation to RD.xml filename
- No method, field, event, property elements
- No preserve="fields", preserve="methods" etc

Related to #882


src/coreclr/tools/aot/ILCompiler/ILLinkDescriptorRootProvider.cs
```diff
@@ -0,0 +1,139 @@
...
+using AssemblyName = System.Reflection.AssemblyName;
+
+namespace ILCompiler
+{
+    /// <summary>
+    /// Compilation root provider that provides roots based on the ILLink descriptor file format.
+    /// Only supports a subset of the ILLink Descriptor file format.
+    /// </summary>
+    /// <remarks>https://github.com/mono/linker/blob/main/docs/data-formats.md</remarks>
+    internal class ILLinkDescriptorRootProvider : ICompilationRootProvider
```
MichalStrehovsky: We have a base parser for the illink formats here: src\coreclr\tools\aot\ILCompiler.Compiler\Compiler\ProcessLinkerXmlBase.cs - this should ideally just derive from that class.

kant2002: @MichalStrehovsky this class is very simplistic and do not account for all possible options in the ILLink.Descriptor.xml
Usages which this parser does not support
- `<assembly preserve="all">`
- `<type fullName="XXX"/>` Does not include all methods in the processing.
- `<type preserve="nothing" />` or `<type preserve="all" />` too not supported.
Question following: 
- Should I try to account for these cases? 
- or I should create simple processing what's possible with current class and try to improve `ProcessLinkerXmlBase`?
- `ProcessLinkerXmlBase` has constructor parameter `_owningModule` which usage is not clear for me. Is this should be entry point for the application?

MichalStrehovsky: We can add whatever virtual methods are needed into ProcessLinkerXmlBase. ILLink has a [similar class](https://github.com/mono/linker/blob/main/src/linker/Linker.Steps/ProcessLinkerXmlBase.cs) that this one is inspired by (we're just not using XPath because WTF). That class is also used as a base for everything.

`_owningModule` is used for the "embedded as a resource in the assembly" case. When the XML is embedded the `<assembly>` element is not mandatory and `<type>` can exist without an owning `<assembly>`. `_owningModule` would be null if the XML is not embedded.

kant2002:
I think made a progress making reading files using `ProcessLinkerXmlBase`. I would like test a bit this implementation. 
Also I would like to have some automated testing for this feature, if that would be practical in your opinion.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILLinkDescriptorRootProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+            var processor = new ILLinkDescriptorProcessor(_context, xmlReader, rootProvider);
+            processor.ProcessXml();
+        }
+
+        private class ILLinkDescriptorProcessor : ProcessLinkerXmlBase
+        {
+            private IRootingServiceProvider _rootProvider;
+
+            internal ILLinkDescriptorProcessor(TypeSystemContext context, XmlReader reader, IRootingServiceProvider rootProvider)
+                : base(context, reader, null, ImmutableDictionary.CreateBuilder<string, bool>().ToImmutable())
```
MichalStrehovsky: We'll want to pass the feature switch values that got passed at the command line instead of passing an empty dictionary.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ProcessLinkerXmlBase.cs
```diff
@@ -59,6 +60,8 @@ private void ProcessAssemblies()
                 if (_reader.Name == "assembly")
                 {
                     string assemblyName = _reader.GetAttribute("fullname");
+                    if (assemblyName == null)
+                        throw new Exception("The \"fullname\" attribute is required on the \"assembly\" ILLink Directive.");
```
MichalStrehovsky: I haven't bothered with validation too much (the assumption was that we would be processing embedded manifests and those were all already checked by ILLink) - these should all actually generate warnings, with whatever warning code il linker generates. We don't want to crash the compiler on invalid input. Maybe let's not add the error handling as part of this pull request.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ProcessLinkerXmlBase.cs
```diff
@@ -106,23 +109,53 @@ protected ModuleDesc GetAssembly(AssemblyName name)
             return _context.ResolveAssembly(name);
         }
 
+        protected virtual void ProcessModuleMetadata(ModuleDesc assembly)
+        {
+        }
+
         private void ProcessAssembly(ModuleDesc assembly)
         {
+            ProcessModuleMetadata(assembly);
```
MichalStrehovsky: Maybe we should make ProcessAssembly virtual and move the new logic into the ILLink.Desciptors.xml processor.

The "no nested elements means all the types" logic only applies to ILLink.Descriptors and so does the `preserve` attribute.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ProcessLinkerXmlBase.cs
```diff
@@ -144,28 +177,44 @@ private void ProcessType(ModuleDesc assembly)
 
                 _reader.Read();
 
-                while (_reader.IsStartElement())
```
MichalStrehovsky, line 147: Maybe let's move the part after this line into a new `protected virtual void ProcessType(TypeDesc type)` and override it for the ILLink.Descriptors-specific logic.

src/coreclr/tools/aot/ILCompiler/Program.cs
```diff
@@ -184,6 +186,7 @@ private ArgumentSyntax ParseCommandLine(string[] args)
                 syntax.DefineOption("waitfordebugger", ref waitForDebugger, "Pause to give opportunity to attach debugger");
                 syntax.DefineOptionList("codegenopt", ref _codegenOptions, "Define a codegen option");
                 syntax.DefineOptionList("rdxml", ref _rdXmlFilePaths, "RD.XML file(s) for compilation");
+                syntax.DefineOptionList("x", ref _illinkDescriptorsFilePaths, "ILLink descriptor file(s) for compilation");
```
MichalStrehovsky: I'm not sure if we want to "waste" the `-x` option on this. Maybe `--linkdescriptor` would be better. I know il linker uses x, but we don't need to be compatible with the il linker command line.

MichalStrehovsky (commented):

kant2002:
@MichalStrehovsky can you take a look when you have time, please?

MichalStrehovsky:
> @MichalStrehovsky can you take a look when you have time, please?

Yes, sorry, I have it in my backlog.

The IL Linker codebase removed their dependency on XPath in the meantime so I've also been thinking whether more code sharing is warranted, similar to how we do this for [dataflow analysis](https://github.com/dotnet/runtimelab/tree/feature/NativeAOT/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow) and [Logging](https://github.com/dotnet/runtimelab/tree/feature/NativeAOT/src/coreclr/tools/Common/Compiler/Logging). I've been postponing thinking about all of this, sorry.

kant2002:
I'm not in rush with this, since WinForms take all my time. At the same time,this may serve as basis to be able use NativeAOT + BenchmarkNet + external libraries. So my hands are itching. Also I read some time ago, that you have life happens to you, so I operate under assumption that you still busy with whatever. Since right now issue not with PR itself, probably nothing which I can do for now, since I'm do not look at IL Linker, and can only trust your opinion on the matter.

kant2002:
@MichalStrehovsky can you take a look when you have time, please?

I may want less mentally exhausting work once I make console app on ARM running. That's important to be able test libraries using NaiveAOT using BDN without hacking it.

kant2002:
@MichalStrehovsky can you take a look when you have time, please? If you think this work should be replaced with other, in other repo, let me know. I would like to have Benchmark.DotNet work with NativeAOT in better format. Manually crafting rd.xml seems to be complicated process. I want simplify that.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ProcessLinkerXmlBase.cs
```diff
@@ -106,26 +107,31 @@ protected ModuleDesc GetAssembly(AssemblyName name)
             return _context.ResolveAssembly(name);
         }
 
-        private void ProcessAssembly(ModuleDesc assembly)
+        protected virtual (bool hasContent, string preserveAttribute) ProcessAssembly(ModuleDesc assembly)
         {
+            var preserveAttribute = _reader.GetAttribute("preserve");
```
MichalStrehovsky, line 112: We already moved past the `<assembly>` element at this point. We're reading `type` or whatever is nested under it. I assume the intention was to read `<assembly preserve="X">`, but that wouldn't work.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILLinkDescriptorRootProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    includeAllTypes = !hasContent;
+                }
+
+                if (includeAllTypes)
+                {
+                    foreach (TypeDesc type in ((EcmaModule)assembly).GetAllTypes())
+                    {
+                        ProcessType(assembly);
+                    }
```
MichalStrehovsky, line 75: Doing a `foreach`, but not actually doing anything with the `type` we get out of it.

MichalStrehovsky (commented):
I had some leftover comments from when I was looking at this earlier, but please just ignore them.

After looking at this more, I think I'll just redo the whole descriptor processing to use ProcessLinkerXmlBase.cs, DescriptorMarker.cs, etc from the linker repo directly (including the all the XPath garbage). I would rather us not spend time on this trying to fix the parser in the current form.

The linker file formats are hot garbage and there's no docs for any of them (the "examples" in the IL Linker repo are nowhere near complete as to what the hell the format is supposed to represent and there's no XSD schema or formal description of all the strings). The only way to make sure we'll not have to keep revisiting this is to take the IL Linker codebase, munge it to how we root things here, and be bug-for-bug compatible with the whole parser, including all the undocumented behaviors.

I don't know if it helps you with Benchmark.net, but you can do pretty good rooting through code. There's `DynamicDependencyAttribute` that can be e.g. attached to `Main` and it can root things. One can also root things by just calling `Type.GetType("SomeType, SomeAssembly").GetMethods()` (this will root all methods on SomeType). Trimming can also be configured through the CSPROJ with TrimmerRootAssembly and the other documented MSBuild items/properties.

kant2002:
Current problem with BDN is that it contains some embedded rd.xml and you either have to hack with source code to override rooting types. Previously we discuss that one way to have RD.xml embedded in the assemblies. Notable examples which create pain is FSharp.Core or Windows Forms(understandable cannot advocate this usecase) . Why I should copy all these helper snippets for each technology which I’m using into applications rd.xml if tooling can be teached to rely on library vendor to supply proper rd.xml

I would like to run ORM benchmarks with minimal efforts under NativeAOT which is problematic right now. 

I start thinking what if go with DefinitelyTyped way of handling that problem. I create some repository where FSharp.Core.rd.xml, System.Runtime.Extensions.xml should be published and community maintained. That way I can have multiple rd.xml files in the project splitted by technology and then I can improve BDN to bypass these files to generated test project. 

I just want seamless benchmarking for NativeAOT for complex projects with zero friction from developer

jkotas:
I believe the current plan is to just share the IL linker code verbatim for this.

PR #931, opened 2021-04-07
Fix write barriers for Windows ARM64

AntonLapounov:
The list of the registers that write barriers allow to trash is controlled by these `define`s:
https://github.com/dotnet/runtimelab/blob/39c78bf0cc9fdc5c444fc0d61f960681b7f6d05d/src/coreclr/jit/target.h#L1369-L1382
Make `x12` → `x17` and `x9` → `x12` replacements to preserve `x9` and obey the contract.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
+++ b/src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
@@ -32,37 +32,37 @@ INVALIDGCVALUE  EQU 0xCCCCCCCD
 
     MACRO
         ;; On entry:
-        ;;  $destReg: location to be updated (cannot be x9,x12)
-        ;;  $refReg: objectref to be stored (cannot be x9,x12)
+        ;;  $destReg: location to be updated (cannot be x12,x17)
+        ;;  $refReg: objectref to be stored (cannot be x12,x17)
         ;;
         ;; On exit:
-        ;;  x9,x12: trashed
+        ;;  x12,x17: trashed
         ;;  other registers are preserved
         ;;
         UPDATE_GC_SHADOW $destReg, $refReg
 
         ;; If g_GCShadow is 0, don't perform the check.
-        adrp    x9, $g_GCShadow
-        ldr     x9, [x9, $g_GCShadow]
-        cbz     x9, %ft1
+        adrp    x12, $g_GCShadow
+        ldr     x12, [x12, $g_GCShadow]
+        cbz     x12, %ft1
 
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
```diff
@@ -328,17 +330,17 @@ ExchangeRetry
...
+;;   x17  : trashed if WRITE_BARRIER_CHECK
 ;;
     LEAF_ENTRY RhpAssignRefArm64, _TEXT
     ALTERNATE_ENTRY RhpAssignRefAVLocation
     ALTERNATE_ENTRY RhpAssignRefX1AVLocation
 
         stlr    x15, [x14]
 
-        INSERT_UNCHECKED_WRITE_BARRIER_CORE x14, x15, x14
+        INSERT_UNCHECKED_WRITE_BARRIER_CORE x14, x15, x15
```
AntonLapounov, line 343: Trash `x15` instead of `x14` for consistency with the two other barriers, `RhpCheckedAssignRefArm64` and 
`RhpByRefAssignRefArm64`

src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
```diff
@@ -362,8 +363,6 @@ ExchangeRetry
 
         INSERT_CHECKED_WRITE_BARRIER_CORE x14, x15, x15
 
-        add     x14, x14, #8
```
AntonLapounov, line 365: `JIT_ByRefWriteBarrier` (= `RhpByRefAssignRefArm64`) is not using this function, so this increment is not needed.

src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.asm
```diff
@@ -250,7 +250,8 @@ INVALIDGCVALUE  EQU 0xCCCCCCCD
 ;;   x0  : original value of objectref
 ;;   x9  : trashed
```
AntonLapounov, line 251: `x9` is OK to trash in functions following the standard calling convention.

MichalStrehovsky (approved):
Thank you!

Are you planning to do the same fixes for the Unix version of this file?

AntonLapounov:
The Unix version does not trash `x9` unless you define `WRITE_BARRIER_CHECK`, which we never define for asm files at the moment.  We will need to reconcile Windows and Unix versions later.

PR #934, opened 2021-04-07
Enable ARM64 → x64 publishing

### LINKED ISSUE #933: Regression 4.21206.3
Error while compiling (regression was introduced with last build 4.21206.3)
> Cant find .ilc.rsp
\tools\ilc" @"obj\x64\Release\net5.0-windows\win-x64\native\project.ilc.rsp"" with code 3

Wrong path?

(no issues with 6.0.0-preview.4.21203.1)

AntonLapounov:
Enable ARM64 → x64 publishing and also fix #933. We cannot trust `RuntimeInformation.ProcessArchitecture` as it returns the `dotnet` binary's architecture. Nor can we trust `RuntimeInformation.OSArchitecture` as it may return `Architecture.X86` when running in x86 emulation mode on ARM64.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.targets
@@ -10,7 +10,8 @@
     <OSIdentifier Condition="'$(OSIdentifier)' == ''">linux</OSIdentifier>
 
     <RuntimeIlcPackageName>runtime.$(OSIdentifier)-$(PlatformTarget).Microsoft.DotNet.ILCompiler</RuntimeIlcPackageName>
-    <IlcHostArch Condition="'$(IlcHostArch)' == ''">$([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant)</IlcHostArch>
+    <IlcHostArch Condition="'$(IlcHostArch)' == '' and ('$(PROCESSOR_ARCHITECTURE)' == 'ARM64' or '$(PROCESSOR_ARCHITEW6432)' == 'ARM64')">arm64</IlcHostArch>
+    <IlcHostArch Condition="'$(IlcHostArch)' == ''">x64</IlcHostArch>
     <IlcHostPackageName>runtime.$(OSIdentifier)-$(IlcHostArch).Microsoft.DotNet.ILCompiler</IlcHostPackageName>
     <IlcCalledViaPackage>true</IlcCalledViaPackage>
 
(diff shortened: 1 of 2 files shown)


src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.targets
```diff
@@ -10,7 +10,8 @@
     <OSIdentifier Condition="'$(OSIdentifier)' == ''">linux</OSIdentifier>
 
     <RuntimeIlcPackageName>runtime.$(OSIdentifier)-$(PlatformTarget).Microsoft.DotNet.ILCompiler</RuntimeIlcPackageName>
-    <IlcHostArch Condition="'$(IlcHostArch)' == ''">$([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant)</IlcHostArch>
+    <IlcHostArch Condition="'$(IlcHostArch)' == '' and ('$(PROCESSOR_ARCHITECTURE)' == 'ARM64' or '$(PROCESSOR_ARCHITEW6432)' == 'ARM64')">arm64</IlcHostArch>
```
MichalStrehovsky, line 13:
```suggestion
    <!-- Cannot use RuntimeInformation.ProcessArchitecture because that's the architecture of the process running MSBuild; assume ARM64 on ARM64 and x64 otherwise. -->
    <IlcHostArch Condition="'$(IlcHostArch)' == '' and ('$(PROCESSOR_ARCHITECTURE)' == 'ARM64' or '$(PROCESSOR_ARCHITEW6432)' == 'ARM64')">arm64</IlcHostArch>
```

MichalStrehovsky (approved):
Thank you!

src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.targets
AntonLapounov: That is sort of obvious in retrospect.  I actually intended to use `RuntimeInformation.OSArchitecture` instead of `ProcessArchitecture`. The former would work for x86 and x64 processes running on x64; however, it would lie for x86 and x64 processes running on ARM64.  Unfortunately, the `PROCESSOR_*` environment variables also lie in case of an x64 process running on ARM64 with recent Windows preview builds.  However, in that case we are at least able to run the x64-hosted ILCompiler, it would be just slower.  As far as I can tell, reading from the registry
```PowerShell
Get-ItemPropertyValue 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment' 'PROCESSOR_ARCHITECTURE'
```
never lies; however, the .NET Core version of MSBuild comes with no registry provider, so we would have to read it by other means, e.g., `reg query`.  That is why I left the option to override `IlcHostArch` if needed.  We will need to review this later.

PR #941, opened 2021-04-09
Align Module.Name behavior with .NET 5 single file

MichalStrehovsky:
We were returning the `ScopeName`. Single-file .NET apps are documented to return `<Unknown>`.


jkotas (approved):
Thanks

PR #945, opened 2021-04-09
NativeAOT-LLVM: Add support for string constants in clrjit LLVM compilation

yowl:
depends on https://github.com/dotnet/runtimelab/pull/890

This PR adds support for accessing string constants from the clrjit/llvm compilation, specifically when `node->IsIconHandle(GTF_ICON_STR_HDL)` is true.  `Bool___cctor` and `String___cctor` can now be compiled.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -3,6 +3,7 @@
 
 #ifdef TARGET_WASM
 #include <string.h>
+#include "alloc.h"
 #include "compiler.h"
 #include "block.h"
 #include "gentree.h"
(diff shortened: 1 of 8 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -130,6 +140,22 @@ Function* getOrCreateRhpAssignRef()
...
+    if (sourceType == valueType)
+        return source;
+
+    Type::TypeID toStoreTypeID = sourceType->getTypeID();
+    Type::TypeID valueTypeKind = valueType->getTypeID();
+
+    if (toStoreTypeID == Type::TypeID::PointerTyID && valueTypeKind == Type::TypeID::PointerTyID)
+    {
+        return builder.CreatePointerCast(source, valueType, "CastPtr");
+    }
```
yowl, line 155: There's a lot more casts in Ilc, but for now pointer casts is all I'm doing.

jkotas (approved):
Thanks

PR #946, opened 2021-04-09
NativeAOT-LLVM: start support for CALL to user functions

yowl:
This PR adds the beginning of support for `CALL` to user functions.  I've only done functions signatures that don't require anything passed on the shadow stack.  Have added some support for GT_LCL_VAR and GT_STORE_LCL_VAR.

Added support for ReadyToRunHelperId.GetThreadStaticBase from clrjit

If you think these PRs are a bit short or long then let me know and I'll adjust what goes in each one.

Depends https://github.com/dotnet/runtimelab/pull/945

### DIFF
--- a/samples/HelloWorld/nuget.config
+++ b/samples/HelloWorld/nuget.config
@@ -4,6 +4,6 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
-    <add key="local" value="e:\github\runtimelab\artifacts\packages\Debug\Shipping" />
+    <add key="dotnet-experimental" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json" />
  </packageSources>
 </configuration>
--- a/samples/HelloWorld/nuget.config
+++ b/samples/HelloWorld/nuget.config
@@ -4,6 +4,6 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
-    <add key="dotnet-experimental" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json" />
+    <add key="local" value="e:\github\runtimelab\artifacts\packages\Debug\Shipping" />
  </packageSources>
 </configuration>
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -32,7 +32,8 @@ static void* _thisPtr; // TODO: workaround for not changing the JIT/EE interface
 static const char* (*_getMangledMethodName)(void*, CORINFO_METHOD_STRUCT_*);
 static const char* (*_getMangledSymbolName)(void*, void*);
 static const char* (*_addCodeReloc)(void*, void*);
-static char* _outputFileName;
+static const bool (*_isRuntimeImport)(void*, CORINFO_METHOD_STRUCT_*);
+static char*     _outputFileName;
 static Function* _doNothingFunction;
 
 Compiler::Info                            _info;
@@ -43,16 +44,21 @@ llvm::IRBuilder<>*                        _builder;
 std::unordered_map<unsigned int, Value*>* _sdsuMap;
 std::unordered_map<unsigned int, Value*>* _localsMap;
 
-extern "C" DLLEXPORT void registerLlvmCallbacks(void* thisPtr, const char* outputFileName, const char* triple, const char* dataLayout,
-    const char* (*getMangledMethodNamePtr)(void*, CORINFO_METHOD_STRUCT_*),
-    const char* (*getMangledSymbolNamePtr)(void*, void*),
-    const char* (*addCodeRelocPtr)(void*, void*)
+extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
+                                                const char* outputFileName,
+                                                const char* triple,
+                                                const char* dataLayout,
+                                                const char* (*getMangledMethodNamePtr)(void*, CORINFO_METHOD_STRUCT_*),
+                                                const char* (*getMangledSymbolNamePtr)(void*, void*),
+                                                const char* (*addCodeRelocPtr)(void*, void*),
+                                                const bool (*isRuntimeImport)(void*, CORINFO_METHOD_STRUCT_*)
     )
 {
     _thisPtr = thisPtr;
     _getMangledMethodName = getMangledMethodNamePtr;
     _getMangledSymbolName = getMangledSymbolNamePtr;
-    _addCodeReloc = addCodeRelocPtr;
+    _addCodeReloc         = addCodeRelocPtr;
+    _isRuntimeImport      = isRuntimeImport;
     if (_module == nullptr) // registerLlvmCallbacks is called for each method to compile, but must only created the module once.  Better perhaps to split this into 2 calls.
     {
         _module = new Module(llvm::StringRef("netscripten-clrjit"), _llvmContext);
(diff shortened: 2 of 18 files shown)


samples/HelloWorld/nuget.config
```diff
@@ -4,6 +4,6 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
-    <add key="dotnet-experimental" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json" />
+    <add key="local" value="e:\github\runtimelab\artifacts\packages\Debug\Shipping" />
```
kant2002: @yowl just nitpicking.

yowl: Thanks

src/coreclr/jit/llvm.cpp
```diff
@@ -175,6 +213,70 @@ Value* buildAdd(llvm::IRBuilder<>& builder, GenTree* node, Value* op1, Value* op
     return mapTreeIdValue(node->gtTreeID, builder.CreateGEP(op1, op2));
 }
 
+FunctionType* getFunctionSignature(CORINFO_CONST_LOOKUP callTarget)
```
jkotas: Unused?

src/coreclr/jit/llvm.cpp
```diff
@@ -175,6 +213,70 @@ Value* buildAdd(llvm::IRBuilder<>& builder, GenTree* node, Value* op1, Value* op
...
+        // added to the Ilc module
+        llvmFunc = Function::Create(getFunctionTypeForMethodHandle(gtCall->gtCallMethHnd), Function::ExternalLinkage,
+                                    0U, symbolName, _module);
+    }
+    std::vector<llvm::Value*> argVec;
+
+    // shadowstack arg first
+    argVec.push_back(_function->getArg(0));
+
+    // TODO Is Args() more appropriate than Operands() ?
```
jkotas: I am not sure. @AndyAyersMS What is the difference between gtCall `Args()` vs `Operands()`?

src/coreclr/jit/llvm.cpp
```diff
@@ -175,6 +213,70 @@ Value* buildAdd(llvm::IRBuilder<>& builder, GenTree* node, Value* op1, Value* op
...
+    {
+        return _builder->getInt({(unsigned int)t->getPrimitiveSizeInBits().getFixedSize(), (uint64_t)gt->AsIntCon()->gtIconVal, true});
+    }
+    failFunctionCompilation();
+}
+
+llvm::Value* buildUserFuncCall(GenTreeCall* gtCall, llvm::IRBuilder<>& builder)
+{
+    const char* symbolName = (*_getMangledSymbolName)(_thisPtr, gtCall->gtEntryPoint.handle);
+    // TODO: how to detect RuntimeImport ?
```
jkotas: A new JIT/EE interface flag or API

src/coreclr/jit/llvm.cpp
AndyAyersMS: `Operands()` is a superset of `Args()`, it gives you (in order)
* `this` pointer, if any;
* `Args()`
* `LateArgs()`
* call cookie (for CT_INDIRECT calls)
* call address (for CT_INDIRECT calls)
* control expression (for other sorts of indirect calls)
 
See 
https://github.com/dotnet/runtime/blob/ab820d7a68b777a2da3dd1d89d4018e9e3d3a0fe/src/coreclr/jit/compiler.hpp#L4432-L4472

src/coreclr/jit/llvm.cpp
```diff
@@ -175,6 +213,70 @@ Value* buildAdd(llvm::IRBuilder<>& builder, GenTree* node, Value* op1, Value* op
     return mapTreeIdValue(node->gtTreeID, builder.CreateGEP(op1, op2));
 }
 
+FunctionType* getFunctionSignature(CORINFO_CONST_LOOKUP callTarget)
+{
+    return FunctionType::get(Type::getInt8PtrTy(_llvmContext), ArrayRef<Type*>(Type::getInt8PtrTy(_llvmContext)), false);
+}
+
+Value* genTreeAsLlvmType(GenTree* gt, Type* t)
```
SingleAccretion: Nit: any reason not to use `tree`, `type` and `value` as variable names?

src/coreclr/jit/llvm.cpp
```diff
@@ -175,6 +213,70 @@ Value* buildAdd(llvm::IRBuilder<>& builder, GenTree* node, Value* op1, Value* op
...
+{
+    Value* v = getTreeIdValue(gt);
+    if (v->getType() == t)
+        return v;
+
+    if (gt->OperGet() == GT_CNS_INT && gt->gtType == var_types::TYP_INT)
```
SingleAccretion: Nit: `gt->IsIntegralConst() && gt->TypeIs(TYP_INT)`.

src/coreclr/jit/llvm.cpp
```diff
@@ -175,6 +213,70 @@ Value* buildAdd(llvm::IRBuilder<>& builder, GenTree* node, Value* op1, Value* op
...
+    {
+        return _builder->getInt({(unsigned int)t->getPrimitiveSizeInBits().getFixedSize(), (uint64_t)gt->AsIntCon()->gtIconVal, true});
```
SingleAccretion: Nit: `gtIconVal` -> `IconValue()`.

src/coreclr/jit/llvm.cpp
```diff
@@ -221,23 +327,80 @@ Value* buildCnsInt(llvm::IRBuilder<>& builder, GenTree* node)
     failFunctionCompilation();
 }
 
+Type*  getLLVMTypeForVarType(var_types type)
```
SingleAccretion: Nit: double space after `Type*`?

src/coreclr/jit/llvm.cpp
```diff
@@ -221,23 +327,80 @@ Value* buildCnsInt(llvm::IRBuilder<>& builder, GenTree* node)
...
+{
+    return castIfNecessary(builder, address, getLLVMTypeForVarType(type)->getPointerTo());
+}
+
+void castingStore(llvm::IRBuilder<>& builder, Value* toStore, Value* address, var_types type)
+{
+    builder.CreateStore(castIfNecessary(builder, toStore, getLLVMTypeForVarType(type)), castToPointerToVarType(builder, address, type));
+}
+
 void importStoreInd(llvm::IRBuilder<>& builder, GenTree* node)
```
SingleAccretion: Nit: make `node` a `GenTreeStoreInd*` and cast at the callsite? You can then use `Addr()` and `Data()` getters instead of `gtOp1` and `gtOp2`.

src/coreclr/jit/llvm.cpp
```diff
@@ -221,23 +327,80 @@ Value* buildCnsInt(llvm::IRBuilder<>& builder, GenTree* node)
...
     }
+    else
+    {
+        castingStore(builder, toStore, address, node->gtType);
+    }
+}
 
-    // RhpAssignRef will never reverse PInvoke, so do not need to store the shadow stack here
-    builder.CreateCall(getOrCreateRhpAssignRef(), ArrayRef<Value*>{address, castIfNecessary(builder, toStore, Type::getInt8PtrTy(_llvmContext))});
+Value* localVar(llvm::IRBuilder<>& builder, GenTree* tree)
```
SingleAccretion: Nit, ditto: make `tree` a `GenTreeLclVar`?

src/coreclr/jit/llvm.cpp
```diff
@@ -221,23 +327,80 @@ Value* buildCnsInt(llvm::IRBuilder<>& builder, GenTree* node)
...
+{
+    GenTreeLclVar* lclVar = tree->AsLclVar();
+    Value*         llvmRef = _localsMap->at(lclVar->GetLclNum());
+    mapTreeIdValue(tree->gtTreeID, llvmRef);
+    return llvmRef;
 }
 
-Value* visitNode(llvm::IRBuilder<> &builder, GenTree* node)
+Value* storeLocalVar(llvm::IRBuilder<>& builder, GenTree* tree)
```
SingleAccretion: Nit, ditto: make `tree` `GenTreeLclVar`?

yowl:
@SingleAccretion @jkotas Thanks for the feedback

src/coreclr/jit/llvm.cpp
yowl: I did this as a another method passed in via registerLlvmCallbacks.

jkotas (approved):
LGTM

PR #951, opened 2021-04-11
Document cross-architecture compilation

AntonLapounov:


docs/using-nativeaot/compiling.md
```diff
@@ -0,0 +1,76 @@
...
+<add key="dotnet-experimental" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json" />
+<add key="nuget" value="https://api.nuget.org/v3/index.json" />
+```
+
+If your project has no `nuget.config` file, it may be created by running
+```bash
+> dotnet new nuget
+```
+
+from the project's root directory. New package sources must be added after the `<clear />` element if you decide to keep it.
```
jkotas, line 18:
```suggestion
from the project's root directory.
```
I do not see a good reason for why this has to be after the `<clear />` element.

docs/using-nativeaot/compiling.md
```diff
@@ -0,0 +1,76 @@
...
+```bash
+> dotnet publish -r win-arm64 -c Release -p:DisableUnsupportedError=true
+```
+
+Similarly, to target linux-arm64 on a Linux x64 host, in addition to the `Microsoft.DotNet.ILCompiler` package reference, also add the `runtime.linux-x64.Microsoft.DotNet.ILCompiler` package reference to get the x64-hosted compiler:
+```xml
+<PackageReference Include="Microsoft.DotNet.ILCompiler; runtime.linux-x64.Microsoft.DotNet.ILCompiler" Version="6.0.0-preview.4.21210.1" />
+```
+
+You also need to specify the sysroot directory for Clang using the `SysRoot` property. For example, assuming you are using one of ARM64-targeting [Docker images](../workflow/building/coreclr/linux-instructions.md#Docker-Images) employed for cross-compilation by the Runtime repo, you may publish for linux-arm64 with the following command:
```
jkotas:
```suggestion
You also need to specify the sysroot directory for Clang using the `SysRoot` property. For example, assuming you are using one of ARM64-targeting [Docker images](../workflow/building/coreclr/linux-instructions.md#Docker-Images) employed for cross-compilation by this repo, you may publish for linux-arm64 with the following command:
```

jkotas (approved):
Looks great. Thank you!

docs/using-nativeaot/compiling.md
AntonLapounov: `dotnet new nuget` creates the `nuget.config` file with the `<clear />` element, which means "ignore the records above". If you add new package sources before the `<clear />` element, they will be ignored as I just confirmed. So you have to either remove the `<clear />` element (thus inheriting the global NuGet package sources) or, if you decide to keep `<clear />`, add new package sources after it.

PR #957, opened 2021-04-13
Skip RemoveAttributeInstances with arguments

MichalStrehovsky:
ILLinkTrim.Attributes format received support for removing attribute instances with specific values of their arguments.

The support for this in illink equals to almost 1000 lines of code. There was no data on how much size savings this can actually provide when it was introduced. My spider-sense says "miniscule savings" (possibly somewhere around 0.1%).

We have plenty of other possible sources of savings with a smaller implementation complexity. This code will make sure to ignore such entries.


jkotas (approved):
LGTM

PR #958, opened 2021-04-13
Don't allow canonical instantiations when reflection-rooting

### LINKED ISSUE #929: Generic virtual method pointer lookup failure in CreateParameterizedConstructor
## Generic virtual method pointer lookup failure in CreateParameterizedConstructor

Repro code: https://github.com/hez2010/SyaBot

1. `cd SyaBot.Server && dotnet publish -c Release -r win-x64`
2. Run nativeaot-built server first and then run client (`cd SyaBot.Client && dotnet run`)
3. See exception from server

Output from server:
```
Raw request string: {"Name":"hello","Data":null,"Type":"SyaBot.Shared.RegisterRequest","Id":"990e15e8-dcba-4d6d-96b8-1767365693b9"}
Request id: 990e15e8-dcba-4d6d-96b8-1767365693b9, type: SyaBot.Shared.RegisterRequest
Generic virtual method pointer lookup failure.

Declaring type handle: EEType:0x00007FF7598B48D0
Target type handle: EEType:0x00007FF7598B5398
Method name: CreateParameterizedConstructor
Instantiation:
  Argument 00000000: EEType:0x00007FF7598B6E50
  Argument 00000001: EEType:0x00007FF7598984C8
  Argument 00000002: EEType:0x00007FF759898488
  Argument 00000003: EEType:0x00007FF759898488
  Argument 00000004: EEType:0x00007FF759898488
```

Seems that it was caused by `JsonSerializer.Deserialze`, but I don't know how to resolve it:
https://github.com/hez2010/SyaBot/blob/8fbaa4ccda964b5f88fe82865d0a4fb64c6037a7/SyaBot.Server/Program.cs#L60

MichalStrehovsky:
When the compiler sees a MakeGenericType/MakeGenericMethod call, it will try to come up with a shared instantiation so that things can at least work if instantiated with reference types at runtime.

We were allowing instantiations over `__Canon`, but this breaks places that use generic dictionaries to identify generic methods (we don't have one for `__Canon` instantiations). Instantiate over a concrete reference type instead.

Fixes #929.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/RootingHelpers.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/RootingHelpers.cs
@@ -113,7 +113,7 @@ namespace ILCompiler
             if (method.OwningType.IsGenericDefinition || method.OwningType.ContainsSignatureVariables(treatGenericParameterLikeSignatureVariable: true))
             {
                 TypeDesc owningType = method.OwningType.GetTypeDefinition();
-                Instantiation inst = TypeExtensions.GetInstantiationThatMeetsConstraints(owningType.Instantiation, allowCanon: method.Instantiation.Length == 0);
+                Instantiation inst = TypeExtensions.GetInstantiationThatMeetsConstraints(owningType.Instantiation, allowCanon: false);
                 if (inst.IsNull)
                 {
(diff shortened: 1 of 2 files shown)


jkotas (approved):
Thank you!

PR #959, opened 2021-04-13
Make System.Net.Http.Primitives part of testhost

aik-jahoda:
Fixes:
- mac build
- missing ref assembly
- missing dll in testhost


scalablecory:
This branch is the old one based on dotnet/runtime -- new one is in https://github.com/dotnet/runtimelab/tree/feature/LLHTTP2

PR #960, opened 2021-04-13
Merge from dotnet/runtime

jkotas:


jkotas:
@jkoritzinsky Could you please take a look or advise?

I have added the nativeaot component by following the pattern used by existing cmake components. `build.cmd -s nativeaot.libs` is failing with
```
 CMake Error at nativeaot/Bootstrap/base/cmake_install.cmake:41 (file):
     file INSTALL cannot find
     "D:/workspace/_work/1/s/artifacts/obj/coreclr/windows.x64.Checked/nativeaot/Bootstrap/base/bootstrapper.lib":
     File exists.
   Call Stack (most recent call first):
     nativeaot/Bootstrap/cmake_install.cmake:42 (include)
     nativeaot/cmake_install.cmake:42 (include)
     cmake_install.cmake:48 (include)
```
I am having hard time figuring out how to tell cmake that it needs to build bootstrapper.lib.

eng/native/functions.cmake
```diff
@@ -328,70 +352,86 @@ function(install_with_stripped_symbols targetName kind destination)
...
   else()
-    install(FILES ${symbol_file} DESTINATION ${destination_path})
+      install(FILES ${symbol_file} DESTINATION ${destination_path} ${ARGN})
   endif()
 endfunction()
 
-function(install_static_library targetName destination)
-  install (TARGETS ${targetName} DESTINATION ${destination})
+function(install_static_library targetName destination component)
+  install (TARGETS ${targetName} DESTINATION ${destination} COMPONENT ${component})
```
jkoritzinsky, line 375:
```suggestion
  if (NOT "${component}" STREQUAL "${targetName}")
    get_property(definedComponents GLOBAL PROPERTY CLR_CMAKE_COMPONENTS)
    list(FIND definedComponents "${component}" componentIdx)
    if (${componentIdx} EQUAL -1)
      message(FATAL_ERROR "The ${component} component is not defined. Add a call to `add_component(${component})` to define the component in the build.")
    endif()
    add_dependencies(${component} ${targetName})
  endif()
  install (TARGETS ${targetName} DESTINATION ${destination} COMPONENT ${component})
```

jkoritzinsky:
You'll need to add an additional target dependency. I've added a suggested change that should do the job well.

jkotas:
@jkoritzinsky Thank you for your help!

PR #965, opened 2021-04-15
Allow RuntimeType with missing EEType in some cases

MichalStrehovsky:
This is a different take on dotnet/runtimelab#620.

We want to make sure `MakeArrayType`/`MakeGenericType` fails if we don't have an EEType for the constructed type because that's the problematic API and it's good UX to have it fail there (as opposed to failing later because e.g. we can't get a `TypeHandle` in a `CreateInstance` call).

But it's really inconvenient to have this fail e.g. when we're hydrating types for the purpose of comparing method signatures. I've seen a failure yesterday where we couldn't find the `string..ctor(char[])` overload because we didn't have an EEType for `ReadOnlySpan<char>` (in a signature of an unrelated overload).

The AOT compiler could create those EETypes, but it's very wasteful. Instead, I'm relaxing the EEType check so that we only do it from the public APIs. It means people can get to "unusable" `RuntimeTypes` now if they're e.g. field types, or parameter types, but I believe those will mostly be corner cases.

I'm adding tests to make sure all of this is diagnosable.


jkotas (approved):
Thanks

PR #970, opened 2021-04-17
Generate metadata for what should be reflectable only

MichalStrehovsky:
This adds a mode (not default for now) where we only generate metadata for things that were specificed as reflectable. I'm seeing 15-20% size on disk savings.

We only do this for methods for now. Fields can follow, although I don't expect this big savings.

The way to specify reflectable things is through the usual avenues:
* Dataflow analysis
* Reflection rooting entire assemblies though our default compat settings
* RD.XML
* Maybe I forgot some

I tested this with `dotnet new webapi` and surpsingly this already works, with ~16% size savings.

It still needs a bit of work so that some invariants can hold (address taken methods should be reflectable so that GetMethodInfo works, all generic instantiations should match reflectability of the definition, etc.).

I'm hoping this will let us get rid of the concept of reflection blocking in the future.

On a high level, this change updates all the places that assumed "I got a method entrypoint therefore the method is reflectable unless blocked" with "you need a ReflectedMethod node to make things reflectable".

Using ReflectedMethodNode simplified reflection rooting in general.


jkotas (approved):
Nice!

PR #971, opened 2021-04-17
Merge from dotnet/runtime

jkotas:


davidfowl:
@jkotas cant we get the bot to manage these?

jkotas:
I doubt that it would save any time. Fixing the conflicts and build breaks is the time consuming part.

PR #972, opened 2021-04-18
Support for PreserveSig = false

kant2002:
Handle just very basic case where return type `void`


kant2002:
@jkotas can you take a look what I should add here as well?

jkotas (approved):
Thank you!

PR #978, opened 2021-04-19
Build ObjWriter with Ninja on Windows

MichalStrehovsky:
Per https://github.com/dotnet/runtime/issues/50794 Ninja is the default generator for building the runtime repo on Windows, so people should have it. Piping through the fallback for MSBuild doesn't seem worth the trouble. I might do it along with the `-vs` option for `build.cmd` next time I need to do an ObjWriter change (I use VS, so not having a vcxproj is going to be annoying).

```
build.cmd nativeaot.objwriter
```

Before: Time Elapsed 00:19:29.02
After: Time Elapsed 00:17:04.22

### DIFF
--- a/src/coreclr/tools/aot/ObjWriter/build.cmd
+++ b/src/coreclr/tools/aot/ObjWriter/build.cmd
@@ -95,10 +95,11 @@ set "ExtraCMakeArgs=%~3"
     -DLLVM_TARGETS_TO_BUILD="AArch64;ARM;X86" ^
     -DLLVM_EXTERNAL_OBJWRITER_SOURCE_DIR="%ScriptDir%\" ^
     -DCORECLR_INCLUDE_DIR="%RepoRoot%src\coreclr\inc" ^
+    -G Ninja ^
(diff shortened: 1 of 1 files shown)


jkotas (approved):
Thanks!

AntonLapounov (approved):

MichalStrehovsky:
ARM64 is failing because Ninja is binplacing into different directories than MSBuild did (the path doesn't include the target details in Ninja since Ninja can't do multiple configurations).

@jkoritzinsky do you happen to know the place in the infra scripts where we decide where to binplace for Ninja vs MSBuild? (I know I know we should ideally use gen-buildsys.cmd/sh and maybe it would happen automagically - I still couldn't find where that is.)

jkoritzinsky:
If you're using the `install` target, then it should binplace to the same location. If you're copying the files out of the tree, then it's a little more confusing.

MichalStrehovsky:
This is less of a priority now that the build time is 25% of what it used to be because of other optimizations.

PR #981, opened 2021-04-20
Use the same configuration env variables as CoreCLR

jkotas:
- Use DOTNET_ prefix
- Make all GC settings configurable


jkotas:
Follow up on #928

MichalStrehovsky (approved):
Thank you!

PR #982, opened 2021-04-20
Consistently mark reflectable generics

MichalStrehovsky:
In the `--reflectedonly` mode that only generates reflection metadata for things that were seen as reflected, ensure that generic things get consistent reflectability information (i.e. if we decide `Foo<int>.Blah()` needs to be available for reflection, but we also compiled `Foo<double>.Blah()`, make sure that we have the reflection mapping for `Foo<double>.Blah()` too).

This is because we often don't know the exact generic instantiation involved - the dataflow analysis operates on definitions only and one can just `object.GetType()` something that already exists in the system and reflect on its members without dataflow analysis objecting.


jkotas (approved):
Thank you!

PR #983, opened 2021-04-20
Revert "Skip test on s390x that runs into OOM killer"

akoeplinger:
Reverts dotnet/runtimelab#892, should work now that the s390x machines have more RAM.


akoeplinger:
FYI @uweigand

Issue #306, opened 2020-11-05
COM interop support


jkotas:
The Native AOT compiler does not support COM interop today. Attempt to use COM interop leads to errors like:
```
Unhandled Exception: System.PlatformNotSupportedException: COM Interop is not supported on this platform.
   at Internal.Runtime.CompilerHelpers.InteropHelpers.ConvertManagedComInterfaceToNative(Object) + 0x42
```

The right way to add support for COM interop to Native AOT runtime is combination of:
- [ ] Implement [ComWrappers](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.comwrappers) in the Native AOT runtime.
    - [x] ~~Add [CoreCLR ComWrappers implementation](https://github.com/dotnet/runtime/tree/master/src/coreclr/interop) to Native AOT build.~~
    - [x] Implement CCW - `CreateComInterfaceForObject` set of APIs
    - [x] Implement RCW - `CreateObjectForComInstance` set of APIs
    - [x] `ComWrappers.RegisterForMarshalling` - needed to replace built-in interop
    - [ ] `ComWrappers.RegisterForTrackerSupport` - WinRT-specific lifetime management
- [x] Implement [IDynamicInterfaceCastable](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.idynamicinterfacecastable) in the Native AOT runtime. CoreRT had a similar experimental feature before that was reverted in https://github.com/dotnet/corert/commit/5383d46c02e2375e181ea95407843c487106a55e . This commit provides approximate blue print for what the implementation should look like.
- [ ] Implemented support for shared generics with `IDynamicInterfaceCastable` https://github.com/dotnet/runtime/issues/72909
- [ ] External tool that generates the required COM marshaling stubs. There are several design options for how this tool can work; or there can even be multiple variants of this tool. The key difference between the different options is the source of truth that the tool consumes.
    - [ ] Source of truth are COM interface definitions in C#. This option is best for making the existing libraries that use COM interop work (e.g. WinForms or Avalonia). Source generator that is extension of [DllImportGenerator](https://github.com/dotnet/runtimelab/tree/feature/DllImportGenerator) may be a good implementation strategy for this option.
    - [ ] Source of truth are COM interface definitions in IDL or TLB file. This option is best for creating new libraries that use COM interop. The tool for this case would be conceptually similar to [cswinrt](https://github.com/microsoft/CsWinRT) tool for WinRT. 

[...]

PR #992, opened 2021-04-21
Add RCW support

kant2002:
See #306
- Implemented RCW creation for COM interfaces.
- Implemented returning original object, if COM interface is ManagedObjectWrapper
- Do not return original object if unique instance requested.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Threading;
 using Internal.Runtime.CompilerServices;
 
@@ -20,7 +21,7 @@ namespace System.Runtime.InteropServices
 
         private readonly ConditionalWeakTable<object, ManagedObjectWrapperHolder> _ccwTable = new ConditionalWeakTable<object, ManagedObjectWrapperHolder>();
         private readonly Lock _lock = new Lock();
-        private readonly Dictionary<IntPtr, WeakReference> _rcwCache = new Dictionary<IntPtr, WeakReference>();
+        private readonly Dictionary<IntPtr, GCHandle> _rcwCache = new Dictionary<IntPtr, GCHandle>();
         private readonly ConditionalWeakTable<object, NativeObjectWrapper> _rcwTable = new ConditionalWeakTable<object, NativeObjectWrapper>();
 
         /// <summary>
@@ -158,27 +159,17 @@ namespace System.Runtime.InteropServices
         internal unsafe class NativeObjectWrapper
         {
             private IntPtr _externalComObject;
+            private ComWrappers _owner;
 
-            public NativeObjectWrapper(IntPtr externalComObject)
+            public NativeObjectWrapper(IntPtr externalComObject, ComWrappers owner)
             {
                 _externalComObject = externalComObject;
             }
 
-            public uint AddRef()
-            {
-                return ((delegate* unmanaged<IntPtr, uint>)IUnknownVtbl[1])(_externalComObject);
-            }
-
-            public uint Release()
-            {
-                return ((delegate* unmanaged<IntPtr, uint>)IUnknownVtbl[2])(_externalComObject);
-            }
-
-            private IntPtr* IUnknownVtbl => (IntPtr*)((IntPtr*)_externalComObject)[0];
-
             ~NativeObjectWrapper()
             {
-                Release();
+                _owner.RemoveComInterfaceFromCache(_externalComObject);
+                Marshal.Release(_externalComObject);
             }
         }
 
@@ -340,6 +331,14 @@ namespace System.Runtime.InteropServices
             return (ComInterfaceDispatch*)comObject;
         }
 
+        private void RemoveComInterfaceFromCache(IntPtr comObject)
+        {
+            if (this._rcwCache.TryGetValue(comObject, out var handle))
+            {
+                handle.Free();
+            }
+        }
+
         /// <summary>
         /// Get the currently registered managed object or creates a new managed object and registers it.
         /// </summary>
@@ -381,11 +380,11 @@ namespace System.Runtime.InteropServices
             {
                 using (LockHolder.Hold(impl._lock))
                 {
-                    if (impl._rcwCache.TryGetValue(externalComObject, out var weakReference))
+                    if (impl._rcwCache.TryGetValue(externalComObject, out var handle))
                     {
-                        if (weakReference.IsAlive)
+                        if (handle.IsAllocated)
                         {
-                            retValue = weakReference.Target;
+                            retValue = handle.Target;
                             return false;
                         }
                     }
@@ -407,18 +406,18 @@ namespace System.Runtime.InteropServices
 
             using (LockHolder.Hold(impl._lock))
             {
-                if (impl._rcwCache.TryGetValue(externalComObject, out var existingWeakReference))
+                if (impl._rcwCache.TryGetValue(externalComObject, out var existingHandle))
                 {
-                    if (existingWeakReference.IsAlive)
+                    if (existingHandle.IsAllocated)
                     {
-                        retValue = existingWeakReference.Target;
+                        retValue = existingHandle.Target;
                     }
                 }
                 else
                 {
-                    NativeObjectWrapper wrapper = new NativeObjectWrapper(externalComObject);
+                    NativeObjectWrapper wrapper = new NativeObjectWrapper(externalComObject, impl);
                     impl._rcwTable.Add(retValue, wrapper);
-                    impl._rcwCache.Add(externalComObject, new WeakReference(retValue));
+                    impl._rcwCache.Add(externalComObject, GCHandle.Alloc(retValue, GCHandleType.Weak));
                 }
             }
 
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -421,7 +421,7 @@ namespace System.Runtime.InteropServices
         {
             using (LockHolder.Hold(_lock))
             {
-                _rcwCache.Remove(_externalComObject);
+                _rcwCache.Remove(comPointer);
             }
         }
 
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -160,17 +160,20 @@ namespace System.Runtime.InteropServices
         {
             private IntPtr _externalComObject;
             private ComWrappers _owner;
+            private GCHandle _handle;
 
-            public NativeObjectWrapper(IntPtr externalComObject, ComWrappers owner)
+            public NativeObjectWrapper(IntPtr externalComObject, ComWrappers owner, GCHandle handle)
             {
                 _externalComObject = externalComObject;
+                _handle = handle;
                 Marshal.AddRef(externalComObject);
             }
 
             ~NativeObjectWrapper()
             {
-                _owner.RemoveComInterfaceFromCache(_externalComObject);
+                _owner._rcwCache.Remove(_externalComObject);
                 Marshal.Release(_externalComObject);
+                _handle.Free();
             }
         }
 
@@ -332,14 +335,6 @@ namespace System.Runtime.InteropServices
             return (ComInterfaceDispatch*)comObject;
         }
 
-        private void RemoveComInterfaceFromCache(IntPtr comObject)
-        {
-            if (this._rcwCache.TryGetValue(comObject, out var handle))
-            {
-                handle.Free();
-            }
-        }
-
         /// <summary>
         /// Get the currently registered managed object or creates a new managed object and registers it.
         /// </summary>
@@ -416,9 +411,10 @@ namespace System.Runtime.InteropServices
                 }
                 else
                 {
-                    NativeObjectWrapper wrapper = new NativeObjectWrapper(externalComObject, impl);
+                    GCHandle handle = GCHandle.Alloc(retValue, GCHandleType.Weak);
+                    NativeObjectWrapper wrapper = new NativeObjectWrapper(externalComObject, impl, handle);
                     impl._rcwTable.Add(retValue, wrapper);
-                    impl._rcwCache.Add(externalComObject, GCHandle.Alloc(retValue, GCHandleType.Weak));
+                    impl._rcwCache.Add(externalComObject, handle);
                 }
             }
 
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
@@ -18,6 +19,8 @@ namespace System.Runtime.InteropServices
         internal static Guid IID_IUnknown = new Guid(0x00000000, 0x0000, 0x0000, 0xc0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46);
 
         private readonly ConditionalWeakTable<object, ManagedObjectWrapperHolder> _ccwTable = new ConditionalWeakTable<object, ManagedObjectWrapperHolder>();
+        private readonly Lock _lock = new Lock();
+        private readonly Dictionary<IntPtr, WeakReference> _rcwCache = new Dictionary<IntPtr, WeakReference>();
         private readonly ConditionalWeakTable<object, NativeObjectWrapper> _rcwTable = new ConditionalWeakTable<object, NativeObjectWrapper>();
 
         /// <summary>
@@ -364,13 +367,28 @@ namespace System.Runtime.InteropServices
             if (flags.HasFlag(CreateObjectFlags.Aggregation))
                 throw new NotImplementedException();
 
-            if (!flags.HasFlag(CreateObjectFlags.UniqueInstance))
+            if (flags.HasFlag(CreateObjectFlags.Unwrap))
             {
                 var comInterfaceDispatch = impl.TryGetComInterfaceDispatch(externalComObject);
                 if (comInterfaceDispatch != null)
                 {
                     retValue = ComInterfaceDispatch.GetInstance<object>(comInterfaceDispatch);
                     return true;
+                }
+            }
+
+            if (!flags.HasFlag(CreateObjectFlags.UniqueInstance))
+            {
+                using (LockHolder.Hold(impl._lock))
+                {
+                    if (impl._rcwCache.TryGetValue(externalComObject, out var weakReference))
+                    {
+                        if (weakReference.IsAlive)
+                        {
+                            retValue = weakReference.Target;
+                            return false;
+                        }
+                    }
                 }
             }
 
@@ -387,17 +405,24 @@ namespace System.Runtime.InteropServices
                 return true;
             }
 
-            NativeObjectWrapper wrapper;
-            if (impl._rcwTable.TryGetValue(retValue, out wrapper))
+            using (LockHolder.Hold(impl._lock))
             {
-                return true;
+                if (impl._rcwCache.TryGetValue(externalComObject, out var existingWeakReference))
+                {
+                    if (existingWeakReference.IsAlive)
+                    {
+                        retValue = existingWeakReference.Target;
+                    }
+                }
+                else
+                {
+                    NativeObjectWrapper wrapper = new NativeObjectWrapper(externalComObject);
+                    impl._rcwTable.Add(retValue, wrapper);
+                    impl._rcwCache.Add(externalComObject, new WeakReference(retValue));
+                }
             }
 
-            wrapper = impl._rcwTable.GetValue(retValue, (c) =>
-            {
-                return new NativeObjectWrapper(externalComObject);
-            });
-            wrapper.AddRef();
+            Marshal.AddRef(externalComObject);
             return true;
         }
 
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -173,7 +173,11 @@ namespace System.Runtime.InteropServices
             ~NativeObjectWrapper()
             {
                 _comWrappers.RemoveRCWFromCache(_externalComObject);
-                _proxyHandle.Free();
+                if (_proxyHandle.IsAllocated)
+                {
+                    _proxyHandle.Free();
+                }
+
                 Marshal.Release(_externalComObject);
             }
         }
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -152,13 +152,6 @@ namespace System.Runtime.InteropServices
             }
         }
 
-        internal unsafe struct IUnknownVftbl
-        {
-            public delegate* unmanaged<IntPtr, ref Guid, out IntPtr, int> QueryInterface;
-            public delegate* unmanaged<IntPtr, uint> AddRef;
-            public delegate* unmanaged<IntPtr, uint> Release;
-        }
-
         internal unsafe class NativeObjectWrapper
         {
             private IntPtr _externalComObject;
@@ -178,7 +171,7 @@ namespace System.Runtime.InteropServices
                 return ((delegate* unmanaged<IntPtr, uint>)IUnknownVtbl[2])(_externalComObject);
             }
 
-            private IntPtr* IUnknownVtbl => (IntPtr*)_externalComObject;
+            private IntPtr* IUnknownVtbl => (IntPtr*)((IntPtr*)_externalComObject)[0];
 
             ~NativeObjectWrapper()
             {
@@ -223,6 +216,12 @@ namespace System.Runtime.InteropServices
             ccwValue = _ccwTable.GetValue(instance, (c) =>
             {
                 ManagedObjectWrapper* value = CreateCCW(this, c, flags);
+                var dispatchesCount = value->UserDefinedCount;
+                if ((flags & CreateComInterfaceFlags.CallerDefinedIUnknown) == CreateComInterfaceFlags.None)
+                {
+                    dispatchesCount++;
+                }
+
                 return new ManagedObjectWrapperHolder(value);
             });
             return ccwValue.ComIp;
@@ -332,6 +331,18 @@ namespace System.Runtime.InteropServices
             return obj!;
         }
 
+        private unsafe ComInterfaceDispatch* TryGetComInterfaceDispatch(IntPtr comObject)
+        {
+            // If the first Vtable entry is part of the ManagedObjectWrapper IUnknown impl,
+            // we know how to interpret the IUnknown.
+            if (((IntPtr*)((IntPtr*)comObject)[0])[0] != ((IntPtr*)DefaultIUnknownVftblPtr)[0])
+            {
+                return null;
+            }
+
+            return (ComInterfaceDispatch*)comObject;
+        }
+
         /// <summary>
         /// Get the currently registered managed object or creates a new managed object and registers it.
         /// </summary>
@@ -345,7 +356,7 @@ namespace System.Runtime.InteropServices
         /// <remarks>
         /// If <paramref name="impl" /> is <c>null</c>, the global instance (if registered) will be used.
         /// </remarks>
-        private static bool TryGetOrCreateObjectForComInstanceInternal(
+        private static unsafe bool TryGetOrCreateObjectForComInstanceInternal(
             ComWrappers impl,
             IntPtr externalComObject,
             IntPtr innerMaybe,
@@ -359,7 +370,13 @@ namespace System.Runtime.InteropServices
             if (flags.HasFlag(CreateObjectFlags.Aggregation))
                 throw new NotImplementedException();
 
-            object? wrapperMaybeLocal = wrapperMaybe;
+            var comInterfaceDispatch = impl.TryGetComInterfaceDispatch(externalComObject);
+            if (comInterfaceDispatch != null)
+            {
+                retValue = ComInterfaceDispatch.GetInstance<object>(comInterfaceDispatch);
+                return true;
+            }
+
             retValue = impl.CreateObject(externalComObject, flags);
             if (retValue == null)
             {
@@ -481,7 +498,7 @@ namespace System.Runtime.InteropServices
                 throw new InvalidOperationException(SR.InvalidOperation_ComInteropRequireComWrapperInstance);
             }
 
-            return s_globalInstanceForMarshalling.CreateObject(externalComObject, CreateObjectFlags.None);
+            return s_globalInstanceForMarshalling.GetOrCreateObjectForComInstance(externalComObject, CreateObjectFlags.None);
         }
 
         [UnmanagedCallersOnly]
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -159,26 +159,21 @@ namespace System.Runtime.InteropServices
         internal unsafe class NativeObjectWrapper
         {
             private IntPtr _externalComObject;
-            private GCHandle _comWrappersHandle;
+            private ComWrappers _comWrappers;
             private GCHandle _proxyHandle;
 
-            public NativeObjectWrapper(IntPtr externalComObject, GCHandle comWrappersHandle, GCHandle proxyHandle)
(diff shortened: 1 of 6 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -332,8 +360,31 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
...
+            if (impl._rcwTable.TryGetValue(retValue, out wrapper))
+            {
+                return true;
+            }
+
+            wrapper = impl._rcwTable.GetValue(retValue, (c) =>
+            {
+                return new NativeObjectWrapper(externalComObject);
+            });
+            wrapper.AddRef();
```
jkotas: Would it make more sense to have this AddRef in constructor since the matching Release is in finalizer?

jkotas:
Add some basic tests?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -332,8 +360,31 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
                 throw new NotImplementedException();
 
             object? wrapperMaybeLocal = wrapperMaybe;
-            retValue = null;
-            throw new NotImplementedException();
+            retValue = impl.CreateObject(externalComObject, flags);
```
jkotas: Should this do a lookup in the list of active objects first before creating the object?

kant2002: I believe I raise this question regarding RCW creation. If caching is responsibility of runtime, then yes, I have to lookup in the list of active objects first.

kant2002: Does that mean that in addition to `ConditionalWeakTable<object, NativeObjectWrapper> _rcwTable` I need separate `ConcurrentDictionary<IntPtr, WeakReference<object>> _rcwCache` ?

jkotas: Yes, something like that. It needs to match this from CoreCLR: https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/coreclr/vm/interoplibinterface.cpp#L708

src/tests/nativeaot/SmokeTests/ComWrappers/ComWrappers.cs
```diff
@@ -83,14 +91,42 @@ public static void TestComInteropReleaseProcess()
             ThrowIfNotEquals(false, comPointerHolder.IsAlive, ".NET object should be disposed by then");
         }
 
+        public static void TestComInteropCCWCreation()
+        {
+            Console.WriteLine("Testing CCW release process");
+            // Lines below was called in previous tests.
```
jkotas: Would it be better to call this in the Main method?

kant2002:
@jkotas I am definitely ready for you to review this.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -331,9 +364,41 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
             if (flags.HasFlag(CreateObjectFlags.Aggregation))
                 throw new NotImplementedException();
 
-            object? wrapperMaybeLocal = wrapperMaybe;
-            retValue = null;
-            throw new NotImplementedException();
+            if (!flags.HasFlag(CreateObjectFlags.UniqueInstance))
+            {
+                var comInterfaceDispatch = impl.TryGetComInterfaceDispatch(externalComObject);
```
jkotas: This should be done only for `CreateObjectFlags.Unwrap`.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
jkotas: Still missing?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -331,9 +364,41 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
...
+            }
+
+            if (flags.HasFlag(CreateObjectFlags.UniqueInstance))
+            {
+                // No need to cache NativeObjectWrapper for unique instances. They are not cached.
+                return true;
+            }
+
+            NativeObjectWrapper wrapper;
+            if (impl._rcwTable.TryGetValue(retValue, out wrapper))
```
jkotas: `impl.CreateObject` above is meant to always create new object. It means that this lookup will never succeed - if it is new object, it could not have been seen before.

kant2002: Again, returning back to responsibility of runtime and implementor of ComWrappers. In the docs https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.comwrappers.createobject?view=net-5.0 `Creates a managed object for the object that externalComObject points to, respecting the values of flags.`. What we do is `not respecting`, i.e. just return always new instances. Maybe wording should be changed somehow? or expanded. I always trying to fall into trap that caching is responsibility of implementor (even if I know that's not true.

kant2002: Also from here https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.comwrappers?view=net-5.0 : `Unlike the built-in features, a majority of the responsibility for accurate lifetime management, dispatching methods, and marshalling of arguments and return values is left to the ComWrappers implementer.` Note  **lifetime management** which for me indicate that all caching/uniquess/addRef should be done by implementer. `The +1 reference count on the native object is assumed to be performed by the managed proxy creator (that is, the ComWrappers implementer) to ensure the associated lifetime between the native object and its managed proxy` from same article, just almost at the end of document.

jkotas: > Unlike the built-in features, a majority of the responsibility for accurate lifetime management, dispatching methods, and marshalling of arguments and return values is left to the ComWrappers implementer.

The next paragraph after that says: 
```
Efficient mapping between a native IUnknown and its managed proxy
```
This mapping is missing in this PR.

> Note lifetime management

It says "a majority of the responsibility for accurate lifetime management". The minority that it is reponsible for is the mapping.

Feel free to submit a PR to the doc repo.

jkotas:
> Add CCW support 

This shold say RCW. I had it swapped in the tracking issue - I have fixed it there.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: This is very delicate topic, and I see myself doing too much errors for now, so I probing my understanding in discussion with you.  I still seek for hidden assumptions, that's why I'm asking same question in different shapes.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Question. Should roundrtip from Marshal API happens automatically? For me this is means that I should create global `ComWrappers` with `CreateObjectFlags.Unwrap` instead of `CreateObjectFlags.None`.
```
var target = new ComObject();
int result = CaptureComPointer(target);
ThrowIfNotEquals(0, result, "Seems to be COM marshalling behave strange.");
result = RetreiveCapturedComPointer(out var capturedObject);
ThrowIfNotEquals(0, result, "Seems to be COM marshalling behave strange.");
if (capturedObject != target)
{
    throw new Exception("Should round-trip");
}
```

kant2002:
> I had it swapped in the tracking issue - I have fixed it there.

Thanks. It was a big linguistic battle inside me. Since I have a lot of other question, in order to not too much distract you, I decide that I do not understand some language tricks.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
jkotas: Probably. I do not remember the details like this. Test what CoreCLR does and match it.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -18,6 +19,9 @@ public abstract partial class ComWrappers
         internal static Guid IID_IUnknown = new Guid(0x00000000, 0x0000, 0x0000, 0xc0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46);
 
         private readonly ConditionalWeakTable<object, ManagedObjectWrapperHolder> _ccwTable = new ConditionalWeakTable<object, ManagedObjectWrapperHolder>();
+        private readonly Lock _lock = new Lock();
+        private readonly Dictionary<IntPtr, WeakReference> _rcwCache = new Dictionary<IntPtr, WeakReference>();
```
jkotas: Do you also need code to delete entries from this cache?

Also, the value can be GCHandle. It will save you an extra allocation.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Will do. It was unexpected twist that you do not know details. From outside appears that you do know details.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: I think this is covered already.

kant2002: I think this is already implemented.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Probably I should not remove entries from the cache, if I want to maintain compatibility with .NET Core. I think idea for that is to maintain identity between Com Object and .NET Object.

jkotas: I am pretty sure .NET Core removes entries from the cache once things become unreachable.

Otherwise, this cache would be a guaranteed memory leak.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -331,9 +366,63 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
...
+                    if (existingHandle.IsAllocated)
+                    {
+                        retValue = existingHandle.Target;
+                    }
+                }
+                else
+                {
+                    NativeObjectWrapper wrapper = new NativeObjectWrapper(externalComObject, impl);
+                    impl._rcwTable.Add(retValue, wrapper);
+                    impl._rcwCache.Add(externalComObject, GCHandle.Alloc(retValue, GCHandleType.Weak));
```
jkotas: With GCHandle, you will also need to cleanup all GCHandles if the ComWrappers instance is garbage collected. 

It may make sense for the GCHandle to be owned by `NativeObjectWrapper` and delete it in `NativeObjectWrapper` finalizer.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Am I read this incorrectly? https://github.com/dotnet/runtime/blob/main/src/tests/Interop/COM/ComWrappers/API/Program.cs#L118

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: This works as I expect. Add that to smoke tests.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
jkotas: This is for the other direction (CCWs). It is not really always present. It is only present for as long as the object instance is alive.

kant2002: Okay. Make sense.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Then it require me to `Dictionary<IntPtr, NativeObjectWrapper> _rcwCache` instead of `Dictionary<IntPtr, GCHandle> _rcwCache`. Would be finalizer for `NativeObjectWrapper` even called in that way? I use GCHandle now to break cycle
`ComProxy` -> `NativeObjectWrapper` -> `ComWrappers` -> `_rcwCache<IntPtr, GCHandle<ComProxy>>`.
If link between `ComProxy` -> `NativeObjectWrapper` is weak, and `NativeObjectWrapper` will be finalized even other parts of code have reference on it, I can do what you propose.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -18,6 +20,17 @@ public abstract partial class ComWrappers
         internal static Guid IID_IUnknown = new Guid(0x00000000, 0x0000, 0x0000, 0xc0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46);
 
         private readonly ConditionalWeakTable<object, ManagedObjectWrapperHolder> _ccwTable = new ConditionalWeakTable<object, ManagedObjectWrapperHolder>();
+        private readonly Lock _lock = new Lock();
+        private readonly Dictionary<IntPtr, GCHandle> _rcwCache = new Dictionary<IntPtr, GCHandle>();
+        private readonly ConditionalWeakTable<object, NativeObjectWrapper> _rcwTable = new ConditionalWeakTable<object, NativeObjectWrapper>();
+
+        ~ComWrappers()
```
jkotas: It is an API breaking change to add finalizer to a non-sealed type that does not have it. The build should fail with public API validation error.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
jkotas: I think the dictionary can still be just `Dictionary<IntPtr, GCHandle>`, even when you do this.

kant2002:
@jkotas I think I'm done with this batch

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -331,9 +368,67 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
...
+                {
+                    if (existingHandle.IsAllocated)
+                    {
+                        retValue = existingHandle.Target;
+                    }
+                }
+                else
+                {
+                    GCHandle proxyHandle = GCHandle.Alloc(retValue, GCHandleType.Weak);
+                    GCHandle comWrappersHandle = GCHandle.Alloc(impl, GCHandleType.Weak);
```
jkotas: Why do you need the ComWrappers reference to be weak?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -151,11 +156,31 @@ public ManagedObjectWrapperHolder(ManagedObjectWrapper* wrapper)
...
+
+            ~NativeObjectWrapper()
+            {
+                if (_comWrappersHandle.IsAllocated)
+                {
+                    ((ComWrappers)_comWrappersHandle.Target)._rcwCache.Remove(_externalComObject);
+                }
+
+                Marshal.Release(_externalComObject);
+                __comWrappersHandle.Free;
```
jkotas: Looks like a typo (double underscore).

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Yes. If ComWrappers garbange collected, my understanding that `NativeObjectWrapper` instances in the `ConditionalWeakTable` would not be immediately collected and just scheduled to run finalizers (?) which may happens a bit after or maybe in different order. So I should not remove anything from `_rcwCache` in that case becasue object already dead.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: thanks. And miss `()`

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
jkotas: I do not see the problem. If you change `NativeObjectWrapper` to have a regular reference to the owning ComWrappers instance,, what is not going to work? Is there a test that is going to fail?

kant2002: As always you are right. I was over-complicating stuff. Remove handle on `ComWrappers`

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -151,11 +156,26 @@ public ManagedObjectWrapperHolder(ManagedObjectWrapper* wrapper)
...
+            {
+                _externalComObject = externalComObject;
+                _comWrappers = comWrappers;
+                _proxyHandle = proxyHandle;
+                Marshal.AddRef(externalComObject);
+            }
+
+            ~NativeObjectWrapper()
+            {
+                _comWrappers._rcwCache.Remove(_externalComObject);
```
jkotas: Does this need to take the lock?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -317,7 +349,7 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
         /// <remarks>
         /// If <paramref name="impl" /> is <c>null</c>, the global instance (if registered) will be used.
         /// </remarks>
-        private static bool TryGetOrCreateObjectForComInstanceInternal(
+        private static unsafe bool TryGetOrCreateObjectForComInstanceInternal(
             ComWrappers impl,
```
jkotas: Can this be an instance method instead? I do not see the point of passing in impl explicitly. (CreateCCW has the same issue.)

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -331,9 +363,66 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
...
+            {
+                // No need to cache NativeObjectWrapper for unique instances. They are not cached.
+                return true;
+            }
+
+            using (LockHolder.Hold(impl._lock))
+            {
+                if (impl._rcwCache.TryGetValue(externalComObject, out var existingHandle))
+                {
+                    if (existingHandle.IsAllocated)
```
jkotas: IsAllocated should be always true here. We should not ever get unallocated GCHandle from the Dictionary.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -331,9 +363,66 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
...
+                    {
+                        retValue = existingHandle.Target;
+                    }
+                }
+                else
+                {
+                    GCHandle proxyHandle = GCHandle.Alloc(retValue, GCHandleType.Weak);
```
jkotas: For symetry, it would be nice to allocated the GCHandle and add it to the table in NativeObjectWrapper constructor - since the opposite is done in the NativeObjectWrapper finalizer.

kant2002:
@jkotas I think I address your feedback.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -151,11 +156,26 @@ public ManagedObjectWrapperHolder(ManagedObjectWrapper* wrapper)
...
 
-        internal unsafe struct IUnknownVftbl
+        internal unsafe class NativeObjectWrapper
         {
-            public delegate* unmanaged<IntPtr, ref Guid, out IntPtr, int> QueryInterface;
-            public delegate* unmanaged<IntPtr, uint> AddRef;
-            public delegate* unmanaged<IntPtr, uint> Release;
+            private IntPtr _externalComObject;
+            private ComWrappers _comWrappers;
+            public GCHandle ProxyHandle;
```
jkotas: I would keep it as `_proxyHandle`. (Some of the places in the file are actually still refering to the field as `_proxyHandle`.)

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -151,11 +156,26 @@ public ManagedObjectWrapperHolder(ManagedObjectWrapper* wrapper)
...
         }
 
-        internal unsafe struct IUnknownVftbl
+        internal unsafe class NativeObjectWrapper
         {
-            public delegate* unmanaged<IntPtr, ref Guid, out IntPtr, int> QueryInterface;
-            public delegate* unmanaged<IntPtr, uint> AddRef;
-            public delegate* unmanaged<IntPtr, uint> Release;
+            private IntPtr _externalComObject;
+            private ComWrappers _comWrappers;
```
jkotas:
```suggestion
            private readonly IntPtr _externalComObject;
            private readonly ComWrappers _comWrappers;
```
Nit

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -151,11 +156,26 @@ public ManagedObjectWrapperHolder(ManagedObjectWrapper* wrapper)
...
+                _externalComObject = externalComObject;
+                _comWrappers = comWrappers;
+                ProxyHandle = GCHandle.Alloc(comProxy, GCHandleType.Weak);
+                Marshal.AddRef(externalComObject);
+            }
+
+            ~NativeObjectWrapper()
+            {
+                _comWrappers.RemoveRCWFromCache(_externalComObject);
+                _proxyHandle.Free();
```
jkotas: This can be unallocated handle if GCHandle.Alloc in the constructor fails. This should get the IsAllocated check.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Done.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Done

kant2002:
@jkotas I think I address your feedback again

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -151,11 +156,30 @@ public ManagedObjectWrapperHolder(ManagedObjectWrapper* wrapper)
...
 
-        internal unsafe struct IUnknownVftbl
+        internal unsafe class NativeObjectWrapper
         {
-            public delegate* unmanaged<IntPtr, ref Guid, out IntPtr, int> QueryInterface;
-            public delegate* unmanaged<IntPtr, uint> AddRef;
-            public delegate* unmanaged<IntPtr, uint> Release;
+            private readonly IntPtr _externalComObject;
+            private readonly ComWrappers _comWrappers;
+            public GCHandle _proxyHandle;
```
jkotas:
```suggestion
            internal GCHandle _proxyHandle;
```
does not need to be public

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -151,11 +156,30 @@ public ManagedObjectWrapperHolder(ManagedObjectWrapper* wrapper)
...
+
+            ~NativeObjectWrapper()
+            {
+                _comWrappers.RemoveRCWFromCache(_externalComObject);
+                if (_proxyHandle.IsAllocated)
+                {
+                    _proxyHandle.Free();
+                }
+
+                Marshal.Release(_externalComObject);
```
jkotas, line 181:
```suggestion
                if (_proxyHandle.IsAllocated)
                {
                    _proxyHandle.Free();
                    Marshal.Release(_externalComObject);
                }
```
When the handle allocation fails, we are going to miss the AddRef as well.

Alternative (and maybe easier to read?) way to fix this would be to reorder the operations in the constructor.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: Curious, what's your rule of thumb on visibility. I usually control visibility mostly using classes and leave all surface of internal classes as public to not bother, until `publicity` happens. Can you share your perspective?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: I'm actually for alternative and reorder operations in the constructor.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
jkotas: I believe that the usual style in runtime libraries is to try to make the visilibity right from the start even for internal types, especially for fields. The hypothetical question to ask - if we were to make this type public for some reason, is the visibility going to be right?

If you search through libraries sources, I think you will find a lot more internal fields in internal types than public fields in internal types.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -331,9 +362,67 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
...
+                    retValue = ComInterfaceDispatch.GetInstance<object>(comInterfaceDispatch);
+                    return true;
+                }
+            }
+
+            if (!flags.HasFlag(CreateObjectFlags.UniqueInstance))
+            {
+                using (LockHolder.Hold(_lock))
+                {
+                    if (_rcwCache.TryGetValue(externalComObject, out var handle))
```
jkotas:
```suggestion
                    if (_rcwCache.TryGetValue(externalComObject, out GCHandle handle))
```
Nit - coding conventions

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -913,17 +913,29 @@ protected override void AllocAndTransformNativeToManaged(ILCodeStream codeStream
...
+            LoadNativeValue(codeStream);
+            codeStream.Emit(ILOpcode.brfalse, lNull);
+
+            LoadNativeValue(codeStream);
+            MethodDesc helper = Context.GetHelperEntryPoint("InteropHelpers", "ConvertNativeComInterfaceToManaged");
+            codeStream.Emit(ILOpcode.call, emitter.NewToken(helper));
+
+            StoreManagedValue(codeStream);
+
+            codeStream.EmitLabel(lNull);
```
jkotas: Is it correct to skip storing the managed value for null?

jkotas:
LGTM otherwise. Thank you!

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
kant2002: I think yes. At least this pattern is everywhere.

kant2002: Managed value initialized as `null` by runtime guarantee, so load/store usually going in pairs. Then somebody do actual return. At least this is how I understand this.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -913,17 +913,29 @@ protected override void AllocAndTransformNativeToManaged(ILCodeStream codeStream
         {
```
jkotas: Can this method be deleted? I think implementing `TransformNativeToManaged` will automatically fill in this one as well.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
jkotas: > Managed value initialized as null by runtime guarantee,

Where is it guaranteed?

I went through all TransformNativeToManaged methods and all of them end with Store. I have not found one that skips the store.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
kant2002: Probably yes. this method do not called yet.

kant2002: Removed it.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
kant2002: LayoutClassPtrMarshaller, AsAnyMarshaller for example

jkotas: The managed value for these marshallers is a pointer to the structure that they are marshalling. It is not the case here.

kant2002: What I was trying to say that
```
.method public hidebysig specialname 
        instance int32  get_Set() cil managed
{
  .maxstack  1
  .locals init (int32 V_0)
  IL_0009:  ldloc.0
  IL_000a:  ret
} // end of method SetProperty::get_Set

```

is equivalent for 
```
int Set => 0;
```

kant2002: I miss your previous message. Will have to check what kind of IL actually generated to validate that what I'm saying is correct.

jkotas: You may be right, but I would still simplify this to a simple ConvertNativeComInterfaceToManaged. If the store is redudandant, the JIT will eliminate it after inlining ConvertNativeComInterfaceToManaged.

jkotas (approved):

PR #999, opened 2021-04-22
Synchronize logging with IL Linker

MichalStrehovsky:
Adds support for the `singlewarn` option that collapses all trimming (and for us, also AOT) warnings in an assembly into a single log entry.

Also includes MSBuild goo that turns on detailed warnings only for user assemblies. Most of that is copied from IL Link targets and can be deleted once we drop .NET 5.0 SDK support.


jkotas (approved):
Thank you

Issue #1011, opened 2021-04-24
[NativeAOT] singlemethodname do not work in non system module


kant2002:
I was trying to use that option to limit generation only to method which use marshallers, but unfortunately I cannot do that. `singlemethodname` used only system module for it's work. Is it possible to have it work for at least entry point assembly?

jkotas:
Try: `--singlemethodtypename "MyType,MyAssembly"`

jkotas:
It uses the custom attribute type name syntax that is defined in ECMA-335 (look for `SerString`).

kant2002:
Will try tomorrow, when I will be around Windows, but I doubt it will works:

https://github.com/dotnet/runtimelab/blob/9f3f548d4fa69215f871d383c1ad5ce228e3003e/src/coreclr/tools/aot/ILCompiler/Program.cs#L864

where 

https://github.com/dotnet/runtimelab/blob/9f3f548d4fa69215f871d383c1ad5ce228e3003e/src/coreclr/tools/aot/ILCompiler/Program.cs#L841-L854

where explicitly SystemModule is used.

And SingleMethod resolved at stage of parsing arguments before even assemblies list collected.

jkotas:
I am pretty sure that it works. I have used it many times. To double check, I have just tried `--singlemethodtypename "HelloWorld.Program,HelloWorld" --singlemethodname Main` with the samples/HelloWorld, and it worked as expected.

> where explicitly SystemModule is used.

SystemModule is the default if the name does not specify any assembly.

> SingleMethod resolved at stage of parsing arguments before even assemblies list collected.

That's not correct.

kant2002:
Ok. I will check that. Thanks.

kant2002:
Indeed, when I specify assembly it works. Thanks.

PR #1016, opened 2021-04-26
Add explanation text for --singlemethodtypename parameter

kant2002:
Should help clarify confusion like in #1011 where incorrect usage was assumed

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler/Program.cs
+++ b/src/coreclr/tools/aot/ILCompiler/Program.cs
@@ -217,7 +217,7 @@ namespace ILCompiler
                 syntax.DefineOption("targetarch", ref _targetArchitectureStr, "Target architecture for cross compilation");
                 syntax.DefineOption("targetos", ref _targetOSStr, "Target OS for cross compilation");
 
-                syntax.DefineOption("singlemethodtypename", ref _singleMethodTypeName, "Single method compilation: full name of the owning type including assembly name. (Format: NameSpace.TypeName,AssemblyName)");
+                syntax.DefineOption("singlemethodtypename", ref _singleMethodTypeName, "Single method compilation: name of the owning type including assembly name. (format: NameSpace.TypeName,AssemblyName)");
                 syntax.DefineOption("singlemethodname", ref _singleMethodName, "Single method compilation: name of the method");
                 syntax.DefineOptionList("singlemethodgenericarg", ref _singleMethodGenericArgs, "Single method compilation: generic arguments to the method");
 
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler/Program.cs
```diff
@@ -217,7 +217,7 @@ private ArgumentSyntax ParseCommandLine(string[] args)
                 syntax.DefineOption("targetarch", ref _targetArchitectureStr, "Target architecture for cross compilation");
                 syntax.DefineOption("targetos", ref _targetOSStr, "Target OS for cross compilation");
 
-                syntax.DefineOption("singlemethodtypename", ref _singleMethodTypeName, "Single method compilation: name of the owning type");
+                syntax.DefineOption("singlemethodtypename", ref _singleMethodTypeName, "Single method compilation: full name of the owning type including assembly name. (Format: NameSpace.TypeName,AssemblyName)");
```
jkotas: "full name" (e.g. `Type.FullName`) usually means type name without assembly name.

jkotas: The next thing somebody will ask is how to specify generics, nested types, or the parametrized types. Would it be better to just say that it is assembly qualified name or that it is same format as `Type.GetType` ?

kant2002: Something like that `Single method compilation: assembly qualified name of the owning type.`, also probably I have to find place in the documentation where I can put examples of usage. Does this qualifies for couple lines in docs?

jkotas: Yes, sounds reasonable.

Nit: It should be `assembly-qualified name`. The official documentation like https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype?view=net-5.0#System_Type_GetType_System_String_ uses spells it with a dash.

For documentation, I think it would be best to piggy back on https://github.com/dotnet/runtime/blob/main/docs/workflow/debugging/coreclr/debugging-crossgen2.md debugging instructions and improve them as necessary. Once/if we more NativeAOT out of the experimental state, I expect that we will merge crossgen2 and ilc into one compiler that supports multiple compilation modes.

kant2002: Given that this documentation is already in NativeAOT branch, I do not see easy way how better educate developer on the need to read that specific part.

src/coreclr/tools/aot/ILCompiler/Program.cs
```diff
@@ -217,7 +217,7 @@ private ArgumentSyntax ParseCommandLine(string[] args)
                 syntax.DefineOption("targetarch", ref _targetArchitectureStr, "Target architecture for cross compilation");
                 syntax.DefineOption("targetos", ref _targetOSStr, "Target OS for cross compilation");
 
-                syntax.DefineOption("singlemethodtypename", ref _singleMethodTypeName, "Single method compilation: name of the owning type");
+                syntax.DefineOption("singlemethodtypename", ref _singleMethodTypeName, "Single method compilation: assembly-qualified name of the owning type.");
```
jkotas:
```suggestion
                syntax.DefineOption("singlemethodtypename", ref _singleMethodTypeName, "Single method compilation: assembly-qualified name of the owning type");
```

jkotas (approved):

PR #1021, opened 2021-04-26
Require VS 2019 for linking the native image

jkotas:
The runtime libraries are built using VS 2019

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/findvcvarsall.bat
+++ b/src/coreclr/nativeaot/BuildIntegration/findvcvarsall.bat
@@ -15,7 +15,7 @@ IF /I "%~1"=="arm64" SET toolsSuffix=ARM64
 FOR /F "tokens=*" %%i IN (
     '"%vswherePath%" -latest -prerelease -products * ^
     -requires Microsoft.VisualStudio.Component.VC.Tools.%toolsSuffix% ^
-    -version [16,16] ^
+    -version [16,17) ^
     -property installationPath'
 ) DO SET vsBase=%%i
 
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/BuildIntegration/findvcvarsall.bat
```diff
@@ -15,6 +15,7 @@ IF /I "%~1"=="arm64" SET toolsSuffix=ARM64
 FOR /F "tokens=*" %%i IN (
     '"%vswherePath%" -latest -prerelease -products * ^
     -requires Microsoft.VisualStudio.Component.VC.Tools.%toolsSuffix% ^
+    -version [16,16] ^
```
AntonLapounov:
```suggestion
    -version [16,17) ^
```
Or `[16,)`. The current version is 16.9.*.

AntonLapounov (approved):

src/coreclr/nativeaot/BuildIntegration/findvcvarsall.bat
jkotas: Let's see whether this is going to work with cmd. I am seeing cmd behaving differently in CI vs. locally on my machine.

src/coreclr/nativeaot/BuildIntegration/findvcvarsall.bat
```diff
@@ -15,6 +15,7 @@ IF /I "%~1"=="arm64" SET toolsSuffix=ARM64
 FOR /F "tokens=*" %%i IN (
     '"%vswherePath%" -latest -prerelease -products * ^
     -requires Microsoft.VisualStudio.Component.VC.Tools.%toolsSuffix% ^
+    -version [16,17) ^
```
AntonLapounov:
```suggestion
    -version [16^,17^) ^
```

src/coreclr/nativeaot/BuildIntegration/findvcvarsall.bat
AntonLapounov: I see why.  We need to escape both `,` and `)`.

PR #1025, opened 2021-04-27
Get rid of a dataflow warning

MichalStrehovsky:
I postponed this while the annotation on Delegate.CreateDelegate was pending in the runtime repo. It has been annotated for a while.


jkotas (approved):
Thanks!

PR #1026, opened 2021-04-27
Update files shared with illink

MichalStrehovsky:
This brings two major things:

* Ability to track contents of arrays in the dataflow analysis. We could later use this to model MakeGenericType and MakeGenericMethod.
* Support for DynamicallyAccessedMembersAttribute on classes and interfaces.

The semantic of DAMAttribute on classes and interface is: when a type has DAMAttribute, calls to `object.GetType` on locations statically typed as something having/inheriting the DAMAttribute return an annotated System.Type value.

The marking part of this required creating a new node in the dependency graph. The node represents "`object.GetType` could have been called on this annotated type". It brings the dependencies mandated by the DAMAttribute.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/MethodBodyScanner.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/MethodBodyScanner.cs
@@ -187,12 +187,6 @@ namespace ILCompiler.Dataflow
             }
         }
 
-        public struct ValueBasicBlockPair
-        {
-            public ValueNode Value;
-            public int BasicBlockIndex;
-        }
-
         private static void StoreMethodLocalValue(
             ValueBasicBlockPair[] valueCollection,
             ValueNode valueToStore,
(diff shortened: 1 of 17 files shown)


src/tests/nativeaot/SmokeTests/Dataflow/Dataflow.cs
```diff
@@ -445,4 +562,19 @@ public static int CountFields(this Type t)
         Justification = "That's the point")]
     public static int CountProperties(this Type t)
         => t.GetProperties(BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly).Length;
+
+    [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2070:UnrecognizedReflectionPattern",
+        Justification = "That's the point")]
```
davidfowl, line 567: 🤣

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/MethodBodyScanner.cs
```diff
@@ -942,5 +988,87 @@ protected virtual void HandleStoreParameter(MethodIL method, int offset, int ind
             int offset,
             ValueNodeList methodParams,
             out ValueNode methodReturnValue);
+
+        // Limit tracking array values to 32 values for performance reasons. There are many arrays much longer than 32 elements in .NET, but the interesting ones for the linker are nearly always less than 32 elements.
+        private const int MaxTrackedArrayValues = 32;
```
jkotas, line 993: This reminds me of .NET Native data flow analyzer that was full of arbitrary cut-offs like this...

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/MethodBodyScanner.cs
MichalStrehovsky: I asked for an arbitrary limit because the other option is that we would model the values of arbitrarily big arrays and risk running out of memory while doing so.

https://github.com/mono/linker/pull/1955#discussion_r611645628

Maybe it would have been better not to model arrays at all...

jkotas: > Maybe it would have been better not to model arrays at all...

Yep. We are getting on a slippery slope with this.

MichalStrehovsky: > > Maybe it would have been better not to model arrays at all...
> 
> Yep. We are getting on a slippery slope with this.

Cc @vitek-karas

vitek-karas: Alternative would be to only model arrays we care about. Which is currently almost exclusively `Type[]`. We could increase the limit (a lot) because nobody will create large arrays of Type statically. My understanding is that the worry with the existing behavior is that it catches on things like `char[]` which can be very large but still fully static.

jkotas: > nobody will create large arrays of Type statically.

Famous last words

vitek-karas: OK OK - I knew this is going to come back to bite me... 😄 
But realistically - we only need to track arrays which are `Type[]` and their length is as long as the largest number of generic parameters on a single type/method. I don't know if reflection/IL has some limits on that, probably not really, so we might as well not put a limit on it either in the linker. If somebody does that ... and linker runs out of memory somehow... I mean there are probably easier ways to force linker to consume lot of memory.

PR #1033, opened 2021-04-28
Ninja-related build fixes

AntonLapounov:
Remove `clr.jit`, which is a proper subset of `clr.alljits`, from the `DefaultNativeAotSubsets` list.  That should fix the race condition in official builds.

Fix the location of the `AsmOffsets.cs` file for non-Ninja builds on Windows.  The condition and the path reflect the following script lines: https://github.com/dotnet/runtimelab/blob/3ad5624dd0affe31a661612cf4e8b7976bf38ad8/src/coreclr/runtime.proj#L30 https://github.com/dotnet/runtimelab/blob/3ad5624dd0affe31a661612cf4e8b7976bf38ad8/src/coreclr/build-runtime.cmd#L258


jkotas (approved):
Thank you!

PR #1036, opened 2021-04-29
Sync files shared with IL Linker

MichalStrehovsky:
This just gets https://github.com/mono/linker/pull/1979 out of the way since it's very noisy and doesn't apply to us.


jkotas (approved):
Thanks

PR #1040, opened 2021-04-29
Fix Benchmarks running

jkoritzinsky:
Fix the benchmark build/run to actually build the forwarder config as the baseline for the benchmarks.

Add benchmarks for the SafeHandle marshaller.

### DIFF
--- a/DllImportGenerator/Benchmarks/Program.cs
+++ b/DllImportGenerator/Benchmarks/Program.cs
@@ -19,7 +19,8 @@ namespace Benchmarks
             var config = DefaultConfig.Instance
                 .AddJob(GetDefaultJob())
                 .AddJob(GetForwarderJob())
-                .AddDiagnoser(MemoryDiagnoser.Default);
+                .AddDiagnoser(MemoryDiagnoser.Default)
+                .WithOptions(ConfigOptions.KeepBenchmarkFiles);
 
             BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args, config);
         }
@@ -44,6 +45,7 @@ namespace Benchmarks
                     new GenerateForwarderBuilder(settings),
                    new DotNetCliExecutor(settings.CustomDotNetCliPath)))
                 .WithId("Built-in")
+                .WithArguments(new []{ new MsBuildArgument("/p:DllImportGenerator_GenerateForwarders=true") })
                 .AsBaseline();
         }
     }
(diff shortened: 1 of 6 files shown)


DllImportGenerator/Benchmarks/SafeHandles.cs
```diff
@@ -0,0 +1,78 @@
...
+        [GeneratedDllImport(NativeExportsNE_Binary, EntryPoint = "modify_handle")]
+        public static partial void ModifyHandle(ref NativeExportsSafeHandle handle, [MarshalAs(UnmanagedType.I1)] bool newHandle);
+    }
+
+    public class SafeHandles
+    {
+        [Benchmark]
+        public bool ReturnSafeHandle()
+        {
+            using NativeExportsNE.NativeExportsSafeHandle handle = NativeExportsNE.AllocateHandle();
```
AaronRobinsonMSFT: This is measuring more than return. It is also including the `Dispose` call and overhead of the `try`/`finally`. I think we need a slightly different test design here.

DllImportGenerator/Benchmarks/SafeHandles.cs
```diff
@@ -0,0 +1,78 @@
...
+            return handle.IsInvalid;
+        }
+
+        [Benchmark]
+        public bool ByRefHandle()
```
AaronRobinsonMSFT: Same as above. We are measuring a lot of non-related code - at least based on what I think we want to measure.

DllImportGenerator/Benchmarks/Program.cs
```diff
@@ -17,61 +10,15 @@ class Program
...
-                .AddJob(GetDefaultJob())
-                .AddJob(GetForwarderJob())
-                .AddDiagnoser(MemoryDiagnoser.Default);
+                .AddJob(Job.Default.WithId("Generated"))
+                .AddJob(Job.Default
+                    .WithCustomBuildConfiguration("Release_Forwarders")
+                    .WithId("Built-in")
+                    .AsBaseline())
+                .AddDiagnoser(MemoryDiagnoser.Default)
+                .WithOptions(ConfigOptions.KeepBenchmarkFiles);
```
jkoritzinsky:
```suggestion
                .AddDiagnoser(MemoryDiagnoser.Default);
```

AaronRobinsonMSFT (approved):

PR #1044, opened 2021-04-30
Allow targeting Windows/Linux ARM64

AntonLapounov:
* Allow targeting Windows/Linux ARM64 without setting `DisableUnsupportedError=true`.
* Bump `objwriter` library version used for developer builds.
* Mark x64-specific tests to avoid building them for ARM64.
* Avoid generating the `nativebuild.proj` file for skipped tests.


jkotas (approved):
Thank you!

PR #1046, opened 2021-04-30
Add detection of managed C++

MichalStrehovsky:
Managed C++ is going to crash the compiler, so might as well have a better message for it. This is in a different spot than crossgen2: https://github.com/dotnet/runtime/pull/38113. Both approaches have their advantages and disadvantages. For us it's probably better to do this lazily.

### DIFF
--- a/src/coreclr/tools/Common/Compiler/CompilerTypeSystemContext.cs
+++ b/src/coreclr/tools/Common/Compiler/CompilerTypeSystemContext.cs
@@ -191,7 +191,7 @@ namespace ILCompiler
             {
                 PEReader peReader = OpenPEFile(filePath, out mappedViewAccessor);
 #if !READYTORUN
-                if ((peReader.PEHeaders.CorHeader.Flags & (CorFlags.ILLibrary | CorFlags.ILOnly)) == 0)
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/Common/Compiler/CompilerTypeSystemContext.cs
```diff
@@ -190,6 +190,11 @@ private EcmaModule AddModule(string filePath, string expectedSimpleName, bool us
             try
             {
                 PEReader peReader = OpenPEFile(filePath, out mappedViewAccessor);
+#if !READYTORUN
+                if (peReader.HasMetadata && (peReader.PEHeaders.CorHeader.Flags & (CorFlags.ILLibrary | CorFlags.ILOnly)) == 0)
+                    throw new NotSupportedException($"Error: managed C++ is not supported: '{filePath}'");
```
jkotas:
```suggestion
                    throw new NotSupportedException($"Error: C++/CLI is not supported: '{filePath}'");
```
C++/CLI is the official name

jkotas (approved):
Thanks!

src/coreclr/tools/Common/Compiler/CompilerTypeSystemContext.cs
MichalStrehovsky: Just copy-pasted what crossgen2 prints, but I don't particularly like the message.

PR #1054, opened 2021-05-01
Fix rpath for macOS

jkotas:
macOS uses different macro for application directory

Unify test setup between OSX and Linux. No need to set env variables for PInvoke smoke test on Linux anymore.

Fixes / related to #1055


jkotas:
cc @kant2002

MichalStrehovsky (approved):
Thank you!

Issue #167, opened 2020-10-09
BSTR marshalling


MichalStrehovsky:
Interop/PInvoke/Array/MarshalArrayAsField/AsByValArray/AsByValArrayTest

kant2002:
I have a question about current implementation.
In the shared marshallers code I see that when using READYTORUN codepath used marshalling from `System.StubHelpers.AnsiBSTRMarshaler` class when ILC code-path used marshalling in the `InteropHelpers` class.
Why difference exists?

kant2002:
Other question which I forgot to ask. Is the BSTRMarshaller and other similar classes should belong to `Marshaller.cs` or in the `Marshaller.Aot.cs`?

jkotas:
> Why difference exists?

We have not done refactoring required to unify them. I expect we will do such refactoring as part https://github.com/dotnet/runtimelab/tree/feature/DllImportGenerator project eventually: We will just make the helpers to follow the same shape as what we will be used for source generators.

jkotas:
>  Is the BSTRMarshaller and other similar classes should belong to Marshaller.cs or in the Marshaller.Aot.cs?

The split between the two is not very clear. Just pick the one you like better.

kant2002:
> We will just make the helpers to follow the same shape as what we will be used for source generators.

So I can try to implement just BSTR for ILC code paths for now as this is what would be path forward?

jkotas:
Yes, that's fine

MichalStrehovsky:
Marshaller.cs is also used in crossgen2. Crossgen2 only implements a subset of the marshallers to avoid bugs. We don't have a choice in NativeAOT since there's no JIT to fall back to. This one should probably be in .Aot.cs. I don't think crossgen2 is going to use it any time soon.

kant2002:
What is `AnsiBSTR`? Is this same BSTR which allocated/deallocated using `SysAllocString / SysFreeString` just actual data not `wchar_t` but `char_t`?

jkotas:
Yes. It have not seen it be ever used, except in a very old Visual Basic code.

PR #1057, opened 2021-05-02
Add BSTR marshaller

kant2002:
Related to #167 and #306

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
@@ -223,34 +223,12 @@ namespace Internal.Runtime.CompilerHelpers
             return PInvokeMarshal.AnsiCharToWideChar(nativeValue);
         }
 
-        internal static unsafe char* StringToBstrBuffer(string str)
-        {
-            if (str == null)
-                return null;
-
-            int stringLength = str.Length;
-
-            char* buffer = (char*)Marshal.AllocCoTaskMem(sizeof(char) * (stringLength + 1) + 4);
-
-            fixed (char* pStr = str)
-            {
-                int size = stringLength * sizeof(char);
-                char* firstCharPtr = buffer + 4;
-                Buffer.MemoryCopy(pStr, firstCharPtr, size, size);
-                *(buffer + stringLength + 2) = '\0';
-                int* pLength = (int*)buffer;
-                *pLength = stringLength * 2;
-            }
-            return buffer;
-        }
-
         internal static unsafe string BstrBufferToString(char* buffer)
         {
             if (buffer == null)
                 return null;
 
-            int stringLength = *(int*)buffer;
-            return new string(buffer + 2, 0, stringLength / 2);
+            return Marshal.PtrToStringBSTR((IntPtr)buffer);
         }
 
         internal static unsafe IntPtr ResolvePInvoke(MethodFixupCell* pCell)
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
@@ -223,6 +223,36 @@ namespace Internal.Runtime.CompilerHelpers
             return PInvokeMarshal.AnsiCharToWideChar(nativeValue);
         }
 
+        internal static unsafe char* StringToBstrBuffer(string str)
+        {
+            if (str == null)
+                return null;
+
+            int stringLength = str.Length;
+
+            char* buffer = (char*)Marshal.AllocCoTaskMem(sizeof(char) * (stringLength + 1) + 4);
+
+            fixed (char* pStr = str)
+            {
+                int size = stringLength * sizeof(char);
+                char* firstCharPtr = buffer + 4;
+                Buffer.MemoryCopy(pStr, firstCharPtr, size, size);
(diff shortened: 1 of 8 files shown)


src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -925,4 +927,29 @@ protected override void TransformManagedToNative(ILCodeStream codeStream)
             throw new NotSupportedException();
         }
     }
+
+    class BSTRStringMarshaller : Marshaller
+    {
```
jkotas, line 932: This also needs to override `CleanupRequired` and provide implementation of `EmitElementCleanup`

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -223,6 +223,36 @@ public static unsafe char AnsiCharToWideChar(byte nativeValue)
...
+            }
+            return buffer;
+        }
+
+        internal static unsafe string BstrBufferToString(char* buffer)
+        {
+            if (buffer == null)
+                return null;
+
+            int stringLength = *(int*)buffer;
```
jkotas: This does not look right. The BSTR pointer points at the first char*, the length is at negative offset.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -223,6 +223,36 @@ public static unsafe char AnsiCharToWideChar(byte nativeValue)
             return PInvokeMarshal.AnsiCharToWideChar(nativeValue);
         }
 
+        internal static unsafe char* StringToBstrBuffer(string str)
+        {
+            if (str == null)
```
jkotas: This can just call `Marshal.StringToBSTR`. (Or the marshaller can just call `Marshal.StringToBSTR` directly.)

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -223,6 +223,36 @@ public static unsafe char AnsiCharToWideChar(byte nativeValue)
...
+            return buffer;
+        }
+
+        internal static unsafe string BstrBufferToString(char* buffer)
+        {
+            if (buffer == null)
+                return null;
+
+            int stringLength = *(int*)buffer;
+            return new string(buffer + 2, 0, stringLength / 2);
```
jkotas: This can just call `Marshal.PtrToStringBSTR`

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
kant2002: Ok. That's my overlook. I try to find that online, but cannot find any explicit reference. Will change accordingly.

kant2002:
@jkotas  I have a question. On Unix implementation https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Unix.cs#L151 `AllocBSTR` use `sizeof(IntPtr)` as length for size. But in the documentation https://docs.microsoft.com/en-us/previous-versions/windows/desktop/automat/bstr#remarks it mentioned that for Length allocated 4 byte. Same here https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/692a42a9-06ce-4394-b9bc-5d2a50440168 Since I today working on Gentoo, I have to de-allocate string using 8 bytes `Length` string. 

My question who is right and should I go back to Runtime repo with fix?

jkotas:
On Windows, BSTRs are implemented in the OS. The Windows OS implementation of BSTRs uses pointer-sized padding, and that is never going to change. Windows OS has APIs to allocate/free BSTRs (SysAllocString/SysFreeString). These APIs should be always used to work with BSTRs on Windows instead of calling the allocator directly. It explains why the Windows documentation does not explain how to allocate BSTRs directly using COM allocator. No code on Windows needs to know that.
  
On Unix, the BSTR type exists just to make it easier to port code from Windows to Unix, so the implementation simply matches Windows, with all its quirks.

src/tests/nativeaot/SmokeTests/PInvoke/PInvokeNative.cpp
```diff
@@ -265,6 +283,63 @@ DLL_EXPORT int __stdcall VerifyUnicodeStringRef(unsigned short **val)
...
+    unsigned short expected[] = { 'H', 'e', 'l', 'l', 'o', ' ', 'W', 'o', 'r', 'l', 'd', 0};
+
+    return CompareUnicodeString(val, expected);
+}
+
+DLL_EXPORT int __stdcall VerifyBSTRStringOut(unsigned short **val)
+{
+    if (val == NULL)
+        return 0;
+    unsigned short *p = (unsigned short *)MemAlloc(sizeof(unsigned short) * 14);
```
jkotas: It would be more correct to use `SysAllocString` / `SysFreeString` on Windows, and have stub implementations of `SysAllocString` / `SysFreeString` methods for Unix.

Alternatively, I think it would be ok to omit the BSTR tests from the NativeAOT smoke test. We have comprehensive BSTR tests in the regular test suite, and the NativeAOT-specific implementation is simple enough and unlikely to break.

jkotas (approved):
Thank you!

PR #1058, opened 2021-05-03
Allow shared libraries with managed code on Linux ARM64

### LINKED ISSUE #1041: Use local-dynamic TLS access model for Linux ARM64
At present we are using the local-exec TLS access model to read `tls_CurrentThread`. That does not work for NativeAOT code running in a shared library.
https://github.com/dotnet/runtimelab/blob/a957342aff3625a7be8a2c4c69814165d2805a1f/src/coreclr/nativeaot/Runtime/unix/unixasmmacrosarm64.inc#L191-L199
We should probably use the following snippet instead:
```asm
adrp    x0, :tlsdesc:var
ldr     xN, [x0, #:tlsdesc_lo12:var]
add     x0, x0, :tlsdesc_lo12:var
.tlsdesccall var
blr     xN
mrs     xN, tpidr_el0
add     xN, xN, x0
```
which trashes the `x0` and `lr` registers.

AntonLapounov:
Use TLS descriptors to access thread-local variables on Linux ARM64.  That is more efficient than calling `__tls_get_addr`.  When the code is linked into an executable, the linker replaces the `add` and `blr` instructions with `nop`s.  Fixes #1041.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/unix/unixasmmacrosarm64.inc
+++ b/src/coreclr/nativeaot/Runtime/unix/unixasmmacrosarm64.inc
@@ -47,17 +47,17 @@ C_FUNC(\Name):
 
 .macro PREPARE_EXTERNAL_VAR Name, HelperReg
         adrp \HelperReg, C_FUNC(\Name)
-        add	\HelperReg, \HelperReg, :lo12:C_FUNC(\Name)
+        add  \HelperReg, \HelperReg, :lo12:C_FUNC(\Name)
 .endm
 
 .macro PREPARE_EXTERNAL_VAR_INDIRECT Name, HelperReg
         adrp \HelperReg, C_FUNC(\Name)
-        ldr	\HelperReg, [\HelperReg, :lo12:C_FUNC(\Name)]
+        ldr  \HelperReg, [\HelperReg, :lo12:C_FUNC(\Name)]
 .endm
 
 .macro PREPARE_EXTERNAL_VAR_INDIRECT_W Name, HelperReg
         adrp x\HelperReg, C_FUNC(\Name)
-        ldr	w\HelperReg, [x\HelperReg, :lo12:C_FUNC(\Name)]
+        ldr  w\HelperReg, [x\HelperReg, :lo12:C_FUNC(\Name)]
 .endm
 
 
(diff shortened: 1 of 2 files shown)


src/coreclr/nativeaot/Runtime/unix/unixasmmacrosarm64.inc
```diff
@@ -188,12 +188,30 @@ C_FUNC(\Name):
...
+// which cannot be x0. Preserves all other registers.
 .macro INLINE_GET_TLS_VAR target, var
-    mrs	\target, tpidr_el0
-    add	\target, \target, #:tprel_hi12:\var, lsl #12
-    add	\target, \target, #:tprel_lo12_nc:\var
+    .ifc \target, x0
+        .error "target cannot be x0"
+    .endif
+
+    stp     x0, lr, [sp,#-0x10]!
```
AntonLapounov, line 198: Alternatively, we could pass two extra registers to trash.  The sequence of instructions below must use `x0`.  In performance-critical use cases we need to preserve both `x0` and `lr`.

jkotas: What does this change do to raw allocation speed `for (int i = 0; i < 1_000_000_000; i++) GC.KeepAlive(new object())`? Would an alternative make this better?

Another alternative would be to change the few places that use this macro on arm64 to use standard lr store/restore in prolog/epilog.

AntonLapounov: I can try to measure.  For an executable this PR replaces the 3 existing instructions with 6 new ones + 2 `nop`s.  Using trash registers would avoid saving to/loading from the stack at the cost of 2 extra `mov` instructions.  The only real alternative I see it to use the old `__tls_get_addr` resolution mechanism, which would replace those 3 instructions with 5–6 new ones, saving two `nop`s at the cost of worse performance in the shared library case.

> Another alternative would be to change the few places that use this macro on arm64 to use standard lr store/restore in prolog/epilog.

The standard prolog would also add `mov fp, sp` and we would still need to preserve `x0`, so we would have to execute more instructions this way.

jkotas: The performance of instructions like ` stp     x0, lr, [sp,#-0x10]!` is sometimes context specific. They are fast when they are part of the standard prolog/epilog pattern. They are sometimes slow in other contexts.

AntonLapounov: With this PR, the `stp` instruction is already the first one in `RhpNewFast`: https://github.com/dotnet/runtimelab/blob/e1674574a82fcf0f528641d4bb3cbf6b8947e524/src/coreclr/nativeaot/Runtime/arm64/AllocFast.S#L21-L28 I do not know if hardware treats that the same way as a standard prolog though; it might depend on the CPU type.

jkotas: Ah ok. I think it would be still useful to do a quick measument of the raw allocation speed to see how much this change is this giving up. But we are probably not going to do anything about it.

jkotas (approved):
Thanks

src/coreclr/nativeaot/Runtime/unix/unixasmmacrosarm64.inc
AntonLapounov: The time to perform one billion `object` allocations on the JIT team's Ubuntu ARM64 server is 13.0 seconds for the baseline and 13.8 seconds for this PR. Performance degradation is about 5.8–5.9%.  Using a pair of trash registers instead of the `stp` and `ldp` instructions results in slightly worse performance degradation: 5.9–6.0%.  I also tried to test the old `__tls_get_addr` resolution mechanism; however, it seems our ARM64-targeting toolchain does not even recognize the required `:tlsgd:` relocation prefix. (GCC allows to specify `-mtls-dialect=trad` to use the old mechanism, but I found no similar support in LLVM.)

If we care enough, we might place two versions of performance-critical routines in the ILCompiler package.  Also, if supporting this scenario is not important at the moment, we may postpone changes or place them under some `#if`.

jkotas: This is good enough for now. Thank you for garthering the numbers!

PR #1060, opened 2021-05-03
A couple native layout fixes

MichalStrehovsky:
I was looking at generating fewer type loader templates in the compiler this weekend (we currently generate a template for everything we compile, but we don't have to). The full fix might be more involved but I stumbled on a couple smaller bugs. See individual commits.


jkotas (approved):
LGTM. Thanks!

MichalStrehovsky:
Ended up finding a pre-existing bug caused by creating empty dictionaries if no template was found. 344c17e is the fix. I added a regression test that would have found it before this pull request too.

PR #1061, opened 2021-05-03
Add support for AnsiBSTR.

kant2002:
Related to #167

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
@@ -231,6 +231,31 @@ namespace Internal.Runtime.CompilerHelpers
             return Marshal.PtrToStringBSTR((IntPtr)buffer);
         }
 
+        internal static unsafe byte* StringToAnsiBstrBuffer(string s)
+        {
+            if (s is null)
+            {
+                return (byte*)IntPtr.Zero;
+            }
+
+            int stringLength = s.Length;
+            fixed (char* pStr = s)
+            {
+                int nativeLength = PInvokeMarshal.GetByteCount(pStr, stringLength);
+                byte* bstr = (byte*)Marshal.AllocBSTR(nativeLength);
+                PInvokeMarshal.ConvertWideCharToMultiByte(pStr, stringLength, bstr, nativeLength, bestFit: false, throwOnUnmappableChar: false);
+                return bstr;
+            }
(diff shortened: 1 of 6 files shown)


kant2002:
This is just a test that failure happens with current implementation. Will provide implementation soon. I just want to better understand that this is proper way how to mark items as implemented in the build

kant2002:
@jkotas or @MichalStrehovsky anybody knows why this is green? In my naivete, I expect this is to fail.

jkotas:
Per-PR CI runs only the smoke tests.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -231,6 +231,31 @@ internal static unsafe string BstrBufferToString(char* buffer)
...
+            if (s is null)
+            {
+                return (byte*)IntPtr.Zero;
+            }
+
+            int stringLength = s.Length;
+            fixed (char* pStr = s)
+            {
+                int nativeLength = PInvokeMarshal.GetByteCount(pStr, stringLength);
+                byte* bstr = (byte*)Marshal.AllocBSTR(nativeLength);
```
jkotas: I do not think that this works. `AllocBSTR` takes length in characters, but you are passing it length in bytes.

src/tests/issues.targets
```diff
@@ -1172,9 +1172,6 @@
         <ExcludeList Include="$(XunitTestBinBase)/Interop/PInvoke/Variant/VariantTest/*">
             <Issue>https://github.com/dotnet/runtimelab/issues/155: VARIANT marshalling</Issue>
         </ExcludeList>
-        <ExcludeList Include="$(XunitTestBinBase)/Interop/StringMarshalling/AnsiBSTR/AnsiBStrTest/*">
```
jkotas, line 1175: Did you run this test locally and verified that it is passing?

MichalStrehovsky:
/azp run runtimelab-nativeaot Pri-0 Tests

This should trigger the P0 test legs. I don't expect them to all pass, but we can see how the test in question is doing.

kant2002:
My apologies. Do not have time today to run the build. Will check how it is going

kant2002:
Even if Pri0 tests fails in unrelated areas, when I run locally, it fails. So I will fix that.

src/tests/issues.targets
kant2002: Yes. I do verify now.

jkotas (approved):
Thanks

PR #1063, opened 2021-05-03
Add support for null arrays in the blittable array marshaller's pinning optimization.

### LINKED ISSUE #1051: Passing null for array of blittable types results in NullReferenceException
Generated code doesn't handle `null` when marshalling an array of blittable types.
```C#
...
NativeExportsNE.Sum(null, 0);
```

In the generated code, we pass null to `MemoryMarshal.GetArrayDataReference` throws `NullReferenceException`:
```C#
fixed (int* __values_gen_native = &System.Runtime.InteropServices.MemoryMarshal.GetArrayDataReference(values))
    __retVal = Sum__PInvoke__(__values_gen_native, numValues);
```

jkoritzinsky:
Example generated code:

```csharp
void Foo(int[] p)
{
   ref int __byref_p = ref (p == null ? ref System.Runtime.CompilerServices.Unsafe.NullRef<int>() : ref System.Runtime.InteropServices.MemoryMarshal.GetArrayDataReference(p));
   fixed (int* __p_gen_native= &p)
   {
        Foo__PInvoke__(__p_gen_native);
   }
}
```

Fixes #1051

### DIFF
--- a/DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
+++ b/DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
@@ -56,6 +56,44 @@ namespace Microsoft.Interop
             var (managedIdentifer, nativeIdentifier) = context.GetIdentifiers(info);
             if (!info.IsByRef && !info.IsManagedReturnPosition && context.PinningSupported)
             {
+                string byRefIdentifier = $"__byref_{managedIdentifer}";
+                if (context.CurrentStage == StubCodeContext.Stage.Marshal)
+                {
+                    // ref <elementType> <byRefIdentifier> = <managedIdentifer> == null ? ref Unsafe.NullRef<<elementType>>() : ref MemoryMarshal.GetArrayDataReference(<managedIdentifer>);
+                    var unsafeNullRef =
+                        InvocationExpression(
+                            MemberAccessExpression(
+                                SyntaxKind.SimpleMemberAccessExpression,
+                                ParseTypeName(TypeNames.System_Runtime_CompilerServices_Unsafe),
+                                GenericName(Identifier("NullRef"))
+                                .WithTypeArgumentList(TypeArgumentList(
+                                    SingletonSeparatedList(GetElementTypeSyntax(info))))));
+
+                    var getArrayDataReference =
+                        InvocationExpression(
+                            MemberAccessExpression(
+                                SyntaxKind.SimpleMemberAccessExpression,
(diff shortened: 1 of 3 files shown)


DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
```diff
@@ -56,6 +56,44 @@ public override IEnumerable<StatementSyntax> Generate(TypePositionInfo info, Stu
...
+                                            SyntaxKind.EqualsExpression,
+                                            IdentifierName(managedIdentifer),
+                                            LiteralExpression(
+                                                SyntaxKind.NullLiteralExpression)),
+                                        RefExpression(unsafeNullRef),
+                                        RefExpression(getArrayDataReference)))))))));
+                }
                 if (context.CurrentStage == StubCodeContext.Stage.Pin)
                 {
                     // fixed (<nativeType> <nativeIdentifier> = &MemoryMarshal.GetArrayDataReference(<managedIdentifer>))
```
AaronRobinsonMSFT: This comment is out of date right?

DllImportGenerator/DllImportGenerator/Marshalling/BlittableArrayMarshaller.cs
```diff
@@ -56,6 +56,44 @@ public override IEnumerable<StatementSyntax> Generate(TypePositionInfo info, Stu
             var (managedIdentifer, nativeIdentifier) = context.GetIdentifiers(info);
             if (!info.IsByRef && !info.IsManagedReturnPosition && context.PinningSupported)
             {
+                string byRefIdentifier = $"__byref_{managedIdentifer}";
+                if (context.CurrentStage == StubCodeContext.Stage.Marshal)
+                {
+                    // ref <elementType> <byRefIdentifier> = <managedIdentifer> == null ? ref Unsafe.NullRef<<elementType>>() : ref MemoryMarshal.GetArrayDataReference(<managedIdentifer>);
```
AaronRobinsonMSFT: We should add an explicit comment about the non-null logic here and that it is an explicit decision for backwards compat and supporting API behavior. I think we typically document changes but this behavior is unusual and non-obvious enough that we should document this as an explicit decision.

jkotas:
>  `ref System.Runtime.CompilerServices.Unsafe.NullRef<int>()`

Cheaper way to create a null ref in this case is `ref *(int*)0`. It does not require inlined method call.

AaronRobinsonMSFT (approved):

PR #1072, opened 2021-05-04
Infer target architecture from RID

### LINKED ISSUE #1013: NativeAOT targets use `PlatformTarget` to find appropriate runtime pack, which is often wrong
The problem is that PlatformTarget is what's normally passed to the C# compiler to choose the appropriate output architecture for the PE file. This is often not the same as the target architecture passed by RuntimeIdentifier because many projects choose to publish for a particular platform but use IL-only binaries, so `PlatformTarget=AnyCPU`.

This seems to result in the ILCompiler package being unable to find the runtime pack and producing errors like:

```
/home/andy/.nuget/packages/microsoft.dotnet.ilcompiler/6.0.0-preview.5.21223.3/build/Microsoft.NETCore.Native.Publish.targets(74,5): error : The PrivateSdkAssemblies ItemGroup is required for _ComputeAssembliesToCompileToNative [/home/andy/code/runtime/src/coreclr/tools/aot/crossgen2/crossgen2.csproj]
```

I think the right long-term fix for this would be to try to share the code that ReadyToRun uses to find their runtime pack.

AntonLapounov:
`dotnet publish` should not use `PlatformTarget` as it may be set to `AnyCPU`. Fixes #1013.

### DIFF
--- a/src/tests/Common/CLRTest.NativeAot.targets
+++ b/src/tests/Common/CLRTest.NativeAot.targets
@@ -152,7 +152,7 @@ if defined RunNativeAot (
     <OutputType>Exe</OutputType>
     <OutputPath>%24(MSBuildProjectDirectory)\</OutputPath>
     <IntermediateOutputPath>%24(MSBuildProjectDirectory)\</IntermediateOutputPath>
-    <PlatformTarget>$(TargetArchitecture)</PlatformTarget>
+    <TargetArchitecture>$(TargetArchitecture)</TargetArchitecture>
     <Optimize>$(Optimize)</Optimize>
     <DebugSymbols>true</DebugSymbols>
     <NETCoreSdkVersion>$(NETCoreSdkVersion)</NETCoreSdkVersion>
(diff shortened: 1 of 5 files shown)


src/tests/Common/CLRTest.NativeAot.targets
```diff
@@ -152,7 +152,7 @@ if defined RunNativeAot (
     <OutputType>Exe</OutputType>
     <OutputPath>%24(MSBuildProjectDirectory)\</OutputPath>
     <IntermediateOutputPath>%24(MSBuildProjectDirectory)\</IntermediateOutputPath>
-    <PlatformTarget>$(TargetArchitecture)</PlatformTarget>
+    <TargetArchitecture>$(TargetArchitecture)</TargetArchitecture>
```
jkotas, line 155:
```suggestion
```

jkotas (approved):
Thank you!

src/tests/Common/CLRTest.NativeAot.targets
AntonLapounov: Are you sure it is not needed?  That is the content of the `CDATA` section, which will be written to the new project file with `$(TargetArchitecture)` replaced with its actual value. For instance, the next line in that `CDATA` section is:
```xml
<Optimize>$(Optimize)</Optimize>
```

jkotas: Ah ok. I have not looked at the context. Ignore my comment then.

AntonLapounov:
An intermittent failure in BasicThreading test?
```
/__w/1/s/artifacts/tests/coreclr/Linux.x64.Checked/nativeaot/SmokeTests/BasicThreading/BasicThreading/BasicThreading.sh: line 413: 28144 Aborted                 (core dumped) $LAUNCHER $ExePath "${CLRTestExecutionArguments[@]}"
```

PR #1074, opened 2021-05-05
Clean up native artifacts by default

MichalStrehovsky:
The condition was written incorrectly. We want to delete compilation artifacts by default.


jkotas (approved):

MichalStrehovsky:
I looked at the new EntrypointNotFoundException failures and they are https://github.com/dotnet/runtime/issues/52444.

PR #1077, opened 2021-05-06
Fix support for BStr marshalling

kant2002:
Previously I incorrectly assume that TBStr is same as BStr.
I have plans to address that after #1061 since TBStr require to switch between BStr or AnsiBStr implementation most likely.


MichalStrehovsky:
Isn't it the same? This might be the complete fix.

https://github.com/dotnet/runtimelab/blob/a828939a5542fc90653c639dadc52778ee476f02/src/coreclr/vm/mlinfo.cpp#L1683-L1694

MichalStrehovsky (approved):

kant2002:
Never expect this be that easy 😃

jkotas (approved):
Thanks

PR #1087, opened 2021-05-07
Mark Array.CreateInstance and Type.MakeArrayType AOT unsafe

MichalStrehovsky:
These potentially require runtime code generation, same as MakeGenericType.


jkotas (approved):
Thank you!

PR #1099, opened 2021-05-09
srm serialization updates

veanes:
Added internal methods Serialize and Deserialize to Regex to enable serialization and deserialization to/from a custom string representation (in visible/safe ascii).

Added ILLink file to declare internal Regex methods for release build, that are accessed from unit tests.

Updated slightly the format of output of serialization that now avoids '\n' so serialized DFA
regexes can be stored as single ascii text lines in a text file. 

Added unit tests to test basic serialization correctness (and also for dgml generation). Added also methods that can be used  to do lightweight performance testing when declared as facts and run as unit tests after the regex solution has been build in release mode.
 
Measured that for the 150 CredScan regexes the total size of the serialized regexes is roughly 
1KB per regex or 150KB in total. 
For these regexes the total construction time is 1.5sec (roughly 10ms per regex).
The total serialization time for them is roughly 15-20ms. (roughly 125 micro sec per regex)
The total deserialization time for them is roughly 35-40ms. (roughly 250 micro sec per regex)
The (de)serialization times seem acceptable and need not be further optimized I believe.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/ILLink/ILLink.Descriptors.LibraryBuild.xml
+++ b/src/libraries/System.Text.RegularExpressions/src/ILLink/ILLink.Descriptors.LibraryBuild.xml
@@ -1,0 +1,10 @@
+<linker>
+  <assembly fullname="System.Text.RegularExpressions">
+    <type fullname="System.Text.RegularExpressions.Regex">
+      <!-- called via reflection from System.Text.RegularExpressions.Tests -->
+      <method name="Serialize" />
+      <method name="Deserialize" />
+      <method name="SaveDGML" />
+    </type>
+  </assembly>
+</linker>
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
@@ -55,6 +55,13 @@ namespace System.Text.RegularExpressions.SRM
         internal System.Text.RegularExpressions.RegexOptions Options { get; set; }
 
         /// <summary>
+        /// Timeout for matching.
+        /// </summary>
+        private TimeSpan _matchTimeout;
+        private int _timeoutOccursAt;
+        private bool _checkTimeout;
+
+        /// <summary>
         /// Set of elements that matter as first element of A.
         /// </summary>
         internal BooleanClassifier A_StartSet;
@@ -247,6 +254,9 @@ namespace System.Text.RegularExpressions.SRM
             sb.Append(Regex.s_top_level_separator);
             //------------ fragment 12 -----------
             dt.Serialize(sb);
+            sb.Append(Regex.s_top_level_separator);
+            //------------ fragemnt 13 -----------
+            sb.Append(_matchTimeout.ToString());
         }
 
         /// <summary>
@@ -271,6 +281,7 @@ namespace System.Text.RegularExpressions.SRM
             A_fixedPrefix_ignoreCase = bool.Parse(fragments[10]);
             Ar_prefix = Base64.DecodeString(fragments[11]);
             dt = Classifier.Deserialize(fragments[12]);
+            _matchTimeout = TimeSpan.Parse(fragments[13]);
             if (A.info.ContainsSomeAnchor)
             {
                 //line anchors are being used when builder.newLinePredicate is different from False
@@ -295,11 +306,12 @@ namespace System.Text.RegularExpressions.SRM
         /// <summary>
         /// Constructs matcher for given symbolic regex
         /// </summary>
-        internal SymbolicRegexMatcher(SymbolicRegexNode<S> sr, CharSetSolver css, BDD[] minterms, RegexOptions options)
+        internal SymbolicRegexMatcher(SymbolicRegexNode<S> sr, CharSetSolver css, BDD[] minterms, RegexOptions options, TimeSpan matchTimeout)
         {
             if (sr.IsNullable)
                 throw new NotSupportedException(SRM.Regex._DFA_incompatible_with + "nullable regex (accepting the empty string)");
 
+            this._matchTimeout = matchTimeout;
             this.Options = options;
             this.StartSetSizeLimit = 1;
             this.builder = sr.builder;
@@ -484,6 +496,13 @@ namespace System.Text.RegularExpressions.SRM
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void CheckTimeout()
+        {
+            if (Environment.TickCount > _timeoutOccursAt)
+                throw new TimeoutException();
+        }
+
         /// <summary>
         /// Generate all matches.
         /// <param name="isMatch">if true return null iff there exists a match</param>
@@ -493,6 +512,13 @@ namespace System.Text.RegularExpressions.SRM
         /// </summary>
         public Match FindMatch(bool isMatch, string input, int startat = 0, int endat = -1)
         {
+            _checkTimeout = (System.Text.RegularExpressions.Regex.InfiniteMatchTimeout != _matchTimeout);
+            if (_checkTimeout)
+            {
+                // Using Environment.TickCount and not Stopwatch similar to the non-DFA case.
+                int _timeout = (int)(_matchTimeout.TotalMilliseconds + 0.5);
+                _timeoutOccursAt = Environment.TickCount + _timeout;
+            }
 #if UNSAFE
             if ((Options & RegexOptions.Vectorize) != RegexOptions.None)
             {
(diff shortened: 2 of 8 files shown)


src/libraries/System.Text.RegularExpressions/src/ILLink/ILLink.Descriptors.LibraryBuild.xml
```diff
@@ -0,0 +1,10 @@
+<linker>
+  <assembly fullname="System.Text.RegularExpressions">
+    <type fullname="System.Text.RegularExpressions.Regex">
+      <!-- called via reflection from System.Text.RegularExpressions.Tests -->
+      <method name="Serialize" />
```
danmoseley, line 5: So these are only exposed for tests -- remind me, how does serialization/deserialization get triggered in the regular case?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
```diff
@@ -484,6 +496,13 @@ private State<S> CreateNewTransition(State<S> q, S atom, int offset)
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
danmoseley, line 499: It's not important right now, but FYI we don't put AggressiveInlining on small methods like this as they will almost surely be inlined.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
```diff
@@ -493,6 +512,13 @@ private State<S> CreateNewTransition(State<S> q, S atom, int offset)
         /// </summary>
         public Match FindMatch(bool isMatch, string input, int startat = 0, int endat = -1)
         {
+            _checkTimeout = (System.Text.RegularExpressions.Regex.InfiniteMatchTimeout != _matchTimeout);
```
danmoseley, line 515: We only use underscores for instance fields (https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md)
But, if this is the normal style in the DFA code already, there's no need to change it all today, as long as it' consistent. It's not important at this point.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -91,72 +91,205 @@ public void TestStartSet()
...
+        //    moveSizes.Add(m);
+        //    alphSizes.Add(k);
+        //}
+        //Console.WriteLine(stateSizes.Count);
+        //string rawregex = @"((decryptionKey|validationKey)=""[a-zA-Z0-9]+"")";
+        //string rawregex = @"(?<Part0>(\b(10|11|12|[1-9])\b))\s*(?<Part1>(:)\s*(?<Part2>([0-5][0-9]))){0,1}\s*(?<Part3>(\b([Tt]he day before yesterday|day before yesterday|[Tt]he day after tomorrow|day after tomorrow|[Yy]esterday|[Tt]oday|[Tt]ommorrow|[Tt]omorrow|[Tt]ommorow|[Tt]mrw)\b))";
+        //string rawregex = @"(?<Part0>(tonight))\s*(?<Part1>(at)){0,1}\s*(?<Part2>(midnight))";
+        */
+
+        private const string inputfile = @"c:\tmp\vsinput.txt";
```
danmoseley, line 143: For temporary files, we make the test class derive from FileCleanupTestBase. Then you call GetTestFileName() one or more times. It will create a random name, in an appropriate path, and take care of deleting them after you tests are done.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -91,72 +91,205 @@ public void TestStartSet()
...
+            }
+            File.AppendAllText(outputfile, string.Format("\ntotal time: DFA:{0}ms, COM:{1}ms\n", totDFA, totCOM));
+        }
+
+        private static int MeasureMatchTime(Regex re, string tag, string input)
+        {
+            File.AppendAllText(outputfile, tag + ": warmup:...");
+            try
+            {
+                int t0 = System.Environment.TickCount;
```
danmoseley, line 192: For tests, it is cleaner to use var sw = Stopwatch.StartNew(), sw.ElapsedTicks, etc.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -91,72 +91,205 @@ public void TestStartSet()
...
+            File.WriteAllLines(serializedout, ser);
+            //deserialize
+            int totDeSerTime = System.Environment.TickCount;
+            //repeat ten times
+            for (int j = 0; j < 10; j++)
+                for (int i = 0; i < k; i++)
+                    drs[i] = Deserialize(ser[i]);
+            totDeSerTime = System.Environment.TickCount - totDeSerTime;
+            totDeSerTime = totDeSerTime / 10;
+            File.AppendAllText(outputfile, string.Format("\nconstr:{0}ms, serialization:{1}ms, deserialization:{2}ms\n", totConstrTime, totSerTime, totDeSerTime));
```
danmoseley, line 256: This is more of a diagnostic/debugging aid than a test, maybe worth keeping such things in their own file so that later when we clean up we can decide whether to #if them out or something.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
danmoseley: Ah, if you want to write out the elapsed time or results etc then the deleting part won't be helpful. I suggest to either put them in Path.GetTempPath() and/or plan to disable these from building later.

danmoseley (approved):
Just nit comments, nothing critical given we're not cleaning up the code yet.

src/libraries/System.Text.RegularExpressions/src/ILLink/ILLink.Descriptors.LibraryBuild.xml
veanes: It would be triggered by using reflection, same as in the unit tests:
MethodInfo _Deserialize = typeof(Regex).GetMethod("Deserialize", BindingFlags.NonPublic | BindingFlags.Static);
Regex Deserialize(string s) => _Deserialize.Invoke(null, new object[] { s }) as Regex;

MethodInfo _Serialize = typeof(Regex).GetMethod("Serialize", BindingFlags.NonPublic | BindingFlags.Instance);
string Serialize(Regex r) => _Serialize.Invoke(r, null) as string;
For example, CredScan would be ok with that, I asked specifically if using reflection would be fine for them.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
veanes: good to know. Will remove those during a later cleanup.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
veanes: In this case, _checkTimeout is in fact an instance field. I'll move its initialization into the constructor.
We will soon do a cleanup phase, going over the codebase and fixing the coding style consistently. 
Right now it is not consistent (but mixed ... as a leftover of various styles).

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: These are temporary changes that I intend to delete later, they are not really used in any of the regular unit tests but as  input/output files is a couple of lightweight profiling entry points for me in couple of unit tests that are actually commented out as unit tests when I do the PR (actually two methods: called TestRunPerformace and TestRunSerialization).
For temporary files I've usually been using in-memory StringWriter, and avoiding actual file creation (eg in the unit test that tests DGML generation).

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: I've been using TickCount in two profiling methods TestRunPerformace and TestRunSerialization. These are not really unit tests, but I can certainly switch to using Stopwatch there.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: Good point! I'll move all my debugging/profiling helper methods into a separate file that can just be deleted later.
I'll do it as the next thing actually. I'll create a new PR to take care of those edits. Currently these are mixed up with the regular unit tests that does create some confusion.
(I also need to add a unit test for detecting timeout of dfa.)

PR #1101, opened 2021-05-09
Support decimal marshalling

### LINKED ISSUE #175: Support Decimal marshalling
Interop/PInvoke/Decimal/DecimalTest

### LINKED ISSUE #969: Struct requires marshalling that is not yet supported by this compiler.
Im working on a .NET solution that involves the development of a native library compiled with NativeAOT.
Everything has been worked fine, but today I decided to add an additional feature to a structure and got the following message.

`Unhandled Exception: EETypeRva:0x00618CF0: Struct 'SvgPathNative' requires marshalling that is not yet supported by this compiler.`

It seems very strange to me because the only thing I changed in that struct was to add a Decimal field to it.  This exception occurs in the native library when trying to generate a delegate from a native pointer to a callback function that receives that struct as parameter.

Delegate:
`public delegate void CreateLayerPathDelegate(Guid processId, Guid layerId, SvgPathNative svgPathNative);`

Struct:

    public struct SvgPathNative : ISvgPath
    {
        public Int32 Index { get; set; }
        public Guid Id { get; set; }
        public IntPtr PointsGuidId { get; set; }
        public Int32 Length { get; set; }
        public Guid CenterId { get; set; }
        public Decimal Area { get; set; }
    }

Exception line:
`CreateLayerPathDelegate createLayerPathDelegate = createLayerPath.GetDelegate<CreateLayerPathDelegate>();`

Extension method:
`public static T GetDelegate<T>(this IntPtr ptr) where T : Delegate =>
            Marshal.GetDelegateForFunctionPointer<T>(ptr);`

Update:
Trying to pass Decimal field as parameter modifiying delegate definition and keeping struct definition got:
`Unhandled Exception: EETypeRva:0x006193D8: Method '[S.P.CompilerGenerated]Internal.CompilerGenerated.PInvokeDelegateWrapper__CreateLayerPathDelegate.ForwardNativeFunctionWrapper__CreateLayerPathDelegate(Guid,Guid,SvgPathNative,Decimal)' requires marshalling that is not yet supported by this compiler.`

hez2010:
Support decimal marshalling.

Contributes to #175
Fixes #969

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
@@ -223,6 +223,16 @@ namespace Internal.Runtime.CompilerHelpers
             return PInvokeMarshal.AnsiCharToWideChar(nativeValue);
         }
 
+        internal static double DateTimeToOleDateTime(DateTime value)
+        {
+            return value.ToOADate();
+        }
+
+        internal static DateTime OleDateTimeToDateTime(double value)
+        {
+            return DateTime.FromOADate(value);
+        }
+
         internal static unsafe string BstrBufferToString(char* buffer)
         {
             if (buffer == null)
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
@@ -256,6 +256,33 @@ namespace Internal.Runtime.CompilerHelpers
             return Marshal.PtrToStringAnsi((IntPtr)buffer, (int)Marshal.SysStringByteLen((IntPtr)buffer));
         }
 
+        internal static unsafe decimal NativeToManagedDecimal(NativeDecimal* value)
+        {
+            byte* pValue = (byte*)value;
+
+            var signscale = *(ushort*)(pValue + 2);
+            var hi32 = *(uint*)(pValue + 4);
+            var lo64 = *(ulong*)(pValue + 8);
+
+            var sign = (byte)((signscale & 0xFF00) >> 8);
+            var scale = (byte)(signscale & 0xFF);
+            var mid32 = (int)((lo64 & 0xFFFFFFFF00000000) >> 32);
+            var lo32 = (int)(lo64 & 0xFFFFFFFF);
+
+            return new decimal(lo32, mid32, (int)hi32, sign != 0, scale);
+        }
+
+        internal static unsafe NativeDecimal ManagedToNativeDecimal(decimal* value)
+        {
+            byte* pValue = (byte*)value;
+
+            var signscale = *(ushort*)(pValue + 2);
+            var hi32 = *(uint*)(pValue + 4);
+            var lo64 = *(ulong*)(pValue + 8);
+
+            return new NativeDecimal { Hi32 = hi32, Lo64 = lo64, SignScale = signscale };
+        }
+
         internal static unsafe IntPtr ResolvePInvoke(MethodFixupCell* pCell)
         {
             if (pCell->Target != IntPtr.Zero)
@@ -555,6 +582,15 @@ namespace Internal.Runtime.CompilerHelpers
             public IntPtr MethodName;
             public ModuleFixupCell* Module;
             public CharSet CharSetMangling;
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct NativeDecimal
+        {
+            public ushort Reserved;
+            public ushort SignScale;
+            public uint Hi32;
+            public ulong Lo64;
         }
     }
 }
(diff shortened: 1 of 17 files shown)


hez2010:
Oops seems that it only works on Windows. Maybe caused by paddings or endians, need to figure out why it doesn't work on other platforms.

jkotas:
> The layout of native decimal is the same with managed decimal struct,

The alignment requirements of the managed Decimal and the unmanaged DECIMAL are different. CoreCLR solves it using https://github.com/dotnet/runtime/blob/main/src/coreclr/System.Private.CoreLib/src/System/StubHelpers.cs#L1082

kant2002:
You can try modify `dirs.proj` by replacing lines

https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/tests/Common/dirs.proj#L36-L38

with 
```
<AllProjects Include="$(TestRoot)Interop/PInvoke/Decimal/DecimalTest.csproj" Exclude="@(DisabledProjects)" />
```
and run `src\tests\build.cmd nativeaot Debug /p:SmokeTestsOnly=true`
then you can run existing tests using workflow for individual tests from https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/docs/workflow/building/coreclr/nativeaot.md#running-tests

hez2010:
@jkotas PTAL. CI failure is unrelated to this PR.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -266,6 +266,33 @@ internal static unsafe string AnsiBstrBufferToString(byte* buffer)
...
+            var signscale = *(ushort*)(pValue + 2);
+            var hi32 = *(uint*)(pValue + 4);
+            var lo64 = *(ulong*)(pValue + 8);
+
+            var sign = (byte)((signscale & 0xFF00) >> 8);
+            var scale = (byte)(signscale & 0xFF);
+            var mid32 = (int)((lo64 & 0xFFFFFFFF00000000) >> 32);
+            var lo32 = (int)(lo64 & 0xFFFFFFFF);
+
+            return new decimal(lo32, mid32, (int)hi32, sign != 0, scale);
```
jkotas: Calling this constructor will do some extra argument validation for `scale`. I do not see it in the CoreCLR decimal marshalers. CoreCLR marshalers just blasts the bits over. It would be nice to match what CoreCLR does.

jkotas:
I would expect the NativeDecimal type to show up in more places (e.g. also in `GetNativeTypeFromMarshallerKind`).

I am actually not sure anymore whether the NativeDecimal type is actually required to deal with the alignment issues. It was required for the old decimal implementation: https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/mscorlib/system/decimal.cs#L145-L148 , but it may not be required for the current one: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Decimal.cs#L106-L108.

What is not going to work if https://github.com/dotnet/runtimelab/pull/1101/files#diff-f8ce84af95c61383c81c774a30efbded7030042bc018027326e1f2c33ad00b22R83 is just changed to return `new BlittableValueMarshaller` ?

hez2010:
Actually I don't see the need of `NativeDecimal`. In current implementation of `System.Decimal`, the layout completely matches the unmanaged DECIMAL.
So I think it's okay to use `StructMarshaller`?

jkotas:
Yes, I think it should be ok to use `StructMarshaller` for decimal.

hez2010:
Updated.
~Does `GetNativeTypeFromMarshallerKind` also need to be updated to return `interopStateManager.GetStructMarshallingNativeType((MetadataType)type)` for decimals?~
Update: I don't think it's needed.

hez2010:
@jkotas I think it's ready for another look.

jkotas:
Could you  please also re-enable the decimal tests by deleting https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/tests/issues.targets#L1139-L1141 ?

hez2010:
Done.

hez2010:
The CI failure seems to be an I/O issue while copying files.

```
D:\workspace\_work\1\s\.dotnet\sdk\6.0.100-preview.3.21202.5\Microsoft.Common.CurrentVersion.targets(4502,5): error MSB3026: Could not copy "D:\workspace\_work\1\s\artifacts\obj\System.Collections.NonGeneric\ref\net6.0-Release\System.Collections.NonGeneric.dll" to "D:\workspace\_work\1\s\artifacts\bin\System.Collections.NonGeneric\ref\net6.0-Release\System.Collections.NonGeneric.dll". Beginning retry 1 in 1000ms. The requested operation cannot be performed on a file with a user-mapped section open. : 'D:\workspace\_work\1\s\artifacts\bin\System.Collections.NonGeneric\ref\net6.0-Release\System.Collections.NonGeneric.dll'  [D:\workspace\_work\1\s\src\libraries\System.Collections.NonGeneric\ref\System.Collections.NonGeneric.csproj]
```

jkotas:
> Contributes to #175

Unfortunately, this is not fully fixed yet. (It fails on BlittableStructPtrMarshaller now.)

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
```diff
@@ -353,7 +353,7 @@ public static partial class MarshalHelpers
                 {
                     if (nativeType == NativeTypeKind.Struct || nativeType == NativeTypeKind.Default)
                         return MarshallerKind.Decimal;
-                    else if (nativeType == NativeTypeKind.LPStruct && !isField && !isReturn)
+                    else if (nativeType == NativeTypeKind.LPStruct && !isField)
```
jkotas, line 356: Minor fix to match CoreCLR behavior that gets us a bit closer.

jkotas (approved):
Thank you!

PR #1102, opened 2021-05-09
Cleanup marshalling handling

### LINKED ISSUE #168: Invalid interop compatibility
Interop/PInvoke/Array/MarshalArrayAsField/AsLPArray/AsLPArrayTest

```
Test FAIL: TestLibrary.AssertTestException: Assert.Fail: Expected 'System.TypeLoadException' to be thrown, however 'System.Exception' was thrown. TakeIntArraySeqStructByVal
   at TestLibrary.Assert.HandleFail(String, String) + 0x5a
   at TestLibrary.Assert.Fail(String, Object[]) + 0x64
   at TestLibrary.Assert.Throws[T](Action, String, AssertThrowsOptions) + 0x1e1
   at Test.RunTest1(String) + 0xd4
   at Test.Main(String[]) + 0x29
```

kant2002:
Related to #167 and
Closes #168

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/TypeLoaderExceptionHelper.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/TypeLoaderExceptionHelper.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Runtime.InteropServices;
 
 using Internal.TypeSystem;
 
@@ -57,6 +58,11 @@ namespace Internal.Runtime
             throw new InvalidProgramException(SR.Format(GetFormatString(id), methodName));
         }
 
+        public static Exception CreateMarshalDirectiveException(ExceptionStringID id)
+        {
+            throw new MarshalDirectiveException(GetFormatString(id));
+        }
+
         // TODO: move to a place where we can share this with the compiler
         private static string GetFormatString(ExceptionStringID id)
         {
@@ -92,6 +98,8 @@ namespace Internal.Runtime
                     return SR.IO_FileNotFound_FileName;
                 case ExceptionStringID.BadImageFormatGeneric:
                     return SR.Arg_BadImageFormatException;
+                case ExceptionStringID.MarshalDirectiveGeneric:
+                    return SR.Arg_MarshalDirectiveException;
                 default:
                     Debug.Assert(false);
                     return "";
--- a/src/coreclr/tools/Common/TypeSystem/Common/ExceptionStringID.cs
+++ b/src/coreclr/tools/Common/TypeSystem/Common/ExceptionStringID.cs
@@ -41,5 +41,8 @@ namespace Internal.TypeSystem
         // BadImageFormatException
         BadImageFormatGeneric,
         BadImageFormatSpecific,
+
+        // MarshalDirectiveException
+        MarshalDirectiveGeneric,
     }
 }
--- a/src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.Aot.cs
+++ b/src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.Aot.cs
@@ -135,5 +135,25 @@ namespace Internal.TypeSystem.Interop
 
             return new PInvokeILStubMethodIL((ILStubMethodIL)emitter.Link(method), isStubRequired: true);
         }
+
+        public static MethodIL EmitExceptionBody(Exception exception, MethodDesc method)
+        {
+            ILEmitter emitter = new ILEmitter();
+
+            TypeSystemContext context = method.Context;
+            MethodSignature ctorSignature = new MethodSignature(0, 0, context.GetWellKnownType(WellKnownType.Void),
+                new TypeDesc[] { context.GetWellKnownType(WellKnownType.String) });
+            Type exceptionType = exception.GetType();
+            MethodDesc exceptionCtor = method.Context.SystemModule
+                .GetKnownType(exceptionType.Namespace, exceptionType.Name)
+                .GetKnownMethod(".ctor", ctorSignature);
+
+            ILCodeStream codeStream = emitter.NewCodeStream();
+            codeStream.Emit(ILOpcode.ldstr, emitter.NewToken(exception.Message));
+            codeStream.Emit(ILOpcode.newobj, emitter.NewToken(exceptionCtor));
+            codeStream.Emit(ILOpcode.throw_);
+
+            return new PInvokeILStubMethodIL((ILStubMethodIL)emitter.Link(method), isStubRequired: true);
+        }
     }
 }
--- a/src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.Aot.cs
+++ b/src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.Aot.cs
@@ -105,7 +105,7 @@ namespace Internal.TypeSystem.Interop
             }
             Debug.Assert(marshalAs != null && marshalAs.SizeConst.HasValue);
 
-            // if SizeConst is not specified, we will default to 1. 
+            // if SizeConst is not specified, we will default to 1.
             // the marshaller will throw appropriate exception
             uint size = 1;
             if (marshalAs.SizeConst.HasValue)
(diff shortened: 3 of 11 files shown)


src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
```diff
@@ -412,6 +412,15 @@ private MethodIL EmitIL()
                 Debug.Assert(!String.IsNullOrEmpty(ex.Message));
                 return MarshalHelpers.EmitExceptionBody(ex.Message, method);
             }
+            catch (System.Runtime.InteropServices.MarshalDirectiveException ex)
```
kant2002: Instead of catching specific exceptions, I can deliver them using `InnerException` of the `InvalidProgramException`.

src/tests/issues.targets
```diff
@@ -1154,13 +1148,10 @@
...
             <Issue>https://github.com/dotnet/runtimelab/issues/155: C++/CLI</Issue>
         </ExcludeList>
-        <ExcludeList Include="$(XunitTestBinBase)/Interop/PInvoke/Miscellaneous/HandleRef/HandleRefTest/*">
-            <Issue>https://github.com/dotnet/runtimelab/issues/168</Issue>
-        </ExcludeList>
         <ExcludeList Include="$(XunitTestBinBase)/Interop/PInvoke/NativeCallManagedComVisible/**/*">
             <Issue>https://github.com/dotnet/runtimelab/issues/155: COM</Issue>
         </ExcludeList>
-        <ExcludeList Include="$(XunitTestBinBase)/Interop/PInvoke/SizeParamIndex/**/*">
+        <ExcludeList Include="$(XunitTestBinBase)/Interop/PInvoke/SizeParamIndex/ReversePInvoke/**/*">
```
kant2002, line 1144: This folder has separate issues IMO. Maybe warrant to change Issue number here.

src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
MichalStrehovsky: We've been avoiding using the built in exception types because it's unclear whether the exception comes from our code here, or whether we hit an actual exception while trying to execute the compiler itself. That's why we have TypeSystemException and the derived types.

Those exception types then get even more complicated because the localization of the compiler doesn't have to match the localization of the BCL, so we avoid trying to come up with exception messages in the compiler and instead defer to the class library with the ExceptionStringID.

It's pretty hairy. If we want to do something with the exceptions, it should be in that direction, but I'm not completely sure if it's worth the trouble.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
```diff
@@ -980,6 +980,16 @@ protected TypeDesc NativeElementType
             }
         }
 
+        protected override void SetupArgumentsForReturnValueMarshalling()
+        {
+            ThrowHelper.ThrowTypeLoadException(ExceptionStringID.ClassLoadGeneral, ManagedType);
```
kant2002: Should I use different `ExceptionStringID` here?

kant2002:
@MichalStrehovsky I hope that I properly understand direction which should I go. I have some doubt points (like adding new ExceptionStringId) and miss new exception messages.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/TypeLoaderExceptionHelper.cs
```diff
@@ -92,6 +98,10 @@ private static string GetFormatString(ExceptionStringID id)
                     return SR.IO_FileNotFound_FileName;
                 case ExceptionStringID.BadImageFormatGeneric:
                     return SR.Arg_BadImageFormatException;
+                case ExceptionStringID.MarshalDirectiveGeneric:
+                    return SR.Arg_BadImageFormatException;
```
jkotas: This should use more appropriate message like `SR.Arg_MarshalDirectiveException` instead of `Arg_BadImageFormatException`.

src/coreclr/tools/Common/TypeSystem/Common/ExceptionStringID.cs
```diff
@@ -41,5 +41,9 @@ public enum ExceptionStringID
         // BadImageFormatException
         BadImageFormatGeneric,
         BadImageFormatSpecific,
+
+        // MarshalDirectiveException
+        MarshalDirectiveGeneric,
+        MarshalDirectiveInvalidSizeParamIndex,
```
jkotas: I am not sure whether we need to special case `MarshalDirectiveInvalidSizeParamIndex` here. I think using `MarshalDirectiveGeneric` for this case would be good enough.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
jkotas: When is this path going to be hit? What is the exception that CoreCLR throws in that case?

kant2002: Can I rebase? I think I need BSTR marshalling to figure out what's going on.

jkotas: Yes, feel free to squash and rebase.

src/coreclr/tools/Common/TypeSystem/Common/ExceptionStringID.cs
kant2002: I personally think that Marshal is complicated enough to have one blanket exception. If at least some areas produce slightly different exceptions it helpful for troubleshooting in my experience. But if you think that's add too much complexity, I will remove that.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
kant2002: This change was not needed. Removed.

src/coreclr/tools/Common/TypeSystem/Common/ExceptionStringID.cs
MichalStrehovsky: We have the [ILVerify tool](https://github.com/dotnet/runtime/tree/main/src/coreclr/tools/ILVerify) that can be used to validate inputs and provide a better experience around it. It could validate p/invoke signatures too. It can provide a lot more granular detailed output than this compiler. I would shoot for the minimally viable bar here.

(I was actually even considering just disabling the parts of the tests that test the invalid p/invokes. I will probably do that with some non-interop tests because we try to be compatible "within reason" - we don't go out of way to be 100% compatible.)

kant2002: Thanks for explanation.

src/coreclr/nativeaot/System.Private.CoreLib/src/Resources/Strings.resx
```diff
@@ -3130,6 +3130,9 @@
   <data name="Arg_AppDomainUnloadedException" xml:space="preserve">
     <value>Attempted to access an unloaded AppDomain.</value>
   </data>
+  <data name="Arg_MarshalDirectiveGeneric" xml:space="preserve">
```
jkotas: There is almost identical `Arg_MarshalDirectiveException` message already. We should just use that instead of introducing this one.

kant2002: @jkotas done.

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.Aot.cs
```diff
@@ -135,5 +135,25 @@ public static MethodIL EmitExceptionBody(string message, MethodDesc method)
 
             return new PInvokeILStubMethodIL((ILStubMethodIL)emitter.Link(method), isStubRequired: true);
         }
+
+        public static MethodIL EmitExceptionBody(Exception exception, MethodDesc method)
```
jkotas: I do not think this should be needed. Incorrectly resolved merge conflict?

kant2002: Yeah. That was left-over when my initial attempt :(

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
```diff
@@ -980,6 +980,11 @@ protected TypeDesc NativeElementType
             }
         }
 
+        protected override void SetupArgumentsForFieldMarshalling()
```
jkotas, line 983: Another left-over?

https://github.com/dotnet/runtimelab/pull/1102#discussion_r636997957

kant2002: No, this was for `AsLPArrayTest`. Actually I now understand why I have `SetupArgumentsForReturnValueMarshalling`. I was under impression that this is needed for symmetry. But there no tests covered returning as LPArray, so maybe I'm wrong here. I did not test on CoreCLR.

kant2002: I create small test and returning `S_INTArray_Seq` lead to 
```
--------------------------------------------------
Debug Assertion Violation

Message: Hardware exception raised inside the runtime.

Expression: 'ASSERT_UNCONDITIONALLY'

File: D:\d\kant\GitHub\dotnet\runtimelab\src\coreclr\nativeaot\Runtime\EHHelpers.cpp, Line: 560
--------------------------------------------------
```
with or without `SetupArgumentsForReturnValueMarshalling`, so that change definitely not helpful.

jkotas: Do you plan to revert this part then?

kant2002: I may confuse you. This line which we discuss are useful. Other one which I remove not.

jkotas (approved):
Thank you!

kant2002:
@jkotas thanks for your patience. I appreciate a lot that you help me push this through.

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
```diff
@@ -434,9 +434,6 @@ public static partial class MarshalHelpers
                 {
                     case NativeTypeKind.Array:
                         {
-                            if (isField || isReturn)
```
jkotas, line 437: This will need more refinement. The IsBlittable check is failing with stack overflow on structs that contain arrays of itself.  https://github.com/dotnet/runtime/pull/53650#discussion_r645939761

kant2002: Where should I make changes? In NativeAOT or I can go directly to crossgen? Probably for me would be more familiar to work with NativeAOT codebase, but if crossgen similarly easy to start with I can go there.

jkotas: Manish rolled back this line in https://github.com/dotnet/runtime/pull/53760 . It is fine for you to work on this in NativeAOT codebase.

PR #1103, opened 2021-05-10
Introduce design doc for Span/ReadOnlySpan<T> marshalling

jkoritzinsky:
This PR introduces a design doc for us to discuss how to implement marshalling support for `Span/ReadOnlySpan<T>` marshalling.

I started implementing Design 1, but when I started writing out the design doc, I realized the limitations of the design. I wrote up Design 2 as a proposal for handling the concept of "collection" marshallers so developers won't have to fork the DllImport marshaller to support default marshalling for generic collection-like types. This way, we could provide support for Span/ReadOnlySpan marshallers without fully inlining them into the generator iself and we can provide additional marshalers for Span and ReadOnlySpan without weird drawbacks and inconsistencies.

### DIFF
--- a/designs/SpanMarshallers.md
+++ b/designs/SpanMarshallers.md
@@ -55,7 +55,7 @@ Cons:
 
 An alternative option to fully inlining the stub would be to extend the model described in the [Struct Marshalling design doc](./StructMarshalling.md) to have custom support for collection-like types. By extending the model to be built with generic collection types in mind, many of the cons of the first approach would be resolved.
 
-Span marshalling would still be implemented with similar semantics as mentioned above in the Empty Spans section.Additional marshallers would still be provided as mentioned in the Additional proposed in-source marshallers section, but the `null` span marshaller would be able to be used in all cases, not just for empty spans.
+Span marshalling would still be implemented with similar semantics as mentioned above in the Empty Spans section. Additional marshallers would still be provided as mentioned in the Additional proposed in-source marshallers section, but the `null` span marshaller would be able to be used in all cases, not just for empty spans.
 
 ### Proposed extension to the custom type marshalling design
 
@@ -92,7 +92,7 @@ The `GenericCollectionMarshallerAttribute` attribute is applied to a generic mar
 
 #### Generic collection marshaller shape
 
-A generic collection marshaller would be required to have the following shape, in addition to the requirements for marshaler types used with the `NativeTypeMarshalllingAttribute`, excluding the constructors.
+A generic collection marshaller would be required to have the following shape, in addition to the requirements for marshaler types used with the `NativeTypeMarshallingAttribute`, excluding the constructors.
 
 ```csharp
 [GenericCollectionMarshaller]
@@ -104,7 +104,7 @@ public struct GenericCollectionMarshallerImpl<T, U, V,...>
     
     public const int StackBufferSize = /* */; // required if the span-based constructor is supplied.
 
-    // These method is required if marshalling from managed to native is supported.
+    // This method is required if marshalling from managed to native is supported.
     public TCollectionElement GetManagedValueAtIndex(int i);
     // This method is required if marshalling from native to managed is supported.
     public void SetManagedValueAtIndex(int i, TCollectionElement value);
@@ -121,7 +121,7 @@ public struct GenericCollectionMarshallerImpl<T, U, V,...>
 }
 ```
 
-The constructors now require an additional `int` parameter specifying the native size of the collection elements, represented as `TCollectionElement` above. As the elements may be marshalled to types with different native sizes than managed, this enables the author of the generic collection marshaller to not need to know how to marshal the elements of the collection, just the collection structure itself.
+The constructors now require an additional `int` parameter specifying the native size of a collection element. The collection element type is represented as `TCollectionElement` above, and can be any type the marshaller defines. As the elements may be marshalled to types with different native sizes than managed, this enables the author of the generic collection marshaller to not need to know how to marshal the elements of the collection, just the collection structure itself.
 
 The `GetManagedValueAtIndex` method and `Count` getter are used in the process of marshalling from managed to native. The generated code will iterate through `Count` elements (retrieved through `GetManagedValueAtIndex`) and assign their marshalled result to the address represented by `GetOffsetForNativeValueAtIndex` called with the same index. Then either the `Value` property getter will be called or the marshaller's `GetPinnableReference` method will be called, depending on if pinning is supported in the current scenario.
 
--- a/designs/SpanMarshallers.md
+++ b/designs/SpanMarshallers.md
@@ -49,10 +49,6 @@ Cons:
 
 ## Design 2: "Out-line" default support with extensions to Native Type Marshalling
 
-<!--
-    Idea: Extend the StructMarshalling proposal with a design for a generic collection-based marshalling pattern. That will enable outlining the Span/ReadOnlySpan/array/etc. marshallers while enabling developers to enable marshalling any collection type they want without us adding types to the generator.
--->
-
 An alternative option to fully inlining the stub would be to extend the model described in the [Struct Marshalling design doc](./StructMarshalling.md) to have custom support for collection-like types. By extending the model to be built with generic collection types in mind, many of the cons of the first approach would be resolved.
 
 Span marshalling would still be implemented with similar semantics as mentioned above in the Empty Spans section. Additional marshallers would still be provided as mentioned in the Additional proposed in-source marshallers section, but the `null` span marshaller would be able to be used in all cases, not just for empty spans.
@@ -162,7 +158,7 @@ The `ElementIndirectionLevel` property is added to support supplying marshalling
 private static partial void Bar([MarshalUsing(CountParameterName = nameof(count)), MarshalUsing(ConstantElementCount = 10, ElementIndirectionLevel = 1), MarshalUsing(typeof(FooMarshaler), ElementIndirectionLevel = 2)] List<List<Foo>> foos, int count);
 ```
 
-Multiple `MarshalUsing` attributes can only be supplied on the same parameter or return value if the `ElementIndirectionLevel` property is set to distinct values.
+Multiple `MarshalUsing` attributes can only be supplied on the same parameter or return value if the `ElementIndirectionLevel` property is set to distinct values. One `MarshalUsing` attribute per parameter or return value can leave the `ElementIndirectionLevel` property unset. This attribute controls the marshalling of the collection object passed in as the parameter.
 
 Alternatively, the `MarshalUsingAttribute` could provide a `Type ElementNativeType { get; set; }` property instead of an `ElementIndrectionLevel` property and support specifying the native type of the element of the collection this way. However, this design would block support for marshalling collections of collections.
 
@@ -217,13 +213,13 @@ This design could also be applied to support the built-in array marshalling if i
 
 #### Possible extension to the above model: Optimized support for sequential collections of blittable types
 
-If both the managed and native representation of a collection is sequential and the contents are blittable, additional optimizations, such as optimized block copying, can be emitted. however the above design does not support the necessary APIs to enable these optimizations. This section proposes that the following members be added to the design above. A `IsSequentialCollection` boolean property should be added to the `GenericCollectionMarshaller` attribute to specify that the collection is sequential in both managed and native representations. Then an additional method should be added to the generic collection model:
+If both the managed and native representation of a collection is sequential and the contents are blittable, additional optimizations, such as optimized block copying, can be emitted. however the above design does not support the necessary APIs to enable these optimizations. This section proposes that the following members be added to the design above. An `IsContiguousCollection` boolean property should be added to the `GenericCollectionMarshaller` attribute to specify that the collection is sequential in both managed and native representations. Then an additional method should be added to the generic collection model:
 
 ```csharp
 public ref TCollectionElement GetOffsetForManagedValueAtIndex(int index);
 ```
 
-This function would be required when `IsSequentialCollection` is `true` and would replace the `Get/SetManagedValueAtIndex` methods. When the elements of the collection are blittable, the marshaller will emit a block copy of the range `MemoryMarshal.CreateSpan(GetOffsetForManagedValueAtIndex(0), Count)` to the destination `MemoryMarshal.CreateSpan(MemoryMarshal.Cast<byte, TCollectionElement>(GetOffsetForNativeValueAtIndex(0), Count))`.
+This function would be required when `IsContiguousCollection` is `true` and would replace the `Get/SetManagedValueAtIndex` methods. When the elements of the collection are blittable, the marshaller will emit a block copy of the range `MemoryMarshal.CreateSpan(GetOffsetForManagedValueAtIndex(0), Count)` to the destination `MemoryMarshal.CreateSpan(MemoryMarshal.Cast<byte, TCollectionElement>(GetOffsetForNativeValueAtIndex(0), Count))`.
 
 When `TCollectionElement` is not blittable, the marshaller will iterate from `0..Count-1` and marshal the elements individually, as specified in the process above when the `Get/SetManagedValueAtIndex` methods are present.
 
--- a/designs/SpanMarshallers.md
+++ b/designs/SpanMarshallers.md
@@ -165,7 +165,7 @@ private static partial void Bar([MarshalUsing(typeof(ListAsArrayMarshaller<List<
 
 Multiple `MarshalUsing` attributes can only be supplied on the same parameter or return value if the `ElementIndirectionLevel` property is set to distinct values. One `MarshalUsing` attribute per parameter or return value can leave the `ElementIndirectionLevel` property unset. This attribute controls the marshalling of the collection object passed in as the parameter. The sequence of managed types for `ElementIndirectionLevel` is based on the elements of the `ManagedValues` span on the collection marshaller of the previous indirection level. For example, for the marshalling info for `ElementIndirectionLevel = 1` above, the managed type is the type of the following C# expression: `ListAsArrayMarshaller<List<Foo>>.ManagedValues[0]`.
 
-Alternatively, the `MarshalUsingAttribute` could provide a `Type ElementNativeType { get; set; }` property instead of an `ElementIndrectionLevel` property and support specifying the native type of the element of the collection this way. However, this design would block support for marshalling collections of collections.
+Alternatively, the `MarshalUsingAttribute` could provide a `Type ElementNativeType { get; set; }` property instead of an `ElementIndirectionLevel` property and support specifying the native type of the element of the collection this way. However, this design would block support for marshalling collections of collections.
 
 #### Example: Using generic collection marshalling for spans
 
--- a/designs/SpanMarshallers.md
+++ b/designs/SpanMarshallers.md
@@ -18,7 +18,7 @@ When a `(ReadOnly)Span<T>` is marshalled from native to managed, we will allocat
 
 ### Empty spans
 
-Since `(ReadOnly)Span<T>` does not have a way to distinguish from an empty collection and a `null` collection, we will provide a dummy pointer for zero-length/`null` spans to match the behavior of zero-length arrays. We will also provide an in-source marshaller that enables marshalling an empty span as `null`.
+We have decided to match the managed semantics of `(ReadOnly)Span<T>` to provide the smoothest default experience for span users in interop. To assist developers who wish to transition from array parameters to span parameters, we will also provide an in-source marshaller that enables marshalling a span that wraps an empty array as a non-null pointer as an opt-in experience.
 
 ### Additional proposed in-source marshallers
 
--- a/designs/SpanMarshallers.md
+++ b/designs/SpanMarshallers.md
@@ -47,11 +47,11 @@ Cons:
 - The MarshalAs attributes will continue to fail to work on spans used in non-source-generated DllImports, so this would be the first instance of enabling the "old" MarshalAs model on a new type in the generated DllImports, which may or may not be undesirable.
   - The existing "native type marshalling" support cannot support marshalling collections of an unknown (at marshaller authoring time) non-blittable element type and cannot specify an element count for collections during unmarshalling.
 
-## Design 2: "Out-line" default support with extensions to Native Type Marshalling
+## Design 2: "Out-line" default support with extensions to Native Type Marshalling for Contiguous Collections
 
 An alternative option to fully inlining the stub would be to extend the model described in the [Struct Marshalling design doc](./StructMarshalling.md) to have custom support for collection-like types. By extending the model to be built with generic collection types in mind, many of the cons of the first approach would be resolved.
 
-Span marshalling would still be implemented with similar semantics as mentioned above in the Empty Spans section. Additional marshallers would still be provided as mentioned in the Additional proposed in-source marshallers section, but the `null` span marshaller would be able to be used in all cases, not just for empty spans.
+Span marshalling would still be implemented with similar semantics as mentioned above in the Empty Spans section. Additional marshallers would still be provided as mentioned in the Additional proposed in-source marshallers section, but the non-`null` span marshaller and the no-alloc span marshaller would be able to be used in all cases, not just for empty spans.
 
 ### Proposed extension to the custom type marshalling design
 
@@ -100,30 +100,26 @@ public struct GenericCollectionMarshallerImpl<T, U, V,...>
     
     public const int StackBufferSize = /* */; // required if the span-based constructor is supplied.
 
-    // This method is required if marshalling from managed to native is supported.
-    public TCollectionElement GetManagedValueAtIndex(int i);
-    // This method is required if marshalling from native to managed is supported.
-    public void SetManagedValueAtIndex(int i, TCollectionElement value);
+    public Span<TCollectionElement> ManagedValues { get; }
 
-    // The getter is required if marshalling from managed to native is supported.
-    // The setter is required if marshalling from native to managed is supported.
-    public int Count { get; set; }
+    // Required only when unmarshalling is supported.
+    public void SetUnmarshalledCollectionLength(int length);
+
+    public IntPtr Value { get; set; }
+
+    public unsafe Span<byte> NativeValueStorage { get; }
 
     // The requirements on the Value property are the same as when used with `NativeTypeMarshallingAttribute`.
     // The property is required with the generic collection marshalling.
     public TNative Value { get; set; }
-
-    public ref byte GetOffsetForNativeValueAtIndex(int index);
 }
 ```
 
 The constructors now require an additional `int` parameter specifying the native size of a collection element. The collection element type is represented as `TCollectionElement` above, and can be any type the marshaller defines. As the elements may be marshalled to types with different native sizes than managed, this enables the author of the generic collection marshaller to not need to know how to marshal the elements of the collection, just the collection structure itself.
 
-The `GetManagedValueAtIndex` method and `Count` getter are used in the process of marshalling from managed to native. The generated code will iterate through `Count` elements (retrieved through `GetManagedValueAtIndex`) and assign their marshalled result to the address represented by `GetOffsetForNativeValueAtIndex` called with the same index. Then either the `Value` property getter will be called or the marshaller's `GetPinnableReference` method will be called, depending on if pinning is supported in the current scenario.
+When the elements of the collection are blittable, the marshaller will emit a block copy of the span `ManagedValues` to the destination `NativeValueStorage`. When the elements are not blittable, the marshaller will emit a loop that will marshal the elements of the managed span one at a time and store them in the `NativeValueStorage` span.
 
-The `SetManagedValueAtIndex` method and the `Count` setter are used in the process of marshalling from native to managed. The `Count` property will be set to the number of elements that the native collection contains, and the `Value` property will be assigned the result value from native code. Then the stub will iterate through the native collection `Count` times, calling `GetOffsetForNativeValueAtIndex` to get the offset of the native value and calling `SetManagedValueAtIndex` to set the unmarshalled managed value at that index.
-
-This design enables marshalling generic collections without requiring the owner of the collection type to know how to marshal every element in the collection.
+This would enable similar performance metrics as the current support for arrays as well as Design 1's support for the span types when the element type is blittable.
 
 #### Providing additional data for collection marshalling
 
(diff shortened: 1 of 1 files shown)


designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+- Inlining the span marshalling fully into the stub increases on-disk IL size.
+- This design does not enable developers to easily define custom marshalling support for their own collection types, which may be desireable.
+- The MarshalAs attributes will continue to fail to work on spans used in non-source-generated DllImports, so this would be the first instance of enabling the "old" MarshalAs model on a new type in the generated DllImports, which may or may not be undesirable.
+  - The existing "native type marshalling" support cannot support marshalling collections of an unknown (at marshaller authoring time) non-blittable element type and cannot specify an element count for collections during unmarshalling.
+
+## Design 2: "Out-line" default support with extensions to Native Type Marshalling
+
+<!--
+    Idea: Extend the StructMarshalling proposal with a design for a generic collection-based marshalling pattern. That will enable outlining the Span/ReadOnlySpan/array/etc. marshallers while enabling developers to enable marshalling any collection type they want without us adding types to the generator.
+-->
```
jkotas: Examples of unmanaged APIs that would be good candidates for the generic collection-based marshaling patterns?

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+
+By default, we will marshal `Span<T>` and `ReadOnlySpan<T>` similarly to array types. When possible, we will pin the `(ReadOnly)Span<T>`'s data and pass that down. When it is not possible to do so, we will try stack allocating scratch space for efficiency or allocate native memory with `Marshal.AllocCoTaskMem` for compatibility with arrays.
+
+To support marshalling from native to managed, we will support the same `MarshalAsAttribute` properties that arrays support today.
+
+When a `(ReadOnly)Span<T>` is marshalled from native to managed, we will allocate a managed array and copy the data from the native memory into the array.
+
+### Empty spans
+
+Since `(ReadOnly)Span<T>` does not have a way to distinguish from an empty collection and a `null` collection, we will provide a dummy pointer for zero-length/`null` spans to match the behavior of zero-length arrays. We will also provide an in-source marshaller that enables marshalling an empty span as `null`.
```
jkotas: I do not think this needs to match behavior of arrays by default, and it makes the behavior unintuitive and the marshaling less efficient than it needs for most users.

I think it would be ok to have this as an optional behavior.

AaronRobinsonMSFT: I would like to have further discussion on this. Part of the `Span<T>` design seemed to be being able to transition to it from an array to gain efficiency. This change in default behavior seems odd because now it require far more care in testing and validation. I'm not against this behavioral asymmetry but I would like to see (a) the performance impact in practice and (b) how impactful this change would be in our first target (i.e., NetCoreApp). I prefer the optimization as an opt-in because it is clear the user has accepted validating assumptions in their consumption of native APIs.

jkotas: This is not just about performance, it is also about intuitive behavior. I think having `default` Span to marshal as null pointer is much more intuitive than having it marshal as a random non-null pointer.

jkotas: And also about consistency between C# `fixed` statement and marshaling behavior.

jkotas: `default(byte[])` will marshal as null pointer today. I expect that there are number of cases where people use null array to pass null pointer into the underlaying API that really needs to see null.

Arrays have a strong distinction between default and zero-length. Span does not have a strong distinction between default and zero-length. We will need to choose which one of those Span cases is going to be less compatible with arrays.

AaronRobinsonMSFT: So is the expected behavior correct then:

```csharp
var t = new int[0];
Span<int> tt = t;  // Assume should be the same as if 't' were passed.
Span<int> ttt = null; // Assume null will be passed.
Span<int> tttt = new Span<int>(null, 0); // Assume null will be passed.
```

I want to make sure users updating arguments from `T[]` to `Span<T>` don't get into trouble here.

jkotas: The `Span` passed into P/Invoke would be comming from ordinary user code some of the time. The ordinary code working with Spans never makes distinction between `IsEmpty` span with null pointer and `IsEmpty` span with non-null pointer. In other words, the pointer value is irelevant for `IsEmpty` spans. It is why we have choosen to normalize the `IsEmpty` spans into null pointer for pinning. It avoids gotcha with behavior differences for the two different `IsEmpty` spans, and it is also good for security (guaranteed AV when accesing zero-sized buffer).

jkotas:
```
var t = new int[0];
Span<int> tt = t;  // Assume should be the same as if 't' were passed.
```
We can make this case work as you have written (same as if `t` were passed) by skipping the normalization and using the raw Span pointer. It would have the risk of behavior differences for the null vs. non-null Spans.

jkoritzinsky: It looks like MemoryMarshal.GetReference will give us the behavior described.

designs/SpanMarshallers.md
jkoritzinsky: Here's a few ideas I had for the patterns:

- Move arrays to use this pattern (intrinsic support in the generator)
- Span -> native array
- List -> linked-list
  - This could be useful if for example the JIT exposed a linked list type across the JIT interface body and crossgen2 needs to use it.
  - LLVM also has a few linked list types that LLVMSharp could use this with if they exposed those APIs
- Dictionary -> array of keys or Dictionary -> array of values
  - One good example of this is the coreclr hosting API
- Dictionary -> array of key-value-pair structures
  - The JVM's `JavaVMInitArgs` `options` field is a good example of this one.
- Win32 `SAFEARRAY` support could be implemented with this model for `SAFEARRAY`s containing a specific element type in the `VARIANT`s.

The nice part about this design is that one developer could provide a "marshal a Dictionary as an array of the keys" and someone else could provide a "marshal an object as the result of it's `ToString` method" and then a user who wants to pass a Dictionary down as the string values of its keys doesn't have to write a marshaller for the whole scenario. They'd only have to say "use the Dictionary keys marshaller for the dictionary, and use the `ToString` marshaller for the elements."

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Design 2: "Out-line" default support with extensions to Native Type Marshalling
+
+<!--
+    Idea: Extend the StructMarshalling proposal with a design for a generic collection-based marshalling pattern. That will enable outlining the Span/ReadOnlySpan/array/etc. marshallers while enabling developers to enable marshalling any collection type they want without us adding types to the generator.
+-->
+
+An alternative option to fully inlining the stub would be to extend the model described in the [Struct Marshalling design doc](./StructMarshalling.md) to have custom support for collection-like types. By extending the model to be built with generic collection types in mind, many of the cons of the first approach would be resolved.
+
+Span marshalling would still be implemented with similar semantics as mentioned above in the Empty Spans section.Additional marshallers would still be provided as mentioned in the Additional proposed in-source marshallers section, but the `null` span marshaller would be able to be used in all cases, not just for empty spans.
```
AaronRobinsonMSFT: ~~section.Additional~~ -> section. Additional

```suggestion
Span marshalling would still be implemented with similar semantics as mentioned above in the Empty Spans section. Additional marshallers would still be provided as mentioned in the Additional proposed in-source marshallers section, but the `null` span marshaller would be able to be used in all cases, not just for empty spans.
```

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+{
+
+}
+```
+
+The `GenericCollectionMarshallerAttribute` attribute is applied to a generic marshaler type with the "collection marshaller" shape described below. Since generic parameters cannot be used in attributes, open generic types will be permitted in the `NativeTypeMarshallingAttribute` constructor as long as they have the same arity as the type the attribute is applied to and generic parameters provided to the applied-to type can also be used to construct the type passed as a parameter.
+
+#### Generic collection marshaller shape
+
+A generic collection marshaller would be required to have the following shape, in addition to the requirements for marshaler types used with the `NativeTypeMarshalllingAttribute`, excluding the constructors.
```
AaronRobinsonMSFT: ~~`NativeTypeMarshalllingAttribute`~~ -> `NativeTypeMarshallingAttribute`

```suggestion
A generic collection marshaller would be required to have the following shape, in addition to the requirements for marshaler types used with the `NativeTypeMarshallingAttribute`, excluding the constructors.
```

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+public struct GenericCollectionMarshallerImpl<T, U, V,...>
+{
+    // these constructors are required if marshalling from managed to native is supported.
+    public GenericCollectionMarshallerImpl(GenericCollection<T, U, V, ...> collection, int nativeSizeOfElements);
+    public GenericCollectionMarshallerImpl(GenericCollection<T, U, V, ...> collection, Span<byte> stackSpace, int nativeSizeOfElements); // optional
+    
+    public const int StackBufferSize = /* */; // required if the span-based constructor is supplied.
+
+    // These method is required if marshalling from managed to native is supported.
+    public TCollectionElement GetManagedValueAtIndex(int i);
```
AaronRobinsonMSFT: How does this work with the `Dictionary` example? A `Tuple<K,V>`?

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+
+    // The requirements on the Value property are the same as when used with `NativeTypeMarshallingAttribute`.
+    // The property is required with the generic collection marshalling.
+    public TNative Value { get; set; }
+
+    public ref byte GetOffsetForNativeValueAtIndex(int index);
+}
+```
+
+The constructors now require an additional `int` parameter specifying the native size of the collection elements, represented as `TCollectionElement` above. As the elements may be marshalled to types with different native sizes than managed, this enables the author of the generic collection marshaller to not need to know how to marshal the elements of the collection, just the collection structure itself.
```
AaronRobinsonMSFT: > native size of the collection elements

The total size of the elements or the "native size of a collection element"?

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+[GenericCollectionMarshaller]
+public struct GenericCollectionMarshallerImpl<T, U, V,...>
+{
+    // these constructors are required if marshalling from managed to native is supported.
+    public GenericCollectionMarshallerImpl(GenericCollection<T, U, V, ...> collection, int nativeSizeOfElements);
+    public GenericCollectionMarshallerImpl(GenericCollection<T, U, V, ...> collection, Span<byte> stackSpace, int nativeSizeOfElements); // optional
+    
+    public const int StackBufferSize = /* */; // required if the span-based constructor is supplied.
+
+    // These method is required if marshalling from managed to native is supported.
```
AaronRobinsonMSFT: ~~These~~ -> This
```suggestion
    // This method is required if marshalling from managed to native is supported.
```

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+
+The `GetManagedValueAtIndex` method and `Count` getter are used in the process of marshalling from managed to native. The generated code will iterate through `Count` elements (retrieved through `GetManagedValueAtIndex`) and assign their marshalled result to the address represented by `GetOffsetForNativeValueAtIndex` called with the same index. Then either the `Value` property getter will be called or the marshaller's `GetPinnableReference` method will be called, depending on if pinning is supported in the current scenario.
+
+The `SetManagedValueAtIndex` method and the `Count` setter are used in the process of marshalling from native to managed. The `Count` property will be set to the number of elements that the native collection contains, and the `Value` property will be assigned the result value from native code. Then the stub will iterate through the native collection `Count` times, calling `GetOffsetForNativeValueAtIndex` to get the offset of the native value and calling `SetManagedValueAtIndex` to set the unmarshalled managed value at that index.
```
AaronRobinsonMSFT: I would be very interested in an example on matrices here - particularly higher order ones.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
++    public int ElementIndirectionLevel { get; set; }
++    public const string ReturnsCountValue = "return-value";
+}
+```
+
+The `MarshalUsingAttribute` will now provide a `CountParameterName` property that will point to a parameter whose value will hold the number of native collection elements, or to the return value if the value of `CountParameterName` is `ReturnsCountValue`. The `ConstantElementCount` property allows users to provide a constant collection length.
+
+> Open Question:
+> Should combining `CountParameterName` and `ConstantElementCount` in the same attribute be allowed?
+> With the `MarshalAs` marshalling, `SizeParamIndex` and `SizeConst` can be combined and the resulting size will be `paramValue(SizeParamIndex) + SizeConst`.
```
AaronRobinsonMSFT, line 156: I think this is a reasonable thing to support - although I would like to see know how many people actually use it.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+
+To support supplying information about collection element counts, a parameterless constructor is added to the `MarshalUsingAttribute` type. The default constructor specifies that the code generator should use the information in the attribute but use the default marshalling rules for the type.
+
+The `ElementIndirectionLevel` property is added to support supplying marshalling info for element types in a collection. For example, if the user is passing a `List<List<Foo>>` from managed to native code, they could provide the following attributes to specify marshalling rules for the outer and inner lists and `Foo` separately:
+
+```csharp
+private static partial void Bar([MarshalUsing(CountParameterName = nameof(count)), MarshalUsing(ConstantElementCount = 10, ElementIndirectionLevel = 1), MarshalUsing(typeof(FooMarshaler), ElementIndirectionLevel = 2)] List<List<Foo>> foos, int count);
+```
+
+Multiple `MarshalUsing` attributes can only be supplied on the same parameter or return value if the `ElementIndirectionLevel` property is set to distinct values.
```
AaronRobinsonMSFT: That isn't true in the above example - `MarshalUsing(CountParameterName = nameof(count))`.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+
+Alternatively, the `MarshalUsingAttribute` could provide a `Type ElementNativeType { get; set; }` property instead of an `ElementIndrectionLevel` property and support specifying the native type of the element of the collection this way. However, this design would block support for marshalling collections of collections.
```
AaronRobinsonMSFT: >However, this design would block support for marshalling collections of collections.

I like the spirit of this, but I am dubious of its value. The interop boundary is typically not the place for complex data structures beyond arrays and matrices of types. I am all for pushing the boundaries here but it does put a fair bit of cost that I am unsure enables something real in practice. Are their existing cases where marshalling a `List<List<T>>` or perhaps a `List<Set<T>>` is warranted?

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+    {
+      if (Value != IntPtr.Zero)
+      {
+          Marshal.FreeCoTaskMem(Value);
+      }
+    }
+}
+```
+
+This design could also be applied to support the built-in array marshalling if it is desired to move that marshalling out of the stub and into shared code.
```
AaronRobinsonMSFT: I am pro this.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+
+#### Possible extension to the above model: Optimized support for sequential collections of blittable types
+
+If both the managed and native representation of a collection is sequential and the contents are blittable, additional optimizations, such as optimized block copying, can be emitted. however the above design does not support the necessary APIs to enable these optimizations. This section proposes that the following members be added to the design above. A `IsSequentialCollection` boolean property should be added to the `GenericCollectionMarshaller` attribute to specify that the collection is sequential in both managed and native representations. Then an additional method should be added to the generic collection model:
```
AaronRobinsonMSFT: > `IsSequentialCollection`

I would propose instead `IsContiguousCollection`. A linked list is technically "sequential". Your point, I believe, is that the memory is contiguous so it can be block copied.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,247 @@
...
+- Collection type owners do not need to know how to marshal the elements of the collection.
+- Custom non-default marshalling of collections of non-blittable types supported with the same code as blittable types.
+- Sharing code for marshalling a given collection type reduces IL size on disk.
+- Developers can easily enable marshalling their own collection types without needing to modify the source generator.
+- Makes no assumptions about native collection layout, so collections like linked lists can be easily supported.
+
+Cons:
+
+- Introduces more attribute types into the BCL.
+- Introduces more complexity in the marshalling type model.
```
AaronRobinsonMSFT, line 311: This complexity should be a sliding scale. I want to understand the breadth of collections we should support here. I agree that supporting arbitrarily complex collections will make this complex, which might be the right thing, but we can start with a tiered approach for increasing levels of complexity. That might help us see where the benefit/complexity curve is at a max.

jkotas: I am looking at this as a performance optimization. 

Collections typically provide methods to convert from/to array or Span. It is not that much burden to tell people to convert from/to array. I think it is easy to explain how to convert from/to array and then marshal vs. how to use GenericCollectionMarshaller that combines these two operations together.

This scheme avoids a potential extra allocation of the array. I think that only really matters for blittable types, and for those there should be a path to avoid it using Span. For non-blittable types, the marshalling of each element will likely dominate the performance profile and so saving the extra allocation is not that big deal.

designs/SpanMarshallers.md
AaronRobinsonMSFT: >The ordinary code working with Spans never makes distinction between IsEmpty span with null pointer and IsEmpty span with non-null pointer.

I did not know this detail. I think I see your point here. @jkoritzinsky What are your thoughts?

designs/SpanMarshallers.md
AaronRobinsonMSFT: > I am looking at this as a performance optimization.
> I think it is easy to explain how to convert from/to array and then marshal vs. how to use GenericCollectionMarshaller that combines these two operations together.

I am interpreting these statements as saying the arbitrarily complex data structures shouldn't be a priority at all - which I can agree with given the potential for complexity in the area. If we look at this purely as performance related then the simpler the API the better.

designs/SpanMarshallers.md
jkoritzinsky: That detail is why I brought up this issue in the first place. I wasn't sure what experience we wanted to offer. Do we want to match how spans are used in regular managed code or how arrays are used in interop?

designs/SpanMarshallers.md
jkoritzinsky:
```suggestion
The constructors now require an additional `int` parameter specifying the native size of a collection element. The collection element type is represented as `TCollectionElement` above, and can be any type the marshaller defines. As the elements may be marshalled to types with different native sizes than managed, this enables the author of the generic collection marshaller to not need to know how to marshal the elements of the collection, just the collection structure itself.
```

designs/SpanMarshallers.md
jkoritzinsky: It depends on how the marshaller author wants to marshal the Dictionary.

Are they marshalling it as an array of key-value pairs? Then the return type would be `KeyValuePair<K, V>` or another pair-like type that has a marshaller defined by default.

Are they marshalling it separately as an array of keys and an array of values in two different parameters (like how the coreclr_initalize function takes configuration values)? Then there would be two marshaller implementations, one that returns `K` from `GetManagedValueAtIndex` and one that returns `V`.

designs/SpanMarshallers.md
AaronRobinsonMSFT: @jkoritzinsky For me I wasn't clear on the normalization that @jkotas is mentioning above. For clarity I am understanding this as `Span<T>(null, 0) == new T[0]` and `Span<T>(non-null, 0) == new T[0]`. The high order bit here is the length and not the pointer value - the normalization logic mentioned. @jkotas is indicating that the above ambiguity is by-design and therefore I assume documented in some manner? If so, I am inclined to agree with him and I believe you that the managed semantics are well understood and my concern about the surprise nature of converting from `T[]` to `Span<T>` isn't as much a concern as I may believe.

With this perspective in mind, I am now looking at this as the managed semantics are the priority here since one of our goals is to focus on the predictability of managed code semantics and continue to move things into managed code. Therefore, I agree prioritizing the semantics of how `Span<T>` is used in regular managed code.

designs/SpanMarshallers.md
jkoritzinsky: A simple method of implementing a matrix is with a jagged array `int[][]`, which would fall into the collection-of-collections model that this section supports.

Another example would be trying to pass an `object[][]` as a SAFEARRAY-of-SAFEARRAYs-of-VARIANTs, which I don't know of any API that does, but it is enabled by the native SAFEARRAY api.

designs/SpanMarshallers.md
jkoritzinsky: If we want to focus exclusively on array or Span-style inputs, then we can simplify the API by removing the `Get/SetManagedValueAtIndex` methods and instead using the `GetOffsetForManagedValueAtIndex` method as specified for the contiguous collections. As part of this, we would make the contiguous collections model the default "generic collection" model and drop the non-contiguous collection support (as well as support for collections where users cannot get a `ref` to the elements).

Then our guidance would be "convert to span or array, then use a collection marshaller to marshal from span or array to your desired native collection format", which in most cases would just be using the default support that marshals to a native array layout.

jkoritzinsky: I've updated the design to primarily focus on contiguous collections, with a blurb at the bottom about supporting non-contiguous collections in the future.

jkoritzinsky:
I've updated Design 2 to focus on contiguous collections and mention non-contiguous ones as an extension we can look at in the future but don't need to adopt today.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Default behavior
+
+This section describes the behavior of the default emitted-in-stub marshallers.
+
+By default, we will marshal `Span<T>` and `ReadOnlySpan<T>` similarly to array types. When possible, we will pin the `(ReadOnly)Span<T>`'s data and pass that down. When it is not possible to do so, we will try stack allocating scratch space for efficiency or allocate native memory with `Marshal.AllocCoTaskMem` for compatibility with arrays.
+
+To support marshalling from native to managed, we will support the same `MarshalAsAttribute` properties that arrays support today.
+
+When a `(ReadOnly)Span<T>` is marshalled from native to managed, we will allocate a managed array and copy the data from the native memory into the array.
```
AaronRobinsonMSFT, line 17: >we will allocate a managed array and copy the data from the native memory into the array.

This is only in the case where the managed callee has a `T[]` as opposed to a `Span<T>` right? I would assume the native caller would provide stable memory and if passes in a `int*,int` we can convert that to a `Span<int>` without allocating anything.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Empty spans
+
+We have decided to match the managed semantics of `(ReadOnly)Span<T>` to provide the smoothest default experience for span users in interop. To assist developers who wish to transition from array parameters to span parameters, we will also provide an in-source marshaller that enables marshalling a span that wraps an empty array as a non-null pointer as an opt-in experience.
+
+### Additional proposed in-source marshallers
+
+As part of this design, we would also want to include some in-box marshallers that follow the design laid out in the [Struct Marshalling design doc](./StructMarshalling.md) to support some additional scenarios:
+
+- A marshaler that marshals an empty span as `null`.
```
AaronRobinsonMSFT: >- A marshaler that marshals an empty span as `null`.

I'm sure I wasn't clear with this. You and Jan convinced me that the `null` for "empty" (length 0) or "null pointer" makes sense here. I think that should be default and the additional marshaler would be to match the `T[]` behavior.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+[NativeTypeMarshalling(typeof(DefaultSpanMarshaler<>))]
+public ref struct Span<T>
+{
+  ...
+}
+
+[GenericCollectionMarshaller]
+public ref struct DefaultSpanMarshaler<T>
+{
+
```
AaronRobinsonMSFT: Add ellipses.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+A generic collection marshaller would be required to have the following shape, in addition to the requirements for marshaler types used with the `NativeTypeMarshallingAttribute`, excluding the constructors.
+
+```csharp
+[GenericCollectionMarshaller]
+public struct GenericCollectionMarshallerImpl<T, U, V,...>
+{
+    // these constructors are required if marshalling from managed to native is supported.
+    public GenericCollectionMarshallerImpl(GenericCollection<T, U, V, ...> collection, int nativeSizeOfElements);
+    public GenericCollectionMarshallerImpl(GenericCollection<T, U, V, ...> collection, Span<byte> stackSpace, int nativeSizeOfElements); // optional
```
AaronRobinsonMSFT: ~~`nativeSizeOfElements`~~ -> `nativeSizeOfElement`

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+    
+    public const int StackBufferSize = /* */; // required if the span-based constructor is supplied.
+
+    public Span<TCollectionElement> ManagedValues { get; }
+
+    // Required only when unmarshalling is supported.
+    public void SetUnmarshalledCollectionLength(int length);
```
AaronRobinsonMSFT, line 112: Let's fully document each of these members as if we were proposing the API.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+
+```csharp
+private static partial void Bar([MarshalUsing(CountParameterName = nameof(count)), MarshalUsing(ConstantElementCount = 10, ElementIndirectionLevel = 1), MarshalUsing(typeof(FooMarshaler), ElementIndirectionLevel = 2)] List<List<Foo>> foos, int count);
+```
+
+Multiple `MarshalUsing` attributes can only be supplied on the same parameter or return value if the `ElementIndirectionLevel` property is set to distinct values. One `MarshalUsing` attribute per parameter or return value can leave the `ElementIndirectionLevel` property unset. This attribute controls the marshalling of the collection object passed in as the parameter.
+
+Alternatively, the `MarshalUsingAttribute` could provide a `Type ElementNativeType { get; set; }` property instead of an `ElementIndrectionLevel` property and support specifying the native type of the element of the collection this way. However, this design would block support for marshalling collections of collections.
```
AaronRobinsonMSFT: ~~`ElementIndrectionLevel`~~ -> `ElementIndirectionLevel`

designs/SpanMarshallers.md
AaronRobinsonMSFT: How would `T<U<A>, V<B>>` work here? I am thinking specifically as it relates to `ElementIndirectionLevel`.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+
+```csharp
+[GenericCollectionMarshaller]
+public ref struct SpanMarshaler<T>
+{
+    private Span<T> managedCollection;
+
+    private int nativeElementSize;
+
+    public SpanMarshaler(Span<T> collection, int nativeSizeOfElements)
```
AaronRobinsonMSFT: ~~`nativeSizeOfElements`~~ -> `nativeSizeOfElement`

designs/SpanMarshallers.md
AaronRobinsonMSFT: I see that the `SetUnmarshalledCollectionLength()` would be passed the unmarshal length that is returned by the native call. How does this get mapped on the actual P/Invoke?

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+
+#### Future extension to the above model: Non-contiguous collection support
+
+If a managed or native representation of a collection has a non-contiguous element layout, then developers currently will need to convert to or from array/span types at the interop boundary. This section proposes an API that would enable developers to convert directly between a managed and native non-contiguous collection layout as part of marshalling.
+
+An `IsContiguousCollection` boolean property could be added to the `GenericCollectionMarshaller` attribute, with a default value of `true`. When the value is `false`, that would specify that the collection is noncontiguous in either managed or native representations. Then additional methods should be added to the generic collection model, and some methods would be removed:
+
+```diff
+- [GenericCollectionMarshaller]
++ [GenericCollectionMarshaller(IsContiguousCollection = false)]
```
AaronRobinsonMSFT: How about renaming `GenericCollectionMarshaller` to be `GenericContiguousCollectionMarshaller`? We can then call the non-contiguous one something else.

designs/SpanMarshallers.md
```diff
@@ -0,0 +1,262 @@
...
+
+```csharp
+private static partial void Bar([MarshalUsing(CountParameterName = nameof(count)), MarshalUsing(ConstantElementCount = 10, ElementIndirectionLevel = 1), MarshalUsing(typeof(FooMarshaler), ElementIndirectionLevel = 2)] List<List<Foo>> foos, int count);
+```
+
+Multiple `MarshalUsing` attributes can only be supplied on the same parameter or return value if the `ElementIndirectionLevel` property is set to distinct values. One `MarshalUsing` attribute per parameter or return value can leave the `ElementIndirectionLevel` property unset. This attribute controls the marshalling of the collection object passed in as the parameter.
+
+Alternatively, the `MarshalUsingAttribute` could provide a `Type ElementNativeType { get; set; }` property instead of an `ElementIndrectionLevel` property and support specifying the native type of the element of the collection this way. However, this design would block support for marshalling collections of collections.
+
+#### Example: Using generic collection marshalling for spans
```
AaronRobinsonMSFT, line 170: Really appreciate this example. Can you provide an example of what the generated marshaller code would look like? I think it helps everyone understand how these mechanics will work from the perspective - here is how our source generator would use this type.

designs/SpanMarshallers.md
jkoritzinsky: Yes, I need to update this section as well. It is out of date.

designs/SpanMarshallers.md
jkoritzinsky: The default marshalling has to handle the blittable and non-blittable cases. The non-blittable case cannot use the native memory directly, so we can't use that in the default case. This proposal has the non-allocating `Span<T>` as an opt-in marshaller.

designs/SpanMarshallers.md
jkoritzinsky: This would be mapped similarly as the SizeParamIndex and SizeConst fields on MarshalAsAttribute (fetching them immediately after the call).

AaronRobinsonMSFT:
This LGTM. The only thing I would like understand better would be https://github.com/dotnet/runtimelab/pull/1103#discussion_r636539273. I didn't find an answer to that but admit I could have missed it due to terminology.

jkoritzinsky:
I added a blurb for that towards the end of the section (if theres a marshaller for `List<T>` named `ListAsArrayMarshaller<T>`):

> The sequence of managed types for `ElementIndirectionLevel` is based on the elements of the `ManagedValues` span on the collection marshaller of the previous indirection level. For example, for the marshalling info for `ElementIndirectionLevel = 1` above, the managed type is the type of the following C# expression: `ListAsArrayMarshaller<List<Foo>>.ManagedValues[0]`.

Does that answer your question?

AaronRobinsonMSFT:
>Does that answer your question?

Yes. I was misinterpreting the utility here. I understand the proposal now.

AaronRobinsonMSFT (approved):

PR #1104, opened 2021-05-10
added experiment file for local lightweight profiling and debugging

veanes:
Separated temporary experiment code from unit tests.
This file can simply be commented out/deleted in the final code, but is 
useful as a playground right now, mid-development.
It is also easy for anyone to use locally, and for example to test on regexes (like CredScan regexes or Cortana regexes) that cannot be checked in.

Updated serialization of timeout info a bit (omitting it when irrelevant).

Did other minor edits, mostly addressing notes in PR #1099.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
@@ -256,7 +256,8 @@ namespace System.Text.RegularExpressions.SRM
             dt.Serialize(sb);
             sb.Append(Regex.s_top_level_separator);
             //------------ fragemnt 13 -----------
-            sb.Append(_matchTimeout.ToString());
+            if (_checkTimeout)
+                sb.Append(_matchTimeout.ToString());
         }
 
         /// <summary>
@@ -281,7 +282,17 @@ namespace System.Text.RegularExpressions.SRM
             A_fixedPrefix_ignoreCase = bool.Parse(fragments[10]);
             Ar_prefix = Base64.DecodeString(fragments[11]);
             dt = Classifier.Deserialize(fragments[12]);
-            _matchTimeout = TimeSpan.Parse(fragments[13]);
+            string potentialTimeout = fragments[13].TrimEnd();
+            if (potentialTimeout == string.Empty)
+            {
+                _matchTimeout = System.Text.RegularExpressions.Regex.InfiniteMatchTimeout;
+                _checkTimeout = false;
+            }
+            else
+            {
+                _matchTimeout = TimeSpan.Parse(potentialTimeout);
+                _checkTimeout = true;
+            }
             if (A.info.ContainsSomeAnchor)
             {
                 //line anchors are being used when builder.newLinePredicate is different from False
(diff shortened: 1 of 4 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
```diff
@@ -496,10 +508,9 @@ private State<S> CreateNewTransition(State<S> q, S atom, int offset)
             }
         }
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private void CheckTimeout()
         {
-            if (Environment.TickCount > _timeoutOccursAt)
+            if (System.Environment.TickCount > _timeoutOccursAt)
```
danmoseley: nit, shouldn't need this prefix.

danmoseley: because TickCount is int, it will overflow at some random point within the next 49 days, so I think you need to do the same logic that is done here
https://github.com/dotnet/runtime/blob/e7b743cae325d4720fceaf7f84f68aaaf7bb7dbe/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexRunner.cs#L376-L400

danmoseley: we can tune the efficiency later, but that's a correctness thing so if you don't fix it, we should add a todo.

src/libraries/System.Text.RegularExpressions/tests/RegexExperiment.cs
```diff
@@ -0,0 +1,193 @@
...
+
+namespace System.Text.RegularExpressions.Tests
+{
+    /// <summary>
+    /// This class is to be ignored wrt unit tests.
+    /// It contains temporary experimental code, such as lightweight profiling and debuggging locally.
+    /// </summary>
+    public class RegexExperiment
+    {
+        private const string tmpWorkingDir = @"c:\tmp\";
```
danmoseley, line 19: At some point you may want this to work on Unix. (Although, generally we find CPU bound single threaded managed code has very similar characteristics on all platforms.)

src/libraries/System.Text.RegularExpressions/tests/RegexExperiment.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        private static int MeasureMatchTime(Regex re, string tag, string input)
+        {
+            File.AppendAllText(outputfile, tag + ": warmup:...");
+            Stopwatch sw = new Stopwatch();
+            int t = 0;
+            try
+            {
+                sw.Start();
```
danmoseley: nit, you can write `var sw = Stopwatch.StartNew();`

danmoseley (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
veanes: Ok, I'll copy the functionality from there and will do the same logic in srm. I'll add this as a fix to this pr.

veanes: will remove

src/libraries/System.Text.RegularExpressions/tests/RegexExperiment.cs
veanes: ok, will do

veanes:
I fixed the comments, in particular timeout checker implementation that now mimics that of RegexRunner. The frequency of checking timeout in DFA mode is lower though (set it to 5 -- instead of 1000) because each check corresponds to one transition of the automaton (so every 5 characters here) and can be a good chunk of work, frequency could even be set to 1 (but 5 should do). Perhaps measuring time with/without timeout could show if this adds any noticeable overhead.

danmoseley:
Sounds good. We can worry about that later.

PR #1105, opened 2021-05-10
Prepend identifiers with @ symbol since the Roslyn Symbol API strips the @ from the name.

### LINKED ISSUE #1096: [DllImportGenerator] Reserved keywords support
```cs
[GeneratedDllImport("Example")]
internal static partial void MyFunc([MarshalAs(UnmanagedType.LPUTF8Str)] string @event);
```

# Actual Result

CS8795: `Partial method 'MyFunc(string)' must have an implementation part because it has accessibility modifiers` in source file
CS0759: `No defining declaration found for implementing declaration of partial method 'MyFunc(string, event)'` in `DllImportGenerator.g.cs`

## Generated code

`internal static partial void MyFunc(string event)`

# Expected Result

## Generated code

`internal static partial void MyFunc(string @event)`

jkoritzinsky:
Fixes https://github.com/dotnet/runtimelab/issues/1096

### DIFF
--- a/DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
+++ b/DllImportGenerator/DllImportGenerator.UnitTests/CodeSnippets.cs
@@ -397,6 +397,18 @@ partial class Test
         public static string ByValueParameterWithModifier<T>(string attributeName) => ByValueParameterWithModifier(typeof(T).ToString(), attributeName);
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


AaronRobinsonMSFT (approved):
Ugh... I thoroughly dislike when keywords as used. However, the fix appears to be simple enough.

Thank you.

PR #1114, opened 2021-05-11
Support for default interface methods

MichalStrehovsky:
The original implementation strategy in #368 didn't have a straightforward extension towards working for canonical default interface methods (hard to inject the instantiation argument). Redoing things differently - this time default implementation methods get injected into the (sealed) vtable of the implementing class. There are no dispatch maps for interfaces and interfaces no longer need physical virtual slots on their EETypes (4 kB saving on a hello world as a bonus).

Support for canonical interfaces is still missing to keep the review reasonably small. Might be better to read commit by commit, since some of this was already reviewed.


jkotas (approved):
Thank you!

PR #1115, opened 2021-05-11
Don't assume everything except AsAny is IUnknown marshalling

MichalStrehovsky:
Should match the logic in mlinfo.cpp.


jkotas (approved):
Thanks

PR #1116, opened 2021-05-11
Use SemanticModel to discover GeneratedDllImportAttribute.

### LINKED ISSUE #1002: Discover the `GeneratedDllImport` via the semantic model rather than syntactic
Update to use the latest Roslyn Source Generator API - @chsienki has details on what new options exist for us.

> Use `GetSymbolInfo` to bind an `AttributeSyntax`, grab the type and compare the FQN of that type to the name you’re looking for. This means the generator will work for any attribute that has the right FQN.

The above approach will ensure that C# type aliasing can be used with our triggering attribute.

jkoritzinsky:
Fixes https://github.com/dotnet/runtimelab/issues/1002

### DIFF
--- a/DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
+++ b/DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
@@ -24,16 +24,11 @@ namespace Microsoft.Interop
 
         public void Execute(GeneratorExecutionContext context)
         {
-            var synRec = context.SyntaxReceiver as SyntaxReceiver;
-            if (synRec is null || !synRec.Methods.Any())
+            if (context.SyntaxContextReceiver is not SyntaxContextReceiver synRec
+                || !synRec.Methods.Any())
             {
                 return;
             }
-
-            // Get the symbol for GeneratedDllImportAttribute. If it doesn't exist in the compilation, the generator has nothing to do.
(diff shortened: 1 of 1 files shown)


DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -24,17 +24,12 @@ public class DllImportGenerator : ISourceGenerator
 
         public void Execute(GeneratorExecutionContext context)
         {
-            var synRec = context.SyntaxReceiver as SyntaxReceiver;
-            if (synRec is null || !synRec.Methods.Any())
+            if (context.SyntaxContextReceiver is not SyntaxContextReceiver synRec
```
AaronRobinsonMSFT, line 27: I guess this is fine. I would like to say though that having a side effect of a negated statement be a declaration that one can use is confusing and in my opinion not as clear as what it was previously. If this is preferred then okay, but if we are going for clarity this syntax is  confusing at compared to what it was previously.

AaronRobinsonMSFT (approved):

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
jkoritzinsky: Personally I like it just so that I don't need to have the mental model of what `as` does on failure in addition to the negative check.

PR #1118, opened 2021-05-12
Fix stack traces when framework resources are stripped

MichalStrehovsky:
Got tired of seeing

```
Unhandled Exception: EETypeRva:0x0011A0D0: TypeInitialization_Type_NoTypeAvailable
StackTrace_AtWordProgram!+0x754cd
StackTrace_AtWordProgram!+0x755be
StackTrace_AtWordProgram!+0xdf7e7
```

in new issues.

This is actually how CoreCLR also does this, so I just stole the code.


jkotas (approved):
Thanks

PR #1119, opened 2021-05-12
Refactor DllImportGenerator project for easier extensibility

jkoritzinsky:
Introduce a new `Microsoft.Interop.SourceGeneration` library which contains all of our marshallers and the default implementation of our marshaller selection algorithm.

Build the DllImport source generator on top of this library and refactor the guts of the P/Invoke body stub generation to be usable for more scenarios outside of `GeneratedDllImport`. For example, this model could be used for scenarios like the following:

- QCalls
  - These can't use local functions since they have compiler-generated names.
- OpenGL/Vulkan wrappers
  - Projects like these could use the refactored generator to generate P/Invokes that invoke function pointers instead of DllImport methods. As a result, this refactoring can enable these wrappers for these libraries to support https://github.com/dotnet/runtime/issues/1906 with their own GeneratedDllImport-like attribute that they provide along with their source generator.


DllImportGenerator/Microsoft.Interop.SourceGeneration/IGeneratorDiagnostics.cs
```diff
@@ -0,0 +1,69 @@
...
+            TypePositionInfo info,
+            string? notSupportedDetails);
+
+        /// <summary>
+        /// Report diagnostic for configuration that is not supported by the DLL import source generator
+        /// </summary>
+        /// <param name="attributeData">Attribute specifying the unsupported configuration</param>
+        /// <param name="configurationName">Name of the configuration</param>
+        /// <param name="unsupportedValue">[Optiona] Unsupported configuration value</param>
+        void ReportConfigurationNotSupported(
```
AaronRobinsonMSFT, line 83: If this is a public API, I would prefer not to have default argument values and instead have method overrides.

DllImportGenerator/Microsoft.Interop.SourceGeneration/Marshalling/MarshallingGeneratorFactory.cs
```diff
@@ -9,133 +7,58 @@
...
+        public IMarshallingGenerator Create(
+            TypePositionInfo info,
+            StubCodeContext context);
     }
 
-    /// <summary>
-    /// Exception used to indicate marshalling isn't supported.
-    /// </summary>
-    internal class MarshallingNotSupportedException : Exception
+    public class DefaultMarshallingGeneratorFactory<TOptions> : IMarshallingGeneratorFactory
```
AaronRobinsonMSFT: Create these inheritance chains makes things difficult for changes. Can we try another pattern?

DllImportGenerator/DllImportGenerator/DllImportGeneratorOptions.cs
```diff
@@ -1,12 +1,15 @@
-using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
-using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Runtime.InteropServices;
 
 namespace Microsoft.Interop
 {
+    record DllImportGeneratorOptions(bool GenerateForwarders, bool UseMarshalType) : InteropGenerationOptions(UseMarshalType)
```
AaronRobinsonMSFT: Public?

DllImportGenerator/Microsoft.Interop.SourceGeneration/Marshalling/MarshallingGeneratorFactory.cs
jkoritzinsky: Would a model more similar to the decorator pattern (composition) be preferable?

DllImportGenerator/DllImportGenerator/DllImportGeneratorOptions.cs
jkoritzinsky: This record is specific to the options for the GeneratedDllImport generator, so I don't think we need to expose it publically.

jkotas:
> extensibility 

Note that analyzers and source generators with dependencies do not work well due to https://github.com/dotnet/roslyn/issues/41785.

jkoritzinsky:
I honestly thought Roslyn had fixed that by now.

jkoritzinsky:
I've converted the Microsoft.Interop.SourceGeneration package to be a source package and have the sources included in DllImportGenerator.

I feel like there's something weird with using resx files in a source package, but I don't remember. If anyone else knows anything about how resource strings should be done with source packages, I'd appreciate any pointers.

jkoritzinsky:
Based on offline conversations as well as the fact that the [source generator cookbook has instructions for generators with dependencies](https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md#use-functionality-from-nuget-packages), we've decided for the .NET 6 time frame to provide the Microsoft.Interop.SourceGeneration package as an assembly instead of as a source package. Since we're not actually planning on shipping it publicly for .NET 6 (same as the DllImportGenerator), we can revisit this decision in the .NET 7 timeframe.

AaronRobinsonMSFT (approved):

PR #1121, opened 2021-05-13
cleanup of experiment code

veanes:
Added method as a "unit test" TestRun() for experimenting with performance and correctness measurements.
This method is commented out but lives in the RegexExperiment class/file.
Currently assuming use of local path 'c:\tmp\runtimelab\' for the experimental data, this can be configured in a cleaner way.
The intent is that, while the code is public, the data and even the regexes are not.

Changed the global .runsettings test timeout limit to 50min, adding a local experiment.runsettings file did not work for some reason. 
Timeout was not updated and tests were still timing out after 5min, making the use of TestRun() impossible.

### DIFF
--- a/eng/testing/.runsettings
+++ b/eng/testing/.runsettings
@@ -2,7 +2,7 @@
 <RunSettings>
   <RunConfiguration>
     <!-- Timeout in ms, 5 minutes -->
-    <TestSessionTimeout>300000</TestSessionTimeout>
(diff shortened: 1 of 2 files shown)


danmoseley (approved):
I'm not really reviewing this test file, I think it's all about "getting the job done for now"

I believe if and when we merge this branch into main, we'll not fail to spot the .runsettings change and revert it.

PR #1122, opened 2021-05-13
Support for default interface methods on generic interfaces

MichalStrehovsky:
Add support for generating instantiating stubs. The implementation strategy is similar to how we do instantiating stubs on generic structs. Adds two kind of methods - the actual instantiating stub, and a fake method that is necessary so that we can access the instantiation argument from IL. The fake method is replaced with the real entrypoint when we ask for the reloc for it.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilerTypeSystemContext.IntefaceThunks.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilerTypeSystemContext.IntefaceThunks.cs
@@ -1,0 +1,287 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+using Internal.TypeSystem;
+using Internal.IL;
+using Internal.IL.Stubs;
+
+using Debug = System.Diagnostics.Debug;
+
+// Default interface method implementation thunks
+//
+// The problem with default interface methods and shared generic code is that for:
+//
+// interface IFoo<T>
+// {
+//     Type GetTheType() => typeof(T);
+// }
+//
+// The actual generated code when instantiated over a shareable instance (like object) is
+// just IFoo<__Canon>.GetTheType, for any shareable argument (there's no unique code
+// generated for IFoo<Object>/IFoo<string>/... - we just have IFoo<__Canon>).
+//
+// For the canonical code to know what the actual T is, we need to provide the instantiation
+// context somehow. We can't easily get it from `this` like we do for reference types
+// since the type might implement multiple instantiations of IFoo (`class Abc : IFoo<object>, IFoo<string> { }`)
+// and we wouldn't know which one we are executing for within the method body.
+//
+// So we end up passing the context same as for shared valuetype code (that also cannot
+// determine context from just `this`) - by adding an extra instantiation
+// argument. The actual code for IFoo<__Canon>.GetTheType looks something like this:
+//
+// Type IFoo__Canon__GetTheType(IFoo<__Canon> instance, EEType* context)
+// {
+//     return Type.GetTypeFromHandle(GetTypeHandleOfTInIFooCanon(context));
+// }
+//
+// Now we have a problem because this method expects an extra `context` argument
+// that will not be provided at the callsite, since the callsite doesn't know
+// where it will dispatch to (could be a non-default-interface-method).
+//
+// We solve this with an instantiating thunk. The instantiating thunk is the thing
+// we place in the vtable of the implementing type. The thunk looks like this:
+//
+// Type Abc_IFoo__Canon__GetTheType_Thunk(IFoo<__Canon> instance)
+// {
+//     return IFoo__Canon__GetTheType(instance, GetOrdinalInterface(instance.m_pEEType, 0));
+// }
+//
+// Notice the thunk now has the expected signature, and some code to compute the context.
+//
+// The GetOrdinalInterface method retrieves the specified interface EEType off the EEType's interface list.
+// The thunks are per-type (since the position in the inteface list is different).
+//
+// We hardcode the position in the interface list instead of just hardcoding the interface type
+// itself so that we don't require runtime code generation when a new type is loaded
+// (e.g. "class Abc<T> : IFoo<T> { }" and we MakeGenericType's a new Abc at runtime) -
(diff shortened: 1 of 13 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilerTypeSystemContext.IntefaceThunks.cs
```diff
@@ -0,0 +1,287 @@
...
+// {
+//     return IFoo__Canon__GetTheType(instance, GetOrdinalInterface(instance.m_pEEType, 0));
+// }
+//
+// Notice the thunk now has the expected signature, and some code to compute the context.
+//
+// The GetOrdinalInterface method retrieves the specified interface EEType off the EEType's interface list.
+// The thunks are per-type (since the position in the inteface list is different).
+//
+// We hardcode the position in the interface list instead of just hardcoding the interface type
```
jkotas, line 58: Nit: I do not think this would strictly require code-generation at runtime. We would store the interface type in the generic dictionary and the instantiated interface type would be fetched via generic dictionary lookup.

jkotas: I think we are basically doing a generic dictionary lookup here, except that we are conveniently using the existing data on EEType for it instead of burning new generic dictionary slots. It is similar how we reuse the instantiation information for `typeof(T)` instead of burning new generic slots.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -299,8 +299,10 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
                 // Add conditional dependencies for interface methods the type implements. For example, if the type T implements
                 // interface IFoo which has a method M1, add a dependency on T.M1 dependent on IFoo.M1 being called, since it's
                 // possible for any IFoo object to actually be an instance of T.
-                foreach (DefType interfaceType in defType.RuntimeInterfaces)
+                for (int interfaceIndex = 0; interfaceIndex < defType.RuntimeInterfaces.Length; interfaceIndex++)
```
jkotas: Cache `defType.RuntimeInterfaces` in a local?

jkotas: I see we have several instances of this pattern in the existing code as well. It would be nice to fix all instances.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilerTypeSystemContext.IntefaceThunks.cs
```diff
@@ -0,0 +1,287 @@
...
+                    key.InterfaceIndex == value.InterfaceIndex;
+            }
+            protected override bool CompareValueToValue(DefaultInterfaceMethodImplementationInstantiationThunk value1, DefaultInterfaceMethodImplementationInstantiationThunk value2)
+            {
+                return Object.ReferenceEquals(value1.TargetMethod, value2.TargetMethod) &&
+                    value1.InterfaceIndex == value2.InterfaceIndex;
+            }
+            protected override DefaultInterfaceMethodImplementationInstantiationThunk CreateValueFromKey(DefaultInterfaceMethodImplementationInstantiationThunkHashtableKey key)
+            {
+                TypeDesc owningTypeOfThunks = key.TargetMethod.Context.GeneratedAssembly.GetGlobalModuleType();
```
jkotas, line 141: Why not park these on the TargetMethod type?

jkotas (approved):
Nice. Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilerTypeSystemContext.IntefaceThunks.cs
MichalStrehovsky: If we were to use actual generic slots, we would need to make sure the injected method is reported from GetAllMethod on the type to make code like this work:

https://github.com/dotnet/runtimelab/blob/27148d32d6f7fbdf675989f4f34e70a1efa53120/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs#L407-L431

It's doable, but it's extra complexity since in GetAllMethods we don't quickly know whether there's any default-implemented methods. The above code is needed to ensure generic dictionaries look the same in multifile builds irrespective of what parts of the generic type were used.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilerTypeSystemContext.IntefaceThunks.cs
MichalStrehovsky: I guess we could do that.

MichalStrehovsky: Okay, I thought about it a little and it complicates the mangling. Right now we can just make this a Prefix-mangled method (the mangled name of the instantiating thunk is the mangled name of the target default interface method and an integer index into the interface list as part of the prefix). If we make these instance methods on the owning type, the mangled name will need to have the (mangled) owning type, target interface method and the index because there's 3 degrees of freedom. We don't have a nice scheme to produce mangled names for that.

jkotas: Make sense. I forgot about mangling.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilerTypeSystemContext.IntefaceThunks.cs
jkotas: I agree that the way you have done it is nice and simple. I was just commenting on the part of the comment that says doing it the other way would require code generation.

MichalStrehovsky: Yeah, I meant that if we were to record the exact instantiations in code - I didn't think about doing a generic lookup.

PR #1123, opened 2021-05-14
fix of startanchor performance bug in srm

veanes:
Start anchors where not eliminated from non-initial start states, i.e., when a regex of say ^[a-z]+ match is started from an input position i>0, then the search regex being used is in fact .*^[a-z]+, and 
it reduces to [] (nothing) because ^ is nullable only when i=0, this information was not stored so search kept using the regex .*^[a-z]+ although it is a dead-end. Bug was visible/caused poor performance in most regexes that start with ^.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/tests/RegexExperiment.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/RegexExperiment.cs
@@ -129,12 +129,12 @@ namespace System.Text.RegularExpressions.Tests
 
         private void TestRunSampleRegex()
         {
-            string rawregex = @"(?<Part0>(every\ week))\s*(?<Part1>(on)){0,1}\s*((?<Part2>(\b([Mm]ondays|[Tt]uesdays|[Ww]ednesdays|[Tt]hursdays|[Ff]ridays|[Ss]aturdays|[Ss]undays)\b))|(?<Part3>((\b([Mm]onday|[Tt]uesday|[Ww]ednesday|[Tt]hursday|[Ff]riday|[Ss]aturday|[Ss]unday)\b)|(\b([Mm]on|[Tt]ues|[Tt]ue|[Ww]ed|[Tt]hurs|[Tt]hur|[Tt]hu|[Ff]ri|[Ss]at|[Ss]un)\b))))";
+            string rawregex = @"^[a-z]+\-[a-z]+$";
             string input = File.ReadAllText(inputfile);
-            TestRunRegex("sample", rawregex, input);
+            TestRunRegex("sample", rawregex, input, true, true);
         }
 
-        private void TestRunRegex(string name, string rawregex, string input)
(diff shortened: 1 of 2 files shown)


src/libraries/System.Text.RegularExpressions/tests/RegexExperiment.cs
```diff
@@ -129,12 +129,12 @@ private static long MeasureMatchTime(Regex re, string input, out Match match)
 
         private void TestRunSampleRegex()
         {
-            string rawregex = @"(?<Part0>(every\ week))\s*(?<Part1>(on)){0,1}\s*((?<Part2>(\b([Mm]ondays|[Tt]uesdays|[Ww]ednesdays|[Tt]hursdays|[Ff]ridays|[Ss]aturdays|[Ss]undays)\b))|(?<Part3>((\b([Mm]onday|[Tt]uesday|[Ww]ednesday|[Tt]hursday|[Ff]riday|[Ss]aturday|[Ss]unday)\b)|(\b([Mm]on|[Tt]ues|[Tt]ue|[Ww]ed|[Tt]hurs|[Tt]hur|[Tt]hu|[Ff]ri|[Ss]at|[Ss]un)\b))))";
+            string rawregex = @"^[a-z]+\-[a-z]+$";
             string input = File.ReadAllText(inputfile);
-            TestRunRegex("sample", rawregex, input);
+            TestRunRegex("sample", rawregex, input, true, true);
```
danmoseley, line 134: style nit, feel fee to ignore: we usually name bool parameters at the caller site (space after colon) as it's self documenting. Occasionally we do this when passing a null for the same reason.
```suggestion
            TestRunRegex("sample", rawregex, input, viewDGML: true, dotStar: true);
```

danmoseley (approved):
I am not yet qualified to review srm/**cs so I'll leave that to @olsaarik

danmoseley:
This was causing a functional bug right, not just perf ? In which case can you please include (or enable for SRM) appropriate tests that it hit.

src/libraries/System.Text.RegularExpressions/tests/RegexExperiment.cs
veanes: Ok, I'll follow that style more consistently going forward, I've sometimes done that.

veanes:
The bug was actually not causing any functionality error, only performance issues. 
Essentially making regexes that start with ^ behave slow (but not incorrect).

veanes:
Just to explain a bit more when using a regex (or derivative) such as .*^[a-z]+. 
anywhere else but at the very beginning of an input it is in reality a dead-end and the search can immediately be terminated with match failed result. 
This did not happen so the search continued until the end of the input, essentially looping on a dead-end state, and finally, at the very end (say after a 100MB or so) determined there is no match when there are no more characters to read.

PR #1127, opened 2021-05-15
Make valuetypes unconditionally sealed

jkotas:


jkotas:
Noticed this while debugging the current integration blocker. Valuetypes that are never boxed are reported as non-sealed that confuses JIT optimizations. For example, `impIsClassExact` will return false for these valuetypes that will pesimize type checks that they are involved in.

jkotas:
There is a potential related issue in `impCastClassOrIsInstToTree` that expands `isinst` as a simple vtable check with no fallback. It won't work for EETypes with duplicate identities. I think it works ok currently since we only duplicate EEType identities for arrays that won't hit this path. Mentioning it here in case I am missing some corner case.

MichalStrehovsky:
Yes, I think it's just for things like strings. Hopefully this assert protects us:

https://github.com/dotnet/runtimelab/blob/e2f3b0b57eab54aa8801bab9ab40a5af80395272/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs#L827-L829

MichalStrehovsky (approved):
Was this for compiler-generated valuetypes? (They should be sealed per spec.)

jkotas:
A small repro:

```csharp
struct Mine<T> { }

static void Main()
{
    if (!(((object)new Mine<object>()) is Nullable<Mine<object>>))
        throw new Exception();
}
```

`Nullable<Mine<object>>` was not in `_constructedTypes`, IsEffectivelySealed returned false for it, and this ended up using the non-inlined `isinst` check using helper.

With this change, it will use the inlined `isinst` check that just compares `EEType*`.

MichalStrehovsky:
Ah. I think we can fix this by just simplifying the logic in the devirtualizer. https://github.com/dotnet/runtimelab/pull/1131

I hope the CI agrees. I didn't test it.

PR #1131, opened 2021-05-17
Allow devirtualizing into unallocated types

MichalStrehovsky:
The purpose of this code was to prevent devirtualization on sealed types that were never seen as allocated (the devirtualization wouldn't be possible at runtime since the `this` would have to be null; and the compiler would crash before we get there because we would attempt to compile code that wasn't scanned).

This should not be needed after tentative instance methods were introduced in dotnet/runtimelab#643.


jkotas (approved):
Thank you!

PR #1132, opened 2021-05-17
Throw BadImageFormat on calls to abstract methods

MichalStrehovsky:
We have a test for this. Also, the string ID doesn't actually have an exception string associated so this was failing to throw in the first place.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
@@ -1062,7 +1062,7 @@ namespace Internal.JitInterface
 
             if (directCall && targetMethod.IsAbstract)
             {
-                ThrowHelper.ThrowInvalidProgramException(ExceptionStringID.InvalidProgramCallAbstractMethod);
+                ThrowHelper.ThrowBadImageFormatException();
             }
 
             if (directCall && !allowInstParam && targetMethod.GetCanonMethodTarget(CanonicalFormKind.Specific).RequiresInstArg())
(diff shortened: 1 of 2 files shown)


jkotas (approved):
Thanks

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
```diff
@@ -1062,7 +1062,7 @@ private void getCallInfo(ref CORINFO_RESOLVED_TOKEN pResolvedToken, CORINFO_RESO
 
             if (directCall && targetMethod.IsAbstract)
             {
-                ThrowHelper.ThrowInvalidProgramException(ExceptionStringID.InvalidProgramCallAbstractMethod);
+                ThrowHelper.ThrowBadImageFormatException();
```
jkotas, line 1065: Should we fix it in all places, and also delete the `InvalidProgramCallAbstractMethod` string ID?

MichalStrehovsky: This is not the only occurrence of the problem. I've been trying to point out that we need to match CoreCLR exactly in code reviews, but places like https://github.com/dotnet/runtime/pull/215 just slip through the cracks. Crossgen2 folks have added quite a bit more of these that don't match CoreCLR behavior and don't actually have an exception string. They all need to be cleaned up at some point because they don't serve their purpose.

jkotas: Should we have an issue on this?

It looks odd that this piece of code here is out-of-sync with the identical code in Scanner and R2R CorInfo.

MichalStrehovsky: Ah. Scanner difference is definitely unintentional. We need to have a CI check that fails if CorInfoImpl is changed and there's no change to scanner, unless one types "I solemnly swear no scanner change is needed" as a PR comment.

I'll file an issue.

Issue #961, opened 2021-04-13
Top level exception filter


distee-ubi:
Hello,
I have an application in two variants. The _NET_ variant consumes a C# assembly via an C++/CLI bridge, while the _AOT_ variant links the assembly after building it into a static native AOT lib.

The C# assembly targets `netcore5.0` and runtime `win-x64`.
The application uses a SEH top level exception filter installed via `SetUnhandledExceptionFilter `to report crashes to developers.

Unhandled exceptions in the _NET_ application actually get reported via the installed top level filter.
Unhandled exceptions in the _AOT_ application never make it to the top level filter and the application is killed silently.
Even with some hacking that ensures that our top level filter cannot be replaced after being installed unhandled exceptions in the _AOT_ application never make it there.

**1)** Is the top level SEH filter mechanism used at all when dealing with unhandled exceptions?
**2)** How does process termination due to unhandled exception work?

I appreciate any answers, help, pointers!

Best,
dirk

jkotas:
NativeAOT does not have interop with SEH on Windows. Unhandled exceptions in NativeAOT compiled code call `RaiseFailFastException`.

MichalStrehovsky:
It's possible that your use-case would be solvable by hooking into the [`AppDomain.UnhandledException`](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.unhandledexception?view=net-5.0) event but I just looked into it and it doesn't look like we call it before the failfast. But we probably should; it's likely an omission.

agocke:
@distee-ubi Out of curiosity, do you anticipate being in the situation where you use Core/Native AOT side-by-side for a while? Is there a particular characteristic that makes you want one over the other?

distee-ubi:
@agocke Yes we will be using it side-by-side for a while. The top reason for trying to switch NativeAOT is performance, but we have some concerns that made us keep Core as a fallback. For example, we already have a Core build copy-protected by a third party, but do not know if this is going to work out for NativeAOT.

So far, we (and in particular I) do not have much experience with NativeAOT. Right now I can not give reason why I would prefer one over the other or keep them side-by-side. Some will pop up down the road, I guess.

distee-ubi:
@jkotas, @MichalStrehovsky Thanks for your insight!

Our _NET_ app uses `AppDomain.UnhandledException` to deal with unhandled exceptions originating in the managed world. If the delegate would be called with _AOT_  exceptions from the managed world would be handled straight away.

[...]

PR #1134, opened 2021-05-17
Added missing invoke for AppDomain.UnhandledException

AlexDoeringEnvision:
#961

### DIFF
--- a/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.cs
+++ b/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.cs
@@ -223,6 +223,26 @@ namespace System.Runtime
             }
         }
 
+        private static void OnUnhandledExceptionViaClassLib(object exception)
+        {
+            IntPtr pOnUnhandledExceptionFunction =
+                (IntPtr)InternalCalls.RhpGetClasslibFunctionFromEEType((IntPtr)exception.EEType, ClassLibFunctionId.OnUnhandledException);
+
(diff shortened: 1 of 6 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=1134) <br/>All CLA requirements met.

jkotas (approved):
Nice. Thank you!

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.cs
```diff
@@ -223,6 +223,26 @@ private static void OnFirstChanceExceptionViaClassLib(object exception)
...
+            if (pOnUnhandledExceptionFunction == IntPtr.Zero)
+            {
+                return;
+            }
+
+            try
+            {
+                ((delegate*<object, void>)pOnUnhandledExceptionFunction)(exception);
+            }
+            catch when (true)
```
Suchiman, line 240: > warning CS7095: Filter expression is a constant 'true', consider removing the filter
```suggestion
            catch
```

jkotas: `catch when (true)` is used in the rest of the file, so I think it is fine to use here as well for consistency.

We use to have a mode where the core runtime was a different type universe from the app (`System.Object` here was not the apps `System.Object`). This is left-over from that mode. `catch when (true)` was needed here to catch all exceptions. Regular `catch` would only catch the exceptions from this type universe.

PR #1137, opened 2021-05-18
Don't do guarded devirtualization for unallocated types

MichalStrehovsky:
PGO data can refer to types we have not seen during scanning. There's no point to do guarded devirtualization for them since they were not allocated in the program. The only thing the guarded devirtualization would do is crash the compiler.


jkotas (approved):
LGTM. Thanks

PR #1143, opened 2021-05-19
Allow more type preinitialization

MichalStrehovsky:
Set of rules to allow executing the static constructor at compile time for ~30 extra types in a Hello World.

* Allow freezing types with reference-typed fields that are readonly (We basically only support it if they end up being null, but it helps a lot for e.g. Array enumerators).
* Allow backwards branches and set an instruction limit instead
* Allow div_un and rem_un
* Allow reading array elements


jkotas (approved):
Nice!

PR #1146, opened 2021-05-20
Fix stresslog lifetime problem

jkotas:
The thread-local StressLog was used even after it was detached.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/thread.cpp
+++ b/src/coreclr/nativeaot/Runtime/thread.cpp
@@ -367,8 +367,20 @@ bool Thread::IsCurrentThread()
     return m_threadId.IsCurrentThread();
 }
 
+void Thread::Detach()
+{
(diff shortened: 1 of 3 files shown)


src/coreclr/nativeaot/Runtime/thread.cpp
```diff
@@ -394,12 +406,10 @@ void Thread::Destroy()
         delete[] m_pThreadLocalModuleStatics;
     }
 
-    RedhawkGCInterface::ReleaseAllocContext(GetAllocContext());
-
-    // Thread::Destroy is called when the thread's "home" fiber dies.  We mark the thread as "detached" here
-    // so that we can validate, in our DLL_THREAD_DETACH handler, that the thread was already destroyed at that
-    // point.
-    SetDetached();
+#ifdef STRESS_LOG
```
jkotas, line 409: RhHandleFree above used StressLog

jkotas:
Found this while tracing down failures in current integration

MichalStrehovsky (approved):
Thank you!

PR #1147, opened 2021-05-20
A minor clarification about the Watch window

WhiteBlackGoose:
I'm not used to opening the Watch windows via main menu, so spent some time trying to understand why there's no `Watch` button in `Debug` -> `Windows`.

(if it's obvious enough... then it's just me ¯\\\_(ツ)_/¯ )


MichalStrehovsky (approved):
Thank you!

PR #1149, opened 2021-05-21
On unhandled runtime exception windows clients can opt to continue se…

### LINKED ISSUE #961: Top level exception filter

distee-ubi:
…arch for handlers
Amends already closed #961

### DIFF
--- a/src/coreclr/nativeaot/Runtime/EHHelpers.cpp
+++ b/src/coreclr/nativeaot/Runtime/EHHelpers.cpp
@@ -449,16 +449,14 @@ int32_t __stdcall RhpHardwareExceptionHandler(uintptr_t faultCode, uintptr_t fau
 
 #else // TARGET_UNIX
 
-static bool g_ContinueSearchOnUnhandledRuntimeException = false;
+static bool g_ContinueOnFatalErrors = false;
 
 // Set the runtime to continue search when encountering an unhandled runtime exception. Once done it is forever.
 // Continuing the search allows any vectored exception handlers or SEH installed by the client to take effect.
 // Any client that does so is expected to handle stack overflows.
-EXTERN_C bool RhpContinueSearchOnUnhandledRuntimeException()
+EXTERN_C void RhpContinueOnFatalErrors()
 {
-    bool prev = g_ContinueSearchOnUnhandledRuntimeException;
-    g_ContinueSearchOnUnhandledRuntimeException = true;
-    return prev;
+    g_ContinueOnFatalErrors = true;
 }
 
 int32_t __stdcall RhpVectoredExceptionHandler(PEXCEPTION_POINTERS pExPtrs)
@@ -489,7 +487,7 @@ int32_t __stdcall RhpVectoredExceptionHandler(PEXCEPTION_POINTERS pExPtrs)
         }
         else if (faultCode == STATUS_STACK_OVERFLOW)
         {
-            if(g_ContinueSearchOnUnhandledRuntimeException)
+            if (g_ContinueOnFatalErrors)
             {
                 // The client is responsible for the handling.
                 return EXCEPTION_CONTINUE_SEARCH;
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=1149) <br/>All CLA requirements met.

src/coreclr/nativeaot/Runtime/EHHelpers.cpp
```diff
@@ -449,6 +449,18 @@ int32_t __stdcall RhpHardwareExceptionHandler(uintptr_t faultCode, uintptr_t fau
 
 #else // TARGET_UNIX
 
+static bool g_ContinueSearchOnUnhandledRuntimeException = false;
+
+// Set the runtime to continue search when encountering an unhandled runtime exception. Once done it is forever.
+// Continuing the search allows any vectored exception handlers or SEH installed by the client to take effect.
+// Any client that does so is expected to handle stack overflows.
+EXTERN_C bool RhpContinueSearchOnUnhandledRuntimeException()
```
jkotas:
```suggestion
EXTERN_C void RhpContinueOnFatalErrors()
```
The previous value is not really useful since the setting is applied forever. Also, I think fatal error better describes what this does.

src/coreclr/nativeaot/Runtime/EHHelpers.cpp
```diff
@@ -477,8 +489,13 @@ int32_t __stdcall RhpVectoredExceptionHandler(PEXCEPTION_POINTERS pExPtrs)
         }
         else if (faultCode == STATUS_STACK_OVERFLOW)
         {
-            // Do not use ASSERT_UNCONDITIONALLY here. It will crash because of it consumes too much stack.
+            if(g_ContinueSearchOnUnhandledRuntimeException)
```
jkotas:
```suggestion
            if (g_ContinueSearchOnUnhandledRuntimeException)
```
Nit

src/coreclr/nativeaot/Runtime/EHHelpers.cpp
```diff
@@ -515,6 +532,10 @@ int32_t __stdcall RhpVectoredExceptionHandler(PEXCEPTION_POINTERS pExPtrs)
         return EXCEPTION_CONTINUE_EXECUTION;
     }
 
+    // The client may have told us to continue to search for custom handlers,
+    // but in general we consider any form of hardware exception within the runtime itself a fatal error.
+    // Note this includes the managed code within the runtime.
+    if(!g_ContinueSearchOnUnhandledRuntimeException)
```
jkotas:
```suggestion
    if (!g_ContinueSearchOnUnhandledRuntimeException)
```
Nit

src/coreclr/nativeaot/Runtime/EHHelpers.cpp
distee-ubi: The return allows you to realize you did a second call to no effect, but I agree that this is not really useful.
I like 'FatalError' better than 'ContinueSearchOnUnh....'. It is also much shorter! I'll change the naming of the setter and the variable it acts on.

src/coreclr/nativeaot/Runtime/EHHelpers.cpp
distee-ubi: ah, habits ... will fix this.

jkotas (approved):
Thank you!

distee-ubi:
Thanks for your help!

PR #1150, opened 2021-05-21
Fixes to COM wrappers

kant2002:
- Create IUnknown with 1 as ref count
- Add error message when cannot QI to desired interface

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
@@ -257,7 +258,7 @@ namespace System.Runtime.InteropServices
             }
 
             mow->Target = RuntimeImports.RhHandleAlloc(instance, GCHandleType.Normal);
-            mow->RefCount = 0;
+            mow->RefCount = 1;
             mow->UserDefinedCount = userDefinedCount;
             mow->UserDefined = userDefined;
             mow->Flags = flags;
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -506,9 +506,11 @@ internal static unsafe IntPtr ComInterfaceForObject(object instance, Guid target
             int resultCode = wrapper->QueryInterface(in targetIID, out comObjectInterface);
             if (resultCode != 0)
             {
```
jkotas, line 511: This path leaks the `unknownPtr` reference.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -506,9 +506,11 @@ internal static unsafe IntPtr ComInterfaceForObject(object instance, Guid target
...
-                throw new PlatformNotSupportedException(SR.PlatformNotSupported_ComInterop);
+                throw new PlatformNotSupportedException(SR.InvalidOperation_InterfaceNotSupported);
```
jkotas: Is this the exception that CoreCLR throws in this situation? (I would expect it to be something else.)

kant2002: Cannot find location where it happens in CoreCLR. Will try again when would be home in hour or so. Will look at marshalling code.

kant2002: Is this proper location where look for object -> interface conversion.

https://github.com/dotnet/runtime/blob/aaaaeed8c8d2d01dccd1f7fc536ea27eee75f472/src/coreclr/vm/ilmarshalers.cpp#L1401

Maybe Marshal class will throw this exception too.
I account for case where ComWrappers does not provide correct interface vtbl

jkotas: The easiest way to find out is to write a small test.

src/coreclr/nativeaot/System.Private.CoreLib/src/Resources/Strings.resx
```diff
@@ -3160,4 +3160,7 @@
   <data name="InvalidOperation_ComInteropRequireComWrapperInstance" xml:space="preserve">
     <value>COM Interop requires ComWrapper instance registered for marshalling.</value>
   </data>
+  <data name="InvalidOperation_InterfaceNotSupported" xml:space="preserve">
```
jkotas: Could you please revert all changes in this file?

kant2002: @jkotas done

jkotas (approved):
Thank you!

Issue #988, opened 2021-04-21
Violated stack walking assertion when using debug ILCompiler build


AntonLapounov:
Create a console app containing just this line:
```C#
throw new System.Exception();
```
and compile it in release mode with a **debug** build of the ILCompiler package:
```cmd
dotnet publish -r win-x64 -c Release
```
Run it and notice the following violated assertion reported:
```
--------------------------------------------------
Debug Assertion Violation

Expression: 'pCurExInfo->m_idxCurClause != MaxTryRegionIdx'

File: src\coreclr\nativeaot\Runtime\StackFrameIterator.cpp, Line: 1850
--------------------------------------------------
```
with the following call stack:
```
Repro!Assert
Repro!RhpSfiNext
Repro!S_P_CoreLib_System_Runtime_RuntimeExports__RhpCalculateStackTraceWorker
Repro!RhpGetCurrentThreadStackTrace
Repro!S_P_CoreLib_System_Runtime_RuntimeExports__RhGetCurrentThreadStackTrace
Repro!S_P_CoreLib_System_Diagnostics_StackTrace__InitializeForCurrentThread
Repro!S_P_CoreLib_System_Diagnostics_StackTrace___ctor_3
Repro!S_P_CoreLib_System_Diagnostics_DebugProvider__Fail
Repro!S_P_CoreLib_System_Diagnostics_Debug__Fail_0
Repro!S_P_CoreLib_System_Diagnostics_Debug__Assert_1
Repro!S_P_CoreLib_System_Diagnostics_Debug__Assert
Repro!S_P_TypeLoader_Internal_Runtime_TypeLoader_TypeLoaderEnvironment__TryGetStaticClassConstructionContext
Repro!S_P_Reflection_Execution_Internal_Reflection_Execution_ReflectionExecutionDomainCallbacksImplementation__TryGetStaticClassConstructionContext
Repro!S_P_CoreLib_System_Runtime_CompilerServices_RuntimeHelpers__RunClassConstructor
Repro!S_P_CoreLib_System_SR__InternalGetResourceString
Repro!S_P_CoreLib_System_SR__GetResourceString_0
Repro!S_P_CoreLib_System_Exception__get_Message
Repro!S_P_CoreLib_System_Exception__AppendExceptionStackFrame
Repro!S_P_CoreLib_System_Runtime_EH__AppendExceptionStackFrameViaClasslib
Repro!S_P_CoreLib_System_Runtime_EH__UpdateStackTrace
[...]

Issue #987, opened 2021-04-21
Infinite recursion in SR.GetResourceString when using debug ILCompiler build


AntonLapounov:
Create a console app containing just this line:
```C#
new System.SystemException();
```
and compile it in release mode with a **debug** build of the ILCompiler package:
```cmd
dotnet publish -r win-x64 -c Release
```
If you run it, the app goes into infinite recursion in the `SR.GetResourceString` method, eventually failing due to stack overflow (that takes a long time though):
```
...
Repro!S_P_CoreLib_System_Diagnostics_Debug__Fail_0
Repro!S_P_CoreLib_System_Diagnostics_Debug__Assert_1
Repro!S_P_CoreLib_System_Diagnostics_Debug__Assert
Repro!S_P_TypeLoader_Internal_Runtime_TypeLoader_TypeLoaderEnvironment__TryGetStaticClassConstructionContext
Repro!S_P_Reflection_Execution_Internal_Reflection_Execution_ReflectionExecutionDomainCallbacksImplementation__TryGetStaticClassConstructionContext
Repro!S_P_CoreLib_System_Runtime_CompilerServices_RuntimeHelpers__RunClassConstructor
Repro!S_P_CoreLib_System_SR__InternalGetResourceString
𝗥𝗲𝗽𝗿𝗼!𝗦_𝗣_𝗖𝗼𝗿𝗲𝗟𝗶𝗯_𝗦𝘆𝘀𝘁𝗲𝗺_𝗦𝗥__𝗚𝗲𝘁𝗥𝗲𝘀𝗼𝘂𝗿𝗰𝗲𝗦𝘁𝗿𝗶𝗻𝗴_𝟬
Repro!S_P_CoreLib_System_Diagnostics_StackFrame__AppendToStackTrace
Repro!S_P_CoreLib_System_Diagnostics_StackTrace__ToString
Repro!S_P_CoreLib_System_Diagnostics_StackTrace__ToString_1
Repro!S_P_CoreLib_System_Diagnostics_DebugProvider__Fail
Repro!S_P_CoreLib_System_Diagnostics_Debug__Fail_0
Repro!S_P_CoreLib_System_Diagnostics_Debug__Assert_1
Repro!S_P_CoreLib_System_Diagnostics_Debug__Assert
Repro!S_P_TypeLoader_Internal_Runtime_TypeLoader_TypeLoaderEnvironment__TryGetStaticClassConstructionContext
[...]

PR #1153, opened 2021-05-22
Fixed wrong assert when getting StackTrace

Miizukii:
- Fixed wrong assert when getting StackTrace with unhandled exception on the stack. #988 

- Improved StackTrace/Frame result by filtering internal frames


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreRT.cs
```diff
@@ -35,9 +34,34 @@ private void InitializeForCurrentThread(int skipFrames, bool needFileInfo)
             IntPtr[] stackTrace = new IntPtr[frameCount];
             int trueFrameCount = RuntimeImports.RhGetCurrentThreadStackTrace(stackTrace);
             Debug.Assert(trueFrameCount == frameCount);
+            skipFrames += CalculateFramesToSkip(stackTrace, trueFrameCount);
```
jkotas: Instead of this guess work, it would be better to have control about the exact number of frames to skip, adding NoInlining as necessary, similar to how we do it in `Environment.StackTrace` or how Mono does it: https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/mono/System.Private.CoreLib/src/System/Diagnostics/StackFrame.Mono.cs#L31 . I think it would be fine to switch regular CoreCLR to the same plan too.

src/coreclr/nativeaot/Runtime/StackFrameIterator.h
```diff
@@ -137,9 +137,11 @@ class StackFrameIterator
         // This is a state returned by Next() which indicates that we just unwound a reverse pinvoke method
         UnwoundReversePInvoke = 0x20,
 
+        IgnoreExCollide = 0x40,
```
jkotas: The ExCollide is not really ignored when this is set. It is respected by the stackwalker in general. It is just the one place that is asking for the colliding clause index that is a problem.

A better way to fix this may be to change arguments of `RhpSfiNext` to pointers, allow passing NULL, and avoid the problem with the colliding clause that way. What do you think?

Miizukii: updated

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreRT.cs
Miizukii: I wanted to do that at first but that would also mean adding NoInlining to the Constructors.

```
Debug win
   at System.Runtime.RuntimeExports.RhGetCurrentThreadStackTrace(IntPtr[]) + 0xa0
   at System.Diagnostics.StackTrace.InitializeForCurrentThread(Int32, Boolean) + 0x7b
   at System.Diagnostics.StackTrace..ctor(Int32, Boolean) + 0x95
   at BringUpTest.FilterWithStackTrace(Exception) + 0x3b
```
```
Checked win
at System.Diagnostics.StackTrace.InitializeForCurrentThread(Int32, Boolean) + 0x5c
   at BringUpTest.FilterWithStackTrace(Exception) + 0x28
```

src/coreclr/nativeaot/Runtime/StackFrameIterator.cpp
```diff
@@ -1845,20 +1845,27 @@ COOP_PINVOKE_HELPER(Boolean, RhpSfiNext, (StackFrameIterator* pThis, uint32_t* p
     if (isValid)
         pThis->CalculateCurrentMethodState();
 
-    if (pThis->m_dwFlags & StackFrameIterator::ExCollide)
+    if (puExCollideClauseIdx)
```
jkotas:
```suggestion
    if (puExCollideClauseIdx != NULL)
```

src/coreclr/nativeaot/Runtime/StackFrameIterator.cpp
```diff
@@ -1845,20 +1845,27 @@ COOP_PINVOKE_HELPER(Boolean, RhpSfiNext, (StackFrameIterator* pThis, uint32_t* p
...
+            pCurExInfo->m_kind = (ExKind)(pCurExInfo->m_kind | EK_SupersededFlag);
+        }
+        else
+        {
+            *puExCollideClauseIdx = MaxTryRegionIdx;
+        }
     }
-    else
+
+    if (pfUnwoundReversePInvoke)
```
jkotas:
```suggestion
    if (pfUnwoundReversePInvoke != NULL)
```
Nit: Follow the usual style in this codebase

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreRT.cs
jkotas: Yes, it is fine to add `NoInlining` to the constructors.

jkotas: (and refactor things as necessary to have known number of stackframes in all cases)

src/coreclr/nativeaot/Runtime/StackFrameIterator.cpp
Miizukii: fixed

src/libraries/System.Private.CoreLib/src/System/SR.cs
```diff
@@ -16,6 +16,7 @@ internal static partial class SR
         private static List<string>? _currentlyLoading;
         private static int _infinitelyRecursingCount;
         private static bool _resourceManagerInited;
+        private static bool _resourceManagerInitFailed;
```
Miizukii: quick test, will remove it after the tests have run

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreRT.cs
Miizukii: I would need to update the constructors in Runtime [libraries/System.Private.CoreLib/src/System/Diagnostics/StackTrace.cs](https://github.com/dotnet/runtime/blob/798949a6f8013fc457cecca1f291a930670cd913/src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackTrace.cs#L33), correct?

jkotas: Yep. It is fine to do the changes like this in this runtimelab branch. We will integrate them up to dotnet/runtime later.

jkotas:
> Removed ToString from the stackTrace since its failing for linux_musl

I am fixing the musl crash in #1156. You can keep the original version of the test.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreRT.cs
Miizukii: updated, still have one problem the Pinvoke gets inlined, if you can point me to the right location in the ILCompiler i would take a look (would be nice to always inline them since it confuses the visual studio debugger), otherwise i changed the guess work to only test for that one extra frame and remove it.

Miizukii:
I run into a infinite loop when i use the repro project to print a stackTrace with a checked win build,
TryGetStaticClassConstructionContext asserts for S_P_CoreLib_System_Resources_ResourceReader, which again tries to get resouceStrings.
Should i create a issue or is that more likely a local setup problem? Test with the same setup run fine.

repro:
build.cmd nativeaot+libs+nativeaot.packages -lc release -rc checked
run checked ilcompiler with @"path\artifacts\bin\repro\x64\checked\compile-with-Release-libs.rsp
run reproNative checked

```
var stack = new StackTrace(0, true);
Debug.WriteLine(stack);$
```
```
reproNative.exe!S_P_CoreLib_System_SR__InternalGetResourceString() Line 22
	at \src\libraries\System.Private.CoreLib\src\System\SR.cs(22)
reproNative.exe!S_P_CoreLib_System_SR__GetResourceString() Line 36
	at \src\libraries\Common\src\System\SR.cs(36)
reproNative.exe!S_P_CoreLib_System_Diagnostics_DebugProvider__WriteAssert() Line 36
	at \src\libraries\System.Private.CoreLib\src\System\Diagnostics\DebugProvider.cs(36)
reproNative.exe!S_P_CoreLib_System_Diagnostics_DebugProvider__Fail() Line 29
	at \src\libraries\System.Private.CoreLib\src\System\Diagnostics\DebugProvider.cs(29)
reproNative.exe!S_P_TypeLoader_Internal_Runtime_TypeLoader_TypeLoaderEnvironment__TryGetStaticClassConstructionContext() Line 463
	at \src\coreclr\nativeaot\System.Private.TypeLoader\src\Internal\Runtime\TypeLoader\TypeLoaderEnvironment.Metadata.cs(463)
reproNative.exe!S_P_Reflection_Execution_Internal_Reflection_Execution_ReflectionExecutionDomainCallbacksImplementation__TryGetStaticClassConstructionContext() Line 118
	at \src\coreclr\nativeaot\System.Private.Reflection.Execution\src\Internal\Reflection\Execution\ReflectionExecutionDomainCallbacksImplementation.cs(118)
reproNative.exe!S_P_CoreLib_System_Runtime_CompilerServices_RuntimeHelpers__RunClassConstructor() Line 32
	at \src\coreclr\nativeaot\System.Private.CoreLib\src\System\Runtime\CompilerServices\RuntimeHelpers.CoreRT.cs(32)
reproNative.exe!S_P_CoreLib_System_SR__InternalGetResourceString() Line 72
	at \src\libraries\System.Private.CoreLib\src\System\SR.cs(72)
reproNative.exe!S_P_CoreLib_System_SR__GetResourceString() Line 36
	at \src\libraries\Common\src\System\SR.cs(36)
reproNative.exe!S_P_CoreLib_System_Diagnostics_StackFrame__AppendToStackTrace() Line 153
	at \src\coreclr\nativeaot\System.Private.CoreLib\src\System\Diagnostics\StackFrame.CoreRT.cs(153)
reproNative.exe!S_P_CoreLib_System_Diagnostics_StackTrace__ToString() Line 116
	at \src\coreclr\nativeaot\System.Private.CoreLib\src\System\Diagnostics\StackTrace.CoreRT.cs(116)
```

MichalStrehovsky:
> I run into a infinite loop when i use the repro project to print a stackTrace with a checked win build,
TryGetStaticClassConstructionContext asserts for S_P_CoreLib_System_Resources_ResourceReader, which again tries to get resouceStrings.

It's #987. Haven't gotten around to it yet, sorry!

MichalStrehovsky:
Just comment out the assert locally if it's blocking. It's harmless.

Miizukii:
@MichalStrehovsky thanks, its not blocking anything and sorry for not searching first

src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackFrame.cs
```diff
@@ -57,6 +58,7 @@ private void InitMembers()
         /// <summary>
         /// Constructs a StackFrame corresponding to the active stack frame.
         /// </summary>
+        [MethodImplAttribute(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
```
jkotas, line 61: Do we need `NoOptimization` here? Other places in this PR (and also existing like `Environment.StackTrace`) just use `NoInlining`.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreRT.cs
```diff
@@ -26,16 +27,32 @@ internal StackTrace(IntPtr[] ipAddresses, int startIndex, int endIndex, bool nee
...
+
             int frameCount = -RuntimeImports.RhGetCurrentThreadStackTrace(Array.Empty<IntPtr>());
             Debug.Assert(frameCount >= 0);
             IntPtr[] stackTrace = new IntPtr[frameCount];
             int trueFrameCount = RuntimeImports.RhGetCurrentThreadStackTrace(stackTrace);
             Debug.Assert(trueFrameCount == frameCount);
-            InitializeForIpAddressArray(stackTrace, skipFrames, frameCount, needFileInfo);
+            if (trueFrameCount > 0 && IsRhGetCurrentThreadStackTraceFrame(stackTrace[0]))
+            {
+                skipFrames++;
```
jkotas: Can we do accurate accounting for `skipFrames` so that `IsRhGetCurrentThreadStackTraceFrame` is not necessary at all?

jkotas: Just read your other comment about this.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreRT.cs
jkotas: I think you should be able to make the PInvoke to be always non-inlined by marking the method with NoInlining (or maybe also NoOptimization). I do not think we have a good to unconditionally enforce optimizations for a method.

Alternatively, you can pass in address of something in the current stack frame into RhpCalculateStackTrace and skip all frames with stack pointer less than that address.

src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackFrame.cs
Miizukii: Without it the  `call InitializeForCurrentThread` can be/gets replaced with a ` jmp`.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.CoreRT.cs
Miizukii: NoInlining and NoOptimization had no effect, i dont think its considered either since the code is generated in place of the normal call.
Will pass in an address.

src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackFrame.cs
jkotas: Hmmm, we are missing the NoInlining check in `CorInfoImpl.*.cs` `canTailCall`.

jkotas (approved):
Looks great. Thank you!

PR #1154, opened 2021-05-23
Cache the ThrowInstanceBodyRemoved helper

MichalStrehovsky:
Improves wallclock time for compilation of a Hello World by ~3%. Looking up things by name is not exactly cheap and we do it a lot here.


jkotas (approved):
Nice!

PR #1157, opened 2021-05-24
fix link to design document

itn3000:
fix link of design document from "proposed" to "accepted"


kant2002:
Is this something which can be merged?

PR #1158, opened 2021-05-24
Add functionality to enumerate virtual methods only

MichalStrehovsky:
We were spending around 8% of time in GetMethods due to various virtual method resolution code. In .NET, not that many methods are virtual, so restricting the iteration to virtual methods can save significant amount of time. The type system didn't have a facility to do this, so I'm adding one.

I'm seeing 2980 ms -> 2740 ms wallclock time improvement when compiling a Hello World with this.


jkotas (approved):
Nice!

jkotas:
Could you please push this to dotnet/runtime too when you get a chance?

PR #1159, opened 2021-05-24
Cache last document seen when iterating sequence points

MichalStrehovsky:
For most methods, the document will be the same through the entire method body. This avoids creating fresh strings using StringBuilder in BlobHeap.GetDocumentName.


jkotas (approved):
Thanks!

PR #1165, opened 2021-05-25
Tailcall from unboxing+instantiating thunks

MichalStrehovsky:
They look weird in stack traces.


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilerTypeSystemContext.BoxedTypes.cs
```diff
@@ -447,6 +447,7 @@ public override MethodIL EmitIL()
 
                 // Call an instance method on the target valuetype that has a fake instantiation parameter
                 // in it's signature. This will be swapped by the actual instance method after codegen is done.
+                codeStream.Emit(ILOpcode.tail);
```
jkotas, line 450: This is ok for now, but if we ever get slow tailcalls implemented, this won't work well anymore.

jkotas:
> They look weird in stack traces.

Is it just in debug builds; or in release builds as well? I would expect that RyuJIT will optimize the call at the end of the method into tailcall when the optimizations are on.

MichalStrehovsky:
Closing in favor of #1317.

Issue #205, opened 2020-10-12
Support covariant returns


MichalStrehovsky:
- [ ] Support in compiler
- [ ] Support in reflection

PR #1169, opened 2021-05-26
Add smoke tests for covariant returns

MichalStrehovsky:
Contributes to #205.

I'll repurpose #205 to just track the disabled CoreCLR Pri0 tests. They're pretty much all failing because we don't throw `TypeLoadException` if the input IL metadata is bogus. I don't think we want to do the kind of validation.

It's a harder fix because we need to refactor the tests so that the useless parts (checking for TypeLoad) can be individually ignored. ILASM apparently supports `ifdef`, but the defines cannot be passed on the command line (wow).


jkotas (approved):
Thanks!

PR #1171, opened 2021-05-26
Refresh sources shared with illink

MichalStrehovsky:


src/tests/nativeaot/SmokeTests/Dataflow/Dataflow.cs
```diff
@@ -166,6 +184,9 @@ static class Type3WithPublicKept
         [RequiresNonPublicMethods(typeof(Type1WithNonPublicKept), AlsoNeeded = typeof(Type2WithAllKept), AndAlsoNeeded = typeof(Type3WithPublicKept))]
         public static void Run()
         {
+            // Make it so that the analysis believes the Run method needs metadata. We wouldn't look at the attributes otherwise.
+            typeof(TestAttributeDataflow).GetMethod(nameof(Run));
```
MichalStrehovsky, line 188: This is unrelated - it was just necessary to make the test pass with `--reflectedonly` (we wouldn't generate metadata or attributes on the Run method otherwise, which would also make us not run the dataflow analysis)

jkotas (approved):
Thank you!

PR #1178, opened 2021-05-27
Replace IndexOf with RegexBoyerMoore in DFA mode

OlliSaarikivi:
Instead of using `string.IndexOf` for finding matches for fixed prefix strings, this PR changes the DFA mode to re-use the existing RegexBoyerMoore class used by the backtracking engines. This addresses some performance outliers we had detected.

The prefix max length is now increased from 20 to RegexBoyerMoore.MaxLimit (50000).

This removes logic that avoided triggering case insensitive prefix matching by using a shorter prefix, which was added originally because IndexOf was significantly slower for case insensitive matching.

Also removes logic in the prefix generation that prevented "I" and "K" from getting included due to semantic differences with case insensitivity between IndexOf and the regex package.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
@@ -1650,30 +1650,6 @@ namespace System.Text.RegularExpressions.SRM
         /// </summary>
         internal string GetFixedPrefix(CharSetSolver css, out bool ignoreCase)
         {
-            var pref = GetFixedPrefix_(css, out ignoreCase);
-            int i = pref.IndexOf('I');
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=1178) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=1178) before we can accept your contribution.<br/><br/>:x: OlliSaarikivi [sign now](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=1178)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/runtimelab?pullRequest=1178) it.</sub>

OlliSaarikivi:
Created from wrong account, closing.

PR #1179, opened 2021-05-27
Replace IndexOf with RegexBoyerMoore in DFA mode

olsaarik:
Instead of using `string.IndexOf` for finding matches for fixed prefix strings, this PR changes the DFA mode to re-use the existing RegexBoyerMoore class used by the backtracking engines. This addresses some performance outliers we had detected.

The prefix max length is now increased from 20 to RegexBoyerMoore.MaxLimit (50000).

This removes logic that avoided triggering case insensitive prefix matching by using a shorter prefix, which was added originally because IndexOf was significantly slower for case insensitive matching.

Also removes logic in the prefix generation that prevented "I" and "K" from getting included due to semantic differences with case insensitivity between IndexOf and the regex package.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/runtimelab?pullRequest=1179) <br/>All CLA requirements met.

veanes (approved):

PR #1180, opened 2021-05-28
Merge from dotnet/runtime

jkotas:


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
```diff
@@ -34,13 +34,6 @@ public static IntPtr OffsetOf(Type t, string fieldName)
             return new IntPtr(RuntimeAugments.InteropCallbacks.GetStructFieldOffset(t.TypeHandle, fieldName));
         }
 
-        private static object PtrToStructureHelper(IntPtr ptr, Type structureType)
-        {
-            object boxedStruct = InteropExtensions.RuntimeNewObject(structureType.TypeHandle);
```
MichalStrehovsky, line 39: Oh, thanks for fixing the dataflow annotations. I didn't realize not running the ctor was a .NET Native bug when I was annotating it.

MichalStrehovsky:
I think the pinvoke test is failing because the assembly is marked as __BlockAllReflectionAttribute. We either need to kick the BlittableClass and NonBlittableClass into a different assembly, or find a suitable pre-existing type in the framework :(. Maybe we can disable the test and file an issue for later?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.CoreRT.cs
jkotas: Each CoreCLR, Mono and NativeAOT had different behavior. Deduping the implementation fixed that :-)

jkotas:
I have fixed up some of the tests to not depend on reflection and commented out the few that cannot be fixed up. Smoketests are not meant to be comprehensive tests anyway.

PR #1182, opened 2021-05-28
Add missing VARIANT marshalling

kant2002:
- support for null marshalling
- marshalling of enums to variant
- Marshalling of IConvertible

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
@@ -559,7 +559,7 @@ namespace Internal.Runtime.CompilerHelpers
             Marshal.DestroyStructure(address, o.GetType());
         }
 
-        internal static unsafe object? OleVariantToObject(IntPtr pSrcNativeVariant)
+        internal static unsafe object? VariantToObject(IntPtr pSrcNativeVariant)
         {
             if (pSrcNativeVariant == IntPtr.Zero)
             {
@@ -575,7 +575,7 @@ namespace Internal.Runtime.CompilerHelpers
 #endif
         }
 
-        internal static unsafe void ConvertObjectToOleVariant(object? obj, IntPtr pDstNativeVariant)
+        internal static unsafe void ConvertObjectToVariant(object? obj, IntPtr pDstNativeVariant)
         {
 #if TARGET_WINDOWS
 #pragma warning disable CA1416
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
@@ -559,6 +559,33 @@ namespace Internal.Runtime.CompilerHelpers
             Marshal.DestroyStructure(address, o.GetType());
         }
 
+        internal static unsafe object? OleVariantToObject(IntPtr pSrcNativeVariant)
+        {
+            if (pSrcNativeVariant == IntPtr.Zero)
+            {
+                return null;
+            }
+
+#if TARGET_WINDOWS
+#pragma warning disable CA1416
+            return Marshal.GetObjectForNativeVariant(pSrcNativeVariant);
+#pragma warning restore CA1416
+#else
+            throw new PlatformNotSupportedException(SR.PlatformNotSupported_ComInterop);
+#endif
+        }
+
+        internal static unsafe void ConvertObjectToOleVariant(object? obj, IntPtr pDstNativeVariant)
+        {
+#if TARGET_WINDOWS
+#pragma warning disable CA1416
+            Marshal.GetNativeVariantForObject(obj, pDstNativeVariant);
+#pragma warning restore CA1416
+#else
+            throw new PlatformNotSupportedException(SR.PlatformNotSupported_ComInterop);
+#endif
+        }
+
         internal static unsafe void CleanupVariant(IntPtr pDstNativeVariant)
         {
 #if TARGET_WINDOWS
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
@@ -208,7 +208,7 @@ namespace System.Runtime.InteropServices
                     data->AsError = DISP_E_PARAMNOTFOUND;
                     break;
                 case IConvertible value:
-                    switch (value->GetTypeCode())
+                    switch (value.GetTypeCode())
                     {
                         case TypeCode.Empty:
                             data->VariantType = VarEnum.VT_EMPTY;
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
@@ -276,7 +276,7 @@ namespace System.Runtime.InteropServices
                     // SAFEARRAY implementation goes here.
                     throw new NotSupportedException();
                 case ValueType:
-                    throw new NotSupportedException();
+                    throw new NotSupportedException("VT_RECORD");
                 default:
                     data->AsDispatch = obj;
                     break;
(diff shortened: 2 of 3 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -559,6 +559,33 @@ internal static void AsAnyCleanupNative(IntPtr address, object o)
             Marshal.DestroyStructure(address, o.GetType());
         }
 
+        internal static unsafe object? OleVariantToObject(IntPtr pSrcNativeVariant)
```
jkotas: Nit: I would call this `Variant` like what it is called in most other places in .NET.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
```diff
@@ -207,6 +207,9 @@ public static unsafe void GetNativeVariantForObject(object? obj, IntPtr pDstNati
                 case System.Reflection.Missing value:
                     data->AsError = DISP_E_PARAMNOTFOUND;
                     break;
+                case Enum value:
+                    data->AsI8 = (long)Convert.ToInt32(value);
```
jkotas: Enums are marshalled as their underlying type. For example:
```csharp
using System;

public enum MyEnum : ushort
{
}

[StructLayout(LayoutKind.Sequential)]
internal struct NativeVariant
{
    public ushort vt;
    public ushort wReserved1;
    public ushort wReserved2;
    public ushort wReserved3;

    public IntPtr data1;
    public IntPtr data2;

}  // struct NativeVariant


class Program
{
    static unsafe void Main(string[] args)
    {
        NativeVariant v = default;
        System.Runtime.InteropServices.Marshal.GetNativeVariantForObject((object)default(MyEnum), (IntPtr)(void*)&v);
        Console.WriteLine(v.vt);

    }
}
```
This prints 18 `VT_UI2` on CoreCLR, but I think it will print 20 `VT_I8` with this change.

jkotas: Another case that this does not support currently are the conversions using `IConvertible`. For example:
```csharp
using System;

public class MyConvertible : IConvertible
{
    public TypeCode GetTypeCode()
    {
        return TypeCode.Int32;
    }

    public bool ToBoolean(IFormatProvider provider)
    {
    }

    public byte ToByte(IFormatProvider provider)
    {
    }

    public char ToChar(IFormatProvider provider)
    {
    }

    public DateTime ToDateTime(IFormatProvider provider)
    {
    }

    public decimal ToDecimal(IFormatProvider provider)
    {
    }

    public double ToDouble(IFormatProvider provider)
    {
    }

    public short ToInt16(IFormatProvider provider)
    {
    }

    public int ToInt32(IFormatProvider provider)
    {
        return 42;
    }

    public long ToInt64(IFormatProvider provider)
    {
    }

    public sbyte ToSByte(IFormatProvider provider)
    {
    }

    public float ToSingle(IFormatProvider provider)
    {
    }

    public string ToString(IFormatProvider provider)
    {
    }

    public object ToType(Type conversionType, IFormatProvider provider)
    {
    }

    public ushort ToUInt16(IFormatProvider provider)
    {
    }

    public uint ToUInt32(IFormatProvider provider)
    {
    }

    public ulong ToUInt64(IFormatProvider provider)
    {
    }
}

[StructLayout(LayoutKind.Sequential)]
internal struct NativeVariant
{
    public ushort vt;
    public ushort wReserved1;
    public ushort wReserved2;
    public ushort wReserved3;

    public IntPtr data1;
    public IntPtr data2;

}  // struct NativeVariant


class Program
{
    static unsafe void Main(string[] args)
    {
        NativeVariant v = default;
        System.Runtime.InteropServices.Marshal.GetNativeVariantForObject(new MyConvertible(), (IntPtr)(void*)&v);
        Console.WriteLine(v.vt);
    }
}
```
This will print 3 /* VT_I4 */ on CoreCLR. Note that enums implement `IConvertible`, so if you implement the more general IConvertible case, enums should just fall out naturally.

jkotas:
While you are on it, you can also flip the Type checks inside `default:` to just be cases in the type switch statement:
```
case Array:
...
case ValueType:
...
default:
...
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
kant2002: Question. Does this warrant having tests somewhere? I test my first implementation on tests from Interop/PInvoke/Variant and nothing was failed. So I naively assume that everything was ok. I would go ahead and implement them, but I'm not sure that you care about this kind of behaviour.

jkotas: `Marshal.GetObjectForNativeVariant` is API. API tests live under src/libraries. The tests for this one are in https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.InteropServices/tests/System/Runtime/InteropServices/Marshal/GetNativeVariantForObjectTests.cs. I can see that it includes the tests for enums.

jkotas: Another easy case to handle is throwing ArgumentException for CriticalHandle and SafeHandle

kant2002:
It looks like will-o'-the-wisp leading me to WinForms support.....

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
```diff
@@ -207,21 +207,78 @@ public static unsafe void GetNativeVariantForObject(object? obj, IntPtr pDstNati
...
+                    break;
+                case CriticalHandle:
+                    throw new ArgumentException();
+                case SafeHandle:
+                    throw new ArgumentException();
+                case Array:
+                    // SAFEARRAY implementation goes here.
+                    throw new NotSupportedException();
+                case ValueType:
+                    throw new NotSupportedException();
```
jkotas:
```suggestion
                    throw new NotSupportedException("VT_RECORD");
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
```diff
@@ -207,21 +207,78 @@ public static unsafe void GetNativeVariantForObject(object? obj, IntPtr pDstNati
...
+                            throw new NotSupportedException();
+                    } 
+                    break;
+                case CriticalHandle:
+                    throw new ArgumentException();
+                case SafeHandle:
+                    throw new ArgumentException();
+                case Array:
+                    // SAFEARRAY implementation goes here.
+                    throw new NotSupportedException();
```
jkotas:
```suggestion
                    // SAFEARRAY implementation goes here.
                    throw new NotSupportedException("VT_ARRAY");
```
to make this easier to diagnose once somebody hit this

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
```diff
@@ -207,21 +207,78 @@ public static unsafe void GetNativeVariantForObject(object? obj, IntPtr pDstNati
                 case System.Reflection.Missing value:
```
jkotas: You can change the `VariantWrapper value` can above to throw ArgumentException too. (It throws ArgumentException on CoreCLR.)

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
```diff
@@ -207,21 +207,78 @@ public static unsafe void GetNativeVariantForObject(object? obj, IntPtr pDstNati
...
+                    throw new ArgumentException();
+                case SafeHandle:
+                    throw new ArgumentException();
+                case Array:
+                    // SAFEARRAY implementation goes here.
+                    throw new NotSupportedException();
+                case ValueType:
+                    throw new NotSupportedException();
+                default:
+                    data->AsDispatch = obj;
```
jkotas, line 280: I suspect that you will pretty soon hit the case that needs to QueryInterface for IDispatch and fallback to IUnknown if the object does not implement IDispatch.

(I know that this is pre-existing issue.)

jkotas (commented):
LGTM otherwise. Thank you!

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Com.cs
kant2002: For now simple WinForms working with this PR. Next thing would be `PreserveSig = false`. Until I play with WebBrowser, hopefully I would not care about IDispatch at all.

jkotas (approved):
Thanks!

PR #1191, opened 2021-06-01
Fix stresslog deadlock

jkotas:
Stress log cannot allocate new memory blocks when the current thread has other threads suspended. Other thread may get suspended with lock taken inside memory allocator and thus attempt to allocate a new memory block on current thread may lead to deadlock.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/thread.h
+++ b/src/coreclr/nativeaot/Runtime/thread.h
@@ -87,6 +87,7 @@ struct ThreadBuffer
     uint64_t                  m_uPalThreadIdForLogging;               // @TODO: likely debug-only
     EEThreadId              m_threadId;
     PTR_VOID                m_pThreadStressLog;                     // pointer to head of thread's StressLogChunks
+    uint32_t                m_canntAlloc;
 #ifdef FEATURE_GC_STRESS
     uint32_t                  m_uRand;                                // current per-thread random number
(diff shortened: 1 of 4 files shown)


src/coreclr/nativeaot/Runtime/thread.h
```diff
@@ -210,6 +210,9 @@ class Thread : private ThreadBuffer
 #ifndef DACCESS_COMPILE
     void                SetThreadStressLog(void * ptsl);
 #endif // DACCESS_COMPILE
+    void                EnterCantAllocRegion();
+    void                LeaveCanntAllocRegion();
```
MichalStrehovsky:
```suggestion
    void                LeaveCantAllocRegion();
```

Also in the name of the field.

Wondering if we should just name it NoAllocRegion?

MichalStrehovsky (approved):

src/coreclr/nativeaot/Runtime/thread.h
jkotas: I forgot to mention that this is copy of CoreCLR logic to deal with the same issue. It is called CantAlloc in CoreCLR. I tried to keep as similar as possible to make eventual unification easier.

jkotas: Thanks for catching the typo!

Issue #1187, opened 2021-05-29
ILC crash on DataGridView


kant2002:
After adding `DataGridView` to simple WinForms application

```
Unhandled exception. System.InvalidOperationException : Internal linker error : processing of call from System.Windows.Forms.DataGridView.BeginEditInternal(Boolean) to System.Type.GetInterface(String) returned value which is not correctly annotated with the expected dynamic member access kinds. [c:\CoreRTWinFormsTestBed\samples\SampleWindowsForms\SampleWindowsForms.csproj]
     at ILCompiler.Dataflow.ReflectionMethodBodyScanner.HandleCall(MethodIL callingMethodBody, MethodDesc calledMethod, ILOpcode operation, Int32 offset, ValueNodeList methodParams, ValueNode& methodReturnValue)
     at ILCompiler.Dataflow.MethodBodyScanner.HandleCall(MethodIL callingMethodBody, ILOpcode opcode, Int32 offset, MethodDesc calledMethod, Stack`1 currentStack, Int32 curBasicBlock)
     at ILCompiler.Dataflow.MethodBodyScanner.Scan(MethodIL methodBody)
     at ILCompiler.Dataflow.ReflectionMethodBodyScanner.ScanAndProcessReturnValue(NodeFactory factory, FlowAnnotations flowAnnotations, Logger logger, MethodIL methodBody)
     at ILCompiler.DependencyAnalysis.DataflowAnalyzedMethodNode.GetStaticDependencies(NodeFactory factory)
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.GetStaticDependenciesImpl(DependencyNodeCore`1 node)
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.GetStaticDependencies(DependencyNodeCore`1 node)
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.ProcessMarkStack()
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.ComputeMarkedNodes()
     at ILCompiler.RyuJitCompilation.CompileInternal(String outputFile, ObjectDumper dumper)
     at ILCompiler.Compilation.ILCompiler.ICompilation.Compile(String outputFile, ObjectDumper dumper)
     at ILCompiler.Program.Run(String[] args)
     at ILCompiler.Program.Main(String[] args)
```

methodReturnValue variable has type `ILCompiler.Dataflow.MergePointValue`

[...]

PR #1193, opened 2021-06-01
Handle different dynamic member access for multiple branches

kant2002:
This case assume that all code branches should have same minimal dynamic access pattern

See #1187


MichalStrehovsky:
Thank you for looking into this - this looks like an actual bug in the codebase we share with IL Linker:

https://github.com/mono/linker/blob/54f4827313384e43bfd15c7ec6a45b6fa76090c3/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L1779-L1790

I would prefer we check that into IL Linker in case the fix end up looking differently and then just port it over when I refresh the IL Linker source next time.

Would you be up for submitting a pull request to the IL linker repo? They'll likely ask you for a unit test. I think we can avoid duplicating the existing if checks - `UniqueValues()` can be called on things that are not merge point too and it just returns itself as a single member of the enumeration.

If you're not up for doing the linker pull request, I'll take care of it later this week.

kant2002:
Filed https://github.com/mono/linker/pull/2076

Issue #1136, opened 2021-05-18
Implement `EnumComparer<T>`


MichalStrehovsky:
See NotImplementedException in:

src\coreclr\nativeaot\System.Private.CoreLib\src\System\Collections\Generic\Comparer.CoreRT.cs

We'll also need to update:

src\coreclr\tools\Common\TypeSystem\IL\Stubs\ComparerIntrinsics.cs
src\coreclr\nativeaot\System.Private.CoreLib\src\Internal\IntrinsicSupport\ComparerHelpers.cs

This should pretty much just follow the same path set by EnumEqualityComparer.

MichalStrehovsky:
This is implemented but we still have a failing test because of private reflection.

PR #1194, opened 2021-06-01
Added implementation for EnumComparer<T>

Miizukii:
see #1136

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Collections/Generic/Comparer.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Collections/Generic/Comparer.CoreRT.cs
@@ -48,8 +48,7 @@ namespace System.Collections.Generic
     {
         public override int Compare(T x, T y)
         {
-            // CORERT-TODO: EnumComparer<T>
-            throw new NotImplementedException();
+            return ComparerHelpers.EnumOnlyCompare(x, y);
         }
     }
 }
--- a/src/coreclr/tools/Common/TypeSystem/IL/CoreRTILProvider.cs
+++ b/src/coreclr/tools/Common/TypeSystem/IL/CoreRTILProvider.cs
@@ -154,6 +154,57 @@ namespace Internal.IL
                             return ComparerIntrinsics.EmitEqualityComparerCreate(method);
                     }
                     break;
+                case "ComparerHelpers":
+                    {
+                        if (owningType.Namespace != "Internal.IntrinsicSupport")
+                            return null;
+
+                        if (methodName == "EnumOnlyCompare")
+                        {
+                            //calls CompareTo for underlyingType to avoid boxing
+
+                            TypeDesc elementType = method.Instantiation[0];
+                            if (!elementType.IsEnum)
+                                return null;
+
+                            ILOpcode convInstruction;
+                            int typeSize = ((DefType)elementType).InstanceFieldSize.AsInt;
+                            if (typeSize <= 4)
+                            {
+                                convInstruction = ILOpcode.conv_i4;
+                            }
+                            else
+                            {
+                                Debug.Assert(typeSize == 8);
+                                convInstruction = ILOpcode.conv_i8;
+                            }
+
+                            TypeDesc underlyingType = elementType.UnderlyingType;
+                            TypeDesc returnType = method.Context.GetWellKnownType(WellKnownType.Int32);
+                            MethodDesc underlyingCompareToMethode = underlyingType.GetKnownMethod("CompareTo",
+                                new MethodSignature(
+                                    MethodSignatureFlags.None,
+                                    genericParameterCount: 0,
+                                    returnType: returnType,
+                                    parameters: new TypeDesc[] {underlyingType}));
+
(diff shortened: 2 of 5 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Collections/Generic/Comparer.CoreRT.cs
```diff
@@ -48,8 +48,7 @@ internal sealed partial class EnumComparer<T> : Comparer<T> where T : struct, En
     {
```
Miizukii, line 48: Do i need to do something special for `EnumComparer<T>` to generate `RuntimeConstructedGenericTypeInfo` and not `RuntimeBlockedTypeInfo`?

https://github.com/dotnet/runtimelab/blob/e5ab2d33f9d64e2ed407391c1b5b724f5c071f04/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/BlockedInternalsBlockingPolicy.cs#L124-L128

Gets blocked there since its internal and has nothing else that keeps it from being blocked, but quick search didn't show any special handling in CoreCLR

jkotas: I do not think you need to do anything special to make `EnumComparer` reflectable. It is supposed to be non-reflectable in NativeAOT.

src/coreclr/tools/Common/TypeSystem/IL/CoreRTILProvider.cs
```diff
@@ -154,6 +154,57 @@ private MethodIL TryGetPerInstantiationIntrinsicMethodIL(MethodDesc method)
...
+
+                            TypeDesc elementType = method.Instantiation[0];
+                            if (!elementType.IsEnum)
+                                return null;
+
+                            ILOpcode convInstruction;
+                            int typeSize = ((DefType)elementType).InstanceFieldSize.AsInt;
+                            if (typeSize <= 4)
+                            {
+                                convInstruction = ILOpcode.conv_i4;
```
jkotas: This should not need any conversions. It should just load the two arguments and call the underlying type CompareTo like: `ldarga 0, ldarg 1, call ..., ret`.

src/coreclr/tools/Common/TypeSystem/IL/CoreRTILProvider.cs
```diff
@@ -154,6 +154,57 @@ private MethodIL TryGetPerInstantiationIntrinsicMethodIL(MethodDesc method)
...
+                            }
+                            else
+                            {
+                                Debug.Assert(typeSize == 8);
+                                convInstruction = ILOpcode.conv_i8;
+                            }
+
+                            TypeDesc underlyingType = elementType.UnderlyingType;
+                            TypeDesc returnType = method.Context.GetWellKnownType(WellKnownType.Int32);
+                            MethodDesc underlyingCompareToMethode = underlyingType.GetKnownMethod("CompareTo",
```
jkotas:
```suggestion
                            MethodDesc underlyingCompareToMethod = underlyingType.GetKnownMethod("CompareTo",
```
Nit

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Collections/Generic/Comparer.CoreRT.cs
Miizukii: While that fine with me, it would mean that the test would still fail since it tries to compare the type
https://github.com/dotnet/runtimelab/blob/e5ab2d33f9d64e2ed407391c1b5b724f5c071f04/src/tests/JIT/opt/Devirtualization/Comparer_get_Default.cs#L221

jkotas: We can fixup the test. @MichalStrehovsky Do you have thoughts on the best way to do it?

jkotas (approved):
LGTM. Thank you!

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Collections/Generic/Comparer.CoreRT.cs
MichalStrehovsky: > We can fixup the test. @MichalStrehovsky Do you have thoughts on the best way to do it?

Nothing good comes to mind. The point of `EnumComparer` is that it doesn't allocate, so maybe we could test that calling `CompareTo` on it will not allocate?

jkotas: > point of EnumComparer is that it doesn't allocate, so maybe we could test that calling CompareTo on it will not allocate?

We typically cover this in performance tests, not in functional tests.

After thinking about it, I am leaning towards suppressing the tests when the type name is not available, e.g. only do this when ToString starts with `System.`.

jkotas: It can be done in separate PR.

PR #1197, opened 2021-06-02
Implement collection marshaller spec

jkoritzinsky:
Implement the spec in SpanMarshallers.md for collections.

Move arrays over to using the collection marshallers internally.

Refactor attribute inspection to more easily support holistic analysis of marshalling attributes (see what's used, what isn't, and what provides conflicting info).

This PR does not add built-in support for marshalling spans. That's planned for a future PR (though the implementation will likely follow the `ArrayMarshaller` in `Ancillary.Interop` very closely).


DllImportGenerator/DllImportGenerator/ManualTypeMarshallingHelper.cs
```diff
@@ -23,17 +27,34 @@ public static bool HasToManagedMethod(ITypeSymbol nativeType, ITypeSymbol manage
                         && !m.IsStatic);
         }
 
-        public static bool IsManagedToNativeConstructor(IMethodSymbol ctor, ITypeSymbol managedType)
+        public static bool IsManagedToNativeConstructor(
+            IMethodSymbol ctor,
+            ITypeSymbol managedType,
+            bool isCollectionMarshaller)
```
AaronRobinsonMSFT: Does an enum make more sense here? If not, I would prefer if we named the parameters during a call. It could be a const local rather than a named parameter if you like.

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
```diff
@@ -94,18 +94,14 @@ internal sealed class StubCodeGenerator : StubCodeContext
 
         public override (string managed, string native) GetIdentifiers(TypePositionInfo info)
         {
-            if (info.IsManagedReturnPosition && !info.IsNativeReturnPosition)
+            if (info.IsManagedReturnPosition)
```
AaronRobinsonMSFT, line 99: This is now doing double duty - similar to the switch statement from a previous review. If we are in a return position the native return is  by definition required right? Let's clarify that logic and reasoning.

DllImportGenerator/TestAssets/SharedTypes/NonBlittable.cs
```diff
@@ -217,4 +219,114 @@ public IntStructWrapperNative(IntStructWrapper managed)
...
+            if (managed is null)
+            {
+                managedList = null;
+                NativeValueStorage = default;
+                return;
+            }
+            managedList = managed;
+            // Always allocate at least one byte when the array is zero-length.
+            int spaceToAllocate = Math.Max(managed.Count * sizeOfNativeElement, 1);
+            if (spaceToAllocate < stackSpace.Length)
```
AaronRobinsonMSFT: `<=`?

DllImportGenerator/DllImportGenerator/DllImportStub.cs
```diff
@@ -157,12 +157,14 @@ public class GeneratedDllImportData
...
 
+            var marshallingAttributeParser = new MarshallingAttributeInfoParser(env.Compilation, diagnostics, defaultInfo, method);
+
             // Determine parameter and return types
             var paramsTypeInfo = new List<TypePositionInfo>();
             for (int i = 0; i < method.Parameters.Length; i++)
             {
                 var param = method.Parameters[i];
-                var typeInfo = TypePositionInfo.CreateForParameter(param, defaultInfo, env.Compilation, diagnostics, method.ContainingType);
+                var typeInfo = TypePositionInfo.CreateForParameter(param, marshallingAttributeParser.ParseMarshallingInfo(param.Type, param.GetAttributes()), env.Compilation);
```
AaronRobinsonMSFT: Can we make a local variable here. Some of this inline logic makes stepping/debugging annoying and reviewing more difficult.

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
```diff
@@ -105,9 +122,616 @@ internal enum SupportedMarshallingMethods
...
+        {
+            this.compilation = compilation;
+            this.diagnostics = diagnostics;
+            this.defaultInfo = defaultInfo;
+            this.contextSymbol = contextSymbol;
+            marshalAsAttribute = compilation.GetTypeByMetadataName(TypeNames.System_Runtime_InteropServices_MarshalAsAttribute)!;
+            marshalUsingAttribute = compilation.GetTypeByMetadataName(TypeNames.MarshalUsingAttribute)!;
+        }
+
+        internal MarshallingInfo ParseMarshallingInfo(
```
AaronRobinsonMSFT: Why `internal` vs `public`? I'm trying to follow the scoping to indicate how we are exposing types and that is really hard in this file. The type is `private` so why not `public`?

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
```diff
@@ -105,9 +122,616 @@ internal enum SupportedMarshallingMethods
...
+            int maxIndirectionLevelUsed = 0;
+            MarshallingInfo info = GetMarshallingInfo(
+                managedType,
+                marshallingAttributesByIndirectionLevel,
+                indirectionLevel: 0,
+                inspectedElements,
+                ref maxIndirectionLevelUsed);
+            if (maxIndirectionLevelUsed < maxIndirectionLevelDataProvided)
+            {
+                diagnostics.ReportConfigurationNotSupported(marshallingAttributesByIndirectionLevel[maxIndirectionLevelDataProvided], "ElementIndirectionLevel", maxIndirectionLevelDataProvided.ToString());
```
AaronRobinsonMSFT: Should this be a `nameof(ElementIndirectionLevel)` somehow?

DllImportGenerator/Ancillary.Interop/GeneratedMarshallingAttribute.cs
```diff
@@ -22,14 +22,36 @@ public NativeMarshallingAttribute(Type nativeType)
         public Type NativeType { get; }
     }
 
-    [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.ReturnValue | AttributeTargets.Field)]
+    [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.ReturnValue | AttributeTargets.Field, AllowMultiple = true)]
     public class MarshalUsingAttribute : Attribute
     {
+        public MarshalUsingAttribute()
+        {
+            CountElementName = null!;
```
AaronRobinsonMSFT: Should we change this to `string.Empty`?

DllImportGenerator/Ancillary.Interop/GeneratedMarshallingAttribute.cs
```diff
@@ -22,14 +22,36 @@ public NativeMarshallingAttribute(Type nativeType)
         public Type NativeType { get; }
     }
 
-    [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.ReturnValue | AttributeTargets.Field)]
+    [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.ReturnValue | AttributeTargets.Field, AllowMultiple = true)]
     public class MarshalUsingAttribute : Attribute
```
AaronRobinsonMSFT: `sealed`?

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaller.cs
```diff
@@ -10,7 +10,6 @@ namespace Microsoft.Interop
 {
     class CustomNativeTypeMarshaller : IMarshallingGenerator
```
AaronRobinsonMSFT: Since we are inheriting from this, would it be possible to make this `abstract` and then have a concrete `sealed` type we can use? The inheritance can get out of control here and I don't want to wind up where we are with the built-in system.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -250,7 +250,7 @@ private static IMarshallingGenerator ValidateByValueMarshalKind(StubCodeContext
                 // Must go before the cases that do not explicitly check for marshalling info to support
                 // the user overridding the default marshalling rules with a MarshalUsing attribute.
                 case { MarshallingAttributeInfo: NativeMarshallingAttributeInfo marshalInfo }:
-                    return CreateCustomNativeTypeMarshaller(info, context, marshalInfo);
+                    return CreateCustomNativeTypeMarshaller(info, context, marshalInfo, options);
```
AaronRobinsonMSFT, line 253: This sequence of arguments reminds me of the built-in system. Can these options be paired to context or something? This is a minor issue relative to other things but it seems we are simply using synonyms for "data". Perhaps an opportunity to collapse and reconcile things?

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaller.cs
AaronRobinsonMSFT: Feel free to push back with this. When I see many new `virtual`s with default behavior I am curious is the inheritance is needed or if there is another model we can apply.

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
jkoritzinsky: We don't have a nameof for it in the generator, but there is a constant I added that I can use here.

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaller.cs
jkoritzinsky: I'll look into re-architecting this a little bit. I agree that the design could be cleaner and I fully understand the desire to limit the inheritance tree or have all classes in the tree either abstract or sealed.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
jkoritzinsky: We could pair it, but it would feel a little odd. I think a better model would be to use the design in #1119 where the method with this giant switch is in a non-static class, and the options are (or could be) passed into the object's constructor instead of passed down as a parameter.

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
AaronRobinsonMSFT: Thanks.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
AaronRobinsonMSFT: Okay. Let's focus on the extensibility PR then to address this.

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
jkoritzinsky: I don't feel like this is doing double duty. The two cases I collapsed here return the same values. I'll add some comments to explain the logic here as I understand it.

DllImportGenerator/DllImportGenerator/ManualTypeMarshallingHelper.cs
jkoritzinsky: For future proofing, it might be worth making this an enum. That way if we ever introduce another marshaller model extension, we can easily extend these APIs cleanly.

DllImportGenerator/DllImportGenerator/Marshalling/CustomNativeTypeMarshaller.cs
jkoritzinsky: I've rewritten this quite drastically to use a combination of the Strategy and Decorator patterns. Let me know what you think of the new design.

DllImportGenerator/DllImportGenerator/ManualTypeMarshallingHelper.cs
```diff
@@ -11,6 +12,22 @@ static class ManualTypeMarshallingHelper
         public const string StackBufferSizeFieldName = "StackBufferSize";
         public const string ToManagedMethodName = "ToManaged";
         public const string FreeNativeMethodName = "FreeNative";
+        public const string ManagedValuesPropertyName = "ManagedValues";
+        public const string NativeValueStoragePropertyName = "NativeValueStorage";
+        public const string SetUnmarshalledCollectionLengthMethodName = "SetUnmarshalledCollectionLength";
+
+        public static class MarshalUsingProperties
+        {
+            public const string ElementIndirectionLevel = "ElementIndirectionLevel";
```
AaronRobinsonMSFT: I _think_ we can do the following &ndash; up to you.

```csharp
public const string ElementIndirectionLevel = nameof(ElementIndirectionLevel);
```

DllImportGenerator/DllImportGenerator/ManualTypeMarshallingHelper.cs
```diff
@@ -62,8 +96,50 @@ public static bool HasFreeNativeMethod(ITypeSymbol type)
         {
             return type.GetMembers(FreeNativeMethodName)
                 .OfType<IMethodSymbol>()
-                .Any(m => m is { Parameters: { Length: 0 } } and
-                    ({ ReturnType: { SpecialType: SpecialType.System_Void } }));
+                .Any(m => m is { IsStatic: false, Parameters: { Length: 0 }, ReturnType: { SpecialType: SpecialType.System_Void } });
+        }
+
+        public static IPropertySymbol? FindManagedValuesProperty(ITypeSymbol type)
```
AaronRobinsonMSFT: Would it make more sense to convert this into a Try function? It would avoid the nullable in the signature. I don't know how impact this change would be.

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
```diff
@@ -105,9 +122,605 @@ internal enum SupportedMarshallingMethods
     /// </summary>
     internal sealed record SafeHandleMarshallingInfo(bool AccessibleDefaultConstructor) : MarshallingInfo;
 
-
     /// <summary>
-    /// Default marshalling for arrays
-    /// </summary>
-    internal sealed record ArrayMarshallingInfo(MarshallingInfo ElementMarshallingInfo) : MarshallingInfo;
+    /// User-applied System.Runtime.InteropServices.NativeMarshalllingAttribute
```
AaronRobinsonMSFT: ~~NativeMarshalllingAttribute~~ -> NativeMarshallingAttribute

DllImportGenerator/DllImportGenerator/MarshallingAttributeInfo.cs
```diff
@@ -105,9 +122,605 @@ internal enum SupportedMarshallingMethods
...
+
+                return new NativeContiguousCollectionMarshallingInfo(
+                    nativeType,
+                    valueProperty?.Type,
+                    methods,
+                    NativeTypePinnable: ManualTypeMarshallingHelper.FindGetPinnableReference(nativeType) is not null,
+                    UseDefaultMarshalling: !isMarshalUsingAttribute,
+                    parsedCountInfo,
+                    elementType,
+                    GetMarshallingInfo(elementType, useSiteAttributes, maxIndirectionLevelUsed = indirectionLevel + 1, inspectedElements, ref maxIndirectionLevelUsed));
```
AaronRobinsonMSFT: > `GetMarshallingInfo(elementType, useSiteAttributes, maxIndirectionLevelUsed = indirectionLevel + 1, inspectedElements, ref maxIndirectionLevelUsed)`

Not a fan of this at all. I am sure the evaluation order in C# is defined but it is weird to see how this works coming from C/C++. Let's avoid these sort of inline assignment and uses.

DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
```diff
@@ -0,0 +1,969 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Microsoft.Interop
+{
+    /// <summary>
+    /// The base interface for implementing various different aspects of the custom native type and collection marshalling specs.
+    /// </summary>
+
```
AaronRobinsonMSFT: Extra line.

AaronRobinsonMSFT (commented):
Overall, I like this a lot more. I am struggling to reconcile the nested collection logic's cost with the user benefit. I am thinking of the servicing burden of owning such a complex piece of logic for the very few users in practice. I wonder if there is some way we could excise that logic and ensure we have the capability for users do to it themselves if it becomes interesting.

jkoritzinsky:
This PR doesn't have any of the work to support nested collections. That's in a separate local branch I haven't pushed up. This PR just has the support for the collections spec as written. It doesn't support collection elements with complex marshallers (marshallers with a Value property). That's also later work.

jkoritzinsky:
Well, it does have support for the ElementIndirectionLevel validation, which is a prereq to nested collections support.

jkoritzinsky:
If we wanted to, we could just block any ElementIndirectionLevel values above 1. Though we'd still need the same validation for levels 0 and 1, so it wouldn't necessarily be much simpler.

AaronRobinsonMSFT:
>Well, it does have support for the ElementIndirectionLevel validation, which is a prereq to nested collections support.

Yes, that is the logic I was referring to.

>If we wanted to, we could just block any ElementIndirectionLevel values above 1. Though we'd still need the same validation for levels 0 and 1, so it wouldn't necessarily be much simpler.

Okay.

AaronRobinsonMSFT (approved):
LGTM.

PR #1200, opened 2021-06-02
Fix memory leaks for return values and some out params.

jkoritzinsky:
Fix some memory leaks we had where we didn't generate code to free native resources for out parameters that use custom native type marshalling or all return values.

Also set a flag in the csproj that enables a nice debugging innerloop option for roslyn components in the csproj settings (in our case running the generator on the integration tests)

### DIFF
--- a/DllImportGenerator/DllImportGenerator/DllImportGenerator.csproj
+++ b/DllImportGenerator/DllImportGenerator/DllImportGenerator.csproj
@@ -10,6 +10,7 @@
     <LangVersion>Preview</LangVersion>
     <Nullable>enable</Nullable>
     <RootNamespace>Microsoft.Interop</RootNamespace>
+    <IsRoslynComponent>true</IsRoslynComponent>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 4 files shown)


AaronRobinsonMSFT (commented):
How are we testing our clean up logic? I am fine adding this in a future PR. I can imagine using some of the `MarshalUsing` logic in some cases. Thoughts?

jkoritzinsky:
Currently, we aren't doing a good job of testing the cleanup logic (I found this when inspecting codegen from https://github.com/dotnet/runtimelab/pull/1197).

With a specially-designed managed/native type pair, we can set something up to explicitly test that we free when expected, at least for the custom native type marshalling (the MarshalUsing scenario). For the other marshallers like the built-in string marshalling, our only testing is manual spot checks. That's one of the reasons I'd like to try moving more of the marshallers to be based on the custom native type marshalling and not directly in the stub if possible. #1197 and some follow-up work I have in a branch in my fork move us more in that direction.

AaronRobinsonMSFT (approved):

PR #1202, opened 2021-06-04
NativeAOT: Typo in var name for 32 bit code

yowl:
This PR corrects a typo in the 32 bit version of the code.  I assume this is a ctrl c/v error that is not picked up as this is not yet merged into NativeAOT-LLVM.  I've not tested it as I'm still working on the merge, and I could put this change in that PR if you prefer.


jkotas (approved):
Thanks

PR #1206, opened 2021-06-07
Implement handling PreserveSig = false for interfaces

### LINKED ISSUE #1183: Unsupported PreserveSig = false combination prevent working with images in WinForms.
Any image related code crashed with following exception
```
System.Exception: Method '[System.Windows.Forms.Primitives]Interop+Ole32.OleCreatePictureIndirect(PICTDESC*,Guid&,BOOL)' requires marshalling that is not yet supported by this compiler.
   at Interop.Ole32.OleCreatePictureIndirect(Interop.Ole32.PICTDESC*, Guid&, Interop.BOOL) + 0x45
```

Lines which crash compiler:

https://github.com/dotnet/winforms/blob/d46ad2e8dc76248739d9ae22b28b399a6a6b299e/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleCreatePictureIndirect.cs#L12-L14

kant2002:
This was done by pretending return value is out argument passed as last value.
We still need to return that value, so this require change in the Marshaller.cs so it would not throw assertion, since previously only Return values can return value.
- For now only primitive types and Interface and Object are supported.
- I opt out of DelegateMarshallingMethodThunk because when these classes created with PreserveSig = false. I assume this is because of laziness during creation of this class because this flag was not used previously. I'm not sure should I optout of DelegateMarshallingMethodThunk or just set PreserveSig flag?
- I add special case for handling blittable types which are return type for PreserveSig=false methods. In this case I fallback to default marshalling implementation.

Closes #1183

### DIFF
--- a/src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
+++ b/src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
@@ -233,14 +233,6 @@ namespace Internal.IL.Stubs
 
         private void EmitPInvokeCall(PInvokeILCodeStreams ilCodeStreams)
         {
-            if (!_flags.PreserveSig)
-            {
-                TypeDesc managedReturnType = _targetMethod.Signature.ReturnType;
-                bool supportedType = managedReturnType.IsPrimitive || managedReturnType.IsInterface || managedReturnType.IsObject;
-                if (!supportedType)
-                    throw new NotSupportedException();
-            }
-
             ILEmitter emitter = ilCodeStreams.Emitter;
             ILCodeStream fnptrLoadStream = ilCodeStreams.FunctionPointerLoadStream;
             ILCodeStream callsiteSetupCodeStream = ilCodeStreams.CallsiteSetupCodeStream;
--- a/src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
+++ b/src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
@@ -97,7 +97,9 @@ namespace Internal.IL.Stubs
                 TypeDesc parameterType = (i == 0)
                     ? methodSig.ReturnType //first item is the return type
                     : methodSig[i - 1];
-                marshallers[i] = Marshaller.CreateMarshaller((i == 0 && !flags.PreserveSig && !parameterType.IsVoid) ? methodSig.Context.GetByRefType(parameterType): parameterType,
+                bool isHRSwapping = !flags.PreserveSig && !parameterType.IsVoid
+                    && targetMethod is not DelegateMarshallingMethodThunk;
+                marshallers[i] = Marshaller.CreateMarshaller((i == 0 && isHRSwapping) ? methodSig.Context.GetByRefType(parameterType): parameterType,
                                                     parameterIndex,
                                                     methodSig.GetEmbeddedSignatureData(),
                                                     MarshallerType.Argument,
@@ -108,8 +110,8 @@ namespace Internal.IL.Stubs
                                                     indexOffset + parameterMetadata.Index,
                                                     flags,
                                                     parameterMetadata.In,
-                                                    (i == 0 && !flags.PreserveSig && !parameterType.IsVoid) ? true : parameterMetadata.Out,
-                                                    (i == 0 && !flags.PreserveSig && !parameterType.IsVoid) ? false : parameterMetadata.Return
+                                                    (i == 0 && isHRSwapping) ? true : parameterMetadata.Out,
+                                                    (i == 0 && isHRSwapping) ? false : parameterMetadata.Return
                                                     );
             }
 
@@ -244,7 +246,8 @@ namespace Internal.IL.Stubs
             ILCodeStream callsiteSetupCodeStream = ilCodeStreams.CallsiteSetupCodeStream;
             TypeSystemContext context = _targetMethod.Context;
 
-            bool isHRSwapping = !_flags.PreserveSig && _targetMethod.Signature.ReturnType != _targetMethod.Context.GetWellKnownType(WellKnownType.Void);
+            bool isHRSwapping = !_flags.PreserveSig && _targetMethod.Signature.ReturnType != _targetMethod.Context.GetWellKnownType(WellKnownType.Void)
+                && _targetMethod is not DelegateMarshallingMethodThunk;
             TypeDesc nativeReturnType = _flags.PreserveSig ? _marshallers[0].NativeParameterType : context.GetWellKnownType(WellKnownType.Int32);
             TypeDesc[] nativeParameterTypes = new TypeDesc[isHRSwapping ? _marshallers.Length : _marshallers.Length - 1];
 
@@ -368,7 +371,8 @@ namespace Internal.IL.Stubs
             cleanupCodestream.BeginHandler(tryFinally);
 
             // Marshal the arguments
-            bool isHRSwapping = !_flags.PreserveSig && _targetMethod.Signature.ReturnType != _targetMethod.Context.GetWellKnownType(WellKnownType.Void);
+            bool isHRSwapping = !_flags.PreserveSig && _targetMethod.Signature.ReturnType != _targetMethod.Context.GetWellKnownType(WellKnownType.Void)
+                && _targetMethod is not DelegateMarshallingMethodThunk;
             for (int i = isHRSwapping ? 1 : 0; i < _marshallers.Length; i++)
             {
                 _marshallers[i].EmitMarshallingIL(pInvokeILCodeStreams);
--- a/src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
+++ b/src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
@@ -97,7 +97,7 @@ namespace Internal.IL.Stubs
                 TypeDesc parameterType = (i == 0)
                     ? methodSig.ReturnType //first item is the return type
                     : methodSig[i - 1];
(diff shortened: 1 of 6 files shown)


kant2002:
I opt out of DelegateMarshallingMethodThunk because when these classes created with `PreserveSig = false`. I assume this is because of laziness during creation of this class because this flag was not used previously. I'm not sure should I optout of DelegateMarshallingMethodThunk or just set PreserveSig flag?

kant2002:
I add special case for handling blittable types which are return type for PreserveSig=false methods. In this case I fallback to default marshalling implementation.

src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
```diff
@@ -231,16 +233,23 @@ private void EmitDelegateCall(DelegateMarshallingMethodThunk delegateMethod, PIn
 
         private void EmitPInvokeCall(PInvokeILCodeStreams ilCodeStreams)
         {
-            if (!_flags.PreserveSig && _targetMethod.Signature.ReturnType != _targetMethod.Context.GetWellKnownType(WellKnownType.Void))
-                throw new NotSupportedException();
+            if (!_flags.PreserveSig)
+            {
+                TypeDesc managedReturnType = _targetMethod.Signature.ReturnType;
+                bool supportedType = managedReturnType.IsPrimitive || managedReturnType.IsInterface || managedReturnType.IsObject;
+                if (!supportedType)
```
jkotas: What prevents the other cases from being supported? I would expect the implementation to work for all marshalers without special casing.

src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
```diff
@@ -231,16 +233,23 @@ private void EmitDelegateCall(DelegateMarshallingMethodThunk delegateMethod, PIn
 
         private void EmitPInvokeCall(PInvokeILCodeStreams ilCodeStreams)
         {
-            if (!_flags.PreserveSig && _targetMethod.Signature.ReturnType != _targetMethod.Context.GetWellKnownType(WellKnownType.Void))
-                throw new NotSupportedException();
+            if (!_flags.PreserveSig)
+            {
+                TypeDesc managedReturnType = _targetMethod.Signature.ReturnType;
+                bool supportedType = managedReturnType.IsPrimitive || managedReturnType.IsInterface || managedReturnType.IsObject;
```
jkotas: Is this going to break `void` return value type?

jkotas:
> I'm not sure should I optout of DelegateMarshallingMethodThunk or just set PreserveSig flag?

It would be nice to make it work the same way as in CoreCLR.

src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
kant2002: No, IsPrimitive handle that because it just look at Enum value here
https://github.com/dotnet/runtime/blob/0fcd7690a14335a8807144d0e7ad937aa83699b0/src/coreclr/tools/Common/TypeSystem/Common/TypeDesc.cs#L198
and 
https://github.com/dotnet/runtime/blob/0fcd7690a14335a8807144d0e7ad937aa83699b0/src/coreclr/tools/Common/TypeSystem/Common/TypeFlags.cs#L31

src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
kant2002: It require more testing. Nothing in principle, just my inability to see all supported cases.
Also https://github.com/dotnet/runtime/blob/1843cdef72e1e7c810c4ea85c677d679ff3b66f4/src/coreclr/vm/dllimport.cpp#L3560-L3564 indicate cases which are not supported, but that does not give me much information what is supported. I think I can add Enum and String without probably much problems if needed.

kant2002:
> It would be nice to make it work the same way as in CoreCLR.

Reverse marshalling I do not fully grasp yet. I do not yet hit case where I was involved enough to implement.

src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
jkotas: All types that can be marshalled should be supported. 

The unsupported cases in CoreCLR are just a legacy baggage from the times where the marshalling stubs were hand-emitted x86 assembly and handling them would require writing a pile of new hand-emitted x86 assembly. We should not need need to worry about artificially throwing exceptions for the cases that are unsupported in CoreCLR for no good reason. 

I think that the transformation for PreserveSig = false should be very regular, it should depend on the regular `[out]` marshalling. It should not depend on the actual type at all.


```
[PreserveSig = false]
RETVAL foo(...);
```
should be transformed to:
```
RETVAL foo(...)
{
   RETVAL retNative = default;

   marshal everything usual

   actual_call(..., &retNative);

   unmarshal everything as usual (including retNative / retManaged)

   return retManaged;
}
```

src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
```diff
@@ -97,7 +97,9 @@ private static Marshaller[] InitializeMarshallers(MethodDesc targetMethod, Inter
                 TypeDesc parameterType = (i == 0)
```
jkotas: `(i == 0 && isHRSwapping)` condition is repeated multiple times here. Also, I do not think you should need the check for `DelegateMarshallingMethodThunk` here. We should make sure to set `PreserveSig=true` for marshalled delegates earlier (`UnmanagedFunctionPointerAttribute` won't allow you to specify `PreserveSig=false` so we do not need to worry about handling it).

I think it would be a bit more efficient and readable like this: 
```
TypeDesc parameterType;
bool isHRSwappedRetVal = false;
if (i == 0)
{
    //first item is the return type
    parameterType = methodSig.ReturnType;
    if (!flags.PreserveSig && !parameterType.IsVoid)
    {
        // PreserveSig = false can only show up an regular forward PInvokes
        Debug.Assert(direction == MarshalDirection.Forward);
    
        parameterType = methodSig.Context.GetByRefType(parameterType);
        isHRSwappedRetVal = true;
    }
}
else
{
    parameterType = methodSig[i - 1];
}
```

src/coreclr/tools/Common/TypeSystem/IL/Stubs/PInvokeILEmitter.cs
```diff
@@ -231,16 +233,15 @@ private void EmitDelegateCall(DelegateMarshallingMethodThunk delegateMethod, PIn
...
         {
-            if (!_flags.PreserveSig && _targetMethod.Signature.ReturnType != _targetMethod.Context.GetWellKnownType(WellKnownType.Void))
-                throw new NotSupportedException();
-
             ILEmitter emitter = ilCodeStreams.Emitter;
             ILCodeStream fnptrLoadStream = ilCodeStreams.FunctionPointerLoadStream;
             ILCodeStream callsiteSetupCodeStream = ilCodeStreams.CallsiteSetupCodeStream;
             TypeSystemContext context = _targetMethod.Context;
 
+            bool isHRSwapping = !_flags.PreserveSig && _targetMethod.Signature.ReturnType != _targetMethod.Context.GetWellKnownType(WellKnownType.Void)
```
jkotas: `== targetMethod.Context.GetWellKnownType(WellKnownType.Void)` can be just `IsVoid`

jkotas:
It would be nice to add one test for this to smoke tests. I think this is tricky enough to have it covered in the smoke tests.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
```diff
@@ -512,7 +512,7 @@ protected virtual void EmitMarshalReturnValueManagedToNative()
 
         public virtual void LoadReturnValue(ILCodeStream codeStream)
         {
-            Debug.Assert(Return);
+            Debug.Assert(Return || Index == 0);
```
jkotas: It would be nice to replace all `Index == 0` checks with a property. It would make the code much easier to understand.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
```diff
@@ -653,9 +653,18 @@ protected void PropagateFromByRefArg(ILCodeStream stream, Home home)
...
+            {
+                // Returning result would be handled by LoadReturnValue
+                return;
+            }
+            else
+            {
+                stream.EmitLdArg(Index - 1);
+                home.LoadValue(stream);
+                stream.EmitStInd(ManagedType);
+            }
```
jkotas:
```suggestion

            stream.EmitLdArg(Index - 1);
            home.LoadValue(stream);
            stream.EmitStInd(ManagedType);
```
Nit

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.cs
```diff
@@ -903,7 +912,7 @@ class BlittableValueMarshaller : Marshaller
     {
         protected override void EmitMarshalArgumentManagedToNative()
         {
-            if (IsNativeByRef && MarshalDirection == MarshalDirection.Forward)
+            if (IsNativeByRef && MarshalDirection == MarshalDirection.Forward && Index > 0)
```
jkotas: I would move the special case first and then leave the rest of the method intact.
```
if (IsHRSwappedRetVal)
{
    ...;
    return;
}
```

jkotas (approved):
LGTM otherwise. Thank you!

PR #1209, opened 2021-06-07
Comment in source that source file is auto-generated.

AaronRobinsonMSFT:
Following guidance from https://github.com/dotnet/roslyn/issues/49326


jkoritzinsky (approved):

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -53,7 +53,12 @@ public void Execute(GeneratorExecutionContext context)
             }
 
             var env = new StubEnvironment(context.Compilation, isSupported, targetFrameworkVersion, context.AnalyzerConfigOptions.GlobalOptions);
+
             var generatedDllImports = new StringBuilder();
+
+            // Mark in source that the file is auto-generated.
+            generatedDllImports.AppendLine(Comment("// <auto-generated/>").ToString());
```
sharwell: 💭 This seems more complex than necessary. If you're just going to call `ToString()` might as well use this:

```suggestion
            generatedDllImports.AppendLine("// <auto-generated/>");
```

elinor-fung (approved):

PR #1216, opened 2021-06-09
Enable marshalling collections of types with marshallers that have a Value property

jkoritzinsky:
Combine the concepts of pinning and stackalloc because we don't guarantee that the Span we provide for the stack-alloc optimization is actually stack allocated (so the only safe way to access is to pin).

Change the concept of "additional temporary state supported" to "additional temporary state lives across stages" and update generators to either handle the new concept or reject working in a scenario where it is unsupported.

This PR can enable us to move more of our built-in marshallers (like our string marshallers) to use the custom type marshalling pattern under the hood.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
+++ b/DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
@@ -927,21 +927,27 @@ namespace Microsoft.Interop
                 NativeIndex = info.NativeIndex
             };
 
-            StatementSyntax marshallingStatement = Block(
-                List(elementMarshaller.Generate(localElementInfo, elementSetupSubContext)
-                    .Concat(elementMarshaller.Generate(localElementInfo, elementSubContext))));
+            List<StatementSyntax> elementStatements = elementMarshaller.Generate(localElementInfo, elementSubContext).ToList();
 
-            if (elementMarshaller.AsNativeType(elementInfo) is PointerTypeSyntax)
+            if (elementStatements.Count != 0)
             {
-                PointerNativeTypeAssignmentRewriter rewriter = new(elementSubContext.GetIdentifiers(localElementInfo).native);
-                marshallingStatement = (StatementSyntax)rewriter.Visit(marshallingStatement);
-            }
+                StatementSyntax marshallingStatement = Block(
+                    List(elementMarshaller.Generate(localElementInfo, elementSetupSubContext)
+                        .Concat(elementStatements)));
 
-            // Iterate through the elements of the native collection to unmarshal them
-            return Block(
-                GenerateNativeSpanDeclaration(info, context),
-                MarshallerHelpers.GetForLoop(collectionIdentifierForLength, IndexerIdentifier)
-                                .WithStatement(marshallingStatement));
+                if (elementMarshaller.AsNativeType(elementInfo) is PointerTypeSyntax elementNativeType)
+                {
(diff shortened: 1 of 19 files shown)


DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
```diff
@@ -192,7 +191,16 @@ public bool UsesNativeIdentifier(TypePositionInfo info, StubCodeContext context)
         public IEnumerable<StatementSyntax> GenerateCleanupStatements(TypePositionInfo info, StubCodeContext context)
         {
             var subContext = new CustomNativeTypeWithValuePropertyStubContext(context);
-            return innerMarshaller.GenerateCleanupStatements(info, subContext);
+
+            if (!context.AdditionalTemporaryStateLivesAcrossStages)
```
AaronRobinsonMSFT, line 198: Please comment on why this must be done prior to the loop.

DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
```diff
@@ -594,7 +624,18 @@ public TypeSyntax AsNativeType(TypePositionInfo info)
 
         public IEnumerable<StatementSyntax> GenerateCleanupStatements(TypePositionInfo info, StubCodeContext context)
         {
-            return innerMarshaller.GenerateCleanupStatements(info, context);
+            if (!context.AdditionalTemporaryStateLivesAcrossStages)
```
AaronRobinsonMSFT, line 633: Comment the logic here.

DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
```diff
@@ -632,6 +673,14 @@ public IEnumerable<StatementSyntax> GenerateUnmarshalStatements(TypePositionInfo
                             IdentifierName(ManualTypeMarshallingHelper.SetUnmarshalledCollectionLengthMethodName)))
                     .AddArgumentListArguments(Argument(numElementsExpression)));
             }
+        }
+
+        public IEnumerable<StatementSyntax> GenerateUnmarshalStatements(TypePositionInfo info, StubCodeContext context)
+        {
+            foreach (var statement in GenerateUnmarshallerCollectionInitialization(info, context))
```
AaronRobinsonMSFT, line 691: Is this an intentional contract? We do the current prior to the inner. If so, we should state that here if it creates functional side-effects.

DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
```diff
@@ -874,22 +928,27 @@ private StatementSyntax GenerateContentsMarshallingStatement(TypePositionInfo in
...
             };
 
-            StatementSyntax marshallingStatement = Block(
-                List(elementMarshaller.Generate(
-                    localElementInfo,
-                    elementSubContext)));
+            List<StatementSyntax> elementStatements = elementMarshaller.Generate(localElementInfo, elementSubContext).ToList();
 
-            if (elementMarshaller.AsNativeType(elementInfo) is PointerTypeSyntax)
+            if (elementStatements.Count != 0)
```
AaronRobinsonMSFT: One of the few uses of Linq I like is the `elementStatements.Any()` &ndash; up to you.

AaronRobinsonMSFT (approved):

jkoritzinsky:
I've added comments to explain the reasoning behind the ordering of the emitted code.

PR #1217, opened 2021-06-09
Add configuration option for Unsafe type name.

jkoritzinsky:
In System.Private.CoreLib, the Unsafe type is internal, in a different namespace, and it is implemented in C# + Intrinsics instead of IL. Use a config option to enable switching the API name the generated code uses so the generator can be used with System.Private.CoreLib.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/Microsoft.Interop.DllImportGenerator.props
+++ b/DllImportGenerator/DllImportGenerator/Microsoft.Interop.DllImportGenerator.props
@@ -15,6 +15,11 @@
         of generating a stub that handles all of the marshalling.
       -->
       <CompilerVisibleProperty Include="DllImportGenerator_GenerateForwarders" />
+      <!--
+        Use the Internal.Runtime.ComplierServices.Unsafe type instead of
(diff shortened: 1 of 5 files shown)


DllImportGenerator/DllImportGenerator/Microsoft.Interop.DllImportGenerator.props
```diff
@@ -15,6 +15,11 @@
         of generating a stub that handles all of the marshalling.
       -->
       <CompilerVisibleProperty Include="DllImportGenerator_GenerateForwarders" />
+      <!--
+        Use the Internal.Runtime.ComplierServices.Unsafe type instead of
+        the System.Runtime.CompilerServices.Unsafe type when emitting code.
+      -->
+    <CompilerVisibleProperty Include="DllImportGenerator_UseInternalUnsafeType" />
```
AaronRobinsonMSFT, line 22:
```suggestion
      <CompilerVisibleProperty Include="DllImportGenerator_UseInternalUnsafeType" />
```

AaronRobinsonMSFT (approved):

elinor-fung (approved):

PR #1219, opened 2021-06-10
Fix ordering of variant and default interface method resolution

MichalStrehovsky:
Default interface methods need to be looked at only if the "old rules" didn't come up with an answer. We need to split default implementations in the dispatch map so that they can be looked at separately. This is to prevent an exact default implementation match from preempting a variant "old rule" match.


jkotas (approved):
Thank you!

PR #1221, opened 2021-06-10
Ignore resource files.

kant2002:
Can I assume that this is mistake of some sort, since this was in CoreRT from inception

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Build.Tasks/ComputeManagedAssembliesToCompileToNative.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Build.Tasks/ComputeManagedAssembliesToCompileToNative.cs
@@ -164,9 +164,12 @@ namespace Build.Tasks
 
                                 if (culture == "" || culture.Equals("neutral", StringComparison.OrdinalIgnoreCase))
                                 {
+                                    list.Add(taskItem);
+                                }
+                                else
+                                {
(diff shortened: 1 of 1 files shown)


jkotas:
What is the repro for the problem that this is trying to fix? Or is it just a cleanup that has no user-visible impact?

kant2002:
This was more of a question. For Windows Forms for example I expect that this do not copy satellite assemblies. But honestly I did not test yet, and use that PR as a question, hoping that I may miss something obvious. I will test tomorrow and let you know about what this change exactly brings

kant2002:
as I said before with this change for WinForms application, publish folder no longer contains satellite assemblies.
Based loading satellite assemblies is not supported.  

Also I notice that seems to be that any assembly which is in ManagedAssemblis should be as part of `AssembliesToSkipPublish` because they are consumed by ILC and no longer needed by final application.

src/coreclr/tools/aot/ILCompiler.Build.Tasks/ComputeManagedAssembliesToCompileToNative.cs
```diff
@@ -138,6 +138,7 @@ public override bool Execute()
                         }
                     }
 
+                    assembliesToSkipPublish.Add(taskItem);
```
kant2002, line 141: This one needed too. I miss that WindowsBase.dll then would be part of published application

jkotas (approved):
Thanks

PR #1222, opened 2021-06-10
Implement marshallers for Span and ReadOnlySpan

jkoritzinsky:
Implement marshallers for Span and ReadOnlySpan to show the extensibility model for a new interop type.

When we move in-box, we can annotate `Span<T>` and `ReadOnlySpan<T>` with `[NativeMarshalling]` attributes pointing to their default marshallers and remove some of the `[MarshalUsing]` attributes.

### DIFF
--- a/DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
+++ b/DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
@@ -7,44 +7,44 @@ namespace System.Runtime.InteropServices.GeneratedMarshalling
     [GenericContiguousCollectionMarshaller]
     public unsafe ref struct ReadOnlySpanMarshaller<T>
     {
-        private ReadOnlySpan<T> managedSpan;
-        private readonly int sizeOfNativeElement;
-        private IntPtr allocatedMemory;
+        private ReadOnlySpan<T> _managedSpan;
+        private readonly int _sizeOfNativeElement;
+        private IntPtr _allocatedMemory;
 
         public ReadOnlySpanMarshaller(int sizeOfNativeElement)
             : this()
         {
-            this.sizeOfNativeElement = sizeOfNativeElement;
+            _sizeOfNativeElement = sizeOfNativeElement;
         }
 
         public ReadOnlySpanMarshaller(ReadOnlySpan<T> managed, int sizeOfNativeElement)
         {
-            allocatedMemory = default;
-            this.sizeOfNativeElement = sizeOfNativeElement;
+            _allocatedMemory = default;
+            _sizeOfNativeElement = sizeOfNativeElement;
             if (managed.Length == 0)
             {
-                managedSpan = default;
+                _managedSpan = default;
                 NativeValueStorage = default;
                 return;
             }
-            managedSpan = managed;
-            this.sizeOfNativeElement = sizeOfNativeElement;
+            _managedSpan = managed;
+            _sizeOfNativeElement = sizeOfNativeElement;
             int spaceToAllocate = managed.Length * sizeOfNativeElement;
-            allocatedMemory = Marshal.AllocCoTaskMem(managed.Length);
-            NativeValueStorage = new Span<byte>((void*)allocatedMemory, spaceToAllocate);
+            _allocatedMemory = Marshal.AllocCoTaskMem(managed.Length);
+            NativeValueStorage = new Span<byte>((void*)_allocatedMemory, spaceToAllocate);
         }
 
         public ReadOnlySpanMarshaller(ReadOnlySpan<T> managed, Span<byte> stackSpace, int sizeOfNativeElement)
         {
-            allocatedMemory = default;
-            this.sizeOfNativeElement = sizeOfNativeElement;
+            _allocatedMemory = default;
+            _sizeOfNativeElement = sizeOfNativeElement;
             if (managed.Length == 0)
             {
-                managedSpan = default;
+                _managedSpan = default;
                 NativeValueStorage = default;
                 return;
             }
-            managedSpan = managed;
+            _managedSpan = managed;
             int spaceToAllocate = managed.Length * sizeOfNativeElement;
             if (spaceToAllocate <= stackSpace.Length)
             {
@@ -52,8 +52,8 @@ namespace System.Runtime.InteropServices.GeneratedMarshalling
             }
             else
             {
-                allocatedMemory = Marshal.AllocCoTaskMem(spaceToAllocate);
-                NativeValueStorage = new Span<byte>((void*)allocatedMemory, spaceToAllocate);
+                _allocatedMemory = Marshal.AllocCoTaskMem(spaceToAllocate);
+                NativeValueStorage = new Span<byte>((void*)_allocatedMemory, spaceToAllocate);
             }
         }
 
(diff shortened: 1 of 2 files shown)


DllImportGenerator/DllImportGenerator.IntegrationTests/SpanTests.cs
```diff
@@ -0,0 +1,170 @@
...
+                    b1 = true,
+                    b2 = true,
+                    b3 = result,
+                },
+            };
+
+            Assert.Equal(result, NativeExportsNE.Span.AndAllMembers(boolValues, boolValues.Length));
+        }
+
+        private static string ReverseChars(string value)
```
AaronRobinsonMSFT: I think we can remove this.

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
+
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System.Runtime.InteropServices.GeneratedMarshalling
+{
+    [GenericContiguousCollectionMarshaller]
+    public unsafe ref struct ReadOnlySpanMarshaller<T>
+    {
+        private ReadOnlySpan<T> managedSpan;
```
AaronRobinsonMSFT: As long as this is a new file and it is on my mind, let's start following the corefx styleguide. It will help me in the long run.

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
...
+            this.sizeOfNativeElement = sizeOfNativeElement;
+            if (managed.Length == 0)
+            {
+                managedSpan = default;
+                NativeValueStorage = default;
+                return;
+            }
+            managedSpan = managed;
+            int spaceToAllocate = managed.Length * sizeOfNativeElement;
+            if (spaceToAllocate < stackSpace.Length)
```
AaronRobinsonMSFT: `<=`

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
...
+        /// <summary>
+        /// Stack-alloc threshold set to 256 bytes to enable small arrays to be passed on the stack.
+        /// Number kept small to ensure that P/Invokes with a lot of array parameters doesn't
+        /// blow the stack since this is a new optimization in the code-generated interop.
+        /// </summary>
+        public const int StackBufferSize = 0x200;
+
+        public Span<T> ManagedValues => MemoryMarshal.CreateSpan(ref MemoryMarshal.GetReference(managedSpan), managedSpan.Length);
+
+        public Span<byte> NativeValueStorage { get; private set; }
```
AaronRobinsonMSFT, line 69: Can we change `private set;` to `init;`?

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
...
+
+        public ref byte GetPinnableReference() => ref MemoryMarshal.GetReference(NativeValueStorage);
+
+        public void SetUnmarshalledCollectionLength(int length)
+        {
+            managedSpan = new T[length];
```
AaronRobinsonMSFT: Should we assert the current state of `managedSpan`?

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
...
+                    NativeValueStorage = new Span<byte>(value, managedSpan.Length * sizeOfNativeElement);
+                }
+            }
+        }
+
+        public ReadOnlySpan<T> ToManaged() => managedSpan;
+
+        public void FreeNative()
+        {
+            if (allocatedMemory != IntPtr.Zero)
```
AaronRobinsonMSFT: We check for `null` so many times in this code path. I would avoid it here.

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
...
+                NativeValueStorage = new Span<byte>((void*)allocatedMemory, spaceToAllocate);
+            }
+        }
+
+        /// <summary>
+        /// Stack-alloc threshold set to 256 bytes to enable small arrays to be passed on the stack.
+        /// Number kept small to ensure that P/Invokes with a lot of array parameters doesn't
+        /// blow the stack since this is a new optimization in the code-generated interop.
+        /// </summary>
+        public const int StackBufferSize = 0x200;
```
AaronRobinsonMSFT, line 65: Do we detect the case where `T` is > `StackBufferSize` during code gen? If we detect that I assume we can trim down the generated code and not use the stack path at all.

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
...
+        public Span<byte> NativeValueStorage
+        {
+            get => inner.NativeValueStorage;
+        }
+
+        public ref byte GetPinnableReference()
+        {
+            if (inner.ManagedValues.Length == 0)
+            {
+                return ref *(byte*)0x1;
```
AaronRobinsonMSFT: Thoughts on using a sentinel value? `0xa5a5a5a5`? `0x11111111`?

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
...
+
+            int spaceToAllocate = managed.Length * Unsafe.SizeOf<T>();
+            allocatedMemory = (T*)Marshal.AllocCoTaskMem(spaceToAllocate);
+            data = managed;
+        }
+
+        public DirectSpanMarshaller(Span<T> managed, Span<byte> stackSpace, int sizeOfNativeElement)
+            :this(sizeOfNativeElement)
+        {
+            data = managed;
```
AaronRobinsonMSFT: Assert that `stackSpace` is empty?

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
```diff
@@ -0,0 +1,398 @@
...
+        }
+
+        public Span<T> ToManaged()
+        {
+            return data;
+        }
+
+        public void FreeNative()
+        {
+            if (allocatedMemory != null)
```
AaronRobinsonMSFT: Remove `null` check.

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
jkoritzinsky: We cannot. The `NativeValueStorage` property is updated in the `Value` property's setter.

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
jkoritzinsky: The `managedSpan` state can either be `null` (byref out or return value case) or still have the value of the original managed span (the byref `ref` case), so I don't think there's anything we can assert here.

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
jkoritzinsky: We do not, but that would be an interesting idea. Technically we don't require that the stack-allocated space is used to store the elements of the collection if possible (see `DirectSpanMarshaller` just using it as a sentinel for supported scenarios), but we could add that requirement and skip the stack buffer if the native size of one element couldn't fit in the buffer.

DllImportGenerator/Ancillary.Interop/SpanMarshallers.cs
jkoritzinsky: I'd be fine using a sentinel. I'll change this to use one.

AaronRobinsonMSFT (approved):

PR #1226, opened 2021-06-10
Implement support for collections of collections marshalling

jkoritzinsky:
The current code in the feature/DllImportGenerator branch almost supports passing collections of collections (like jagged arrays) by value or by-ref `in` (it only needs the first commit that handles nested for loops), but it does not correctly support passing collections of collections in a way that requires unmarshalling.

This PR adds the functionality required to enable marshalling collections of collections like jagged arrays from native code back to managed code.

This code is designed such that if n levels of indirection is supported, then n+1 levels is also supported. The `CollectionsOfCollectionsStress` code snippet test validates this by testing 0-11 dimensional jagged arrays.

To enable safely tracing dependencies of "CountElementName" metadata values, this PR enhances the cycle-breaking in the MarshallingAttributeInfoParser to correctly break cycles.

To identify dependencies between parameters/return values, this PR also introduces a topological sorting algorithm to order the parameter/return value unmarshalling in an order that ensures dependencies are marshalled before elements that depend on them.

This algorithm is specified generically so developers of interop source generators could use this mechanism to support their own scenarios. For example, WinRT's implicit array length parameters could have part of their implementation supported by the topological sort.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/ContiguousCollectionElementMarshallingCodeContext.cs
+++ b/DllImportGenerator/DllImportGenerator/ContiguousCollectionElementMarshallingCodeContext.cs
@@ -2,7 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-
+using System.Text;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -12,13 +12,14 @@ namespace Microsoft.Interop
 {
     internal sealed class ContiguousCollectionElementMarshallingCodeContext : StubCodeContext
     {
-        private readonly string indexerIdentifier;
         private readonly string nativeSpanIdentifier;
         private readonly StubCodeContext parentContext;
 
         public override bool SingleFrameSpansNativeContext => false;
 
         public override bool AdditionalTemporaryStateLivesAcrossStages => false;
+
+        public string IndexerIdentifier { get; }
 
         /// <summary>
         /// Create a <see cref="StubCodeContext"/> for marshalling elements of an collection.
@@ -29,12 +30,11 @@ namespace Microsoft.Interop
         /// <param name="parentContext">The parent context.</param>
         public ContiguousCollectionElementMarshallingCodeContext(
             Stage currentStage,
-            string indexerIdentifier,
             string nativeSpanIdentifier,
             StubCodeContext parentContext)
         {
             CurrentStage = currentStage;
-            this.indexerIdentifier = indexerIdentifier;
+            IndexerIdentifier = CalculateIndexerIdentifierBasedOnParentContext(parentContext);
             this.nativeSpanIdentifier = nativeSpanIdentifier;
             this.parentContext = parentContext;
         }
@@ -48,20 +48,36 @@ namespace Microsoft.Interop
         {
             var (_, native) = parentContext.GetIdentifiers(info);
             return (
-                $"{native}.ManagedValues[{indexerIdentifier}]",
-                $"{nativeSpanIdentifier}[{indexerIdentifier}]"
+                $"{native}.ManagedValues[{IndexerIdentifier}]",
+                $"{nativeSpanIdentifier}[{IndexerIdentifier}]"
             );
         }
 
         public override string GetAdditionalIdentifier(TypePositionInfo info, string name)
         {
-            return $"{nativeSpanIdentifier}__{indexerIdentifier}__{name}";
+            return $"{nativeSpanIdentifier}__{IndexerIdentifier}__{name}";
         }
 
         public override TypePositionInfo? GetTypePositionInfoForManagedIndex(int index)
         {
             // We don't have parameters to look at when we're in the middle of marshalling an array.
             return null;
+        }
+
+        private static string CalculateIndexerIdentifierBasedOnParentContext(StubCodeContext parentContext)
+        {
+            int i = 0;
+            while (parentContext is ContiguousCollectionElementMarshallingCodeContext collectionContext)
+            {
+                i++;
+                parentContext = collectionContext.parentContext;
+            }
+
+            // Follow a progression of indexers of the following form:
+            // __i, __j, __k, __ii, __jj, __kk, etc.
+            StringBuilder indexerIdentifier = new StringBuilder("__");
+            indexerIdentifier.Append((char)('i' + (i % 3)), 1 + (i / 3));
+            return indexerIdentifier.ToString();
         }
     }
 }
--- a/DllImportGenerator/DllImportGenerator/ContiguousCollectionElementMarshallingCodeContext.cs
+++ b/DllImportGenerator/DllImportGenerator/ContiguousCollectionElementMarshallingCodeContext.cs
@@ -76,10 +76,8 @@ namespace Microsoft.Interop
             }
 
             // Follow a progression of indexers of the following form:
-            // __i, __j, __k, __ii, __jj, __kk, etc.
-            StringBuilder indexerIdentifier = new StringBuilder("__");
-            indexerIdentifier.Append((char)('i' + (i % 3)), 1 + (i / 3));
-            return indexerIdentifier.ToString();
+            // __i0, __i1, __i2, __i3, etc/
+            return $"__i{i}";
         }
     }
 }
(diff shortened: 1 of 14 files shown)


DllImportGenerator/DllImportGenerator/ContiguousCollectionElementMarshallingCodeContext.cs
```diff
@@ -46,22 +45,41 @@ internal sealed class ContiguousCollectionElementMarshallingCodeContext : StubCo
...
+                {
+                    i++;
+                }
+                parentContext = context.ParentContext;
+            }
+
+            // Follow a progression of indexers of the following form:
+            // __i, __j, __k, __ii, __jj, __kk, etc.
+            StringBuilder indexerIdentifier = new StringBuilder("__");
+            indexerIdentifier.Append((char)('i' + (i % 3)), 1 + (i / 3));
```
AaronRobinsonMSFT: I think I would prefer to simply append the depth - `__i0`, `__i1`, ...,  `__iN`, etc. Computing the depth from the current form appears much harder. Is there a benefit of current transform?

DllImportGenerator/DllImportGenerator/ContiguousCollectionElementMarshallingCodeContext.cs
```diff
@@ -12,14 +12,14 @@ namespace Microsoft.Interop
...
     {
-        private readonly string indexerIdentifier;
         private readonly string nativeSpanIdentifier;
-        private readonly StubCodeContext parentContext;
 
         public override bool SingleFrameSpansNativeContext => false;
 
         public override bool AdditionalTemporaryStateLivesAcrossStages => false;
 
+        public string IndexerIdentifier { get; }
```
AaronRobinsonMSFT, line 21: `init;`?

DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
```diff
@@ -127,26 +127,24 @@ public IEnumerable<StatementSyntax> GeneratePinStatements(TypePositionInfo info,
...
-
         public CustomNativeTypeWithValuePropertyStubContext(StubCodeContext parentContext)
         {
-            this.parentContext = parentContext;
+            ParentContext = parentContext;
             CurrentStage = parentContext.CurrentStage;
         }
 
-        public override bool SingleFrameSpansNativeContext => parentContext.SingleFrameSpansNativeContext;
+        public override bool SingleFrameSpansNativeContext => ParentContext!.SingleFrameSpansNativeContext;
```
AaronRobinsonMSFT, line 136: I don't see the `?` - is the `!` needed?

DllImportGenerator/DllImportGenerator/Marshalling/MarshallerHelpers.cs
```diff
@@ -86,10 +88,152 @@ public static TypeSyntax GetCompatibleGenericTypeParameterSyntax(this TypeSyntax
...
+        /// <returns>A topologically sorted collection of the elemens of the <paramref name="elements"/> collection.</returns>
+        /// <exception cref="InvalidOperationException">The graph of <paramref name="elements"/> nodes and the edges produced by <paramref name="getDependentIndicesFn"/> has cycles.</exception>
+        public static IEnumerable<T> GetTopologicallySortedElements<T, U>(
+            ICollection<T> elements,
+            Func<T, U> keyFn,
+            Func<T, IEnumerable<U>> getDependentIndicesFn)
+        {
+            Dictionary<U, int> elementIndexToEdgeMapNodeId = new(elements.Count);
+            List<T> nodeIdToElement = new(elements.Count);
+            EdgeMap edgeMap = new EdgeMap(elements.Count);
```
AaronRobinsonMSFT: Let's be consistent here with `new` if possible.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallerHelpers.cs
```diff
@@ -86,10 +88,152 @@ public static TypeSyntax GetCompatibleGenericTypeParameterSyntax(this TypeSyntax
...
+                    {
+                        S.Add(nodeIdToElement[m]);
+                    }
+                }
+            }
+
+            // If we have edges left, then we have a cycle.
+            if (edgeMap.AnyEdges)
+            {
+                throw new InvalidOperationException();
```
AaronRobinsonMSFT: Let's add a nice message here. This should be considered a catastrophic error for the source generator.

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -392,21 +392,72 @@ ExpressionSyntax GetExpressionForParam(TypePositionInfo? paramInfo)
...
+                    {
+                        indexerStack.Push(collectionContext.IndexerIdentifier);
+                    }
+                    lastContext = currentContext;
+                    currentContext = currentContext.ParentContext;
+                }
+
+                numIndirectionLevels = indexerStack.Count;
+
+                ExpressionSyntax indexedNumElements = IdentifierName(lastContext!.GetIdentifiers(numElementsInfo).managed);
```
AaronRobinsonMSFT: Why is the `!` needed here?

DllImportGenerator/DllImportGenerator/Marshalling/MarshallingGenerator.cs
```diff
@@ -392,21 +392,72 @@ ExpressionSyntax GetExpressionForParam(TypePositionInfo? paramInfo)
...
+                while (indexerStack.Count > 0)
+                {
+                    indexedNumElements = ElementAccessExpression(indexedNumElements).AddArgumentListArguments(Argument(IdentifierName(indexerStack.Pop())));
```
AaronRobinsonMSFT: Please break this nesting up.

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
```diff
@@ -76,6 +77,16 @@ internal sealed class StubCodeGenerator : StubCodeContext
             // Get marshaller for return
             this.retMarshaller = CreateGenerator(retTypeInfo);
 
+
+            List<(TypePositionInfo TypeInfo, IMarshallingGenerator Generator)> allMarshallers = new(this.paramMarshallers);
+            allMarshallers.Add(retMarshaller);
+
+            this.sortedMarshallers = MarshallerHelpers.GetTopologicallySortedElements(
```
AaronRobinsonMSFT, line 106: Comment why we are doing this with an example.

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
```diff
@@ -202,43 +235,57 @@ public BlockSyntax GenerateSyntax()
...
+                            }
+                            else
+                            {
+                                statements.AddRange(generatedStatements);
+                            }
                         }
                     }
                 }
 
+
```
AaronRobinsonMSFT: Extra line.

DllImportGenerator/DllImportGenerator/ContiguousCollectionElementMarshallingCodeContext.cs
jkoritzinsky: I like the idea of appending the depth. That's a better idea. I'll switch it to that.

DllImportGenerator/DllImportGenerator/Marshalling/ICustomNativeTypeMarshallingStrategy.cs
jkoritzinsky: `ParentContext` is nullable, so yes this is needed.

AaronRobinsonMSFT: Oh casing.

DllImportGenerator/DllImportGenerator/ContiguousCollectionElementMarshallingCodeContext.cs
jkoritzinsky: This property is calculated internally, so we don't want external code that creates `ContiguousCollectionElementMarshallingCodeContext` instances to assign it.

AaronRobinsonMSFT: Oh. I didn't realize that `init` was also subject to scope - I thought it was limited to constructor.

AaronRobinsonMSFT (approved):

PR #1227, opened 2021-06-11
Respect RuntimeHostConfigurationOption

MichalStrehovsky:
Setting RuntimeHostConfigurationOption in the project file is observable in AppContext. Pipe the values to ILC and preserve them in the executable.


jkotas (approved):
Thank you!

PR #1232, opened 2021-06-12
Try removing `/EHs`

MichalStrehovsky:
Just wondering whether we need it since we do native EH through `AddVectoredExceptionHandler` and unwind ourselves.

I haven't actually tried locally...


jkotas (approved):
Does it make any observable difference?

MichalStrehovsky:
3 kB size saving, which is meh, but it also removes dependency on `__CxxFrameHandler4` from the C++ runtime library.

PR #1237, opened 2021-06-14
NativeAOT-LLVM remove restriction of 0 params for compiling with clrjit

yowl:
This PR removes the restriction on attempting to compile with clrjit any method with > 0 parameters.  Adds the generation of the managed to native thunk for runtime export functions compiled with clrjit.


jkotas (approved):
LGTM. Thank you!

PR #1243, opened 2021-06-15
Couple interop fixes

MichalStrehovsky:
* When propagating from byref arguments, we need to propagate the native value in case of reverse p/invoke.
* SafeHandle marshalling with an abstract SafeHandle type throws MarshalDirective.


MichalStrehovsky:
(This is covered by Pri0 CoreCLR tests)

jkotas (approved):
Thank you!

PR #1245, opened 2021-06-15
Delete .NET Native-specific DebugEventSource

MichalStrehovsky:
Deletes the ETW-based mechanism by which exception handling communicates with the .NET Native debugger.


jkotas (approved):
Thanks

PR #1246, opened 2021-06-15
NativeAOT-LLVM: Pass this to calls - initial support for GT_IND and GT_NE

yowl:
This PR adds support for passing `this` to instance methods from the clrjit->LLVM module.  Also starts support for GT_IND and GT_NE.  I moved a few methods around to prevent forward references, so sorry about that churn.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -35,7 +35,6 @@ static const char* (*_addCodeReloc)(void*, void*);
 static const uint32_t (*_isRuntimeImport)(void*, CORINFO_METHOD_STRUCT_*);
 static char*     _outputFileName;
 static Function* _doNothingFunction;
-static unsigned int                       _pointerSize;
 Compiler::Info                            _info;
 Compiler*                                 _compiler;
 Function*                                 _function;
@@ -65,8 +64,6 @@ extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
         _module = new Module(llvm::StringRef("netscripten-clrjit"), _llvmContext);
         _module->setTargetTriple(triple);
         _module->setDataLayout(dataLayout);
-        // TODO: better to get this from compiler, but where is it?
-        _pointerSize = _module->getDataLayout().getPointerSize();
         _outputFileName = (char*)malloc(strlen(outputFileName) + 7);
         strcpy(_outputFileName, "1.txt"); // ??? without this _outputFileName is corrupted
         strcpy(_outputFileName, outputFileName);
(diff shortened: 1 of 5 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -64,6 +65,8 @@ extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
         _module = new Module(llvm::StringRef("netscripten-clrjit"), _llvmContext);
         _module->setTargetTriple(triple);
         _module->setDataLayout(dataLayout);
+        // TODO: better to get this from compiler, but where is it?
```
SingleAccretion: I don't think we have a dynamic value for it. Forms I've personally seen in the past: `TARGET_POINTER_SIZE` define, `genTypeSize(TYP_I_IMPL)`. Seems like the define is most popular.

src/coreclr/jit/llvm.cpp
```diff
@@ -235,8 +283,65 @@ Value* genTreeAsLlvmType(GenTree* tree, Type* type)
     failFunctionCompilation();
 }
 
-llvm::Value* buildUserFuncCall(GenTreeCall* gtCall, llvm::IRBuilder<>& builder)
+int getTotalParameterOffset(CORINFO_SIG_INFO &sigInfo)
```
```suggestion
int getTotalParameterOffset(CORINFO_SIG_INFO& sigInfo)
```

src/coreclr/jit/llvm.cpp
```diff
@@ -235,8 +283,65 @@ Value* genTreeAsLlvmType(GenTree* tree, Type* type)
...
+{
+    if (offset == 0)
+        return shadowStack;
+
+    return _builder->CreateGEP(shadowStack, _builder->getInt32(_pointerSize));
+}
+
+bool isThisArg(GenTreeCall* gtCall, GenTree* operand)
 {
+    // TODO: is the this arg always a late arg?
```
SingleAccretion: I think you can make this logic not dependent on the lateness of `this`:
```c++
if (gtCall->gtCallThisArg == nullptr)
{
    return false;
}
   
return _compiler->gtGetThisArg(gtCall) == operand;
```
Also nit: `gtCall` -> `call` (here and elsewhere). The purpose of `gt`/`fg`/`gen` and others is the namespacing of sorts for when everything is a `Compiler` member. I don't think there is need to use the prefixes for locals.

yowl: Much better, thanks.

src/coreclr/jit/llvm.cpp
yowl: Done.

yowl:
Having problems with the CI...

SingleAccretion:
The CI <-> GitHub flow has broken recently, but the pipeline is still running. See https://dev.azure.com/dnceng/public/_build?definitionId=841.

jkotas (approved):
LGTM

PR #1248, opened 2021-06-16
RuntimeType devirtualization when using different reflection stack

MichalStrehovsky:
* Don't pass reference to System.Private.Reflectin.Core when reflection is disabled
* Give System.Type to RyuJIT (RyuJIT asserts if we give back null). This should be fine since RyuJIT has to assume inexact type.


jkotas (approved):
Thank you

PR #1251, opened 2021-06-17
NativeAOT-LLVM: support for args passed on shadow stack

yowl:
This PR adds support for passing and retrieving of CLASS and BYREF type arguments on the shadow stack, following the rules used in ILImporter.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -24,6 +24,12 @@ using llvm::ArrayRef;
 using llvm::Module;
 using llvm::Value;
 
+struct LlvmArgInfo
+{
+    int m_argIx; // -1 indicates not in the LLVM arg list, but on the shadow stack
+    unsigned int m_shadowStackOffset;
+};
+
 typedef JitHashTable<BasicBlock*, JitPtrKeyFuncs<BasicBlock>, llvm::BasicBlock*> BlkToLlvmBlkVectorMap;
 
 static Module* _module = nullptr;
@@ -43,6 +49,7 @@ llvm::IRBuilder<>*                        _builder;
 std::unordered_map<unsigned int, Value*>* _sdsuMap;
 std::unordered_map<unsigned int, Value*>* _localsMap;
 CORINFO_SIG_INFO                          _sigInfo; // sigInfo of function being compiled
+llvm::IRBuilder<>*                        _prologBuilder;
 
 extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
                                                 const char* outputFileName,
@@ -135,10 +142,7 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
         case CorInfoType::CORINFO_TYPE_ULONG:
             return Type::getInt64Ty(_llvmContext);
 
-        // needs to go on the shadow stack.  TODO when Ilc module is gone, as a performance improvement can we pass byrefs to non struct value types on the llvm stack?
         case CorInfoType::CORINFO_TYPE_BYREF:
-            failFunctionCompilation();
-
         case CorInfoType::CORINFO_TYPE_CLASS:
             return Type::getInt8PtrTy(_llvmContext);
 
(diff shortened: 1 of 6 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -452,7 +640,27 @@ Value* buildInd(llvm::IRBuilder<>& builder, GenTree* node, Value* ptr)
...
-    return mapTreeIdValue(node->gtTreeID, builder.CreateICmpNE(op1, op2));
+    // TODO: when the next integer binary operator is implemented, factor out the widening
+    Type* op1Type = op1->getType();
+    Type* op2Type = op2->getType();
+    if (op1Type == op2Type)
+    {
+        // no widening required
+        return mapTreeIdValue(node->gtTreeID, builder.CreateICmpNE(op1, op2));
+    }
+    else
```
SingleAccretion: I am curious: where are you seeing the type mismatches? The frontend should normalize them in almost all cases, except for shifts.

src/coreclr/jit/llvm.cpp
```diff
@@ -135,31 +142,135 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
...
+    return padOffset(corInfoType, atOffset) + size;
+}
+
+/// <summary>
+/// Returns true if the type can be stored on the LLVM stack
+/// instead of the shadow stack in this method.
+/// </summary>
+bool canStoreTypeOnLlvmStack(CorInfoType corInfoType)
+{
+    // TODO: CORINFO_TYPE_VALUECLASS maps to a c# struct?
```
SingleAccretion: > CORINFO_TYPE_VALUECLASS maps to a c# struct?

Yes.

src/coreclr/jit/llvm.cpp
```diff
@@ -135,31 +142,135 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
...
+        failFunctionCompilation();
+    }
+
+    if (corInfoType == CorInfoType::CORINFO_TYPE_BYREF || corInfoType == CorInfoType::CORINFO_TYPE_CLASS)
+    {
+        return false;
+    }
+    if (corInfoType == CorInfoType::CORINFO_TYPE_REFANY || corInfoType == CorInfoType::CORINFO_TYPE_VAR)
+    {
+        //TODO: what scenarios have these
```
SingleAccretion: `CORINFO_TYPE_VAR` used to be used for verification:
```
// CORINFO_TYPE_VAR is for a generic type variable.
// Generic type variables only appear when the JIT is doing
// verification (not NOT compilation) of generic code
// for the EE, in which case we're running
// the JIT in "import only" mode.
```
Verification is no more, so you should never see it.

`CORINFO_TYPE_REFANY` is a `TypedReference`.

src/coreclr/jit/llvm.cpp
```diff
@@ -248,6 +359,56 @@ void castingStore(llvm::IRBuilder<>& builder, Value* toStore, Value* address, va
...
+            if (lclNum == i)
+            {
+                llvmArgInfo.m_shadowStackOffset = shadowStackOffset;
+                return llvmArgInfo;
+            }
+
+            shadowStackOffset += TARGET_POINTER_SIZE; // TODO size of arg, for now only handles byrefs and class types
+        }
+    }
+    failFunctionCompilation(); // lclNum not an argument, TODO: how to indicate a failure in compilation
```
SingleAccretion: Failure from bad, like bad IL? I suppose, the frontend should've failed on it. Though here it looks like you would like to assert the path is unreachable?

We have an `unreachable()` macro, or you can `assert(!"lclNum not an argument");` (more informative in my opinion). There is also a `noway_assert`, which is an `assert` that's enabled in Release builds (for our purposes, in CoreCLR it also makes the EE try to compile the function in MinOpts).

src/coreclr/jit/llvm.cpp
```diff
@@ -135,31 +142,135 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
...
+    }
+    else
+    {
+        retLlvmType   = getLlvmTypeForCorInfoType(sigInfo.retType);
+    }
+
     CORINFO_ARG_LIST_HANDLE  sigArgs = sigInfo.args;
-    argVec[0]                        = Type::getInt8PtrTy(_llvmContext); // shadowstack arg
+
+    //TODO: not attempting to compile generic signatures with context arg via clrjit yet, is it the same as sigInfo.hasTypeArg()?
```
SingleAccretion: > is it the same as sigInfo.hasTypeArg()?

Yes.

src/coreclr/jit/llvm.cpp
```diff
@@ -135,31 +142,135 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
...
+    else
+    {
+        // TODO: value type field alignment - this is the ILToLLVMImporter logic:
+        //var fieldAlignment = type is DefType && type.IsValueType ? ((DefType)type).InstanceFieldAlignment
+        //                                                         : type.Context.Target.LayoutPointerSize;
+        //var alignment      = LayoutInt.Min(fieldAlignment, new LayoutInt(ComputePackingSize(type))).AsInt;
+        //var padding        = (atOffset + (alignment - 1)) & ~(alignment - 1);
+        failFunctionCompilation();
+    }
+    return (atOffset + (alignment - 1)) & ~(alignment - 1);
```
```suggestion
    return roundUp(atOffset, alignment);
```

src/coreclr/jit/llvm.cpp
yowl: I hit this with I think the short circuit evaluation of e.g. https://github.com/dotnet/runtimelab/blob/5054e33a8bc08e05da77233ed2d2ca29a2400a56/src/coreclr/nativeaot/System.Private.CoreLib/src/System/RuntimeExceptionHelpers.cs#L239.  The IR being processed starts with
```
------------ BB01 [000..008) -> BB03 (cond), preds={} succs={BB02,BB03}
               [000076] ------------                 IL_OFFSET void   IL offset: 0x0
N001 (  1,  1) [000000] ------------                 NO_OP     void
N001 ( 14,  2) [000001] H-CXG-------         t1 =    CALL help byref  HELPER.CORINFO_HELP_READYTORUN_STATIC_BASE Zero Fseq[Value]
                                                  /--*  t1     byref
N002 ( 18,  5) [000004] ---XG-------         t4 = *  IND       bool
N003 (  1,  1) [000005] ------------         t5 =    CNS_INT   int    0
                                                  /--*  t4     bool
                                                  +--*  t5     int
N004 ( 20,  7) [000006] J--XG--N----         t6 = *  NE        int
                                                  /--*  t6     int
N005 ( 22,  9) [000007] ---XG-------              *  JTRUE     void
```
Where the NE operates on `t4` and `t5`.  `t4` is a bool which I currently translate to an `i8` in LLVM and `t5` is an `int`, LLVM `i32` .  The LLVM `icmp` won't like that.  I suppose in this case as `t5` is 0, that could go to an `i8` which would be more optimal and avoid the widening.  Maybe I need to look harder and there is more information available in `CNS_INT   int    0` but not dumped here, that tells me 8 bits is enough.

src/coreclr/jit/llvm.cpp

src/coreclr/jit/llvm.cpp
SingleAccretion: > N002 ( 18,  5) [000004] ---XG-------         t4 = *  IND       bool

Ah, of course, the indirections off of small types. Those always implicitly extend in IR (so while in IR the type is e. g. `short`, the type that the user consumes is `int`).

I would personally go with just upcasting the indirections (and a couple other nodes that are indir-like: `CLS_VAR`, `LCL_FLD`), since that should be the only case where this happens (well, there's one more case: morph can retype relops as `TYP_UBYTE` sometimes, but I have a PR which moves this to lowering). But up to you if you want to follow the IR semantics or build on top of them here.

>  Maybe I need to look harder and there is more information available in CNS_INT int 0 but not dumped here, that tells me 8 bits is enough.

I don't think so. Constants in this phase are always `LONG` or `INT` (`BYREF`/`REF`), there are no small types for them (only codegen sometimes retypes them to bytes to emit "narrow" instructions on XARCH).

src/coreclr/jit/llvm.cpp
yowl: Thanks, I'll ignore `CORINFO_TYPE_VAR ` and place `CORINFO_TYPE_REFANY` on the shadow stack.

src/coreclr/jit/llvm.cpp
yowl: I guess I was thinking, more failure in my own programming skills, and hence a bug in my code.  I'll go with `assert`.

yowl: followed by `UNREACHABLE()`

src/coreclr/jit/llvm.cpp

src/coreclr/jit/llvm.cpp
SingleAccretion: FWIW, `unreachable()` is usually used as an escape hatch from compiler warnings, or for truly "unreachable" stuff. I would think an assert is enough?

SingleAccretion: One thing to note is that `unreachable()` is on in Release builds, but does that matter given `gtTreeID` is `DEBUG`-only, so Release builds are not that big a concern?

yowl: "gtTreeID is DEBUG-only"  Oh that's a shame.  We used to support Release builds in the CoreRT repo.  It got dropped in the move to runtimelab, but I would like to enable them again.  I'd better look at removing that in the next PR.  For this change, I'll need to return something, or call something that never returns.  I'll add a `failFunctionCompilation` so the compiler is ok with no return value.  That will also be ok for Release builds while the transition from IL to IR is in progress.

SingleAccretion: > 'll need to return something, or call something that never returns

Oh I see. Then `unreachable()` is the macro to use.

src/coreclr/jit/llvm.cpp
yowl: Thanks, have gone for upcasting the indirection.

src/coreclr/jit/llvm.cpp
yowl: At the risk of continuing the discussion, I replaced the `for` with  a `do/while` to remove the multiple returns and while still a bit of an odd structure, I think it is a small improvement.

src/coreclr/jit/llvm.cpp
yowl: Done, thanks

src/coreclr/jit/llvm.cpp
```diff
@@ -447,7 +638,8 @@ Value* buildCnsInt(llvm::IRBuilder<>& builder, GenTree* node)
 
 Value* buildInd(llvm::IRBuilder<>& builder, GenTree* node, Value* ptr)
 {
-    return mapTreeIdValue(node->gtTreeID, builder.CreateLoad(castIfNecessary(builder, ptr, getLLVMTypeForVarType(node->TypeGet())->getPointerTo())));
+    // pass true to widen small ints to i32, TODO: will need to do the same for CLS_VAR, LCL_FLD
```
SingleAccretion: > pass true to widen small ints to i32

An alternative design would be to pass `genActualType(node->TypeGet())` (for now, can be simplified to just `genActualType(node)` once you merge `main`) and not modify `getLLVMTypeForVarType`. `genActualType` maps all integers to `int/long` (see also `gtlist.h`).

yowl: Thanks, that's better.

jkotas (approved):
LGTM. Thank you!

jkotas:
@SingleAccretion Thank you for your feedback!

PR #1255, opened 2021-06-18
added full support for cultures in SRM

veanes:
This fixed the bug in SRM that all regexes were compiled 
(and special characters (i,I) were treated) as if always in en-US culture.
This affected correctness of case-insensitivity under other cultures like "tr" or "".
Now all cultures are supported. When a regex is created in DFA mode its culture 
(e.g. InvariantCulture if passed through regex options) and is "frozen" 
and will be part of the regex, and is also preserved when serialized/deserialized, 
so the special internal serialization was updated to include culture info. 

Added unit test to validate new culture support, using cultures "", "tr" and "en-US"


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/CharSetSolver.cs
```diff
@@ -97,7 +58,9 @@ public BDD MkCharConstraint(char c, bool ignoreCase = false)
         /// </summary>
         public BDD MkCharSetFromRange(char m, char n)
         {
-            return MkSetFromRange((uint)m, (uint)n, _bw-1);
+            if (m == n)
+                return MkCharConstraint(m);
+            return MkSetFromRange((uint)m, (uint)n, 15);
```
olsaarik, line 63: Are these changes from `_bw-1` to 15 related to the culture support?

olsaarik (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/CharSetSolver.cs
veanes: Not really, more part of cleanup, I eliminated the _bw field. It is always 16. One could define the constant maxcharbit somewhere as 15  instead of using 15 explicitly.

PR #1256, opened 2021-06-18
NativeAOT-LLVM: add null terminator to name strings for c++

yowl:
This PR fixes an issue passing byte arrays from managed to c++ where the array is interpreted as a string and was not previously null terminated.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
@@ -75,9 +75,9 @@ namespace Internal.JitInterface
             node.AppendMangledName(_this._compilation.NameMangler, sb);
             if (node is FrozenStringNode)
             {
-                sb.Append("___SYMBOL");
+                sb.Append("___SYMBOL\0");
             }
-            return (byte*)_this.GetPin(AppendNullByte(sb.UnderlyingArray));
+            return (byte*)_this.GetPin(sb.UnderlyingArray);
         }
 
         [UnmanagedCallersOnly]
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -68,7 +77,7 @@ public static void addCodeReloc(IntPtr thisHandle, void* handle)
             {
                 sb.Append("___SYMBOL");
             }
-            return (byte*)_this.GetPin(sb.UnderlyingArray);
+            return (byte*)_this.GetPin(AppendNullByte(sb.UnderlyingArray));
```
jkotas: You can append the null byte directly to the `Utf8StringBuilder` here. It will save a string copy.

yowl: Thanks

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -66,7 +75,7 @@ public static void addCodeReloc(IntPtr thisHandle, void* handle)
             node.AppendMangledName(_this._compilation.NameMangler, sb);
             if (node is FrozenStringNode)
             {
-                sb.Append("___SYMBOL");
+                sb.Append("___SYMBOL\0");
```
jkotas: Do you need to append the `0` even when it is not `FrozenStringNode` ?

yowl: Yes, sorry that was careless.  Moved after the `if` now.

PR #1258, opened 2021-06-19
fixed SRM case-insensitivity bug

veanes:
The bug was that a regex such as (?i:[\0-@B-\uFFFF]) is turned into (?i:[^A]) -- the regex parser 
makes this transformation implicitly which then for SRM loses the distinction if it was (?i:[^A]) originally or not
since (?i:[\0-@B-\uFFFF]) and (?i:[^A]) have different semantics although [^A] and [\0-@B-\uFFFF] are 
otherwise equivalent. Somehow this transformation works correctly in non-DFA 
mode as the info can be recovered by some means that seems not available in SRM.

To fix the bug I turned off an optimization in the parser in Canonicalize in DFA mode that normalizes [\0-@B-\uFFFF] to [^A] internally as a special implicit representation (IsSingletonInverse).

Also cleaned up the initial SRM creation avoiding calls in Regex initialization not needed for SRM.

Added unit tests many of which failed before the fixes.
Also fixed a bug that .* was marked lazy (that showed up here but was otherwise an unrelated bug).


danmoseley:
It's unfortunate to teach more of the parser etc about DFS. Is there any way to keep it unified?

Btw I'm still curious why RegexCharClass today contains the original casing as well as the lower casing in insensitive mode.

veanes:
Regarding the issue of treating character classes in DFA mode in a **unified** way as in the rest of the engine.
It is still a fact that DFA does use and depends on the preprocessing of the RegexCharClass. 
It could _stop the preprocessing_ at the point when the non-DFA engine prepares it for actual matching in Case-Insensitive (CI) mode  -- i.e. when it also incorporates the characters needed to capture enough characters so that it can work in CI mode. For example if the interval is [A-Z] in CI mode it needs no extra info as it will just test 
the input character in lower case [a-z] and will do the right thing for i (depending on culture). But, this is all fine for DFA mode as well except that the **IsSingletonInverse** representation, essentially muddies the difference between not{c} and not{c,C} in this particular case (because of regexes _not having referential transparency_ ... I think due to bad legacy decisions how to treat ?i, note that not{c} in invariant mode adds back C if c is case-sensitive otherwise not). In the non-DFA case this difference is still maintained internally by  essentially keeping original casing as well as the lower casing in ?i mode in singleton sets essentially to recognize the IsSingletonInverse representation (so that when executing it will have the right semantics). 

However, in the DFA mode this information seems no longer accessible, at least it is a very specific internal AST representation issue that should not be relevant (unfortunately there is no simple IsSingletonInverse flag for example), or that the DFA should not be concerned with, so it made more sense not to apply the IsSingletonInverse representation in the first place to avoid the whole problem. 

Essentially, when a regex  (?i:[\0-AC-\uFFFF]) is turned into (?i:[^B]) (but in IsSingletonInverse form ... which is a predicate that can be applied to the string representation of the AST of (?i:[^B])) 
Original input regex  (?i:[^B]) on theother hand  is not in IsSingletonInverse form itself. 
Internally I think the difference is maintained by keeping both {a,A} vs only one {a}. 
In the DFA engine both cases look the same (when looking it as an AST) or so it seems. 
Never new of the IsSingletonInverse special form until the bug surfaced.

DFA uses a pre-generated relation R that is a BDD (but also is as A BDDxBDD relation) that in CI mode, given any predicate P (as a BDD), transforms it into P' = {c | exists x in P such that tolower(c)=tolower(x)}. 
The relation R itself is now culture-specific (3 cases matter, i.e., there are three different Rs, I think I'll optimize this representation a bit more, the three Rs only differ wrt how 'i','I','\u0130','\u0131' are treated --- **no other differences**).

Hopefully the above clarifies a bit. Overall I don't see this as a big issue. The situation is somewhat analogous to turning a loop into an atomic loop... that is correct only in non-DFA mode (so it had to be turned off in DFA mode).
The above text could be added somewhere in the documentation of the code also.

veanes:
Observe that, in the end it is in fact the case that  (?i:[\0-AC-\uFFFF]) is the same as [\0-\uFFFF] because 
b adds B back due to ?i. In other words, (?i:[^B]) acts the same as  [\0-\uFFFF] (or '.' in singleline mode) **if** 
IfSingletonInverse((?i:[^B])) holds**.
However (?i:[^B]) means  [\0-AC-ac-\uFFFF]
... I seriously don't like this semantics ... but it is what it is 

the SingletonInverse((?i:[^@])) is mostly relevant perhaps for CI-independent characters such as 
SingletonInverse((?i:[^\n])), but then again the ?i flag does not matter, so I'm puzzled by why it is there, it just seems to muddy the semantics under CI for the special case of a **complement-of-single-character-set** when it always becomes [\0-\xFFFF] because the "other" character will add back the one being complemented out. ?????

veanes:
Another possibility is to get rid of SigletonInverse, by 
turning (?i:[\0-\u{c-1}\u{c+1}C-\uFFFF]) into
1) [\0-\uFFFF] when c is case-sensitive (in the given culture), else into 
2) [^c] (just normal singleton-complement in the AST, which is not the same as SigletonInverse) (here without ?i) 
This would elimate the need to do the above fix in the first place.

veanes:
I checked that all the build failures are due to failure of
System.Net.Primitives.Unit.Tests.CookieContainerTest.SetCookies_Success

(So unrelated to this PR)

veanes:
The last commit derives specific culture case-insensitivity from the default table, instead of constructing it from scratch. This removed the need to store the other large strings for representing case-insensitivity under invariant culture and under Turkish alphabet cultures. It makes the logic more streamlined and **explicit what the exact differences are**, although added some code to do the derivation from the default table that uses BDD operations.

The SRM culture unit test, still tests the correctness of this change, that I also validated separately (namely that the produced tables are identical to the ones that were earlier precomputed).

Observe! It is all about special/different handling of the letters 'i', 'I', '\u0130' and '\u0131' under the different culture groups 
1) "" (invariant culture), 
2) "tr" and 7 other Turkish alphabet cultures, 
3) "en-US" (and all other cultures). 
The case-invariance rules are exactly the same for all the other Unicode letters across all supported 813 cultures.

olsaarik:
If I understood correctly, the two ways discussed to fix this bug without getting rid of the singleton inverse form were:

1. Pass the `isDFA` boolean into the parser and use that to skip the IsSingletonInverse transformation for DFA mode.
2. Make DFA mode aware of IsSingletonInverse.

I do agree with @danmoseley that passing in the `isDFA` flag into the parser is a bit unfortunate. Ideally the frontend would be mostly independent of what backend is going to get used with it. However, duplicating logic to recognize IsSingletonInverse in DFA sounds dangerous too as a potential source of future confusion in case the parser gets modified without full understanding of all of its dependencies.

@veanes would it be possible for the DFA mode to use [this IsSingletonInverse function](https://github.com/dotnet/runtime/blob/ab7492c4a77f0315d384b4c7648d0cefa36b18d1/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCharClass.cs#L621)? It is a public static method of `RegexCharClass`, so that should be fine, but does it do the right thing for the DFA mode?

veanes:
Not exactly sure about it,  the IsSingletonInverse function applies to the string output, I can try to reproduce the string (I hope it is the ToString of the AST node, not sure). But I think the whole IsSingletonInverse concept itself is unnecessary to begin with.

veanes:
I added a bit more unit tests but kept the changes.

I may have misunderstood IsSingletonInverse a bit, it is not the culprit, it is in general just a predicate that says that the set represents complement of a single char, it is independent of ignore-case ... it is unclear how the engine recognizes the difference between the expressions (?i:[\0-AC-\uFFFF]) and (?i:[^B]) later after it has transformed (?i:[\0-AC-\uFFFF]) into (?i:[^B]). The culprit is the rule that does the unsound transformation from (?i:[\0-AC-\uFFFF]) into (?i:[^B]).

Well, it transforms **(?i:[\0-AC-\uFFFF]) into (?i:[^B])** and it transforms **(?i:[^B]) into (?i:[^b])**. Not sure what exact rules are used later to know that the input was not (?i:[^b]). What if the character is such that it has two other case-versions, for example, when **c != ToLower(c) AND c!= ToUpper(c)?**

Long story short, the IsSingletonInverse transformation **is unsound under ?i and DFA mode**, and therefore what I do (skip the IsSingleton transformation) is **much safer** than try to mimic and recognize the special (undocumented) representation trick that somehow works in the nonDFA case -- a clue I have is that it uses lower-uppercase representation distinction somehow, but also this is not really enough. DFA is completely agnostic to what characters are --- if lower or upper or etc, it does not care.

I tried to implement a check but it did not work. This was a serious attempt, took the me the whole day. This is where I draw the line. I'm happy if someone else suggests another fix, else I'm done with it. I would like to move on with other changes and not waste more time on this.

veanes:
Just to add a bit more. I tried Olli's suggestion, that I thought first would work, but the fact is that the IsSingletonInverse predicate in isolation is fine actually (it does not implicitly embed that it is a singleton gap between to two intervals -- that I first thought it did and I was wrong). 
**My overall take on this code is that it is very brittle to changes.** 
What makes it brittle also is that there is no predicate for: **IsCaseSensitive(c)**, else the above issue could be avoided and cleaned up as I mentioned above (essentially by trivially eliminating ?i).

veanes:
There is one more important detail, not sure you know. 
IsCaseSensitive(c) != (tolower(c) != toupper(c))
You might ask why?
Well in Regex, there seems to be an **undocumented** rule that
  if tolower(c) = c and tolower(toupper(c)) != c **then c is classified as case-insensitive in regex.**
Let mw know if I'm wrong in some way, the above I found out by experimenting at it applies to 22 characters across all cultures. For example
ToLower(\u00B5)=\u00B5 (micro sign),
ToUpper(\u00B5)=\u039C (Greek capital M letter)
ToLower(\u039C) = \u03BC (Greek mu letter)
but \u00B5 does not match any of the other two characters in (?i...) context.

olsaarik:
Given the complexities involved with the case insensitive canonicalization, I do agree with skipping it for DFA mode (as there's no benefit for DFA mode anyway).

olsaarik (approved):

danmoseley:
@stephentoub  thoughts?

veanes:
Here is one final thought before I merge. 
To me it would make sense that (under any give culture) 
the following simple rule would hold for all characters x:
(tolower(x) matches (?i:x) AND  toupper(x) matches (?i:x))
as well as 
(toupper(tolower(x)) matches (?i:x) AND  tolower(toupper(x)) matches (?i:x))
I thought this was the case initially but learned otherwise.
Perhaps there is a legacy reason why this is not the case.

PR #1259, opened 2021-06-19
Add ntdll entrypoints used by libraries to the direct import list

jkotas:


jkotas:
This won't work. These entrypoints are not in the import libraries.

PR #1264, opened 2021-06-21
Optimize out full EEType in the scanner if they're used for comparison

MichalStrehovsky:
* If a result of a typeof is fed to Type::Equals, ask for a "necessary" EEType as opposed to a "constructed" EEType in the scanner. Necessary EEType doesn't have a vtable and if nothing else asks for a full EEType, we just saved some space.
* Give answers based on whole program view to RyuJIT during compilation phase.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Reflection/Augments/ReflectionAugments.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Reflection/Augments/ReflectionAugments.cs
@@ -91,7 +91,7 @@ namespace Internal.Reflection.Augments
             if (type == typeof(decimal))
                 return TypeCode.Decimal;
 
-            if (eeType == DBNull.Value.EETypePtr)
+            if (type == typeof(DBNull))
                 return TypeCode.DBNull;
 
             return TypeCode.Object;
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/AnalysisBasedMetadataManager.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/AnalysisBasedMetadataManager.cs
@@ -28,15 +28,12 @@ namespace ILCompiler
         private readonly Dictionary<FieldDesc, MetadataCategory> _reflectableFields = new Dictionary<FieldDesc, MetadataCategory>();
         private readonly HashSet<ReflectableCustomAttribute> _reflectableAttributes = new HashSet<ReflectableCustomAttribute>();
 
-        private readonly HashSet<TypeDesc> _ldtokenReferenceableTypes;
-
         public AnalysisBasedMetadataManager(CompilerTypeSystemContext typeSystemContext)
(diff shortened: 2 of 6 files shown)


jkotas:
What would it take to recognize the type comparison pattern so that this optimization just works for all code out there?

MichalStrehovsky:
> What would it take to recognize the type comparison pattern so that this optimization just works for all code out there?

We had that in https://github.com/dotnet/corert/pull/7975. But it was lost as we picked up the IL Linker's IL analyzer.

The ideal solution would be to teach RyuJIT about it instead but I don't know if I want to teach RyuJIT about unconstructed EETypes. We would also need to do the same detection in the scanner. The transformation is not exactly trivial (ldtoken/GetTypeFromHandle/find arguments to op_Equality).

jkotas:
Would it better to restore the change from dotnet/corert#7975? Or is that particularly difficult?

Why would RyuJIT need to know about this? I would expect that the scanner should be enough.

MichalStrehovsky:
> Would it better to restore the change from dotnet/corert#7975? Or is that particularly difficult?

dotnet/corert#7975 worked because we always gave RyuJIT the unconstructed version and relied on something else to put a constructed EEType into the graph later. The reflection method body scanner ran on all the code that RyuJIT saw.

We don't run the IL Linker MethodBodyScanner on all code, only code that does something "interesting". The analysis is more expensive. We also don't run the analysis on instantiated generic method bodies - just the definitions. This would make us miss places that should do an upgrade.

Unless we add another scanner that is cheaper to run and looks at all code that RyuJIT is looking at, we would need to do this in RyuJIT.

jkotas:
Would ILImporter.Scanner.cs work?

We recognize `ldtoken` + `Type.GetTypeFromHandle` sequence in `ILImporter.Scanner.cs` already. Can we extend it to also recognize `Type::op_Equality`? The whole ILImporter.Scanner.cs is basically a workaround that would be ideally folded into RyuJIT.

MichalStrehovsky:
> Can we extend it to also recognize Type::op_Equality?

It will not be pretty. We don't model the execution stack in the scanner and the patterns we had to look at so far are basically "look at immediately preceding or immediately following instruction".

Equality comparison with a typeof might have an arbitrary thing on one side of the comparison that we might need to skip over so that we can connect the ldtoken with op_Equality.

Are you suggesting we start modeling the execution stack within the IL scanner?

jkotas:
I would be ok with recognizing `something == typeof(Program)` and not `typeof(Program) == something`. `something == typeof(Program)` does not require modeling the IL stack, and it is how code like this is written typically.

I know that it is not ideal to be pattern sensitive like this, but it would not be the first place in .NET where we have this problem (e.g. bound checks tend to exhibit similar pattern sensitivity).

MichalStrehovsky:
Yeah, that might be doable then. I'll take a look at that.

MichalStrehovsky:
Yep, looks better this way.

jkotas (approved):
Nice!

PR #1265, opened 2021-06-21
Use NetCoreAppToolCurrent for ilc

jkotas:
- Allows us to dogfood .NET 6, crossgen2 is on the same plan
- Undo custom System.Reflection.Metadata version that is no longer needed


MichalStrehovsky (approved):
LGTM if we can still build and debug the compiler in a non-preview Visual Studio.

jkotas:
> LGTM if we can still build and debug the compiler in a non-preview Visual Studio.

It works fine. You have to have Tool/Options/Environment/Preview Features/Use previews of the .NET SDK checked and the .NET SDK preview installed globally.

PR #1267, opened 2021-06-21
Forward special P/Invoke attributes to the target DllImport method

jkoritzinsky:
If the GeneratedDllImport-attributed method has SuppressGCTransition or UnmanagedCallConv attributes, apply the attributes to the generated DllImport method.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
+++ b/DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
@@ -130,6 +130,7 @@ namespace Microsoft.Interop
                     generatorDiagnostics.ReportConfigurationNotSupported(lcidConversionAttr, nameof(TypeNames.LCIDConversionAttribute));
                 }
 
+                // TODO: These need to be propagaged to the P/Invoke target, which is this does not do correctly.
                 List<AttributeSyntax> additionalAttributes = GenerateSyntaxForForwardedAttributes(suppressGCTransitionAttribute, unmanagedCallConvAttribute);
 
                 // Create the stub.
--- a/DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
+++ b/DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
@@ -32,6 +32,10 @@ namespace Microsoft.Interop
 
             INamedTypeSymbol? lcidConversionAttrType = context.Compilation.GetTypeByMetadataName(TypeNames.LCIDConversionAttribute);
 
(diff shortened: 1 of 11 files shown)


jkoritzinsky:
@AaronRobinsonMSFT looks like Clang warns about unused attributes (like calling conventions on x64), and since Clang uses a pretty standard warning format, MSBuild recognizes it and elevates it to an error since we have WarnAsError set to true.

Is there an easy way to suppress the warning from DNNE? Also, is it worth having integration tests for the calling convention work since we validate that we forwarded the attribute in the unit tests and the calling conventions would only have a difference in x86, which we don't run against?

AaronRobinsonMSFT:
Updated DNNE to avoid setting these values - https://www.nuget.org/packages/DNNE/1.0.22.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -112,8 +130,11 @@ public void Execute(GeneratorExecutionContext context)
                     generatorDiagnostics.ReportConfigurationNotSupported(lcidConversionAttr, nameof(TypeNames.LCIDConversionAttribute));
                 }
 
+                // TODO: These need to be propagaged to the P/Invoke target, which is this does not do correctly.
```
AaronRobinsonMSFT: Why don't these get propagated properly? Spelling and grammar check please.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -127,6 +148,44 @@ public void Initialize(GeneratorInitializationContext context)
             context.RegisterForSyntaxNotifications(() => new SyntaxContextReceiver());
         }
 
+        private List<AttributeSyntax> GenerateSyntaxForForwardedAttributes(AttributeData? suppressGCTransitionAttribute, AttributeData? unmanagedCallConvAttribute)
+        {
+            const string CallConvsField = "CallConvs";
+            // Manually rehydrate the forwarded attributes with fully qualified types so we don't have to worry about any using directives.
+            List<AttributeSyntax> attributes = new List<AttributeSyntax>();
```
AaronRobinsonMSFT: `new()`?

AaronRobinsonMSFT (approved):

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
jkoritzinsky: This is an old todo that I need to remove (I've fixed this)

PR #1271, opened 2021-06-23
Qualify `new` with `nothrow`

MichalStrehovsky:
It's obviously what we want there.


src/coreclr/nativeaot/Runtime/CachedInterfaceDispatch.cpp
```diff
@@ -435,7 +435,7 @@ void ReclaimUnusedInterfaceDispatchCaches()
 // One time initialization of interface dispatch.
 bool InitializeInterfaceDispatch()
 {
-    g_pAllocHeap = new AllocHeap();
+    g_pAllocHeap = new (std::nothrow) AllocHeap();
```
jkotas:
```suggestion
    g_pAllocHeap = new (nothrow) AllocHeap();
```
This is what we are using everywhere else.

jkotas (approved):
Thanks

PR #1276, opened 2021-06-23
NativeAOT-LLVM: Remove dependency on gtTreeID

yowl:
This PR changes the mapping of gen tree nodes to LLVM Values by replacing the use of `gtTreeID` with the `GenTree` structure address.
As `gtTreeId` is degug only, this unblocks creating a release configuration job.


jkotas (approved):
LGTM

Issue #537, opened 2021-01-07
Investigate lazy generic virtual methods


MichalStrehovsky:
The Avalonia app mentioned in https://github.com/dotnet/runtimelab/pull/536 compiles to a 128 MB EXE. Most of this is caused by explosion of generics.

With this line commented out:

https://github.com/dotnet/runtimelab/blob/0ae2d4b9cc24ca399aa19d89190ca6d30d474121/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs#L85

it compiles to 80 MB. So a lot of the bloat is caused by data structures required to satisfy generic virtual method dispatch for all theoretical call targets. 80 MB is not great either, but the project is rooting entire Avalonia and using a fraction of it.

In theory, things should still work with that line commented out as we should be able to fallback to the type loader. In practice, things don't work. We should look at that.

PR #1281, opened 2021-06-26
Prevent infinite virtual generic method recursion

### LINKED ISSUE #1124: Ilc Net5.0 (6.0.0-preview.5.21258.1)
<a href="https://imgbb.com/">iLc</a><br /><a target='_blank' href='https://imgbb.com/'>eye one share download</a><br />
`Dim Test_DataBase As New ConcurrentDictionary(Of String, Long)
        Dim Test_Data As New ConcurrentBag(Of String)
        Dim Test_Ilc As Object = Encoding.UTF8.GetBytes(Join(Test_DataBase.AsParallel.Select(Function(z) z.Key & "?" & z.Value).Concat(Test_Data.AsParallel.Select(Function(z) New FileInfo(z.Substring(z.IndexOf(">"c) + 1).Trim).FullName & "?" & New FileInfo(z.Substring(z.IndexOf(">"c) + 1).Trim).LastWriteTime.Ticks)).ToArray, "*"))`

`Test_Data.AsParallel.Select` - This cause it :)

kant2002:
Closes dotnet/runtime#92329

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
@@ -43,7 +43,7 @@ namespace ILCompiler.DependencyAnalysis
             DependencyList dependencies = null;
 
             context.MetadataManager.GetDependenciesDueToVirtualMethodReflectability(ref dependencies, context, _method);
-            
+
             if (!_method.IsAbstract)
             {
                 MethodDesc instantiatedMethod = _method;
@@ -89,10 +89,11 @@ namespace ILCompiler.DependencyAnalysis
                             return dependencies;
                         }
 
-                        if (!instantiatedMethod.IsGenericDepthGreaterThan(CanonGVMDepthHeuristic_NonCanonDepth))
+                        if (!instantiatedMethod.IsGenericDepthGreaterThan(SpecificCanonGVMDepthHeuristic_NonCanonDepth))
                         {
                             dependencies.Add(context.MethodGenericDictionary(instantiatedMethod), "GVM Dependency - Dictionary");
                         }
+
                         dependencies.Add(context.NativeLayout.TemplateMethodEntry(canonMethodTarget), "GVM Dependency - Template entry");
                         dependencies.Add(context.NativeLayout.TemplateMethodLayout(canonMethodTarget), "GVM Dependency - Template");
                     }
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
@@ -89,7 +89,10 @@ namespace ILCompiler.DependencyAnalysis
                             return dependencies;
                         }
 
-                        dependencies.Add(context.MethodGenericDictionary(instantiatedMethod), "GVM Dependency - Dictionary");
+                        if (!instantiatedMethod.IsGenericDepthGreaterThan(CanonGVMDepthHeuristic_NonCanonDepth))
+                        {
+                            dependencies.Add(context.MethodGenericDictionary(instantiatedMethod), "GVM Dependency - Dictionary");
+                        }
                         dependencies.Add(context.NativeLayout.TemplateMethodEntry(canonMethodTarget), "GVM Dependency - Template entry");
                         dependencies.Add(context.NativeLayout.TemplateMethodLayout(canonMethodTarget), "GVM Dependency - Template");
                     }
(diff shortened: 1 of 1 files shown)


jkotas:
This will replace one problem with a different one. Programs can hit this condition even without infinite generic expansion.

kant2002:
That make sense to me, just want to understand where I made mistake or that's going on in first place.

What I was trying to do:
I thought that it would be nice to somehow shortciruit compiler, so it would produce some error that such and such method does not supported.
1. I guess that this is proper place to break cycle and try to set `new NotImplementedException()`. Unfortunately that break compiler completely and it crashed. Maybe that's right thing to do.
2. Other option which I see is to have after some depth, generate some `not implemented stub` essentially dealing in similar way as with unsupported marshalling. But I do not find a way how to do that 
3. So I decide try that way and in specific case it works. Reasonably nobody is happy with this.

Question. 
Is SupportsUniversalCanon indicate that compiler supports `universal shared code`? I see that it is hardcoded as false. Anything which I can read for my education what's prevent this from implementing?

jkotas:
>  Anything which I can read for my education what's prevent this from implementing?

USG would be pretty complex RyuJIT feature to implement. We are not convinced that it is a good idea for NativeAOT. USG makes things "work", but the code is slow and big - not something one expects from NativeAOT.

In any case, the first step for the infinite generic recursion is to detect it. Once we detect it, we can choose what to do about it (fail with error, succeed with incomplete program, ...).

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
```diff
@@ -82,6 +83,12 @@ public override IEnumerable<DependencyListEntry> GetStaticDependencies(NodeFacto
                             return dependencies;
                         }
 
+                        if (!context.TypeSystemContext.SupportsUniversalCanon && instantiatedMethod.IsGenericDepthGreaterThan(SpecificCanonGVMDepthHeuristic_NonCanonDepth))
+                        {
+                            // Do not support too deep instantiation.
+                            return dependencies;
+                        }
+
                         dependencies.Add(context.MethodGenericDictionary(instantiatedMethod), "GVM Dependency - Dictionary");
```
MichalStrehovsky: Doing this instead would be an acceptable thing to try:

```suggestion
                        if (!instantiatedMethod.IsGenericDepthGreaterThan(CanonGVMDepthHeuristic_NonCanonDepth))
                        {
                            dependencies.Add(context.MethodGenericDictionary(instantiatedMethod), "GVM Dependency - Dictionary");
                        }
```

I would set CanonGVMDepthHeuristic_NonCanonDepth to e.g. 256.

This would still fix #1124.

We should be able to build the generic dictionary at runtime and most of the time we actually can - there are some rough edges tracked in #537. Setting this number high should mitigate that bug (real world scenarios will unlikely be so deep, and even if they are, chances are decent things will still work because we can compute these most of the time).

hez2010: Can `CanonGVMDepthHeuristic_NonCanonDepth` be made configurable in csproj?

MichalStrehovsky: I imagine once we fix #537, we would want to not do the MethodGenericDictionary line whenever we're optimizing for size in addition to this "things are getting too deep" heuristic - it's exactly the kind of thing we can get rid of when size is more important than startup.

I don't see much value in being able to exactly specify the value beyond that. What would you use it for?

kant2002: My question would be, do we need two `CanonGVMDepthHeuristic_NonCanonDepth` ? One for Universal, and one for Specific like I do, or I otherthink things and only 2 depth parameters needed?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
```diff
@@ -82,7 +83,17 @@ public override IEnumerable<DependencyListEntry> GetStaticDependencies(NodeFacto
                             return dependencies;
                         }
 
-                        dependencies.Add(context.MethodGenericDictionary(instantiatedMethod), "GVM Dependency - Dictionary");
+                        if (!context.TypeSystemContext.SupportsUniversalCanon && instantiatedMethod.IsGenericDepthGreaterThan(SpecificCanonGVMDepthHeuristic_NonCanonDepth))
+                        {
+                            // Do not support too deep instantiation.
+                            return dependencies;
+                        }
+
```
MichalStrehovsky:
```suggestion
```

MichalStrehovsky (approved):
Thanks!

Issue #1266, opened 2021-06-21
Question: NativeAOT-LLVM:  Argument order in GenTreeCall.Operands()


yowl:
I naively thought that the `Operands` method on `GenTreeCall` would be in the IL order but looking at the IR for https://github.com/dotnet/runtimelab/blob/a21e41e80ec07dcea7b18a6b1b8c90220f54af31/src/libraries/System.Private.CoreLib/src/System/OutOfMemoryException.cs#L15-L21
```
------------ BB01 [000..01A) (return), preds={} succs={}
               [000016] ------------                 IL_OFFSET void   IL offset: 0x0
N002 ( 14,  2) [000001] --CXG-------         t1 =    CALL      ref    System.SR.get_Arg_OutOfMemoryException
                                                  /--*  t1     ref
N004 ( 18,  5) [000011] DA-XG-----L-              *  STORE_LCL_VAR ref    V02 tmp1
N005 (  3,  2) [000012] ------------        t12 =    LCL_VAR   ref    V02 tmp1
N006 (  3,  2) [000000] ------------         t0 =    LCL_VAR   ref    V00 this
                                                  /--*  t12    ref    arg1 in rdx
                                                  +--*  t0     ref    this in rcx
N007 ( 41, 15) [000002] --CXG-------              *  CALL      void   System.SystemException..ctor
N001 (  1,  1) [000003] ------------                 NO_OP     void
N001 (  1,  1) [000004] ------------                 NO_OP     void
N003 (  3,  2) [000005] ------------         t5 =    LCL_VAR   ref    V00 this
N004 (  1,  4) [000006] ------------         t6 =    CNS_INT   int    -0x7FF8FFF2
                                                  /--*  t5     ref    this in rcx
                                                  +--*  t6     int    arg1 in rdx
N005 ( 18, 11) [000007] --CXG-------              *  CALL      void   System.Exception.set_HResult
N001 (  1,  1) [000008] ------------                 NO_OP     void
N001 (  0,  0) [000009] ------------                 RETURN    void

```
I can see that the 2nd and 3rd `CALL`s have `this` in different positions.  That doesn't matter I suppose when the args are in registers.  If I want them in IL order do I need something like
```
   fgArgTabEntry* curArgTabEntry = gtArgEntryByLateArgIndex(call, lateArgIndex);
  // use curArgTabEntry->argNum
```
Or is there a more correct way?

yowl:
[...]

PR #1286, opened 2021-06-29
NativeAOT-LLVM: sort call operands by argNum

yowl:
This PR sorts the operands by argNum to ensure the call order matches the signature.  This addresses the issue with the current code mentioned https://github.com/dotnet/runtimelab/issues/1266 

I've closed that issue as some actual code to look at is often better.  Apart from the general issue of whether there is a better way to access the operands sorted by arg number, I've used a `std::vector` and `emplace` to perform an insert sort.   This is not the best, but the number of args is typically small, and I don't know if any other strategy would be better.  That's mute of course if there is a way to access them by argNum.  I could `reserve` on the vector a small number, e.g. 5, which may improve the `emplace` performance, but I left that out as sometimes there are no args, and the internal algorithm in `emplace` for creating more capacity may be better anyway.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -519,7 +519,7 @@ void storeOnShadowStack(llvm::IRBuilder<>& builder, GenTree* operand, Value* sha
 }
 
 llvm::Value* buildUserFuncCall(GenTreeCall* call, llvm::IRBuilder<>& builder)
-    {
+{
     const char* symbolName = (*_getMangledSymbolName)(_thisPtr, call->gtEntryPoint.handle);
     if (_isRuntimeImport(_thisPtr, call->gtCallMethHnd))
     {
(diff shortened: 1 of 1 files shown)


yowl:
Timed out, restarting

SingleAccretion:
We have to undo what morph so carefully arranged for us because it is in morph that the calling convention is decided, and we have a bespoke calling convention, so I think this approach is fine.

That said, why not allocate a buffer of `GenTree*` `fgArgInfo->argTableSize` in size and then walk the `fgArgInfo->argTable`, collecting nodes (via `GetNode()`) into entries indexed by `tabEntry->argNum`?

yowl:
> That said, why not allocate a buffer of `GenTree*` `fgArgInfo->argTableSize` in size and then walk the `fgArgInfo->argTable`, collecting nodes (via `GetNode()`) into entries indexed by `tabEntry->argNum`?

Thanks!.  That's much neater.

yowl:
I think there is a problem with the argument indexing elsewhere.  I'll have to fix that soon.  The problem at the moment is that there are only around 80 methods compiled so there's little confidence that the existing code is without errors.   I've got a few more PRs lined up which increase the coverage to around 300-400 methods so hopefully they will force the error out.

jkotas (approved):
Thanks

PR #1287, opened 2021-07-01
Merge from dotnet/runtime

jkotas:


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ArrayMapNode.cs
```diff
@@ -57,12 +57,15 @@ public override ObjectData GetData(NodeFactory factory, bool relocsOnly = false)
 
                 var arrayType = (ArrayType)type;
 
+                // This optimization is not compatible with canInlineTypeCheck on JIT/EE interface returning
```
jkotas, line 60: cc @MichalStrehovsky I have done this to make the tests pass. The alternative was to change canInlineTypeCheck, but that would lead to suboptimal code in number of places.

MichalStrehovsky: Makes sense. Thank you!

Miizukii:
[Remove some redundant / unused binplace directories (#54890)](https://github.com/dotnet/runtimelab/pull/1287/commits/f399076cb971fcfdab397d8b3786f321acb4b4d5) breaks the repro project, since `compile-with-Release-libs.rsp` is using `artifacts\bin\pkg\net6.0\runtime\windows-Release-x64` which has been removed.

PR #1289, opened 2021-07-01
Don't emit SkipLocalsInit when it's already active

jkoritzinsky:
If SkipLocalsInit is already applied on the module, type, or method level, do not add it when we emit the stub.

### DIFF
--- a/DllImportGenerator/DllImportGenerator/DllImportStub.cs
+++ b/DllImportGenerator/DllImportGenerator/DllImportStub.cs
@@ -16,7 +16,8 @@ namespace Microsoft.Interop
         Compilation Compilation,
         bool SupportedTargetFramework,
         Version TargetFrameworkVersion,
-        AnalyzerConfigOptions Options);
+        AnalyzerConfigOptions Options,
+        bool ModuleSkipLocalsInit);
 
     internal class DllImportStub
     {
@@ -218,7 +219,7 @@ namespace Microsoft.Interop
             var additionalAttrs = new List<AttributeListSyntax>();
 
             // Define additional attributes for the stub definition.
-            if (env.TargetFrameworkVersion >= new Version(5, 0) && !IsSkipLocalsInit(method))
+            if (env.TargetFrameworkVersion >= new Version(5, 0) && !MethodIsSkipLocalsInit(env, method))
             {
                 additionalAttrs.Add(
                     AttributeList(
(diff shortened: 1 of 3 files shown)


DllImportGenerator/DllImportGenerator/DllImportStub.cs
```diff
@@ -240,5 +241,31 @@ public class GeneratedDllImportData
...
+
+            for (INamedTypeSymbol type = method.ContainingType; type is not null; type = type.ContainingType)
+            {
+                if (type.GetAttributes().Any(a => IsSkipLocalsInitAttribute(a)))
+                {
+                    return true;
+                }
+            }
+
+            if (method.ContainingModule.GetAttributes().Any(a => IsSkipLocalsInitAttribute(a)))
```
AaronRobinsonMSFT: I would check the module first.

DllImportGenerator/DllImportGenerator/DllImportStub.cs
```diff
@@ -217,7 +218,7 @@ public class GeneratedDllImportData
             var additionalAttrs = new List<AttributeListSyntax>();
 
             // Define additional attributes for the stub definition.
-            if (env.TargetFrameworkVersion >= new Version(5, 0))
+            if (env.TargetFrameworkVersion >= new Version(5, 0) && !IsSkipLocalsInit(method))
```
AaronRobinsonMSFT: Do we still need the target framework version check?

DllImportGenerator/DllImportGenerator/DllImportStub.cs
```diff
@@ -240,5 +241,31 @@ public class GeneratedDllImportData
                 AdditionalAttributes = additionalAttrs.ToArray(),
             };
         }
+
+        private static bool IsSkipLocalsInit(IMethodSymbol method)
```
AaronRobinsonMSFT: This checks really seems expensive to be honest. Is there a way we can optimize for the module level case?

DllImportGenerator/DllImportGenerator/DllImportStub.cs
jkoritzinsky: We still have our tests around validating against downlevel usage, so I'd like to keep the target framework version check until we drop that.

DllImportGenerator/DllImportGenerator/DllImportStub.cs
jkoritzinsky: I'll go precalculate the module case.

AaronRobinsonMSFT (approved):

PR #1290, opened 2021-07-01
updated reversal of regexes and simplified DFA state repr

veanes:
Simplified state representation by eliminating the reverse flag. 
Instead, reversing anchors systematically, as in the paper that documents the algorithm:
rev(^)=$, rev($)=^, ..., rev(\Z)=\a
For that had to introduce rev(\Z)=\a anchor (matching start or after very first \n). 
This anchor happens e.g. in rev("abc\Z") = "\acba"

Generalized ReplaceStartAnchorByBottom() to PruneAnchors(..) that also takes 
care of other common situations where anchors become infeasible. The pruning is
applied every time a state is constructed and thus needs no special handling any more.


veanes:
The last commit eliminated the restriction that in DFA mode the regex was not allowed to have an empty match or an empty input. This commit removed these special checks and streamlined the code with non-DFA code.
Updated SRM_ATT unit tests accordingly and added some more unit tests for SRM to test the new 
allowed cases.
Also cleaned up some code (In particular: deleted unused code that was under USAFE flag.)

veanes:
The build failures are unrelated to the PR

veanes:
The build failures are unrelated. (again the failing System.Net.Primitives.Unit.Tests.CookieContainerTest)

veanes:
Decided to push the last commit on top of this PR. 
The commit regarding Att and Mono tests is somewhat related to the updates because it 
simplified all of them with a very clean separation of concerns, essentially, the only tests not supported would be ones that use non-supported features that are now cleanly separated. 
(rather than also having to consider regexes that are nullable/nonnulllable)
In the Att case a handful of tests also give a different Group[0] value, which is expected in those cases because in DFA mode the order of choices does not matter, while is non-DFA mode it does matter, e.g. regex a|ab vs. ab|a. 
These are now encoded as part of the Att and Mono test sets. The earlier SRM specific copies of these tests are hereby deleted.

veanes:
Added Dan as a reviewer also, mainly due to the final commit regarding Att and Mono tests.

olsaarik (approved):

Issue #833, opened 2021-03-18
Support for ARM32 in NativeAOT


bjornen77:
Is there any plans to support ARM32? Basically so that you from a PC running Linux(amd64) could do:
dotnet publish -r linux-arm -c release

Or could you already today compile applications for ARM32 using NativeAOT?

jkotas:
The core team working on the NativeAOT experiment in this repo does not have plans for it, but we are happy to accept contributions to make it work.

josephmoresena:
I recall CoreRT had support on ARM but I never used it. I was actually looking for that cause I want to experiment with JNI libs but using Native NOT.

jkotas:
Yes, we have received some contributions for ARM support in the past but it did not get pushed far enough to get official builds enabled, etc.

michaldobrodenka:
Would be great if at least this could be the way for NET to support ARMv6. Compile to llvm -> ARMv6 or ARMv7 native.

trampster:
Arm is potentially a better use case for NativeAOT than x86. Arm devices tend to be used in resource constrained environments like IOT devices, raspberry pi's, etc.  And can benefit more from the advantages of NativeAOT.

The team I work on has an embedded linux device using OpenEmbedded as the operating system on an arm32 chip. We had to reject .NET as an option because we simply do not have the space for it, and we need fast startup time.

NativeAOT seems like the perfect solution but we can't use it because it doesn't support arm32.

kant2002:
Right now building of NativeAOT on Raspberry 3 (I assume that this is ARM32) fails with following error

```
  CMake Error at nativeaot/Runtime/Full/CMakeLists.txt:25 (add_library):
    Cannot find source file:

      /home/pi/runtimelab/src/coreclr/nativeaot/Runtime/arm/GcProbe.S

    Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
    .hpp .hxx .in .txx


  CMake Error at nativeaot/Runtime/Full/CMakeLists.txt:25 (add_library):
    No SOURCES given to target: Runtime.ServerGC


  CMake Error at nativeaot/Runtime/Full/CMakeLists.txt:23 (add_library):
    No SOURCES given to target: Runtime


  CMake Generate step failed.  Build files cannot be regenerated correctly.
  ~/runtimelab/src/coreclr
```

ChrML:
[...]

PR #1294, opened 2021-07-03
Add basic support for compiling to ARM32

kant2002:
This is what I do to make it compile on Raspbian on my Raspberry Pi 3B+
- GcProbe.S is just copy from ARM64
- __cdecl is not defined on ARM, so I suppress that
- I resurrect RareFlags from CoreRT in hopes that this is make it works, but no luck
I try to resurrect https://github.com/dotnet/corert/commit/d9847aff807f50e3fbd9393742ee7f9a24a21ded in changes to regdisp.h but seems to be I do wrong thing here, and I should resurrect definitions from there.

I would like to understand is RareFlags are still valid way, and if not what's the proper way.

See #833

### DIFF
--- a/src/coreclr/nativeaot/Runtime/PalRedhawk.h
+++ b/src/coreclr/nativeaot/Runtime/PalRedhawk.h
@@ -30,6 +30,25 @@
 #define __PN__MACHINECALL_CDECL_OR_DEFAULT __cdecl
 #endif
 
+#ifndef _MSC_VER
+
+// Note:  Win32-hosted GCC predefines __stdcall and __cdecl, but Unix-
+// hosted GCC does not.
+
+#ifdef __i386__
+
+#if !defined(__cdecl)
+#define __cdecl        __attribute__((cdecl))
+#endif
+
+#else   // !defined(__i386__)
+
+#define __cdecl
+
+#endif  // !defined(__i386__)
+
+#endif // !_MSC_VER
+
 #ifndef _INC_WINDOWS
 //#ifndef DACCESS_COMPILE
 
--- a/src/coreclr/nativeaot/Runtime/RuntimeInstance.cpp
+++ b/src/coreclr/nativeaot/Runtime/RuntimeInstance.cpp
@@ -447,7 +447,7 @@ COOP_PINVOKE_HELPER(PTR_VOID, RhGetRuntimeHelperForType, (EEType * pEEType, int 
     {
     case RuntimeHelperKind::AllocateObject:
 #ifdef HOST_ARM
-        if ((pEEType->get_RareFlags() & EEType::RareFlags::RequiresAlign8Flag) == EEType::RareFlags::RequiresAlign8Flag)
+        if ((pEEType->RequiresAlign8())
         {
             if (pEEType->HasFinalizer())
                 return INDIRECTION(RhpNewFinalizableAlign8);
--- a/src/coreclr/nativeaot/Runtime/arm/GcProbe.S
+++ b/src/coreclr/nativeaot/Runtime/arm/GcProbe.S
@@ -1,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#include <unixasmmacros.inc>
+#include "AsmOffsets.inc"
+
+    .global RhpGcPoll2
+
+    LEAF_ENTRY RhpGcPoll
+        PREPARE_EXTERNAL_VAR_INDIRECT_W RhpTrapThreads, 0
+        cbnz    w0, RhpGcPollRare // TrapThreadsFlags_None = 0
+        ret
+    LEAF_END RhpGcPoll
+
+    NESTED_ENTRY RhpGcPollRare, _TEXT, NoHandler
+        PUSH_COOP_PINVOKE_FRAME x0
+        bl RhpGcPoll2
+        POP_COOP_PINVOKE_FRAME
+        ret
+    NESTED_END RhpGcPollRare
--- a/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/TypeBuilderState.cs
+++ b/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/TypeBuilderState.cs
@@ -1071,7 +1071,7 @@ namespace Internal.Runtime.TypeLoader
 #if TARGET_ARM
                 if (TypeBeingBuilt is DefType)
                 {
-                    return ((DefType)TypeBeingBuilt).IsHfa;
+                    return ((DefType)TypeBeingBuilt).IsHomogeneousAggregate;
                 }
                 else
                 {


jkotas:
What do you need the rare flags for?

src/coreclr/nativeaot/Runtime/PalRedhawk.h
```diff
@@ -647,7 +647,9 @@ inline uint8_t * PalNtCurrentTeb()
 // conditional compilation (upto and including defining an export of functionality that isn't a supported
 // intrinsic on that platform).
 //
-
+#if defined(HOST_ARM)
+#define __cdecl
```
jkotas: Where do the other architectures (x64 or arm64) get the cdecl from?

src/coreclr/inc/regdisp.h
```diff
@@ -155,7 +155,7 @@ inline TADDR GetRegdisplayStackMark(REGDISPLAY *display) {
 
 #elif defined(TARGET_64BIT)
 
-#if defined(TARGET_ARM64)
+#if defined(TARGET_ARM64) || defined(__ARM_HOST)
```
jkotas: Where is `__ARM_HOST` defined?

kant2002: This one I take from this answer https://stackoverflow.com/questions/23934862/what-predefined-macro-can-i-use-to-detect-the-target-architecture-in-clang I suspect that I should use `TARGET_XXX` since this is specific to runtime repo. Do not have time to properly look for the appropriate target.

src/coreclr/nativeaot/Runtime/PalRedhawk.h
kant2002: I assume this is MSVC specific calling convention, and probably this code was never compiled on non-Windows.

kant2002:
This is most puzzling thing for me. I notice that this is missing between CoreRT and NativeAOT. Actual copy start after I hit compilation errors here https://github.com/dotnet/runtimelab/blob/e8ca007febe797f5f8a5303ec21fb15a3a9a3c64/src/coreclr/nativeaot/Runtime/RuntimeInstance.cpp#L449-L459

This was only place where RareFlags was used. Sorry that I forgot to mention that in the original issue description. Whole PR is more question, then actual work.

src/coreclr/nativeaot/Runtime/PalRedhawk.h
kant2002: I assume that mean I cannot find anything which make me think otherwise. I looking for gcc and clang but seems to be it's only MSVC specific

src/coreclr/inc/regdisp.h
kant2002: I think I should use `TARGET_ARM` here

kant2002:
When building using on x64 Linux with command line `./build.sh nativeaot -a arm -rc Debug -lc Release`

error produced
```
~/runtimelab/.dotnet/sdk/6.0.100-preview.4.21255.9/Microsoft.Common.CurrentVersion.targets(4964,5): error MSB3030: Could not copy the file "~/.nuget/packages/runtime.linux-arm.microsoft.dotnet.ilcompiler/6.0.0-preview.5.21222.2/tools/libobjwriter.so" because it was not found. [~/runtimelab/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj]
~/runtimelab/.dotnet/sdk/6.0.100-preview.4.21255.9/Microsoft.Common.CurrentVersion.targets(4964,5): error MSB3030: Could not copy the file "~/runtimelab/artifacts/bin/coreclr/Linux.arm.Debug/libjitinterface_arm.so" because it was not found. [~/runtimelab/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj]```
I assume that `libjitinterface_arm` this is 2 separate issues
- package `runtime.linux-arm.microsoft.dotnet.ilcompiler` does not produced (there nothing I can do now), and
- `libjitinterface_arm` does not built as part of product build. This one I have to look. Need a bit of time to get inside build system.

File: src/coreclr/inc/regdisp.h [hunk 3]
```diff
@@ -155,7 +155,7 @@ inline TADDR GetRegdisplayStackMark(REGDISPLAY *display) {
 
 #elif defined(TARGET_64BIT)
 
-#if defined(TARGET_ARM64)
+#if defined(TARGET_ARM64) || defined(TARGET_ARM)
```
jkotas: This does not look right. `TARGET_64BIT` above should not be defined for `TARGET_ARM`. 32-bit ARM has very different volatile context pointers.

src/coreclr/nativeaot/Runtime/PalRedhawk.h
jkotas: Right, this file is included for Linux x64 or arm64 as well. Why is clang not complaining about this cdecl on existing Linux targets?

jkotas:
> This was only place where RareFlags was used. 

We should rewrite this in C#. The extra indirection is not needed anymore, so this should just directly return the function pointer.

src/coreclr/nativeaot/Runtime/PalRedhawk.h
kant2002: This is what I still cannot get where this is included. 
I found following 3 files where __cdecl is redefined if needed
```
src\coreclr\pal\inc\pal_mstypes.h
src\coreclr\inc\palclr.h
src\coreclr\inc\coredistools.h
```
Last one seems to be not connected to building Runtime/Full, but other two can be somehow included. I try to track down 

I will continue hunt down, since on my Linux arm64 with command `./build.sh nativeaot -a arm64 -rc Debug -lc Release` I pass that point, so somewhere I should change macro conditions.

Side-node: Not sure if this is expected,
```
error MSB3030: Could not copy the file "~/runtimelab/artifacts/bin/coreclr/Linux.arm64.Debug/libjitinterface_arm64.so" because it was not found. [/~/runtimelab/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj]```
```

kant2002:
> We should rewrite this in C#. The extra indirection is not needed anymore, so this should just directly return the function pointer.

@jkotas is this related to this whole function, or to only offending part?

jkotas:
> @jkotas is this related to this whole function, or to only offending part?

The whole function

src/coreclr/nativeaot/Runtime/PalRedhawk.h
jkotas: >  Could not copy the file "~/runtimelab/artifacts/bin/coreclr/Linux.arm64.Debug/libjitinterface_arm64.so"

It is not expected. We do official arm64 builds via cross-compilation on Linux x64 so it may explain why we do not see it.

kant2002: Things become more interesting. It seems that I know the reason why ARM not working.
Difference in compilation flags give me hint to the problem.
ARM
```
clang test.cpp -lstdc++ -fms-extensions -march=armv7-a
```
Literally anything else
```
clang test.cpp -lstdc++ -fms-extensions
```
Irregardless of `-march` presence it produce warning on ARM, and only way to suppress that is to undefine `__cdecl`.
So options are following:
- Should I place __cdecl augmentation in the `src/coreclr/nativeaot/Runtime/PalRedhawk.h` or `src/coreclr/nativeaot/Runtime/PalRedhawkCommon.h`
- Or maybe I can include `src\coreclr\pal\inc\pal_mstypes.h` or `src\coreclr\inc\palclr.h` in `PalRedhawkXXX.h`

jkotas: > Should I place __cdecl augmentation in the src/coreclr/nativeaot/Runtime/PalRedhawk.h or src/coreclr/nativeaot/Runtime/PalRedhawkCommon.h

Yes, that sounds good.

src/coreclr/nativeaot/Runtime/RuntimeInstance.cpp
```diff
@@ -447,7 +447,7 @@ COOP_PINVOKE_HELPER(PTR_VOID, RhGetRuntimeHelperForType, (EEType * pEEType, int
     {
     case RuntimeHelperKind::AllocateObject:
 #ifdef HOST_ARM
-        if ((pEEType->get_RareFlags() & EEType::RareFlags::RequiresAlign8Flag) == EEType::RareFlags::RequiresAlign8Flag)
+        if ((pEEType->RequiresAlign8())
```
kant2002, line 450: This is not working, since method not present. Require reworking in C#.

jkotas (approved):
This looks good to me as is. Is it fine to merge? (The PR is still marked [DRAFT].)

kant2002:
This is still draft. I have other recommendations replace with C# and most
likely this is requirement. That’s another can of worm for me. I can show
want I have because I only barely understand what’s going on yet.

jkotas:
> I have other recommendations replace with C#

It is fine to do that in separate PR. This PR is a good incremental step in its current state. I would be happy to merge it if you drop "DRAFT" from the title.

kant2002:
This is not yet compiling. Given that it was broken, I would not mind, but maybe not what you expect.

kant2002:
But I'm more then happy with that, since I working on update libunwind, so I let it hang until you merge that PR then. And really rework of `RhGetRuntimeHelperForType` would be huge toll on me.

jkotas:
Thank you!

PR #1308, opened 2021-07-09
Update libunwind

kant2002:
Updates from https://github.com/llvm/llvm-project/tree/main/libunwind at f7d8754312bf3d10fb902d94f2bc84e3adf12ba4

Before that I have compilation error
```
  make[3]: *** [nativeaot/Runtime/Full/CMakeFiles/Runtime.ServerGC.dir/build.make:804: nativeaot/Runtime/Full/CMakeFiles/Runtime.ServerGC.dir/__/unix/UnixNativeCodeManager.cpp.o] Error 1
  /home/pi/runtimelab/src/coreclr/nativeaot/libunwind/src/libunwind.cpp:96:61: error: format specifies type 'unsigned int' but the argument has type 'unsigned long' [-Werror,-Wformat]
                         static_cast<void *>(cursor), regNum, (unsigned long)value);
```


src/coreclr/nativeaot/Runtime/unix/UnixContext.cpp
```diff
@@ -365,69 +365,9 @@ bool InitializeUnwindContextAndCursor(REGDISPLAY* regDisplay, unw_cursor_t* curs
     return true;
 }
 
-// Update context pointer for a register from the unw_cursor_t.
-static void GetContextPointer(unw_cursor_t *cursor, unw_context_t *unwContext, int reg, PTR_UIntNative *contextPointer)
-{
-    unw_save_loc_t saveLoc;
```
kant2002, line 371: I suspect of this is dummy move. I probably should use `unw_get_reg` instead of `unw_get_save_loc` to restore registry data.

kant2002: Now I see that WASM also do not have this function. Is this mean that it's gone for good?

kant2002: So appears that new version of libUnwind is also used for WASM, or `unw_get_save_loc` is undesired operation. I remove that defines for symmetry, and if this is desired function, I may have to make that function account for WASM too.

jkotas:
This work should continue in dotnet/runtime repo.

PR #1310, opened 2021-07-09
Another round of changes required to compile on ARM

kant2002:
- libunwind change is fix compilation type mismatch.  Present in upstream https://github.com/llvm/llvm-project/blob/d480f968ad8b56d3ee4a6b6df5532d485b0ad01e/libunwind/src/libunwind.cpp#L102
- gcinfodecoder.cpp This one looks like leftover from other times? Would like to be educated. This location assumes that `src\coreclr\inc\regdisp.h` included, but `src\coreclr\nativeaot\Runtime\regdisplay.h` included. I again do not understand how this happens and why this is working for other targets.

I try search in runtime repo, https://github.com/dotnet/runtime/search?q=volatileCurrContextPointers but results very limited and looks like this variables not used a lot, so this again adds to the puzzle.

### DIFF
--- a/src/coreclr/vm/gcinfodecoder.cpp
+++ b/src/coreclr/vm/gcinfodecoder.cpp
@@ -1486,23 +1486,23 @@ OBJECTREF* GcInfoDecoder::GetRegisterSlot(
     _ASSERTE(regNum >= 0 && regNum <= 14);
     _ASSERTE(regNum != 13);  // sp
 
-    DWORD **ppReg;
+    uint32_t **ppReg;
 
     if(regNum <= 3)
     {
-        ppReg = &pRD->volatileCurrContextPointers.R0;
+        ppReg = &pRD->pR0;
         return (OBJECTREF*)*(ppReg + regNum);
     }
     else if(regNum == 12)
     {
-        return (OBJECTREF*) pRD->volatileCurrContextPointers.R12;
+        return (OBJECTREF*) pRD->pR12;
     }
     else if(regNum == 14)
     {
-        return (OBJECTREF*) pRD->pCurrentContextPointers->Lr;
+        return (OBJECTREF*) pRD->pLr;
     }
 
-    ppReg = &pRD->pCurrentContextPointers->R4;
+    ppReg = &pRD->pR4;
 
     return (OBJECTREF*)*(ppReg + regNum-4);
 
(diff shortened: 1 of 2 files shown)


kant2002:
This is bare minimum until I make https://github.com/dotnet/runtimelab/pull/1308 work.

src/coreclr/vm/gcinfodecoder.cpp
```diff
@@ -1486,23 +1486,23 @@ OBJECTREF* GcInfoDecoder::GetRegisterSlot(
     _ASSERTE(regNum >= 0 && regNum <= 14);
     _ASSERTE(regNum != 13);  // sp
 
-    DWORD **ppReg;
+    uint32_t **ppReg;
```
jkotas: This needs `#ifdef FEATURE_REDHAWK` similar to other `GcInfoDecoder::GetRegisterSlot` in this file

kant2002: How I miss that. Now it make sense!

src/coreclr/vm/gcinfodecoder.cpp
```diff
@@ -1486,6 +1486,11 @@ OBJECTREF* GcInfoDecoder::GetRegisterSlot(
     _ASSERTE(regNum >= 0 && regNum <= 14);
     _ASSERTE(regNum != 13);  // sp
 
+#ifdef FEATURE_REDHAWK
+    PTR_UIntNative* ppReg = &pRD->pR0;
+
```
jkotas: This needs to special case `regNum == 14`

kant2002: I have a question. I do not fully understand your remark. But after looking at the source code at least I think I can ask reasonable question.

1. What is `regNum == 14`? If I take `REGDISPLAY` from `src\coreclr\nativeaot\Runtime\regdisplay.h` this is `SP`.
2. In the assert above written that `_ASSERTE(regNum != 13);  // sp`, but 13 index is `pLR`.
3. Why special case? I take a look at `TARGET_ARM64` and they just look at plan structure. I probably miss some concept here.

kant2002: Looks like I got it. I found this definition and that's from where `regNum` values should be used.
https://github.com/dotnet/runtimelab/blob/dc24a19f55ebe8546bce15db4d208ce2b3053355/src/coreclr/nativeaot/libunwind/include/libunwind.h#L603-L605

jkotas: Right, the REGDISPLAY structure does not have SP in the usual position, and so the code has to compensate for everything above SP. It does not have much to with libunwind.h.

jkotas (approved):
Looks good. Thank you!

PR #1313, opened 2021-07-11
Improved Win Statics debugging

Miizukii:
- Fixed static name mangling to something debugger understands
- Fixed GCStatics incorrect double indirection
- Added helper struct and natvis for thread static data

fixes dotnet/corert#5575

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
@@ -789,7 +789,6 @@ namespace ILCompiler
                         }
                     };
 
-                    _objectWriter.GetClassTypeIndex(classTypeDescriptor);
                     staticFieldRegionTypeIndex = _objectWriter.GetCompleteClassTypeIndex(helperClassTypeDescriptor, helperFieldsDescriptor, helperFields, null);
                     staticFieldRegionSymbolTypeIndex = staticFieldRegionTypeIndex;
                     staticFieldForm = WindowsNodeMangler.ThreadStaticIndexName;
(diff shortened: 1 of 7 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UserDefinedTypeDescriptor.cs
```diff
@@ -750,17 +751,57 @@ private void InsertStaticFieldRegionMember(List<DataFieldDescriptor> fieldDescs,
...
+                        },
+                        new DataFieldDescriptor
+                        {
+                            FieldTypeIndex = GetVariableTypeIndex(defType.Context.GetWellKnownType(Is64Bit? WellKnownType.Int64 : WellKnownType.Int32), true),
+                            Offset = (ulong)NodeFactory.Target.PointerSize,
+                            Name = "ClassIndex"
+                        }
+                    };
+
+                    _objectWriter.GetClassTypeIndex(classTypeDescriptor);
```
MichalStrehovsky: Is this line needed?

MichalStrehovsky (approved):
Looks great, thanks!

Seems like these only show up in WinDBG (I can access them in VS through e.g. `___this.__GCSTATICS`, but they don't show in the locals window in VS). We can work on those details later; I think this is a good direction.

Miizukii:
using watch on the class/struct works best since ___this is sometimes optimized away,
also added natvis intrinsic to keep watch of a thread static fields readable
statics

>but they don't show in the locals window in VS

i have no idea how to tell debugger what fields are relevant for local or whats the current namespace, but if you can point me in the right direction i would take a look

MichalStrehovsky:
> i have no idea how to tell debugger what fields are relevant for local or whats the current namespace, but if you can point me in the right direction i would take a look

I was hoping we could get the same experience like when we're doing managed debugging:

image

But it looks like Visual C++ can't do it (can't get it to show like that for C++).

So this is probably as good as we can make it.

Thank you!

PR #1317, opened 2021-07-13
Do not tailcall from interface instantiation stub

MichalStrehovsky:


jkotas (approved):
I assume that the generated code is same as before with optimizations on. Is that right?

MichalStrehovsky:
> I assume that the generated code is same as before with optimizations on. Is that right?

I didn't check - I assumed we don't want explicit tail prefix for reason specified in #1165. Would it make difference if it's no longer tail optimized?

jkotas:
The implicit tailcall optimization gets sometime derailed by non-obvious factors. I was asking just to make sure that we are not hitting one of those factors by accident.

MichalStrehovsky:
Yes, this is optimized to a tail call in release build. It's no longer a tail call on debug though.

PR #1321, opened 2021-07-14
Analyze GVM dependencies from native layout

MichalStrehovsky:
If we're generating native layout for a potential† GVM callsite, we should ensure the GVM analysis is aware of the callsite and generates necessary templates in template constructable types implementing the GVM.

Contributes to #537 (Avalonia now works with fully dynamic GVM resolution).

†We can't tell between LDTOKEN of a GVM and a GVM call right now and it's probably fine.


jkotas:
> Contributes to #537 

Is the remaining part about reducing number of generic dictionaries saved into the image?

jkotas (approved):
Thanks

MichalStrehovsky:
> Is the remaining part about reducing number of generic dictionaries saved into the image?

Yes, basically exposing an option to not do this line at compile time + any other bug tail this might expose:

https://github.com/dotnet/runtimelab/blob/0ae2d4b9cc24ca399aa19d89190ca6d30d474121/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs#L85

Avalonia is a pretty good stress test, but there might be more to this

PR #1324, opened 2021-07-14
NativeAOT-LLVM: Add support for instance methods

yowl:
This PR adds support to pass the `this` argument into called methods.  This increases the % of methods compiled in the clrjit module for `HelloWasm` to 5.2% and 881 methods.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
@@ -28,6 +28,7 @@ namespace Internal.JitInterface
                         if (_this._compilation.HasLazyStaticConstructor(target))
                         {
                             var nonGcStaticSymbol = _this._compilation.NodeFactory.TypeNonGCStaticsSymbol(target);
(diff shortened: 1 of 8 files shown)


src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -28,6 +28,7 @@ public static void addCodeReloc(IntPtr thisHandle, void* handle)
                         if (_this._compilation.HasLazyStaticConstructor(target))
                         {
                             var nonGcStaticSymbol = _this._compilation.NodeFactory.TypeNonGCStaticsSymbol(target);
+                            _this._codeRelocs.Add(new Relocation(RelocType.IMAGE_REL_BASED_REL32, 0, nonGcStaticSymbol));
```
yowl, line 31: I'm revisiting https://github.com/dotnet/runtimelab/pull/890#issuecomment-813105021 as although the GCStatics symbol was added, the NonGCStatics symbol was added as a symbol, but never made it to a dependency so was never initialised.

jkotas (approved):
Thanks

PR #1328, opened 2021-07-15
Delete ExportForm and related code

jkotas:
Not used in NativeAOT


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/CompilationModuleGroup.Aot.cs
```diff
@@ -30,19 +23,6 @@ partial class CompilationModuleGroup : IInliningPolicy
         /// <summary>
         /// If true, "type" is exported by the set of input assemblies being compiled
         /// </summary>
```
MichalStrehovsky:
```suggestion
```

MichalStrehovsky (approved):

PR #1330, opened 2021-07-16
added unit test for ignore case relation of DFA

veanes:
Testing across all cultures that the only relevant special handling is required for the letters are i, I, and the Turkish i's, \u0130 and \u0131,
This test should catch any changes made to ordinal table of ignoring case of special characters that may depend on culture in the future.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
@@ -93,7 +93,7 @@ namespace System.Text.RegularExpressions.Tests
         /// This test is to make sure that the generated IgnoreCaseRelation table for DFA does not need to be updated.
         /// It would need to be updated/regenerated if this test fails.
         /// </summary>
-        //[OuterLoop("May take several seconds due to large number of cultures tested")]
+        [OuterLoop("May take several seconds due to large number of cultures tested")]
         [Fact]
         public void TestIgnoreCaseRelation()
         {
@@ -137,7 +137,7 @@ namespace System.Text.RegularExpressions.Tests
             }
         }
 
-        //[OuterLoop("May take tens of seconds")]
+        [OuterLoop("May take tens of seconds")]
         [Fact]
         public void TestIgnoreCaseRelationBorderCasesInDFAmode()
         {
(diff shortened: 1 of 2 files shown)


veanes:
marked the unit test as OuterLoop, it runs 7sec on a surface

veanes:
The build failures are unrelated to my changes, they are all due to: Test System.Net.Primitives.Unit.Tests.CookieContainerTest.SetCookies_Success has failed.

veanes:
The new unit test I just added I also marked as OuterLoop, it runs around 20sec on a surface.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -29,6 +29,148 @@ static void WriteLine(string s)
         private const char Turkish_i_withoutDot = '\u0131';
         private const char Kelvin_sign = '\u212A';
 
+        /// <summary>
+        /// Maps each character c to the set of all of its equivalent characters if case is ingored or null if c in case-insensitive
```
danmoseley:
```suggestion
        /// Maps each character c to the set of all of its equivalent characters if case is ignored or null if c in case-insensitive
```

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -29,6 +29,148 @@ static void WriteLine(string s)
...
+
+        /// <summary>
+        /// This test is to make sure that the generated IgnoreCaseRelation table for DFA does not need to be updated.
+        /// It would need to be updated/regenerated if this test fails.
+        /// </summary>
+        [OuterLoop("May take several seconds due to large number of cultures tested")]
+        [Fact]
+        public void TestIgnoreCaseRelation()
+        {
+            //these 22 characters are considered case-insensitive by regex, while they are case-sensitive outside regex
```
danmoseley: nit, our style is a single space after //. it's not a big deal as we can bulk format them as part of cleaning up.

veanes: I fixed some in the last pr, but more of such format changes are needed throughout the whole codebase of srm.

veanes:
I merged this PR, it seemed Dan you were ok with the edits, although you did not approve it explicitly. The remaining build failure: System.Net.Primitives.Unit.Tests.CookieContainerTest.SetCookies_Success
is independent, but could be fixed soon to avoid having to look into the cause of the failure each time because there can also be other legitimate reasons.

danmoseley:
Oops, I got distracted and didn't finish review. Fine.

PR #1332, opened 2021-07-16
Fix more issues with native layout dependencies

MichalStrehovsky:
This is in the same spirit as #1321.

Native layout wasn't properly declaring dependencies for type handles and method generic dictionaries.

The type handles case can be hit with the existing test coverage (need to turn off eager generic dictionary building for GVMs).

The generic dictionary case required a new test (again, can only be hit once we disable eager GVM dictionaries).

These are prerequisites for making GVM dictionaries fully lazy.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GenericMethodsTemplateMap.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GenericMethodsTemplateMap.cs
@@ -51,21 +51,15 @@ namespace ILCompiler.DependencyAnalysis
             nativeSection.Place(hashtable);
 
 
-            foreach (MethodDesc method in factory.MetadataManager.GetCompiledMethods())
+            foreach (var methodEntryNode in factory.MetadataManager.GetTemplateMethodEntries())
             {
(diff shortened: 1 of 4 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GenericMethodsTemplateMap.cs
```diff
@@ -51,29 +51,23 @@ public override ObjectData GetData(NodeFactory factory, bool relocsOnly = false)
             nativeSection.Place(hashtable);
 
 
-            foreach (MethodDesc method in factory.MetadataManager.GetCompiledMethods())
```
MichalStrehovsky, line 54: Going over compiled method bodies was flawed because we have cases where we need native layout for dictionaries of methods that weren't compiled. It has to do with the fact that we compute generic dictionary layouts without considering the effects of inlining - we can end up with generic dictionaries that have dead references to other generic dictionaries of methods that got inlined.

`Unsafe.As` is very famous for being in this category. Can't wait for it to be a RyuJIT intrinsic so that I can intrinsify it in the scanner too.

jkotas (approved):
Thank you!

PR #1335, opened 2021-07-16
Fixes compilation error

### LINKED ISSUE #1334: Compilation error
When on `.NET SDK 6.0.100-preview.7.21359.1`

```
error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.Enumerable.Select<TSource, TResult>(System.Collections.Generic.IEnumerable<TSource>, System.Func<TSource, TResult>)' and 'Internal.CommandLine.Enumerable.Select<T, U>(System.Collections.Generic.IEnumerable<T>, System.Func<T, U>)'
```

kant2002:
after introduction of these functions to LINQ
This is requires bump of expected version in global.json

Closes #1334

### DIFF
--- a/src/coreclr/tools/Common/CommandLine/Enumerable.cs
+++ b/src/coreclr/tools/Common/CommandLine/Enumerable.cs
@@ -18,16 +18,6 @@ namespace Internal.CommandLine
             return Linq.Enumerable.Empty<T>();
         }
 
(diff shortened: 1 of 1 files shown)


jkotas:
This file is shared with CoreCLR. It would be better to submit this fix to dotnet/runtime repo - it has the same problem.

PR #1337, opened 2021-07-16
Fix memory leak caused by ThreadPool.ThreadCount counter

jkotas:
new ThreadLocal<bool>(trackAllValues: true) does not release the storage for dead threads. It meant that
ThreadPool.ThreadCount was counting the number of threads that were ever used by threadpool and it was
leaking memory in bursty workloads where threadpool threads come and go.

Replaced the implemention with more manual counters.


MichalStrehovsky (approved):
Thank you!

PR #1340, opened 2021-07-17
Use shared copy of CustomAttribute family of types

jkotas:


src/libraries/System.Private.CoreLib/src/System/Reflection/CustomAttributeTypedArgument.cs
```diff
@@ -92,5 +92,9 @@ internal string ToString(bool typed)
 
         public Type ArgumentType => m_argumentType;
         public object? Value => m_value;
+
+#if CORERT
+        private static object? CanonicalizeValue(object? value) => (value is Enum e) ? e.GetValue() : value;
```
jkotas: I have PR in dotnet/runtime to add this everywhere

src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/CustomAttributes/NativeFormat/NativeFormatCustomAttributeData.cs
```diff
@@ -189,11 +190,29 @@ internal sealed override IList<CustomAttributeNamedArgument> GetNamedArguments(b
...
-                customAttributeNamedArguments.Add(ReflectionAugments.CreateCustomAttributeNamedArgument(this.AttributeType, memberName, isField, typedValue));
+
+                customAttributeNamedArguments.Add(CreateCustomAttributeNamedArgument(this.AttributeType, memberName, isField, typedValue));
             }
             return customAttributeNamedArguments;
         }
 
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2080:UnrecognizedReflectionPattern",
+            Justification = "Metadata generation ensures fields/properties referenced from attributes are preserved.")]
+        private static CustomAttributeNamedArgument CreateCustomAttributeNamedArgument(Type attributeType, string memberName, bool isField, CustomAttributeTypedArgument typedValue)
```
jkotas, line 201: This will resolve the MemberInfo a bit earlier than before, but I do not think that it matters in practice. It is unlikely that somebody asks for CustomAttributeNamedArgument and won't look at the MemberInfo.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/Attribute.CoreRT.cs
```diff
@@ -0,0 +1,161 @@
...
+
+using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
+using System.Collections.Generic;
+using Internal.LowLevelLinq;
+using Internal.Reflection.Extensions.NonPortable;
+
+namespace System
+{
+    public abstract partial class Attribute
```
jkotas, line 12: This flipped the direction to match CoreCLR: `Attribute` has the actual implementation and `CustomAttributeExtensions` is the wrapper.

src/coreclr/nativeaot/System.Private.Reflection.Core/src/System/Reflection/Runtime/CustomAttributes/NativeFormat/NativeFormatCustomAttributeData.cs
MichalStrehovsky: They can look at `MemberName` instead. It's actually easier to match up that way if I were to crack CustomAttributeNamedArguments myself. But that's probably a perf optimization that should be done in the runtime repo first.

MichalStrehovsky (approved):
Thank you!

PR #1343, opened 2021-07-20
Delete !FEATURE_SYNCTABLE

jkotas:


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Monitor.CoreRT.cs
```diff
@@ -24,11 +24,6 @@ public static partial class Monitor
     {
         #region Object->Lock/Condition mapping
 
-#if !FEATURE_SYNCTABLE
-        private static ConditionalWeakTable<object, Lock> s_lockTable = new ConditionalWeakTable<object, Lock>();
```
jkotas, line 28: This path does not even work anymore. ConditionalWeakTable uses C# lock and so it hits infinite recursion.

MichalStrehovsky (approved):
Thank you!

PR #1344, opened 2021-07-20
added nfa mode to SRM as a fallback when DFA state-space explodes

veanes:
Added a DFA state limit of 10000 that triggers NFA mode. Default DFA state space size is 1024 and this is increased by 1024 each time more states are needed to be stored. When 10000 state limit is reached then states are only added for non-union states. For example a regex (state) (.{10}$|.{7}$) becomes is treated as a union of two states .{10}$ and .{7}$. In NFA mode SRM works with unions that slows down the overall matching.

NFA mode is highly unlikely to happen but avoids running out of memory for large counters and long inputs of millions of characters when each input character may cause a new state.

The change does not affect the overall semantics of SRM.

Also refactored the code a bit so that handling of states and transitions (delta) happens in one place now (in the builder) instead of states in the builder and transitions in the matcher -- that caused the resizing logic of the corresponding arrays to be unintuitive and difficult to work with in particular to ensure thread-safety.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
@@ -236,6 +236,8 @@ namespace System.Text.RegularExpressions.SRM
             }
         }
 
+        internal SymbolicRegexNode<S> ToANF() => throw new NotImplementedException();
+
         /// <summary>
         /// Converts a concatenation into an array,
         /// returns a non-concatenation in a singleton array.
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
@@ -236,8 +236,6 @@ namespace System.Text.RegularExpressions.SRM
             }
         }
 
-        internal SymbolicRegexNode<S> ToANF() => throw new NotImplementedException();
-
         /// <summary>
         /// Converts a concatenation into an array,
         /// returns a non-concatenation in a singleton array.
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
@@ -987,8 +987,9 @@ namespace System.Text.RegularExpressions.SRM
         /// </summary>
         /// <param name="elem">given element wrt which the derivative is taken</param>
         /// <param name="context">immediately surrounding character context that affects nullability of anchors</param>
+        /// <param name="antimirov">if true uses Antimirov derivatives</param>
         /// <returns></returns>
-        internal SymbolicRegexNode<S> MkDerivative(S elem, uint context)
+        internal SymbolicRegexNode<S> MkDerivative(S elem, uint context, bool antimirov)
         {
             if (this == builder.dotStar || this == builder.nothing)
                 return this;
@@ -1005,7 +1006,7 @@ namespace System.Text.RegularExpressions.SRM
                     case SymbolicRegexKind.Loop:
                         {
                             #region d(a, R*) = d(a,R)R*
-                            var step = left.MkDerivative(elem, context);
+                            var step = left.MkDerivative(elem, context, antimirov);
                             if (step == builder.nothing || upper == 0)
                             {
                                 return builder.nothing;
@@ -1038,10 +1039,17 @@ namespace System.Text.RegularExpressions.SRM
                     case SymbolicRegexKind.Concat:
                         {
                             #region d(a, AB) = d(a,A)B | (if A nullable then d(a,B))
-                            var first = builder.MkConcat(left.MkDerivative(elem, context), right);
+                            var leftd = left.MkDerivative(elem, context, antimirov);
+                            var first = builder.nothing;
+                            if (antimirov && leftd.kind == SymbolicRegexKind.Or)
+                                // push concatenations into the union
+                                foreach (var d in leftd.alts)
+                                    first = builder.MkOr(first, builder.MkConcat(d, right));
+                            else
+                                first = builder.MkConcat(leftd, right);
                             if (left.IsNullableFor(context))
                             {
-                                var second = right.MkDerivative(elem, context);
+                                var second = right.MkDerivative(elem, context, antimirov);
                                 var deriv = builder.MkOr2(first, second);
                                 return deriv;
                             }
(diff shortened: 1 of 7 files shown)


stephentoub:
>> NFA mode

> The change does not affect the overall semantics of SRM.

I'm not sure how to rationalize these two statements.  Can this degrade into exponential execution times?

veanes:
It does not degrade to exponential execution. **NFA mode does not mean backtracking**. It simply means that OR-states are being recomputed instead of being  cached. It is a tradeoff between memory and speed. Almost all DFA engines do this as a fallback.
Thus if you have a disjunctive state {p|q} you consider transition from p -> P and q -> Q
and then work again with P u Q (the size of which can in the worst case be linear in the size of  the input).

veanes:
NFA mode may kick in for regexes with counters such as  a.{100} over long inputs. (see unit test that I added).
But in fact there can be no more than in the order of 100 states in this case in an NFA of this regex while in a DFA the full state space is of size 2^{100}.
So the number of states here formally only depends on the size of the regex, not the input. But you many of course have a regex like a.{10000000}

veanes:
The limit 10k states is rather arbitrary. (It could also be 100k.)
It is based on an observation that I've yet to see DFAs in practice that have more than a few thousands states.

stephentoub:
Ok, thanks.

veanes:
The last commit makes sure that in nfa mode we use Antimirov derivatives. 
In particular this means that taking a derivative 
  d_a((A|B)C) = d_a(A)C | d_a(B)C 
instead of 
  d_a((A|B)C) = (d_a(A)|d_a(B))C 
the latter happens otherwise in Brzozowski derivatives (as the default). 
Observe that  d_a(A)C | d_a(B)C  lifts the choices to the top-level that exposes them as different states, while in Brzozowski they remain hidden (which is what you want in DFA mode but not in NFA mode -- it would partially defeat the purpose of the NFA mode).

veanes:
The last commit is to avoid the confusion of using the term NFA (as it is sometimes used to refer to backtracking-based search), because what is really going on here is use of "Antimirov" derivatives, so I changed the name of the flag 'nfamode' to 'antimirov'.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
```diff
@@ -236,6 +236,8 @@ public static void Serialize(SymbolicRegexNode<S> node, System.Text.StringBuilde
             }
         }
 
+        internal SymbolicRegexNode<S> ToANF() => throw new NotImplementedException();
+
```
olsaarik: Is this a leftover piece of code, or something to be filled in?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
```diff
@@ -451,14 +412,38 @@ private State<S> Delta(string input, int i, State<S> q)
...
-            int atom_id = (c == 10 && i == input.Length - 1 && q.Node.info.StartsWithLineAnchor ? atoms.Length : dt.Find(c));
-            int offset = (q.Id << K) | atom_id;
-            var p = delta[offset];
-            if (p == null)
-                //transition atom False means that this is \Z
-                return CreateNewTransition(q, atom_id == atoms.Length ? builder.solver.False : atoms[atom_id], offset);
+            int atom_id = (c == 10 && i == input.Length - 1 && q.StartsWithLineAnchor ? builder.atoms.Length : dt.Find(c));
+            // atom=False represents \Z
+            S atom = atom_id == builder.atoms.Length ? builder.solver.False : builder.atoms[atom_id];
+            if (antimirov && q.Node.Kind == SymbolicRegexKind.Or)
```
olsaarik: We might want to check how this extra branch in our innermost matching loop affects the performance of the non-antimirov mode. If it does, then we could bubble up this decision to outside the matching loop. In C++ we could use a bool template parameter to avoid copying code, but I'm not sure if C# has any way to get a similar effect. So we might need to specialize some code by hand.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
```diff
@@ -469,24 +454,17 @@ private State<S> CreateNewTransition(State<S> q, S atom, int offset)
...
-                        StateLimit = builder.statearray.Length;
-                        Array.Resize(ref delta, StateLimit << K);
-                    }
-                    delta[offset] = p;
+                    // this is the only place in code where the Next method is called in the matcher
+                    p = q.Next(atom, antimirov);
+                    builder.delta[offset] = p;
+                    //switch to antimirov mode if the maximum bound has been reached
+                    if (p.Id == s_STATEMAXBOUND)
+                        antimirov = true;
```
olsaarik: Regarding getting the antimirov check out of the inner loop, throwing an exception here and catching it outside the matching loop to call into an antimirov specific version might be a way to avoid complicating the control flow.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexNode.cs
veanes: I thought I'd use ToAntimirovNormalForm (ToANF) but then instead added the antimirov flag during derivative creation to avoid duplicating work afterwards. So this is never going to be used, will delete this now with a commit.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
veanes: I can set the antimirov flag **in the builder** -- I'll do that -- then it will not be passed around as an extra argument at all in the rest of the code and will only be checked when node.kind is Or (I guess an overhead that should be negligible). I agree with c++ -- templates can nicely solve the issue, but throwing an exception and later using a separate loop for Antimirov mode in C# will cause a lot of code duplication because much of the code will be essentially identical.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
veanes: See prior comment, I will move the antimirov flag from the matcher to the builder. In this way it is never passed around as an argument and only used in the derivative code when **OR** is involved, because the builder is shared in all of this. After I do this commit, you could rerun one of the experiments to confirm that there are no perf regressions.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
olsaarik: Ok! I'll rerun one of the microbenchmarks to see how this fares.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/matcher/SymbolicRegexMatcher.cs
olsaarik: Sounds good!

olsaarik (approved):

PR #1345, opened 2021-07-20
Use dependent handle for implicit locks

jkotas:
The current implementation of implicit locks can keep the Lock and the associated event alive
for a long time after the owning object is dead, It can appear as a leak when observing
the performance metrics of the process.

Switch to managing the implicit locks lifetime using dependent handles. It has characteristics
that are much closer to CoreCLR implementation. The sync block index used by CoreCLR implementation
is effectively a dependent handle with extra hardcoded behavior. We may want to switch to the same
plan as CoreCLR eventually, but this change should improve the situation for now.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Lock.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Lock.cs
@@ -8,7 +8,7 @@ using System.Runtime.CompilerServices;
 namespace System.Threading
 {
     [ReflectionBlocked]
-    public sealed class Lock
+    public sealed class Lock : IDisposable
     {
         // The following constants define characteristics of spinning logic in the Lock class
         private const int SpinningNotInitialized = 0;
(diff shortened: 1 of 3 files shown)


jkotas:
Test:
```
using System;
using System.Diagnostics;
using System.Threading;

class Program
{
    static int pending;

    static void Work(object o) { Monitor.TryEnter(o, 1); Interlocked.Decrement(ref pending); }

    static void CreateContendedLocks() {
        for (int i = 0; i < 10000; i++) {
            Interlocked.Increment(ref pending);
            Object o = new object();
            Monitor.Enter(o);
            ThreadPool.QueueUserWorkItem(Work, o);
        }
    }

    {
        for (;;) {
            Console.WriteLine($"HandleCount: {Process.GetCurrentProcess().HandleCount}");
            CreateContendedLocks();
            while (pending != 0) Thread.Sleep(1);
            GC.Collect();
            GC.WaitForPendingFinalizers();
        }
    }
}
```
Before this change: 20,000 handles
With this change: 892 handles

CoreCLR stabilizes at around 300 handles.

MichalStrehovsky (approved):
Looks good to me but maybe - @AntonLapounov - could you have a look as well?

AntonLapounov:
Looks good to me. I do not recall all the details about this code though.

PR #1347, opened 2021-07-21
Include a couple more native layout dependencies

MichalStrehovsky:
In the same spirit as #1321 and #1332.

The concern is that if we ever end up in situations where we only generate canonical code and not a single dictionary that matches it, we might miss artifacts required by native layout at runtime.

The above situation can happen with lazy generic virtual methods (that we would like to enable soon) or with lazy generics in general (nice to have in the future).

I'm finding all of these places by running an aggressive mode where only native artifacts the compiler considers necessary receive template layout. (The compiler currently generates template layout for everything generic, which tends to paper over deficiencies like this.)

I verified this is a no-size-diff change for a basic WebAPI app (modulo reordering things), but it fixes issues for the aggressive mode where previously we would miss artifacts in the middle of type building.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/NodeFactory.NativeLayout.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/NodeFactory.NativeLayout.cs
@@ -212,6 +212,18 @@ namespace ILCompiler.DependencyAnalysis
             // of just necessary. (Which is what the actual templates signatures will ensure)
             public IEnumerable<IDependencyNode> TemplateConstructableTypes(TypeDesc type)
             {
+                // Array types are the only parameterized types that have templates
+                if (type.IsSzArray && !type.IsArrayTypeWithoutGenericInterfaces())
+                {
+                    TypeDesc arrayCanonicalType = type.ConvertToCanonForm(CanonicalFormKind.Specific);
+
+                    // Add a dependency on the template for this type, if the canonical type should be generated into this binary.
+                    if (arrayCanonicalType.IsCanonicalSubtype(CanonicalFormKind.Any) && !_factory.NecessaryTypeSymbol(arrayCanonicalType).RepresentsIndirectionCell)
+                    {
+                        yield return _factory.NativeLayout.TemplateTypeLayout(arrayCanonicalType);
+                    }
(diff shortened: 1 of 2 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/NodeFactory.NativeLayout.cs
```diff
@@ -212,6 +212,18 @@ private void CreateNodeCaches()
...
+                    TypeDesc arrayCanonicalType = type.ConvertToCanonForm(CanonicalFormKind.Specific);
+
+                    // Add a dependency on the template for this type, if the canonical type should be generated into this binary.
+                    if (arrayCanonicalType.IsCanonicalSubtype(CanonicalFormKind.Any) && !_factory.NecessaryTypeSymbol(arrayCanonicalType).RepresentsIndirectionCell)
+                    {
+                        yield return _factory.NativeLayout.TemplateTypeLayout(arrayCanonicalType);
+                    }
+                }
+
                 while (type.IsParameterizedType)
```
jkotas, line 227: This while loop looks a bit suspect. Let's say we have `T[][][]`. Do we need to ensure the templates for all `T[]`, `T[][]` and `T[][][]` ?

jkotas (approved):
Thanks

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/NodeFactory.NativeLayout.cs
MichalStrehovsky: Yeah, maybe we're at risk of missing the `__Canon[]` template here, but that will probably never happen in practice.

Template for `T[]` is `X[]` (where X is the thing T canonicalizes into).

Template for T[][][] is `__Canon[]`, no matter the T (because the element type is a reference type).

MichalStrehovsky: (This whole TemplateConstructableTypes business is probably just someone getting tired of fixing small bugs within the dependencies. I think it creates a lot more templates than really needed. I would like to get rid of it at some point.)

PR #1348, opened 2021-07-21
Fix annotations on custom attribute handling

MichalStrehovsky:
We were lucking out that default constructors on synthetic attributes were kept. Hit this in a more aggressive compilation mode that was stripping them.


jkotas (approved):
Thanks

PR #1350, opened 2021-07-22
Make PropertyValue public

MichalStrehovsky:
ReferenceTypeHelper is a target of reflection. Likely none of that worked at runtime because non-public details of corelib are blocked from reflection.

Noticed this by accident in the "generate template type metadata lazily" mode - I didn't subject the template type metadata to reflection blocking, and I was seeing extra template metadata for this compared to baseline. In the baseline, `ReferenceTypeHelper` is unusable with reflection completely - in the lazy mode, we generated unreachable template type metadata for it (and still blocked the reflection data).

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/TraceLogging/PropertyValue.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/TraceLogging/PropertyValue.cs
@@ -21,7 +21,13 @@ namespace System.Diagnostics.Tracing
     ///
     /// To get the value of a property quickly, use a delegate produced by <see cref="PropertyValue.GetPropertyGetter(PropertyInfo)"/>.
     /// </summary>
-    internal readonly unsafe struct PropertyValue
+#if CORERT
+    [CLSCompliant(false)]
+    public
+#else
+    internal
+#endif
+    readonly unsafe struct PropertyValue
     {
         /// <summary>
         /// Union of well-known value types, to avoid boxing those types.
@@ -206,7 +212,12 @@ namespace System.Diagnostics.Tracing
             return helper.GetPropertyGetter(property);
         }
 
-        private abstract class TypeHelper
+#if CORERT
+        public
+#else
+        private
+#endif
+        abstract class TypeHelper
         {
             public abstract Func<PropertyValue, PropertyValue> GetPropertyGetter(PropertyInfo property);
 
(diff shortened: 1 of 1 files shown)


src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/TraceLogging/PropertyValue.cs
```diff
@@ -21,7 +21,13 @@ namespace System.Diagnostics.Tracing
     ///
     /// To get the value of a property quickly, use a delegate produced by <see cref="PropertyValue.GetPropertyGetter(PropertyInfo)"/>.
     /// </summary>
-    internal readonly unsafe struct PropertyValue
+#if CORERT
+    [CLSCompliant(false)]
+    public
```
jkotas:
```suggestion
    public // On CoreRT, this must be public to prevent it from getting reflection blocked.
```

src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/TraceLogging/PropertyValue.cs
```diff
@@ -206,7 +212,12 @@ public int ScalarLength
             return helper.GetPropertyGetter(property);
         }
 
-        private abstract class TypeHelper
+#if CORERT
+        public
```
jkotas:
```suggestion
        public // On CoreRT, this must be public to prevent it from getting reflection blocked.
```

src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/TraceLogging/PropertyValue.cs
```diff
@@ -220,7 +231,12 @@ protected static Delegate GetGetMethod(PropertyInfo property, Type propertyType)
             }
         }
 
-        private sealed class ReferenceTypeHelper<TContainer> : TypeHelper where TContainer : class?
+#if CORERT
+        public
```
jkotas:
```suggestion
        public // On CoreRT, this must be public to prevent it from getting reflection blocked.
```

jkotas (approved):
I hope we will eventually invent some better plan for these. Looks good to me for now. Thanks!

MichalStrehovsky:
> I hope we will eventually invent some better plan for these. Looks good to me for now. Thanks!

The easiest way to solve it would be to make `--reflectedonly` the default and remove the concept of reflection blocking (well, we might still need to do some blocking to e.g. not show any fields on System.Object).

The other option I've been thinking about is to switch from blocking to "automatic metadata eligibility". User code and public framework surface (outside the `--reflectedonly` mode) would be automatically eligible to get metadata based on method code/artifact presence. Framework internals would only get metadata if they're visible targets of reflection, or transitively needed from visible targets of reflection. Basically, we would put framework internals on the `--reflectedonly` plan and everything else would stay as-is. That _might_ work.

PR #1351, opened 2021-07-22
Don't generate artifacts for reflection-blocked things

MichalStrehovsky:
We can often end up in situations when the compiler decides we need to generate artifacts due to reflection (e.g. all methods on a type including those inherited from `System.Object`).

We would end up generating artifacts that we then reject in the appropriate mapping tables.

Optimized build (where such `ReflectedMethodNodes` would not be added as roots) can mostly mop up after this, but this still results in a cool 100 kB saving on an ASP.NET WebApi app (the extra method added to the graph could have caused new vtable slots to be needed, etc. that we can't get rid of afterwards anymore).


jkotas (approved):
Nice!

PR #1352, opened 2021-07-22
Do not statically build generic dictionaries for generic virtual methods

### LINKED ISSUE #537: Investigate lazy generic virtual methods

MichalStrehovsky:
The generic dictionaries can be computed at runtime at the time of first dispatch.

I'm reasonably confident that this shouldn't cause trouble after the past few days of tiny fixes.

Saves 2.5% in size for the default WebApi template. Saves a lot more for e.g. Avalonia apps that are generic virtual method heavy. Also removes a possible source of big generic explosions seen in e.g. System.Linq.Paralell.

Fixes #537.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GVMDependenciesNode.cs
@@ -21,7 +21,6 @@ namespace ILCompiler.DependencyAnalysis
     {
         private const int UniversalCanonGVMDepthHeuristic_NonCanonDepth = 2;
         private const int UniversalCanonGVMDepthHeuristic_CanonDepth = 2;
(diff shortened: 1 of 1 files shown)


jkotas (approved):
Woo-hoo!

PR #1353, opened 2021-07-23
Feature/native aot: Fix Dwarf location generation for Reg+Reg, Reg+Stk and Stk+Reg type

RalfKornmannEnvision:


src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfGen.cpp
```diff
@@ -368,7 +368,7 @@ static void EmitVarLocation(MCObjectStreamer *Streamer,
       DwarfRegNum  = GetDwarfRegNum(ArchType, VarInfo.loc.vlRegReg.vlrrReg1);
       DwarfRegNum2 = GetDwarfRegNum(ArchType, VarInfo.loc.vlRegReg.vlrrReg2);
 
-      Len = (1 /* DW_OP_reg */ + 1 /* DW_OP_piece */ + 1 /* Reg size */) * 2 + 1;
+      Len = (1 /* DW_OP_reg */ + 1 /* DW_OP_piece */ + 1 /* Reg size */) * 2;
```
jkotas, line 371: Does `ICorDebugInfo::VLT_STK_REG` case have the same problem?

RalfKornmannEnvision: I honestly don't know as it was not generated so far. As I had no way to test it I didn't want to touch it. That said I have the suspicion that it is although off by one.

jkotas: I think it would be better to fix it proactively even if we are not able to hit it.

jkotas (approved):
Thank you!

PR #1358, opened 2021-07-25
NativeAOT-LLVM: Support for non void LLVM methods

yowl:
This PR starts to add support in the RyuJIT module for calling methods where the return value can be returned on the LLVM stack, i.e. through the `ret` instruction.  Just ints handled in this PR.  Also fixes a bug when the only use of the statics symbol was from non-GC fields, then then symbol was not expanded.

Also widen some LLVM int operations, e.g. `ICmpNE ` which returns an `i8` to `i32` at the point of storing the result in a local.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -519,7 +519,7 @@ void storeOnShadowStack(llvm::IRBuilder<>& builder, GenTree* operand, Value* sha
 }
 
 llvm::Value* buildUserFuncCall(GenTreeCall* call, llvm::IRBuilder<>& builder)
-    {
+{
     const char* symbolName = (*_getMangledSymbolName)(_thisPtr, call->gtEntryPoint.handle);
     if (_isRuntimeImport(_thisPtr, call->gtCallMethHnd))
     {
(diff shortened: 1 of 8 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -317,6 +317,7 @@ Type* getLLVMTypeForVarType(var_types type)
     switch (type)
     {
         case var_types::TYP_BOOL:
+            return Type::getInt1Ty(_llvmContext);
```
SingleAccretion: Note that `TYP_BOOL` is a bit of a lie. There is no actual requirement for `TYP_BOOL` variables to have only the lowest bit set (there is some frontend code which assumes this, it will gradually be made not to do that).

It is safest to assume `TYP_BOOL` is the same as `TYP_UBYTE`.

yowl: Thanks have mapped `TYP_BOOL` as suggested.  I thought this would be more involved due to LLVM wanting i1s for its branch instructions but the tests are passing so guess not.

jkotas (approved):
Thanks!

PR #1361, opened 2021-07-26
Fix how we check for slot defining generic virtual methods

MichalStrehovsky:
F# compiler apparently doesn't bother setting the `newslot` flag. While being technically correct, it's not helping the type system resolve things faster...


jkotas (approved):
Thank you!

PR #1362, opened 2021-07-26
Fix assembly on ARM32

kant2002:
When I copy over that file, I do not realize that these 2 functions do not compile properly

So far left following
- Produce libObjWriter.so for ARM
- Rewrite RhGetRuntimeHelperForType in C# as per this discussion https://github.com/dotnet/runtimelab/pull/1294#issuecomment-874430938

### DIFF
--- a/src/coreclr/nativeaot/Runtime/arm/GcProbe.S
+++ b/src/coreclr/nativeaot/Runtime/arm/GcProbe.S
@@ -7,16 +7,16 @@
     .global RhpGcPoll2
 
     LEAF_ENTRY RhpGcPoll
-        ; @todo: I'm assuming it's not OK to trash any register here. If that's not true we can optimize the
-        ; push/pops out of this fast path.
+        // @todo: I'm assuming it's not OK to trash any register here. If that's not true we can optimize the
+        // push/pops out of this fast path.
         push        {r0}
         ldr         r0, =RhpTrapThreads
         ldr         r0, [r0]
         tst         r0, #TrapThreadsFlags_TrapThreads
-        bne         %0
+        bne         LOCAL_LABEL(RhpGcPollRareCall)
         pop         {r0}
         bx          lr
-0
+LOCAL_LABEL(RhpGcPollRareCall):
         pop         {r0}
         b           RhpGcPollRare
     LEAF_END RhpGcPoll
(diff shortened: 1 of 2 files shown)


jkotas:
GcProbe.asm is for Windows. GcProbe.S is for non-Windows.

kant2002:
So far, outstanding items assuming that I write assembly correctly.
- `libobjwriter` built just fine using `./build.sh nativeaot.objwriter`. That's probably expected.
- Rewrite RhGetRuntimeHelperForType in C# (required for completeness, not for hacking)
- I built just `nativeaot` subset, other blocked. See below.
- I now stuck on Nuget issue, since I cannot build `libs` subset on Raspberry, [due to](https://github.com/NuGet/Home/issues/10968) memory limitations. I catch command line which execute restore and request 1.2Gb in total, so maybe I can find a way how to reduce memory pressure. 
- Once I build `libs` I believe I can run `nativeaot.packages` just fine. Building that subset blocked on `libs`.

src/coreclr/nativeaot/Runtime/arm/GcProbe.S
```diff
@@ -7,14 +7,23 @@
     .global RhpGcPoll2
 
     LEAF_ENTRY RhpGcPoll
-        PREPARE_EXTERNAL_VAR_INDIRECT_W RhpTrapThreads, 0
-        cbnz    w0, RhpGcPollRare // TrapThreadsFlags_None = 0
-        ret
+        // @todo: I'm assuming it's not OK to trash any register here. If that's not true we can optimize the
```
jkotas: I do not think that this assumption is correct. It should be fine to trash any volatile registers here.

kant2002: Should I fix that in windows asm too?

src/coreclr/nativeaot/Runtime/arm/GcProbe.S
```diff
@@ -7,14 +7,16 @@
     .global RhpGcPoll2
 
     LEAF_ENTRY RhpGcPoll
-        PREPARE_EXTERNAL_VAR_INDIRECT_W RhpTrapThreads, 0
```
kant2002, line 10: Technically I can create similar macro for ARM

src/coreclr/nativeaot/Runtime/arm/GcProbe.S
jkotas: Yes, it would be nice.

src/coreclr/nativeaot/Runtime/arm/GcProbe.S
jkotas: I do not think that it is necessary.

kant2002:
Thanks

PR #1369, opened 2021-07-30
NativeAOT-LLVM: Add dwarf info to the clrjit module

yowl:
This PR adds DWARF info for line numbers only to the clrjit module.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -370,37 +370,23 @@ Type* getLLVMTypeForVarType(var_types type)
     }
 }
 
-Value* castIfNecessary(llvm::IRBuilder<>& builder, Value* source, Type* targetType)
+Value* castIfNecessary(llvm::IRBuilder<>& builder, Value* source, Type* valueType)
 {
     Type* sourceType = source->getType();
-    if (sourceType == targetType)
+    if (sourceType == valueType)
         return source;
 
     Type::TypeID sourceTypeID = sourceType->getTypeID();
-    Type::TypeID targetTypeId = targetType->getTypeID();
+    Type::TypeID valueTypeKind = valueType->getTypeID();
 
-    if (targetTypeId == Type::TypeID::PointerTyID)
+    if (valueTypeKind == Type::TypeID::PointerTyID)
     {
         switch (sourceTypeID)
         {
             case Type::TypeID::PointerTyID:
-                return builder.CreatePointerCast(source, targetType, "CastPtrToPtr");
+                return builder.CreatePointerCast(source, valueType, "CastPtrToPtr");
             case Type::TypeID::IntegerTyID:
-                return builder.CreateIntToPtr(source, targetType, "CastPtrToInt");
-            default:
-                failFunctionCompilation();
-        }
-    }
-    if(targetTypeId == Type::TypeID::IntegerTyID)
-    {
-        switch (sourceTypeID)
-        {
-            case Type::TypeID::IntegerTyID:
-                if (sourceType->getPrimitiveSizeInBits() > targetType->getPrimitiveSizeInBits())
-                {
-                    // assumes bools in int8/16/32s have the least significant bit set i.e. not just != 0
-                    return builder.CreateTrunc(source, targetType, "truncInt");
-                }
+                return builder.CreateIntToPtr(source, valueType, "CastPtrToInt");
             default:
                 failFunctionCompilation();
         }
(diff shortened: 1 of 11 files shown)


jkotas:
Could you please resolve conflicts?

src/coreclr/jit/llvm.cpp
```diff
@@ -319,6 +356,7 @@ Type* getLLVMTypeForVarType(var_types type)
     switch (type)
     {
         case var_types::TYP_BOOL:
+            return Type::getInt1Ty(_llvmContext);
```
jkotas: Is this change right? I thought you changed bool to map to Int8 in previous PR.

yowl: Thanks for spotting, looks wrong and a merge error on my part.

yowl: put back to i8

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -110,8 +164,26 @@ public void RegisterLlvmCallbacks(IntPtr corInfoPtr, string outputFileName, stri
...
+                &getDocumentFileName,
+                &firstSequencePointLineNumber,
+                &getOffsetLineNumber
+            );
+        }
+
+        public void InitialiseDebugInfo(MethodDesc method, MethodIL methodIL)
+        {
+            MethodIL uninstantiatiedMethodIL = methodIL.GetMethodILDefinition();
+            if (methodIL != uninstantiatiedMethodIL)
```
jkotas: I think `methodCodeNodeNeedingCode.Method` should be the right instantiation and none of this should be needed here. The right place to fix the instantiation is at the point where the dependencies are added to the graph, not here. Mismatch between `methodCodeNodeNeedingCode.Method` and the actual method instantiation that the code is compiled for is just going to cause more problems later.

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -126,6 +126,7 @@ private unsafe void CompileSingleMethod(CorInfoImpl corInfo, LLVMMethodCodeNode
             {
                 var sig = method.Signature;
                 corInfo.RegisterLlvmCallbacks((IntPtr)Unsafe.AsPointer(ref corInfo), _outputFile, Module.Target, Module.DataLayout);
+                    corInfo.InitialiseDebugInfo(method, GetMethodIL(method));
```
jkotas:
```suggestion
                corInfo.InitialiseDebugInfo(method, GetMethodIL(method));
```

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
yowl: Thanks, removed.

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilationBuilder.cs
```diff
@@ -70,6 +70,7 @@ public override ICompilation ToCompilation()
             {
                 case OptimizationMode.None:
                     jitFlagBuilder.Add(CorJitFlag.CORJIT_FLAG_DEBUG_CODE);
+                    jitFlagBuilder.Add(CorJitFlag.CORJIT_FLAG_DEBUG_INFO);
```
jkotas, line 73: This should be based on the command line settings and not tied to the optimization level, e.g.:
```
            // Do not bother with debug information if the debug info provider never gives anything.
            if (!(_debugInformationProvider is NullDebugInformationProvider))
                jitFlagBuilder.Add(CorJitFlag.CORJIT_FLAG_DEBUG_INFO);
```
(This can be fixed in follow-up PR.)

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -1,14 +1,18 @@
 ﻿using System;
+using System.Linq;
```
jkotas, line 2: We typically stay away from Linq since it is a performance trap.

src/coreclr/jit/llvm.cpp
```diff
@@ -909,6 +947,85 @@ void generateProlog()
...
+
+void startImportingNode(llvm::IRBuilder<>& builder)
+{
+    if (_compiler->opts.compDbgInfo)
+    {
+        const char* documentFileName = _getDocumentFileName(_thisPtr);
+
+        if (documentFileName && *documentFileName != '\0')
+        {
+            unsigned int lineNo = _getOffsetLineNumber(_thisPtr, _currentOffset);
```
jkotas, line 1019: This looks like `O(method_size^2)`. It can take a while for larger methods, but I guess it won't be the only quadratic algorithm in LLVM compilation pipeline.

jkotas (approved):
Thank you

src/coreclr/jit/llvm.cpp
yowl: I probably should have done something better than just copying this from the IL module.  I'll get and store it once per method.

yowl:
Thanks for reviewing.  I'll address the comments.

Issue #1364, opened 2021-07-26
Question: NativeAOT-LLVM: Emit Dwarf from clrjit


yowl:
In order to get DWARF instructions written to the LLVM module from the clrjit dll is it `get/setBoundaries` that I should be looking at?  I think targets like x64 write the dwarf out separately at `PublishCode` in `CorInfoImpl` but I think I need to set the locations as I build the LLVM.

MichalStrehovsky:
RyuJIT uses `setBoundaries` to tell us the mapping from native offsets to IL offsets. The compiler uses that to map native offsets to line numbers and passes the info to the object writer.

I don't know if it's convenient to hook into that mechanism.

The other option would be to add an API to the JitInterface that tells RyuJIT the IL to source mapping and pipe that through to LLVM. The debug information is available in `_debugInfo` on the CorInfoImpl.cs side.

yowl:
Thanks, I'm just a bit reluctant to jump straight to extending the API in case I've missed something, but here it looks like the way to go.

PR #1370, opened 2021-08-01
NativeAOT-LLVM: store ref locals on shadow stack

yowl:
This PR pushes reference type locals on to the shadow stack.  The LLVM clrjit supports conservative GC only, so any GC references need to go on the shadow stack.  This change also allows these lclVars to be used across different blocks which is coming in another PR.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -42,25 +42,22 @@ struct SpilledExpressionEntry
     CorInfoType m_CorInfoType;
 };
 
+enum LocalVarLocation
+{
+    LlvmStack,
+    ShadowStack
+};
+
 // local vars that don't need to be on the shadow stack
 struct LocalVar
 {
     Value* llvmValue;
+    LocalVarLocation location;
 
-    LocalVar(Value* llvmValue) : llvmValue(llvmValue) {}
+    LocalVar(LocalVarLocation location, Value* llvmValue) : llvmValue(llvmValue), location(location) {}
     virtual Value* getValue(llvm::IRBuilder<>& builder)
     {
-        return llvmValue;
-    }
-};
-
-// local vars that need to be on the shadow stack
-struct ShadowStackLocalVar : LocalVar
-{
-    ShadowStackLocalVar(Value* llvmAddress) : LocalVar(llvmAddress) {}
-    Value* getValue(llvm::IRBuilder<>& builder) override
-    {
-        return builder.CreateLoad(llvmValue);
+        return location == LocalVarLocation::LlvmStack ? llvmValue : builder.CreateLoad(llvmValue);
     }
 };
 
@@ -82,7 +79,7 @@ Function*                                           _function;
 BlkToLlvmBlkVectorMap*                              _blkToLlvmBlkVectorMap;
 llvm::IRBuilder<>*                                  _builder;
 std::unordered_map<GenTree*, Value*>*               _sdsuMap;
-std::unordered_map<unsigned int, struct LocalVar*>* _localsMap;
+std::unordered_map<unsigned int, LocalVar>*         _localsMap;
 CORINFO_SIG_INFO                                    _sigInfo; // sigInfo of function being compiled
 llvm::IRBuilder<>*                                  _prologBuilder;
 std::vector<SpilledExpressionEntry>                 _spilledExpressions;
@@ -137,25 +134,12 @@ void Llvm::llvmShutdown()
 //    Module.Verify(LLVMVerifierFailureAction.LLVMAbortProcessAction);
 }
 
-void cleanUp()
-{
-    if (_localsMap != nullptr)
-    {
-        // These structs are new'ed to they need to be delete'd
-        for (std::pair<unsigned int, struct LocalVar*> element : *_localsMap)
-        {
-            delete element.second;
-        }
-    }
-}
-
 [[noreturn]] void failFunctionCompilation()
 {
     if (_function != nullptr)
     {
         _function->deleteBody();
     }
-    cleanUp();
     fatal(CORJIT_SKIPPED);
 }
 
@@ -202,15 +186,15 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
     }
 }
 
-// TODO: Seems like this is not a great idea, but when looking at a sigInfo from eeGetMethodSig we have CorInfoType(s) but when looking at lclVars we have LclVarDsc or var_type(s), it would be better if a common
-// denomiator was used, but I couldn't find one so introduced this method to workaround.
+// When looking at a sigInfo from eeGetMethodSig we have CorInfoType(s) but when looking at lclVars we have LclVarDsc or var_type(s), This method exists to allow both to map to LLVM types.
 CorInfoType toCorInfoType(var_types varType)
 {
     switch (varType)
     {
         case var_types::TYP_BYREF:
             return CorInfoType::CORINFO_TYPE_BYREF;
-        case var_types::TYP_LCLBLK: // TODO: what is this in the lvaTable?   Dont think it needs to be on the shadow stack, so map to int.
+        case var_types::TYP_LCLBLK: // TODO: outgoing args space - need to get an example compiling, e.g. https://github.com/dotnet/runtimelab/blob/40f9ff64ae80596bcddcec16a7e1a8f57a0b2cff/src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs#L3492 to see what's
+            // going on.  CORINFO_TYPE_VALUECLASS is a better mapping but if that is mapped as of now, then canStoreTypeOnLlvmStack will fail compilation for most methods.
         case var_types::TYP_INT:
             return CorInfoType::CORINFO_TYPE_INT;
         case var_types::TYP_REF:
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -114,12 +137,25 @@ void Llvm::llvmShutdown()
...
 }
 
+void cleanUp()
+{
+    if (_localsMap != nullptr)
+    {
+        // These structs are new'ed to they need to be delete'd
+        for (std::pair<unsigned int, struct LocalVar*> element : *_localsMap)
+        {
+            delete element.second;
```
SingleAccretion: So, I am a bit puzzled by this `delete`. Usually we do not `delete` things in the Jit, neither do we have destructors. All allocations are arena-based and utilize the compiler's allocator (`compiler->getAllocator`), which has a no-op `free`.

All memory is then released when the function is compiled. Do we have a specific reason why we need (what I am guessing is) STL's `new` instead of compiler's here?

src/coreclr/jit/llvm.cpp
```diff
@@ -166,11 +202,32 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
     }
 }
 
+// TODO: Seems like this is not a great idea, but when looking at a sigInfo from eeGetMethodSig we have CorInfoType(s) but when looking at lclVars we have LclVarDsc or var_type(s), it would be better if a common
+// denomiator was used, but I couldn't find one so introduced this method to workaround.
```
SingleAccretion: The rest of the Jit deals in `var_types` exclusively, and converts to that as soon as possible.

Since we're juggling 3 types of "types" here (var, CorInfo and LLVM's), I think it is fine to have these helper functions to convert between them, but we'll have to be careful about rationalizing some of the finer differences, if and when those come up.

src/coreclr/jit/llvm.cpp
```diff
@@ -166,11 +202,32 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
...
+CorInfoType toCorInfoType(var_types varType)
+{
+    switch (varType)
+    {
+        case var_types::TYP_BYREF:
+            return CorInfoType::CORINFO_TYPE_BYREF;
+        case var_types::TYP_LCLBLK: // TODO: what is this in the lvaTable?   Dont think it needs to be on the shadow stack, so map to int.
```
SingleAccretion: `LCLBLK` is the outgoing args area, aka the bytes in the function's frame that the callee can use as their own. It is only used for that one local and for nothing else (honestly, we should probably just delete it to save on memory in VN's static arrays).

It is a pretty low-level detail of the calling convention whether you'll want to have outgoing args implemented explicitly or not, but if you do, the best match for it would be `CorInfoType::CORINFO_TYPE_VALUECLASS` (aka opague block of bytes).

src/coreclr/jit/llvm.cpp
```diff
@@ -513,15 +567,20 @@ int getTotalParameterOffset(CORINFO_SIG_INFO& sigInfo)
...
+        LclVarDsc tableVar = _compiler->lvaTable[i];
+        if (!tableVar.lvIsParam)
+        {
+            CorInfoType corInfoType = toCorInfoType(tableVar.TypeGet());
+            if (!canStoreTypeOnLlvmStack(corInfoType))
+            {
+                offset = padNextOffset(corInfoType, offset);
+            }
+        }
+    }
```
SingleAccretion: The "modern" way to write this loop would something like this:
```c++
{
}
```
Note that we never work with `LclVarDsc`s themselves, only `LclVarDsc*`, since they're pretty large (100+ bytes).

src/coreclr/jit/llvm.cpp
```diff
@@ -820,7 +878,50 @@ Value* widenIntIfNecessary(llvm::IRBuilder<>& builder, Value* intValue)
...
+    if (canStoreTypeOnLlvmStack(localCorInfoType))
+    {
+        offset = -1;
+    }
+    else
+    {
+        offset = 0;
+        for (unsigned int i = 0; i < lclVar->GetLclNum(); i++)
+        {
+            LclVarDsc tableVar = _compiler->lvaTable[i];
```
SingleAccretion: Ditto with the loop over local vars.

src/coreclr/jit/llvm.cpp
```diff
@@ -786,7 +844,7 @@ Value* localVar(llvm::IRBuilder<>& builder, GenTreeLclVar* lclVar)
                 }
             }
 
-            _localsMap->insert({lclNum, llvmRef});
+            _localsMap->insert({lclNum, new struct LocalVar(llvmRef)});
```
SingleAccretion: I understand we dynamically allocate these because of the desire for polymorphic behavior right now, so I have a question: do you anticipate `LocalVar` getting bigger in size? If not, I would consider making the behavior polymorphic via a descriminator enum (`LocalVarLocation` or something), it would avoid the need for dynamic allocation (well, the map would still allocate dynamically, but there would be one less indirection).

Also, if the dynamic allocation is there to stay, and if you are to use the compiler's allocator, it would look something like this:
```cs
// We have a global overloaded `new`
new (_compiler) LocalVar(llvmRef);
```
But for this every dynamic allocation needs to be done via the compiler's allocator.

Also nit, here and elsewhere: why use `struct` explicitly? I believe C++ should be able to cope with just the type name fine...

src/coreclr/jit/llvm.cpp
yowl: That sounds good, I replace the `new` by using the allocator and I can delete the `delete`.   Will do, thanks.

SingleAccretion: Expanding on the dynamic allocation bit...

Basically, we don't use STL in the Jit, and so we don't allocate dynamically with plain `new`. The benefit from that is we don't need to `delete`, the downside is we don't have the STL classes. It is up to you if you want to diverge from the Jit or not, I can see benefits both ways.

src/coreclr/jit/llvm.cpp
yowl: "why use struct explicitly" oh dear.  You've spotted by 30 year old C experience. :-)

yowl: I don't think it needs to get bigger, this is a bit analogous to https://github.com/dotnet/runtimelab/blob/40f9ff64ae80596bcddcec16a7e1a8f57a0b2cff/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/EvaluationStack.cs#L444 so maybe the `var_type` might be needed.  Hope not.  Will change as suggested.  Thanks

yowl:
Thanks for reviewing, appreciate the help.  I believe all comments are addressed in the latest commit.

yowl:
This or #1364 will probably merge conflict depending on which merges first.  I'll keep an eye out and resolve as necessary.

src/coreclr/jit/llvm.cpp
```diff
@@ -42,6 +42,25 @@ struct SpilledExpressionEntry
     CorInfoType m_CorInfoType;
 };
 
+enum LocalVarLocation
```
```suggestion
enum class LocalVarLocation
```
We prefer `enum class`es over raw `enum`s as they do not put their member into the scope they're defined in.

src/coreclr/jit/llvm.cpp
```diff
@@ -42,6 +42,25 @@ struct SpilledExpressionEntry
...
+};
+
+// local vars that don't need to be on the shadow stack
+struct LocalVar
+{
+    Value* llvmValue;
+    LocalVarLocation location;
+
+    LocalVar(LocalVarLocation location, Value* llvmValue) : llvmValue(llvmValue), location(location) {}
+    virtual Value* getValue(llvm::IRBuilder<>& builder)
```
```suggestion
    Value* getValue(llvm::IRBuilder<>& builder)
```
No need for `virtual` anymore?

src/coreclr/jit/llvm.cpp
```diff
@@ -42,6 +42,25 @@ struct SpilledExpressionEntry
     CorInfoType m_CorInfoType;
 };
 
+enum LocalVarLocation
+{
+    LlvmStack,
+    ShadowStack
+};
+
+// local vars that don't need to be on the shadow stack
```
```suggestion
// Helper for unifying vars that do and don't need to be on the shadow stack.
```

SingleAccretion (approved):
LGTM, modulo minor nits.

I think in the longer term splitting the translation into multiple phases should be considered. Since we're an AOT compiler, and LLVM is not the fastest code generator, I would expect we can get away with it TP-wise.

I think modeling the LLVM backend after the existing backend makes sense here, with something like the following:
```scala
// General lowering required for any and all LLVM targets.
// Here SSA will be restored (rationalizer deletes PHIs now...),
// args morphing undone as needed, explicit null checks inserted, etc.
// This need not be a single pass (I expect the SSA rebuilding will be a separate IR walk).
lowerllvm.cpp
// WASM-specific lowering, in particular rewriting `STORE_LCL_VAR(V00 <on shadow stack>)`
// to `STOREIND` and munging of the calling convention.
lowerllvmwasm.cpp
// Generate LLVM representation using the emitter, insert everything not explicit
// in IR (ordinarily that'd be prologs, epilogs, profiler calls, etc).
codegenllvm.cpp
// Wrapper for data structures that translate IR to LLVM IR (aka wrapper over IRBuilder).
// Ordinarily the emitter would also generate unwind info, exception info, gc info, not
// sure how much of that will apply to this LLVM emitter.
// It is yet to be seen if the codegen-emit separation will prove useful, but I have a feeling it will.
emitllvm.cpp
```
In general the goal would be to have the IR by codegen match LLVM one-to-one as perfectly as is practical.

src/coreclr/jit/llvm.cpp
yowl: thanks, missed that.

yowl:
> I think in the longer term splitting the translation into multiple phases should be considered. Since we're an AOT compiler, and 

Sounds interesting.  I hope you are prepared for a plethora of my ignorant questions :-)  Seriously though, I want to get a PR for `JTRUE` in, look at how Wasm Interface Types might fit in (for now with witx-bindgen) and then I can try this, unless someone else wants to and beats me to it.

SingleAccretion:
> Sounds interesting. I hope you are prepared for a plethora of my ignorant questions :-)

Sure :). I am available at most times on the DotNetEvolution discord.

>  Seriously though, I want to get a PR for JTRUE in, look at how Wasm Interface Types might fit in (for now with witx-bindgen) and then I can try this, unless someone else wants to and beats me to it.

I see.

jkotas:
Could you please resolve the conflict?

yowl:
Done.

jkotas (approved):

PR #1374, opened 2021-08-02
Move to using the new Roslyn IIncrementalGenerator API for better in-VS performance

jkoritzinsky:
Through some local testing, it was determined that using the `ISourceGenerator` V1 API for DllImportGenerator will cause the same VS performance issues as https://github.com/dotnet/runtime/pull/50741. Since the source our generator generates is required at design-time, we cannot take the same resolution as https://github.com/dotnet/runtime/pull/50741 or use a simple non-incrementable direct port to the V2 API with the new "Implementation" source type to avoid impacting design time builds (https://github.com/dotnet/roslyn/pull/54798).

As a result, this PR moves our source generator to a more incremental design.

I've split the pipeline of the generator into a few sections:

- Checking if TFM is supported
- Finding all attributed methods
- Determining the marshalling info for parameters and return values, as well as the `GeneratedDllImport` attribute info
  - From this point onwards, we no longer use `ISymbol`-implementing types. Symbols are not equatable across compilations, so any stages where they are in the output are not cacheable between generator runs.
- Generating syntax for the stub from the marshalling info
- Adding whitespace to the syntax
  - This operation is apparently very expensive, so by caching before we do this, we can skip this expensive step when the stub doesn't change.
- Concatenating all of the stubs into one file.
- Reporting the new file and all diagnostics.

As part of this refactoring, I moved `TypePositionInfo` to use a new type model to represent the managed type of the parameter/return value instead of symbols due to the caching issues mentioned above.

The testing for the incremental support is an ad-hoc testing model that will be replaced when https://github.com/dotnet/roslyn/issues/54832 is approved, implemented, and available.

I have validated with this branch that there is no typing lag when working in the System.Private.CoreLib solution from dotnet/runtime on a Surface Book 1 i7 (my work laptop).


DllImportGenerator/Benchmarks/Benchmarks.csproj
```diff
@@ -16,7 +17,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Ancillary.Interop\Ancillary.Interop.csproj" />
-    <ProjectReference Include="..\DllImportGenerator\DllImportGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+	<ProjectReference Include="..\DllImportGenerator\DllImportGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
AaronRobinsonMSFT:
```suggestion
    <ProjectReference Include="..\DllImportGenerator\DllImportGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```

DllImportGenerator/DllImportGenerator/Comparers.cs
```diff
@@ -0,0 +1,73 @@
...
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+
+namespace Microsoft.Interop
+{
+    internal static class Comparers
+    {
+        public static IEqualityComparer<ImmutableArray<(string, ImmutableArray<Diagnostic>)>> GeneratedSourceSet = new ImmutableArraySequenceEqualComparer<(string, ImmutableArray<Diagnostic>)>(new CustomValueTupleElementComparer<string, ImmutableArray<Diagnostic>>(EqualityComparer<string>.Default, new ImmutableArraySequenceEqualComparer<Diagnostic>(EqualityComparer<Diagnostic>.Default)));
```
AaronRobinsonMSFT: Please document these instances.

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
```diff
@@ -8,11 +9,14 @@
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+using static Microsoft.Interop.StubCodeContext;
 
 namespace Microsoft.Interop
 {
     internal sealed class StubCodeGenerator : StubCodeContext
     {
+        record struct BoundGenerator(TypePositionInfo TypeInfo, IMarshallingGenerator Generator);
```
AaronRobinsonMSFT:
```suggestion
        private record struct BoundGenerator(TypePositionInfo TypeInfo, IMarshallingGenerator Generator);
```

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
```diff
@@ -35,40 +39,61 @@ internal sealed class StubCodeGenerator : StubCodeContext
...
             this.dllImportData = dllImportData;
-            this.paramsTypeInfo = paramsTypeInfo.ToList();
-            this.diagnostics = generatorDiagnostics;
             this.options = options;
 
-            // Get marshallers for parameters
-            this.paramMarshallers = paramsTypeInfo.Select(p => CreateGenerator(p)).ToList();
+            List<BoundGenerator> allMarshallers = new();
+            List<BoundGenerator> paramMarshallers = new();
+            bool foundNativeRetMarshaller = false, foundManagedRetMarshaller = false;
```
AaronRobinsonMSFT:
```suggestion
            bool foundNativeRetMarshaller = false;
            bool foundManagedRetMarshaller = false;
```

DllImportGenerator/DllImportGenerator/StubCodeGenerator.cs
```diff
@@ -35,40 +39,61 @@ internal sealed class StubCodeGenerator : StubCodeContext
...
 
         public StubCodeGenerator(
-            IMethodSymbol stubMethod,
-            DllImportStub.GeneratedDllImportData dllImportData,
-            IEnumerable<TypePositionInfo> paramsTypeInfo,
-            TypePositionInfo retTypeInfo,
-            GeneratorDiagnostics generatorDiagnostics,
-            AnalyzerConfigOptions options)
+            GeneratedDllImportData dllImportData,
+            IEnumerable<TypePositionInfo> elements,
```
AaronRobinsonMSFT: `elements` -> `argTypes`?

DllImportGenerator/DllImportGenerator/GeneratorDiagnostics.cs
```diff
@@ -174,12 +175,9 @@ public class Ids
...
                 description: GetResourceString(nameof(Resources.TargetFrameworkNotSupportedDescription)));
 
-        private readonly GeneratorExecutionContext context;
+        private readonly List<Diagnostic> diagnostics = new List<Diagnostic>();
 
-        public GeneratorDiagnostics(GeneratorExecutionContext context)
-        {
-            this.context = context;
-        }
+        public IReadOnlyList<Diagnostic> Diagnostics => diagnostics;
```
AaronRobinsonMSFT: Just an assumption here, but I imagine the we simply loop through the `Diagnostics` so why not just `IEnumerable<Diagnostic>`?

DllImportGenerator/DllImportGenerator/DllImportStubContext.cs
```diff
@@ -206,41 +187,34 @@ public class GeneratedDllImportData
...
-                returnTypeInfo = managedRetTypeInfo,
-                paramsTypeInfo = paramsTypeInfo,
-                StubTypeNamespace = stubTypeNamespace,
-                StubContainingTypes = containingTypes,
-                StubCode = code,
-                AdditionalAttributes = additionalAttrs.ToArray(),
-            };
+        public override int GetHashCode()
+        {
+            return StubTypeNamespace?.GetHashCode() ?? 0;
```
AaronRobinsonMSFT: I'd like a comment here. There are a lot of `?` - which I dislike greatly - coupled with an implementation of `GetHasCode` makes me leery of what is going on.

AaronRobinsonMSFT: We should also make sure that `StubTypeNamespace` is set in a constructor or better yet compute this earlier. Another thing that makes me a bit confused here is the `DllImportStubContext` now has the same hash code as any other instance with the same namespace? Coupled with the fact that now instances with no associated namespace also have the same hash code? I think we need a bit more context here.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -1,164 +1,183 @@
...
-            // Store a mapping between SyntaxTree and SemanticModel.
-            // SemanticModels cache results and since we could be looking at
-            // method declarations in the same SyntaxTree we want to benefit from
-            // this caching.
-            var syntaxToModel = new Dictionary<SyntaxTree, SemanticModel>();
-
-            var generatorDiagnostics = new GeneratorDiagnostics(context);
+            public override int GetHashCode()
+            {
+                return (StubContext, DllImportData, ForwardedAttributes.Length, Diagnostics.Length).GetHashCode();
```
AaronRobinsonMSFT: Please doc this - why it is needed etc.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -1,164 +1,183 @@
...
+                        context.ReportDiagnostic(
+                            Diagnostic.Create(
+                                GeneratorDiagnostics.TargetFrameworkNotSupported,
+                                Location.None,
+                                MinimumSupportedFrameworkVersion.ToString(2)));
                     }
-                    else if (attr.AttributeClass.ToDisplayString() == TypeNames.GeneratedDllImportAttribute)
+                });
+
+            var stubEnvironment = compilationAndTargetFramework
```
AaronRobinsonMSFT, line 108: This logic needs extensive comments below. What is `Left` and `Right`?

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -1,164 +1,183 @@
...
                     }
-                    else if (lcidConversionAttrType is not null && SymbolEqualityComparer.Default.Equals(attr.AttributeClass, lcidConversionAttrType))
+                )
+                .WithComparer(Comparers.CalculatedContextWithSyntax)
+                .Combine(context.AnalyzerConfigOptionsProvider)
+                .Select(
+                    (data, ct) =>
                     {
-                        lcidConversionAttr = attr;
+                        IncrementalTracker?.RecordExecutedStep(new IncrementalityTracker.ExecutedStepInfo(IncrementalityTracker.StepName.GenerateSingleStub, data));
```
AaronRobinsonMSFT, line 140: I don't see why the `IncrementalTracker` should ever be `null`. Is there a case where we don't need tracking? If so, I would like to see a default/nop implementation.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -293,96 +312,193 @@ private DllImportStub.GeneratedDllImportData ProcessGeneratedDllImportAttribute(
                     default:
                         Debug.Fail($"An unknown member was found on {GeneratedDllImport}");
                         continue;
-                    case nameof(DllImportStub.GeneratedDllImportData.BestFitMapping):
-                        stubDllImportData.BestFitMapping = (bool)namedArg.Value.Value!;
-                        stubDllImportData.IsUserDefined |= DllImportStub.DllImportMember.BestFitMapping;
+                    case nameof(GeneratedDllImportData.BestFitMapping):
+                        stubDllImportData = stubDllImportData with
```
AaronRobinsonMSFT: All this `with` logic is simply continually making copies. What is the win here? I don't see the value as opposed to just updating an instance before it leaves scope. If the language can't express that we shouldn't defer to multiple copies being made.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
```diff
@@ -293,96 +312,193 @@ private DllImportStub.GeneratedDllImportData ProcessGeneratedDllImportAttribute(
...
+                || !syntaxNode.IsKind(SyntaxKind.MethodDeclaration))
+            {
+                return false;
+            }
+
+            var methodSyntax = (MethodDeclarationSyntax)syntaxNode;
+
+            // Verify the method has no generic types or defined implementation
+            // and is marked static and partial.
+            if (!(methodSyntax.TypeParameterList is null)
```
AaronRobinsonMSFT: Do we want the `is not` logic here?

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
jkoritzinsky: The reason for switching to `with` expressions is to allow us to update the instance while using `init` setters to enforce that once the object is constructed, it is immutable.

I can change `GeneratedDllImportData` to be a `record struct` so the copies can be non-allocating and possibly be elided by the JIT completely. Alternatively, I can make an alternative data type that enforces the immutability.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
jkoritzinsky: We only need tracking for the test case. The tracking is additional work that we don't need (and can't check the results of anyway) in IDE and command-line compiler scenarios. Additionally, keeping global state on a generator object is technically against the recommended Roslyn policy, so I want to keep this as test-only if possible.

DllImportGenerator/Benchmarks/Benchmarks.csproj
sharwell: 💡 It seems likely that **.editorconfig** needs to be updated if this is occurring.

DllImportGenerator/DllImportGenerator/Comparers.cs
sharwell:
```suggestion
        public static readonly IEqualityComparer<ImmutableArray<(string, ImmutableArray<Diagnostic>)>> GeneratedSourceSet = new ImmutableArraySequenceEqualComparer<(string, ImmutableArray<Diagnostic>)>(new CustomValueTupleElementComparer<string, ImmutableArray<Diagnostic>>(EqualityComparer<string>.Default, new ImmutableArraySequenceEqualComparer<Diagnostic>(EqualityComparer<Diagnostic>.Default)));
```

DllImportGenerator/DllImportGenerator/DllImportStubContext.cs
sharwell: This is a really common pattern in `GetHashCode`, but you can avoid it by using:

```suggestion
            return EqualityComparer<string>.Default.GetHashCode(StubTypeNamespace);
```

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
AaronRobinsonMSFT: Okay. Please ensure that instance/member is documented as test only.

DllImportGenerator/DllImportGenerator/DllImportStubContext.cs
jkoritzinsky: This override of `GetHashCode` is never used, but the C# compiler requires us to provide an override when `Equals` is overridden in a `record`. To keep the implementation simple, I just picked the simplest member to use in the calculation.

This is the case for all of the `GetHashCode` implementations that you've seen in this PR. They're all provided since they're each compiler-required. If they were actually used, then I'd consider pulling in the `Microsoft.Bcl.HashCode` package to allow us to use a really good hash-code creator. But since they're unused, I just threw together some simple implementations.

AaronRobinsonMSFT: If this is never used, can we instead throw an exception? I'd rather an exception be thrown than have a not intentionally equivalent hash code generated and returned.

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
jkoritzinsky: Left and Right are the names provided for the values returned by the `Combine` method. The `Left` member is the data on the left side of the `Combine` method (the `this` parameter), and `Right` is on the right side of the `Combine` method (the parameter passed directly to the method).

Do you want me to add explicit comments about the types of `Left` and `Right`? I'm a little apprehensive about that since it would get out of date quickly if we change the pipeline at all.

DllImportGenerator/Benchmarks/Benchmarks.csproj
jkoritzinsky: We're going to re-align and unify our style when we move out of runtimelab and into the product repo this will live in (most likely runtime).

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
AaronRobinsonMSFT: This is simply ignorance on my part. I've avoided Linq functions for years due to overhead compared to explicit logic - I am assuming `Combine` is some Linq function? I would appreciate a comment for each independent statement in this function. There are baked in assumptions here about how we efficiently work with Roslyn and I'd like to ensure we state those assumptions or at least describe the thought process for how we got here. This implementation, at least to me, feels like there is some opportunity for the IDE team to improve the contract. Will all source generators implementations need to think about this if they want to hit IDE performance metrics? If so, is the methodology documented or captured in an official example?

jkoritzinsky: The `Select`, `WithComparer`, `Combine`, and `Where` methods here aren't actually LINQ methods. They're provided by Roslyn on the `IncrementalValue(s)Provider` types and are named like the LINQ methods for consistency with the BCL. They're actually the only way to do any sort of manipulation of the provider types to define an incremental generator pipeline.


See the source in Roslyn: 
https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/SourceGeneration/IncrementalValueProvider.cs

https://github.com/dotnet/roslyn/blob/main/src/Compilers/Core/Portable/SourceGeneration/Nodes/ValueSourceExtensions.cs

AaronRobinsonMSFT: >They're provided by Roslyn on the IncrementalValue(s)Provider types and are named like the LINQ methods for consistency with the BCL.

Ah. This is one of those things where it would have been obvious if I loaded it in VS or another IDE. Thanks for letting me know. It is annoying that C# requires an IDE to actually do a good review :-(

DllImportGenerator/DllImportGenerator/DllImportGenerator.cs
AaronRobinsonMSFT: Is `GeneratedDllImportData` a public type? If so, then let's try the `record struct` method. If not, I don't think we should constrain the implementation based on a limitation of the `init` feature. The methodology here can be constrained to our usage and policy without much issue.

jkoritzinsky: At the moment it is, but it likely won't need to be once we finish the extensibility work.

My bigger concern is around ensuring that we don't break assumptions in how the incremental pipeline works. Since `GeneratedDllImportData` instances are passed between the phases of the incremental generator, they are cached by Roslyn between runs. As a result, we can end up incorrectly reusing cached results if we accidentally add code that changes the value of one of these instances. This isn't too bad right now since incremental results are only used in the IDE, but with https://github.com/dotnet/roslyn/pull/55023, this type of bug is much scarier.

I personally don't like relying on implicit policy when compiler enforcement is possible. It usually ends up in someone not noticing a misuse of a type and a hard-to-track-down bug.

I'll rewrite this to use locals internally to remove all of the copies and construct the `GeneratedDllImportData` object at the end.

jkoritzinsky:
@sharwell I can't seem to repro the failures locally. It looks like some sort of infra issue with the CodeAnalysis.Testing package. Can you take a quick look and see if you can figure out what's going on?

DllImportGenerator/DllImportGenerator.UnitTests/TestUtils.cs
```diff
@@ -95,7 +119,7 @@ public static (ReferenceAssemblies, MetadataReference) GetReferenceAssemblies()
                     "net6.0",
                     new PackageIdentity(
                         "Microsoft.NETCore.App.Ref",
-                        "6.0.0-preview.6.21317.4"),
+                        "6.0.0-preview.7.21376.23"),
```
sharwell: 📝 This version doesn't exist right now:
https://www.nuget.org/packages/Microsoft.NETCore.App.Ref/

jkoritzinsky: It exists on the net6 Azure DevOps feed, which I think is where my machine pulled it from...

I guess we'll wait to merge this until Preview 7 comes out and I can update to a release version of P7.

sharwell: https://github.com/dotnet/roslyn-sdk/pull/639 was supposed to enable modified configurations, but if the APIs you need are present in the public release I would recommend using that. There is no need to have the current project build target match the reference assemblies used in tests.

AaronRobinsonMSFT (approved):

PR #1377, opened 2021-08-04
Change how we write to stderr from CoreLib

### LINKED ISSUE #1277: FailFast doesn't fail fast enough
We hit a segfault when ICU is missing on Linux instead of the helpful FailFast message ("Couldn't find a valid ICU package installed on the system.").

We should probably short-circuit Console.Error and do file writes to stderr directly.

```
Thread 1 "bflat" received signal SIGSEGV, Segmentation fault.
0x0000000001a03895 in System_Console_System_Text_EncodingHelper__GetCharset ()
(gdb) bt
#0  0x0000000001a03895 in System_Console_System_Text_EncodingHelper__GetCharset ()
#1  0x0000000001a037a9 in System_Console_System_Text_EncodingHelper__GetEncodingFromCharset ()
#2  0x00000000019fd729 in System_Console_System_ConsolePal__GetConsoleEncoding ()
#3  0x00000000019fc738 in System_Console_System_Console__get_OutputEncoding ()
#4  0x00000000019fc80b in System_Console_System_Console__CreateOutputWriter ()
#5  0x00000000019fca43 in System_Console_System_Console___get_Error_g__EnsureInitialized_27_0 ()
#6  0x00000000018465b0 in S_P_DeveloperExperience_Console_Internal_Runtime_CompilerHelpers_LibraryInitializer___c___InitializeLibrary_b__0_0 ()
#7  0x0000000001108255 in S_P_CoreLib_System_RuntimeExceptionHelpers__FailFast_1 ()
#8  0x00000000011c354d in S_P_CoreLib_System_Globalization_GlobalizationMode_Settings___cctor ()
#9  0x00000000011b38d9 in S_P_CoreLib_System_Runtime_CompilerServices_ClassConstructorRunner__EnsureClassConstructorRun
    ()
#10 0x00000000011b37e9 in S_P_CoreLib_System_Runtime_CompilerServices_ClassConstructorRunner__CheckStaticClassConstructionReturnNonGCStaticBase ()
#11 0x0000000001149393 in S_P_CoreLib_System_Globalization_CultureData__CreateCultureWithInvariantData ()
#12 0x000000000114945a in S_P_CoreLib_System_Globalization_CultureData__get_Invariant ()
#13 0x000000000115dc5c in S_P_CoreLib_System_Globalization_TextInfo___cctor ()
#14 0x00000000011b38d9 in S_P_CoreLib_System_Runtime_CompilerServices_ClassConstructorRunner__EnsureClassConstructorRun
    ()
#15 0x00000000011b37d9 in S_P_CoreLib_System_Runtime_CompilerServices_ClassConstructorRunner__CheckStaticClassConstructionReturnGCStaticBase ()
#16 0x0000000001a0388e in System_Console_System_Text_EncodingHelper__GetCharset ()
#17 0x0000000001a037a9 in System_Console_System_Text_EncodingHelper__GetEncodingFromCharset ()
#18 0x00000000019fd729 in System_Console_System_ConsolePal__GetConsoleEncoding ()
#19 0x00000000019fc738 in System_Console_System_Console__get_OutputEncoding ()
#20 0x00000000019fc80b in System_Console_System_Console__CreateOutputWriter ()
#21 0x00000000019fc9a3 in System_Console_System_Console___get_Out_g__EnsureInitialized_25_0 ()
#22 0x00000000019fc8e0 in System_Console_System_Console__WriteLine_12 ()
#23 0x00000000017f88b2 in bflat_Program__Run ()
#24 0x00000000017fc572 in bflat_Program__Main ()
#25 0x0000000001dbf225 in bflat__Module___StartupCodeMain ()
#26 0x0000000000dec7ae in main ()
```

MichalStrehovsky:
We write to stderr from CoreLib for two reasons:

1. FailFast
2. Unhandled exception (also FailFast in a way)

We're using a roundabout way to use `Console.Error.WriteLine` but that has problems for FailFast because we then don't fail fast enough (#1277).

This inlines enough of writing to stderr that we can get rid of System.Console dependency.

This is complicated on Windows because output can be redirected and then we need to deal with code pages. System.Console uses the encoding of the Console even when we're redirected to a file/pipe. I'm making this go through the same infrastructure. If output is not redirected, I'm just calling into `WriteConsole` directly, without having to mess with encodings.

Fixes #1277.


src/libraries/System.Private.CoreLib/src/Internal/Console.Windows.cs
```diff
@@ -8,16 +8,55 @@ namespace Internal
...
+            else
+            {
+                fixed (char* pChars = s)
+                {
+                    Interop.Kernel32.WriteConsole(handle, (byte*)pChars, s.Length, out _, IntPtr.Zero);
+                }
+            }
+        }
+
+        private static bool IsHandleRedirected(IntPtr handle)
```
jkotas: The CoreCLR unmanaged equivalent (`PrintToStdErr` - used by unmanaged failfast) does not have any of this special handling for redirected vs. non-redirected output.

src/libraries/Common/src/System/Text/OSEncoding.Windows.cs
```diff
@@ -10,6 +10,7 @@ namespace System.Text
     internal sealed class OSEncoding : Encoding
     {
         private readonly int _codePage;
+
```
MichalStrehovsky:
```suggestion
```

jkotas (approved):
Nice!

PR #1380, opened 2021-08-05
Merge from dotnet/runtime

jkotas:


jkotas:
@MichalStrehovsky What is the best way to deal with this one?
```
Generating compatible native code. To optimize for size or speed, visit https://aka.ms/OptimizeCoreRT
Unhandled exception. System.NotSupportedException: Specified method is not supported.
    at ILCompiler.ProcessLinkerXmlBase.ProcessAssemblies() in ILCompiler.Compiler.dll:token 0x60003f6+0x4b
    at ILCompiler.ProcessLinkerXmlBase.ProcessXml() in ILCompiler.Compiler.dll:token 0x60003f3+0x46
...
```

MichalStrehovsky:
> @MichalStrehovsky What is the best way to deal with this one?

Added a workaround. We'll want to reprocess manifest XML embedded in CoreLib for all assemblies. CoreLib is the only assembly allowed to embed this assembly name.

PR #1384, opened 2021-08-05
Don't copy custom attribute from Main to MainMethodWrapper

MichalStrehovsky:
Not sure why that was done. We would copy RequiresUnreferencedCode and crash later when we try to crack it open (this is not an `EcmaMethod`).


jkotas (approved):
Thanks

PR #1385, opened 2021-08-05
Fix compilation of libunwind on ARM

kant2002:
This is last issue which prevent compilation on ARM. 
So far, I have crash when publish, but that's on .NET 6 Preview 5 I hope. ILC not even running yet.

### DIFF
--- a/src/coreclr/nativeaot/libunwind/src/Unwind-EHABI.cpp
+++ b/src/coreclr/nativeaot/libunwind/src/Unwind-EHABI.cpp
@@ -91,9 +91,11 @@ _Unwind_Reason_Code ProcessDescriptors(
       case Descriptor::LU32:
         descriptor = getNextWord(descriptor, &length);
         descriptor = getNextWord(descriptor, &offset);
+        break;
       case Descriptor::LU16:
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/libunwind/src/Unwind-EHABI.cpp
```diff
@@ -91,9 +91,11 @@ _Unwind_Reason_Code ProcessDescriptors(
       case Descriptor::LU32:
         descriptor = getNextWord(descriptor, &length);
         descriptor = getNextWord(descriptor, &offset);
+        break;
```
jkotas, line 94: It would be nice to submit this patch to https://github.com/llvm/llvm-project/blob/main/libunwind/src/Unwind-EHABI.cpp as well (it is where this file came from).

jkotas (approved):
Thanks

kant2002:
Already submitted.

kant2002:
@jkotas  this link probably should be proper answer. https://reviews.llvm.org/D107537 There should be additional work probably in that direction, so send you the link in case this is interesting.

Issue #1372, opened 2021-08-02
Regression after 6.0.0-rc.1.21373.1 (all versions)


Bxaa:
```
Friend Module Distinct 'Enhanced Distinct
    <Extension()>
    Public Function c_Distinct(Of TSource)(Input As IEnumerable(Of TSource), Comparator As PropertyInfo) As IEnumerable(Of TSource)
        Dim e_Comp As Func(Of Object, String) = Function(x) CStr(x?.GetType.GetProperty(Comparator.Name).GetValue(x))
        If Comparator IsNot Nothing Then
            Return Input.AsParallel.GroupBy(Function(x) e_Comp(x)?.ToString).Select(Function(x) x.FirstOrDefault)
        Else Return Input
        End If
    End Function
End Module

internal static class Distinct // Enhanced Distinct
{
    public static IEnumerable<TSource> c_Distinct<TSource>(this IEnumerable<TSource> Input, PropertyInfo Comparator)
    {
        Func<object, string> e_Comp = x => System.Convert.ToString(x?.GetType.GetProperty(Comparator.Name).GetValue(x));
        if (Comparator != null)
            return Input.AsParallel().GroupBy(x => e_Comp(x)?.ToString).Select(x => x.FirstOrDefault());
        else
            return Input;
    }
}

**Reproduce:**
Some IEnumerable(Of FileInfo).c_Distinct(GetType(FileInfo).GetProperties()(4)))

...c_Distinct(typeof(FileInfo).GetProperties()(4))
```
Notice: Properties enum sequence is different in net versions
C#\VB Net5 Win10x64
Kernel unhandled error?
AsParallel? (GetType?)

(6.0.0-rc.1.21373.1 work perfect)

jkotas:
@Bxaa Could you please share exact steps for how to reproduce problem? Also, could you please share the stacktrace of the crash or exception that you see?

I have tried to compile and run the following program, and it works fine for me:
```
using System;
using System.IO;
using System.Reflection;
using System.Collections.Generic;
using System.Linq;

internal static class Distinct // Enhanced Distinct
{
    public static IEnumerable<TSource> c_Distinct<TSource>(this IEnumerable<TSource> Input, PropertyInfo Comparator)
    {
        Func<object, string> e_Comp = x => System.Convert.ToString(x?.GetType().GetProperty(Comparator.Name).GetValue(x));
        if (Comparator != null)
[...]

PR #1386, opened 2021-08-05
Improve GVM lookup failure message in type loader

hez2010:
In favor of https://github.com/dotnet/runtimelab/issues/1372#issuecomment-893239032

Not sure it's in a correct way.


MichalStrehovsky:
We've been avoiding calling into reflection from the type loader. Type loader is a pretty low level component and it can get called into from pretty arbitrary spots. That's why we don't call `i.ToString()` here for example:

https://github.com/dotnet/runtimelab/blob/4424f855c5b80b0bdf0f46c60824891be4367144/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/TypeLoaderEnvironment.GVMResolution.cs#L522

calling into globalization could result in recursions. Same applies when calling into reflection.

There is `GetTypeNameDebug` that should be safe, but we don't include that infrastructure in non-DEBUG builds.

This failfast is a bug that normal users can't do much about.

hez2010:
Thanks for the explanation. I'm closing this PR.

PR #1389, opened 2021-08-06
Include canonical EETypes in GVM analysis

### LINKED ISSUE #1372: Regression after 6.0.0-rc.1.21373.1 (all versions)

MichalStrehovsky:
We might no longer have all the concrete EETypes to do analysis - need to include the canonical EETypes we have in the system. The main part of the change is the move of `InterestingForDynamicDependencyAnalysis` from `ConstructedEETypeNode` to `EETypeNode` and limiting it so that it only applies to Constructed and Canonical EETypes (that's the new if at the beginning).

The rest is just mopping up after the change.

Fixes #1372

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
@@ -24,66 +24,6 @@ namespace ILCompiler.DependencyAnalysis
 
         protected override bool EmitVirtualSlotsAndInterfaces => true;
 
-        public override bool InterestingForDynamicDependencyAnalysis
-        {
-            get
-            {
-                if (_type.IsInterface)
-                    return _type.HasGenericVirtualMethods();
-
-                if (_type.IsDefType)
-                {
-                    // First, check if this type has any GVM that overrides a GVM on a parent type. If that's the case, this makes
-                    // the current type interesting for GVM analysis (i.e. instantiate its overriding GVMs for existing GVMDependenciesNodes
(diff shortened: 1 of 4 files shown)


jkotas (approved):
Thank you!

PR #1392, opened 2021-08-07
Clean up BDD code

olsaarik:
This PR cleans up code in `BDD.cs` belonging to the DFA mode we're developing for `System.Text.RegularExpressions`.

I've followed the coding styles from the [coding guidelines](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md) and formatted the code with the `.editorconfig` in the root directory. The .NET Codeformatter Tool didn't work out of the box for me, IIRC reading the internet it seemed it may have to do with the current `.csproj` format not being supported anymore.

I've also added comments as I saw fit. A good thing for reviewing this might be to look at the more complex functions like `TopologicalSort` and `Serialize` and evaluate if these seem like someone new 10 years from now will have a good time figuring them out.

Please add any comments on style or code comments that I should keep in mind when I go through the rest of the DFA codebase.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -7,7 +7,7 @@ namespace System.Text.RegularExpressions.SRM
 {
     /// <summary>
     /// Represents a Binary Decision Diagram (BDD). Supports multi-terminal BDDs, i.e. ones where the leaves are
-    /// something else the True or False, which are used for representing classifiers.
+    /// something other than True or False, which are used for representing classifiers.
     /// </summary>
     internal sealed class BDD : IComparable
     {
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -130,7 +130,8 @@ namespace System.Text.RegularExpressions.SRM
                     set = set.One;
                 }
                 else
-                { // Otherwise, leaving the bit as 0 gives the smaller bitvector.
+                {
+                    // Otherwise, leaving the bit as 0 gives the smaller bitvector
                     set = set.Zero;
                 }
             }
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -123,7 +123,8 @@ namespace System.Text.RegularExpressions.SRM
             while (!set.IsLeaf)
             {
                 if (set.Zero.IsEmpty)
-                { // The bit must be set to 1 when the zero branch is False.
+                { 
+                    // The bit must be set to 1 when the zero branch is False
                     res = res | ((ulong)1 << set.Ordinal);
                     // If zero is empty then by the way BDDs are constructed one is not.
                     set = set.One;
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -15,6 +15,7 @@ namespace System.Text.RegularExpressions.SRM
         /// The ordinal for the True special value.
         /// </summary>
         private const int TrueOrdinal = -2;
+        
         /// <summary>
         /// The ordinal for the False special value.
         /// </summary>
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -1,26 +1,33 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
 using System.Collections.Generic;
-using System.Runtime.CompilerServices;
-using System.IO;
 
 namespace System.Text.RegularExpressions.SRM
 {
     /// <summary>
-    /// Represents a Binary Decision Diagram.
+    /// Represents a Binary Decision Diagram (BDD). Supports multi-terminal BDDs, i.e. ones where the leaves are
+    /// something else the True or False, which are used for representing classifiers.
     /// </summary>
-    internal class BDD : IComparable
+    internal sealed class BDD : IComparable
     {
+        /// <summary>
+        /// The ordinal for the True special value.
+        /// </summary>
+        private const int TrueOrdinal = -2;
+        /// <summary>
+        /// The ordinal for the False special value.
+        /// </summary>
+        private const int FalseOrdinal = -1;
+
         /// <summary>
         /// The unique BDD leaf that represents the full set or true.
         /// </summary>
-        public static BDD True = new BDD(-2, null, null);
+        public static BDD True = new BDD(TrueOrdinal, null, null, specialValue: true);
         /// <summary>
         /// The unique BDD leaf that represents the empty set or false.
         /// </summary>
-        public static BDD False = new BDD(-1, null, null);
+        public static BDD False = new BDD(FalseOrdinal, null, null, specialValue: true);
 
         /// <summary>
         /// The encoding of the set for lower ordinals for the case when the current bit is 1.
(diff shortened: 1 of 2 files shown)


veanes (approved):
Edits look good to me overall.
Perhaps one thing to think about is if we should replace AutomataException with some other exception (either existing one or rename the exception). In most cases it indicates an internal error of some kind. This note applies to the whole SRM codebase and perhaps should be revisited later.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -1,26 +1,33 @@
...
 using System.Collections.Generic;
-using System.Runtime.CompilerServices;
-using System.IO;
 
 namespace System.Text.RegularExpressions.SRM
 {
     /// <summary>
-    /// Represents a Binary Decision Diagram.
+    /// Represents a Binary Decision Diagram (BDD). Supports multi-terminal BDDs, i.e. ones where the leaves are
+    /// something else the True or False, which are used for representing classifiers.
```
danmoseley:
```suggestion
    /// something other than True or False, which are used for representing classifiers.
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -1,26 +1,33 @@
...
     /// </summary>
-    internal class BDD : IComparable
+    internal sealed class BDD : IComparable
     {
+        /// <summary>
+        /// The ordinal for the True special value.
+        /// </summary>
+        private const int TrueOrdinal = -2;
+        /// <summary>
```
danmoseley:
```suggestion
        private const int TrueOrdinal = -2;
        
        /// <summary>
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -120,47 +115,22 @@ public ulong GetMin()
...
+            // Starting from all 0, bits will be flipped to 1 as necessary.
             ulong res = 0;
 
+            // Follow the minimum path throught the branches to a True leaf.
             while (!set.IsLeaf)
             {
-                if (set.Zero.IsEmpty) //the bit must be set to 1
-                {
+                if (set.Zero.IsEmpty)
+                { // The bit must be set to 1 when the zero branch is False.
```
danmoseley:
```suggestion
                { 
                    // The bit must be set to 1 when the zero branch is False
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -35,24 +42,38 @@ internal class BDD : IComparable
...
             One = one;
             Zero = zero;
             Ordinal = ordinal;
-            //precompute a hashchode value that respects BDD equivalence
-            //i.e. two equivalent BDDs will always have the same hashcode
-            //that is independent of object id values of the BDD objects
-            hashcode = (ordinal, one, zero).GetHashCode();
+            // Precompute a hash code value that respects BDD equivalence i.e.
+            // two equivalent BDDs will always have the same hashcode that is
+            // independent of object id values of the BDD objects.
```
danmoseley, line 76: Not worth fixing today,  but we generally don't end an inline comment with a period unless it's more than one sentence.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -120,47 +115,22 @@ public ulong GetMin()
...
-                {
+                if (set.Zero.IsEmpty)
+                { // The bit must be set to 1 when the zero branch is False.
                     res = res | ((ulong)1 << set.Ordinal);
-                    set = set.One;  //must follow the 1-branch
+                    // If zero is empty then by the way BDDs are constructed one is not.
+                    set = set.One;
                 }
                 else
+                { // Otherwise, leaving the bit as 0 gives the smaller bitvector.
```
danmoseley:
```suggestion
                {
                    // Otherwise, leaving the bit as 0 gives the smaller bitvector
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -1,26 +1,33 @@
...
-using System;
 using System.Collections.Generic;
-using System.Runtime.CompilerServices;
-using System.IO;
 
 namespace System.Text.RegularExpressions.SRM
 {
     /// <summary>
-    /// Represents a Binary Decision Diagram.
+    /// Represents a Binary Decision Diagram (BDD). Supports multi-terminal BDDs, i.e. ones where the leaves are
```
danmoseley, line 9: maybe explain what MTBDD is (multi terminal?)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -169,163 +139,184 @@ public ulong GetMax()
...
                 else
-                {
-                    if (nonterminals[node.Ordinal] == null)
-                        nonterminals[node.Ordinal] = new List<BDD>();
-                    nonterminals[node.Ordinal].Add(node);
-                    if (set.Add(node.Zero))
-                        stack.Push(node.Zero);
-                    if (set.Add(node.One))
-                        stack.Push(node.One);
+                { // Non-terminals are grouped by their ordinal first.
```
danmoseley, line 190: We don't put comments on the same line as open braces -- I wont mark them all.

I know this is all trivia, but it catches eyeballs. It's funny how we converge on this stuff.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -169,163 +139,184 @@ public ulong GetMax()
...
+            return sorted.ToArray();
         }
 
         #region Serialization
-        private static readonly long[] s_False_repr = new long[] { 0 };
-        private static readonly long[] s_True_repr = new long[] { 1 };
         /// <summary>
-        /// Serialize this BDD in a flat ulong array.
-        /// The BDD may have at most 2^k ordinals and 2^n nodes, st. k+2n&lt;64.
+        /// Serialize this BDD in a flat ulong array. The BDD may have at most 2^k ordinals and 2^n nodes, st. k+2n&lt;64.
```
danmoseley:
```suggestion
        /// Serialize this BDD in a flat ulong array. The BDD may have at most 2^k ordinals and 2^n nodes, such that k+2n &lt; 64
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
danmoseley: Probably not worth bothering with fixing though. There are better uses of time.. 😸

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -461,72 +458,78 @@ public bool Contains(int input)
...
             {
-                var node = stack.Pop();
+                var node = toVisit.Pop();
                 if (node.IsEmpty)
                     continue;
 
                 if (node.IsFull)
                 {
-                    //contains the True leaf
+                    // Contains the True leaf, so not essentially boolean in this method's sense.
```
danmoseley, line 494: Another nit not worth fixing globally. We seem to usually start inline comments with lower case.

danmoseley (approved):
Looks good to me!

I don't yet understand the mechanism, I'll have to look through that all up when you've finished the style cleanup.

olsaarik:
@danmoseley Thanks for all the comments! The nitpicks are definitely useful, since at least some of these were things I changed when going through the code. I'll keep these in mind best I can when working through the rest. I'll close this PR for now and submit a new one when I have a slightly larger section of the code cleaned up (maybe the algebras folder).

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
olsaarik: Good to know, that's what Margus has been doing too I think, so I won't be "fixing" those then.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
olsaarik: Agreed! I'll add an explanation.

Issue #1397, opened 2021-08-08
ARM: JIT crash


kant2002:
During samples compilation. ARM built from https://github.com/dotnet/runtimelab/commit/4424f855c5b80b0bdf0f46c60824891be4367144

During publish of HelloWorld sample. No modifications except local feed change.

```
  /home/pi/runtimelab/src/coreclr/jit/lclvars.cpp:2618
  Assertion failed 'varTypeIsStruct(varDsc)' in 'System.SpanHelpers:Fill(byref,int,int)' during 'Morph - Global' (IL size 1139)
```

kant2002:
Not sure, but maybe #1388 gone. I may change clang version on new Raspberry or maybe something changed in the runtime repo.

jkotas:
What is the `T` that it is crashing on? The same crash is likely going to repro on regular CoreCLR as well.

kant2002:
`T` is `int`. Code below enough for crash.
```
var x = new int[10];
var s = new Span<int>(x);
s.Fill(0);
```

Also #1388 is not gone,since if I publish empty console app, I still receive segmentation fault.

kant2002:
I cannot reproduce that issue in not-NativeAOT context. Regular `dotnet run` just works. What route should I take?

jkotas:
I would capture JIT dumps of the method in NativeAOT and in non-Native (make sure to turn off tiered JIT and R2R, and use same CoreLib build flavor in both cases), and then compare the two JIT dumps to find out why things are different in NativeAOT.

kant2002:
I think I close to finding root cause of the issue. Part of JIT dump in NativeAOT.
```
Morphing BB41 of 'System.SpanHelpers:Fill(byref,int,int)'

fgMorphTree BB41, STMT00093 (before)
               [000358] S-CXG--------             *  CALL      void   System.Runtime.Intrinsics.Vector256.AsVector
               [000361] ------------- arg0        +--*  LCL_VAR_ADDR byref  V10 loc7         
               [000357] --CXG-------- arg1        \--*  OBJ       struct<System.Runtime.Intrinsics.Vector256`1[System.Byte], 32>
               [000721] -------------                \--*  ADDR      byref 
               [000722] -------N-----                   \--*  LCL_VAR   int   (AX) V09 loc6         
Initializing arg info for 358.CALL:
ArgTable for 358.CALL after fgInitArgInfo:
[...]

PR #1399, opened 2021-08-09
Fixes when cross compile ARM on Windows

kant2002:
- Interlocked.asm is copied Interlocked.S
- gcrhenv.cpp contains fix for signed/unsigned typed.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/arm/Interlocked.asm
+++ b/src/coreclr/nativeaot/Runtime/arm/Interlocked.asm
@@ -1,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#include "AsmMacros.h"
+
+// WARNING: Code in EHHelpers.cpp makes assumptions about this helper, in particular:
+// - Function "InWriteBarrierHelper" assumes an AV due to passed in null pointer will happen at RhpLockCmpXchg32AVLocation
+// - Function "UnwindSimpleHelperToCaller" assumes no registers were pushed and LR contains the return address
+// r0 = destination address
+// r1 = value
+// r2 = comparand
+LEAF_ENTRY RhpLockCmpXchg32, _TEXT
+          dmb
+ALTERNATE_ENTRY RhpLockCmpXchg32AVLocation
+LOCAL_LABEL(CmpXchg32Retry):
+          ldrex        r3, [r0]
+          cmp          r2, r3
+          bne          LOCAL_LABEL(CmpXchg32Exit)
+          strex        r12, r1, [r0]
+          cmp          r12, #0
+          bne          LOCAL_LABEL(CmpXchg32Retry)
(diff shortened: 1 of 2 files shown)


jkotas:
.NET on Windows ARM32 is unsupported since .NET 5. The devices that are able to run it are increasingly rare. Why do you care about it?

kant2002:
I do not care at all. I was trying to debug ARM on Linux in more pleasant environment. I was trying follow suggestion from @MichalStrehovsky how I can debug Jit issue for ARM on Windows. Seems to be it should be more involved then just couple arguments to ILC. So essentiallly I somehow want attach debug symbols for libjitinterface_arm.so on ARM to troubleshoot #1397

MichalStrehovsky:
> Seems to be it should be more involved then just couple arguments to ILC

No, it should be just a couple arguments to ILC. What I was suggesting:

* build.cmd nativeaot+libs
* Open ILC in VS and set up the response file for the repro project. Add `--targetarch arm --targetos linux` to the command line so that it crosscompiles.
* Done

When you launch debugging with F5, it's going to compile the repro project targeting ARM32 Linux. You might be able to hit your JIT crash there (the output of the compilation will be garbage because we're using the Windows CoreLib, but at this point we'll not try to run the produced executable, so it doesn't matter)

If the JIT crash doesn't repro, copy over RSP file from your Pi, along with all the files the RSP refers to. Adjust paths in the RSP file as needed. Add `--targetos:linux --targetarch:arm`. Hit F5 to debug again.

MichalStrehovsky:
In particular, notice that I didn't pass `-arch` to `build.cmd nativeaot+libs`. We want an x64 compiler and x64 JIT so that you can run it on your machine.

kant2002:
Thanks. Unfortunately that's what I initially do. When I build NativeAOT on Windows as usual. i.e as I do for Windows. No special build
Then I run just ILC with additional `--targetos linux --targetarch arm`.
When I run like that I experience crashes 
```
Code generation failed for method '[S.P.CoreLib]Internal.Runtime.CompilerHelpers.StartupCodeHelpers.InitializeGlobalTablesForModule(TypeManagerHandle,int32,object[])'
Code generation failed for method '[S.P.CoreLib]Internal.Runtime.CompilerHelpers.StartupCodeHelpers.InitializeEagerClassConstructorsForModule(TypeManagerHandle)'
Expected method 'IDiv' not found on type '[S.P.CoreLib]Internal.Runtime.CompilerHelpers.MathHelpers'
```

I take a look at 
https://github.com/dotnet/runtimelab/blob/4ffe2eb2b6c972697fa3888620bca27e6c391578/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/MathHelpers.cs#L289-L294
and notice TARGET_ARM so I "smartly" decide that I should add command line arguments to `build.cmd`

kant2002:
I think I can work with ARM on windows, so this PR can be closed if nobody need it.

PR #1405, opened 2021-08-11
Speed up generic virtual method analysis

MichalStrehovsky:
I'm ramping up on generic virtual method implementation so that I can add default interface method support and this is a thing I noticed. I used this work as a way to prove to myself I understand how things work in the compiler.

For apps that use generic virtual methods extensively (e.g. Avalonia), we spend a considerable amount of time analyzing generic virtual methods.

As I was looking at things, I realized we're doing a lot more work than needed. This is basically two things:

1. `TypeGVMEntriesNode` is used to build a data structure that describes things like: IFoo<!1>.Frob is implemented by Foo.Frob on Foo. We were creating a `TypeGVMEntriesNode` object for each EEType that is interesting to GVM analysis. But for generic types, the various instantiations weren't contributing anything since this knowledge is the same irrespective of the T's. We only need one` TypeGVMEntriesNode` per type definition.
2. `GVMDependenciesNode` describes "a generic virtual method was used". Since we stopped eagerly pregenerating generic dictionaries for these methods, we only need one `GVMDependenciesNode` per canonical form of the method.

Speeds up compilation of a basic Avalonia app by 8%. Size of the output stays the same.


jkotas (commented):
Nice!

jkotas (approved):

PR #1406, opened 2021-08-11
Generate method metadata even if we can't make a body

### LINKED ISSUE #1402: GetMethods(BindingFlags.Public | BindingFlags.Static) does not root generic methods
```
using System;
using System.Reflection;

class Program
{
    {
        foreach (MethodInfo m in typeof(Enum).GetMethods(BindingFlags.Public | BindingFlags.Static)) Console.WriteLine(m);
    }
}
```

Expected: Output includes `TryParse[TEnum]`
Actual: No `TryParse[TEnum]` in the output

MichalStrehovsky:
Fixes #1402.


jkotas:
Failing tests...

jkotas (approved):
Thank you!

PR #1408, opened 2021-08-12
Do not throw when when see foreign type

### LINKED ISSUE #1403: InvalidProgramException during type pre-init
I receive exception at function
`[S.P.CoreLib]System.DateTime.DateToTicks(int32,int32,int32)` during PreInit
https://github.com/dotnet/runtimelab/blob/4ffe2eb2b6c972697fa3888620bca27e6c391578/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs#L1345

This happens when those compiling those classes 
```
{[S.P.CoreLib]System.Globalization.HijriCalendar}
{[S.P.CoreLib]System.Globalization.JapaneseCalendar}
{[S.P.CoreLib]System.Globalization.HebrewCalendar}
{[S.P.CoreLib]System.Globalization.PersianCalendar}
{[S.P.CoreLib]System.Globalization.UmAlQuraCalendar}
{[S.P.CoreLib]System.Globalization.CalendricalCalculationsHelper}
```

kant2002:
Static fields from other types cannot be pre-initialized and this is wel-known case so do not throw.
Closes #1403

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
@@ -1340,6 +1340,10 @@ namespace ILCompiler
                             {
                                 return Status.Fail(methodIL.OwningMethod, opcode, "Null array");
                             }
+                            else if (array.Value is ForeignTypeInstance)
+                            {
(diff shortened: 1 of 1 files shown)


MichalStrehovsky (approved):
Thanks!

PR #1411, opened 2021-08-13
Use struct with default constructor in CreateInstance

MichalStrehovsky:
Fixes https://github.com/dotnet/runtimelab/pull/849/files#r600874149.


jkotas (approved):
Thanks

Issue #1388, opened 2021-08-06
Crash in ObjWriter on ARM during publish


kant2002:
After #1387 during publish ILC crashed with following stack trace.

```
Thread 1 "ilc" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x76c36510 (LWP 26921)]
0x268d53f8 in ?? ()
(gdb) backtrace
#0  0x268d53f8 in ?? ()
#1  0x460addb4 in llvm::MCExpr::evaluateAsAbsolute(long long&, llvm::MCAssembler const&) const () from /home/pi/runtimelab/samples/HelloWorld/pkg/runtime.linux-arm.microsoft.dotnet.ilcompiler/6.0.0-dev/tools/libobjwriter.so
#2  0x453a3e40 in ObjectWriter::EmitSymbolRef(char const*, RelocType, int) () from /home/pi/runtimelab/samples/HelloWorld/pkg/runtime.linux-arm.microsoft.dotnet.ilcompiler/6.0.0-dev/tools/libobjwriter.so
#3  0x453a36b0 in EmitSymbolRef () from /home/pi/runtimelab/samples/HelloWorld/pkg/runtime.linux-arm.microsoft.dotnet.ilcompiler/6.0.0-dev/tools/libobjwriter.so
#4  0x52a19706 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

I try to look at locals, but without luck 
```
(gdb) frame 2
#2  0x453a3e40 in ObjectWriter::EmitSymbolRef(char const*, RelocType, int) () from /home/pi/runtimelab/samples/HelloWorld/pkg/runtime.linux-arm.microsoft.dotnet.ilcompiler/6.0.0-dev/tools/libobjwriter.so
(gdb) info locals
No symbol table info available.
```
Even if I build LLVM and ObjWriter on this Raspberry using `./build.sh nativeaot.objwriter -rc Debug -lc Release`. It will take a bit of time to figure out what's going on.

kant2002:
When I try run ILC on Windows, On the ObjWriter stage probably I receive `LLVM ERROR: unsupported relocation on symbol`

That's seems to be from one of these places
https://github.com/llvm/llvm-project/blob/d480f968ad8b56d3ee4a6b6df5532d485b0ad01e/llvm/lib/Target/ARM/MCTargetDesc/ARMELFObjectWriter.cpp#L90
or
[...]

PR #1412, opened 2021-08-15
Prevent crash on ARM32

kant2002:
This change has 2 parts, one which is JIT part and likely require submission to runtime.
I keep that change here to give full scope of changes. Will extract if everything would be fine.
JIT for ARM now start numering floating point registers starting from 256

Second part is DWARF generation
In order to produce DWARF I made following changes
- Account for register numbers more then 31, by use DW_OP_bregx and DW_OP_regx for large regnum
- Combine VLT_FPSTK and VLT_STK handling.

See #1388

### DIFF
--- a/src/coreclr/jit/unwind.cpp
+++ b/src/coreclr/jit/unwind.cpp
@@ -188,11 +188,11 @@ void Compiler::unwindPushPopMaskCFI(regMaskTP regMask, bool isFloat)
 
     for (regNumber regNum = isFloat ? REG_FP_FIRST : REG_FIRST; regNum < REG_COUNT;
 #if TARGET_ARM
-         regNum = isFloat ? REG_NEXT(REG_NEXT(regNum)) : REG_NEXT(regNum),
+         regNum = isFloat ? REG_NEXT(REG_NEXT(regNum)) : REG_NEXT(regNum), regBit <<= isFloat ? 2 : 1
 #else
-         regNum = REG_NEXT(regNum),
+         regNum = REG_NEXT(regNum), regBit <<= 1
 #endif
-         regBit <<= 1)
+         )
     {
         if (regBit > regMask)
         {
--- a/src/coreclr/jit/unwind.cpp
+++ b/src/coreclr/jit/unwind.cpp
@@ -188,7 +188,7 @@ void Compiler::unwindPushPopMaskCFI(regMaskTP regMask, bool isFloat)
 
     for (regNumber regNum = isFloat ? REG_FP_FIRST : REG_FIRST; regNum < REG_COUNT;
 #if TARGET_ARM
-         regNum = isFloat ? ((regNumber)((unsigned)(regNum) + 2)) : REG_NEXT(regNum),
+         regNum = isFloat ? REG_NEXT(REG_NEXT(regNum)) : REG_NEXT(regNum),
 #else
          regNum = REG_NEXT(regNum),
 #endif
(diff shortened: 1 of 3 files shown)


src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfGen.cpp
```diff
@@ -331,6 +367,7 @@ static void EmitVarLocation(MCObjectStreamer *Streamer,
       IsByRef = true;
     case ICorDebugInfo::VLT_STK2:
       IsStk2 = true;
+    case ICorDebugInfo::VLT_FPSTK:
```
jkotas, line 381: Did you actually hit this case? I would not expect it to be hit on. Floating point stack is a legacy 32-bit x86 that are not using anymore even on 32-bit x86.

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfGen.cpp
```diff
@@ -284,6 +284,42 @@ static int GetDwarfFpRegNum(Triple::ArchType ArchType)
...
+    return 0;
+  }
+}
+
+static void EmitBreg(MCObjectStreamer* Streamer, int DwarfRegNum) {
+  if (DwarfRegNum <= 31) {
+    Streamer->EmitIntValue(DwarfRegNum + dwarf::DW_OP_breg0, 1);
+  }
+  else {
+    Streamer->EmitIntValue(dwarf::DW_OP_bregx, 1);
```
jkotas, line 308: According to the dwarf manual: "The DW_OP_bregx operation has two operands: a register which is specified by an unsigned LEB128 number, followed by a signed LEB128 offset.". It would be nice to emit the offset as part of this function instead of emitting it in the calling code.

src/coreclr/jit/unwind.cpp
```diff
@@ -187,7 +187,12 @@ void Compiler::unwindPushPopMaskCFI(regMaskTP regMask, bool isFloat)
     regMaskTP regBit = isFloat ? genRegMask(REG_FP_FIRST) : 1;
 
     for (regNumber regNum = isFloat ? REG_FP_FIRST : REG_FIRST; regNum < REG_COUNT;
-         regNum           = REG_NEXT(regNum), regBit <<= 1)
+#if TARGET_ARM
+         regNum = isFloat ? ((regNumber)((unsigned)(regNum) + 2)) : REG_NEXT(regNum),
```
jkotas:
```suggestion
         regNum = isFloat ? REG_NEXT(REG_NEXT(regNum)) : REG_NEXT(regNum),
```

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfGen.cpp
kant2002: It is hit. I was tired of seeing assertion. This change was fishy to me too. But I do not know where to start looking. With that hint, I try to pin-point what can cause code to hit that location.

src/coreclr/jit/unwind.cpp
```diff
@@ -187,7 +187,12 @@ void Compiler::unwindPushPopMaskCFI(regMaskTP regMask, bool isFloat)
...
+#else
+         regNum = REG_NEXT(regNum),
+#endif
+         regBit <<= 1)
```
jkotas: Does this need to shift the bit by 2 for floats as well?

src/coreclr/tools/aot/ObjWriter/debugInfo/dwarf/dwarfGen.cpp
jkotas: I have looked around. `VLT_FPSTK` is actually used for regular local variables that do not live on floating point stack. It does not look right; but it may done that way for some legacy reasons. So your change is ok.

kant2002:
@jkotas If you don't have any other suggestions, can I split this PR on two parts? One for runtime, second for this repo.

jkotas:
I do not have any other comments.

>  can I split this PR on two parts? One for runtime, second for this repo.

Yes, it would be nice. Thank you!

kant2002:
Superseeded by #1413 and https://github.com/dotnet/runtime/pull/57443

PR #1413, opened 2021-08-15
Fix generation of DWARF when regnum >31

kant2002:
Also for ARM use D0-D31 registers for encoding S0-S31
See #1388


jkotas (approved):
Thanks

PR #1416, opened 2021-08-15
Fix handling diamond case in GVMs

MichalStrehovsky:
`targetTypeHandle` would be null if there's no destination.

Fixes two of the default interface methods tests in the Pri-0 test suite.


jkotas (approved):
Thanks

PR #1418, opened 2021-08-16
Fix invoking instance methods on interfaces

MichalStrehovsky:
Non-virtual instance methods on interfaces need an instantiation argument. Also, stop assuming all interface methods are virtual.


jkotas (approved):
Thanks

PR #1419, opened 2021-08-16
Improve test project files

MichalStrehovsky:
* Reflection-root the entrypoint assembly. There are a couple Pri-0 failing tests that are reflecting on things we didn't compile. I think this is a regression - we used to root them before unification with IL Linker targets.
* Copy RuntimeHostConfigurationOptions from the source compilation project to native compilation project. Interop\PInvoke\Variant\VariantTestComWrappers expects to see the AppContext switches.


jkotas (approved):
Thanks

Issue #1424, opened 2021-08-17
ARM Linker issues


kant2002:
After I fix scanner issues, I have following error from linker
```
/usr/bin/ld: sdk/libRuntime.a(EHHelpers.cpp.o):(.data+0x1c): undefined reference to `RhpInterfaceDispatchAVLocation1'
/usr/bin/ld: sdk/libRuntime.a(EHHelpers.cpp.o):(.data+0x20): undefined reference to `RhpInterfaceDispatchAVLocation2'
/usr/bin/ld: sdk/libRuntime.a(EHHelpers.cpp.o):(.data+0x24): undefined reference to `RhpInterfaceDispatchAVLocation4'
/usr/bin/ld: sdk/libRuntime.a(EHHelpers.cpp.o):(.data+0x28): undefined reference to `RhpInterfaceDispatchAVLocation8'
/usr/bin/ld: sdk/libRuntime.a(EHHelpers.cpp.o):(.data+0x2c): undefined reference to `RhpInterfaceDispatchAVLocation16'
/usr/bin/ld: sdk/libRuntime.a(EHHelpers.cpp.o):(.data+0x30): undefined reference to `RhpInterfaceDispatchAVLocation32'
/usr/bin/ld: sdk/libRuntime.a(EHHelpers.cpp.o):(.data+0x34): undefined reference to `RhpInterfaceDispatchAVLocation64'
```

I assume that I should provide

kant2002:
I take a look and seems that I need add `ALTERNATE_ENTRY RhpInterfaceDispatchAVLocation\entries` somewhere here
https://github.com/dotnet/runtimelab/blob/d648860b17173467e6f76116f8906596172e8ddc/src/coreclr/nativeaot/Runtime/arm/StubDispatch.S#L15-L21

I look how that's done on ARM64 and I see lot less ceremony with saving registers. That maybe due to fact that this is newer code, or because this is 64-bit arch. For example

https://github.com/dotnet/runtimelab/blob/d648860b17173467e6f76116f8906596172e8ddc/src/coreclr/nativeaot/Runtime/arm64/StubDispatch.S#L32-L39

Also on ARM64 this entry is `NESTED_ENTRY` whereas ARM uses `LEAF_ENTRY`. I need some education on that parts or what direction should I take.

jkotas:
> I look how that's done on ARM64 and I see lot less ceremony with saving registers.

That's because ARM64 has enough scratch registers. 

[...]

PR #1427, opened 2021-08-17
Add access violation location in ASM

kant2002:
That location used in `InInterfaceDispatchHelper` to check for possible NRE
See #1424

### DIFF
--- a/src/coreclr/nativeaot/Runtime/EHHelpers.cpp
+++ b/src/coreclr/nativeaot/Runtime/EHHelpers.cpp
@@ -376,6 +376,10 @@ static uintptr_t UnwindSimpleHelperToCaller(
     pContext->SetSp(sp+sizeof(uintptr_t)); // pop the stack
 #elif defined(HOST_ARM) || defined(HOST_ARM64)
     uintptr_t adjustedFaultingIP = pContext->GetLr();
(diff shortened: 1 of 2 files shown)


jkotas:
I would be also nice to update `UnwindSimpleHelperToCaller` to match. It should be:

```
...
    uintptr_t adjustedFaultingIP = pContext->GetLr();
#if defined(HOST_ARM)
    if (InInterfaceDispatchHelper(faultingIP))
        pContext->SetSp(pContext->GetSp()+2*sizeof(uintptr_t));
#endif
...
```

kant2002:
That's for manually pop data from stack which was previously added. Correct?

jkotas:
BTW: An alternative way to deal with this would be to use redzone to save the registers (8 bytes below the current sp). I see that the code later in the function is using redzone already. It would save you from dealing with pushing and popping the registers.

jkotas:
> That's for manually pop data from stack which was previously added. Correct?

Right

kant2002:
> BTW: An alternative way to deal with this would be to use redzone to save the registers

I hope I understand the idea correctly.

jkotas (approved):
Yes, I think you got it right. Thanks!

PR #1430, opened 2021-08-18
Implement TryResolveConstraintMethodApprox

MichalStrehovsky:
I tried to implement the full behavior (that wouldn't throw from forceRuntimeLookup), but that started to be hairy (https://github.com/MichalStrehovsky/runtimelab/commit/7d632d43ddd1f5b21383c4123b924a5623d83b5e) so I'll leave it in a branch: the dictionary entries look different from what we would need (not fat pointers), we need to intern the ConstrainedCallInfo, and who knows what else is needed on top of what's in that commit.


jkotas (approved):
Looks good, once the tests are passing

Issue #162, opened 2020-10-09
Support IDynamicInterfaceCastable


MichalStrehovsky:
New feature in .NET 5.

Depends on #161.

MichalStrehovsky:
Leftover work tracked in dotnet/runtime#72909.

PR #1433, opened 2021-08-20
Implement IDynamicInterfaceCastable

MichalStrehovsky:
Adds almost full support for IDynamicInterfaceCastable. The missing part is support for shared interface methods - we need to provide generic context somehow and instantiating stubs are not it because we don't have them.

I used https://github.com/dotnet/corert/pull/8103 as a blueprint, with a small difference - instead of hardcoding slot numbers into the MethodTable, I'm making runtime call into CoreLib so that we can call methods on `IDynamicInterfaceCastable` as usual. Being in the CoreLib makes various error handling easier too.

The implementation in the compiler is very straightforward:

* Allow `DynamicInterfaceCastableImplementation` interfaces to have a dispatch map.
* Allow `DynamicInterfaceCastableImplementation` to have a sealed vtable.
* Pretty much all that's needed.

At runtime, we consult the dispatch map to find the proper implementation, given an implementation interface.

Contributes to #162.

### DIFF
--- a/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/CachedInterfaceDispatch.cs
+++ b/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/CachedInterfaceDispatch.cs
@@ -29,7 +29,17 @@ namespace System.Runtime
             IntPtr pTargetCode = RhResolveDispatchWorker(pObject, (void*)pCell, ref cellInfo);
             if (pTargetCode != IntPtr.Zero)
             {
-                return InternalCalls.RhpUpdateDispatchCellCache(pCell, pTargetCode, pObject.MethodTable, ref cellInfo);
+                // We don't update the dispatch cell cache if this is IDynamicInterfaceCastable because this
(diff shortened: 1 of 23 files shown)


src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/CachedInterfaceDispatch.cs
```diff
@@ -29,7 +29,17 @@ private static IntPtr RhpCidResolve_Worker(object pObject, IntPtr pCell)
             IntPtr pTargetCode = RhResolveDispatchWorker(pObject, (void*)pCell, ref cellInfo);
             if (pTargetCode != IntPtr.Zero)
             {
-                return InternalCalls.RhpUpdateDispatchCellCache(pCell, pTargetCode, pObject.MethodTable, ref cellInfo);
+                // We don't update the dispatch cell cache if this is IDynamicInterfaceCastable because this
```
MichalStrehovsky, line 32: We didn't have this check in `ICastable`. Probably a bug we never hit.

jkotas (approved):
Nice!

Issue #1241, opened 2021-06-15
NativeAOT breaking change announcements


MichalStrehovsky:
This issue will be a running list of potentially breaking changes in NativeAOT. If you use NativeAOT in your project, make sure to click the "Subscribe" button on this issue.

MichalStrehovsky:
The AppContextSwitchOverrides ItemGroup in project files was replaced with the officially documented RuntimeHostConfigurationOption. Migration path is this:

old value: `<AppContextSwitchOverrides Include="Switch.System.Reflection.Disabled.DoNotThrowForNames" />`
new value: `<RuntimeHostConfigurationOption Include="Switch.System.Reflection.Disabled.DoNotThrowForNames" Value="true" />`.

https://github.com/dotnet/runtimelab/pull/1227

MichalStrehovsky:
The ILCompiler package version was bumped to 7.0. If you're using one of 6.0 wildcards to reference the ILCompiler NuGet package, please switch to 7.0 like outlined in https://github.com/dotnet/runtimelab/pull/1452 to keep getting the latest.

MichalStrehovsky:
In the past, setting `<TrimMode>link</TrimMode>` in the csproj would trim the entire app. This was a shortcut because .NET 5 there was a no convenient way to express "I want to trim the whole app".

As of #1596 we now align with the [documented](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options) trimming settings.

To trim the entire app you now need to:

On .NET 6, set the `<TrimmerDefaultAction>link</TrimmerDefaultAction>` property.

On .NET 5 you need to write a custom target (this also works on .NET 6, but why would you do this to yourself?):

```xml
  <Target Name="ConfigureTrimming"
          BeforeTargets="PrepareForILLink">
    <ItemGroup>
      <ManagedAssemblyToLink>
        <TrimMode>link</TrimMode>
      </ManagedAssemblyToLink>
    </ItemGroup>
  </Target>
```

MichalStrehovsky:
We’ve moved the NativeAOT project out of the dotnet/runtimelab repo to the dotnet/runtime repo. As part of the move, the NuGet feed changed from the dotnet-experimental one to the dotnet7 one.

To keep getting ILCompiler package updates, please update your NuGet feed from (old feed):

https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json

To (new feed):

[...]

PR #1441, opened 2021-08-21
Merge from dotnet/runtime

jkotas:


hez2010:
Will this bump ilc version to 7.0.0-*?

jkotas:
> Will this bump ilc version to 7.0.0-*?

Yes, it is a good candidate for #1241

MichalStrehovsky:
Let's update the docs and add a record to the breaking changes issue once such NuGet package is out.

jkotas:
I was wrong about that this will bump ilc version. The ilc version is still 6.0 even after this PR.

hez2010:
I think the ilc version has been bumped to 7.0.0-*
image

jkotas:
Ah, your right. I have looked in the wrong place.

PR #1443, opened 2021-08-23
Work towards shared generics support in IDynamicIntefaceCastable

MichalStrehovsky:
This is enough to get all our test to pass.

The rest is tracked in https://github.com/dotnet/runtime/issues/72909.

It's a bunch of work and at this point I don't know how much we need it.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/SharedCodeHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/SharedCodeHelpers.cs
@@ -19,6 +19,7 @@ namespace Internal.Runtime.CompilerHelpers
         public static unsafe MethodTable* GetCurrentSharedThunkContext()
         {
             // TODO: We should return the current context from the ThunkPool
+            // https://github.com/dotnet/runtimelab/issues/1442
             return null;
         }
     }
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/SharedCodeHelpers.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/SharedCodeHelpers.cs
@@ -15,5 +15,11 @@ namespace Internal.Runtime.CompilerHelpers
             Debug.Assert(interfaceIndex <= pType->NumInterfaces);
             return pType->InterfaceMap[interfaceIndex].InterfaceType;
         }
+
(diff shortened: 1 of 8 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/SharedCodeHelpers.cs
```diff
@@ -15,5 +15,11 @@ internal static class SharedCodeHelpers
             Debug.Assert(interfaceIndex <= pType->NumInterfaces);
             return pType->InterfaceMap[interfaceIndex].InterfaceType;
         }
+
+        public static unsafe MethodTable* GetCurrentSharedThunkContext()
+        {
+            // TODO: We should return the current context from the ThunkPool
```
jkotas, line 21: Add a link to the issue you have opened? 

Maybe also add assert? Or is it actually exercised by the tests and we are just getting lucky that nothing is accessing the context?

jkotas (approved):

jkotas:
> It's unclear to me if anyone uses IDynIntfCast with generics.

I think cswinrt is going to use it with generics. However, I am not sure whether it is going to hit this specific case with missing context.

MichalStrehovsky:
> I think cswinrt is going to use it with generics. However, I am not sure whether it is going to hit this specific case with missing context.

I tried CsWinRT but at this point it's AOT unfriendly beyond any workarounds.

One thing is all the `MakeGenericType` and `MakeGenericMethod` calls and string concatenations passed to `GetType`. Those can be worked around with RD.XML. It's not pretty but it's workable.

But then there's "[declaring a delegate type requires writing a whole new line in C# so let's ask Expressions to reflection-emit a new delegate type at runtime for us instead](https://github.com/microsoft/CsWinRT/blob/dcec6e0f71bb0b9eeeec6cb25aeee20eab61744f/src/WinRT.Runtime/Projections/IReadOnlyList.net5.cs#L209-L210)".

I miss MCG.

MichalStrehovsky:
Let's not merge yet, there's a bug.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/SharedCodeHelpers.cs
MichalStrehovsky: It is exercised by one of the tests, but the test doesn't do anything with the context.

AaronRobinsonMSFT:
>I miss MCG.

Said no one ever :-)

PR #1444, opened 2021-08-23
Garbage collect unused runtime imports/exports

MichalStrehovsky:
Noticed while I was looking at thunk pools.

There's a couple more around restricted callouts, but I assume we're going to need that for ComWrappers lifetime tracking at some point.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/RuntimeImports.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/RuntimeImports.cs
@@ -236,19 +236,6 @@ namespace System.Runtime
             return h;
         }
 
-        // Allocate variable handle with its initial type.
-        [MethodImpl(MethodImplOptions.InternalCall)]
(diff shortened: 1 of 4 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/RuntimeImports.cs
```diff
@@ -236,19 +236,6 @@ internal static IntPtr RhHandleAllocDependent(object primary, object secondary)
             return h;
         }
 
-        // Allocate variable handle with its initial type.
-        [MethodImpl(MethodImplOptions.InternalCall)]
-        [RuntimeImport(RuntimeLibrary, "RhpHandleAllocVariable")]
-        private static extern IntPtr RhpHandleAllocVariable(object value, uint type);
```
jkotas, line 242: This may be also needed for ComWrappers lifetime tracking eventually, but it is fine to delete it for now.

jkotas (approved):
Thanks!

Issue #1426, opened 2021-08-17
ARM Empty application crashed.


kant2002:
After I speculatively insert ALTERNATE ENTRY here:
https://github.com/dotnet/runtimelab/blob/d648860b17173467e6f76116f8906596172e8ddc/src/coreclr/nativeaot/Runtime/arm/StubDispatch.S#L24

I compile application
```
using System;

namespace HelloWorld
{
    class Program
    {
        static void Main(string[] args)
        {
//            Console.WriteLine("Hello World!"); // I still have JIT issue #1397 so have to comment out.
        }
    }
}
```

And receive crash. In GDB application crashed here:
```
Thread 1 "HelloWorld" received signal SIGSEGV, Segmentation fault.
0x0042578c in vtable for __Array<S_P_Reflection_Execution_Internal_Reflection_Execution_ExecutionEnvironmentImplementation_FunctionPointerOffsetPair> ()
```

jkotas:
Try to step through the assembly code that leads to the crash.

kant2002:
I will take a look at this when I be at home, but post just in case if this is something obvious.

```
(gdb) disas 0x42578c
Dump of assembler code for function _ZTV124__Array<S_P_Reflection_Execution_Internal_Reflection_Execution_ExecutionEnvironmentImplementation_FunctionPointerOffsetPair>:
=> 0x0042578c:  andgt   r0, r2, r8
   0x00425790:  andeq   r0, r0, r12
   0x00425794:  eorseq  r0, pc, r4, asr #11
   0x00425798:  andeq   r0, r11, r3
   0x0042579c:  strbne  r0, [pc, #2805]!        ; 0x426299
   0x004257a0:  andeq   r4, r10, r1, ror r7
   0x004257a4:  muleq   r10, r1, r7
   0x004257a8:  strdeq  r4, [r10], -r1
   0x004257ac:  mlaseq  pc, r4, lr, r6  ; <UNPREDICTABLE>
   0x004257b0:  eorseq  r6, pc, r4, lsl lr      ; <UNPREDICTABLE>
   0x004257b4:  eorseq  r6, pc, r8, ror #29
   0x004257b8:  eorseq  r6, pc, r12, lsl #30
   0x004257bc:  eorseq  r6, pc, r8, lsr #30
   0x004257c0:  eorseq  r1, pc, r0, asr #19
   0x004257c4:  subeq   r2, r0, r8, ror #27
[...]

PR #1446, opened 2021-08-23
Fix codegen on ARM

kant2002:
NativeAOT has different ABI, so it produce Relative Indirection Virtual Stub when invoke virtual methods on generic interface.

Actual idea by @SingleAccretion
If this is okay, I will submit PR to runtime.

See #1426

### DIFF
--- a/src/coreclr/jit/codegenarmarch.cpp
+++ b/src/coreclr/jit/codegenarmarch.cpp
@@ -2450,7 +2450,9 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
         assert(!call->IsTailCall());
 
         regNumber tmpReg = call->GetSingleTempReg();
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/codegenarmarch.cpp
```diff
@@ -2450,7 +2450,9 @@ void CodeGen::genCallInstruction(GenTreeCall* call)
         assert(!call->IsTailCall());
```
jkotas, line 2450: The comment above (Generate a direct call to a non-virtual user defined or helper method) is wrong - copy&paste bug. It should be either deleted or changed to talk about indirect call.

jkotas:
Yes, it looks good to me. It would be great if you can submit a PR to dotnet/runtime.

kant2002:
Submitted as https://github.com/dotnet/runtime/pull/57948

PR #1448, opened 2021-08-23
Add instructions for building ObjWriter

kant2002:


docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -54,6 +54,10 @@ To run an individual test (after it was built), navigate to the `artifacts\tests
 
 `$repo_root` is the root of your clone of the repo.
 
+## Building ObjWriter
+
```
jkotas: It is not clear when one does one need to run this from this description.

I am wondering whether this would be better as a new bullet point in the `# Building` chapter in this doc.

kant2002: Does something like that works?

jkotas (approved):
Looks good! Thanks

PR #1454, opened 2021-08-24
Fix pointer type validation

MichalStrehovsky:
CsWinRT was triggering the validation for a `sizeof(delegate*<blah>)` statement because we validate everything seen by the JIT.

Consider function pointer types valid, but still throw if we try to make an EEType.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
@@ -76,6 +76,12 @@ namespace ILCompiler.DependencyAnalysis
             _hasConditionalDependenciesFromMetadataManager = factory.MetadataManager.HasConditionalDependenciesDueToEETypePresence(type);
 
             factory.TypeSystemContext.EnsureLoadableType(type);
+
+            // We don't have a representation for function pointers right now
+            if (WithoutParameterizeTypes(type).IsFunctionPointer)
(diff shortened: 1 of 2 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -76,6 +76,12 @@ public EETypeNode(NodeFactory factory, TypeDesc type)
             _hasConditionalDependenciesFromMetadataManager = factory.MetadataManager.HasConditionalDependenciesDueToEETypePresence(type);
 
             factory.TypeSystemContext.EnsureLoadableType(type);
+
+            // We don't have a representation for function pointers right now
```
jkotas, line 80: Is this going to print System.IntPtr like in CoreCLR?
```
unsafe
{
    Console.WriteLine(typeof(delegate* <void>));
}
```

jkotas (approved):
Thanks

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
MichalStrehovsky: No, it didn't do it before this change and it won't do it after. The expected behavior is that we rewrite this method body to a throw...

MichalStrehovsky:
Looks like the "sometimes there's some kind of race while building libraries tests" now happens regularly. I'll try to have a look later. I'm puzzled because we're only supposed to build tests and the race is because we try to rebuild libs for whatever reason).

PR #1456, opened 2021-08-24
Fix constrained method calls

### LINKED ISSUE #1455: [NativeAOT] Bug on enumeration (Dispose)
Here's a minimal repro:

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="7.0.0-*" />
  </ItemGroup>

</Project>
```

```csharp
using System.Collections.Immutable;

var list = ImmutableList<string>.Empty.Add("Hello,").Add( "World!");
foreach (var s in list) 
{
    Console.Write(s);
}
```

Output:
```
Hello,World!Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Collections.Immutable.ImmutableList`1.Enumerator.Dispose() + 0x3f
   at <Program>$.<Main>$(String[]) + 0x121
   at tmp!<BaseAddress>+0x1df6ba
```

---

In my actual code this reproduced problem goes away when avoiding the Dispose call (which is done by the foreach lowering), however, soon after it hits what is possibly another different bug:

```
Process terminated. Generic virtual method pointer lookup failure.

Declaring type handle: MethodTable:0x00007FF74C7567B8
Target type handle: MethodTable:0x00007FF74C77A4E0
Method name: System.Linq.IOrderedEnumerable<TElement>.CreateOrderedEnumerable
Instantiation:
  Argument 00000000: MethodTable:0x00007FF74C760670
```

MichalStrehovsky:
Fixes #1455.

### DIFF
--- a/src/coreclr/tools/Common/TypeSystem/Common/TypeSystemHelpers.cs
+++ b/src/coreclr/tools/Common/TypeSystem/Common/TypeSystemHelpers.cs
@@ -274,7 +274,7 @@ namespace Internal.TypeSystem
                         if (genInterfaceMethod.OwningType != interfaceType)
                             exactInterfaceMethod = context.GetMethodForInstantiatedType(
                                 genInterfaceMethod.GetTypicalMethodDefinition(), (InstantiatedType)interfaceType);
-                        method = canonType.ResolveVariantInterfaceMethodToVirtualMethodOnType(exactInterfaceMethod);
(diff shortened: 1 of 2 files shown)


jkotas (approved):
Thanks

Issue #1458, opened 2021-08-25
ARM missing RhpStackProbe


kant2002:
With 7.0 version
```
  /usr/bin/ld: obj/Release/net6.0/linux-arm/native/HelloWorld.o: in function `S_P_CoreLib_System_Number__Dragon4':
  /home/pi/runtimelab/src/libraries/System.Private.CoreLib/src/System/Number.Dragon4.cs:117: undefined reference to `RhpStackProbe'
```

kant2002:
I see implementation only on x64, Windows and Linux, so no ARM and ARM64.

jkotas:
It should be straightforward to copy over the CoreCLR implementation.

MichalStrehovsky:
(They're named JIT_StackProbe in CoreCLR.)

PR #1460, opened 2021-08-26
Fix name of the RhpStackProbe

kant2002:
When looking for #1458, I found this naming discrepancies. I assume that's just some rename happens over time


kant2002:
It's not that I care that much about this one. This was trivial fix. Close if you don't think that's not needed at all.

MichalStrehovsky (approved):
Thanks!

PR #1461, opened 2021-08-26
Update LLVM 5.0 -> 12.0

MichalStrehovsky:
Picks up 4 years of LLVM evolution.

Size of objectwriter.dll grew from 17 MB to 26 MB with this update, while giving us very little in return. I found that unacceptable.

I had a look at what we're building and how we're building it. Updated things to use the LLVM CMake mechanism and properly declared dependencies. Removed the dependency on `AssemblyPrinter` (that we only used to get target pointer size (!) and `TargetMachine` (that we only used to get the target triple (!!!)).

The size of objectwriter.dll is now about 5.5 MB. `build.cmd nativeaot.objwriter` now takes about 6 minutes (down from 20).


MichalStrehovsky:
Cc @kant2002 @RalfKornmannEnvision this updates LLVM used in ObjectWriter to 12. I hope I didn't break ARM64 (or ARM32). The good news is that building objwriter will get almost 4x faster than it was.

AntonLapounov:
Excellent work and great size reduction!  I guess cloning the 12.0 enlistment should be noticeably slower than cloning the 5.0 enlistment due to its much bigger size, is that correct?

MichalStrehovsky:
> I guess cloning the 12.0 enlistment should be noticeably slower than cloning the 5.0 enlistment due to its much bigger size, is that correct?

Yeah, but it still finishes under 2 minutes so it's not terrible. We could probably optimize the checkout to a sparse one to speed that up a little since we don't need clang, libc and friends but maybe in another pull request.

src/coreclr/tools/aot/ObjWriter/objwriter.cpp
```diff
@@ -846,9 +835,10 @@ void ObjectWriter::EmitDwarfFunctionInfo(const char *FunctionName,
 void ObjectWriter::EmitDebugFileInfo(int FileId, const char *FileName) {
   assert(FileId > 0 && "FileId should be greater than 0.");
   if (ObjFileInfo->getObjectFileType() == ObjFileInfo->IsCOFF) {
-    Streamer->EmitCVFileDirective(FileId, FileName);
+    ArrayRef<uint8_t> ChecksumAsBytes;
+    Streamer->EmitCVFileDirective(FileId, FileName, ChecksumAsBytes, 0);
```
AntonLapounov, line 841: Nit: We might want to add a TODO comment to actually calculate the check sum.

kant2002:
I need LLVM 13 or manually port my favorite patch to LLVM https://github.com/llvm/llvm-project/commit/08a5ac36b956edeb989b4a65269a829eac26a5a2

kant2002:
Aah.  I see, that's without libunwind. Probably will try compile that branch on ARM32.

src/coreclr/tools/aot/ObjWriter/debugInfo/codeView/codeViewTypeBuilder.cpp
```diff
@@ -52,16 +52,16 @@ void UserDefinedCodeViewTypesBuilder::EmitTypeInformation(
...
+    StringRef S(reinterpret_cast<const char *>(Record.data().data()), Record.data().size());
+    Streamer->emitBinaryData(S);
   });
 }
 
 unsigned UserDefinedCodeViewTypesBuilder::GetEnumFieldListType(
     uint64 Count, const EnumRecordTypeDescriptor *TypeRecords) {
-  FieldListRecordBuilder FLRB(TypeTable);
-  FLRB.begin();
+  ContinuationRecordBuilder FLRB;
```
AntonLapounov: `FLRB` stands for `FieldListRecordBuilder`; we might want to update the names of corresponding locals and parameters.

src/coreclr/tools/aot/ObjWriter/README.md
```diff
@@ -0,0 +1,7 @@
+# Native Object Writer library
+
+This directory contains a native object file writer library based on LLVM. This library exposes C APIs to emit ELF/Mach-O/PE object files.
+
+The implementation is based on LLVM's assembler APIs - these are APIs intended to be consumed by assembly language compilers and are therefore close to the underlying object file formats. When in doubt, look at how the assemblers (e.g. llvm-ml in the tools directory of LLVM) use these APIs.
+
+The build script clones the appropriate version of the LLVM source tree, applies a patch that adds a couple things we need, and builds the repo with this directory added as an out-of-tree tool.
```
AntonLapounov: Nit: a couple _of_ things.

AntonLapounov (approved):
Looks great, thank you a lot!

kant2002:
> We could probably optimize the checkout to a sparse one to speed that up a little since we don't need clang, libc and friends but maybe in another pull request.

Right now the full checkout take too much space to my small 16Gb Pi. I have to cleanup a lot.  And not sure that would be enough space for the build. llvm-project is at least 1.2Gb and I restarting checkout again. It may happens that if this lands, it may halt my work on ARM.
Or I immediately start working on sparse checkout.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -14,7 +14,7 @@ parameters:
 steps:
 - ${{ if ne(parameters.archType, 'arm64') }}:
 
-  - ${{ if eq(parameters.buildConfig, 'Release') }}:
+  - ${{ if eq(parameters.buildConfig, 'ReleaseDoNotRun') }}:
```
jkotas: Was this meant to be a temporary change?

jkotas:
> Right now the full checkout take too much space to my small 16Gb Pi.

One more workaround: Submit WIP PR that enables cross-compiled Linux ARM build in CI. You can then use the CI system to build the object writer for you, and just download it from the artifacts.

kant2002:
Thanks. Will try that.

kant2002:
I was able to compile. 
16Gb was enough. 
Empty app compiles, but seg fault when run (so something changed).
Some new error location

Not sure, should it hold this PR and what's the reasons.

src/coreclr/tools/aot/ObjWriter/debugInfo/codeView/codeViewTypeBuilder.h
```diff
@@ -61,11 +62,11 @@ class UserDefinedCodeViewTypesBuilder : public UserDefinedTypesBuilder {
   unsigned GetEnumFieldListType(uint64 Count,
                                 const EnumRecordTypeDescriptor *TypeRecords);
 
-  void AddBaseClass(FieldListRecordBuilder &FLBR, unsigned BaseClassId);
-  void AddClassVTShape(FieldListRecordBuilder &FLBR);
+  void AddBaseClass(ContinuationRecordBuilder&CRB, unsigned BaseClassId);
+  void AddClassVTShape(ContinuationRecordBuilder&CRB);
```
AntonLapounov: Nit: Better to keep the space.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
MichalStrehovsky: Yup. We only run with a live objwriter in Release builds. But Windows Release CI hasn't been green in several days because libraries tests (that run before any other tests) are failing. I need the CI to validate this.

AntonLapounov:
I have tried Windows ARM64 and `bl` and `adrp/add` targets are not resolved. For instance, look at the code produced for `StartupCodeHelpers.InitializeModules`. Before:
```asm
00007ff6`8856ce6c 940000c5 bl          test!S_P_CoreLib_Internal_Runtime_CompilerHelpers_StartupCodeHelpers__CreateTypeManagers (00007ff6`8856d180)
00007ff6`8856ce70 aa0003f3 mov         x19,x0
00007ff6`8856ce74 93407e81 sxtw        x1,w20
00007ff6`8856ce78 b0000f60 adrp        x0,test!__Array<S_P_TypeLoader_Internal_Runtime_TypeLoader_GenericDictionaryCell>::`vftable'+0xb0 (00007ff6`88759000)
00007ff6`8856ce7c 911b8000 add         x0,x0,#0x6E0
00007ff6`8856ce80 97faa5ad bl          test!RhpNewArray (00007ff6`88416534)
```
After:
```asm
00007ff7`8f9c4efc 94000000 bl          test!S_P_CoreLib_Internal_Runtime_CompilerHelpers_StartupCodeHelpers__InitializeModules+0x4c (00007ff7`8f9c4efc)
00007ff7`8f9c4f00 aa0003f3 mov         x19,x0
00007ff7`8f9c4f04 93407e81 sxtw        x1,w20
00007ff7`8f9c4f08 90000000 adrp        x0,test!S_P_CoreLib_Internal_Runtime_CompilerServices_FunctionPointerOps___cctor (00007ff7`8f9c4000)
00007ff7`8f9c4f0c 91000000 add         x0,x0,#0
00007ff7`8f9c4f10 94000000 bl          test!S_P_CoreLib_Internal_Runtime_CompilerHelpers_StartupCodeHelpers__InitializeModules+0x60 (00007ff7`8f9c4f10)
```
Note that all offsets are zero.

MichalStrehovsky:
> I have tried Windows ARM64 and `bl` and `adrp/add` targets are not resolved.

Thank you! Looks like the problem was that LLVM came up with their own `AArch64AsmBackend::getFixupKind` that differed from the one in the old llvm.patch. The LLVM's one looked more complete so I took it, but it only works for ELF because it aparently uses the "literal relocs" as opposed to "target-specific relocs". The literal relocs use the literal reloc numbering from the output object file format. The fix was to use the original llvm.patch implementation outside ELF.

I've also tweaked things in the objwriter to assert if this ever happens again.

Includes @kant2002 fix for ARM32 (Thanks!)

Fingers crossed these are the last issues!

AntonLapounov:
Hello-world does work on Windows ARM64 now.

src/coreclr/tools/aot/ObjWriter/llvm.patch
```diff
@@ -454,26 +153,11 @@ index 31762b9e4cd50e429725a3f32323cc5a555dfe2f..d53ab3e3b61c2fd73e7333682289d310
...
  
-diff --git a/lib/Target/ARM/MCTargetDesc/ARMAsmBackend.cpp b/lib/Target/ARM/MCTargetDesc/ARMAsmBackend.cpp
-index a77df7a2598f4176ccfda98c72c35471a679285b..e1aa7526f9b07205ca27ab0489e11a6081bcf8e9 100644
---- a/lib/Target/ARM/MCTargetDesc/ARMAsmBackend.cpp
-+++ b/lib/Target/ARM/MCTargetDesc/ARMAsmBackend.cpp
-@@ -48,6 +48,14 @@ public:
- };
- } // end anonymous namespace
- 
-+Optional<MCFixupKind> ARMAsmBackend::getFixupKind(StringRef Name) const {
```
AntonLapounov, line 465: Should we keep this as well instead of the change in `ObjectWriter::EmitSymbolRef`?

AntonLapounov:
I am getting a stream of linker errors while trying to build hello-world on Linux ARM64:
```
  obj/debug/net5.0/linux-arm64/native/test.o: In function `S_P_TypeLoader_System_Reflection_Runtime_General_QHandle____GetFieldHelper':
  (__managedcode+0x50): relocation truncated to fit: R_AARCH64_ADR_PREL_LO21 against symbol `vtable for Object' defined in .data section in obj/debug/net5.0/linux-arm64/native/test.o
```
Are you sure the fix in `AArch64AsmBackend::getFixupKind` should be Windows-only?

AntonLapounov:
Hello-world works on Linux ARM64 if the fix in `AArch64AsmBackend::getFixupKind` is applied to ELF as well.

kant2002:
@AntonLapounov can you show what's in the output of after you use fix for Linux + Win implementation.
```
objdump -r obj/debug/net5.0/linux-arm64/native/test.o | grep RhpReversePInvoke2
```
Even better if somehow you extract all 3 (?) types of relocations from obj file. I suspect that we are using incorrect names for relocations. So if we change them we can use LLVM code without patches.

Maybe for last problem you can use
```
objdump -r obj/debug/net5.0/linux-arm64/native/test.o | grep "vtable for Object"
```

MichalStrehovsky:
I think @kant2002 is right - the name we use for the reloc is wrong. We say "`R_AARCH64_ADR_PREL_LO21` maps to `fixup_aarch64_pcrel_adrp_imm21`", but `fixup_aarch64_pcrel_adrp_imm21` actually maps to `R_AARCH64_ADR_PREL_PG_HI21`.

I'll check if the same problem is in the other relocs. It's all very confusing.

I spent half a day trying to see if we can get away without these names (the names are actually names of the ELF relocs, so it's weird using those names for PE/Mach-O, but LLVM doesn't have the names hooked up for anything else), but it would require a lot more refactoring - the actual target specific reloc enums are only available in lib/Target. They're not supposed to be visible to end users because they're used in instruction encoding. But we don't use instruction encoding APIs. So the only option we have left is to use this LLVM API that is meant to support `.reloc X, Y, Z` assembler directives (and those take opaque strings from source code where the string basically means the reloc name for the given output file format).

AntonLapounov:
@kant2002 Before my fix (Linux ARM64):
```
RELOCATION RECORDS FOR [__managedcode]:
OFFSET           TYPE              VALUE
0000000000000050 R_AARCH64_ADR_PREL_LO21  _ZTV6Object
0000000000000054 R_AARCH64_ADD_ABS_LO12_NC  _ZTV6Object
...
00000000000997e0 R_AARCH64_JUMP26  RhpReversePInvoke2
```
After the fix:
```
RELOCATION RECORDS FOR [__managedcode]:
OFFSET           TYPE              VALUE
0000000000000050 R_AARCH64_ADR_PREL_PG_HI21  _ZTV6Object
0000000000000054 R_AARCH64_ADD_ABS_LO12_NC  _ZTV6Object
...
00000000000997e0 R_AARCH64_CALL26  RhpReversePInvoke2
```

MichalStrehovsky:
I've pushed out an update that appears to fix ARM64 (at least statically it looks okay now). I've renamed the textual strings we use to refer to the reloc to match the actual ELF reloc we need to emit (Anton's dump seems to confirm it's the right reloc name).

PR #1462, opened 2021-08-26
Collection tests for dfa

veanes:
Support for Regex.Replace in DFA mode. (this support was missing before)
Added some new unit tests for collections, replace and constructor.
Lifted old unit tests for ctor, replace, and collections to include DFA mode.
This is part of extending more unit tests to DFA mode.
Found and fixed a bug related to incorrect use of match length 
when match was not started from the beginning of the input.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
@@ -53,7 +53,7 @@ namespace System.Text.RegularExpressions
         /// <param name="hideStateInfo">if true then hide state info</param>
         /// <param name="addDotStar">if true then pretend that there is a .* at the beginning</param>
         /// <param name="inReverse">if true then unwind the regex backwards (addDotStar is then ignored)</param>
-        /// <param name="onlyDFAinfo">if true then compute and save only genral DFA info</param>
+        /// <param name="onlyDFAinfo">if true then compute and save only general DFA info</param>
         /// <param name="writer">dgml output is written here</param>
         /// <param name="maxLabelLength">maximum length of labels in nodes anything over that length is indicated with .. </param>
         internal void SaveDGML(TextWriter writer, int bound, bool hideStateInfo, bool addDotStar, bool inReverse, bool onlyDFAinfo, int maxLabelLength)
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
@@ -92,6 +92,7 @@ namespace System.Text.RegularExpressions
         /// Generates two files IgnoreCaseRelation.cs and UnicodeCategoryRanges.cs for the namespace System.Text.RegularExpressions.SRM.Unicode
         /// in the given directory path. Only avaliable in DEBUG mode.
         /// </summary>
+        [Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage(Justification = "Debug only")]
         internal static void GenerateUnicodeTables(string path)
         {
             SRM.Unicode.IgnoreCaseRelationGenerator.Generate("System.Text.RegularExpressions.SRM.Unicode", "IgnoreCaseRelation", path);
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
@@ -102,7 +102,7 @@ namespace System.Text.RegularExpressions
                 // Ignore Compiled flag if DFA is used
                 // this is to make sure the pattern is not being compiled as well as being used in SRM
                 options = options & ~RegexOptions.Compiled;
-            } 
+            }
             this.pattern = pattern;
             internalMatchTimeout = matchTimeout;
             roptions = options;
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
@@ -96,6 +96,12 @@ namespace System.Text.RegularExpressions
             ValidateOptions(options);
             ValidateMatchTimeout(matchTimeout);
 
+            _useSRM = (options & RegexOptions.DFA) != 0;
+            if (_useSRM)
+                // Ignore Compiled flag if DFA is used
+                // this is to make sure the pattern is not being compiled as well as being used in SRM
+                options = options & ~RegexOptions.Compiled;
+
             this.pattern = pattern;
             internalMatchTimeout = matchTimeout;
             roptions = options;
@@ -113,24 +119,19 @@ namespace System.Text.RegularExpressions
             // Extract the relevant information
             capnames = tree.CapNames;
             capslist = tree.CapsList;
+            // code related info is relevant in DFA mode only regarding caps
+            // that are used in RegexReplacement to check absense of subtitutions
+            _code = RegexWriter.Write(tree);
+            caps = _code.Caps;
+            capsize = _code.CapSize;
+
+            InitializeReferences();
 
             // If SRM is used then construct the SMR.Regex matcher.
             // This construction fails and throws a NotSupportedException
             // if constructs that are not compatible with DFA are being used in the pattern.
-            _useSRM = (options & RegexOptions.DFA) != 0;
             if (_useSRM)
-            {
                 _srm = InitializeSRM(tree.Root, roptions & ~RegexOptions.DFA, matchTimeout, culture);
-            }
-            else
-            {
-                // code related info is only set in non-DFA mode
-                _code = RegexWriter.Write(tree);
-                caps = _code.Caps;
-                capsize = _code.CapSize;
-            }
-
-            InitializeReferences();
         }
 
         /// <summary>
@@ -147,9 +148,6 @@ namespace System.Text.RegularExpressions
             // TBD: this could also be supported easily, but is not of priority right now
             if ((options & RegexOptions.ECMAScript) != 0)
                 throw new NotSupportedException(SRM.Regex._DFA_incompatible_with + RegexOptions.ECMAScript);
-            // TBD: this will eventually be supported
-            if ((options & RegexOptions.Compiled) != 0)
-                throw new NotSupportedException(SRM.Regex._DFA_incompatible_with + RegexOptions.Compiled);
 
             return SRM.Regex.Create(rootNode, options, matchTimeout, culture);
         }
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
@@ -120,7 +120,7 @@ namespace System.Text.RegularExpressions
             capnames = tree.CapNames;
             capslist = tree.CapsList;
             // code related info is relevant in DFA mode only regarding caps
-            // that are used in RegexReplacement to check absense of subtitutions
+            // that are used in RegexReplacement to check absence of subtitutions
             _code = RegexWriter.Write(tree);
             caps = _code.Caps;
             capsize = _code.CapSize;
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
@@ -110,9 +110,9 @@ namespace System.Text.RegularExpressions
             // Parse the input
             RegexTree tree = RegexParser.Parse(pattern, roptions, culture ?? ((options & RegexOptions.CultureInvariant) != 0 ? CultureInfo.InvariantCulture : CultureInfo.CurrentCulture));
 
-            // if SRM is used then construct the SMR.Regex matcher
-            // this construction fails and throws a NotSupportedException
-            // if unsupported constructs are being used in the original regex
+            // If SRM is used then construct the SMR.Regex matcher.
+            // This construction fails and throws a NotSupportedException
+            // if constructs that are not compatible with DFA are being used in the original regex.
             _useSRM = (options & RegexOptions.DFA) != 0;
             if (_useSRM)
             {
@@ -126,9 +126,9 @@ namespace System.Text.RegularExpressions
                 _code = RegexWriter.Write(tree);
                 caps = _code.Caps;
                 capsize = _code.CapSize;
-
-                InitializeReferences();
             }
+
+            InitializeReferences();
         }
 
         /// <summary>
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
@@ -98,10 +98,11 @@ namespace System.Text.RegularExpressions
 
             _useSRM = (options & RegexOptions.DFA) != 0;
             if (_useSRM)
+            {
                 // Ignore Compiled flag if DFA is used
                 // this is to make sure the pattern is not being compiled as well as being used in SRM
                 options = options & ~RegexOptions.Compiled;
-
+            } 
             this.pattern = pattern;
             internalMatchTimeout = matchTimeout;
             roptions = options;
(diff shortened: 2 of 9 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
```diff
@@ -92,6 +92,7 @@ private Regex(SRM.Regex srm)
         /// Generates two files IgnoreCaseRelation.cs and UnicodeCategoryRanges.cs for the namespace System.Text.RegularExpressions.SRM.Unicode
         /// in the given directory path. Only avaliable in DEBUG mode.
         /// </summary>
+        [Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage(Justification = "Debug only")]
```
danmoseley, line 95: Nit, no need to fix today,  but we would add a using statement. `[ExcludeFromCodeCoverage(Justification = "Debug only")]`

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
```diff
@@ -96,6 +96,12 @@ private void Init(string pattern, RegexOptions options, TimeSpan matchTimeout, C
             ValidateOptions(options);
             ValidateMatchTimeout(matchTimeout);
 
+            _useSRM = (options & RegexOptions.DFA) != 0;
+            if (_useSRM)
+                // Ignore Compiled flag if DFA is used
+                // this is to make sure the pattern is not being compiled as well as being used in SRM
+                options = options & ~RegexOptions.Compiled;
+
```
danmoseley:
```suggestion
            if (_useSRM)
            {
                // Ignore Compiled flag if DFA is used
                // this is to make sure the pattern is not being compiled as well as being used in SRM
                options = options & ~RegexOptions.Compiled;
            } 
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
```diff
@@ -110,25 +116,22 @@ private void Init(string pattern, RegexOptions options, TimeSpan matchTimeout, C
...
 
-            // if SRM is used then construct the SMR.Regex matcher
-            // this construction fails and throws a NotSupportedException
-            // if unsupported constructs are being used in the original regex
-            _useSRM = (options & RegexOptions.DFA) != 0;
+            // Extract the relevant information
+            capnames = tree.CapNames;
+            capslist = tree.CapsList;
+            // code related info is relevant in DFA mode only regarding caps
+            // that are used in RegexReplacement to check absense of subtitutions
```
danmoseley:
```suggestion
            // that are used in RegexReplacement to check absence of subtitutions
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexReplacement.cs
```diff
@@ -114,6 +120,9 @@ private struct FourStackStrings // used to do the equivalent of: Span<string> st
             if (!replRef.TryGetTarget(out repl) || !repl.Pattern.Equals(replacement))
             {
                 repl = RegexParser.ParseReplacement(replacement, roptions, caps, capsize, capnames);
+                if (repl.HasGroups && ((roptions & RegexOptions.DFA) != 0))
+                    throw new NotSupportedException(SRM.Regex._DFA_incompatible_with + "replacement pattern with substitutions");
```
danmoseley, line 124: Please move _DFA_incompatible_with and any other user visible strings into the .resx and then you can use the `SR.XXX` syntax to load it (or `SR.Format(...,)`)

You will probably need one string here, and one that you can reuse for all the incompatible regex options, eg.,
`SR.Format(SR.DfaIncompatibleWith, RegexOptions.Compiled)`

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexReplacement.cs
```diff
@@ -213,7 +222,30 @@ public string Replace(Regex regex, string input, int count, int startat)
...
+
+                while (match.Success)
+                {
+                    if (state.prevat < match.Index)
+                        state.segments.Add(state.inputMemory.Slice(state.prevat, match.Index - state.prevat));
+                    state.prevat = match.Index + match.Length;
+                    ReplacementImpl(ref state.segments, match);
+
+                    if (state.count > 0) state.count -= 1;
+                    if (state.count == 0) break;
```
danmoseley, line 239:
```suggestion
                    if (--state.count <= 0) break;
```
maybe (and above)

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
```diff
@@ -10,6 +10,16 @@ namespace System.Text.RegularExpressions.Tests
     public class RegexReplaceTests
     {
         public static IEnumerable<object[]> Replace_String_TestData()
+        {
+            foreach (var data in Replace_String_TestData_Core())
+            {
+                yield return data;
+                if (!((string)data[0]).Contains(@"G") && !((string)data[2]).Contains("$") &&
```
danmoseley: this is pretty ugly. ~~Can we just have two enumerables, Replace_String_TestData and Replace_String_TestData_Not_DFA ? and tests can use one or both as needed (it's OK for a test to have multiple `[MemberData]` on it). Then no need for this code nor a wrapper around both enumerables.~~

also the string is `@"\Ga"` so should the G be expected in `data[1]` ?

danmoseley: scratch that. I mean, you have two enumerables, and this wrapper here enumerates over all of Replace_String_TestData  (which works for both) then over it again with `| RegexSRMTests.DFA`, and then over Replace_String_TestData_Not_DFA without  `| RegexSRMTests.DFA`

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
```diff
@@ -144,6 +154,18 @@ public void Replace(string pattern, string input, string replacement, RegexOptio
...
 
         public static IEnumerable<object[]> Replace_MatchEvaluator_TestData()
+        {
+            foreach (var data in Replace_MatchEvaluator_TestData_Core())
+            {
+                yield return data;
+                // Duplicate the data for DFA mode
+                // but ignore both RightToLeft and Compiled as these cases are either not supported or ignored in DFA mode
+                // also avoid \b for now
+                if ((((RegexOptions)data[3]) & (RegexOptions.RightToLeft| RegexOptions.Compiled)) == 0 && !((string)data[0]).Contains(@"\b"))
```
danmoseley: same comment -- suggest two clearly named enumerables rather than sniffing and filtering individual entries.

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
```diff
@@ -315,6 +345,9 @@ public void Replace_Invalid()
 
             AssertExtensions.Throws<ArgumentOutOfRangeException>("startat", () => new Regex("pattern").Replace("input", "replacement", 0, 6));
             AssertExtensions.Throws<ArgumentOutOfRangeException>("startat", () => new Regex("pattern").Replace("input", new MatchEvaluator(MatchEvaluator1), 0, 6));
+
+            // Sustitutions not supported in DFA mode
```
danmoseley:
```suggestion
            // Substitutions not supported in DFA mode
```

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
```diff
@@ -332,5 +365,57 @@ public static string MatchEvaluator3(Match match)
...
+        [InlineData("[ab]+", "012aaabb34bba56", "###", 0, "012aaabb34bba56")]
+        [InlineData("[ab]+", "012aaabb34bba56", "###", -1, "012###34###56")]
+        [InlineData("[ab]+", "012aaabb34bba56", "###", 1, "012###34bba56")]
+        [InlineData(@"\b", "Hello World!", "#", 2, "#Hello# World!")]
+        [InlineData(@"\b", "Hello World!", "#$$#", -1, "#$#Hello#$# #$#World#$#!")]
+        [InlineData(@"", "hej", "  ", -1, "  h  e  j  ")]
+        [InlineData(@"\bis\b", "this is it", "${2}", -1, "this ${2} it")]
+        private void TestReplaceCornerCases(string pattern, string input, string replacement, int count, string expectedoutput)
+        {
+            RegexOptions[] testoptions = new RegexOptions[] { RegexSRMTests.DFA, RegexOptions.None, RegexOptions.Compiled };
```
danmoseley: this is probably better using `[MemberData]`  -- an enumerator over the 7 inputs you have in member data, wrapped in an enumerator over the 3 options. the reason that is better is that then xunit treats each combinatoin as its own test and reports on it individually.

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
```diff
@@ -332,5 +365,57 @@ public static string MatchEvaluator3(Match match)
...
+
+        [Theory]
+        [InlineData(@"(\$\d+):(\d+)", "it costs $500000:55 I think", "$$???:${2}", "it costs $???:55 I think")]
+        [InlineData(@"(\d+)([a-z]+)", "---12345abc---", "$2$1", "---abc12345---")]
+        private void TestReplaceWithSubstitution(string pattern, string input, string replacement, string expectedoutput)
+        {
+            RegexOptions[] testoptions = new RegexOptions[] { RegexOptions.None, RegexOptions.Compiled };
+            foreach (var opt in testoptions)
+            {
+                var output = new Regex(pattern, opt).Replace(input, replacement, -1);
```
danmoseley, line 440: incidentally, why do we accept -1 as the count in this API?
we even have an error string `Count cannot be less than -1.`
yet we aren't even truthful elsewhere:
```c#
73              if (count < 0)
   74              {
   75:                 ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.count, ExceptionResource.CountTooSmall);
   76              }
```
does it do anything different from 0? and why do we accept 0 either, just as a convenience?

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
```diff
@@ -332,5 +365,57 @@ public static string MatchEvaluator3(Match match)
...
+            Assert.Throws<NotSupportedException>(() => Regex.Replace(input, pattern, replacement, RegexSRMTests.DFA));
+            Assert.Throws<NotSupportedException>(() => new Regex(pattern, RegexSRMTests.DFA).Replace(input, replacement, -1));
+        }
+
+        [Theory]
+        [InlineData(@"(\bis\b)", "this is it", "this IS it")]
+        private void TestReplaceWithToUpperMatchEvaluator(string pattern, string input, string expectedoutput)
+        {
+            MatchEvaluator f = new MatchEvaluator(m => m.Value.ToUpper());
+            RegexOptions[] testoptions = new RegexOptions[] { RegexSRMTests.DFA, RegexOptions.None, RegexOptions.Compiled };
```
danmoseley: @stephentoub for these cases where there are different inputs that are multiplexed over different options, do you favor nested enumerators for MemberData? or just internally loop like here?

seems like when there's just a few variations it's not worth the hassle of member data, but if there's a lot then it might be nice to have xunit break them out as separate tests.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -23,12 +23,38 @@ static void WriteLine(string s)
...
         }
 
-        internal static RegexOptions DFA = (RegexOptions)0x400;
+        internal const RegexOptions DFA = (RegexOptions)0x400;
 
         private const char Turkish_I_withDot = '\u0130';
         private const char Turkish_i_withoutDot = '\u0131';
         private const char Kelvin_sign = '\u212A';
 
+        //[Theory]
```
danmoseley: We usually use `[ActiveIssue]` instead of commenting out. That way the code stays compiling.

Shouldn't the first test run for the other engine at least?

danmoseley (approved):

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
veanes: sounds good, I'll clean this up

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
veanes: ok, will do

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
veanes: -1 means there is no bound, 0 is just no replacements allowed (useless bound)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
veanes: sounds good, this code is used to, generate/update the pregenerated serialized BDD relations, so the code is in fact definitely covered

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexReplacement.cs
veanes: Ok. This applies to the whole SRM code base, I think there should be a separate code cleanup item to clean up all error messages and to create proper .resx entries.

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
veanes: sounds good, done, added separate Replace_String_TestData_DFA. Indeed most of the cases were irrelevant for DFA anyways because they used substitutions.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexReplacement.cs
veanes: the case >= 0 means there is an upper bound. -1 means there is no bound.

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
danmoseley: ah, we should probably doc that. and at some point fix the Regex.Split case to use a string that says "Count cannot be less than 0".

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
veanes: added separate test enumerator for DFA mode, also extracted out one case as ActiveIssue

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: I uncommented  these as you suggest and then rewrote/simplified to be more to the point. 
Marked one case as ActiveIssue (the one that was failing for None).
The point the active issue is making that Regex.IsMatch(" AB\u200cCD ", @"\b\w+\b") should succeed no matter what interpretation \u200c is given (\w or \W).

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
veanes: sounds good, that is in the public call user must provide a number >= 0.

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
veanes: fixed those, using `MemberData` attribute.

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
veanes: fixed, using `MemberData`

veanes:
The remaining item is fixing proper handling of error messages such as _DFA_incompatible_with_ .
This applies to the whole SRM code base. Fixed other things. Will merge if all checks complete successfully.

PR #1464, opened 2021-08-26
Disable COM specific marshalers on Unix

jkotas:
Match CoreCLR behavior and prevent compiler from crashing

### DIFF
--- a/src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
+++ b/src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
@@ -266,7 +266,10 @@ namespace Internal.TypeSystem.Interop
                                 return MarshallerKind.CBool;
 
                             case NativeTypeKind.VariantBool:
-                                return MarshallerKind.VariantBool;
+                                if (type.Context.Target.IsWindows)
+                                    return MarshallerKind.VariantBool;
+                                else
+                                    return MarshallerKind.Invalid;
 
                             default:
                                 return MarshallerKind.Invalid;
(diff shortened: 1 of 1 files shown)


jkotas:
`[DllImport("abc")] extern static void f(object o);` was crashing the compiler. Kennedy hit this issue when trying to compile for Linux.

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
```diff
@@ -266,7 +266,10 @@ public static partial class MarshalHelpers
                                 return MarshallerKind.CBool;
 
                             case NativeTypeKind.VariantBool:
-                                return MarshallerKind.VariantBool;
+                                if (type.Context.Target.IsWindows)
```
MichalStrehovsky:
```suggestion
                                if (context.Target.IsWindows)
```

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
```diff
@@ -580,12 +583,18 @@ public static partial class MarshalHelpers
             {
                 if (nativeType == NativeTypeKind.AsAny)
                     return isAnsi ? MarshallerKind.AsAnyA : MarshallerKind.AsAnyW;
-                else if ((isField && nativeType == NativeTypeKind.Default)
-                    || nativeType == NativeTypeKind.Intf
-                    || nativeType == NativeTypeKind.IUnknown)
-                    return MarshallerKind.ComInterface;
                 else
-                    return MarshallerKind.Variant;
+                if (type.Context.Target.IsWindows)
```
MichalStrehovsky:
```suggestion
                if (context.Target.IsWindows)
```

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
```diff
@@ -641,7 +650,10 @@ public static partial class MarshalHelpers
             }
             else if (type.IsInterface)
             {
-                return MarshallerKind.ComInterface;
+                if (type.Context.Target.IsWindows)
```
MichalStrehovsky:
```suggestion
                if (context.Target.IsWindows)
```

MichalStrehovsky (approved):
Thank you!

PR #1467, opened 2021-08-27
NativeAOT: Only mark debugger step in/through points on Windows

TheSpydog:
Disabling the DebuggerSteppingHelpers on non-Windows fixes a bug where breakpoints were not working correctly with the Switch debugger. The helpers are only useful on Windows anyway, so this should be a safe change.


MichalStrehovsky (approved):
Thank you!

PR #1468, opened 2021-08-29
Specify ObjWriter build configuration in single location

kant2002:


MichalStrehovsky:
I'm planning to submit a pull request for https://github.com/MichalStrehovsky/runtimelab/commit/3706eb5657f3f830329ec1919521a25d04c9e365 once the LLVM update is merged so that objwriter builds in the same configuration as the rest of the bits. I think it will address the thing you're after with this pull request.

I think the only reason why we hardcoded ObjWriter to always build Release was that the debug build was broken due to LLVM devs [buying too much into the "modern C++" garbage and hitting static initialization corner case bugs in the MSVC toolchain](https://bugs.llvm.org/show_bug.cgi?id=41367). We picked up fixes for that (first in llvm.patch, now in LLVM proper) and the problem no longer exists.

PR #1472, opened 2021-08-30
Bug fix of dfa lazy dot star

veanes:
Fixed bug of .*? being treated as .* in DFA mode that came up testing of Replace and was related to and fixed an active issue.
Added more unit tests for lazy loops that then discovered another active issue that needs to be fixed where an empty match exists
and where there is no start-set and match fails for DFA (perhaps related to use of lazy loops, but it is not clear if that is so).

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
@@ -37,6 +37,43 @@ namespace System.Text.RegularExpressions.Tests
             Assert.True(Regex.IsMatch(" AB\u200dCD ", @"\b\w+\b"));
         }
 
+        public static IEnumerable<object[]> TestLazyLoops_Data()
+        {
+            foreach (var options in new RegexOptions[] { RegexOptions.Singleline, RegexOptions.Compiled | RegexOptions.Singleline, DFA | RegexOptions.Singleline })
+            {
+                yield return new object[] { options, "seq 012 of 3 digits", @"\W.*?\D", true, 3, 5 };
+                yield return new object[] { options, "seq 012 of 3 digits", @"\W.+?\D", true, 3, 5 };
+                yield return new object[] { options, "seq 012 of 3 digits", @"\W.{1,7}?\D", true, 3, 5 };
+                yield return new object[] { options, "seq 012 of 3 digits", @"\W.{1,2}?\D", true, 7, 3 };
+                yield return new object[] { options, "digits:0123456789", @"\W.*?\b", true, 6, 1 };
+                yield return new object[] { options, "e.g:abc", @"\B.*?\B", true, 5, 0 };
+                yield return new object[] { options, "e.g:abc", @"\B\W+?", false, 0, 0 };
+            }
+            yield return new object[] { RegexOptions.Singleline, "e.g:abc", @"\B\W*?", true, 5, 0 };
+            yield return new object[] { RegexOptions.Compiled | RegexOptions.Singleline, "e.g:abc", @"\B\W*?", true, 5, 0 };
+        }
+
+        [Theory]
+        [MemberData(nameof(TestLazyLoops_Data))]
+        public void TestLazyLoops(RegexOptions options, string input, string pattern, bool success, int index, int length)
+        {
+            Match m = Regex.Match(input, pattern, options);
+            Assert.Equal(success, m.Success);
+            Assert.Equal(index, m.Index);
+            Assert.Equal(length, m.Length);
+        }
+
+        [Theory]
+        [InlineData(DFA, "e.g:abc", @"\B\W*?", true, 5, 0)]
+        [ActiveIssue("bug with use of startset finding no match while there is an empty match")]
+        public void TestLazyLoops_AvtiveIssue(RegexOptions options, string input, string pattern, bool success, int index, int length)
+        {
+            Match m = Regex.Match(input, pattern, options);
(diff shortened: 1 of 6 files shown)


src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -37,6 +37,43 @@ public void TestBoundary()
...
+            Match m = Regex.Match(input, pattern, options);
+            Assert.Equal(success, m.Success);
+            Assert.Equal(index, m.Index);
+            Assert.Equal(length, m.Length);
+        }
+
+        [Theory]
+        [InlineData(DFA, "e.g:abc", @"\B\W*?", true, 5, 0)]
+        [ActiveIssue("bug with use of startset finding no match while there is an empty match")]
+        public void TestLazyLoops_AvtiveIssue(RegexOptions options, string input, string pattern, bool success, int index, int length)
```
danmoseley:
```suggestion
        public void TestLazyLoops_ActiveIssue(RegexOptions options, string input, string pattern, bool success, int index, int length)
```

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -37,6 +37,43 @@ public void TestBoundary()
             Assert.True(Regex.IsMatch(" AB\u200dCD ", @"\b\w+\b"));
         }
 
+        public static IEnumerable<object[]> TestLazyLoops_Data()
```
danmoseley: presumably we want to end up with only SRM-specific tests (which are hopefully relatively few) in the RegexSRMTests.cs file. 

could clean up in a followup if you prefer.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -37,6 +37,43 @@ public void TestBoundary()
...
+                yield return new object[] { options, "seq 012 of 3 digits", @"\W.*?\D", true, 3, 5 };
+                yield return new object[] { options, "seq 012 of 3 digits", @"\W.+?\D", true, 3, 5 };
+                yield return new object[] { options, "seq 012 of 3 digits", @"\W.{1,7}?\D", true, 3, 5 };
+                yield return new object[] { options, "seq 012 of 3 digits", @"\W.{1,2}?\D", true, 7, 3 };
+                yield return new object[] { options, "digits:0123456789", @"\W.*?\b", true, 6, 1 };
+                yield return new object[] { options, "e.g:abc", @"\B.*?\B", true, 5, 0 };
+                yield return new object[] { options, "e.g:abc", @"\B\W+?", false, 0, 0 };
+            }
+            yield return new object[] { RegexOptions.Singleline, "e.g:abc", @"\B\W*?", true, 5, 0 };
+            yield return new object[] { RegexOptions.Compiled | RegexOptions.Singleline, "e.g:abc", @"\B\W*?", true, 5, 0 };
```
danmoseley: when we fix the DFA bug we just adjust these two lines so they're inside the loop above, right? we would not have the TestLazyLoops_ActiveIssue test.

if so maybe that test is not worth having, and instead you just want to make sure there's an issue open to fix this. you could put the link to the issue here if you like.

danmoseley: in fact, should these just go into Match_Basic_TestData? no need for a new test. all variations can get merged in there.

danmoseley (approved):
I'm not able to review the SRM changes yet, but if the tests pass..

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: Right, I've been making some tests apply to None and Compiled as well each time, to get more confidence in expected result. Thus, they've become not just DFA tests. In the end those test could indeed be moved to where they best belong -- I would say -- as a follow-up cleanup. In some cases I'm not sure where they best belong.

danmoseley: looks like most or all of RegexSRMTests.cs is relevant to the other engine. And more importantly than where it lives, it would be good to make those tests run against the existing engine as another test against discrepancies.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: I can also just leave the test where it was and let it fail until fixed. I was planning to do it next actually. Most probably part of this pr. Then I'll just put it back where it was -- change the TestLazyLoops_Data back to what it was initially when I added the test.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: I'll go over the tests after the bug fix and generalize them. There are some very specific DFA test though.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: I'll move the test to Match_Basic_TestData.

veanes:
Added Olli as an extra pair of eyes also to doublecheck the last edit that disables use of start-set for nullable regexes. It is a small change that should have no noticeable perf implications (typical regexes are not nullable - do not accept empty matches).

PR #1477, opened 2021-08-31
Include canonical forms of types involved in GVM dispatch

### LINKED ISSUE #1473: [NativeAOT] Generic virtual method pointer lookup failure
@MichalStrehovsky Follow-up for issue #1455

Repro:

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
    <Nullable>enable</Nullable>
    <!-- Removing these next two causes the bug to go away -->
    <IlcDisableReflection>true</IlcDisableReflection>
    <IlcGenerateStackTraceData>false</IlcGenerateStackTraceData>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="7.0.0-*" />
  </ItemGroup>

</Project>
```

Code

```csharp
var list = new List<string>() { "ABC", "X", "XYZ", "UV", "XA" };
var ordered = list.OrderBy(s => s.Length).ThenBy(s => s.GetHashCode());
foreach (var s in ordered)
{
    Console.WriteLine(s);
}
```

MichalStrehovsky:
The GVM analysis within the compiler assumes we can do analysis on top of canonical forms and we get analysis holes if the canonical form doesn't exist at compile time.

GVM dispatch at runtime might end up building new types as well, so the extra template isn't strictly just bloat but I do admit there's some laziness in just throwing the extra template in to fix the problem.

This wasn't hittable outside reflection-free mode because we generate canonical forms of everything outside reflection-free mode anyway.

As I was looking for a place to put this in, I realized the `TypeNeedsGVMTableEntries` logic needs to run on canonical types as well (those are not `ConstructedEETypeNode`), so I moved it to a common spot. That part was not necessary to fix the reported problem, but it provides a convenient spot to put the canonical type dependency.

Fixes #1473.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ConstructedEETypeNode.cs
@@ -73,14 +73,6 @@ namespace ILCompiler.DependencyAnalysis
                 }
             }
 
-            // Generated type contains generic virtual methods that will get added to the GVM tables
-            if (TypeGVMEntriesNode.TypeNeedsGVMTableEntries(_type))
-            {
-                dependencyList.Add(new DependencyListEntry(factory.TypeGVMEntries(_type.GetTypeDefinition()), "Type with generic virtual methods"));
-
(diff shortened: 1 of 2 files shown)


jkotas (approved):
Thanks you

PR #1478, opened 2021-08-31
fixed deep recursion crash in SRM

veanes:
Conversion of deeply nested concatenations of RegexNodes to Symbolic Regexes was handled poorly/naively and caused crash (most probably out-of-stack-memory) when pattern has nested concats inside captures, say 1000-fold deep as in the added unit test. Discovered this by lifting over a related stress test that was used for Replace (but that initially failed earlier and used substitutions, so was out of scope because of the focus on substitutions), but now revisited and fixed the core issue behind the earlier crash.

Also added a cache for already created predicates from internal strings representing intervals and character classes, that in this unit test below gets called 1000x but exposed again the problem of repeated nontrivial computations that were not cached. This should improve construction time considerably for larger patterns. The cache itself is local to a regex matcher and should not be a concern.

Added the unit test StressTestSymbolicRegexConversion (this test would have crashed earlier).

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/tests/Regex.GetGroupNames.Tests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/Regex.GetGroupNames.Tests.cs
@@ -114,13 +114,31 @@ namespace System.Text.RegularExpressions.Tests
                 new int[] { 0, 15 },
                 new string[] { "Ryan Byington", "Byington" }
             };
+
+            yield return new object[]
+            {
+                "(?'15'\\S+)\\s(?'15'\\S+)", "Ryan Byington",
+                new string[] { "0" },
+                new int[] { 0 },
+                new string[] { "Ryan Byington" },
+                RegexSRMTests.DFA
+            };
+
+            yield return new object[]
+            {
+                "(?<first_name>\\S+)\\s(?<last_name>\\S+)", "Ryan Byington",
+                new string[] { "0" },
+                new int[] { 0 },
+                new string[] { "Ryan Byington" },
+                RegexSRMTests.DFA
+            };
         }
 
         [Theory]
         [MemberData(nameof(GroupNamesAndNumbers_TestData))]
-        public void GroupNamesAndNumbers(string pattern, string input, string[] expectedNames, int[] expectedNumbers, string[] expectedGroups)
+        public void GroupNamesAndNumbers(string pattern, string input, string[] expectedNames, int[] expectedNumbers, string[] expectedGroups, RegexOptions options = RegexOptions.None)
         {
-            Regex regex = new Regex(pattern);
+            Regex regex = new Regex(pattern, options);
             Match match = regex.Match(input);
             Assert.True(match.Success);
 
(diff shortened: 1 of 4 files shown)


danmoseley:
Another case where a smart fuzzer or other stress apparatus might help us (?)

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -1001,5 +1001,14 @@ public static IEnumerable<object[]> ValidateSRMRegex_NotSupportedCases_Data()
             yield return new object[] { @"\G(\w+\s?\w*),?", RegexOptions.None, "contiguous matches" };
             yield return new object[] { @"(?>a*).", RegexOptions.None, "atomic" };
         }
+
+        [Fact]
+        public void StressTestSymbolicRegexConversion()
+        {
+            string pattern = string.Concat(Enumerable.Repeat("([a-z]", 1000).Concat(Enumerable.Repeat(")", 1000)));
```
danmoseley, line 1011: does this fail on NFA? If not shall we make it generic to both

src/libraries/System.Text.RegularExpressions/tests/Regex.GetGroupNames.Tests.cs
```diff
@@ -9,10 +9,11 @@ namespace System.Text.RegularExpressions.Tests
     public class GetGroupNamesTests
     {
         [Theory]
-        [InlineData("(?<first_name>\\S+)\\s(?<last_name>\\S+)", new string[] { "0", "first_name", "last_name" })]
-        public void GetGroupNames(string pattern, string[] expectedGroupNames)
+        [InlineData("(?<first_name>\\S+)\\s(?<last_name>\\S+)", RegexOptions.None, new string[] { "0", "first_name", "last_name" })]
```
danmoseley, line 12: maybe Compiled also?

danmoseley (approved):
As usual I'm only really reviewing tests..

src/libraries/System.Text.RegularExpressions/tests/Regex.GetGroupNames.Tests.cs
veanes: I can add that here but in many other cases the code being tested is really independent of options (only dependent if the DFA option is used, in those cases I started using options as the last parameter whose default value is None unless provided otherwise, in which cases I don't have to change the existing data generators, just add extra  ones with last parameter being the DFA option at the end)

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: Variant of this one already appears as part of Replace tests (in Replace_String_TestData). So it is covered in backtracking case as well. We had an SRM meeting today with Olli and discussed that over time we should collect and add more such stress tests, say with deeply nested loops etc -- as robustness tests on the initial parsing/conversion phase, also for the BDD algebra.

veanes: I added None and Compiled in any case because the test here is slightly different.

veanes:
merging at this point, moving on with other changes.

PR #1479, opened 2021-09-01
Export debug header with a pragma

MichalStrehovsky:
Updating my ZeroSharp samples and this is getting in the way. The fix we did in https://github.com/dotnet/runtimelab/pull/743 was insufficient (I want things to be debuggable in ZeroSharp).


jkotas (approved):
LGTM

PR #1480, opened 2021-09-01
Rename RegexOptions.DFA to RegexOptions.NonBacktracking

stephentoub:
And fix execution of tests on .NET Framework.


veanes (approved):
Looks good to me. I'll start using RegexHelpers.RegexOptionNonBacktracking going forward in new edits/unit tests I've been working on. Probably comments might use DFA occasionally as a synonym -- well this is true for many places in the code in any case.

PR #1481, opened 2021-09-01
Fix compiling with Test.CoreLib

MichalStrehovsky:
* We have new exports that are needed.
* Unshare IDynamicInterfaceCastableSupport. I thought it would be shareable, but it uses SR and throws various exceptions. Maybe not so shareable after all.


jkotas (approved):
Thanks

PR #1486, opened 2021-09-02
Initial round of clean up on Regex SRM

stephentoub:
I started going through the SRM code to understand it, and along the way cleaned things up to be more in line with dotnet/runtime style and practices.  There's still a lot more of that to be done, but this is as far as I got between meetings today.

- Replaced exceptions thrown only in DEBUG with Debug.Asserts
- Avoid a lot of temporary strings by writing to an existing builder
- Moved some exception messages into resources
- Replace uses of string.Split with span slicing
- Removed unused namespaces
- Changed string.Formats to use string interpolation where relevant
- Sealed types
- Made fields readonly
- Renamed fields with _ and s_ naming
- Replaced some hashcode creationg with HashCode.Combine
- Use expression body syntax in more places
- Cleaned up some brace usage, e.g. always using braces for a block if another branch uses braces
- Use compound and null coalescing operators where relevant
- Delete unused code
- Using explicit types instead of var
...


veanes (approved):

src/libraries/System.Text.RegularExpressions/src/Resources/Strings.resx
```diff
@@ -225,4 +225,31 @@
...
     <value>Unterminated (?#...) comment.</value>
   </data>
+  <data name="NotSupported_NonBacktrackingAndReplacementsWithSubstitutions" xml:space="preserve">
+    <value>Regex replacements with substitutions are not supported with RegexOptions.NonBacktracking.</value>
+  </data>
+  <data name="NotSupported_NonBacktrackingConflictingOption" xml:space="preserve">
+    <value>RegexOptions.NonBacktracking is not supported in conjunction with RegexOptions.{0}.</value>
+  </data>
+  <data name="NotSupported_NonBacktrackingConflictingExpression" xml:space="preserve">
+    <value>RegexOptions.NonBacktracking is not supported in conjunction with expressions containing {0}.</value>
```
danmoseley, line 235: as an aside, this is concatenating sentence fragments, which is not localizable. they either need to follow a period, or be in quotes.

stephentoub: We can add single quotes.  It's not meant to be sentence fragments, but rather a description of the thing not supported.  It could be `: '{0}'` if that's better.

danmoseley: that would work fine.

PR #1488, opened 2021-09-02
Implement Control Flow Guard annotations

MichalStrehovsky:
This is a set of changes required to enable control flow guard enforcement within the process. Control flow guard is a security mitigation feature that validates that indirect calls only land on addresses that are valid targets of indirect calls. It has two parts: identifying valid targets of indirect calls within the process, and checking whether target of indirect call is valid before dispatching to it.

This implements annotations and enforcement within the unmanaged parts of the NativeAOT runtime and the annotation-only part for the managed code. Enforcement will follow later.

Three kinds of changes:

* A new version of Runtime.lib that enables `/guard:cf` flag. This is in addition to the existing libraries since we don't want code to pay the perf penalty if CFG is not enabled.
* Annotating methods as valid CFG targets in the AOT compiler and object file writer.
* MSBuild support for new `<ControlFlowGuard>Guard</ControlFlowGuard>` property that enables all of this (passes a switch to the AOT compiler, selects the guarded version of runtime libraries to link with, and passes a switch to link.exe to enable CFG for the process).

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
@@ -19,6 +19,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <CppLibCreator>lib</CppLibCreator>
     <FullRuntimeName>Runtime</FullRuntimeName>
     <FullRuntimeName Condition="'$(ServerGarbageCollection)' != ''">Runtime.ServerGC</FullRuntimeName>
+    <FullRuntimeName Condition="'$(ControlFlowGuard)' == 'Guard'">Runtime.ServerGC.GuardCF</FullRuntimeName>
     <EntryPointSymbol Condition="'$(EntryPointSymbol)' == ''">wmainCRTStartup</EntryPointSymbol>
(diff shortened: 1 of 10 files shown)


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
```diff
@@ -19,6 +19,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <CppLibCreator>lib</CppLibCreator>
     <FullRuntimeName>Runtime</FullRuntimeName>
     <FullRuntimeName Condition="'$(ServerGarbageCollection)' != ''">Runtime.ServerGC</FullRuntimeName>
+    <FullRuntimeName Condition="'$(ControlFlowGuard)' == 'Guard'">Runtime.ServerGC.GuardCF</FullRuntimeName>
```
jkotas, line 22: Do we also need GuardCF version of the corefx .libs (globalization, compression)?

jkotas: (It can be follow up PR.)

jkotas (approved):
Thanks!

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Windows.props
MichalStrehovsky: Ah yes. We also need it for the bootstrapper.

PR #1489, opened 2021-09-03
Use sparse checkout for LLVM

MichalStrehovsky:
Shaves down the clone/checkout to ~45 seconds.

### DIFF
--- a/src/coreclr/tools/aot/ObjWriter/build.sh
+++ b/src/coreclr/tools/aot/ObjWriter/build.sh
@@ -8,6 +8,8 @@ TargetArch="$4"
 BuildType="${5:-Release}"
 CompilerId="$(echo "$6" | tr "[:upper:]" "[:lower:]")"
 
+LLVMBranch="release/12.x"
+
 # Check that we have enough arguments
 if [ $# -lt 4 ]; then
     echo "Usage: $(basename $0) ArtifactsDir RepoRoot BuildArch TargetArch [BuildType [CompilerId]]"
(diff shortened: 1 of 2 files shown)


jkotas (approved):
Nice!

src/coreclr/tools/aot/ObjWriter/build.sh
```diff
@@ -37,8 +39,13 @@ PatchApplied=0
...
 if [ ! -d llvm-project ]; then
     # Clone the LLVM repo
-    git clone --depth 1 -b release/12.x https://github.com/llvm/llvm-project.git || exit 1
-    cd llvm-project/llvm || exit 1
+    git clone --no-checkout --depth 1 -b $LLVMBranch https://github.com/llvm/llvm-project.git || exit 1
+    cd llvm-project || exit 1
+    # Purposefuly ignoring exit codes of sparse-checkout so that this works with git lower than 2.26
+    git sparse-checkout init
+    git sparse-checkout set /llvm/\* !/llvm/test/\*
+    git checkout $LLVMBranch || exit 1
```
AntonLapounov, line 47: Not worth restarting the build, but it is a good practice to always quote variables:
```suggestion
    git clone --no-checkout --depth 1 -b "$LLVMBranch" https://github.com/llvm/llvm-project.git || exit 1
    cd llvm-project || exit 1
    # Purposefuly ignoring exit codes of sparse-checkout so that this works with git lower than 2.26
    git sparse-checkout init
    git sparse-checkout set /llvm/\* !/llvm/test/\*
    git checkout "$LLVMBranch" || exit 1
```

AntonLapounov (approved):
Looks great, thank you!

PR #1490, opened 2021-09-03
Fix thread-safety of timeouts in SymbolicRegexMatcher

stephentoub:


veanes (approved):
When I wrote this code, if I recall correctly I mimicked what happens in backtracking code, but I do not recall the exact details of handling of the timeoutOccursAt info in that case, perhaps good to doublecheck.

PR #1492, opened 2021-09-03
Commenting and code cleanup for SRM's algebras

olsaarik:
MintermGenerator in particular got a lot of comments, since it is a non-trivial algorithm.

BDDAlgebra likely needs more comments, but I'm delaying that until after I've introduced the ReaderWriterLock there.

BDDRangeConverter is only partially commented, the rest will come soon.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -328,11 +328,50 @@ namespace System.Text.RegularExpressions.SRM
                 return arcs[0] == 0 ? False : True;
 
             if (algebra is null)
-                return Deserialize_(arcs, MkBDD);
+                return DeserializeWithFactory(arcs, MkBDD);
 
             lock (algebra)
             {
-                return Deserialize_(arcs, algebra.MkBDD);
+                return DeserializeWithFactory(arcs, algebra.MkBDD);
+            }
+
+            static BDD DeserializeWithFactory(long[] arcs, Func<int, BDD, BDD, BDD> mkBDD)
+            {
+                // the number of bits used for ordinals and node identifiers are stored in the first two values
+                int k = arcs.Length;
+                int ordinal_bits = (int)arcs[0];
+                int node_bits = (int)arcs[1];
+                // create bit masks for the sizes of ordinals and node identifiers
+                long ordinal_mask = (1 << ordinal_bits) - 1;
+                long node_mask = (1 << node_bits) - 1;
+                BitLayout(ordinal_bits, node_bits, out int zero_node_shift, out int one_node_shift, out int ordinal_shift);
+
+                // store BDD nodes by their id when they are created
+                BDD[] nodes = new BDD[k];
+                nodes[0] = False;
+                nodes[1] = True;
+
+                for (int i = 2; i < k; i++)
+                {
+                    long arc = arcs[i];
+                    if (arc <= 0)
+                    {
+                        // this is an MTBDD leaf. Its ordinal was serialized negated
+                        nodes[i] = mkBDD((int)-arc, null, null);
+                    }
+                    else
+                    {
+                        // reconstruct the ordinal and child identifiers for a non-terminal
+                        int ord = (int)((arc >> ordinal_shift) & ordinal_mask);
+                        int oneId = (int)((arc >> one_node_shift) & node_mask);
+                        int zeroId = (int)((arc >> zero_node_shift) & node_mask);
+                        // the BDD nodes for the children are guaranteed to exist already due to the topological order
+                        nodes[i] = mkBDD(ord, nodes[oneId], nodes[zeroId]);
+                    }
+                }
+
+                //the result is the final BDD in the nodes array
+                return nodes[k - 1];
             }
         }
 
@@ -342,45 +381,6 @@ namespace System.Text.RegularExpressions.SRM
             Debug.Assert(one is not null || zero is null);
             Debug.Assert(one is null || zero is not null);
             return new BDD(ordinal, one, zero);
-        }
-
-        private static BDD Deserialize_(long[] arcs, Func<int, BDD, BDD, BDD> mkBDD)
-        {
-            // the number of bits used for ordinals and node identifiers are stored in the first two values
-            int k = arcs.Length;
-            int ordinal_bits = (int)arcs[0];
-            int node_bits = (int)arcs[1];
-            // create bit masks for the sizes of ordinals and node identifiers
-            long ordinal_mask = (1 << ordinal_bits) - 1;
-            long node_mask = (1 << node_bits) - 1;
-            BitLayout(ordinal_bits, node_bits, out int zero_node_shift, out int one_node_shift, out int ordinal_shift);
-
-            // store BDD nodes by their id when they are created
-            BDD[] nodes = new BDD[k];
-            nodes[0] = False;
-            nodes[1] = True;
-
-            for (int i = 2; i < k; i++)
-            {
-                long arc = arcs[i];
-                if (arc <= 0)
-                {
-                    // this is an MTBDD leaf. Its ordinal was serialized negated
-                    nodes[i] = mkBDD((int)-arc, null, null);
-                }
-                else
-                {
-                    // reconstruct the ordinal and child identifiers for a non-terminal
-                    int ord = (int)((arc >> ordinal_shift) & ordinal_mask);
-                    int oneId = (int)((arc >> one_node_shift) & node_mask);
-                    int zeroId = (int)((arc >> zero_node_shift) & node_mask);
-                    // the BDD nodes for the children are guaranteed to exist already due to the topological order
-                    nodes[i] = mkBDD(ord, nodes[oneId], nodes[zeroId]);
-                }
-            }
-
-            //the result is the final BDD in the nodes array
-            return nodes[k - 1];
         }
 
         /// <summary>
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -328,11 +328,11 @@ namespace System.Text.RegularExpressions.SRM
                 return arcs[0] == 0 ? False : True;
 
             if (algebra is null)
-                return DeserializeWithFactory(arcs, MkBDD);
+                return DeserializeWithFactory(arcs, CreateBDD);
 
             lock (algebra)
             {
-                return DeserializeWithFactory(arcs, algebra.MkBDD);
+                return DeserializeWithFactory(arcs, algebra.GetOrCreateBDD);
             }
 
             static BDD DeserializeWithFactory(long[] arcs, Func<int, BDD, BDD, BDD> mkBDD)
@@ -375,7 +375,7 @@ namespace System.Text.RegularExpressions.SRM
             }
         }
 
-        private static BDD MkBDD(int ordinal, BDD one, BDD zero)
+        private static BDD CreateBDD(int ordinal, BDD one, BDD zero)
         {
             Debug.Assert(one != zero || one is null);
             Debug.Assert(one is not null || zero is null);
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -11,7 +11,7 @@ namespace System.Text.RegularExpressions.SRM
     /// nodes have an Ordinal, which indicates the position of the bit the node relates to (0 for the least significant
     /// bit), and two children, One and Zero, for the cases of the current bit being 1 or 0, respectively. An integer
     /// belongs to the set represented by the BDD if the path from the root following the branches that correspond to
-    /// the bits of the integer leads to the True leaf. This class also supports multi-terminal BDDs, i.e. ones where
+    /// the bits of the integer leads to the True leaf. This class also supports multi-terminal BDDs (MTBDD), i.e. ones where
     /// the leaves are something other than True or False, which are used for representing classifiers.
     /// </summary>
     internal sealed class BDD : IComparable
@@ -202,7 +202,7 @@ namespace System.Text.RegularExpressions.SRM
             while (toVisit.Count > 0)
             {
                 BDD node = toVisit.Pop();
-                // true and False are not included in the result
+                // True and False are not included in the result
                 if (node.IsFull || node.IsEmpty)
                     continue;
 
@@ -262,21 +262,21 @@ namespace System.Text.RegularExpressions.SRM
             Debug.Assert(nodes.Length <= (1 << 24));
 
             // As few bits as possible are used to for ordinals and node identifiers for compact serialization.
-            // use at least a nibble (4 bits) to represent the ordinal and count how many are needed
+            // Use at least a nibble (4 bits) to represent the ordinal and count how many are needed.
             int ordinal_bits = 4;
             while (Ordinal >= (1 << ordinal_bits))
             {
                 ordinal_bits += 1;
             }
 
-            // use at least 2 bits to represent the node identifier and count how many are needed
+            // Use at least 2 bits to represent the node identifier and count how many are needed
             int node_bits = 2;
             while (nodes.Length >= (1 << node_bits))
             {
                 node_bits += 1;
             }
 
-            // reserve space for all nodes plus 2 extra: index 0 and 1 are reserved for False and True
+            // Reserve space for all nodes plus 2 extra: index 0 and 1 are reserved for False and True
             long[] res = new long[nodes.Length + 2];
             res[0] = ordinal_bits;
             res[1] = node_bits;
@@ -292,7 +292,7 @@ namespace System.Text.RegularExpressions.SRM
                 [False] = 0
             };
 
-            // give all nodes ascending identifiers and produce their serializations into the result
+            // Give all nodes ascending identifiers and produce their serializations into the result
             for (int i = 0; i < nodes.Length; i++)
             {
                 BDD node = nodes[i];
@@ -306,7 +306,7 @@ namespace System.Text.RegularExpressions.SRM
                 }
                 else
                 {
-                    // combine ordinal and child identifiers according to the bit layout
+                    // Combine ordinal and child identifiers according to the bit layout
                     long v = (((long)node.Ordinal) << ordinal_shift) | (idmap[node.One] << one_node_shift) | (idmap[node.Zero] << zero_node_shift);
                     Debug.Assert(v >= 0);
                     res[i + 2] = v; // children ids are well-defined due to the topological order of nodes
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
@@ -190,8 +190,7 @@ namespace System.Text.RegularExpressions.SRM
             if (IsLeaf)
                 return new BDD[] { this };
 
-            //order the nodes according to their ordinals
-            //into the nonterminals array
+            // Order the nodes according to their ordinals into the nonterminals array
             var nonterminals = new List<BDD>[Ordinal + 1];
             var sorted = new List<BDD>();
             var toVisit = new Stack<BDD>();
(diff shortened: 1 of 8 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -7,18 +7,34 @@
 namespace System.Text.RegularExpressions.SRM
 {
     /// <summary>
-    /// Represents a Binary Decision Diagram.
+    /// Represents nodes in a Binary Decision Diagram (BDD), which compactly represent sets of integers. All non-leaf
+    /// nodes have an Ordinal, which indicates the position of the bit the node relates to (0 for the least significant
+    /// bit), and two children, One and Zero, for the cases of the current bit being 1 or 0, respectively. An integer
+    /// belongs to the set represented by the BDD if the path from the root following the branches that correspond to
+    /// the bits of the integer leads to the True leaf. This class also supports multi-terminal BDDs, i.e. ones where
```
stephentoub: Nit: later on you abbreviate multi-terminal BDDs as MTBDD, so consider adding a `(MTBDD)` parenthetical here.
```suggestion
    /// the bits of the integer leads to the True leaf. This class also supports multi-terminal BDDs (MTBDD), i.e. ones where
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -194,90 +193,90 @@ public BDD[] TopSort()
...
 
-            stack.Push(this);
+            toVisit.Push(this);
 
-            while (stack.Count > 0)
+            while (toVisit.Count > 0)
             {
-                BDD node = stack.Pop();
+                BDD node = toVisit.Pop();
+                // true and False are not included in the result
```
stephentoub:
```suggestion
                // True and False are not included in the result
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -194,90 +193,90 @@ public BDD[] TopSort()
...
                 if (node.IsFull || node.IsEmpty)
                     continue;
 
                 if (node.IsLeaf)
                 {
-                    nodes.Add(node);
+                    // MTBDD terminals can be directly added to the sorted nodes
```
stephentoub: Can you add a parenthetical saying why?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -194,90 +193,90 @@ public BDD[] TopSort()
...
 
-            BDD[] nodes = TopSort();
+            BDD[] nodes = TopologicalSort();
 
             Debug.Assert(nodes[nodes.Length - 1] == this);
             Debug.Assert(nodes.Length <= (1 << 24));
 
-            //use fewer bits when possible, starting with nibble size
+            // As few bits as possible are used to for ordinals and node identifiers for compact serialization.
+            // use at least a nibble (4 bits) to represent the ordinal and count how many are needed
```
stephentoub:
```suggestion
            // Use at least a nibble (4 bits) to represent the ordinal and count how many are needed.
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -194,90 +193,90 @@ public BDD[] TopSort()
...
             int ordinal_bits = 4;
             while (Ordinal >= (1 << ordinal_bits))
             {
                 ordinal_bits += 1;
             }
 
-            //use as few bits as possible for node identifiers, starting with 2 bits
-            //this will give smaller and more compact serialized representations
+            // use at least 2 bits to represent the node identifier and count how many are needed
```
stephentoub:
```suggestion
            // Use at least 2 bits to represent the node identifier and count how many are needed
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -194,90 +193,90 @@ public BDD[] TopSort()
...
             int node_bits = 2;
             while (nodes.Length >= (1 << node_bits))
             {
                 node_bits += 1;
             }
 
-            //add 2 extra positions: index 0 and 1 are reserved for False and True
+            // reserve space for all nodes plus 2 extra: index 0 and 1 are reserved for False and True
```
stephentoub:
```suggestion
            // Reserve space for all nodes plus 2 extra: index 0 and 1 are reserved for False and True
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -293,22 +292,24 @@ public long[] Serialize()
                 [False] = 0
             };
 
+            // give all nodes ascending identifiers and produce their serializations into the result
```
stephentoub:
```suggestion
            // Give all nodes ascending identifiers and produce their serializations into the result
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -293,22 +292,24 @@ public long[] Serialize()
...
                 {
-                    //this is MTBDD leaf: negate the value (it may be 0)
-                    //because True and False are excluded from TopSort()
+                    // This is MTBDD leaf. Negating it should make it less than or equal to zero, as True and False are
+                    // excluded here and MTBDD Ordinals are required to be non-negative.
                     res[i + 2] = -node.Ordinal;
                 }
                 else
                 {
+                    // combine ordinal and child identifiers according to the bit layout
```
stephentoub:
```suggestion
                    // Combine ordinal and child identifiers according to the bit layout
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
```diff
@@ -345,13 +346,16 @@ private static BDD MkBDD(int ordinal, BDD one, BDD zero)
 
         private static BDD Deserialize_(long[] arcs, Func<int, BDD, BDD, BDD> mkBDD)
```
stephentoub: This could be made into a local function within Deserialize

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDDAlgebra.cs
```diff
@@ -249,6 +251,9 @@ private BDD MkBinBoolOP_rec(BoolOp op, BDD a, BDD b)
                     if (b == True)
                         return MkNot_rec(a);
                     break;
+                default:
+                    Debug.Assert(false, "Unhandled binary BoolOp case");
```
stephentoub:
```suggestion
                    Debug.Fail("Unhandled binary BoolOp case");
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDDAlgebra.cs
```diff
@@ -303,7 +308,7 @@ private BDD MkNot_rec(BDD a)
                 return neg;
 
             neg = a.IsLeaf ?
-                MkBDD(CombineTerminals(BoolOp.NOT, a.Ordinal, 0), null, null) : //muti-terminal case
+                MkBDD(CombineTerminals(BoolOp.NOT, a.Ordinal, 0), null, null) : // multi-terminal case
```
stephentoub: Is the "Mk" prefix short for "Make"?  Could we spell it out, or use "Create" or something a little more standard?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
```diff
@@ -8,19 +8,26 @@
 namespace System.Text.RegularExpressions.SRM
 {
     /// <summary>
-    /// Represents a bitvector of given Length (number of bits).
+    /// Represents a bitvector of arbitrary length (i.e. number of bits).
```
stephentoub, line 11: Could we use `BitArray` instead?  If not, what would it take to make `BitArray` suitable?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
```diff
@@ -159,59 +156,34 @@ public static BV Mk(params bool[] bits)
...
-            if (_hashcode == 0)
+            // if the hash code hasn't been calculated yet, do so before returning it
+            if (_hashcode == null)
             {
                 _hashcode = Length.GetHashCode();
                 for (int i = 0; i < _blocks.Length; i++)
                 {
-                    _hashcode = (_hashcode << 1) ^ _blocks[i].GetHashCode();
+                    _hashcode = HashCode.Combine(_hashcode, _blocks[i].GetHashCode());
                 }
```
stephentoub, line 184: Could this be:
```C#
HashCode c = default;
c.Add(Length);
c.AddBytes(MemoryMarshal.AsBytes(_blocks));
_hashcode = c.ToHashCode();
```
?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV64Algebra.cs
```diff
@@ -131,15 +136,19 @@ public BDD ConvertToCharSet(ICharAlgebra<BDD> solver, ulong pred)
             if (_partition == null)
                 throw new NotImplementedException(nameof(ConvertToCharSet));
 
+            // the result will be the union of all minterms in the set
```
stephentoub, line 139: We probably need a glossary somewhere in the code that describes what various terms mean in the context used in the source: minterms, algebra, etc.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/MintermGenerator.cs
```diff
@@ -41,236 +42,262 @@ public MintermGenerator(IBooleanAlgebra<TPredicate> ba)
...
             {
-                foreach (PartitonTree<TPredicate> leaf in _left.GetLeaves())
-                    yield return leaf;
-
-                foreach (PartitonTree<TPredicate> leaf in _right.GetLeaves())
-                    yield return leaf;
+                if (_left != null)
+                    foreach (PartitionTree<TPredicate> leaf in _left.GetLeaves())
+                        yield return leaf;
+                if (_right != null)
```
stephentoub, line 299:
```suggestion

                if (_right != null)
```

stephentoub (approved):

stephentoub:
Thanks!

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
olsaarik: To clarify, is the style to use for comments "always capitalized, but lone single-line comments without a period"?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDDAlgebra.cs
olsaarik: Yes, it is short for Make. I have no strong preference between Make or Create. If Create feels least surprising to you that could be a good choice.

The prefix is also used in the IBooleanAlgebra interface for logical operations, but unlike `MkBDD` those aren't guaranteed to be "making" anything. Rather, depending on the algebra in question they might be operating on value types (BV64Algebra) or sometimes just returning something from a cache (BDDAlgebra), so the prefix might be a bit misleading. The prefix also isn't used quite consistently, e.g. the following line uses the BDDAlgebra, but the additional function `ShiftLeft` supplied by the BDDAlgebra doesn't use the prefix.
```
_IgnoreCaseRel_inv = _solver.MkAnd(inv_table, _solver.MkNot(_solver.ShiftLeft(tr_I_withdot_BDD, 16)));
```
Any suggestions here? One option would be to drop the Mk prefix and just have that line be:
```
_IgnoreCaseRel_inv = _solver.And(inv_table, _solver.Not(_solver.ShiftLeft(tr_I_withdot_BDD, 16)));
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
stephentoub: Space between the // and the start and starting with a captial.

We're inconsistent about whether things end with a period or not.

As for line length, we're also inconsistent, though for borderline cases we typically err on the side of longer comment lines rather than more lines.  I've not counted our average line length, but it's probably a bit more than the 80 of yesteryear :)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDDAlgebra.cs
stephentoub: > One option would be to drop the Mk prefix and just have that line be

I like that (I think :)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
olsaarik: Yes, I think `BitArray` should be used instead. The whole `BV` class could probably get removed and BVAlgebra could implement `ICharAlgebra<BitArray>` instead. @veanes do you remember if there's any code that depends on the exact representation of `BV`?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
olsaarik: Great, thanks! I set my editor to wrap at 120, which should fit into GitHub code views.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV64Algebra.cs
olsaarik: Would likely be helpful. For minterms the MintermGenerator class' XML comment also includes a short description/example, which other XML comments could link to with `<see cref="MintermGenerator{TPredicate}"/>`. Visual Studio at least seems to make that a clickable link.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BDD.cs
veanes: The reason is that adding leaves here preserves the topological order of nodes. 
The topological order being that a child precedes its parent in the nodes list.
I will not edit it myself but Olli could add that comment.

olsaarik: Done!

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
veanes: No, BV is used instead of BV64 (i.e. ulong) when there are more than 64 parts in the partition (that almost never happens but could and there are unit tests that trigger its use). 
But other than that, the details of what the type actually is should never leak through. Replacing BV by BitArray implies reimplementing all the Boolean operations in the algebra: Bitwise-OR, Bitwise-AND, complement, etc... all the interface methods of the algebra. Then providing a convenienf pretty printer etc. ... So some amount of programming needed, but should be pretty straightforward.

**One important aspect being that BitArray can be use as a threadsafe valuetype (that might be a dealbraker here). I highly doubt also that is will be more efficient than BV. Usually BV will have say to (or at most 3) ulongs tat it consists ofe --- very low overhead**

veanes: I looked at the datatype. I don't think it would be useful to replace BV. What concerns me most is that BitArray is not a valuetype. Adding wrappers that make it a valuetype would defeat the purpose. For example implementing bitwise conjunction. With BV it is essentially bitwise-AND of say two ulongs (say when you have < 128 parts, which is covers all the cases I can think of).

olsaarik: I think the motivation is mostly to reduce the amount of code. The term type has to be used immutably, so memory allocation for operations is inevitable and I'd guess that's the most expensive thing here.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
veanes: Yes, this looks cleaner and more efficient. If I understand it correctly, the point here is to calculate a hashcode for a 
sequence (_blocks) of ulongs (typical case being _blocks.Length = 2 or at most 3).

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV64Algebra.cs
veanes: I can help Olli with some of these missing comments.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
olsaarik: I looked into the `BitArray` implementation and it's pretty much all the same logic as `BV`, except it uses 32-bit ints instead of 64-bit ones. It also does have code paths for vectorization, but those don't get triggered with less than 256 bits.

BitArray not being immutable is a concern. It would probably be best to have the actual term type be an opaque wrapper around BitArray that doesn't expose the mutating functions to other code, just so no-one in the future introduces code outside BVAlgebra that mutates the terms.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
olsaarik: Just a note, if we replace BV with BitVector we still need a separate hash code implementation since BitVector doesn't implement one.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
veanes: OK. Then I leave the decision up to you. If adding the wrapper for the immutable type and the operations is worth the effort, I agree and **am all for** reducing code size (i.e. getting rid of BV). 
First, there would need to be operations such as C = A & B where A, B, C are BitArrays and C is the resulting bitwise-AND of A and B where neither A nor B is changed. Currently such operations are missing in the datatype.

veanes: ... making **BV be the immutable wrapper** and basing its implementation on BitArray, is what I would suggest then.

veanes: ... Perhaps renaming it BitVector

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
veanes: I think so, yes.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/algebras/BV.cs
stephentoub: > I think the motivation is mostly to reduce the amount of code

Yup.  My goal is to reduce duplication, and wherever possible, centralize on shared implementations rather than having one-offs that need to be maintained and that don't benefit from any improvements to that centralized thing.  Additionally, as use cases present themselves that the current implementation may not be suitable for, it can drive us to determine what would enable it to be suitable, and then augment the centralized thing, making it better for everyone.

veanes: Just a note: The following two tests are (currently) the only two related unit test to be checked more closely later, that will be affected by this change: `SRMTest_BV `and `SRMTest_BV_WideLatin`. More unit tests could of course be added for more corner cases of BV. (e.g. adding a case reaching up to or more than 256 parts)

PR #1494, opened 2021-09-03
fixed failing DGML test and turned back serialize-deserialize in DEBUG

veanes:
One test was failing because when an SRM internal regex is serialized and then deserialized with the internal base64 coder the 
direct pretty-printing info of individual predicates and chars is lost and only preserved at an abstract level.
Relaxed the test to check this, and turned the serialization/deserializtion back on in DEBUG mode so that 
whenever an SRM internal regex is created it is run through the serialize/deserialize loop. 
The release build will of course not do that, 
so both forms (original one and ser-deser-version) are in fact being tested in the end.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
@@ -466,11 +466,23 @@ namespace System.Text.RegularExpressions.Tests
         {
             StringWriter sw = new StringWriter();
             var re = new Regex(".*a+", RegexOptions.NonBacktracking | RegexOptions.Singleline);
+            RegexExperiment.ViewDGML(re, name : "TestDGMLGeneration");
             SaveDGML(re, sw);
             string str = sw.ToString();
             Assert.StartsWith("<?xml version=\"1.0\" encoding=\"utf-8\"?>", str);
             Assert.Contains("DirectedGraph", str);
+#if DEBUG
+            // in debug mode re may be serialized and then deserialized internally 
+            // if that happens the predicate label .*a+ becomes .*[2]+
+            // the partition of characters in this regex is into two sets (in binary):
+            //   01 or [1] representing [^a]  (first part)
+            //   10 or [2] representinng 'a' (second part) which is where 2 comes from
+            //   (3rd part would be 100 = [4], 4th 1000 = [8] etc)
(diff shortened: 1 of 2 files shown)


src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -466,11 +466,23 @@ public void TestDGMLGeneration()
...
+            // if that happens the predicate label .*a+ becomes .*[2]+
+            // the partition of characters in this regex is into two sets (in binary):
+            //   01 or [1] representing [^a]  (first part)
+            //   10 or [2] representinng 'a' (second part) which is where 2 comes from
+            //   (3rd part would be 100 = [4], 4th 1000 = [8] etc)
+            // '.' here is the union of all parts, i.e. 01 and 10 that is 11 (in binary internally) but printed as '.' also.
+            Assert.True(str.Contains(".*a+") || str.Contains(".*[2]+"));
+#else
             Assert.Contains(".*a+", str);
+#endif
```
stephentoub: While it's not commonly done, technically there's no requirement that a debug test build use a debug src build.  Since the debug version of this is more accomodating, should we just keep that one and get rid of the else?

stephentoub (approved):

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: I think it is still useful to also check that the original regex be tested. The reason is that the deserialized clone gets rid of all caches for example, and restarts from a clean slate, and has **no dependency** on the original BDD solver.

stephentoub: Ok.  We do generally try to avoid having logic in tests that will fail if run against a different configuration (e.g. debug tests against release src, or vice versa), but we can leave it for now.

veanes: I realized I actually misunderstood you a bit, I thought you meant getting rid of the deserialize(serialize(regex)) in creating the regex in debug mode. What you meant is getting rid of the conditional in the test. Yes, I'll do that.

stephentoub: > What you meant is getting rid of the conditional in the test. Yes, I'll do that.

Exactly.  Thanks.

Issue #232, opened 2020-10-15
Investigate EETypePtr overhead


jkotas:
Context: https://github.com/dotnet/runtimelab/pull/225#discussion_r505246415

RuJIT is not able to optimize out the EETypePtr wrapper completely. In the example above the EETypePtr wrapper:
- Disables tailcall optimization. RyuJIT is not able to prove that the byrefs taken by EETypePtr are not passed into the worker method
- Results into less efficient register allocation and instruction selection (e.g. `cmp [rcx], rdx` vs. `mov rax, [rcx] cmp rax, rdx`)

jkotas:
Another factor to consider is code sharing with full CoreCLR. MethodTable and EEType are 80% the same between. We need a plan how to share the code that uses these. For example, the `Array.Copy` impl could have been shared except this difference.

EEType is a better name. Unfortunately, MethodTable is too entrenched in CoreCLR n all code, documentation and user facing surface (e.g. DumpMT command in SOS).

PR #1496, opened 2021-09-03
Speed up StElemRef if there's no covariance

MichalStrehovsky:
Makes storing into arrays 40% faster if there's no covariance.

I added `Object.MethodTableOf<T>` because `EETypePtrOf<T>` was causing a spill into a temporary and was messing up codegen and we'll want this for #232 anyway. I wanted to add `MethodTable.Of<T>`, but `Object` has a property named `MethodTable` and that was breaking the pattern. So it's on `Object`. Oh well.

### DIFF
--- a/src/coreclr/nativeaot/Runtime.Base/src/System/Object.cs
+++ b/src/coreclr/nativeaot/Runtime.Base/src/System/Object.cs
@@ -45,6 +45,9 @@ namespace System
             }
         }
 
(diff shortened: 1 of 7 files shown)


jkotas (approved):
Nice!

MichalStrehovsky:
@yowl this will need an adjustment on wasm side too - treat MethodTableOf same as EETypePtrOf (they're the same, just result is typed differently).

PR #1497, opened 2021-09-03
Clean up Regex SRM unicode folder

stephentoub:
And some code it touches.


veanes (approved):
It might be useful to add a unit test that checks that the updated Unicode table generator is correct -- a bit tricky here. The test should essentially run the code generator and then run the generated code that is responsible for generating the Unicode tables and then check that the generated tables are equivalent to the existing ones. This should all be possible through internally exposed helpers. If you think that is relevant I'll think how to expose that. But approving now.

stephentoub:
> It might be useful to add a unit test that checks that the updated Unicode table generator is correct

I agree that's a nice to have.  But at the same time, the very next time the src is built, won't all of the tests validate that it worked correctly?

veanes:
Not really, the code generator your cleanup touched is not run, that unit test is in fact missing. So what I'm mostly concerned about is making changes now to the generator and not rechecking its correctness -- although the cleanup you did seems all equivalent to what was there - it is still a change in generator-code that is being touched. The generator might not be needed, but if it is needed in say in 2 years (after say some Unicode categories are changed -- which happens now and then) it would be great that it actually runs. Right now now it should both run and produce equivalent tables.

stephentoub:
> Not really, the code generator your cleanup touched is not run, that unit test is in fact missing

It is, as long as you set the Enabled property to return true.

My point was simply that with a few clicks you can get a new version to replace the old, and build/test should tell you its correctness (if not, we're missing tests).

> So what I'm mostly concerned about is making changes now to the generator and not rechecking its correctness

Sure.  But changes to the generator don't matter until you take the output and apply it over the current code, at which point the very next build/test will validate its correctness.

I'm not saying we shouldn't invest in improving it, just that it doesn't strike me as a high priority.

If, however, our tests for regex in general are insufficient to validate that the tables built into the code are correct, that is higher priority to address.

veanes:
The current tests should be sufficient (even exhaustive for ignore-case relation: `TestIgnoreCaseRelation`) to check correctness of current tables,  the aspect of recreating (you are right this is the test: `RegenerateUnicodeTables`). The only step missing to "close the loop" is to validate that the generated tables are equal to the current ones, but this is indeed perhaps overkill (not high priority) -- The missing part is to test that the strings generated by  `RegenerateUnicodeTables` are equal to the existing strings in, 
`SRM.Unicode.IgnoreCaseRelation` and `SRM.Unicode.UnicodeCategoryRanges`.

veanes:
... Such additional test could be part of the existing RegenerateUnicodeTables test as additional asserts that would take care of the comparison of the generated strimgs to the existing static fields in the above classes. I can extend the test in such a manner.

PR #1505, opened 2021-09-04
Rename Runtime.lib to Runtime.WorkstationGC.lib

MichalStrehovsky:
"Runtime" as the target name is problematic because it's also the component name to build the CoreCLR-proper runtime.

There's CMakeLists entries that do things like `add_dependencies(runtime singlefilehost)` which were causing us to build a garbage single file host library that nobody needs in this branch.


jkotas (approved):
Good catch!

PR #1506, opened 2021-09-04
support for Regex.Split in NonBacktracking mode

veanes:
Implemented `Split `for `NonBacktracking `mode. 
Lifted all `Split `unit tests to the `NonBacktracking  `mode (with some adjustments because captures are not supported).
Added some new tests for (`None`, `Compiled`, and `NonBacktracking`) to verify correct/same behavior of unusual cases involving Split based on anchors (such tests were missing before).

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
@@ -109,7 +109,6 @@ namespace System.Text.RegularExpressions
                         break;
                     }
 
-                    // Omit the matched substring itself and find the next match
                     match = match.NextMatch();
                 }
 
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
@@ -87,7 +87,36 @@ namespace System.Text.RegularExpressions
             count--;
             (List<string> results, int prevat, string input, int count) state = (results: new List<string>(), prevat: 0, input, count);
 
-            if (!regex.RightToLeft)
+            if (regex._srm is not null)
+            {
+                Match match = regex.Match(input, startat, input.Length - startat);
+                if (!match.Success)
+                {
+                    // If there is no match then return the input as the only part in the split
+                    return new[] { input };
+                }
+
+                while (match.Success)
+                {
+                    // Add the substring before the found match, this may be empty
+                    state.results.Add(state.input.Substring(state.prevat, match.Index - state.prevat));
+                    state.prevat = match.Index + match.Length;
+
+                    if (--state.count == 0)
+                    {
+                        // If the maximum number of allowed mathes is reached break the splitting
+                        // If the initial value of count is 0 or negative this value just keeps decreasing further
+                        break;
+                    }
+
+                    // Omit the matched substring itself and find the next match
+                    match = match.NextMatch();
+                }
+
+                // Add the final suffix that either did not contain any matches or was cut off due to max split count being reached, this may be empty
+                state.results.Add(state.input.Substring(state.prevat, input.Length - state.prevat));
+            }
+            else if (!regex.RightToLeft)
             {
                 regex.Run(input, startat, ref state, static (ref (List<string> results, int prevat, string input, int count) state, Match match) =>
                 {
--- a/src/libraries/System.Text.RegularExpressions/tests/Regex.Split.Tests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/Regex.Split.Tests.cs
@@ -57,7 +57,59 @@ namespace System.Text.RegularExpressions.Tests
             yield return new object[] { @"(?<=\G..)(?=..)", "aabbccdd", RegexOptions.None, 8, 0, new string[] { "aa", "bb", "cc", "dd" } };
         }
 
+        // Test data addjusted from Split_TestData for RegexOptions.NonBacktracking
+        public static IEnumerable<object[]> Split_TestData_NonBacktracking()
+        {
+            yield return new object[] { "", "", RegexHelpers.RegexOptionNonBacktracking, 0, 0, new string[] { "", "" } };
+            yield return new object[] { "123", "abc", RegexHelpers.RegexOptionNonBacktracking, 3, 0, new string[] { "abc" } };
+
+            yield return new object[] { "    ", "word0    word1    word2    word3", RegexHelpers.RegexOptionNonBacktracking, 32, 0, new string[] { "word0", "word1", "word2", "word3" } };
+
+            yield return new object[] { ":", "kkk:lll:mmm:nnn:ooo", RegexHelpers.RegexOptionNonBacktracking, 19, 0, new string[] { "kkk", "lll", "mmm", "nnn", "ooo" } };
+            yield return new object[] { ":", "kkk:lll:mmm:nnn:ooo", RegexHelpers.RegexOptionNonBacktracking, 0, 0, new string[] { "kkk", "lll", "mmm", "nnn", "ooo" } };
+
+            // RegexOptions.NonBacktracking is similar to RegexOptions.ExplicitCapture when there are no explicit capture names because NonBacktracking does not support captures
+            // This is the reason for including the RegexOptions.ExplicitCapture options after the tests to check that the results are the same in this case
+            yield return new object[] { @"(\s)?(-)", "once -upon-a time", RegexHelpers.RegexOptionNonBacktracking, 17, 0, new string[] { "once", "upon", "a time" } };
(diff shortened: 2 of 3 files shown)


src/libraries/System.Text.RegularExpressions/tests/Regex.Split.Tests.cs
```diff
@@ -57,7 +57,59 @@ public static IEnumerable<object[]> Split_TestData()
...
+                yield return new object[] { @"\b", "Hello World!", options, 3, 6, new string[] { "Hello ", "World", "!" } };
+                yield return new object[] { @"\b", "Hello World!", options, 0, 0, new string[] { "", "Hello", " ", "World", "!" } };
+                yield return new object[] { @"^", "Hello \nWorld!", options | RegexOptions.Multiline, 0, 0, new string[] { "", "Hello \n", "World!" } };
+                yield return new object[] { @"$", "Hello \nWorld!", options | RegexOptions.Multiline, 0, 0, new string[] { "Hello ", "\nWorld!", "" } };
+            }
+        }
+
         [Theory]
+        [MemberData(nameof(Split_TestData_AnchorMatches))]
+        [MemberData(nameof(Split_TestData_NonBacktracking))]
```
stephentoub, line 119: I'm curious how you're running tests locally?  I do:
```C#
dotnet build /t:test
```
That will run the tests first on .NET Core and then on .NET Framework... I expect this will result in a lot of failures on .NET Framework.

I can keep fixing them up after the fact, but I'm curious why you're not seeing these failures locally.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
```diff
@@ -87,7 +87,36 @@ private static string[] Split(Regex regex, string input, int count, int startat)
...
+                    state.prevat = match.Index + match.Length;
+
+                    if (--state.count == 0)
+                    {
+                        // If the maximum number of allowed mathes is reached break the splitting
+                        // If the initial value of count is 0 or negative this value just keeps decreasing further
+                        break;
+                    }
+
+                    // Omit the matched substring itself and find the next match
```
stephentoub: What does "Omit the matched substring itself" mean here?

src/libraries/System.Text.RegularExpressions/tests/Regex.Split.Tests.cs
```diff
@@ -57,7 +57,59 @@ public static IEnumerable<object[]> Split_TestData()
             yield return new object[] { @"(?<=\G..)(?=..)", "aabbccdd", RegexOptions.None, 8, 0, new string[] { "aa", "bb", "cc", "dd" } };
         }
 
+        // Test data addjusted from Split_TestData for RegexOptions.NonBacktracking
```
stephentoub:
```suggestion
        // Test data adjusted from Split_TestData for RegexOptions.NonBacktracking
```

src/libraries/System.Text.RegularExpressions/tests/Regex.Split.Tests.cs
```diff
@@ -57,7 +57,59 @@ public static IEnumerable<object[]> Split_TestData()
...
+            yield return new object[] { "", "", RegexHelpers.RegexOptionNonBacktracking, 0, 0, new string[] { "", "" } };
+            yield return new object[] { "123", "abc", RegexHelpers.RegexOptionNonBacktracking, 3, 0, new string[] { "abc" } };
+
+            yield return new object[] { "    ", "word0    word1    word2    word3", RegexHelpers.RegexOptionNonBacktracking, 32, 0, new string[] { "word0", "word1", "word2", "word3" } };
+
+            yield return new object[] { ":", "kkk:lll:mmm:nnn:ooo", RegexHelpers.RegexOptionNonBacktracking, 19, 0, new string[] { "kkk", "lll", "mmm", "nnn", "ooo" } };
+            yield return new object[] { ":", "kkk:lll:mmm:nnn:ooo", RegexHelpers.RegexOptionNonBacktracking, 0, 0, new string[] { "kkk", "lll", "mmm", "nnn", "ooo" } };
+
+            // RegexOptions.NonBacktracking is similar to RegexOptions.ExplicitCapture when there are no explicit capture names because NonBacktracking does not support captures
+            // This is the reason for including the RegexOptions.ExplicitCapture options after the tests to check that the results are the same in this case
```
stephentoub: It'd be better here to avoid duplicating all the test entries, and instead do something like:
```C#
foreach (RegexOptions options in new[] { RegexHelpers.RegexOptionNonBacktracking, RegexOptions.ExplicitCapture })
{
    ... // each yield return using options
}
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
```diff
@@ -87,7 +87,36 @@ private static string[] Split(Regex regex, string input, int count, int startat)
             count--;
             (List<string> results, int prevat, string input, int count) state = (results: new List<string>(), prevat: 0, input, count);
 
-            if (!regex.RightToLeft)
+            if (regex._srm is not null)
```
stephentoub, line 90: Subsequently we should spend a little time seeing how we might better abstract away the implementations to not need these special cases sprinkled throughout the top-level APIs.  What you have here is good for now, though.

src/libraries/System.Text.RegularExpressions/tests/Regex.Split.Tests.cs
veanes: When I run them locally I just right-click on the test project and select Run Tests. So far this has worked well. So I did not know I need to do anything else special.

veanes: But also, all 8 checks have passed. I would have expected there to be some failures otherwise in the cloud build process.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
veanes: Well, the match itself is excluded from the output, but that I guess is the whole point of Split, so perhaps not such a helpful comment.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
veanes: I was thinking the same, but found it too difficult myself to do a deeper integration through the Run method of Regex. It really matters mostly for Replace and Split where things could be more unified I guess, they have similar top-level loops, but there are also some differences (mainly concerning wrt captures), and the new case is integrated similarly.

src/libraries/System.Text.RegularExpressions/tests/Regex.Split.Tests.cs
veanes: sounds good, will update.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
veanes: I will remove this comment.

src/libraries/System.Text.RegularExpressions/tests/Regex.Split.Tests.cs
veanes: More precisely. I select `Run Tests` from  the context menu of the `System.Text.RegularExpressions.Tests` project and then inspect the Tests output. With the latest updates for example it says:
```
========== Starting test run ==========
[xUnit.net 00:00:00.02] xUnit.net VSTest Adapter v2.4.2+2d84eb3141 (64-bit .NET 7.0.0-dev)
[xUnit.net 00:00:01.01]   Starting:    System.Text.RegularExpressions.Tests (parallel test collections = on, max threads = 56)
[xUnit.net 00:00:10.15]   Finished:    System.Text.RegularExpressions.Tests
========== Test run finished: 9168 Tests (9168 Passed, 0 Failed, 0 Skipped) run in 10.2 sec ==========
```
This is the way I have use the solution and run tests so far.

veanes: So I guess I need to add additional checks that under net48 the test with the new option are not run. So far I've not done that, but will do going forward. I'll see from your updates what needs to be done. 
I do see the failures when I now run `dotnet build /t:test` in the test directory (I did not run test like that before).

stephentoub (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Split.cs
stephentoub: >> we should spend a little time seeing how we might better abstract away the implementations to not need these special cases sprinkled throughout the top-level APIs

> I was thinking the same, but found it too difficult myself to do a deeper integration through the Run method of Regex

I'm working on this.

PR #1508, opened 2021-09-04
NativeAOT-LLVM: Reflection fix - add test and dependency for GetDependenciesDueToAccess

yowl:
This PR adds a test and a fix for the missing GetDependenciesDueToAccess call.

From a visual comparison of ILScanner https://github.com/dotnet/runtimelab/blob/23feaa85546690e611531ee0d6113077f8d64708/src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs#L682 to ILToLLVMImporter I also thought that there should be this dependency, `GVMLookupForSlot`, at https://github.com/dotnet/runtimelab/blob/23feaa85546690e611531ee0d6113077f8d64708/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs#L2186-L2189

```
            if (canonMethod.HasInstantiation)
            {
                exactContextNeedsRuntimeLookup = callee.IsSharedByGenericInstantiations;
                MethodDesc entryPoint = _compilation.TypeSystemContext.SystemModule.GetKnownType("System.Runtime", "TypeLoaderExports").GetKnownMethod("GVMLookupForSlot", null);
                _dependencies.Add(_compilation.NodeFactory.MethodEntrypoint(entryPoint), "LLVM GVM dependency");
            }
```

 But it's not currently present and nothing breaks, so interested if there is a test that might highlight this.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -3,12 +3,9 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
 using Internal.TypeSystem;
 using ILCompiler;
 using LLVMSharp.Interop;
@@ -17,7 +14,6 @@ using ILCompiler.DependencyAnalysis;
 using ILCompiler.DependencyAnalysisFramework;
 using ILCompiler.LLVM;
 using Internal.IL.Stubs;
-using Internal.ReadyToRunConstants;
 using Internal.TypeSystem.Ecma;
 
 namespace Internal.IL
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -17,6 +17,7 @@ using ILCompiler.DependencyAnalysis;
 using ILCompiler.DependencyAnalysisFramework;
 using ILCompiler.LLVM;
 using Internal.IL.Stubs;
+using Internal.ReadyToRunConstants;
 using Internal.TypeSystem.Ecma;
 
 namespace Internal.IL
@@ -1801,6 +1802,8 @@ namespace Internal.IL
         {
             MethodDesc runtimeDeterminedMethod = (MethodDesc)_methodIL.GetObject(token);
             MethodDesc callee = (MethodDesc)_canonMethodIL.GetObject(token);
+            _compilation.NodeFactory.MetadataManager.GetDependenciesDueToAccess(ref _dependencies, _compilation.NodeFactory, _canonMethodIL, callee);
+
             if (callee.IsIntrinsic)
             {
                 if (ImportIntrinsicCall(callee, runtimeDeterminedMethod))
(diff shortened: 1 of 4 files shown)


yowl:
cache error, restarting hoping its a one off

MichalStrehovsky:
Similar GetDependenciesDueToAccess should be added when importing a field - it triggers reflection analysis of the method body if the field was "interesting" (i.e. had the .NET 5 DynamicallyAccessedMembersAttribute annotation).

> to ILToLLVMImporter I also thought that there should be this dependency, GVMLookupForSlot, at

I think GVMLookupForSlot ends up being a normal managed method call for WASM, so it shows up as a dependency implicitly. It's a transparent helper call for RyuJIT.

yowl:
Thanks.  For a field of type string, is this something like

```
    public class DerivedClass1<T> : BaseClass<T>
    {
        public override sealed string GVMethod1<U>(T t, U u) { return "DerivedClass1.GVMethod1"; }
        
       [DynamicallyAccessedMembersAttribute]
       public string MethodName;
    }

    static void TestGetDependenciesDueToAccess()
    {
        StartTest("Test GetDependenciesDueToAccess (MakeGenericMethod)");
        var instance = new DerivedClass1<string>(), new[] { "", "" };
        instance.MethodName = "GVMethod1";
        MethodInfo gvm1 = typeof(DerivedClass1<string>).GetTypeInfo().GetDeclaredMethod(instance.MethodName).MakeGenericMethod(typeof(string));
        EndTest(gvm1.Invoke(instance ) == "DerivedClass1.GVMethod1");
    }
```
?

MichalStrehovsky:
You should see a test failure in #1516 without the fix for fields.

yowl:
Thanks for the test.  Yes that looks good now so I can enable the DataFlowTests for LLVM/Wasm.

yowl:
I added the test in this branch manually but that should resolve eventually when yours is merged and I merged that to this branch.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
```diff
@@ -17,6 +17,7 @@
 using ILCompiler.DependencyAnalysisFramework;
 using ILCompiler.LLVM;
 using Internal.IL.Stubs;
+using Internal.ReadyToRunConstants;
```
MichalStrehovsky: Is this needed?

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -373,6 +373,8 @@ private static unsafe int Main(string[] args)
 
         TestBoolCompare();
 
+        TestGetDependenciesDueToAccess();
```
MichalStrehovsky: I'll leave it up to you but if you're enabling the Dataflow tests, they'll cover this scenario too and the failure mode of the test is probably easier to troubleshoot.

MichalStrehovsky (approved):
Looks good otherwise!

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
yowl: Thanks, removed as suggested.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
yowl: No, removed (any a few others that weren't needed).

MichalStrehovsky (approved):
Thank you!

PR #1509, opened 2021-09-05
Add ARM to CI

kant2002:


eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -12,7 +12,7 @@ parameters:
   uploadIntermediateArtifacts: false
 
 steps:
-- ${{ if ne(parameters.archType, 'arm64') }}:
+- ${{ if and(ne(parameters.archType, 'arm'), ne(parameters.archType, 'arm64')) }}:
```
kant2002, line 15: If I use `eq(parameters.archType, 'x64')` would this cover `Linux_musl_x64`? Even if not, maybe better be explicit about arches where tests runs?

jkotas: Ideally, the tests would run everywhere and this condition would not be needed. This looks good to me as is.

jkotas (approved):
Thank you!

kant2002:
Thanks!

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
kant2002: Reason why I'm asking about that is that CI has financial and hardware limitations which may not be resolved as easy as technical one. Somebody has to approve bill to allocate hardware for ARM64 for example. And given that I see FreeBSD and s390x brewing, and RISK-V in distant future too, I do not see that this would be added to hardware pool so easy, given that even ARM64 not running tests by default.

Again, that's just a question which do not have practical application for me.

jkotas: runtimelab repo is low-traffic and CI does not cost much currently (dotnet/runtimelab costs <1% of what dotnet/runtime costs). The CI costs for runtimelab are not a concern at this point.

> ARM64 not running tests by default.

The reason behind this is that it is a non-trivial amount of work to get the ARM/ARM64 tests running for native AOT in the CI (we would want the tests to be cross-compiled on x64, just like the regular ARM/ARM64 build is cross-compiled). If you get the setup figure out, it would be fine to get the tests enabled  on ARM/ARM64.

PR #1510, opened 2021-09-05
NativeAOT-LLVM: Do not use mmap/munmap

yowl:
This PR adds an emscripten friendly VirtualReserve/Release implementation.  Emscripten does not fully implement mmap/munmap in particular it does not zero the memory in `mmap` for the `MAP_ANONYMOUS` flag.  This continues from https://github.com/dotnet/corert/issues/8330.

When trying to merge the latest into NativeAOT-LLVM random memory crashes, usually inside the GC, have started to appear, so now seems a good time to ask for feedback on this change.   This change does stop those crashes appearing so far at least.

Emscripten reports the page size as 0x4000, whereas `CommonMacros.h` has

```
#elif defined(HOST_WASM)

#define DATA_ALIGNMENT  4
#ifndef OS_PAGE_SIZE
#define OS_PAGE_SIZE    0x4
#endif
```
I don't know what the implications are there, if anything maybe just performance as `posix_memalign` does not refer to the page size.

### DIFF
--- a/src/coreclr/gc/wasm/gcenv.wasm.cpp
+++ b/src/coreclr/gc/wasm/gcenv.wasm.cpp
@@ -7,48 +7,6 @@
 // Emscripten does not provide a complete implementation of mmap and munmap: munmap cannot unmap partial allocations
 // Emscripten does provide an implementation of posix_memalign which is used here.
 
-struct _reserved
-{
-    void * address;
-    size_t size;
-    int used;
-};
-
-typedef struct _reserved reserved;
-
-static int nextPos = 0;
-static reserved blocks[1000]; // better to allocate and grow this as needed (until malloc fails)?
-static int init = 0;
-
-static void Init()
-{
-    int i;
-    for(i = 0; i < 1000; i++)
-    {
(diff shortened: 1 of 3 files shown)


src/coreclr/gc/wasm/gcenv.wasm.cpp
```diff
@@ -0,0 +1,181 @@
...
+{
+    void * address;
+    size_t size;
+    int used;
+};
+
+typedef struct _reserved reserved;
+
+static int nextPos = 0;
+static reserved blocks[1000]; // better to allocate and grow this as needed (until malloc fails)?
```
jkotas: This is really only used for argument validation. The GC should not be ever calling these APIs with invalid arguments.

I think you can omit this structure, and instead directly call `posix_memalign` / `free` / `memset` to implement `reserve` / `release` / `decommit`.

yowl: Thanks, works with without that.

jkotas (approved):
LGTM. Thanks!

PR #1514, opened 2021-09-06
Progress towards Android build

MichalStrehovsky:


jkotas (approved):
Thanks

kant2002:
How can I try to make Android build?

MichalStrehovsky:
```
export NDK_VER=r21b
export ANDROID_NDK_ROOT=~/android-ndk-${NDK_VER}
curl https://dl.google.com/android/repository/android-ndk-${NDK_VER}-linux-x86_64.zip -L --output ~/andk.zip
unzip ~/andk.zip -d $(dirname ${ANDROID_NDK_ROOT}) && rm -rf ~/andk.zip
build.sh nativeaot.libs+libs -arch arm64 -os Android
```

It requires
https://github.com/dotnet/runtime/pull/58701
https://github.com/dotnet/runtime/pull/58703
https://github.com/dotnet/runtime/pull/58700
And a couple others fixes that depend on whether the above can be upstreamed or not.

I'm on the verge of just abandoning this effort because it looks like https://github.com/dotnet/runtime/pull/58700 is not upstreamable and I'm not interested in a solution that would introduce extra merge conflicts in runtimelab when runtime repo gets around to do the "grand plan for Android". I tend to do very strict timeboxing for all of these side projects and I just about ran out of time for this.

MichalStrehovsky:
Here's my branch, but it doesn't have future unless the pieces that affect the shared engineering parts can be upstreamed: https://github.com/MichalStrehovsky/runtimelab/tree/androidBuild

kant2002:
Can you educate me why #58700 needed? I run `build.sh nativeaot.libs+libs -arch arm64 -os Android
` on clean NativeAOT , and it only produce errors in C# code, so I cannot understand where need to have C++ tweaks appears? I have to start talking about shared eng plans on Mono, but I willing to do smallest possible part there to unblock you here, but I need understand why this needed in first place.

MichalStrehovsky:
> on clean NativeAOT

You mean on my branch? That one has the fix I'm trying to upstream. It has like a dozen commits.

The src/libraries currently build by including android.toolchain.cmake and using the CMake support in the Android NDK. This is how src/mono also builds on Android but it duplicates all of this logic. I'm trying to move that logic to /eng so that we can also build NativeAOT that way without duplicating it like the Mono build does (the nice thing about using the builtin cmake support is that it would likely also work when building on Windows targeting Android with very little effort, but that's tangential).

But seems like including android.toolchain.cmake is not how people want to build Android going forward. They would instead like to use "CoreCLR way" that involves passing `-cross` to build.sh and providing a sysroot that is created using a separate Bash script. Not sure when that will happen but until that work happens, we can't make much progress here.

kant2002:
No, I do `build.sh nativeaot.libs+libs -arch arm64 -os Android` on `features/NativeAOT`.
I was under impression that this is how it should be done. That build produce small amount of C# errors about missing functions mostly. That's why I do not get why this changes #58700 was needed, so I try to better understand what's going one.

Right now you point me that Android build using `android.toolchain.cmake` which probably include additional configuration.

> Not sure when that will happen but until that work happens, we can't make much progress here.

It is crystal clear for me, that's why I'm trying to capture as much information as possible about the current situation, so if I go and start asking questions, maybe somebody would respond to me. 

At least you give me a bit of hints, so I will look here and there, and maybe comeback again

MichalStrehovsky:
It's likely not actually building for Android. It would have failed during CMake configure step. We need to configure the entire coreclr directory and the coreclr single file file host references libraries that don't exist on Android, for example.

PR #1520, opened 2021-09-07
More cleanup of Regex SRM

stephentoub:


veanes:
I went through all the changes. All looked good to me. At some point I will get rid of the `IfThenElse` node and replace it with a `Complement` node in the SymbolicRegexNode AST. I will do this after most of the cleanup is done.

veanes (approved):

PR #140, opened 2020-10-02
Fix NuGet packages

MichalStrehovsky:
* Linux NuGet package wasn't including the static libraries.
* The non-RID-specific package had RID specific stuff in it.

Part of this is basically a revert of dotnet/runtime#41966. I got lost in trying to figure out what the .depsproj architecture is trying to achieve so I'm planning to just add this to the unification tech debt issue as we merge this.

PR #1523, opened 2021-09-07
Add `/guard:cf` flavor of bootstrapper

MichalStrehovsky:


MichalStrehovsky:
I also spent a bit time thinking how to do System.IO.Compression.Native (also with respect to #106).

Seems like S.I.C.lib with `/guard:cf` is almost what we need for the single-file host (the only other difference is LTCG - not sure if it's giving us anything for this lib though).

I see two options:

* Build two versions of S.I.C.Native static lib: one with CFG, the other without. Disable LTCG. Single-file host will use the one with CFG. Maybe we should name the one without CFG as `.NoGuardCF` for CoreCLR's benefit. (i.e. the invert the suffix logic I introduced for NativeAOT bits).
* Build three versions S.I.C.Native static libs: for single file host, and NativeAOT GuardCF/NoGuard.

In either case, we should probably switch to the plan that single file host is on and build the shims as part of coreclr partition build (i.e. the `STATIC_LIBS_ONLY` plan) so that we can control what flavors of libs build.

What do you think?

MichalStrehovsky:
Putting the shims on the STATIC_LIBS_ONLY plan will also help with some other aspects of the tech debt since we don't want to rely on the libraries build to produce and keep these libraries for us (it requires twisting the framework packaging in a way it's not meant to work - https://github.com/dotnet/runtimelab/pull/140). STATIC_LIBS_ONLY plan lets us treat these libraries same as Bootstrapper or Runtime when it comes to packaging.

jkotas (approved):

jkotas:
> the only other difference is LTCG - not sure if it's giving us anything for this lib though

The compiler and linker have to match with LTCG. We have disabled LTCG to make things work when the compiler in the build lab and linker at the end user machine are not exactly the same. If we were to enable LTCG, we would need to figure out how to download the exact right version of the linker.

MichalStrehovsky:
I meant whether LTCG is giving any benefit in the CoreCLR config and whether we need an extra lib flavor just to have one with LTCG for CoreCLR (it's the difference between the "build 3 libs" and "build 2 libs" above is whether we can get away without LTCG for all products)

jkotas:
I suspect that LTCG (and the whole program optimization that comes with it) is going to benefit CoreCLR config in some cases.

PR #1524, opened 2021-09-07
Remove RhpReversePInvoke

kant2002:
I do not see where this is being used. Only RhpReversePInvoke2 is used. Also this function nowhere in .S files.


jkotas:
Thank you! It would be also nice to rename `RhpReversePInvoke2` to `RhpReversePInvoke` in a follow up change.

jkotas (approved):

PR #1525, opened 2021-09-07
generalized most SRM unit tests to None and Compiled

veanes:
Cleaned up (most) of the unit tests that used to be specific for `NonBacktracking` in `RegexSRMTests` by generalizing them to both `None` and `Compiled` options. 
Added a new test called `AllMatches` in `Regex.Match.Tests` that consolidates most prior SRM specific tests and lifts **all these test also to None and Compiled** (thus further checking identical results in may corner cases -- that were actually not previously formally tested)
Also lifted over some other tests. 
Some further cleanup may still be needed for SRM specific tests but this was the bulk of them.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
@@ -10,6 +10,24 @@ namespace System.Text.RegularExpressions.Tests
 {
     public class RegexKnownPatternTests
     {
+        public static IEnumerable<RegexOptions> RegexOptionsExtended()
+        {
+            if (!PlatformDetection.IsNetFramework)
+            {
+                yield return RegexHelpers.RegexOptionNonBacktracking;
+            }
+
+            yield return RegexOptions.None;
+            yield return RegexOptions.Compiled;
+        }
+
+        public static IEnumerable<object[]> RegexOptions_TestData()
+        {
+            foreach (RegexOptions options in RegexOptionsExtended())
+            {
+                yield return new object[] { options };
+            }
+        }
         //
         // These patterns come from the Regex documentation at docs.microsoft.com.
         //
--- a/src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using Xunit;
@@ -9,6 +10,24 @@ namespace System.Text.RegularExpressions.Tests
 {
     public class RegexKnownPatternTests
     {
+        public static IEnumerable<RegexOptions> RegexOptionsExtended()
+        {
+            if (!PlatformDetection.IsNetFramework)
+            {
+                yield return RegexHelpers.RegexOptionNonBacktracking;
+            }
+
+            yield return RegexOptions.None;
+            yield return RegexOptions.Compiled;
+        }
+
+        public static IEnumerable<object[]> RegexOptions_TestData()
+        {
+            foreach (RegexOptions options in RegexOptionsExtended())
+            {
+                yield return new object[] { options };
+            }
+        }
         //
         // These patterns come from the Regex documentation at docs.microsoft.com.
         //
@@ -294,8 +313,7 @@ namespace System.Text.RegularExpressions.Tests
 
         // https://docs.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions#noncapturing-groups
         [Theory]
-        [InlineData(RegexOptions.None)]
-        [InlineData(RegexOptions.Compiled)]
+        [MemberData(nameof(RegexOptions_TestData))]
         public void Docs_GroupingConstructs_NoncapturingGroups(RegexOptions options)
         {
             const string Pattern = @"(?:\b(?:\w+)\W*)+\.";
@@ -309,8 +327,7 @@ namespace System.Text.RegularExpressions.Tests
 
         // https://docs.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions#group-options
         [Theory]
-        [InlineData(RegexOptions.None)]
-        [InlineData(RegexOptions.Compiled)]
+        [MemberData(nameof(RegexOptions_TestData))]
         public void Docs_GroupingConstructs_GroupOptions(RegexOptions options)
         {
             const string Pattern = @"\b(?ix: d \w+)\s";
@@ -614,8 +631,7 @@ namespace System.Text.RegularExpressions.Tests
 
         // https://docs.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions#linear-comparison-without-backtracking
         [Theory]
-        [InlineData(RegexOptions.None)]
-        [InlineData(RegexOptions.Compiled)]
+        [MemberData(nameof(RegexOptions_TestData))]
         public void Docs_Backtracking_LinearComparisonWithoutBacktracking(RegexOptions options)
         {
             const string Pattern = @"e{2}\w\b";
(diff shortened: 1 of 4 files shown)


src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
```diff
@@ -647,13 +663,20 @@ public void Docs_Backtracking_WithOptionalQuantifiersOrAlternationConstructs(Reg
...
-            Assert.False(Regex.IsMatch(Input, "^((?>[0-9a-fA-F]{1,4}:)*(?>[0-9a-fA-F]{1,4}))*(::)$", options)); // non-backtracking
+            if (options == RegexHelpers.RegexOptionNonBacktracking)
+            {
+                // With NonBacktracking option this test becomes feasible: from almost a minute to less than a second
+                Assert.False(Regex.IsMatch(Input, "^(([0-9a-fA-F]{1,4}:)*([0-9a-fA-F]{1,4}))*(::)$", options));
+            }
+            else
+            {
+                Assert.False(Regex.IsMatch(Input, "^((?>[0-9a-fA-F]{1,4}:)*(?>[0-9a-fA-F]{1,4}))*(::)$", options)); // forced non-backtracking
+            }
```
stephentoub, line 660:
```suggestion
            Assert.False(Regex.IsMatch(Input, "^((?>[0-9a-fA-F]{1,4}:)*(?>[0-9a-fA-F]{1,4}))*(::)$", options)); // forced non-backtracking
```

src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
```diff
@@ -647,13 +663,20 @@ public void Docs_Backtracking_WithOptionalQuantifiersOrAlternationConstructs(Reg
 
         // https://docs.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions#nonbacktracking-subexpression
         [Theory]
-        [InlineData(RegexOptions.None)]
-        [InlineData(RegexOptions.Compiled)]
+        [MemberData(nameof(RegexOptions_TestData))]
         public void Docs_Backtracking_WithNestedOptionalQuantifiers(RegexOptions options)
         {
             const string Input = "b51:4:1DB:9EE1:5:27d60:f44:D4:cd:E:5:0A5:4a:D24:41Ad:";
             // Assert.False(Regex.IsMatch(Input, "^(([0-9a-fA-F]{1,4}:)*([0-9a-fA-F]{1,4}))*(::)$")); // takes too long due to backtracking
```
stephentoub:
```suggestion
```

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -1312,5 +1312,166 @@ public void Match_Wordchar()
...
+            Assert.True(re.IsMatch(input));
+        }
+
+        public static IEnumerable<object[]> AllMatches_TestData()
+        {
+            foreach (RegexOptions options in new[] { RegexHelpers.RegexOptionNonBacktracking, RegexOptions.None, RegexOptions.Compiled })
+            {
+                if (PlatformDetection.IsNetFramework && options == RegexHelpers.RegexOptionNonBacktracking)
+                {
+                    yield break;
```
stephentoub: Did you really mean `yield break;` here?  Shouldn't this be:
```suggestion
                    continue;
```
?

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -1312,5 +1312,166 @@ public void Match_Wordchar()
...
+        [Theory]
+        [InlineData(RegexOptions.None)]
+        [InlineData(RegexOptions.Compiled)]
+        [InlineData(RegexHelpers.RegexOptionNonBacktracking)]
+        public void StressTestDeepNestingOfConcat(RegexOptions options)
+        {
+            if (PlatformDetection.IsNetFramework && options == RegexHelpers.RegexOptionNonBacktracking)
+            {
+                return;
+            }
```
stephentoub: Could we move the RegexOptionsExtended and RegexOptions_TestData into a shared test class (I believe we already have one where we put helpers)?  Then this can just use that same MemberData, and we can get rid of some duplication elsewhere as well.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -1312,5 +1312,166 @@ public void Match_Wordchar()
...
+
+            string pattern = string.Concat(Enumerable.Repeat("([a-z]", 1000).Concat(Enumerable.Repeat(")", 1000)));
+            string input = string.Concat(Enumerable.Repeat("abcde", 200));
```
stephentoub: Nit: could you lift these constant values out into named const locals?  I'm also wondering if 1000 is sufficient, especially on Linux where stacks are typically much larger.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: It turns out that `MemberData` can only reference a method in the same class where the test itself is defined, so it seems moving `RegexOptions_TestData` to `RegexHelpers` class is not allowed. I'll move `RegexOptionsExtended`.

src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
veanes: The atomic loop annotation  `(?>...)` is not supported in `NonBacktracking` mode. So the `IsMatch` will throw an exception if not guarded by `else`.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: I'm updating the `Regex.Match.Tests` to use the new `RegexHelpers.RegexOptionsExtended()` method instead to clean this up, whenever possible. The intent here was simply that the tests must not run in the `NonBacktracking` mode when `IsNetFramework` is true.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
stephentoub: > It turns out that MemberData can only reference a method in the same class where the test itself is defined

You can set a MemberType property on the attribute, e.g.
https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Linq.Parallel/tests/QueryOperators/OfTypeTests.cs#L33

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
stephentoub: > The intent here was simply that the tests must not run in the NonBacktracking mode when IsNetFramework is true.

Right, but `yield break;` will permanently exit the enumeration.

src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
stephentoub: Ah, ok.  Can you remind me why?  Rather than writing it one-off here, I think we should add a README.md to the root of the SRM folder that outlines:
- What features aren't supported (and why)?
- Key terminology employed in the implementation
- Other interesting details to help understand the implementation

We can do that subsequently.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: I introduced `RegexHelpers.StressTestNestingDepth` and set it  to 10000. I'm getting 
`The program '[55724] dotnet.exe' has exited with code 3221225477 (0xc0000005) 'Access violation'.`
so need to investigate this further.

veanes: This happens only in `NonBacktracking` mode, the other cases seem fine.

veanes: I'm investigating this next, but will set it back to 1000 for now to finish these edits first,

src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
veanes: Short reply is that it changes the meaning of the loop -- cutting of choices that would exist with backtracking. In `NonBacktracking` mode it messes with/contradicts the DFA semantics because all search paths are preserved there -- just that backtracking is not being used to preserve those paths. We discussed this briefly at some point in the spring. Yes, **documentation** is the next big item on the todo list.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: Good to know. Thanks! I'll use that attribute.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: This is not what I intended, should be fixed now with the commit I'm pushing next.

stephentoub:
(Pushed a rebase to handle conflicts with other changes)

stephentoub (approved):
Thanks.

PR #1529, opened 2021-09-07
More cleanup for Regex SRM

stephentoub:
- Change GenerateMinterms to only return BDD, not a tuple; the bool[] was never used, so we don't need to exert energy creating it.
- Avoid recursion in MintermGenerator.GetLeaves
- Remove some dead code in GetIgnoreCaseRel
- Change exceptions for not supported to use NotSupported rather than NotImplemented
- Remove some unused parameters/fields
- Remove some more bounds checking by iterating from 0 to arr.Length rather than to a const
- Make a few properties into auto-props
- Clean up some naming and use of braces
- Make some types nested where they're only used by one consumer
- Make some structs readonly
- Remove more unused usings
- Clean up some exception throwing


veanes (approved):
Looks good.

PR #1531, opened 2021-09-08
Compile NativeAOT on x86

kant2002:
Thanks to suggestions from dotnet/runtime#86573 I was able to compile NativeAOT on Windows x86.
I apply just `USE_GC_INFO_DECODER` for now.


kant2002:
I need to uncomment parts of 
https://github.com/dotnet/runtimelab/blob/67927bd15799cccf4ab77a8bcd38a508c2953cd7/src/coreclr/nativeaot/Runtime/i386/WriteBarriers.asm#L223-L236

But still ILC fails to process `CORINFO_HELP_ENDCATCH`. I would need advice where I can take parts for that implementation in ASM or somewhere elese.

kant2002:
Also do I send something like dotnet/runtimelab#1530 and dotnet/runtimelab#1509 so I would not think too much about ObjWriter?

src/coreclr/nativeaot/Runtime/gcinfodecoder.h
```diff
@@ -1,4 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+// USE_GC_INFO_DECODER does not defined for Win-x86 in CoreCLR
```
jkotas: This won't be enough. You will also need to define it when building RyuJIT.

src/coreclr/nativeaot/Runtime/windows/PalRedhawkCommon.cpp
```diff
@@ -172,7 +172,7 @@ REDHAWK_PALEXPORT int32_t PalGetModuleFileName(_Out_ const TCHAR** pModuleNameOu
     return 0;
 }
 
-REDHAWK_PALEXPORT uint64_t __cdecl PalGetTickCount64()
+REDHAWK_PALEXPORT uint64_t REDHAWK_PALAPI PalGetTickCount64()
```
jkotas, line 175: PalGetTickCount64 is defined as cdecl in C#. It has to match.

src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/CallConverterThunk.cs
```diff
@@ -119,10 +120,11 @@ private static unsafe bool CallConverterThunk_LazyCctor()
 #if TARGET_UNIX
             // TODO
 #else
-            CallingConventionConverter_GetStubs(out ReturnVoidReturnThunk, out ReturnIntegerPointReturnThunk, out CommonInputThunkStub
+            CallingConventionConverter_GetStubs(out ReturnVoidReturnThunk, out ReturnIntegerPointReturnThunk,
 #if CALLDESCR_FPARGREGSARERETURNREGS
+                                                out CommonInputThunkStub
 #else
```
jkotas, line 126: I do not see why this change is needed.

kant2002: This is purely to satisfy style cop. I may suppress it on these code. It did not like comma before other parameters.

src/coreclr/nativeaot/Runtime/gcinfodecoder.h
kant2002: I assume this is NativeAOT specific customization to https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/coreclr/jit/CMakeLists.txt

Also I did not see any reference to `USE_GC_INFO_DECODER` in jit folder. Are you talking that I need to use `FEATURE_EH_FUNCLETS` then how I make that change purely NativeAOT specific ?

src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/CallConverterThunk.cs
jkotas: Suppressing the warning would be better. The ifdefs do not make logical sense with this change.

src/coreclr/nativeaot/Runtime/windows/PalRedhawkCommon.cpp
kant2002: REDHAWK_PALAPI defined as StdCall on non-Unix, and empty value on Unix. Does this mean, that I should change to StdCall in C#?

jkotas:
> Also do I send something like dotnet/runtimelab#1530 and dotnet/runtimelab#1509 so I would not think too much about ObjWriter?

I would not want to add x86 to the official builds until it is close to working.

jkotas:
> `CORINFO_HELP_ENDCATCH`

This helper is only needed for `!FEATURE_EH_FUNCLETS`. If you plan to go with `FEATURE_EH_FUNCLETS`, it won't be needed.

src/coreclr/nativeaot/Runtime/gcinfodecoder.h
jkotas: > I assume this is NativeAOT specific customization to https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/coreclr/jit/CMakeLists.txt

Yes. It becomes tricky because it would mean that the JIT binary used by CoreCLR will be different from the JIT binary used by AOT. It is actually a bunch of binaries since we build multiple variants to support cross-compilation. There are short-term hacks possible in this branch. Once/if we move NativeAOT to dotnet/runtime, we will need to be able to use the same JIT binary for both regular CoreCLR and NativeAOT. If we keep the ABI differences, we may need to replace the ifdefs with dynamic checks in the JIT. Other option is to just stick with the same ABI as CoreCLR and clean it up as necessary (ie avoid forking the ABI for NativeAOT).

> Also I did not see any reference to USE_GC_INFO_DECODER in jit folder

JIT has `JIT32_GCENCODER`. `JIT32_GCENCODER = !USE_GC_INFO_DECODER`

src/coreclr/nativeaot/Runtime/windows/PalRedhawkCommon.cpp
jkotas: You can add new `RhpGetTickCount64` that is cdecl and call that from C# instead of PalGetTickCount64. And then this change can stay.

src/coreclr/nativeaot/Runtime/gcinfodecoder.h
kant2002: I think I found a way how we can postpone decision temporary. Instead of this particular change, I remove these lines from .h files

https://github.com/dotnet/runtimelab/blob/6f2cfcdb8a4484f5a62234353923ee8073c57593/src/coreclr/jit/jit.h#L468-L470

and 

https://github.com/dotnet/runtimelab/blob/6f2cfcdb8a4484f5a62234353923ee8073c57593/src/coreclr/inc/gcinfodecoder.h#L22-L24

And will control these defines from CMake files. That change maybe be submitted to Runtime repo I think, and all hacks would be in CMake files in NativeAOT repo. That way we can explore all 3 options without too much commitment to either of them, if I only apply CMake hacks locally. What do you think?

jkotas: I do not think that there is a big difference in controlling it in cmake files vs. in header files. Both of these are short-term hacks that would be ok in this branch, but not in dotnet/runtime. You end up with a JIT build that is different from regular CoreCLR JIT build in either case.

src/coreclr/nativeaot/Runtime/windows/PalRedhawkCommon.cpp
kant2002: So you propose declare `RhpGetTickCount64` in c++, for which I would add corresponding C# function, and use that function instead of `PalGetTickCount64`? I probably can just define `__cdecl` to nothing for `TARGET_X86` in that file then.

src/coreclr/nativeaot/Runtime/gcinfodecoder.h
kant2002: Ok, will try that. I think about JIT as black box for now, so it was a scary change for me. I think I understand direction now.

src/coreclr/nativeaot/Runtime/windows/PalRedhawkCommon.cpp
jkotas: Yes. There is number of methods like that (e.g. `RhpReleaseCastCacheLock`). They are using `__cdecl` already.

kant2002:
There bunch of assertions with only 2 kinds of them during ILC run with latest changes. Not sure if this is indication that I have to do something extra, or this is indication that I do something wrong

https://github.com/dotnet/runtimelab/blob/6f2cfcdb8a4484f5a62234353923ee8073c57593/src/coreclr/jit/gcencode.cpp#L4594

and

https://github.com/dotnet/runtimelab/blob/6f2cfcdb8a4484f5a62234353923ee8073c57593/src/coreclr/gcinfo/gcinfoencoder.cpp#L1158

Last one has additional information in form of this text message `Assertion failed 'genRegPtrTemp->rpdGCtypeGet() != GCT_NONE' in 'System.IO.StreamWriter:.cctor()' during 'Emit GC+EH tables' (IL size 27)`

kant2002:
At least

https://github.com/dotnet/runtimelab/blob/6f2cfcdb8a4484f5a62234353923ee8073c57593/src/coreclr/gcinfo/gcinfoencoder.cpp#L1158

related to `#define FEATURE_FIXED_OUT_ARGS 0` on x86. 
If I define  `#define FEATURE_FIXED_OUT_ARGS 1` on x86. 
next assertion is 

https://github.com/dotnet/runtimelab/blob/6f2cfcdb8a4484f5a62234353923ee8073c57593/src/coreclr/jit/lower.cpp#L1204

Which seems to be related to `FEATURE_PUT_STRUCT_ARG_STK` being defined. If I undefine it for Win-x86 configuration, then `src\coreclr\jit\codegenxarch.cpp` does not ready for this challenge now. I was try to add missing #ifdefs but next issues start appear in `src\coreclr\jit\lowerxarch.cpp` and I think rabbit hole is too deep. Maybe I miss something simple

kant2002:
Let's start with IL
```
IL to import:
IL_0000  16                ldc.i4.0    
IL_0001  6a                conv.i8     
IL_0002  7e 23 04 00 04    ldsfld       0x4000423
IL_0007  73 37 0a 00 06    newobj       0x6000A37
IL_000c  80 9d 01 00 04    stsfld       0x400019D
```

Which translated to following statements
```
***** BB01
STMT00001 (IL   ???...  ???)
               [000011] I-CXG-------              *  CALL      void   System.DateTimeOffset..ctor (exactContextHnd=0x4000000000420029)
               [000010] ------------ this in ecx  +--*  ADDR      byref 
               [000009] -------N----              |  \--*  LCL_VAR   struct<System.DateTimeOffset, 12> V01 tmp1         
               [000001] ------------ arg1         +--*  CNS_LNG   long   0x0000000000000000
               [000005] --CXG------- arg2         \--*  OBJ       struct<System.TimeSpan, 8>
               [000004] --CXG-------                 \--*  ADD       byref 
               [000002] H-CXG-------                    +--*  CALL help byref  HELPER.CORINFO_HELP_READYTORUN_STATIC_BASE
               [000003] ------------                    \--*  CNS_INT   int    0 Fseq[Zero]

***** BB01
STMT00002 (IL 0x00C...  ???)
               [000018] -ACXG-------              *  ASG       struct (copy)
               [000017] --CXG-------              +--*  BLK       struct<System.DateTimeOffset, 12>
               [000015] --CXG-------              |  \--*  ADD       byref 
               [000013] H-CXG-------              |     +--*  CALL help byref  HELPER.CORINFO_HELP_READYTORUN_STATIC_BASE
               [000014] ------------              |     \--*  CNS_INT   int    0 Fseq[MinValue]
               [000012] ------------              \--*  LCL_VAR   struct<System.DateTimeOffset, 12> V01 tmp1         
```


Eventually it transformed to 
```
***** BB01
STMT00000 (IL 0x000...0x00C)
N003 (  5,  4) [000008] IA--G---R---              *  ASG       struct (init) $VN.Void
N002 (  3,  2) [000006] D---G--N----              +--*  LCL_VAR   struct<System.DateTimeOffset, 12>(AX) V01 tmp1         
N001 (  1,  1) [000007] ------------              \--*  CNS_INT   int    0 $41

***** BB01
STMT00001 (IL   ???...  ???)
N014 ( 47, 27) [000011] -ACXG-------               *  CALL      void   System.DateTimeOffset..ctor $VN.Void
N010 ( 22, 11) [000058] -ACXG---R-L- arg2 SETUP    +--*  ASG       struct (copy) $VN.Void
N009 (  3,  2) [000056] D------N----               |  +--*  LCL_VAR   struct<System.TimeSpan, 8> V04 tmp4         d:1
N008 ( 18,  8) [000005] -ACXG--N----               |  \--*  IND       struct <l:$285, c:$284>
N007 ( 15,  6) [000077] -ACXG-------               |     \--*  COMMA     byref  $240
N005 ( 14,  5) [000075] -ACXG---R---               |        +--*  ASG       byref  $VN.Void
N004 (  1,  1) [000074] D------N----               |        |  +--*  LCL_VAR   byref  V07 cse0         d:1 $240
N003 ( 14,  5) [000002] H-CXG-------               |        |  \--*  CALL help byref  HELPER.CORINFO_HELP_READYTORUN_STATIC_BASE Zero Fseq[Zero] $240
N006 (  1,  1) [000076] ------------               |        \--*  LCL_VAR   byref  V07 cse0         u:1 Zero Fseq[Zero] $240
N011 (  3,  2) [000059] ------------ arg2 in out+08+--*  LCL_VAR   struct<System.TimeSpan, 8> V04 tmp4         u:1 (last use) $2c2
N012 (  3,  3) [000010] ------------ this in ecx   +--*  LCL_VAR_ADDR byref  V01 tmp1          $102
N013 (  2,  2) [000001] ------------ arg1 in out+00\--*  CNS_LNG   long   0x0000000000000000 $300

```

And this is last state of this part of tree before it start asserting. I mark argument which lowering complains about.

```
lowering call (before):
N006 (  1,  1) [000076] ------------        t76 =    LCL_VAR   byref  V07 cse0         u:1 Zero Fseq[Zero] $240
                                                  /--*  t76    byref  
N008 ( 18,  8) [000005] -c-XG--N----         t5 = *  IND       struct <l:$285, c:$284>
               [000100] Dc-----N----       t100 =    LCL_VAR_ADDR byref  V04 tmp4         
                                                  /--*  t100   byref  
                                                  +--*  t5     struct 
N010 ( 22, 11) [000058] sA--------L-              *  STORE_BLK struct<System.TimeSpan, 8> (copy) (Unroll)
N011 (  3,  2) [000059] ------------        t59 =    LCL_VAR   struct<System.TimeSpan, 8> V04 tmp4         u:1 (last use) $2c2
N012 (  3,  3) [000010] ------------        t10 =    LCL_VAR_ADDR byref  V01 tmp1          $102
N013 (  2,  2) [000001] ------------         t1 =    CNS_INT   int    0
               [000093] ------------        t93 =    CNS_INT   int    0
                                                  /--*  t1     int    
                                                  +--*  t93    int    
               [000094] ------------        t94 = *  LONG      long  
                                                  /--*  t59    struct arg2 in out+08  <-- !!!!!  ASSERT ON THIS ARG !!!!!
                                                  +--*  t10    byref  this in ecx
                                                  +--*  t94    long   arg1 in out+00
N014 ( 47, 27) [000011] --CXG-------              *  CALL      void   System.DateTimeOffset..ctor $VN.Void
```

My understanding that assert do not like that "local variable" passed as struct, and this should be "morphed" earlier. If morphing is indeed proper place for turning structs into I will look there. I also will re-read responsibilities of each phase, so maybe I will get better understanding what's going on.

uno-studio:
Has this pull request been merged into the master branch?

kant2002:
@unofficialdev I was stuck with that. I do not have enough cycles to make that work properly, as such this is only half done. I still have appetite for that, just delay this excercise. If you want to pursue, you are free base that on current work,

jkotas:
This work should continue in dotnet/runtime repo.

yowl:
Why was `RhpGetTickCount64` added as `void` not `uint64_t` ?  `wasm-ld` doesn't like the mismatch between the c# declaration in `InternalCalls.cs` of `ulong` .

jkotas:
It was an oversight. Would you like to submit a PR with a fix?

yowl:
Thanks: https://github.com/dotnet/runtime/pull/79538

larkliy:
Could you tell us how you were able to achieve NativeAOT compilation in x86? I want to link a C++ static library to my x86 project.

yowl:
> Could you tell us how you were able to achieve NativeAOT compilation in x86? I want to link a C++ static library to my x86 project.

Hi, work for x86 continues in the runtime repo, not here in runtimelab.

PR #1371, opened 2021-08-02
Rework RhGetRuntimeHelperForType in C#

kant2002:
as per this discussion https://github.com/dotnet/runtimelab/pull/1294#issuecomment-874430938

PR #1532, opened 2021-09-08
Delete indirections

MichalStrehovsky:
Follows up to #1371 and #1328. This code was used to support the fragile/nonversionable SharedLibrary.dll. It's unlikely we would want this in NativeAOT. If we ever add support for compiling apps that consist of multiple native libraries, it will probably follow a ReadyToRun scheme instead of fragile references.


MichalStrehovsky:
Cc @yowl - fat calls will do one less dereference (see the RyuJIT change), and so will accessing the static bases from generic dictionaries (this one may just fall out - it fell out here - no change to ReadyToRunGenericLookupHelper was needed).

jkotas (approved):
Nice!

PR #1533, opened 2021-09-08
Move SRM construction to be behind RegexRunnerFactory abstraction

stephentoub:
This moves the Regex SRM implementation to be better integrated with the existing regex library, putting it behind the existing RegexRunner{Factory} abstraction.

It also fixes a thread-safety issue that's been causing me to hit occasional null reference exceptions and asserts due to IgnoreCaseTransformer assigning multiple fields non-atomically.


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
```diff
@@ -1,63 +1,32 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
-using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions.SRM;
```
danmoseley, line 6: Later, we should probably globally rename SRM to NonBacktracking.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
```diff
@@ -1,63 +1,32 @@
...
-                if (match.Success)
-                {
-                    var m = new Match(this, 1, input, beg, length, startat);
-                    m._matches[0][0] = match.Index;
-                    m._matches[0][1] = match.Length;
-                    m._matchcount[0] = 1;
-                    m.Tidy(match.Index + match.Length);
-                    return m;
-                }
+                throw new NotSupportedException();
```
danmoseley, line 20: for whatever reason, we use PlatformNotSupportedException for the existing serialize API. NSE does seem more correct but it's not consistent?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Replace.cs
```diff
@@ -172,43 +172,7 @@ private static string Replace(MatchEvaluator evaluator, Regex regex, string inpu
...
-                }
-
-                // final segment
-                if (state.prevat < input.Length)
-                {
-                    state.segments.Add(state.input.AsMemory(state.prevat, input.Length - state.prevat));
-                }
-            }
-            else if (!regex.RightToLeft)
+            if (!regex.RightToLeft)
```
danmoseley, line 175: Aside, I notice SRM doesn't support RTL | EcmaScript. Do we expect to ship that way? Although my care level is low as it's likely very rarely used.

I also noticed we today don't seem to have any tests with RTL | EcmaScript, although I would expect the combination is unlikely to throw up any bugs.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
stephentoub: From my perspective,  this whole function is just a placeholder. It's currently not used anywhere, and assuming it's worth it's weight,  I think our goal should be for it to be an implementation detail as part of source generating non-backtracking regexes. (Though even there I'd hope we could come up with a better way to spit out code specific to the regex. )

danmoseley:
Aside - @veanes do you plan to remove the ""DFAINCOMPATIBLE" markers and simply move those into their own enumeration? that seems prettier. also by doing this you can remove the check for EcmaScript | RTL in MonoRegexTests.cs. Which seems to not do nothing anyway as I don't (I think) see any such tests.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.SRM.cs
stephentoub: Yes. Namespaces and folders are still on the docket to be cleaned up.  You can see multiple PRs of me incrementally working through the codebase.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Replace.cs
stephentoub: > although I would expect the combination is unlikely to throw up any bugs

To be clear,  it doesn't support either option,  not just both together. 

I think it's fine to ship that way.  We can always add in support later if it's really deemed impactful.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
```diff
@@ -312,26 +291,26 @@ public static string Unescape(string str)
...
-                }
+                return capslist.AsSpan().ToArray();
             }
-            else
+
+            var result = new string[capsize];
+            for (int i = 0; i < result.Length; i++)
             {
-                result = capslist.AsSpan().ToArray();
+                result[i] = ((uint)i).ToString();
```
danmoseley, line 312: why the cast?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
```diff
@@ -403,15 +384,16 @@ public int GroupNumberFromName(string name)
                 ThrowHelper.ThrowArgumentNullException(ExceptionArgument.name);
             }
 
-            if (_srm is not null)
+            if ((roptions & RegexOptions.NonBacktracking) != 0)
             {
                 // In NonBacktracking mode there is a single top level group 0
+                // See comment in GetGroupNames.
```
danmoseley, line 390: should this magic 0 value be a const field maybe?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/srm/SymbolicRegexMatcher.cs
```diff
@@ -9,9 +9,37 @@
...
+        public abstract void Serialize(StringBuilder sb);
+
+        /// <summary>
+        /// Unwind the regex of the matcher and save the resulting state graph in DGML
+        /// </summary>
+        /// <param name="bound">roughly the maximum number of states, 0 means no bound</param>
+        /// <param name="hideStateInfo">if true then hide state info</param>
+        /// <param name="addDotStar">if true then pretend that there is a .* at the beginning</param>
+        /// <param name="inReverse">if true then unwind the regex backwards (addDotStar is then ignored)</param>
+        /// <param name="onlyDFAinfo">if true then compute and save only genral DFA info</param>
```
danmoseley, line 34: nit typo genral. feel free to ignore for now..

danmoseley (approved):
LGTM

veanes (approved):
Overall, really intricate and great improvements. 
Clearly shows you went throw the code **in-depth** (up to the core details even in the BDD ignore-case-table semantics).
Actually, going over this PR was a great learning experience for me. 
(I need to start get using the very convenient and presumably more efficient new C# language features like ?? for dealing with uninitialized fields.)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
veanes: The top level group is always 0 and "0" as a string. Not sure these need separate constants.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Replace.cs
veanes: On a related note, conceptually EcmaScript would be really easy to add. What it essentially affects is the meaning of \w and \d (probably also \s), all in a very systematic manner. In all those cases it is simply the intersection of the involved BDDs with the BDD for ascii and this happens when the BDDs are created from the original regex parse tree. 
(May be there are corner cases I don't understand. Not sure exactly what happens with \W and \D etc).

stephentoub:
> As I understand Interlocked.CompareExchange makes sure that no two threads can simultaneously update the static tables which fixes the thread safety issue here.

Yes and no.  The primary fix here is avoiding the pattern of:
```C#
if (a is null)
{
    a = new ...;
    b = new ...;
}

Use(a);
Use(b);
```
as if this code is executed from multiple threads, it's possible the second thread sees that a isn't null but sees b as still being null.  That can be worked around by changing the order of the writes, e.g.
```C#
if (a is null)
{
    b = new ...;
    a = new ...;
}

Use(a);
Use(b);
```
such that the thing being checked for nullness is the one written after everything else guarded by the check.  In a sequential consistency world, that's likely sufficient, but according to ECMA the reads and writes here can be reordered, so doing it correctly requires that a also be volatile (or that the reads/writes with it be volatile).

That doesn't explain the Interlocked, though.  I added the Interlocked on top of using volatile in order to ensure that every thread ends up using the same instance.   With the above code, even if everything is volatile, it's possible for two threads to both execute this at the exact same time and both see that a is null, and both create their own instances that'll end up being used.  I wasn't sure whether anything in the code required that `a` actually be the same for everyone.  By using an Interlocked.CompareExchange, I ensure that even in the face of such races to initialize the data, every thread ends up using the same instances because only one instance is ever actually published.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
stephentoub: `i` is always non-negative, and formatting a uint is more efficient than formatting an int.  (This code is the same as in .NET 5, I just reordered the blocks.)

veanes:
Thanks for the explanation. The underlying BDDs will actually be the same (identical) because the same BDD algebra is being used that guarantees **extensionality** -- any two equivalent BDDs are actually identical.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs
danmoseley: TIL. I wonder whether that pattern (format inside a for-loop from 0 upwards) is common enough for an analyzer. I guess not.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Replace.cs
danmoseley: As I recall from when we did some data mining in the past, even EcmaScript without RTL is pretty rare. I've not seen it in code either, eg., when folks open bugs. So I'd say this is fine to defer in favor of more impactful work.

stephentoub:
>  The underlying BDDs will actually be the same (identical) because the same BDD algebra is being used that guarantees extensionality -- any two equivalent BDDs are actually identical.

Ok, in that case we can remove the interlockeds.  I'll do that in my next go-around.

Issue #1534, opened 2021-09-08
cleanup namespaces and folder structure of new regex engine


danmoseley:
We likely want to remove SRM from all identifiers and folders. We generally want to use NonBacktracking.

PR #1543, opened 2021-09-09
Move code around in Regex SRM

stephentoub:
Fixes https://github.com/dotnet/runtimelab/issues/1534

- Capitalized folders under Symbolic
- Deleted Utils folder
- Deleted StringUtility.cs, moving Escape into CharSetSolver where it's used
- Renamed Base64 to Base64Utility so as to not conflict with built-in Base64 type
- Renamed a few methods on Base64Utility to include 32 suffix on Int
- Changed Base64Utility.CustomBase64Encoding to be a blittable span (compiler trick)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Base64Utility.cs
```diff
@@ -2,46 +2,26 @@
...
-        };
-
         /// <summary>
         /// Custom Base64 encoder for int.
         /// It uses the standard base 64 alphabet but reordered to 0-9A-Za-z+/ so numbers up to 15 (0-9A-F) coincide with hexadecimal digits.
         /// Negative numbers start with '-'.
         /// </summary>
         public static void Encode(int n, StringBuilder builder)
         {
+            Debug.Assert(n != int.MinValue);
```
stephentoub, line 19: @veanes, I added this assert because Encode will not work correctly in this case, but does anything prevent it from happening?

veanes: From where Encode is used currently (essentially for small integers) the assert cannot arise (unless there is a buggy scenario I cannot see ... what the assert is for). I don't think it is even used for **any** negative numbers although the type is int -- but I'm not 100% sure here ,, but **definitely not** int.Minvalue.
(used for character codes, length and similar).

veanes (approved):
Overall slick changes, I really liked the edits to the base64 encoder/decoder using spans. One thing to keep in mind, is if we get rid of Regex Serialize and Deserialise (where this is primarily used), the only remaining useage would be serializing BDD where what happens is BDD.Deserialize invokes Base64.DecodeInt64Array. That essentially serializes the string to long[]. This could use another (builtin serializer).

veanes:
I meant the initial serialization needs long[] ==> string and then deserilzer string ==> long[].
The essential piece is really BDD ==> long[], and long[] ==> BDD, but how the actual long[] is serialized/deserialized is immaterial. It is used to deserialize
the IgnoreCaseEnUsSerializedBDD, as well as the strings in UnicodeCategoryRanges.

veanes:
Another option would be no string serialization/deserialization for BDD ulong[] arrays. This is what it used to be initially. 
For example the type of  AllCategoriesSerializedBDD would then be ulong[][]. And the type of IgnoreCaseEnUsSerializedBDD would be long[].
The custom base64 string representation for the ulong[] arrays just turned out to be much more compact.

veanes:
As I may edit related code, I'm merging  it.

Issue #1453, opened 2021-08-24
Get CsWinRT going


MichalStrehovsky:
[CsWinRT](https://github.com/microsoft/cswinrt) is a necessary prerequisite for WinUI support.

We're close to getting some of it going but it doesn't work yet. The below code should pop a notification. Nothing happens instead. I haven't tried to troubleshoot where the problem is (is there a problem in the XmlDocument? Maybe try saving it to a StorageFile first? Etc.)

Eventually we'll also need to make CsWinRT AOT and trimming friendly. It's neither of those right now.

CSPROJ:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
    <TrimMode>link</TrimMode>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="7.0.0-*" />
    <RdXmlFile Include="rd.xml" />
  </ItemGroup>
</Project>
```

RD.XML:

```xml
<Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
  <Application>
    <Assembly Name="Microsoft.Windows.SDK.NET">
      <Type Name="ABI.Windows.UI.Notifications.IToastNotificationManagerStatics" Dynamic="Required All" />
      <Type Name="ABI.Windows.Data.Xml.Dom.IXmlDocument" Dynamic="Required All" />
      <Type Name="Windows.Data.Xml.Dom.XmlNodeList" Dynamic="Required All" />
      <Type Name="ABI.Windows.Data.Xml.Dom.IXmlNodeList" Dynamic="Required All" />

      <Type Name="ABI.Windows.Data.Xml.Dom.IXmlNode" Dynamic="Required All" />
      <Type Name="ABI.Windows.Data.Xml.Dom.IXmlNodeSerializer" Dynamic="Required All" />
      <Type Name="ABI.Windows.Data.Xml.Dom.IXmlNodeSelector" Dynamic="Required All" />
      <Type Name="ABI.Windows.Data.Xml.Dom.IXmlDocumentIO" Dynamic="Required All" />
      <Type Name="ABI.Windows.Data.Xml.Dom.IXmlDocumentIO2" Dynamic="Required All" />
[...]

PR #1544, opened 2021-09-09
Implement ComWrappers.RegisterForTrackerSupport to be able create CCW

kant2002:
This moves a bit towards #306 and #1453

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -720,7 +720,6 @@ namespace System.Runtime.InteropServices
         private static unsafe IntPtr CreateDefaultIReferenceTrackerTargetVftbl()
         {
             IntPtr* vftbl = (IntPtr*)RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(ComWrappers), 7 * sizeof(IntPtr));
-            GetIUnknownImpl(out vftbl[0], out vftbl[1], out vftbl[2]);
             vftbl[0] = (IntPtr)(delegate* unmanaged<IntPtr, Guid*, IntPtr*, int>)&ComWrappers.IReferenceTrackerTarget_QueryInterface;
             vftbl[1] = (IntPtr)(delegate* unmanaged<IntPtr, uint>)&ComWrappers.IUnknown_AddRef;
             vftbl[2] = (IntPtr)(delegate* unmanaged<IntPtr, uint>)&ComWrappers.IUnknown_Release;
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -489,9 +489,6 @@ namespace System.Runtime.InteropServices
             if (flags.HasFlag(CreateObjectFlags.Aggregation))
                 throw new NotImplementedException();
 
-            if (flags.HasFlag(CreateObjectFlags.TrackerObject))
-                throw new NotImplementedException();
-
             if (flags.HasFlag(CreateObjectFlags.Unwrap))
             {
                 var comInterfaceDispatch = TryGetComInterfaceDispatch(externalComObject);
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
@@ -196,12 +196,12 @@ namespace System.Runtime.InteropServices
 
             private unsafe IntPtr AsRuntimeDefined(in Guid riid)
             {
-                int i = UsedDefinedCount;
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -114,12 +196,30 @@ public unsafe void Destroy()
...
+                    {
+                        if (riid == IID_IReferenceTrackerTarget)
+                        {
+                            return (IntPtr)(Dispatches + UserDefinedCount + 1);
+                        }
+                    }
+                }
+                else
+                {
+                    if ((Flags & CreateComInterfaceFlagsEx.TrackerSupport) == CreateComInterfaceFlagsEx.TrackerSupport)
```
jkotas: It would be nice to avoid this duplication by maintaining a running index:

```
int i = UsedDefinedCount;

if ((Flags & CreateComInterfaceFlagsEx.CallerDefinedIUnknown) == CreateComInterfaceFlagsEx.None)
{
    if (riid == IID_IUnknown)
    {
        return (IntPtr)(Dispatches + i);
    }
    i++;
}

if ((Flags & CreateComInterfaceFlagsEx.TrackerSupport) == CreateComInterfaceFlagsEx.TrackerSupport)
{
....
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -240,6 +365,11 @@ public unsafe IntPtr GetOrCreateComInterfaceForObject(object instance, CreateCom
                 runtimeDefinedVtable[runtimeDefinedCount++] = DefaultIUnknownVftblPtr;
             }
 
+            if ((flags & CreateComInterfaceFlags.TrackerSupport) == CreateComInterfaceFlags.TrackerSupport)
```
jkotas:
```suggestion
            if ((flags & CreateComInterfaceFlags.TrackerSupport) != 0)
```
Shorted and less duplication

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -114,11 +196,22 @@ public unsafe void Destroy()
...
+
+                    i++;
+                }
+
+                if ((Flags & CreateComInterfaceFlagsEx.TrackerSupport) != 0)
+                {
+                    if (riid == IID_IReferenceTrackerTarget)
+                    {
+                        return (IntPtr)(Dispatches + i);
                     }
```
jkotas, line 215:
```suggestion
                    }

                    i++;
```
For symmetry, in case more interfaces are added later.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -364,6 +489,9 @@ public object GetOrRegisterObjectForComInstance(IntPtr externalComObject, Create
             if (flags.HasFlag(CreateObjectFlags.Aggregation))
                 throw new NotImplementedException();
 
+            if (flags.HasFlag(CreateObjectFlags.TrackerObject))
```
jkotas: Just to make sure that I understand this right: This blocks the newly added code from getting executed for now. Is that correct?

jkotas (approved):
LGTM

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
kant2002: I implement only CCW. I thought that throw that for RCW was a right way, but if I remove that lines, then app from #1453 is basically working (except it does not produce any Toast for me). But exit code == 0. So I would remove these lines, and let it leak. At least you wasn't agains that previously.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/ComWrappers.CoreRT.cs
```diff
@@ -554,11 +675,60 @@ internal static unsafe uint IUnknown_Release(IntPtr pThis)
...
         {
             IntPtr* vftbl = (IntPtr*)RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(ComWrappers), 3 * sizeof(IntPtr));
             GetIUnknownImpl(out vftbl[0], out vftbl[1], out vftbl[2]);
             return (IntPtr)vftbl;
         }
+
+        private static unsafe IntPtr CreateDefaultIReferenceTrackerTargetVftbl()
+        {
+            IntPtr* vftbl = (IntPtr*)RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(ComWrappers), 7 * sizeof(IntPtr));
+            GetIUnknownImpl(out vftbl[0], out vftbl[1], out vftbl[2]);
```
jkotas: The pointers returned by `GetIUnknownImpl` are immediately overridden by next 3 lines. Redundant?

kant2002: Yes. Removed.

PR #1547, opened 2021-09-10
Srm unit test edits

veanes:
Lifted more unit test to also handle NonBacktracking, including most Culture tests and all MultiMatching test (only excluding cases that are not supported with NonBacktracking).
Currently investigating a failing case for Turkish i, that should clearly be working because it is specifically supported in NonBacktracking mode, but it seems that some culture option is not being propagated correctly.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/tests/Regex.MultipleMatches.Tests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/Regex.MultipleMatches.Tests.cs
@@ -48,11 +48,16 @@ namespace System.Text.RegularExpressions.Tests
             }
         }
 
-        public static IEnumerable<object[]> Matches_TestData()
+        public static IEnumerable<object[]> Matches_TestData_NonBacktracking() =>
+            PlatformDetection.IsNetFramework ? RegexHelpers.NoTestData() : Matches_TestData_WithOptions(RegexHelpers.RegexOptionNonBacktracking);   
+
+        public static IEnumerable<object[]> Matches_TestData() => Matches_TestData_WithOptions(RegexOptions.None);   
+
+        private static IEnumerable<object[]> Matches_TestData_WithOptions(RegexOptions options)
         {
             yield return new object[]
             {
-                "[0-9]", "12345asdfasdfasdfljkhsda67890", RegexOptions.None,
+                "[0-9]", "12345asdfasdfasdfljkhsda67890", options,
                 new CaptureData[]
                 {
                     new CaptureData("1", 0, 1),
@@ -70,7 +75,7 @@ namespace System.Text.RegularExpressions.Tests
 
             yield return new object[]
             {
-                "[a-z0-9]+", "[token1]? GARBAGEtoken2GARBAGE ;token3!", RegexOptions.None,
+                "[a-z0-9]+", "[token1]? GARBAGEtoken2GARBAGE ;token3!", options,
                 new CaptureData[]
                 {
                     new CaptureData("token1", 1, 6),
@@ -91,7 +96,7 @@ namespace System.Text.RegularExpressions.Tests
 
             yield return new object[]
             {
-                @"\b\w*\b", "handling words of various lengths", RegexOptions.None,
+                @"\b\w*\b", "handling words of various lengths", options,
                 new CaptureData[]
                 {
                     new CaptureData("handling", 0, 8),
@@ -109,7 +114,7 @@ namespace System.Text.RegularExpressions.Tests
 
             yield return new object[]
             {
-                @"\b\w{2}\b", "handling words of various lengths", RegexOptions.None,
+                @"\b\w{2}\b", "handling words of various lengths", options,
                 new CaptureData[]
                 {
                     new CaptureData("of", 15, 2),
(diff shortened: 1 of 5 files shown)


veanes:
I added extra unit test for Turkish i in NonBacktracking mode. 
It seems a no-brainer (to me) that this is how the behavior should be. 
Please correct me if I'm wrong and suggest a fix then.
I'm a bit concerned about the difference. 
To me the other test is sort of like declaring what looks like a bug as an expected behavior.
Well, this is not the only place, \b vs \w is clearly the other case.

veanes:
I added additional failing unit test (`TurkishCulture_Handling_Of_IgnoreCase`) as active issue. This is totally independent of the NonBacktracking case. It clearly points to a bug and is further evidence supporting my claim above that the unit test is testing incorrect behavior.

veanes:
As I was lifting the SRM culture related tests (in the last commit) to None and Compiled, I found some discrepancies. I spent some time on figuring this out. Seems in all cases `NonBacktracking` gives the right result but in some cases `None` and `Compiled` give not only wrong but **different** answers in `InvariantCulture`. I marked a sample unit test 
`Match_InvariantCulture_None_vs_Compiled` as an **active issue**.

danmoseley:
can you rebase on head? that would take Stephen's change out of the diff.

src/libraries/System.Text.RegularExpressions/tests/Regex.MultipleMatches.Tests.cs
```diff
@@ -127,19 +132,23 @@ public static IEnumerable<object[]> Matches_TestData()
                 }
             };
 
-            yield return new object[]
+            // RegexOptions.RightToLeft is not supported with NonBacktracking
+            if (options != RegexHelpers.RegexOptionNonBacktracking)
```
danmoseley: would it be easier to segregate these into their own enumerator? Matches_TestData_WithOptions and Matches_TestData_WithOptions_Backtracking maybe? then you can reason about what tests are in what by just looking at which enumerators are used, with no if's.

src/libraries/System.Text.RegularExpressions/tests/RegexCultureTests.cs
```diff
@@ -119,5 +141,137 @@ Regex[] Create(string input, CultureInfo info, RegexOptions additional)
...
+        [Theory]
+        [MemberData(nameof(Match_In_Different_Cultures_TestData))]
+        public void Match_In_Different_Cultures(string pattern, RegexOptions options, CultureInfo culture, string input, string match_expected)
+        {
+            Regex r = Create(pattern, culture, options)[0];
+            var match = r.Match(input);
+            Assert.Equal(match_expected, match.Value);
+        }
+
+        [ActiveIssue("Incorrect result of match in complied mode in Invariant culture, https://github.com/dotnet/runtime/issues/58956")]
```
danmoseley, line 295: Maybe have a 2nd test with ActiveIssue disabling it and the commented line below uncommented. So when the bug is fixed, we just remove ActiveIssue.

danmoseley: Oh never mind, I didn't see that Create was creating several. Maybe a name like CreateCompiledAndInterpretedRegexes?

src/libraries/System.Text.RegularExpressions/tests/RegexCultureTests.cs
```diff
@@ -119,5 +141,137 @@ Regex[] Create(string input, CultureInfo info, RegexOptions additional)
...
+                    yield return new object[] { "(?i:iI+)", options, turkish, "abc\u0130Iixyz", "\u0130I" };            // <-- failing for None, Compiled
+                }
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(Match_In_Different_Cultures_TestData))]
+        public void Match_In_Different_Cultures(string pattern, RegexOptions options, CultureInfo culture, string input, string match_expected)
+        {
+            Regex r = Create(pattern, culture, options)[0];
```
danmoseley: this can just new up the regex itself here?

danmoseley (approved):

src/libraries/System.Text.RegularExpressions/tests/RegexCultureTests.cs
veanes: right, it was just convenient to use the existing Create method for passing the culture, else I need to set and then undo setting the thread culture. But I can add another Create1 method that does not create 2 regexes.

src/libraries/System.Text.RegularExpressions/tests/Regex.MultipleMatches.Tests.cs
veanes: Yes, will do.

src/libraries/System.Text.RegularExpressions/tests/RegexCultureTests.cs
veanes: Yes, will do.

src/libraries/System.Text.RegularExpressions/tests/RegexCultureTests.cs
danmoseley: Or just have one CreateRegex method that takes culture & options and takes care of the culture reset. Tests that need both call it twice once with Compiled and once without. Would probably be clearer in the test.

veanes: I added `RegexHelpers.CreateRegexInCulture(pattern, options, culture)`

PR #1552, opened 2021-09-12
Srm update

veanes:
- Updated `ToString` of `SymbolicRegexNode` to use `StringBuilder` internally.

- Eliminated `IfThenElse` node from the AST of `SymbolicRegexNode`. 
Replaced it by more restricted `Not` node.
Now the general case is not supported. 
It will now also be very easy to only allow And and Not in DEBUG build. 
with a small localized change to make this 
part conditional in DEBUG mode:

```
                    if (node.Type == RegexNode.Testgroup)
                    {
                        // Try to extract the special case representing complement or intersection
                        if (IsComplementedNode(node))
                        {
                            return _builder.MkNot(Convert(node.Child(0), false));
                        }
                        List<RegexNode> conjuncts = new();
                        if (TryGetIntersection(node, conjuncts))
                        {
                            return _builder.MkAnd(Array.ConvertAll(conjuncts.ToArray(), x => Convert(x, false)));
                        }
                    }
```

I also added the lines
```
        /// <summary>test conditional (?( test-pattern ) yes-pattern | no-pattern )</summary>
        internal static string @ExpressionDescription_IfThenElse => GetResourceString("@ExpressionDescription_IfThenElse", @"test conditional (?( test-pattern ) yes-pattern | no-pattern )");
```
into `System.SR.cs` but then noticed this file is autogenerated, not sure where/waht to edit so this gets added. So I'lll comment out the following line (or I guess else the build will fail)
        `//RegexNode.Testgroup => SR.ExpressionDescription_IfThenElse,`
in `RegexNodeToSymbolicConverter`

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
@@ -278,20 +278,22 @@ namespace System.Text.RegularExpressions.Symbolic
                 case RegexNode.Nothing:
                     return _builder._nothing;
 
-                default:
-                    if (node.Type == RegexNode.Testgroup)
+                case RegexNode.Testgroup:
+                    // Try to extract the special case representing complement or intersection
+                    if (IsComplementedNode(node))
                     {
-                        // Try to extract the special case representing complement or intersection
-                        if (IsComplementedNode(node))
-                        {
-                            return _builder.MkNot(Convert(node.Child(0), false));
-                        }
-                        List<RegexNode> conjuncts = new();
-                        if (TryGetIntersection(node, conjuncts))
-                        {
-                            return _builder.MkAnd(Array.ConvertAll(conjuncts.ToArray(), x => Convert(x, false)));
-                        }
+                        return _builder.MkNot(Convert(node.Child(0), false));
                     }
+
+                    List<RegexNode>? conjuncts;
+                    if (TryGetIntersection(node, out conjuncts))
+                    {
+                        return _builder.MkAnd(Array.ConvertAll(conjuncts.ToArray(), x => Convert(x, false)));
+                    }
+
+                    goto default;
+
+                default:
                     throw new NotSupportedException(SR.Format(SR.NotSupported_NonBacktrackingConflictingExpression, node.Type switch
                     {
                         RegexNode.Testgroup => SR.ExpressionDescription_IfThenElse,
@@ -440,13 +442,15 @@ namespace System.Text.RegularExpressions.Symbolic
 
             bool IsIntersect(RegexNode node) => node.Type == RegexNode.Testgroup && IsNothing(node.Child(2));
 
-            bool TryGetIntersection(RegexNode node, List<RegexNode> conjuncts)
+            bool TryGetIntersection(RegexNode node, [Diagnostics.CodeAnalysis.NotNullWhen(true)] out List<RegexNode>? conjuncts)
             {
                 if (!IsIntersect(node))
                 {
+                    conjuncts = null;
                     return false;
                 }
 
+                conjuncts = new();
                 conjuncts.Add(node.Child(0));
                 node = node.Child(1);
                 while (IsIntersect(node))
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
@@ -433,7 +433,7 @@ namespace System.Text.RegularExpressions.Symbolic
                 return _builder.MkLoop(body, isLazy, node.M, node.N);
             }
 
-            // TODO: recognizing strictly only [] (RegexNode.Nothing), for example [0-[0]] would not be regonized
+            // TODO: recognizing strictly only [] (RegexNode.Nothing), for example [0-[0]] would not be recognized
             bool IsNothing(RegexNode node) => node.Type == RegexNode.Nothing || (node.Type == RegexNode.Set && ConvertSet(node).IsNothing);
 
             bool IsDotStar(RegexNode node) => node.Type == RegexNode.Setloop && Convert(node, false).IsDotStar;
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
@@ -294,7 +294,7 @@ namespace System.Text.RegularExpressions.Symbolic
                     }
                     throw new NotSupportedException(SR.Format(SR.NotSupported_NonBacktrackingConflictingExpression, node.Type switch
                     {
-                        RegexNode.Testgroup => SR.ExpressionDescription_IfThenElse,
+                        //RegexNode.Testgroup => SR.ExpressionDescription_IfThenElse,
                         RegexNode.Ref => SR.ExpressionDescription_Backreference,
                         RegexNode.Testref => SR.ExpressionDescription_Conditional,
                         RegexNode.Require => SR.ExpressionDescription_PositiveLookaround,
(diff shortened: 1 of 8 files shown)


stephentoub:
> noticed this file is autogenerated

It's generated from the Strings.resx file.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
```diff
@@ -421,6 +432,32 @@ SymbolicRegexNode<BDD> ConvertSetloop(RegexNode node, bool isLazy)
...
+            bool IsDotStar(RegexNode node) => node.Type == RegexNode.Setloop && Convert(node, false).IsDotStar;
+
+            bool IsIntersect(RegexNode node) => node.Type == RegexNode.Testgroup && IsNothing(node.Child(2));
+
+            bool TryGetIntersection(RegexNode node, List<RegexNode> conjuncts)
+            {
+                if (!IsIntersect(node))
+                {
+                    return false;
+                }
```
stephentoub, line 451:
```suggestion
                }

```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
```diff
@@ -421,6 +432,32 @@ SymbolicRegexNode<BDD> ConvertSetloop(RegexNode node, bool isLazy)
...
+                conjuncts.Add(node.Child(0));
+                node = node.Child(1);
+                while (IsIntersect(node))
+                {
+                    conjuncts.Add(node.Child(0));
+                    node = node.Child(1);
+                }
```
stephentoub, line 460:
```suggestion
                }

```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexSet.cs
```diff
@@ -316,37 +316,72 @@ public override bool Equals([NotNullWhen(true)] object? obj)
...
+                        {
+                            sb.Append("(?(");
+                            node.ToString(sb);
+                            sb.Append(')');
+                            node = enumerator.Current;
+                            count++;
+                        }
+                        node.ToString(sb);
+                        while (count-- > 0)
+                            sb.Append($"|{SymbolicRegexNode<S>.EmptyCharClass})");
```
stephentoub:
```suggestion

                        while (count-- > 0)
                        {
                            sb.Append($"|{SymbolicRegexNode<S>.EmptyCharClass})");
                        }
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexSet.cs
```diff
@@ -316,37 +316,72 @@ public override bool Equals([NotNullWhen(true)] object? obj)
...
+            {
+                sb.Append(".*");
+            }
+            else
             {
-                if (_kind == SymbolicRegexKind.Or)
+                Enumerator enumerator = GetEnumerator();
+                bool nonempty = enumerator.MoveNext();
+                // Collection must be nonempty because IsNothing is false and IsEverything is false
+                Debug.Assert(nonempty);
```
stephentoub:
```suggestion
                Debug.Assert(nonempty, "Collection must be nonempty because IsNothing is false and IsEverything is false");
```

veanes:
In the last commit addressed two deep recursion issues that showed up when I temporarily increased the the limit uised in `RegexMatchTests.StressTestDeepNestingOfConcat` to `RegexHelpers.StressTestNestingDepth * 10`. Removed this increase, since the test became too slow (but no longer crashed). The two issues were **deep recursion crashes** in `SymbolicRegexNode.GetHashCode()` and also in the computation of `SymbolicRegexNode.GetPrefix()`.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
```diff
@@ -282,8 +279,22 @@ public SymbolicRegexNode<BDD> Convert(RegexNode node, bool topLevel)
...
+                        if (IsComplementedNode(node))
+                        {
+                            return _builder.MkNot(Convert(node.Child(0), false));
+                        }
+                        List<RegexNode> conjuncts = new();
+                        if (TryGetIntersection(node, conjuncts))
+                        {
+                            return _builder.MkAnd(Array.ConvertAll(conjuncts.ToArray(), x => Convert(x, false)));
+                        }
+                    }
```
stephentoub: Can you do this instead as:
```C#
case RegexNode.Testgroup:
    // Try to extract the special case representing complement or intersection
    if (IsComplementedNode(node))
    {
        return _builder.MkNot(Convert(node.Child(0), false));
    }

    List<RegexNode> conjuncts = new();
    if (TryGetIntersection(node, conjuncts))
    {
        return _builder.MkAnd(Array.ConvertAll(conjuncts.ToArray(), x => Convert(x, false)));
    }

    goto default;

default:
    throw new NotSupportedException(...
```
?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
```diff
@@ -421,6 +432,34 @@ SymbolicRegexNode<BDD> ConvertSetloop(RegexNode node, bool isLazy)
                 SymbolicRegexNode<BDD> body = _builder.MkSingleton(moveCond);
                 return _builder.MkLoop(body, isLazy, node.M, node.N);
             }
+
+            // TODO: recognizing strictly only [] (RegexNode.Nothing), for example [0-[0]] would not be regonized
```
stephentoub:
```suggestion
            // TODO: recognizing strictly only [] (RegexNode.Nothing), for example [0-[0]] would not be recognized
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
```diff
@@ -421,6 +432,34 @@ SymbolicRegexNode<BDD> ConvertSetloop(RegexNode node, bool isLazy)
...
+            bool IsNothing(RegexNode node) => node.Type == RegexNode.Nothing || (node.Type == RegexNode.Set && ConvertSet(node).IsNothing);
+
+            bool IsDotStar(RegexNode node) => node.Type == RegexNode.Setloop && Convert(node, false).IsDotStar;
+
+            bool IsIntersect(RegexNode node) => node.Type == RegexNode.Testgroup && IsNothing(node.Child(2));
+
+            bool TryGetIntersection(RegexNode node, List<RegexNode> conjuncts)
```
stephentoub: Rather than having the caller always allocate the `List<RegexNode>`, even though it'll only be used when TryGetIntersection returns true, it seems like it'd be better to make the parameter be `[NotNullWhen(true)] out List<RegexNode>? conjuncts`, and have the method set it to null when returning false and `new()` when returning true and populating it.

veanes: will do

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
veanes: ok

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
```diff
@@ -865,6 +839,11 @@ internal SymbolicRegexNode<S> MkDerivative(S elem, uint context)
         }
 
         public override int GetHashCode()
+        {
+            return _hashcode;
+        }
+
+        private void EnsureHashCode()
```
stephentoub: Since we're now always calling this in the ctor, can this instead be `private int ComputeHashCode();` and then the field can be `private readonly int _hashcode;` and the ctor can do `_hashcode = ComputeHashCode();`?

stephentoub (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
veanes: yes, was considering that actually, will do

PR #1554, opened 2021-09-14
fixed some deep recursion related bugs

veanes:
- Added internalization of regexes. This was needed in particular to avoid deep recursion of `SymbolicRegexNode.Equals` causing out of stack memory crashes. This affected some constructor logic. Added a related stress test that was failing before the edits.
- Cleaned up some more unit tests by migrating and generalizing them from SRM to their appropriate places.
- Deleted unused left-over code/methods related functionality not used any more.


veanes:
In the last commit, added local caching of already computed derivatives in symbolic regex nodes. Typically these caches will only have a handful of entries each and are only created for already internalized regexes. In particular this addresses a case that came up that for very long (say malicious input) concat `R_{n+1} = A_n(R_n})` where each `A_i` is nullable, say `n>100` and `A_i = [a-z]*`, then `D(x,R_{k+1}) = D(x,A_k)R_k|D(x,R_k)` will  otherwise create in the order of `O(n^2)` behavior of re-computing the same derivatives over and over again.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
```diff
@@ -737,13 +717,36 @@ internal SymbolicRegexNode<S> MkDerivative(S elem, uint context)
...
+                // Internalize the node before proceeding
+                // this node could end up being internalized or replaced by
+                // an already previously created object (!= this)
+                SymbolicRegexNode<S> this_internalized = Internalize(this);
+                Debug.Assert(this_internalized._isInternalizedUnion);
+                if (this_internalized != this)
+                {
+                    return this_internalized.MkDerivative(elem, context);
+                }
+            }
```
olsaarik: This code looks like it might cause states in Antimirov mode to be internalized. I'm guessing Antimirov mode doesn't actually directly call this function on the top level element? A comment on that here would be helpful.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
```diff
@@ -1404,10 +1411,12 @@ S ComputeStartSet()
                         {
                             Debug.Assert(_left is not null && _right is not null);
                             S startSet = _left.GetStartSet();
+                            // To avoid deep recursion of trying to get the startset of right
+                            // just return True when left can be nullable
+                            // It is always correct to return True as the startset
```
olsaarik: Does this change disable the startset deduction for patterns like `a?bR`? This should have a startset of `[ab]` but with the new change would have `.` if I understand this correctly, so wouldn't benefit from startset optimization. If so this piece of code is probably a target for an iterative rewrite.

veanes: You are right, there should be a smarter check here. Not sure at this point what that would be.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
veanes: fixed in the last commit

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
veanes: After some thought, I know how to fix this cleanly. I'll compute start-sets for all nodes at construction time. Similar to CoputeHashCode. This adds virtually no overhead and eliminates the recursion issue and is more efficient also. Startsets are used frequently so its seems not an overkill to add this extra info to each node.

veanes: fixed

olsaarik (approved):

PR #1556, opened 2021-09-16
password search unit test with conjunction and complement

veanes:
A sophisticated unit test demonstrating the capability and the performance of handling intersection and complement by NonBacktracking. Arguably the same search is not possible by other means **and no other regex matcher**. **If I'm wrong I would really like to see how**. Pay attention to the printed timing output -- the search is very performant.
(I've added the constraint so that random text in between would have a match with **very low** likelihood so the test may fail very rarely -- something to fix with appropriate fixed input rather than randomly generated blobs.)

I really like this test. It somehow highlights a hidden gem with the engine. Something we should definitely revisit later. I added a `RegexOptions.Extended` note that I discussed with Olli in the board also.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
@@ -409,6 +409,21 @@ namespace System.Text.RegularExpressions.Symbolic
             }
         }
 
+        /// <summary>Returns true if this is equivalent to [\0-\xFFFF] </summary>
+        public bool IsDot
+        {
+            get
+            {
+                if (_kind == SymbolicRegexKind.Singleton)
+                {
+                    Debug.Assert(_set is not null);
+                    return _builder._solver.AreEquivalent(_builder._solver.True, _set);
+                }
+
+                return false;
+            }
+        }
+
         /// <summary>Returns true if this is equivalent to [0-[0]]</summary>
         public bool IsNothing
         {
@@ -482,8 +497,11 @@ namespace System.Text.RegularExpressions.Symbolic
         internal static SymbolicRegexNode<S> MkNWBAnchor(SymbolicRegexBuilder<S> builder) =>
             Create(builder, SymbolicRegexKind.NWBAnchor, null, null, -1, -1, default, null, SymbolicRegexInfo.Mk(startsWithBoundaryAnchor: true, canBeNullable: true));
 
-        internal static SymbolicRegexNode<S> MkDotStar(SymbolicRegexBuilder<S> builder, SymbolicRegexNode<S> body) =>
+        internal static SymbolicRegexNode<S> MkStar(SymbolicRegexBuilder<S> builder, SymbolicRegexNode<S> body) =>
             Create(builder, SymbolicRegexKind.Loop, body, null, 0, int.MaxValue, default, null, SymbolicRegexInfo.Loop(body._info, 0, false));
+
+        internal static SymbolicRegexNode<S> MkPlus(SymbolicRegexBuilder<S> builder, SymbolicRegexNode<S> body) =>
+           Create(builder, SymbolicRegexKind.Loop, body, null, 1, int.MaxValue, default, null, SymbolicRegexInfo.Loop(body._info, 1, false));
 
         #endregion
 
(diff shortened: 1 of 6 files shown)


src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -465,5 +472,81 @@ public void SRMTest_ComplementFindMatch()
...
+                _output.WriteLine($@"k={k}, t={t}ms");
+
+                Assert.True(match1.Success);
+                Assert.Equal(expextedMatch1Index, match1.Index);
+                Assert.Equal(expextedMatch1Length, match1.Length);
+                Assert.Equal(password1, match1.Value);
+
+                Assert.True(match2.Success);
+                Assert.Equal(expextedMatch2Index, match2.Index);
+                Assert.Equal(expextedMatch2Length, match2.Length);
```
danmoseley, line 545:
```suggestion
                int expectedMatch1Index = (2 * k) + almostPassw1.Length + 3;
                int expectedMatch1Length = password1.Length;

                int expectedMatch2Index = (3 * k) + almostPassw1.Length + password1.Length + 5;
                int expectedMatch2Length = password2.Length;

                // Random text hiding almostPassw and password
                int t = System.Environment.TickCount;
                Match match1 = re.Match(input);
                Match match2 = match1.NextMatch();
                Match match3 = match2.NextMatch();
                t = System.Environment.TickCount - t;

                _output.WriteLine($@"k={k}, t={t}ms");

                Assert.True(match1.Success);
                Assert.Equal(expectedMatch1Index, match1.Index);
                Assert.Equal(expectedMatch1Length, match1.Length);
                Assert.Equal(password1, match1.Value);

                Assert.True(match2.Success);
                Assert.Equal(expectedMatch2Index, match2.Index);
                Assert.Equal(expectedMatch2Length, match2.Length);
```

danmoseley (approved):

danmoseley:
Could you help me understand the `And` method? Why does it need to nest the patterns with alternations with empty (ie `|[0-[0]]`) (rather than just listing them all as a single alternation (ie a string with `|` between each) as I believe credscan does)

veanes:
> Could you help me understand the `And` method? Why does it need to nest the patterns with alternations with empty (ie `|[0-[0]]`) (rather than just listing them all as a single alternation (ie a string with `|` between each) as I believe credscan does)

We are currently just temporarily using (?(A)B|[0-[0]]) to represent the intersection (A & B) (which does not exist), also cannot write (?(A)B|[]) because this is not allowed by the parser.
I also think that (?(A)?(B)?(C)|[0-[0]]) is invalid syntax if I tried instead of
(?(A)(?(B)?C|[0-[0]])|[0-[0]])
to avoid the (rather ugly) nesting.
 But this weird looking way of writing conjunction gives a way to test the algorithm at least. This will only be enabled in DEBUG mode. In release we will turn off allowing if-then-else until we figure out how to best support & (intersection) and ~ (complement).
The last commit adds a dual test (that discovered a need for computing NNF (Negation Normal Form) when constructing a complement. NNF uses extended deMorgan's laws (the ones we prove in the paper actually). 
Adding Olli here to look this code over also.

veanes:
Note also, the tests and the use of **conjunction** I added is not something CredScan uses today. They are interested in using conjunction though. 
It is not possible to get around this without complement. 
The dual test writes ~(~A|~B) to encode (A & B) --- that is possible, but also ugly since 
~A is represented by (?(A)[0-[0]]|[\0-\xFFFF]*)  (not using .* because meaning of . may depend on options)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
```diff
@@ -409,6 +409,21 @@ public bool IsDotStar
             }
         }
 
+        /// <summary>Returns true if this is equivalent to [\0-\xFFFF] </summary>
+        public bool IsDot
```
stephentoub: This naming is a little confusing, since `.` maps to `[^\n]` unless RegexOptions.Singleline is provided.  This calls it "AnyClass":
https://github.com/dotnet/runtime/blob/c5e31e676b5f486741543c4e54f6b762cfae6c12/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCharClass.cs#L73
Maybe we do the same here?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
```diff
@@ -572,8 +590,95 @@ internal static SymbolicRegexNode<S> MkConcat(SymbolicRegexBuilder<S> builder, S
...
+            // Do this transformation in-line without recursion, to avoid any chance of deep recursion
+            // OBSERVE: NNF[node] represents the Negation Normal Form of ~node
+            Dictionary<SymbolicRegexNode<S>, SymbolicRegexNode<S>> NNF = new();
+            Stack<(SymbolicRegexNode<S>, bool)> todo = new();
+            todo.Push((root, false));
+            while (todo.Count > 0)
+            {
+                var top = todo.Pop();
+                bool secondTimePushed = top.Item2;
+                var node = top.Item1;
```
stephentoub: Nit: we try to avoid using `var` unless the type is explicit on the right-hand side

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
```diff
@@ -465,5 +472,159 @@ public void SRMTest_ComplementFindMatch()
...
+            var re = new Regex($@"\b{all}\b", RegexOptions.NonBacktracking | RegexOptions.Singleline);
+
+            // Does not qualify because of 123 and connot end between 2 and 3 because of \b
+            var almostPassw1 = "P@ssW0rd123";
+            // Does not have at least two uppercase
+            var almostPassw2 = "P@55w0rd";
+
+            // These two qualify
+            var password1 = "P@55W0rd";
+            var password2 = "Pa5$w00rD";
```
stephentoub: That goes for tests as well :)

stephentoub (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
veanes: Sounds good. I'll rename IsDot to IsAnyChar and IsDotStar to IsAnyStar

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
veanes: ok, will fix.

src/libraries/System.Text.RegularExpressions/tests/RegexSRMTests.cs
veanes: Ok, will fix. I'll try to keep this in mind.

PR #1557, opened 2021-09-16
Add support for module initializers

### LINKED ISSUE #193: Support module constructors
We'll probably need to run them at process start to avoid perf impact of having to check everywhere (implementing a "module load emulator").

.NET Native simply did a topological sort of the dependencies of the individual cctors and ran them in that order.

Now it's somewhat higher priority since C# finally added support for them.

Loader/classloader/regressions/529206/vsw529206ModuleCctor
Loader/classloader/regressions/vsw529206/vsw529206ModuleCctor

Also a failing System.Runtime test.

MichalStrehovsky:
Fixes #193.

Since we don't have a module load event (everything is "loaded" at compile time), we can't get the same module initializer timing as CoreCLR. Since the spec allows running them at arbitrary time before accessing things within the module, we simply run them before `Main`. We order them so that if module X references module Y, we run Y's initializer before X.

We could also potentially look into running them at compile time (same as other cctors).


src/coreclr/nativeaot/Common/src/Internal/Runtime/CompilerHelpers/StartupCodeHelpers.cs
```diff
@@ -208,6 +208,27 @@ private static unsafe void RunEagerClassConstructors(IntPtr cctorTableStart, int
             }
         }
 
+        internal static unsafe void RunModuleInitializers()
```
jkotas, line 211: Would it be better to just emit the calls directly instead of going through the pointers?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/CodeBasedDependencyAlgorithm.cs
```diff
@@ -23,6 +23,12 @@ public static void AddDependenciesDueToMethodCodePresence(ref DependencyList dep
 
             factory.InteropStubManager.AddDependeciesDueToPInvoke(ref dependencies, factory, method);
 
+            if (method.OwningType is MetadataType mdOwningType)
+            {
+                dependencies ??= new DependencyList();
+                dependencies.Add(factory.ModuleUse(mdOwningType.Module), "Module with a used method");
```
jkotas: This will add a lot of dependencies that are useless in nearly 100% cases. These dependencies only matter when the module has static constructor. Should we skip adding these dependencies when the module does not have static constructor (e.g. return null from `ModuleUse`)?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ModuleInitializerListNode.cs
```diff
@@ -0,0 +1,225 @@
...
+
+            // We have a list of modules with a class constructor.
+            // Do a topological sort based on the assembly references of each module.
+            // This is an approximation that tries to deal with module initializers that might have
+            // dependencies on each other. The spec doesn't guarantee any ordering so this is best effort.
+            List<ModuleDesc> sortedModules = new List<ModuleDesc>();
+
+            var graphFactory = new ModuleGraphFactory();
+
+            // This is a list because we want to keep a stable sort order.
```
jkotas, line 65: This is populated from `HashSet. Do we also need to do something to make the HashSet order stable?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ModuleInitializerListNode.cs
```diff
@@ -0,0 +1,225 @@
...
+            if (relocsOnly)
+                return new ObjectData(Array.Empty<byte>(), Array.Empty<Relocation>(), 1, new ISymbolDefinitionNode[] { this });
+
+            
+            var modulesWithCctor = new HashSet<ModuleDesc>();
+
+            foreach (var methodNode in factory.MetadataManager.GetCompiledMethodBodies())
+            {
+                if (methodNode.Method.IsStaticConstructor && methodNode.Method.OwningType is MetadataType mdType
+                    && mdType.IsModuleType)
```
jkotas: Nit: `IsModuleType` check is probably faster. We may want to do it first.

src/coreclr/nativeaot/Common/src/Internal/Runtime/CompilerHelpers/StartupCodeHelpers.cs
jkotas: (It would be insufficient for multi-module mode, but the multi-module mode needs a more complex data structure to deal with the sorting problem anyway.)

MichalStrehovsky: My line of thinking was that the current structure is closer to what we'll need for multi-module (and it actually works for multi-module, but with no ordering guarantees). If we emit the calls directly, it stops working for multi-module completely and cannot be extended to work for it.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/CodeBasedDependencyAlgorithm.cs
MichalStrehovsky: Yeah, maybe we can just get rid of ModuleUse and emit a dependency on the cctor directly.

src/coreclr/nativeaot/Common/src/Internal/Runtime/CompilerHelpers/StartupCodeHelpers.cs
jkotas: sounds reasonable

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/ModuleInitializerListNode.cs
MichalStrehovsky: Yes, good catch. Made it a list.

jkotas (approved):
Thank you!

PR #1558, opened 2021-09-17
Remove custom serialization mechanism from symbolic regex

stephentoub:


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BV64Algebra.cs
```diff
@@ -211,7 +188,17 @@ public string PrettyPrint(ulong bv)
             {
                 var sb = new StringBuilder();
                 sb.Append('[');
```
veanes: The following should work instead for `PrettyPrint`, because now `CharSetSolver` is never `null` and `_partition` is never `null` in the algebra, because the algebra cannot arise via deserialization any more, and this will print the true meaning of the label (not the abstract one based on `ulong` representation of the minterms):
`public string PrettyPrint(ulong bv) => CharSetSolver.PrettyPrint(ConvertToCharSet(CharSetSolver, bv));`

Esssentially, what I'm saying is that the code in `...`: 
`if (_partition == null || bddalgebra == null) {...}`
will be dead-code.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BVAlgebra.cs
```diff
@@ -247,7 +184,23 @@ public string PrettyPrint(BV bv)
             ICharAlgebra<BDD> bddalgebra = SymbolicRegexRunner.s_unicode._solver;
```
veanes, line 184: As in the `ulong` case, the following should work now:
`public string PrettyPrint(BV bv) => CharSetSolver.PrettyPrint(ConvertToCharSet(CharSetSolver, bv));`

veanes (approved):
I left 2 minor comments to simplify pretty-printing in two cases. Otherwise it looks good. This is essentially a rather big change, making it rather high priority to get code gen instead. We made a snapshot of the code before this change, in case we would need to keep the serialization still available for a while in the SRM nuget package for CredScan. Overall it is good to have it gone.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BV64Algebra.cs
stephentoub: Thanks, changed.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BVAlgebra.cs
stephentoub: Thanks, changed.

PR #1564, opened 2021-09-17
Delete unnecessary binplace of libraries .lib files

jkotas:
This was needed to make System.IO.Compression.lib work. It is not needed after the ControlFlowGuard changes


MichalStrehovsky:
Ah. I was wondering where that went. I couldn't find any diffs related to that in /libraries or /installer so I thought we now binplace these in the runtime repo too. Didn't occur to me to search in /eng. Joys of a monorepo.

MichalStrehovsky (approved):
Thank you!

PR #1565, opened 2021-09-17
Fix GC polls around Thread.Spinwait

jkotas:
Avoid GC starvation when Thread.Spinwait is called in a tight loop


src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/Augments/RuntimeAugments.cs
```diff
@@ -1061,11 +1061,6 @@ public static void RhHandleFree(IntPtr handle)
             RuntimeImports.RhHandleFree(handle);
         }
 
-        public static IntPtr RhGetOSModuleForMrt()
```
jkotas, line 1064: Unrelated cleanup

MichalStrehovsky (approved):
Thank you!

PR #1566, opened 2021-09-17
Fix bunch of CS8600

kant2002:
I left only these warnings, which require some thinking. Will be chipping piece by piece these warnings.


jkotas (approved):
Thank you!

PR #1568, opened 2021-09-17
Delete NativeFormat-based type system

MichalStrehovsky:
I was lead to moving the entire directory but only a single file was included from here.

The reason why we included it is comparing public keys/tokens. I don't think we even need that.


jkotas (approved):
Thanks

PR #1569, opened 2021-09-17
Make GetFixedPrefix in SymbolicRegexNode iterative

olsaarik:
This removes the reliance on the max bound of the length of the prefix for avoiding stack overflows. The new algorithm is also likely more efficient since it avoids materializing the long prefix of one path only to truncate it to the short prefix of another path.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
@@ -14,11 +14,6 @@ namespace System.Text.RegularExpressions.Symbolic
     {
         internal const string EmptyCharClass = "[]";
 
-        // Limit the maximum prefix length in the NonBacktracking case to 1000
-        // TODO: alternative is to rewrite SymbolicRegexNode.GetPrefixSequence
-        // to avoid deep recursion
-        internal const int MaxPrefixLength = 1000; // RegexBoyerMoore.MaxLimit;
-
         internal readonly SymbolicRegexBuilder<S> _builder;
         internal readonly SymbolicRegexKind _kind;
         internal readonly int _lower;
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
@@ -937,7 +937,7 @@ namespace System.Text.RegularExpressions.Symbolic
                     return;
 
                 case SymbolicRegexKind.NWBAnchor:
-                   sb.Append("\\B");
+                    sb.Append("\\B");
                     return;
 
                 case SymbolicRegexKind.EndAnchorZ:
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
@@ -1341,16 +1341,27 @@ namespace System.Text.RegularExpressions.Symbolic
                         return true;
                     case SymbolicRegexKind.Concat:
                         Debug.Assert(node._left is not null && node._right is not null);
-                        if (node._left._kind == SymbolicRegexKind.Singleton)
+                        if (!node._left.CanBeNullable)
                         {
-                            nextPaths.Add(node._right);
-                            Debug.Assert(node._left._set is not null);
-                            set = node._left._set;
-                            return true;
+                            if (node._left.GetFixedLength() == 1)
+                            {
+                                set = node._left.GetStartSet();
+                                // Left side had just one character, can use just right side as path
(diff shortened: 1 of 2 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
```diff
@@ -1395,83 +1390,128 @@ internal bool StartsWithLoop(int upperBoundLowestValue = 1)
...
+                    case SymbolicRegexKind.Concat:
+                        Debug.Assert(node._left is not null && node._right is not null);
+                        if (node._left._kind == SymbolicRegexKind.Singleton)
                         {
-                            ImmutableList<S> p = enumerator.Current.GetPrefixSequence(ImmutableList<S>.Empty, lengthBound);
-                            int prefix_length = altsPrefix.TakeWhile((x, i) => i < p.Count && x.Equals(p[i])).Count();
-                            altsPrefix = altsPrefix.RemoveRange(prefix_length, altsPrefix.Count - prefix_length);
+                            nextPaths.Add(node._right);
+                            Debug.Assert(node._left._set is not null);
+                            set = node._left._set;
```
veanes: Here I thought one could use more generally node._startSet, to cover some cases when 
the concat does not start with a singleton (but e.g. (a|A) rather than [aA]) but I'm not exactly sure if that matters, it might already be optimized away by the constructor (or, rather, (a|A)  **should** be optimized to [aA] at construction time).
Unfortunately, the mimimum/maximum length information is not part of info that you could just check that 
the length of _left here is 1 and then use the _startSet. This is not important though.

veanes (approved):
I looked through. Seems good to me. Some cases are not covered, such as 
`a(bB|bC)d` would give prefix `"a"` rather than prefix `"ab"`, but I think this is fine. 
This way the logic of the algorithm remains cleaner to not consider non-singleton _left in concats. Perhaps a related update would be to make sure that `MkOr(x,y)` creates `[xy]` instead of `(x|y)` when both `x` and `y` are singletons (predicates) in the case of say `[aA](b|B)[cC]` to get the prefix `"abc"` with `IgnoreCase=true` instead of the prefix `"a"`

A related thing is to add a couple of unit tests that make sure to exercise the corner cases of the new code.

veanes:
A related addition could be to add the following line to `StressTestDeepNestingOfConcat_TestData`:
 `yield return new object[] { "[aA]", "", options, "aAaAa", 1000, 200 };`
It should stress-test the new code to make sure there are no hidden issues with the yields. The prior cases in `StressTestDeepNestingOfConcat_TestData` should exit immediately with prefix "" but this one should create a prefix of 1000 'a's with IgnoreCase = true. (or even try 10000 instead of 1000, but I would just leave 1000 in the test itself with 10000 the line would need to be:
`yield return new object[] { "[aA]", "", options, "aAaAa", 10000, 2000 };`

veanes:
I'm updating MkOr(...) to join separate cases like (a|A) into [aA]. Just saying it here so we don't both duplicate this work as I mentioned it above also.

stephentoub:
> Some cases are not covered, such as
a(bB|bC)d would give prefix "a" rather than prefix "ab"

This should already be handled by the parser/optimizer:
https://github.com/dotnet/runtime/blob/a842e7a1dc6c241c928c6291411393cdb2516608/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexNode.cs#L880
Is that not happening?

veanes:
I think that some of the optimizations (`ExtractCommonPrefix()` being one such) are not being used _after_ the conversion from `RegexNode` to `SymbolicRegexNode` that has had its own prefix algorithm. But to be honest, I'm not 100% sure, some small unit tests to check this out and step through the code in debugging mode here would be good though to make sure.

veanes:
I mean ideally a regex such as `(a|A)(b(c|C)|Bc|BC)` should also give the prefix `"abc"` with `IgnoreCase=true`, not sure of this case is really useful, but there are practical examples of the form `a(bcd|bcefg)e` that the current impl I think would miss out on (even in out unit tests) (but that used to work), because in handling the concatenation `(bcd|bcefg)e` the left side is not a singleton.

veanes:
I've also discussed another more general way of handling prefix computation with Olli. I would say the code is a bit unfinished at this point.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
olsaarik: Oh right, using _startSet is a great idea. With the current algorithm the prefix does need to end there though, since for patterns like (aa|AA)b the prefix needs to be aab (+ignore case). The current algorithm can only track a path that only descends further into the syntax tree. This could be improved by, instead of having a single SymbolicRegexNode for each path, having a stack of nodes where the top one is followed until it ends and the path ends when the stack empties.

olsaarik:
I found a way to address patterns like `a(bcd|bcefg)e` without using either a stack for tracking paths or going to the transition regexes solution (which we discussed with Margus). Basically, when handling `(bcd|bcefg)e` it is safe (though not complete) to just consider the left hand side and forget about the `e`, then the existing handling for disjunctions produces the desired prefix.

PR #1570, opened 2021-09-18
added special treatment of singletons in SymbolicRegexSet

veanes:
Added special case for representing unions of singletons so that for example (a|b) is represented internally in the same way as [ab].

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexSet.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexSet.cs
@@ -24,6 +24,9 @@ namespace System.Text.RegularExpressions.Symbolic
         /// </remarks>
         private readonly Dictionary<(SymbolicRegexNode<S>, SymbolicRegexNode<S>, bool), int> _loops;
 
+        /// <summary> the union (or intersection) of all singletons in the collection if any or null if none</summary>
+        private readonly SymbolicRegexNode<S>? _singleton;
+
         internal readonly SymbolicRegexKind _kind;
 
         private int _hashCode;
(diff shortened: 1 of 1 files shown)


stephentoub:
> Added special case for representing unions of singletons so that for example (a|b) is represented internally in the same way as [ab].

The parser/optimizer should already transform (a|b) to [ab]. Is it not doing so?
https://github.com/dotnet/runtime/blob/a842e7a1dc6c241c928c6291411393cdb2516608/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexNode.cs#L760

veanes:
So In the case of a regex such as `(a|ab|c|d|dfd)` there is no internal conversion in a `RegexNode` of creating `([acd]|ab|dfd)` because this would change the order of the nodes and thus affect _match generation semantics_. Well not in the trivial case of `(a|b)` being `[ab]` but that is too simple example. In a `SymbolicRegexNode` converting `(a|ab|c|d|dfd)` to `([acd]|ab|dfd)` is a valid transformation because there alternation-order does not matter.

veanes:
In other words. In `NonBacktracking` mode `|` is both associative and _commutative_ while otherwise it is not commutative. Commutativity essentially enables that internal optimization. I should have made more clear that this is what this update is about (not just the case of a union of singletons, but about the internal representation of `SymbolicRegexSet` that implements this extra optimization when singletons are present)

olsaarik:
It seems like this change introduces a second representation for singletons, so there will be the plain singleton and the singleton wrapped inside SymbolicRegexSet. How necessary is this? I'm just thinking that when writing down theory we'd just say (a|A) -> [aA] and be fine with simplifying the intersection into a singleton. Would it be possible to do that here too, maybe in the builder function for sets? I'm guessing the problem might be other code that assumes the builder function for sets actually returns a set instance.

veanes:
Currently the problem was that `HashSet<SymbolicRegexNode<S>> _set` will otherwise contain 
`a` and `A` separately as two separate` SymbolicRegexNode` singletons when the set is created for example with arguments `a` and `B` separately, say after a `b`-derivative of `(bA|ba)`.
In other words, `SymbolicRegexSet` already has the dual representation problem anyways, this was an attempt to normalize/canonicalize the internal representation. Else for example, a `SymbolicRegexSet` containing the singletons `[aA]` and `b` separately would be different from a `SymbolicRegexSet` containing say `[ab]` and `A` separately.

veanes:
Alternative is to get rid of `SymbolicRegexSet ` and to instead reuse `_left` and `_right` and to represent union (and intersection) as **binary** operators. Then we would need to order those (say by `_hashcode`) and keep those say in right-associative form. That might actually be a better alternative that to use HashSets and Dictionaries inside. But that is a bigger change.

olsaarik:
> Currently the problem was that `HashSet<SymbolicRegexNode<S>> _set` will otherwise contain
> `a` and `A` separately as two separate` SymbolicRegexNode` singletons when the set is created for example with arguments `a` and `B` separately, say after a `b`-derivative of `(bA|ba)`.
> In other words, `SymbolicRegexSet` already has the dual representation problem anyways, this was an attempt to normalize/canonicalize the internal representation. Else for example, a `SymbolicRegexSet` containing the singletons `[aA]` and `b` separately would be different from a `SymbolicRegexSet` containing say `[ab]` and `A` separately.

Ok yeah. Do I understand correctly that with this change also something like A|[ab]|hello would get simplified to [aAb]|hello, i.e., the singleton part can coexist with non-singleton children? So that makes this not just a redundant representation for singletons.

If we later need to, we can introduce a simplification rule that unwraps sets that just contain the singleton part. That might for some regexes result in modest reduction of the state space size.

veanes:
That is correct. The current wrapping is essentially needed because of the expected types.

olsaarik (approved):

PR #1572, opened 2021-09-18
Another bunch of nullability fixes

kant2002:
Fix all 
- CS8601
- CS8605
and some more other issues.


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Exception.CoreRT.cs
```diff
@@ -166,14 +166,14 @@ private static void AppendExceptionStackFrame(object exceptionObj, IntPtr IP, in
...
-            IntPtr[] stackTrace = _corDbgStackTrace;
+            IntPtr[]? stackTrace = _corDbgStackTrace;
             if (stackTrace != null)
             {
                 IntPtr[] newStackTrace = new IntPtr[stackTrace.Length];
                 Array.Copy(stackTrace, 0, newStackTrace, 0, stackTrace.Length);
                 stackTrace = newStackTrace;
             }
-            return new DispatchState(stackTrace);
+            return new DispatchState(stackTrace!);
```
jkotas:
```suggestion
            return new DispatchState(stackTrace);
```
?

kant2002: Seems to be I made mistake. Will change declaration of  `DispatchState` constructor.

src/coreclr/nativeaot/Common/src/System/Collections/Concurrent/ConcurrentUnifierW.cs
```diff
@@ -112,7 +112,7 @@ public V GetOrAdd(K key)
                     // and only one value may exist per key.
                     Debug.Assert(checkedFound);
                     Debug.Assert(object.ReferenceEquals(checkedValue, value));
-                    GC.KeepAlive(value);
+                    GC.KeepAlive(value!);
```
jkotas:
```suggestion
                    GC.KeepAlive(value);
```
KeepAlive takes `object?`. `!` should not be needed here.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Array.CoreRT.cs
```diff
@@ -581,7 +581,7 @@ private static unsafe void CopyImplValueTypeArrayWithInnerGcRefs(Array sourceArr
 
                     object boxedValue = RuntimeImports.RhBox(sourceElementEEType, ref *pSourceElement);
                     if (reliable)
-                        boxedElements[i] = boxedValue;
+                        boxedElements![i] = boxedValue;
```
jkotas: We can avoid `!` here by changing the condition to `if (boxedElements != null)`

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Array.CoreRT.cs
```diff
@@ -601,7 +601,7 @@ private static unsafe void CopyImplValueTypeArrayWithInnerGcRefs(Array sourceArr
                     byte* pDestinationElement = pDstArray + (nuint)destinationIndex * cbElementSize;
                     for (int i = 0; i < length; i++)
                     {
-                        RuntimeImports.RhUnbox(boxedElements[i], ref *pDestinationElement, sourceElementEEType);
+                        RuntimeImports.RhUnbox(boxedElements![i], ref *pDestinationElement, sourceElementEEType);
```
jkotas: Same here

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Array.CoreRT.cs
kant2002: Great!

src/coreclr/nativeaot/System.Private.CoreLib/src/System/InvokeUtils.cs
```diff
@@ -570,7 +570,7 @@ public static object DynamicInvokeParamHelperCore(ref ArgSetupState argSetupStat
             if (index >= parametersLength)
                 throw new System.Reflection.TargetParameterCountException();
 
-            object incomingParam = argSetupState.parameters[index];
+            object? incomingParam = argSetupState.parameters![index];
```
jkotas:
```suggestion
            Debug.Assert(argSetupState.parameters != null);

            object? incomingParam = argSetupState.parameters[index];
```
and also remove `!` from the rest of the `parameters` occurrences below? (I am not 100% sure whether it is going to work.)

kant2002: Will try. That's worst change which I have to do to make it work.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/ThreadPool.Windows.cs
```diff
@@ -180,8 +180,9 @@ private void FinishUnregistering()
             GC.SuppressFinalize(this);
         }
 
-        private void FinishUnregisteringAsync(object waitObject)
+        private void FinishUnregisteringAsync(object? waitObject)
         {
+            Debug.Assert(waitObject != null);
```
jkotas: This is likely causing the CI failures. The code assumes that `safeWaitHandle` can be null below.

jkotas (approved):
Thank you!

PR #1576, opened 2021-09-19
Log progress when writing object file with `--verbose`

MichalStrehovsky:


jkotas:
Anything particular that made you to add this?

jkotas (approved):

MichalStrehovsky:
I couldn't repro the hang from the internal email thread but object writing takes about as much as the compilation phase for those inputs (loooong) and probably appeared as a hang. I saw the customer had verbose turned on. I'm yet to write a response to the email.

But the short version is the app is huge.

PR #1577, opened 2021-09-20
Remove FixedArgument from metadata schema

MichalStrehovsky:
SchemaDef.cs is the meat of this change, most of the rest is just generated code and a small adjustment on the reader and writer side.

* `FixedArgumentAttributes` (`FixedArgument.Flags`) was unused
* The type of a fixed argument is redundant with the constructor signature (and is therefore not captured in the ECMA format). We shouldn't duplicate it.
* `NamedArgument` having a `FixedArgument` field was an odd shortcut.


jkotas (approved):
LGTM

PR #1582, opened 2021-09-20
Simplify integration of NonBacktracking into Regex

stephentoub:
Initially NonBacktracking was tacked on with its own processing loop and special-cased directly from Regex with code that detected the option and called into special functions.  I previously changed that to instead derive from RegexRunner, and added new virtual methods for it to override.  But such integration doesn't go far enough in terms of being able to plug into the currently public entry points, which will be required when we implement source generator support for NonBacktracking.

This PR fixes that up by:
- Undoing the previously added internal virtuals
- Changing SymbolicRegexRunner to instead override FindFirstChar and Go
- Deleting the custom Scan overrides
- Creating matches using the exposed Capture method rather than doing so manually

Doing this, and in particular changing it to use Capture, meant that the caps information in the Regex had to be correct, so it also:
- Fixes the Init method to properly populate these members based on the expected NonBacktracking semantics (i.e. no subcaptures)

That then required being able to pass the originally computed capture size from the pattern to Regex.Replace operations, as it uses that number to interpret whether constructs are backreferences or just text, which meant I had to store that value separately on the Regex.  And as I never like increasing the size of the object, I looked for things I could offset it with:
- Deleted the _refsInitialized member.  It was only used to record whether _replref had been initialized, which is obvious based on whether the field is non-null
- Made _replref itself initialized lazily, so that regex instances don't pay for that WeakReference object allocation unless Replace is actually being used
- That then made the old InitializeReferences protected method useless, so I made it a nop and removed the associated references

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexRunner.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexRunner.cs
@@ -1,7 +1,6 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Text.RegularExpressions.Symbolic.Unicode;
@@ -12,14 +11,18 @@ namespace System.Text.RegularExpressions.Symbolic
     {
         /// <summary>The unicode component, including the BDD algebra.</summary>
         internal static readonly UnicodeCategoryTheory<BDD> s_unicode = new UnicodeCategoryTheory<BDD>(new CharSetSolver());
-
+        /// <summary>The matching engine.</summary>
         internal readonly SymbolicRegexMatcher _matcher;
+        /// <summary>Minimum length computed</summary>
+        private readonly int _minRequiredLength;
 
-        private SymbolicRegexRunner(RegexNode rootNode, RegexOptions options, TimeSpan matchTimeout, CultureInfo culture)
+        private SymbolicRegexRunner(RegexCode code, RegexOptions options, TimeSpan matchTimeout, CultureInfo culture)
         {
             var converter = new RegexNodeToSymbolicConverter(s_unicode, culture);
             var solver = (CharSetSolver)s_unicode._solver;
-            SymbolicRegexNode<BDD> root = converter.Convert(rootNode, topLevel: true);
+            SymbolicRegexNode<BDD> root = converter.Convert(code.Tree.Root, topLevel: true);
+
+            _minRequiredLength = code.Tree.MinRequiredLength;
 
             BDD[] minterms = root.ComputeMinterms();
             if (minterms.Length > 64)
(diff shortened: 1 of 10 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexRunner.cs
```diff
@@ -65,85 +68,42 @@ public static RegexRunnerFactory CreateFactory(RegexNode rootNode, RegexOptions
...
+            // TODO: In order to produce a SymbolicMatch with a valid Index/Length, FindMatch needs to do additional
+            // work once it's found the end of a match to find the beginning of the match.  If this is an IsMatch
+            // call, however, the details of the match will be ignored, with the only important information being
+            // that there was a match.  We need a mechanism of supplying that information to this operation. At the
+            // moment the only obvious answer to that is new protected surface area, either additional protected
+            // field/property that could be consulted, or a new virtual Scan method that this implementation could
+            // override instead of using the base Scan method that invokes FindFirstChar/Go.  We could choose to bypass
+            // this public/protected surface area for the in-memory implementation, e.g. having it look at some
+            // new private protected field we add to RegexRunner, but for now we simply don't benefit from this optimization.
+            const bool calledFromIsMatch = false;
```
veanes, line 90: There are some cases where the perf difference can be substantial. 
Example regex: `.{20}a[a-z]*$` , `input = "01234567890123456789asomelongrandomtext...goingonandon"`
The forward pass will be very easy to test that a match exists (a linear pass). But finding where the match starts will involve using the reverse of this regex that is  `^[a-z]*a.{20}` (walking the input in reverse) that is the classical regex causing bowup in the reverse DFA (that may then switch to Antimirov mode) in order to find the start of the match, that in the end is actually irrelevant.

veanes (approved):
Not sure I fully understand the API logic, but seems correct to me and is clearly a deeper integration than before. The only concern I have, I also added a note and example in the note, is defining `IsMatch` via `Match`. My main concern is use of the reverse regex in backward search to find the start position of the match can make a big difference between performance of `IsMatch` vs the full `Match`. But I believe we are all aware of that and that this is a temporary solution.

stephentoub:
> Not sure I fully understand the API logic

Basically this just moves the logic into the existing Go method that's called from the RegexRunner.Scan loop, e.g. along the lines of:
```C#
while (FindFirstChar())
{
    ...
    Go();
    ...
}
```

> My main concern is use of the reverse regex in backward search to find the start position of the match can make a big difference between performance of IsMatch vs the full Match

Yeah, I'd like to find a solution for that, too.  If this were just about the in-memory implementation, we could add an internal field to RegexRunner and be done with it. But we also need to support the source generator, which wouldn't have access to such an internal.  I expect we're going to need to add new surface area to do this well, and I want us to not rush that.  One thing I want to explore more is if we could make all the engines better by adding a new virtual Scan method to replace the separate FindFirstChar/Go methods... there's no great reason they need to be separate, other than some initialization code that I expect we can find others way to make work.  And we could then pass such information into Scan as long as we're exposing a new API.

PR #1584, opened 2021-09-21
Make BDDAlgebra lock reads too

olsaarik:
Switches the locking in BDDAlgebra to use `ReaderWriterLockSlim`. The operation caches were also unified into a single `_opCache`, which made the locking code slightly cleaner and I don't think should affect performance much in either direction.

Also removed an overload of And in IBooleanAlgebra that wasn't used anywhere.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDD.cs
@@ -343,51 +343,43 @@ namespace System.Text.RegularExpressions.Symbolic
                 return arcs[0] == 0 ? False : True;
             }
 
-            algebra.Lock.EnterWriteLock();
-            try
+            // the number of bits used for ordinals and node identifiers are stored in the first two values
+            int k = arcs.Length;
+            int ordinal_bits = (int)arcs[0];
+            int node_bits = (int)arcs[1];
+
+            // create bit masks for the sizes of ordinals and node identifiers
+            long ordinal_mask = (1 << ordinal_bits) - 1;
+            long node_mask = (1 << node_bits) - 1;
+            BitLayout(ordinal_bits, node_bits, out int zero_node_shift, out int one_node_shift, out int ordinal_shift);
+
+            // store BDD nodes by their id when they are created
+            BDD[] nodes = new BDD[k];
+            nodes[0] = False;
+            nodes[1] = True;
+
+            for (int i = 2; i < k; i++)
             {
-                // the number of bits used for ordinals and node identifiers are stored in the first two values
-                int k = arcs.Length;
-                int ordinal_bits = (int)arcs[0];
-                int node_bits = (int)arcs[1];
-
-                // create bit masks for the sizes of ordinals and node identifiers
-                long ordinal_mask = (1 << ordinal_bits) - 1;
-                long node_mask = (1 << node_bits) - 1;
-                BitLayout(ordinal_bits, node_bits, out int zero_node_shift, out int one_node_shift, out int ordinal_shift);
-
-                // store BDD nodes by their id when they are created
-                BDD[] nodes = new BDD[k];
-                nodes[0] = False;
-                nodes[1] = True;
-
-                for (int i = 2; i < k; i++)
+                long arc = arcs[i];
+                if (arc <= 0)
                 {
-                    long arc = arcs[i];
-                    if (arc <= 0)
-                    {
-                        // this is an MTBDD leaf. Its ordinal was serialized negated
-                        nodes[i] = algebra.GetOrCreateBDD((int)-arc, null, null);
-                    }
-                    else
-                    {
-                        // reconstruct the ordinal and child identifiers for a non-terminal
-                        int ord = (int)((arc >> ordinal_shift) & ordinal_mask);
-                        int oneId = (int)((arc >> one_node_shift) & node_mask);
-                        int zeroId = (int)((arc >> zero_node_shift) & node_mask);
-
-                        // the BDD nodes for the children are guaranteed to exist already due to the topological order
-                        nodes[i] = algebra.GetOrCreateBDD(ord, nodes[oneId], nodes[zeroId]);
-                    }
+                    // this is an MTBDD leaf. Its ordinal was serialized negated
+                    nodes[i] = algebra.GetOrCreateBDD((int)-arc, null, null);
                 }
+                else
+                {
+                    // reconstruct the ordinal and child identifiers for a non-terminal
+                    int ord = (int)((arc >> ordinal_shift) & ordinal_mask);
+                    int oneId = (int)((arc >> one_node_shift) & node_mask);
+                    int zeroId = (int)((arc >> zero_node_shift) & node_mask);
 
-                //the result is the final BDD in the nodes array
-                return nodes[k - 1];
+                    // the BDD nodes for the children are guaranteed to exist already due to the topological order
+                    nodes[i] = algebra.GetOrCreateBDD(ord, nodes[oneId], nodes[zeroId]);
+                }
             }
-            finally
-            {
-                algebra.Lock.ExitWriteLock();
-            }
+
+            //the result is the final BDD in the nodes array
+            return nodes[k - 1];
         }
 
         /// <summary>
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDD.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDD.cs
@@ -336,30 +336,15 @@ namespace System.Text.RegularExpressions.Symbolic
         /// which implies that all BDD nodes other than True and False are new BDD objects
         /// that have not been internalized or cached.
         /// </summary>
-        public static BDD Deserialize(long[] arcs, BDDAlgebra? algebra = null)
+        public static BDD Deserialize(long[] arcs, BDDAlgebra algebra)
         {
             if (arcs.Length == 1)
             {
                 return arcs[0] == 0 ? False : True;
             }
 
-            if (algebra is null)
-            {
-                return DeserializeWithFactory(arcs, (int ordinal, BDD? one, BDD? zero) =>
-                {
-                    Debug.Assert(one != zero || one is null);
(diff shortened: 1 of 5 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDD.cs
```diff
@@ -392,13 +377,17 @@ static BDD DeserializeWithFactory(long[] arcs, Func<int, BDD?, BDD?, BDD> create
...
                     }
                 }
 
                 //the result is the final BDD in the nodes array
                 return nodes[k - 1];
             }
+            finally
+            {
+                algebra.Lock.ExitWriteLock();
+            }
```
stephentoub, line 379: What is it we're protecting access to here with the GetOrCreateBDD call?  Is it just the state here, or are we trying to synchronize access across multiple calls, or something else?

I ask because it seems like the `HashSet<BDD>` could be replaced by a `ConcurrentDictionary<BDD, BDD>`, and GetOrCreateBDD could become:
```C#
var key = new BDD(ordinal, one, zero);
return _bddCache.GetOrAdd(key, key);
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
```diff
@@ -90,58 +92,69 @@ public BDD GetOrCreateBDD(int ordinal, BDD? one, BDD? zero)
...
+                        Lock.ExitWriteLock();
+                    }
+                }
+                return result;
+            }
+            finally
+            {
+                Lock.ExitUpgradeableReadLock();
+            }
+        }
```
stephentoub: Similar question here.  I'm wondering if it might be better for _opCache to be a ConcurrentDictionary; then we wouldn't lock while doing reads, and would just use a normal lock if we had a cache miss, e.g.
```C#
if (!_opCache.TryGetValue(key, out BDD? result)
{
    lock (_opCache)
    {
        if (!_opCache.TryGetValue(key, out result)
        {
            result = CreateBoolOp_lock(key);
        }
    }
}
return result;
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
```diff
@@ -497,30 +507,38 @@ private BDD Shift_rec(Dictionary<ShiftOpKey, BDD> shiftCache, BDD set, int k)
...
-                if (maxBit < 31)
-                {
-                    uint filter = (mask << 1) - 1;
-                    m &= filter;
-                    n &= filter;
-                }
+            uint mask = (uint)1 << maxBit;
 
+            //filter out bits greater than maxBit
+            if (maxBit < 31)
```
veanes, line 315: Unrelated to your changes, but a left-over that I just recalled. There is an implicit assumption that has to do with 31, obviously related to `maxBit` of `uint` but why `<`. Essentially the intervals->BDD algo is missing some intuition (although I wrote it myself 10y back or so, _I forgot_, there are some internal assumptions, e.g. that it works correctly for characters, i.e., ushort, but actually I don't think it works correctly for arbitrary `uint` intervals due to some overflow) . So definitely some `Debug.Assert`'s are missing (something todo later).

veanes (approved):
I left a comment but this is more of a reminder that some intuition is due, that I need to recall.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDD.cs
olsaarik: Thinking about this more I do think ConcurrentDictionary might be the better approach. The main concern is indeed the synchronization of that datastructure, ensuring a unique representation for each BDD. I'll change this.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
olsaarik: Yes, similarly to _bddCache this seems like a good idea. Thinking about this there's actually a further choice here on the granularity of locking. The actual construction of the BDD operations could also be not locked, with all updates to the cache using `GetOrAdd` like:
```
result = _opCache.GetOrAdd(key, key => {
    // Compute the actual operation here
});
```
The computation of the operation does include recursive calls to GetOrAdd, but it seems like recursive calls to GetOrAdd are fine: https://stackoverflow.com/a/62093870 .

A slight worry here is that with many threads you might sometimes have threads racing through the construction of an operation at a similar pace, thus duplicating work, but that's only an efficiency problem. I'm thinking avoiding the lock for writes too might be a win since I'm guessing we are mostly expecting single threaded usage.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
olsaarik: I created a todo on our project board for this.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
olsaarik: I talked with Margus about this too and I think he was convinced it should work, I'll implement the fine grained synchronization.

stephentoub: >  with all updates to the cache using GetOrAdd

If from a correctness perspective it's ok to have the same key being processed concurrently,  I agree this is the way to go.

olsaarik:
With `ConcurrentDictionary` I was able to simplify the whole BDDAlgebra implementation a lot, since the boolean operations can now just recurse directly on themselves.

veanes (approved):
I think overall these are good improvements of readability that I think help better understand how the algorithm interval->BDD algo works.

olsaarik:
There was a failed test run that had a timeout. I retried the run, but even if it goes through its a bit concerning given the changes have to do with synchronization. I'll spend more time double checking the code.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
```diff
@@ -58,170 +54,45 @@ internal abstract class BDDAlgebra : IBooleanAlgebra<BDD>
...
-            if (!_bddCache.TryGetValue(key, out BDD? set))
-            {
-                set = key;
-                _bddCache.Add(set);
-            }
-
-            return set;
-        }
+        public BDD GetOrCreateBDD(int ordinal, BDD? one, BDD? zero) =>
+            _bddCache.GetOrAdd(new BDDKey(ordinal, one, zero), key => new BDD(ordinal, one, zero));
```
stephentoub: This is going to incur closure/delegate allocations even for the fast path where the object already exists.  You can avoid that in a few ways, but the simplest here is probably to just use the overload that let's you pass in additional state, e.g.
```suggestion
        public BDD GetOrCreateBDD(int ordinal, BDD? one, BDD? zero) =>
            _bddCache.GetOrAdd(new BDDKey(ordinal, one, zero), (key, state) => new BDD(state.ordinal, state.one, state.zero), (ordinal, one, zero));
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
```diff
@@ -58,170 +54,45 @@ internal abstract class BDDAlgebra : IBooleanAlgebra<BDD>
...
-                    res = t == f ? t : GetOrCreateBDD(a.Ordinal, t, f);
-                }
-
-                _binOpCache[key] = res;
-                return res;
-            }
-        }
+            _opCache.GetOrAdd(new(BoolOp.Not, a, null), key => a.IsLeaf ?
+                GetOrCreateBDD(CombineTerminals(BoolOp.Not, a.Ordinal, 0), null, null) : // multi-terminal case
+                GetOrCreateBDD(a.Ordinal, Not(a.One), Not(a.Zero)));
```
stephentoub:
```suggestion
            _opCache.GetOrAdd(new(BoolOp.Not, a, null), (key, a) => a.IsLeaf ?
                GetOrCreateBDD(CombineTerminals(BoolOp.Not, a.Ordinal, 0), null, null) : // multi-terminal case
                GetOrCreateBDD(a.Ordinal, Not(a.One), Not(a.Zero)),
                a);
```

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
```diff
@@ -266,62 +137,45 @@ private BDD CreateBinBoolOP_rec(BoolOp op, BDD a, BDD b)
...
-                return res;
-
-            if (a.IsLeaf && b.IsLeaf)
+            // Order operands by hash code to increase cache hits
+            if (a.GetHashCode() > b.GetHashCode())
             {
-                //multi-terminal case, we know here that a is neither True nor False
-                int ord = CombineTerminals(op, a.Ordinal, b.Ordinal);
-                res = GetOrCreateBDD(ord, null, null);
+                var tmp = a;
```
stephentoub:
```suggestion
                BDD tmp = a;
```

stephentoub (approved):
Thanks!

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
olsaarik: Good to know, thanks! I've gotten too used to C++ where these would end up being stack allocations.

PR #1586, opened 2021-09-22
Replace IRuntimeImplemented with RuntimeType and RuntimeAssembly basetype

jkotas:
To match how reflection is structured in CoreCLR and Mono. It should be also a step towards eliminating System.Private.DisabledReflection.


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Enum.CoreRT.cs
```diff
@@ -186,14 +171,11 @@ private static bool TryGetUnboxedValueOfEnumOrInteger(object value, out ulong re
             }
         }
 
-        public static string? GetName<TEnum>(TEnum value) where TEnum : struct, Enum
```
jkotas, line 189: I have just done minimal work on sharing, just enough to fix build breaks. There is more that can be shared in the Enum implementation after this change.

MichalStrehovsky (approved):

PR #1589, opened 2021-09-23
Add guards against stack overflows in various NonBackTracking algorithms

olsaarik:
I implemented the approach that was discussed that guards against stack overflows by calling the function on another thread when the stack is too deep. I found an implementation of it in System.Linq.Expressions which I adapted: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/StackGuard.cs

I also updated the stress tests we had for deep nestings of concatenations. The constants were not large enough to actually trigger the overflows. I also added an analogous test for nesting a couple kinds of loops. 

There were several algorithms that were encountering stack overflows after I updated the tests, which I added the guards to.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
+using System.Runtime.CompilerServices;
 
 namespace System.Text.RegularExpressions.Symbolic
 {
@@ -200,8 +201,12 @@ namespace System.Text.RegularExpressions.Symbolic
         public SymbolicRegexNode<BDD> Convert(RegexNode node, bool topLevel)
         {
             // Guard against stack overflow due to deep recursion
-            if (StackHelper.CallOnEmptyStackIfNecessary(() => Convert(node, topLevel), out SymbolicRegexNode<BDD>? result))
-                return result!;
+            if (!RuntimeHelpers.TryEnsureSufficientExecutionStack())
+            {
+                RegexNode localNode = node;
+                bool localTopLevel = topLevel;
+                return StackHelper.CallOnEmptyStack(() => Convert(localNode, localTopLevel));
+            }
 
             switch (node.Type)
             {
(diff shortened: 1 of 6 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
```diff
@@ -197,9 +197,12 @@ code switch
             }
         }
 
-        // TODO https://github.com/dotnet/runtimelab/issues/1537: Avoid deep recursion
         public SymbolicRegexNode<BDD> Convert(RegexNode node, bool topLevel)
         {
+            // Guard against stack overflow due to deep recursion
+            if (StackHelper.CallOnEmptyStackIfNecessary(() => Convert(node, topLevel), out SymbolicRegexNode<BDD>? result))
```
stephentoub: Due to the way Roslyn works, the fact that the lambda closes node and topLevel will cause the closure to be allocated at the beginning of the method, even if we don't follow this code path.  I think we should change the pattern to be something like:
```C#
if (!RuntimeHelpers.TryEnsureSufficientExecutionStack())
{
    RegexNode localNode = node;
    bool localTopLevel = topLevel;
    return CallOnEmptyStack(() => Convert(localNode, localTopLevel));
}
```
That way, everything is guarded behind the if, and when we do capture, it's locals in the guarded scope, which means the closure will only be allocated in that rare case rather than the common case.

veanes (approved):
Regarding unit tests, edits look good to me. The new bounds of 2000 seem reasonable. I tried earlier with 10000 but that made the tests too slow, so 2000 seems like a good compromise.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
olsaarik: Thanks! I'll change the pattern.

olsaarik: Oh, does capturing locals rather than parameters matter?

olsaarik: Found my answer: https://stackoverflow.com/questions/65526964/avoid-closure-on-fast-path-by-storing-state-into-scoped-locals

stephentoub: > Oh, does capturing locals rather than parameters matter?

It's the scope of where they're defined that matters.  No matter where they're used, a closure that captures them is defined at the same level they are.

stephentoub (approved):
Thanks!

PR #1590, opened 2021-09-23
Switch to shared System.Enum implementation

jkotas:


src/libraries/System.Private.CoreLib/src/System/Enum.cs
```diff
@@ -1413,6 +1410,7 @@ private static RuntimeType ValidateRuntimeType(Type enumType)
                 throw new ArgumentException(SR.Arg_MustBeType, nameof(enumType));
 #if CORERT
             // Check for the unfortunate "typeof(Outer<>).InnerEnum" corner case.
```
jkotas:
```suggestion
            // Check for the unfortunate "typeof(Outer<>.InnerEnum)" corner case.
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System/RuntimeType.cs
```diff
@@ -2,11 +2,112 @@
...
 
 using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
 
 namespace System
 {
     // Base class for runtime implemented Type
     public abstract class RuntimeType : TypeInfo
     {
+        public override string GetEnumName(object value)
```
MichalStrehovsky: Make all of these sealed overrides? (RyuJIT will be able to devirtualize if this comes from a typeof or GetType.)

MichalStrehovsky (approved):
Thank you!

PR #1593, opened 2021-09-23
Transition regex support

veanes:
- Added `TransitionRegex `as a separate abstraction of a branching structure of symbolic derivatives.
- Added `MkDerivative()` that computes the `TransitionRegex` from a symbolic regex.
- Extended `DGML` support to debug the new additions, and updated some unit tests `ViewSampleRegexInDGML()` in particular good to look (change `RegexExperiment.s_tmpWorkingDir` locally to some local tmp folder and then open the dgml files created there to look at the output) -- it now calls repeatedly into `MkDerivative()` to explore the whole regex and creates an **NFA** from it (not **DFA**) i.e., possibly nondeterministic.

The size of the NFA can be used to determine if the regex is **secure** - say has less than 10k AST nodes. The size (nr of states) of the NFA is roughly proportional to the number of nodes of the AST of the regex when counters are unwound. Currently the full NFA creation is part of the dgml creation code, but this part is rather straightforward DFS search. 

Can also support **lookarounds** going forward. 
Using MkDerivative() to explore the whole regex is in a nutshell the core idea for the code-generator. As said, we have done this before with Olli in this style. The somewhat new thing here is the branching structure and actually that **minterms** are not needed, but also this we are familiar with from earlier works with Olli.

The new code is really only reachable right now via dgml creation, but as I said, its real purpose is to be used in code-gen and possibly as an alternative backend to the current `DfaMatchingState.Next()` method. The previous character kind is no longer stored or needed there so state becomes just the SymbolicrRegexNode and there is a special lookaround conditional that currently only involves anchors (but could be other stuff -- lookahed/lookbehind in the future). So this is pretty important stuff from that point of view.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Dgml/RegexDFA.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Dgml/RegexDFA.cs
@@ -10,16 +10,20 @@ namespace System.Text.RegularExpressions.Symbolic.DGML
     /// <summary>
     /// Used by DgmlWriter to unwind a regex into a DFA up to a bound that limits the number of states
     /// </summary>
-    internal sealed class RegexDFA<T> : IAutomaton<T> where T : notnull
+    internal sealed class RegexDFA<T> : IAutomaton<(SymbolicRegexNode<T>?, T)> where T : notnull
     {
         private readonly DfaMatchingState<T> _q0;
         private readonly List<int> _states = new();
         private readonly HashSet<int> _stateSet = new();
-        private readonly List<Move<T>> _moves = new();
+        private readonly List<Move<(SymbolicRegexNode<T>?, T)>> _moves = new();
         private readonly SymbolicRegexBuilder<T> _builder;
+        private readonly List<SymbolicRegexNode<T>> _nfaStates = new();
+        private readonly Dictionary<SymbolicRegexNode<T>, int> _nfaStateId = new();
+        private readonly bool _asNFA;
 
-        internal RegexDFA(SymbolicRegexMatcher<T> srm, int bound, bool addDotStar, bool inReverse)
+        internal RegexDFA(SymbolicRegexMatcher<T> srm, int bound, bool addDotStar, bool inReverse, bool asNFA)
         {
+            _asNFA = asNFA;
             _builder = srm._builder;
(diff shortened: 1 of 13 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Dgml/RegexDFA.cs
```diff
@@ -10,84 +10,128 @@ namespace System.Text.RegularExpressions.Symbolic.DGML
     /// <summary>
     /// Used by DgmlWriter to unwind a regex into a DFA up to a bound that limits the number of states
     /// </summary>
-    internal sealed class RegexDFA<T> : IAutomaton<T> where T : notnull
+    internal sealed class RegexDFA<T> : IAutomaton<(SymbolicRegexNode<T>?, T)> where T : notnull
```
olsaarik: The name of this class should change since it's both DFAs and NFAs now.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
```diff
@@ -808,6 +825,161 @@ internal SymbolicRegexNode<S> MkDerivative(S elem, uint context)
...
         }
 
+        private TransitionRegex<S>? _transitionRegex;
+        /// <summary>Computes the symbolic derivative as a transition regex</summary>
+        internal TransitionRegex<S> MkDerivative()
+        {
+            if (_transitionRegex is not null)
+            {
+                return _transitionRegex;
+
```
olsaarik:
```suggestion
```

olsaarik (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Dgml/RegexDFA.cs
veanes: I'll rename it to RegexAutomatonForDGML

veanes: Or maybe just RegexAutomaton, as it is already in  the Dgml namespace.

PR #1595, opened 2021-09-24
Some cleanup/commenting in SymbolicRegexMatcher

stephentoub:
Didn't get all the way through it, so there's more commenting/renaming to be done.  I could use some help understanding the meaning here of words like "partition" and "atom", so if you've got a few minutes @olsaarik to do some renaming and commenting, that would be helpful.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/CharKind.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/CharKind.cs
@@ -5,6 +5,12 @@ namespace System.Text.RegularExpressions.Symbolic
 {
     internal static class CharKind
     {
+        /// <summary>Number of kinds of chars.</summary>
+        internal const int CharKindCount = 5;
+
+        /// <summary>All characters other than those in the four other kinds.</summary>
+        internal const uint General = 0;
+
         /// <summary>Start or Stop of input (bit 0 is 1)</summary>
         internal const uint StartStop = 1;
 
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexMatcher.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexMatcher.cs
@@ -6,17 +6,18 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Runtime.CompilerServices;
+using System.Threading;
 
 namespace System.Text.RegularExpressions.Symbolic
 {
-    /// <summary>Provides IsMatch and Matches methods.</summary>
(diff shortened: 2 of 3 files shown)


veanes (approved):

veanes:
In general we use _partition_ to mean a complete division of a full (character) set into pairwise disjoint nonempty sets (https://en.wikipedia.org/wiki/Partition_of_a_set). A part of a _partition_ is sometimes called an _atom_ (or _minterm_ , _atom_ is a bit arbitrary it reflects that it is sort of an unbreakable unit, we should perhaps be sticking to _minterm_ that reflects how the partition arises). Minterms are like _nonempty regions of a Venn diagram_. The partition originally arises as the _minterms over the predicates that occur in the regex_. For example if the regex is `\d\w+` then the partition is {`\d`, `\W`, `[\w-[\d]]`} we order minterms based on the smallest element in them, in this case the order is `\W`, `\d`, `[\w-[\d]]`. This order is then used to lift them to this abstract representation using just 3-bit bit-vectors for efficiency `\W` is  `001`,  `\d` is `010` , and  `[\w-[\d]]`  is `100`. Those bitvectors are then represented as numbers (using `ulong` if the number of bits < 65 which is, practically speaking, always the case).

veanes:
Another good example source for `partition` and `minterms` is the regex `\b\w+\b`. It used to be that this regex resulted in the partition {`\W`,`\w`}. With the edits for `\b` (and `\B`) (that I will make a PR for shortly) it will instead be {`[\W-[\u200C\u200D]]`,`\w`,`[\u200C\u200D]`}

stephentoub:
Thanks, @veanes.  Trying to internalize what you wrote, but I'm missing some pieces.  I think it'd be helpful as part of a readme / write-up about how this all works to walk through a simple example in excruciating detail, and with the assumption that the reader has zero knowledge about not just these terms but also the terms you're using to explain the terms :), e.g. what is a "predicate" in this context.  That would include not just what the partitions are, but how the implementation derives them, how is that abstract representation interpreted / evaluated, etc.  I'm very much looking forward to reading that :) Regardless, thanks for taking the time to define some of the terms.

veanes:
> I think it'd be helpful as part of a readme / write-up about how this all works to walk through a simple example in excruciating detail,

Absolutely, a bit of a context switch, but this is on the todo list as soon as I'm done with the code edits. Regarding the word _predicate_ this is really just another word for _character class_, but a bit more general because it depends on the Boolean algebra being used, in the case of the predominant `BDDAlgebra`, a _predicate_ is a `BDD`, in the case of `BV64Algebra`, a _predicate_ is a `ulong`, etc.  We could also use some different terminology. At some point `BDD` was actually called `CharSet`, but `BDD`s are also useful outside of the `CharSetSolver` that derives from the `BDDAlgebra` (e.g. as compact representations of sets of unsigned integers in other contexts and also as _Multi Terminal BDD_ or _MTBDD_)

Yes, there is a lot of terminology here, some of which may be simplified a bit I guess, and some words may conflict with existing terminology (or conflict with a perceived understanding that may confuse) -- I think is another issue.

PR #1596, opened 2021-09-24
Align trimming settings with IL Linker

MichalStrehovsky:
Aligns the ILCompiler targets with ILLink.

It's all rather complex because:

* One can specify that an assembly is trimmable using a custom attribute. This doesn't mean it will be trimmed in IL Link, it just marks it.
* One can specify a default trimming action (keep everything or trim) on the command line for assemblies not marked as trimmable
* One can specify a trimming action for listed assemblies on the command line
* In ILLink, one can also specify trimming action for assemblies marked IsTrimmable. For ILCompiler, I'm making this assume the action is trim. ILLink can also say to root. I don't see much point.

I tested this with both .NET 5 and .NET 6 SDKs.

There's a difference in behavior - in the past, setting `TrimMode=link` would trim the whole app, but that's not what illinker does. For .NET 5, one has to do the [custom target dance](https://github.com/davidfowl/LinkedAspNetCoreApplication/blob/fd7454f23e7e48a4302fb239f72a914ce4c9dc73/LinkedAspNetCoreApplication/LinkedAspNetCoreApplication.csproj#L14-L21). On .NET 6, one can just specify TrimmerDefaultAction=link.


jkotas (approved):
This looks complex enough for it to be right. Thank you the due diligence to match the trimming behaviors!

PR #1598, opened 2021-09-25
NativeAOT-LLVM: Add SSA form

### LINKED ISSUE #1404: Question: NativeAOT-LLVM: SSA for branching
I was going to submit a PR to implement `JTRUE` but I'm now thinking that it would be better to look at https://github.com/dotnet/runtimelab/pull/1370#pullrequestreview-719979777 first.  Specifically this comment

" Here SSA will be restored (rationalizer deletes PHIs now...),"

For the method https://github.com/dotnet/runtimelab/blob/1bf0fb49967fb36b69877cda7f8ff8c9f238c68f/src/libraries/System.Private.CoreLib/src/System/Globalization/DateTimeFormatInfo.cs#L1854 the IR at the point the LLVM compilation is called ends in

```
N006 (  8, 10) [000004] N--XG--N-U--         t4 = *  NE        int
                                                  /--*  t4     int
N007 ( 10, 12) [000005] ---XG-------              *  JTRUE     void

------------ BB03 [009..011) -> BB05 (always), preds={BB02} succs={BB05}
               [000028] ------------                 IL_OFFSET void   IL offset: 0x9
N002 (  3,  2) [000013] ------------        t13 =    LCL_VAR   ref    V00 this
                                                  /--*  t13    ref    this in rcx
N003 ( 17,  6) [000014] --CXG-------        t14 = *  CALL      int    System.Globalization.DateTimeFormatInfo.InitializeFormatFlags
                                                  /--*  t14    int
N005 ( 21,  9) [000016] DA-XG-------              *  STORE_LCL_VAR int    V03 tmp2
               [000029] ------------                 IL_OFFSET void   IL offset: 0xf
N001 (  0,  0) [000018] ------------                 NOP       void
N001 (  3,  2) [000017] ------------        t17 =    LCL_VAR   int    V03 tmp2
                                                  /--*  t17    int
N003 (  7,  5) [000020] DA----------              *  STORE_LCL_VAR int    V02 tmp1

------------ BB04 [011..017), preds={BB02} succs={BB05}
               [000030] ------------                 IL_OFFSET void   IL offset: 0x11
N001 (  3,  2) [000006] ------------         t6 =    LCL_VAR   ref    V00 this
N002 (  1,  4) [000025] ------------        t25 =    CNS_INT   int    176 field offset Fseq[formatFlags]
                                                  /--*  t6     ref
                                                  +--*  t25    int
N003 (  4,  6) [000026] -------N----        t26 = *  ADD       byref
                                                  /--*  t26    byref
N004 (  6,  8) [000007] ---XG-------         t7 = *  IND       int
                                                  /--*  t7     int
N006 ( 10, 11) [000009] DA-XG-------              *  STORE_LCL_VAR int    V02 tmp1

------------ BB05 [017..018) (return), preds={BB03,BB04} succs={}
N001 (  3,  2) [000011] ------------        t11 =    LCL_VAR   int    V02 tmp1
                                                  /--*  t11    int
N002 (  4,  3) [000012] ------------              *  RETURN    int
```
`V02 tmp1` is set in one of the 2 branches.  If that is compiled to LLVM I have to currently create some storage, and put the result in storage so it can be read from the return branch, `BB05`.  However I'm wondering if SSA and phis where introduced that problem would go away and the return value would be based on a phi instead?

@SingleAccretion

yowl:
closes #1404

This PR adds the SSA form generation to the LLVM compilation in RyuJIT.  This simplifies the LLVM generation. 

The majority of the changes to ssabuilder are to do with STORE instead of ASG and LIR node walking instead of statements.

The call to `fgExtendDbgLifetimes` is removed to prevent the liveness of vars being extended and confusing the phi arg candidate logic.  Maybe this will have an adverse affect for the DWARF debugging experience later.

For the SSA generation there are a lot of `#if`s in `ssabuilder.cpp` (and elsewhere), an alternative is to create a `ssabuilder_[lir|llvm].cpp`.

In the process the following others changes ard included

- Support for JTRUE/EQ/CAST(initial)
- Do not load the ref arguments and call return values to get synchronicity with constant handles for phis.
- Some build infra changes to get -msbuild option working as that was the easiest way I found to produce the vcxproj for the clrjit browser/wasm config.


SingleAccretion:
> For the SSA generation there are a lot of `#ifs` in `ssabuilder.cpp` (and elsewhere), an alternative is to create a `ssabuilder_[lir|llvm].cpp`.

I've looked at the SSA builder changes and while they are unfortunately `#ifdef`y, it seems fine enough to leave them as is. I feel alternatives will run into different problems.

src/coreclr/jit/compiler.h
```diff
@@ -4980,6 +4988,9 @@ class Compiler
 
     bool backendRequiresLocalVarLifetimes()
     {
+#if TARGET_WASM
+        if (this->compRationalIRForm) return true;
```
SingleAccretion: Just `return true` under `TARGET_WASM` (also, should probably be `#ifdef` and not `#if`)?

src/coreclr/jit/compiler.hpp
```diff
@@ -1716,13 +1716,16 @@ inline unsigned Compiler::lvaGrabTempWithImplicitUse(bool shortLifetime DEBUGARG
...
-    {
-        // Note, at least, that there is at least one reference.
-        lvImplicitlyReferenced = 1;
-        return;
-    }
+#if defined(TARGET_WASM)
+    if (!comp->compRationalIRForm) // TODO: introduce bool comp->opts.OptimizationDisabled() && !comp->compRationalIRForm to make clearer?
+#endif
+        // In minopts and debug codegen, we don't maintain normal ref counts.
+        if ((state == RCS_NORMAL) && comp->opts.OptimizationDisabled())
```
SingleAccretion: It think it would be best to introduce `Compiler::PreciseRefCountsRequired()` (or something) and call here (`&& !Compiler::PreciseRefCountsRequred())` instead of `OptimizationDisabled()`). You can then fold the `#ifdef`s into it.

src/coreclr/jit/lclvars.cpp
```diff
@@ -4411,11 +4411,14 @@ void Compiler::lvaMarkLocalVars()
     const bool isRecompute = false;
     lvaComputeRefCounts(isRecompute, setSlotNumbers);
 
-    // If we're not optimizing, we're done.
-    if (opts.OptimizationDisabled())
-    {
-        return;
-    }
+#if defined(TARGET_WASM)
+    if (!this->compRationalIRForm) // always want to run for LLVM/LIR
```
SingleAccretion: Another case for `PreciseRefCountsRequired`.

src/coreclr/jit/lclvars.cpp
```diff
@@ -4432,7 +4435,10 @@ void Compiler::lvaMarkLocalVars()
     }
 
 #if ASSERTION_PROP
-    assert(opts.OptimizationEnabled());
+#if defined(TARGET_WASM)
+    if (!this->compRationalIRForm) // dont assert for LLVM/LIR
+#endif
+        assert(opts.OptimizationEnabled());
```
SingleAccretion: Would be `assert(PreciseRefCountsRequired())`.

src/coreclr/jit/lclvars.cpp
```diff
@@ -4572,7 +4581,9 @@ void Compiler::lvaComputeRefCounts(bool isRecompute, bool setSlotNumbers)
     {
         if (block->IsLIR())
         {
+#if !defined(TARGET_WASM) // TODO: maybe we should be setting isRecompute to true?
```
SingleAccretion: Not necessarily (it would be a "lie" if we didn't run any refcounts before). I think it is fine to leave the `#ifdef` (or, to be honest, it should be fine to delete the assert altogether; it seems low value).
```suggestion
#if !defined(TARGET_WASM)
```

src/coreclr/jit/lclvars.cpp
```diff
@@ -4466,6 +4472,9 @@ void Compiler::lvaComputeRefCounts(bool isRecompute, bool setSlotNumbers)
     unsigned   lclNum = 0;
     LclVarDsc* varDsc = nullptr;
 
+#if defined(TARGET_WASM)
+    if (!compRationalIRForm) // for LIR never do this fast path
+#endif
```
SingleAccretion: Another case for `PreciseRefCountsRequired`.

src/coreclr/jit/liveness.cpp
```diff
@@ -172,11 +172,17 @@ void Compiler::fgLocalVarLivenessInit()
 {
     JITDUMP("In fgLocalVarLivenessInit\n");
 
-    // Sort locals first, if we're optimizing
-    if (opts.OptimizationEnabled())
-    {
+#if defined (TARGET_WASM)
+    if (compRationalIRForm) // for LIR, always sort
```
SingleAccretion: And another case `PreciseRefCountsRequired`, though I am not sure why we need to sort here.

src/coreclr/jit/liveness.cpp
```diff
@@ -2201,7 +2207,10 @@ void Compiler::fgRemoveDeadStoreLIR(GenTree* store, BasicBlock* block)
         blockRange.Remove(store);
     }
 
-    assert(!opts.MinOpts());
+#if defined(TARGET_WASM)
+    if (!compRationalIRForm) // conditionally do assert
+#endif
+        assert(!opts.MinOpts());
```
SingleAccretion: It seems that the method should be restructurable such that dead stores are not eliminated even when computing precise liveness for LLVM. It is a somewhat common debugging technique to have unused variables assigned some call return value or something so that they can be looked at in the debugger.

Also would get rid of the unfortunate formatting here.

src/coreclr/jit/liveness.cpp
```diff
@@ -2541,10 +2550,13 @@ void Compiler::fgInterBlockLocalVarLiveness()
      * reported scope, so that it will be visible over the entire scope
      */
 
-    if (opts.compDbgCode && (info.compVarScopesCount > 0))
-    {
-        fgExtendDbgLifetimes();
-    }
+#if defined(TARGET_WASM)
+    if (!compRationalIRForm) // dont extend for the LIR pass - will this break the debugging experience?
```
SingleAccretion: Yeah, kinda... (one of the things) the liveness extension allows for is looking at "dead" variables in the debugger (see the last value they were assigned, etc).

Actually conservative liveness should not be a problem for SSA, it would just result in a lot of dead PHI inputs. Should be fine to tackle this later though (also a good candidate for a `TODO-LLVM`).

src/coreclr/jit/llvm.cpp
```diff
@@ -184,21 +206,35 @@ void Llvm::llvmShutdown()
...
+LocatedLlvmValue getGenTreeValue(GenTree* op)
 {
     return _sdsuMap->at(op);
 }
 
+LocatedLlvmValue getSsaLocalForPhi(unsigned lclNum, unsigned ssaNum)
+{
+    // when the phi arg is a forward reference this fails, so fail those methods.  TODO: would this actually be valid LLVM and if not how would, e.g., DefaultBinder.GetHierarchyDepth be represented in SSA form?
+    if (_localsMap->find({lclNum, ssaNum}) == _localsMap->end())
+        failFunctionCompilation();
```
SingleAccretion: > when the phi arg is a forward reference

Forward reference as it hasn't been defined yet? Hmm, I wonder if this is because we're generating the code in layout order. Generating it in pre-order of the dominator tree would presumably work, as would pre-recording all SSA locals (in a separate walk or in some other way).

src/coreclr/jit/llvm.cpp
```diff
@@ -759,62 +806,156 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
     failFunctionCompilation();
 }
 
-Value* buildCnsInt(llvm::IRBuilder<>& builder, GenTree* node)
+void buildCast(llvm::IRBuilder<>& builder, GenTreeCast* cast)
+{
+    if (cast->gtCastType == TYP_BOOL && cast->TypeGet() == TYP_INT && cast->gtGetOp1()->TypeGet() == TYP_INT)
+    {
+        Value* intValue = builder.CreateZExt(getGenTreeValue(cast->gtGetOp1()).getValue(builder), getLLVMTypeForVarType(TYP_INT));
+        mapGenTreeToValue(cast, intValue); // nothing to do except map the source value to the destination GenTree
```
```suggestion
    if ((cast->CastToType() == TYP_BOOL) && cast->TypeIs(TYP_INT) && cast->CastOp()->TypeIs(TYP_INT))
    {
        Value* intValue = builder.CreateZExt(getGenTreeValue(cast->CastOp()).getValue(builder), getLLVMTypeForVarType(TYP_INT));
        mapGenTreeToValue(cast, intValue); // nothing to do except map the source value to the destination GenTree
```

So when we call `getGenTreeValue(cast->CastOp()).getValue(builder)`, what type does it have? I would expect `int`, but then documentation states that we must use a smaller type for `zext` source than the destination... So how does it really work here?

(This all comes down to `CAST(BOOL)` actually being `int -> bool -> int`, and similar for all other small types)

src/coreclr/jit/llvm.cpp
```diff
@@ -759,62 +806,156 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
+    else
+    {
+        // TODO: other casts
+        failFunctionCompilation();
+    }
+}
+
+void buildCnsInt(llvm::IRBuilder<>& builder, GenTree* node)
 {
     if (node->gtType == var_types::TYP_INT)
```
SingleAccretion: Unrelated to this PR: the `var_types` can be dropped.

src/coreclr/jit/llvm.cpp
```diff
@@ -759,62 +806,156 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
     {
-        return mapGenTreeToValue(node, builder.getInt32(node->AsIntCon()->gtIconVal));
+        mapGenTreeToValue(node, builder.getInt32(node->AsIntCon()->gtIconVal));
+        return;
```
SingleAccretion: A bit more portable way would be to add a `static_cast<int32_t>`, since `gtIconVal` is a `nint` in C# terms, but it works as is and will error out when porting to 64 bit, so, eh.
```suggestion
        mapGenTreeToValue(node, builder.getInt32(node->AsIntCon()->IconValue()));
        return;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -759,62 +806,156 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
+#endif
+
+        // adding a cast as LLVM has GTF_ICON_STR_HDL as i32*, whereas `STORE_LCL_VAR ref` are i8*  This wont work as the cast ends up after the phi.
+        // The methods that come through here must be failing compilation on something else, but this at least allows compilation to continue withouth an LLVM type mismatch assert.
+        // TODO: Fix and possibly make GTF_ICON_STR_HDL i8*?
+        llvmPhiNode->addIncoming(castIfNecessary(builder, localPhiArg.getRawValue(), llvmPhiNode->getType()),
+                                 getLLVMBasicBlockForBlock(phiArg->gtPredBB));
+        i++;
+    }
+    Value* phiValue = requiresLoad ? (Value* )builder.CreateLoad(llvmPhiNode) : llvmPhiNode;
```
```suggestion
    Value* phiValue = requiresLoad ? (Value*)builder.CreateLoad(llvmPhiNode) : llvmPhiNode;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1012,25 +1175,26 @@ Value* visitNode(llvm::IRBuilder<>& builder, GenTree* node)
...
-    llvmBlock = llvm::BasicBlock::Create(_llvmContext, "", _function);
-    _blkToLlvmBlkVectorMap->Set(block, llvmBlock);
-    return llvmBlock;
+    _currentBlock = block;
 }
 
 void endImportingBasicBlock(BasicBlock* block)
 {
-    if (block->bbJumpKind == BBjumpKinds::BBJ_NONE && block->bbNext)
+    if ((block->bbJumpKind == BBjumpKinds::BBJ_NONE) && block->bbNext)
```
```suggestion
    if ((block->bbJumpKind == BBjumpKinds::BBJ_NONE) && (block->bbNext != nullptr))
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1012,25 +1175,26 @@ Value* visitNode(llvm::IRBuilder<>& builder, GenTree* node)
...
     {
         _builder->CreateBr(getLLVMBasicBlockForBlock(block->bbNext));
+        return;
+    }
+    if ((block->bbJumpKind == BBjumpKinds::BBJ_ALWAYS) && block->bbJumpDest)
```
```suggestion
    if ((block->bbJumpKind == BBjumpKinds::BBJ_ALWAYS) && (block->bbJumpDest != nullptr))
```

src/coreclr/jit/llvm.cpp
```diff
@@ -53,27 +53,48 @@ struct SpilledExpressionEntry
...
+typedef std::pair<unsigned, unsigned> ssaPair;
+
+struct ssaPairHash
+{
+    template <class T1, class T2>
+    std::size_t operator()(const std::pair<T1, T2>& pair) const
+    {
+        return std::hash<T1>()(pair.first) ^ std::hash<T2>()(pair.second);
+    }
+};
```
SingleAccretion: Any particular reason for naming inconsistencies? All other types are `PascalCase`d.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -106,16 +106,33 @@ void Compiler::fgResetForSsa()
             }
         }
 
-        for (Statement* const stmt : blk->Statements())
+#if defined(TARGET_WASM)
+        if (compRationalIRForm)
         {
-            for (GenTree* const tree : stmt->TreeList())
+            for (GenTree* tree = blk->GetFirstLIRNode(); tree; tree = tree->gtNext)
```
SingleAccretion: Here and elsewhere: `compRationalIRForm` -> `blk->IsLIR()`, `for (GenTree* tree = blk->GetFirstLIRNode(); tree; tree = tree->gtNext)` -> `for (GenTree* tree : LIR::AsRange(blk))`.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -482,27 +500,62 @@ void SsaBuilder::ComputeIteratedDominanceFrontier(BasicBlock* b, const BlkToBlkV
...
+#if defined(TARGET_WASM)
+GenTree* SsaBuilder::GetPhiNode(BasicBlock* block, unsigned lclNum)
+#else
 static GenTree* GetPhiNode(BasicBlock* block, unsigned lclNum)
+#endif
 {
-    // Walk the statements for phi nodes.
-    for (Statement* const stmt : block->Statements())
+#if defined(TARGET_WASM)
+    if (m_pCompiler->compRationalIRForm)
```
SingleAccretion: Ditto: `IsLIR()`, `LIR:AsRange`, will also allow to get rid of the `static`/`not-static` `#ifdef`.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -482,27 +500,62 @@ void SsaBuilder::ComputeIteratedDominanceFrontier(BasicBlock* b, const BlkToBlkV
...
     {
-        // A prefix of the statements of the block are phi definition nodes. If we complete processing
-        // that prefix, exit.
-        if (!stmt->IsPhiDefnStmt())
+        // Walk the lir node for phi nodes.
```
SingleAccretion:
```suggestion
```
Don't think there's much value in these comments, especially as they're copied.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -482,27 +500,62 @@ void SsaBuilder::ComputeIteratedDominanceFrontier(BasicBlock* b, const BlkToBlkV
...
+        for (GenTree* tree = block->GetFirstLIRNode(); tree; tree = tree->gtNext)
         {
-            break;
-        }
+            // A prefix of the statements of the block are phi definition nodes. If we complete processing
+            // that prefix, exit.
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -482,27 +500,62 @@ void SsaBuilder::ComputeIteratedDominanceFrontier(BasicBlock* b, const BlkToBlkV
...
+            {
+                break;
+            }
 
-        GenTree* tree = stmt->GetRootNode();
+            // skip the phi node to get to the GT_STORE_LCL_VAR
+            if (tree->OperGet() == GT_PHI_ARG || tree->OperGet() == GT_PHI)
+            {
+                continue;
+            }
```
SingleAccretion:
```suggestion
            // skip the phi node to get to the GT_STORE_LCL_VAR
            if (tree->OperIs(GT_PHI_ARG, GT_PHI))
            {
                continue;
            }
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -482,27 +500,62 @@ void SsaBuilder::ComputeIteratedDominanceFrontier(BasicBlock* b, const BlkToBlkV
...
 
-        GenTree* phiLhs = tree->AsOp()->gtOp1;
-        assert(phiLhs->OperGet() == GT_LCL_VAR);
-        if (phiLhs->AsLclVarCommon()->GetLclNum() == lclNum)
+            assert(tree->OperGet() == GT_STORE_LCL_VAR);
```
SingleAccretion:
```suggestion
            assert(tree->OperIs(GT_STORE_LCL_VAR));
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -482,27 +500,62 @@ void SsaBuilder::ComputeIteratedDominanceFrontier(BasicBlock* b, const BlkToBlkV
...
+
+            GenTree* tree   = stmt->GetRootNode();
+            GenTree* phiLhs = tree->AsOp()->gtOp1;
+            assert(phiLhs->OperGet() == GT_LCL_VAR);
+            if (phiLhs->AsLclVarCommon()->GetLclNum() == lclNum)
+            {
+                return tree->AsOp()->gtOp2;
+            }
         }
+#if defined(TARGET_WASM)
```
SingleAccretion: An alternative to this style of `#ifdef`ing is to `return nullptr` after the LIR loop (will also get rid of the indentation changes).

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -550,6 +603,31 @@ void SsaBuilder::InsertPhi(BasicBlock* block, unsigned lclNum)
     JITDUMP("Added PHI definition for V%02u at start of " FMT_BB ".\n", lclNum, block->bbNum);
 }
 
+#if defined(TARGET_WASM)
+void SsaBuilder::InsertPhiToRationalIRForm(BasicBlock* block, unsigned lclNum)
+{
+    assert(m_pCompiler->compRationalIRForm);
```
SingleAccretion:
```suggestion
    assert(block->IsLIR());
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -550,6 +603,31 @@ void SsaBuilder::InsertPhi(BasicBlock* block, unsigned lclNum)
...
+
+    var_types type = m_pCompiler->lvaGetDesc(lclNum)->TypeGet();
+
+    // PHIs and all the associated nodes do not generate any code so the costs are always 0
+    GenTree* phi = new (m_pCompiler, GT_PHI) GenTreePhi(type);
+    phi->SetCosts(0, 0);
+    GenTree* storeLcl = (GenTree*)m_pCompiler->gtNewStoreLclVar(lclNum, phi);
```
SingleAccretion:
```suggestion
    GenTree* phi      = new (m_pCompiler, GT_PHI) GenTreePhi(type);
    GenTree* storeLcl = m_pCompiler->gtNewStoreLclVar(lclNum, phi);
```
LIR doesn't have a way to maintain costs anyway, so, no point setting them.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -550,6 +603,31 @@ void SsaBuilder::InsertPhi(BasicBlock* block, unsigned lclNum)
...
+
+    storeLcl->SetCosts(0, 0);
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -550,6 +603,31 @@ void SsaBuilder::InsertPhi(BasicBlock* block, unsigned lclNum)
...
+
+    GenTree* firstNode = block->GetFirstLIRNode();
+    block->SetFirstLIRNode(phi);
+    phi->gtNext       = storeLcl;
+    storeLcl->gtPrev  = phi;
+    storeLcl->gtNext  = firstNode;
+    firstNode->gtPrev = storeLcl;
```
SingleAccretion: This should use `LIR::AsRange` and the appropriate `InsertBefore/After` functions (will also make the code easier to understand I suppose).

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -581,18 +659,40 @@ void SsaBuilder::AddPhiArg(
     // will be first in linear order as well.
     phi->gtUses = new (m_pCompiler, CMK_ASTNode) GenTreePhi::Use(phiArg, phi->gtUses);
 
-    GenTree* head = stmt->GetTreeList();
-    assert(head->OperIs(GT_PHI, GT_PHI_ARG));
-    stmt->SetTreeList(phiArg);
-    phiArg->gtNext = head;
-    head->gtPrev   = phiArg;
+#if defined(TARGET_WASM)
+    if (m_pCompiler->compRationalIRForm)
```
SingleAccretion:
```suggestion
    if (block->IsLIR())
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -581,18 +659,40 @@ void SsaBuilder::AddPhiArg(
...
+    {
+        // TODO: multiple phis in one block
+        GenTree* firstNode = block->GetFirstLIRNode();
+        block->SetFirstLIRNode(phiArg);
+        phiArg->gtNext    = firstNode;
+        firstNode->gtPrev = phiArg;
```
SingleAccretion: Ditto `LIR::AsRange` and the insertion functions.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -581,18 +659,40 @@ void SsaBuilder::AddPhiArg(
...
+        head->gtPrev   = phiArg;
+#if defined(TARGET_WASM)
+    }
+#endif
 
 #ifdef DEBUG
-    unsigned seqNum = 1;
-    for (GenTree* const node : stmt->TreeList())
+#if defined(TARGET_WASM)
+    if (!m_pCompiler->compRationalIRForm)
```
SingleAccretion:
```suggestion
    if (!block->IsLIR())
```
This double `#ifdef` is kind of ugly... Not clear how to make it better though, sadly.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -667,7 +767,12 @@ void SsaBuilder::InsertPhiFunctions(BasicBlock** postOrder, int count)
                 {
                     // We have a variable i that is defined in block j and live at l, and l belongs to dom frontier of
                     // j. So insert a phi node at l.
+#if defined(TARGET_WASM)
+                    m_pCompiler->compRationalIRForm ? InsertPhiToRationalIRForm(bbInDomFront, lclNum)
```
SingleAccretion: Ditto: `IsLIR()`.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -743,7 +848,23 @@ void SsaBuilder::RenameDef(GenTreeOp* asgNode, BasicBlock* block)
 
     GenTreeLclVarCommon* lclNode;
     bool                 isFullDef;
-    bool                 isLocal = asgNode->DefinesLocal(m_pCompiler, &lclNode, &isFullDef);
+#if defined(TARGET_WASM)
+    bool isLocal;
+    if (m_pCompiler->compRationalIRForm)
```
SingleAccretion:
```suggestion
    if (block->IsLIR())
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -743,7 +848,23 @@ void SsaBuilder::RenameDef(GenTreeOp* asgNode, BasicBlock* block)
...
+    {
+        isLocal   = asgNode->OperIsLocalStore();
+        lclNode   = asgNode->AsLclVarCommon();
+        // copied from IsPartialLclFld but with both GT_STORE_LCL_FLD and GT_LCL_FLD allowed - TODO : is this right?
+        isFullDef = lclNode != nullptr && !(((asgNode->gtOper == GT_STORE_LCL_FLD || asgNode->gtOper == GT_LCL_FLD) &&
+                     (m_pCompiler->lvaTable[lclNode->GetLclNum()].lvExactSize != genTypeSize(asgNode->gtType))));
```
SingleAccretion: So canonical "partial definitions" occur with assignments to struct fields, so the check here should simply be `lclNode->OperIs(GT_STORE_LCL_FLD)`. I don't know how LLVM IR handles SSA for aggregates though, it may be that we may just not care about these concenpts anyway.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -787,7 +908,18 @@ void SsaBuilder::RenameDef(GenTreeOp* asgNode, BasicBlock* block)
             // If necessary, add "lclNum/ssaNum" to the arg list of a phi def in any
             // handlers for try blocks that "block" is within.  (But only do this for "real" definitions,
             // not phi definitions.)
+#if defined(TARGET_WASM)
+            if (m_pCompiler->compRationalIRForm)
```
SingleAccretion:
```suggestion
            if (block->IsLIR())
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -902,29 +1034,64 @@ void SsaBuilder::AddDefToHandlerPhis(BasicBlock* block, unsigned lclNum, unsigne
 #ifdef DEBUG
                 bool phiFound = false;
 #endif
-                // A prefix of blocks statements will be SSA definitions.  Search those for "lclNum".
-                for (Statement* const stmt : handler->Statements())
+                // TODO: llvm.cpp cant handle fin blocks, but this at least does not assert
+#if defined(TARGET_WASM)
+                if (m_pCompiler->compRationalIRForm)
```
SingleAccretion:
```suggestion
                if (block->IsLIR())
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -902,29 +1034,64 @@ void SsaBuilder::AddDefToHandlerPhis(BasicBlock* block, unsigned lclNum, unsigne
...
                 {
-                    // If the tree is not an SSA def, break out of the loop: we're done.
-                    if (!stmt->IsPhiDefnStmt())
+                    // Walk the lir node for phi nodes.
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -902,29 +1034,64 @@ void SsaBuilder::AddDefToHandlerPhis(BasicBlock* block, unsigned lclNum, unsigne
...
+                    for (GenTree* tree = handler->GetFirstLIRNode(); tree; tree = tree->gtNext)
```
SingleAccretion: Ditto: `LIR::AsRange`.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -902,29 +1034,64 @@ void SsaBuilder::AddDefToHandlerPhis(BasicBlock* block, unsigned lclNum, unsigne
...
                     {
-                        break;
-                    }
+                        // A prefix of the statements of the block are phi definition nodes. If we complete processing
+                        // that prefix, exit.
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -902,29 +1034,64 @@ void SsaBuilder::AddDefToHandlerPhis(BasicBlock* block, unsigned lclNum, unsigne
...
+                        if (!tree->IsPhiNode())
+                        {
+                            break;
+                        }
 
-                    GenTree* tree = stmt->GetRootNode();
+                        // skip the phi node to get to the GT_STORE_LCL_VAR
+                        if (tree->OperGet() == GT_PHI_ARG || tree->OperGet() == GT_PHI)
```
SingleAccretion:
```suggestion
                        if (tree->OperIs(GT_PHI_ARG, GT_PHI))
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -902,29 +1034,64 @@ void SsaBuilder::AddDefToHandlerPhis(BasicBlock* block, unsigned lclNum, unsigne
...
+                        {
+                            continue;
+                        }
 
-                    assert(tree->IsPhiDefn());
+                        assert(tree->OperGet() == GT_STORE_LCL_VAR);
```
SingleAccretion:
```suggestion
                        assert(tree->OperIs(GT_STORE_LCL_VAR));
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -1053,12 +1220,13 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
         }
     }
 
-    // Walk the statements of the block and rename definitions and uses.
-    for (Statement* const stmt : block->Statements())
+#if defined(TARGET_WASM)
+    if (m_pCompiler->compRationalIRForm)
```
SingleAccretion:
```suggestion
    if (block->IsLIR())
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -1053,12 +1220,13 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
...
     {
-        for (GenTree* const tree : stmt->TreeList())
+        // Walk the statements of the block and rename definitions and uses.
+        for (GenTree* tree = block->GetFirstLIRNode(); tree; tree = tree->gtNext)
```
SingleAccretion: Ditto: `LIR:AsRange`.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -1116,43 +1306,90 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
 //
 void SsaBuilder::AddPhiArgsToSuccessors(BasicBlock* block)
 {
+
     for (BasicBlock* succ : block->GetAllSuccs(m_pCompiler))
     {
-        // Walk the statements for phi nodes.
-        for (Statement* const stmt : succ->Statements())
+#if defined(TARGET_WASM)
+        if (m_pCompiler->compRationalIRForm)
```
SingleAccretion:
```suggestion
        if (succ->IsLIR())
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -1116,43 +1306,90 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
...
         {
-            // A prefix of the statements of the block are phi definition nodes. If we complete processing
-            // that prefix, exit.
-            if (!stmt->IsPhiDefnStmt())
+            // Walk the nodes for phi nodes.
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -1116,43 +1306,90 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
...
+            for (GenTree* tree = succ->GetFirstLIRNode(); tree; tree = tree->gtNext)
```
SingleAccretion: Ditto: `LIR::AsRange`.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -1116,43 +1306,90 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
...
             {
-                break;
-            }
+                // A prefix of the statements of the block are phi definition nodes. If we complete processing
+                // that prefix, exit.
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -1116,43 +1306,90 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
...
-            // (Can we assert that its the head of the list?  This should only happen when we add
-            // during renaming for a definition that occurs within a try, and then that's the last
-            // value of the var within that basic block.)
+                unsigned    lclNum = tree->AsLclVarCommon()->GetLclNum();
+                GenTreePhi* phi    = tree->gtGetOp1()->AsPhi();
+                unsigned    ssaNum = m_renameStack.Top(lclNum);
+                // Search the arglist for an existing definition for ssaNum.
+                // (Can we assert that its the head of the list?  This should only happen when we add
+                // during renaming for a definition that occurs within a try, and then that's the last
+                // value of the var within that basic block.)
```
SingleAccretion:
```suggestion

```
Mainly just because they're duplicates.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -155,7 +172,8 @@ int SsaBuilder::TopologicalSort(BasicBlock** postOrder, int count)
     DBEXEC(VERBOSE, comp->fgDispBasicBlocks());
     DBEXEC(VERBOSE, comp->fgDispHandlerTab());
 
-    auto DumpBlockAndSuccessors = [](Compiler* comp, BasicBlock* block) {
+    auto DumpBlockAndSuccessors = [](Compiler* comp, BasicBlock* block)
```
SingleAccretion: Unintended formatting change?

src/coreclr/jit/llvm.cpp
```diff
@@ -759,62 +806,156 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
+void buildPhi(llvm::IRBuilder<>& builder, GenTreePhi* phi)
 {
-    return mapGenTreeToValue(node, builder.CreateICmpNE(op1, op2));
+    llvm::PHINode* llvmPhiNode = nullptr;
+    unsigned i = 0;
+    bool requiresLoad = false;
+    for (GenTreePhi::Use& use : phi->Uses())
+    {
+        assert(use.GetNode()->OperIs(GT_PHI_ARG));
+        GenTreePhiArg* phiArg = use.GetNode()->AsPhiArg();
```
```suggestion
        GenTreePhiArg* phiArg = use.GetNode()->AsPhiArg();
```
Assert redundant with the one inside `AsPhiArg()`.

src/coreclr/jit/llvm.cpp
```diff
@@ -759,62 +806,156 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
+        unsigned       lclNum = phiArg->GetLclNum();
+        unsigned       ssaNum = phiArg->GetSsaNum();
+
+        LocatedLlvmValue localPhiArg = getSsaLocalForPhi(lclNum, ssaNum);
+        if (i == 0)
+        {
+            requiresLoad = localPhiArg.valueRequiresLoad();
+            Type* phiType = getLLVMTypeForVarType(phi->TypeGet());
+
+            llvmPhiNode = builder.CreatePHI(requiresLoad ? phiType->getPointerTo() : phiType, phi->NumChildren());
```
SingleAccretion: `NumChildren` does linear counting, making the loop quadratic. It should be cached instead.

src/coreclr/jit/llvm.cpp
```diff
@@ -759,62 +806,156 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
+        }
+#ifdef DEBUG
+        else
+        {
+            // phi args must be all direct, or all indirect
+            assert(requiresLoad == localPhiArg.valueRequiresLoad());
+        }
+#endif
```
SingleAccretion: No real need for `#ifdef DEBUG` I suppose?

SingleAccretion (commented):
Just some comments on style and code beautification.

eng/build.ps1
```diff
@@ -258,7 +258,7 @@ if ($os -eq "Browser") {
   # override default arch for Browser, we only support wasm
   $arch = "wasm"
 
-  if ($msbuild -eq $True) {
+  if ($msbuild -eq $True -and $runtimeFlavor -ne "CoreCLR") {
     Write-Error "Using the -msbuild option isn't supported when building for Browser on Windows, we need need ninja for Emscripten."
```
shahid-pk, line 262: nit: double need

src/coreclr/jit/liveness.cpp
yowl: That looks like a correct assessment.  It gets a null pointer from ` block->GetFirstLIRNode()` in `InsertPhiToRationalIRForm`

src/coreclr/jit/llvm.cpp
yowl: I'm not sure.  The IR for this is
```
------------ BB03 [005..01D) -> BB03 (cond), preds={BB02,BB03} succs={BB04,BB03}
     (  0,  0) [000070] ------------        t70 =    PHI_ARG   ref    V02 loc1         u:3
     (  0,  0) [000069] ------------        t69 =    PHI_ARG   int    V01 loc0         u:3
     (  0,  0) [000068] ------------        t68 =    PHI_ARG   ref    V02 loc1         u:1
     (  0,  0) [000067] ------------        t67 =    PHI_ARG   int    V01 loc0         u:1
                                                  /--*  t69    int
                                                  +--*  t67    int
     (  0,  0) [000065] ------------        t65 = *  PHI       int
                                                  /--*  t65    int
     (  0,  0) [000066] DA----------              *  STORE_LCL_VAR int    V01 loc0         d:2
                                                  /--*  t70    ref
                                                  +--*  t68    ref
     (  0,  0) [000063] ------------        t63 = *  PHI       ref
                                                  /--*  t63    ref
     (  0,  0) [000064] DA----------              *  STORE_LCL_VAR ref    V02 loc1         d:2
               [000052] ------------                 IL_OFFSET void   IL offset: 0x5
               [000053] ------------                 IL_OFFSET void   IL offset: 0x6
N001 (  3,  2) [000009] ------------         t9 =    LCL_VAR   int    V01 loc0         u:2 (last use)
N002 (  1,  1) [000010] ------------        t10 =    CNS_INT   int    1
                                                  /--*  t9     int
                                                  +--*  t10    int
N003 (  5,  4) [000011] ------------        t11 = *  ADD       int
                                                  /--*  t11    int
N005 (  9,  7) [000013] DA----------              *  STORE_LCL_VAR int    V01 loc0         d:3
               [000054] ------------                 IL_OFFSET void   IL offset: 0xa
N001 (  1,  4) [000015] H-----------        t15 =    CNS_INT(h) int    0x4300F8 ftn
                                                  /--*  t15    int
N003 (  5,  7) [000017] DA----------              *  STORE_LCL_VAR int    V06 tmp1         d:1
N002 (  3,  2) [000014] ------------        t14 =    LCL_VAR   ref    V02 loc1         u:2 (last use)
N003 (  3,  2) [000018] ------------        t18 =    LCL_VAR   int    V06 tmp1         u:1 (last use)
                                                  /--*  t14    ref    this in rcx
                                                  +--*  t18    int    calli tgt
N004 ( 23,  8) [000019] --CXG-------        t19 = *  CALL ind nullcheck ref
                                                  /--*  t19    ref
N006 ( 27, 11) [000021] DA-XG-------              *  STORE_LCL_VAR ref    V07 tmp2         d:1
               [000055] ------------                 IL_OFFSET void   IL offset: 0x10
N001 (  3,  2) [000022] ------------        t22 =    LCL_VAR   ref    V07 tmp2         u:1 (last use)
                                                  /--*  t22    ref
N003 (  7,  5) [000024] DA----------              *  STORE_LCL_VAR ref    V02 loc1         d:3
               [000056] ------------                 IL_OFFSET void   IL offset: 0x11
               [000057] ------------                 IL_OFFSET void   IL offset: 0x12
N003 (  3,  2) [000026] ------------        t26 =    LCL_VAR   ref    V02 loc1         u:3
N004 (  1,  1) [000027] ------------        t27 =    CNS_INT   ref    null
                                                  /--*  t26    ref    arg0 in rcx
                                                  +--*  t27    ref    arg1 in rdx
N005 ( 18,  7) [000028] --CXG-------        t28 = *  CALL      int    System.Type.op_Inequality
                                                  /--*  t28    int
N006 ( 19,  9) [000029] ---XG-------        t29 = *  CAST      int <- bool <- int
                                                  /--*  t29    int
N008 ( 23, 12) [000031] DA-XG-------              *  STORE_LCL_VAR int    V08 tmp3         d:1
               [000058] ------------                 IL_OFFSET void   IL offset: 0x19
N001 (  3,  2) [000032] ------------        t32 =    LCL_VAR   int    V08 tmp3         u:1 (last use)
                                                  /--*  t32    int
N002 (  4,  4) [000048] ------------        t48 = *  CAST      int <- bool <- int
                                                  /--*  t48    int
N004 (  8,  7) [000034] DA----------              *  STORE_LCL_VAR int    V03 loc2         d:1
               [000059] ------------                 IL_OFFSET void   IL offset: 0x1a
N001 (  3,  2) [000035] ------------        t35 =    LCL_VAR   int    V03 loc2         u:1 (last use)
N002 (  1,  1) [000036] ------------        t36 =    CNS_INT   int    0
                                                  /--*  t35    int
                                                  +--*  t36    int
N003 (  5,  4) [000037] J------N----        t37 = *  NE        int
                                                  /--*  t37    int
N004 (  7,  6) [000038] ------------              *  JTRUE     void
```

Some of the args are in the same block so would reordering make any difference?

src/coreclr/jit/llvm.cpp
yowl: An example
```
------------ BB02 [000..007) (return), preds={BB01} succs={}
               [000010] ------------                 IL_OFFSET void   IL offset: 0x0
N002 (  3,  2) [000001] ------------         t1 =    LCL_VAR   ref    V00 this         u:1 (last use)
                                                  /--*  t1     ref    this in rcx
N003 ( 17,  6) [000002] --CXG-------         t2 = *  CALL      int    System.Type.get_IsSignatureType
                                                  /--*  t2     int
N004 ( 18,  8) [000003] ---XG-------         t3 = *  CAST      int <- bool <- int
                                                  /--*  t3     int
N006 ( 22, 11) [000005] DA-XG-------              *  STORE_LCL_VAR int    V02 tmp1         d:1
               [000011] ------------                 IL_OFFSET void   IL offset: 0x6
N001 (  3,  2) [000006] ------------         t6 =    LCL_VAR   int    V02 tmp1         u:1 (last use)
                                                  /--*  t6     int
N002 (  4,  4) [000009] ------------         t9 = *  CAST      int <- bool <- int
                                                  /--*  t9     int
N003 (  5,  5) [000007] ------------              *  RETURN    int
```
`CastOp` :
image
And it gets from `getGenTreeValue` an `i8`.
image

When creating the call the `CORINFO_SIG_INFO` is
image
 so we end up ok, an i8 to i32
```
  %6 = call i8 @S_P_CoreLib_System_Type__get_IsSignatureType(i8* %4), !dbg !3637
  %7 = zext i8 %6 to i32, !dbg !3637
```
All the example I've seen are from a call, so maybe just getting lucky so far.

src/coreclr/jit/llvm.cpp
yowl: Thanks, will do that.

src/coreclr/jit/llvm.cpp
yowl: nope, corrected.

src/coreclr/jit/llvm.cpp
SingleAccretion: I see.

> All the example I've seen are from a call, so maybe just getting lucky so far.

I would say yes. I suppose `bool`s are special because they do not actually exist. One other source of casts with them is "normalization on store", but that is harder to hit because C# only allows boolean expressions on the RHS, and those do not need normalization. Anyhow, this will need to be tackled in any case when other casts are implemented.

src/coreclr/jit/llvm.cpp
SingleAccretion: Hmm, I wonder then if this is due to `compInitMem`'s hidden definition again.

src/coreclr/jit/ssabuilder.cpp
yowl: Thanks, I did wonder if there was a `for` preference.

yowl:
Thanks for the feedback and improvements,   Still got the last comment to look at.  Namely https://github.com/dotnet/runtimelab/pull/1598#discussion_r717075221

src/coreclr/jit/ssabuilder.cpp
yowl: Missed this, accepting yours.

eng/build.ps1
yowl: Sorry, dont understand.  mono might need ninja but we dont need it here.

src/coreclr/jit/llvm.cpp
yowl: I should include the previous blocks.
```

------------ BB02 [000..005), preds={BB01} succs={BB03}
               [000049] ------------                 IL_OFFSET void   IL offset: 0x0
               [000050] ------------                 IL_OFFSET void   IL offset: 0x1
N001 (  1,  1) [000002] ------------         t2 =    CNS_INT   int    0
                                                  /--*  t2     int
N003 (  5,  4) [000004] DA----------              *  STORE_LCL_VAR int    V01 loc0         d:1
               [000051] ------------                 IL_OFFSET void   IL offset: 0x3
N001 (  3,  2) [000005] ------------         t5 =    LCL_VAR   ref    V00 arg0         u:1 (last use)
                                                  /--*  t5     ref
N003 (  7,  5) [000007] DA----------              *  STORE_LCL_VAR ref    V02 loc1         d:1
```

yowl: I don't think there is a problem here as 
```
define hidden i32 @main() #0 {
entry:
  %ssa1 = alloca i32
  %phiarg1 = load i32, i32* %ssa1
  br label %loop

loop:
  %phi = phi i32 [ %phiarg1, %entry ], [ %phiarg2, %loop ]
  %ssa2 = alloca i32
  %phiarg2 = load i32, i32* %ssa2
  %eq = icmp eq i32 %phi, 0
  br i1 %eq, label %loop, label %ret

ret:
  ret i32 0
}
```
Seems to be valid LLVM anyway.  The problem is just in `llvm.cpp` that needs to handle this scenario.

eng/build.ps1
shahid-pk: i mean to say that in the statement

>  we need need ninja for Emscripten

need should be once.  A minor typo. I found while reading the code.

yowl: Thanks, its not part of this change, and its upstream in the runtime repo (https://github.com/dotnet/runtime/blob/72d643d05ab23888f30a57d447154e36f979f3d1/eng/build.ps1#L276).  Should be fixed there really.

src/coreclr/jit/llvm.cpp
SingleAccretion: I think I see what is going on here now.

We encounter `d:3` while building the PHI, but `d:3` is only defined after the PHI, and thus we fail. And we need the `Value*` for `d:3` (`phiarg2`), but we cannot have that. Or can we? Would it be possible to insert "missing" inputs into the map while building the PHI?

yowl: Yes, precisely that.  Sorry I wasn't clear.  We should be able to do this, I don't see why we can't create the `Value*` when the phi arg is processed and then pick that up for the `d:3`.

yowl: I think just allowing the phi arg to add to the map should be enough.

yowl: Wasn't quite as simple as I'd hoped, but nothing horrible.  I store the phis that need "filling in" in a map, and when the `GT_STORE_LCL_VAR` is processed I fill it in.  If all the phi args are in LIR processed after the phi, then it breaks.  A fix for that might be to just do all phi args at the end of the block regardless of where the definition occurs.  I've not done that here, on the assumption it wont happen much and I can put it off for a bit.  I also added `ADD` for ints and a type conversion to facilitate my testing.

yowl:
Sometimes I seem to not get a local build/test cleanly done and it runs something old.  The tests failed and I'm looking at that.

src/coreclr/jit/llvm.cpp
yowl: Enabling these methods has unearthed some problems.  Can you take a look at this excerpt LIR:
```
----------- BB10 [044..047) -> BB11 (always), preds={BB08,BB09} succs={BB11}
     (  0,  0) [000116] ------------       t116 =    PHI_ARG   int    V10 tmp5         u:3
     (  0,  0) [000115] ------------       t115 =    PHI_ARG   int    V10 tmp5         u:2
                                                  /--*  t116   int
                                                  +--*  t115   int
               [000110] ------------       t110 = *  PHI       int
                                                  /--*  t110   int
               [000111] DA----------              *  STORE_LCL_VAR int    V10 tmp5         d:1
N001 (  3,  2) [000049] ------------        t49 =    LCL_VAR   int    V10 tmp5         u:1 (last use)
                                                  /--*  t49    int
N002 (  4,  4) [000088] ------------        t88 = *  CAST      int <- bool <- int
                                                  /--*  t88    int
N004 (  8,  7) [000051] DA----------              *  STORE_LCL_VAR int    V04 loc3         d:2
               [000106] ------------                 IL_OFFSET void   IL offset: 0x45

------------ BB11 [047..049) (return), preds={BB03,BB10} succs={}
     (  0,  0) [000119] ------------       t119 =    PHI_ARG   bool   V04 loc3         u:3
     (  0,  0) [000114] ------------       t114 =    PHI_ARG   bool   V04 loc3         u:2
                                                  /--*  t119   bool
                                                  +--*  t114   bool
               [000108] ------------       t108 = *  PHI       bool
                                                  /--*  t108   bool
               [000109] DA----------              *  STORE_LCL_VAR bool   V04 loc3         d:1
```
The `PHI_ARG` for `t114` is `bool`, but `V04`,`d:2` is `int`.  One of these int<-bool<-int is in play here. Is this valid, or have I messed up in ssabuilder?

SingleAccretion: It is valid, yes. The reason for this is that PHI definitions always use the "real" type of the local, while regular assignments/stores can (and, in most cases, but not all, do) use the "actual" type of the local (`genActualType`).

Of course, a `stloc <bool local>` truncates to the width of one byte, and that's what the cast represents.

yowl: Thanks,  I'll adjust the LLVM to cope.

yowl: I insert a cast if required before the end of the block where the local is defined.

yowl:
🤞 this should now pass.  I fixed a couple of problems surfaced by this change enabling more code to be compiled by the RyuJIT backend.  Arguments were cast in the first block encountered, instead of the prolog.  And `i1`s were being sign extended causing, strangely, just a problem with default interface generic vitrual methods.

jkotas:
@SingleAccretion Could you please review this change?

src/coreclr/jit/liveness.cpp
```diff
@@ -2182,6 +2182,9 @@ bool Compiler::fgTryRemoveNonLocal(GenTree* node, LIR::Range* blockRange)
 //
 void Compiler::fgRemoveDeadStoreLIR(GenTree* store, BasicBlock* block)
 {
+    if (!PreciseRefCountsRequired())
+        return;
+
```
SingleAccretion, line 2188: Ideally we should not be depending on dead store elimination for correctness (it is also a bit weird - ref counts are only tangentially related to liveness and are really for CSE and register allocator to use). Fine to defer though.
```suggestion
    // TODO-LLVM: we should not need this for correctness, investigate removing.
    if (!PreciseRefCountsRequired())
        return;

```

src/coreclr/jit/llvm.cpp
```diff
@@ -759,64 +834,196 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
+        mapGenTreeToValue(node, builder.CreateIntToPtr(builder.getInt32(intCon), Type::getInt8PtrTy(_llvmContext))); // TODO: wasm64
+        return;
     }
     failFunctionCompilation();
 }
 
-Value* buildInd(llvm::IRBuilder<>& builder, GenTree* node, Value* ptr)
+void buildInd(llvm::IRBuilder<>& builder, GenTree* node, Value* ptr)
 {
     // TODO: Simplify genActualType(node->TypeGet()) to just genActualType(node) when main is merged
```
```suggestion
```

src/coreclr/jit/llvm.cpp
```diff
@@ -759,64 +834,196 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
     // first cast the pointer to create the correct load instructions, then cast the result incase we are loading a small int into an int32
-    return mapGenTreeToValue(node,
-                             castIfNecessary(builder, builder.CreateLoad(
+    mapGenTreeToValue(node, castIfNecessary(builder, builder.CreateLoad(
                                  castIfNecessary(builder, ptr,
-                                                 getLLVMTypeForVarType(node->TypeGet())->getPointerTo())), getLLVMTypeForVarType(genActualType(node->TypeGet()))));
+                                     getLLVMTypeForVarType(node->TypeGet())->getPointerTo())), getLLVMTypeForVarType(genActualType(node->TypeGet()))));
```
```suggestion
                                     getLLVMTypeForVarType(node->TypeGet())->getPointerTo())), getLLVMTypeForVarType(genActualType(node))));
```

src/coreclr/jit/llvm.cpp
```diff
@@ -759,64 +834,196 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
+        {
+            op2 = builder.CreateIntToPtr(op2, op1->getType());
+        }
+        else if (op2->getType()->isPointerTy() && op1->getType()->isIntegerTy())
+        {
+            op1 = builder.CreateIntToPtr(op1, op2->getType());
+        }
+        else
+        {
+            // TODO: other valie LIR comparisons
```
```suggestion
            // TODO-LLVM: other valid LIR comparisons
```

src/coreclr/jit/llvm.cpp
```diff
@@ -759,64 +834,196 @@ Value* buildCall(llvm::IRBuilder<>& builder, GenTree* node)
...
+            }
+            else
+            {
+                phiRealArgValue = localPhiArg.getRawValue();
+            }
+            llvmPhiNode->addIncoming(phiRealArgValue, getLLVMBasicBlockForBlock(phiArg->gtPredBB));
+        }
+        i++;
+    }
+    assert(requiresLoadSet); // TODO-LLVM: If all the phi args are in this or later blocks, then this will be hit, and this load/no load swtich will have to be inserted when finishing the block, or something...
```
SingleAccretion, line 1017: Alternative solution to this problem is to turn all locals from the shadow stack into `IND`s in a pre-codegen pass (this will also be needed for "address exposed" locals).

src/coreclr/jit/llvm.cpp
```diff
@@ -896,9 +1106,11 @@ N011 ( 34, 17) [000016] DA-XG-------              *  STORE_LCL_VAR int    V03 lo
 */
 Value* widenIntIfNecessary(llvm::IRBuilder<>& builder, Value* intValue)
 {
-    if (intValue->getType()->getPrimitiveSizeInBits() < TARGET_POINTER_SIZE * 8)
+    llvm::TypeSize intSize = intValue->getType()->getPrimitiveSizeInBits();
+    if (intSize < TARGET_POINTER_SIZE * 8)
     {
-        return builder.CreateIntCast(intValue, Type::getInt32Ty(_llvmContext), true);
+        bool isSigned = intSize > 1; // bools are not signed;
```
SingleAccretion: What about other small unsigned types?

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -550,6 +587,26 @@ void SsaBuilder::InsertPhi(BasicBlock* block, unsigned lclNum)
...
 }
 
+#if defined(TARGET_WASM)
+void SsaBuilder::InsertPhiToRationalIRForm(BasicBlock* block, unsigned lclNum)
+{
+    assert(block->IsLIR());
+
+    var_types type = m_pCompiler->lvaGetDesc(lclNum)->TypeGet();
+
+    // PHIs and all the associated nodes do not generate any code so the costs are always 0
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -550,6 +587,26 @@ void SsaBuilder::InsertPhi(BasicBlock* block, unsigned lclNum)
...
+    GenTree* phi      = new (m_pCompiler, GT_PHI) GenTreePhi(type);
+    GenTree* storeLcl = (GenTree*)m_pCompiler->gtNewStoreLclVar(lclNum, phi);
```
SingleAccretion:
```suggestion
    GenTree* storeLcl = m_pCompiler->gtNewStoreLclVar(lclNum, phi);
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -550,6 +587,26 @@ void SsaBuilder::InsertPhi(BasicBlock* block, unsigned lclNum)
...
+
+    LIR::Range& lirRange = LIR::AsRange(block);
+    GenTree*    firstNode = lirRange.FirstNode();
```
SingleAccretion:
```suggestion
    LIR::Range& lirRange  = LIR::AsRange(block);
    GenTree*    firstNode = lirRange.FirstNode();
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -743,7 +815,22 @@ void SsaBuilder::RenameDef(GenTreeOp* asgNode, BasicBlock* block)
...
     bool                 isFullDef;
-    bool                 isLocal = asgNode->DefinesLocal(m_pCompiler, &lclNode, &isFullDef);
+#if defined(TARGET_WASM)
+    bool isLocal;
+    if (block->IsLIR())
+    {
+        isLocal   = asgNode->OperIsLocalStore();
+        lclNode   = asgNode->AsLclVarCommon();
+        isFullDef = lclNode != nullptr && !(lclNode->OperIs(GT_STORE_LCL_FLD) &&
+                     (m_pCompiler->lvaTable[lclNode->GetLclNum()].lvExactSize != genTypeSize(asgNode->gtType)));
```
SingleAccretion:
```suggestion
        isLocal   = asgNode->OperIsLocalStore();
        lclNode   = isLocal ? asgNode->AsLclVarCommon() : nullptr;
        isFullDef = isLocal && lclNode->OperIs(GT_STORE_LCL_VAR);
```
Let's simplify this a bit and prevent crashes (`AsLclVarCommon` has an assert inside it that uses `this`).

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -902,29 +999,61 @@ void SsaBuilder::AddDefToHandlerPhis(BasicBlock* block, unsigned lclNum, unsigne
 #ifdef DEBUG
                 bool phiFound = false;
 #endif
-                // A prefix of blocks statements will be SSA definitions.  Search those for "lclNum".
-                for (Statement* const stmt : handler->Statements())
+                // TODO: llvm.cpp cant handle fin blocks, but this at least does not assert
```
SingleAccretion:
```suggestion
                // TODO-LLVM: llvm.cpp cant handle fin blocks, but this at least does not assert.
```

src/coreclr/jit/compiler.h
```diff
@@ -5102,6 +5114,9 @@ class Compiler
     // rather than the "use" SSA number recorded in the tree "lcl".
     inline unsigned GetSsaNumForLocalVarDef(GenTree* lcl);
 
+    // 
+    inline bool Compiler::PreciseRefCountsRequired();
```
```suggestion
    inline bool PreciseRefCountsRequired();
```

src/coreclr/jit/compiler.hpp
```diff
@@ -4188,6 +4188,15 @@ unsigned Compiler::GetSsaNumForLocalVarDef(GenTree* lcl)
     }
 }
 
+bool Compiler::PreciseRefCountsRequired()
```
SingleAccretion:
```suggestion
inline bool Compiler::PreciseRefCountsRequired()
```

SingleAccretion (approved):
LGTM, modulo comments and suggestions.

Very nice work! Out of curiosity, how many methods is the RyuJit-based codegen able to handle now?

src/coreclr/jit/llvm.cpp
yowl: Thanks adding a link to this comment

yowl:
> how many methods is the RyuJit-based codegen able to handle now?

Thanks, its around 10/11%  which from memory in HelloWasm, is 1600 methods.

yowl:
Hitting an assert
` Assertion failed '(lclNode->gtFlags & GTF_VAR_USEASG) != 0' in 'System.Span`1:.ctor(byref,int):this' during 'SSA: insert phis' (IL size 34)  `

https://dev.azure.com/dnceng/public/_build/results?buildId=1399816&view=logs&j=7d83a857-5248-5c91-a515-9f1763cf5d33&t=e8c9661b-399a-55ca-e631-825c2e966bbb&l=667

Looking into that.

yowl:
> Thanks, its around 10/11% which from memory in HelloWasm, is 1600 methods.

We'd get a lot more I think if I sorted out the return in a spill slot for REF types.  Next job I think.

src/coreclr/jit/llvm.cpp
yowl: Yes, this was broken I think but the methods in question were failing on other missing functionality.  I changed this to always be a zero extend and changed the method name.  There is some clean up to be done here, but I've just remarked on it and added a TODO.

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -743,7 +814,22 @@ void SsaBuilder::RenameDef(GenTreeOp* asgNode, BasicBlock* block)
...
     bool                 isFullDef;
-    bool                 isLocal = asgNode->DefinesLocal(m_pCompiler, &lclNode, &isFullDef);
+#if defined(TARGET_WASM)
+    bool isLocal;
+    if (block->IsLIR())
+    {
+        isLocal   = asgNode->OperIsLocalStore();
+        lclNode   = isLocal ? asgNode->AsLclVarCommon() : nullptr;
+        isFullDef = isLocal && (lclNode->OperIs(GT_STORE_LCL_VAR)
+            || (m_pCompiler->lvaTable[lclNode->GetLclNum()].lvExactSize == genTypeSize(asgNode->gtType)));
```
yowl, line 824: I added this condition in for the size comparison as it was hitting the asserts on `GTF_VAR_DEF` below.

yowl:
In addition to the comments in the commit.  This latest commit groups the phis together.  LLVM doesn't allow anything in between.  The easy option was to insert each phi at the start of the block, this reverses the phis compared to the IR, but I couldn't think of why one phi would depend on another in the same block, so thought that was ok.

yowl:
windows failed on file copy user mapped section open problem.  Closing and reopening.

yowl:
This is the PR that keeps on giving.  I don't see the error only `Process terminated.`   I'll try again tomorrow

yowl:
Passes now, 😌

src/coreclr/jit/compiler.hpp
```diff
@@ -4188,6 +4188,15 @@ unsigned Compiler::GetSsaNumForLocalVarDef(GenTree* lcl)
     }
 }
 
+inline bool Compiler::PreciseRefCountsRequired()
```
jkotas, line 4191: Would it make sense to introduce `PreciseRefCountsRequired` upstream? It looks like general goodness.

jkotas (approved):
Thank you!

src/coreclr/jit/compiler.hpp
yowl: I'm happy to create the PR to see what others think.  Just leave
https://github.com/dotnet/runtimelab/blob/9ff990c48ccae699b4ef6faeda70672d0e9b22c5/src/coreclr/jit/liveness.cpp#L2185-L2187
out?

jkotas: Yep

PR #1603, opened 2021-09-28
Merge main into feature/regexsrm

stephentoub:
This merges dotnet/runtime main into feature/regexsrm, which we haven't done in a while.  Most impactfully, it brings in the regex source generator, which required some overhauling to how tests were structured.  I've adapted to that in this PR.  Note this does significantly increase test execution time for regex; on my machines, tests now take closer to a minute.  I'll be looking for ways to decrease that.
cc: @veanes, @olsaarik


src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -541,10 +597,16 @@ public static IEnumerable<object[]> Match_VaryingLengthStrings_MemberData()
...
         [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework, "Takes several minutes on .NET Framework")]
-        [Theory]
+        [ConditionalTheory]
         [MemberData(nameof(Match_VaryingLengthStrings_MemberData))]
         public async Task Match_VaryingLengthStrings(RegexEngine engine, RegexOptions options)
         {
+            if (RegexHelpers.IsNonBacktracking(engine))
+            {
+                // TODO-NONBACKTRACKING: Stack overflowing
+                throw new SkipTestException("ActiveIssue: Stack overflows on RegexOptions.NonBacktracking");
```
stephentoub, line 607: @olsaarik, fyi, this test hits a stack overflow for me with the NonBacktracking engine.  We'll need to investigate why and fix it.  For now, I'm skipping it.,

src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
```diff
@@ -58,6 +67,13 @@ public async Task Docs_Examples_ScanningHrefs(RegexEngine engine)
         [MemberData(nameof(RegexHelpers.AvailableEngines_MemberData), MemberType = typeof(RegexHelpers))]
         public async Task Docs_Examples_MDYtoDMY(RegexEngine engine)
         {
+            if (RegexHelpers.IsNonBacktracking(engine))
+            {
+                // named group replacements not supported
+                // TODO-NONBACKTRACKING: This is producing an incorrect result... should it throw instead?
```
stephentoub, line 73: @veanes, @olsaarik, this test ends up producing a string from Regex.Replace, but it's not the right answer (it treats the names as text rather than as replacements and then throwing).  Presumably that's something we need to fix in the implementation, to throw.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -45,464 +46,505 @@ public static IEnumerable<object[]> Match_Basic_TestData_LazyLoops()
...
+                yield return new object[] { engine, @".+a", "baa", RegexOptions.None, 0, 3, true, "baa" };
+                yield return new object[] { engine, @"[ab]+a", "cacbaac", RegexOptions.None, 0, 7, true, "baa" };
+                yield return new object[] { engine, @"^(\d{2,3}){2}$", "1234", RegexOptions.None, 0, 4, true, "1234" };
+                yield return new object[] { engine, @"(\d{2,3}){2}", "1234", RegexOptions.None, 0, 4, true, "1234" };
+                yield return new object[] { engine, @"((\d{2,3})){2}", "1234", RegexOptions.None, 0, 4, true, "1234" };
+                if (!RegexHelpers.IsNonBacktracking(engine))
+                {
+                    // TODO-NONBACKTRACKING: Produce "1234" rather than "123", which could be considered an alternative valid result given greedy semantics
+                    yield return new object[] { engine, @"(\d{2,3})+", "1234", RegexOptions.None, 0, 4, true, "123" };
+                    yield return new object[] { engine, @"(\d{2,3})*", "123456", RegexOptions.None, 0, 4, true, "123" };
```
stephentoub, line 181: @veanes, @olsaarik, the NonBacktracking engine is producing a different answer here than the other engines.  I assume we'd consider its answer to be valid, but wanted to double-check it's intentional.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -45,464 +46,505 @@ public static IEnumerable<object[]> Match_Basic_TestData_LazyLoops()
...
+                yield return new object[] { engine, "^b", "abc", RegexOptions.None, 0, 3, false, string.Empty };
+
+                // Trim leading and trailing whitespace
+                yield return new object[] { engine, @"\s*(.*?)\s*$", " Hello World ", RegexOptions.None, 0, 13, true, " Hello World " };
+
+                if (!RegexHelpers.IsNonBacktracking(engine))
+                {
+                    // TODO-NONBACKTRACKING: This test fails by successfully matching.
+                    // < in group
+                    yield return new object[] { engine, @"(?<cat>cat)\w+(?<dog-0>dog)", "cat_Hello_World_dog", RegexOptions.None, 0, 19, false, string.Empty };
```
stephentoub, line 298: Seems something is amiss here in the NonBacktracking implementation... the test is expecting matching to fail but NonBacktracking actually produces a match.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -45,464 +46,505 @@ public static IEnumerable<object[]> Match_Basic_TestData_LazyLoops()
...
+                yield return new object[] { engine, @"[ab\-\[cd-[-[]]]]", "-]]", RegexOptions.None, 0, 3, false, string.Empty };
+                yield return new object[] { engine, @"[ab\-\[cd-[-[]]]]", "`]]", RegexOptions.None, 0, 3, false, string.Empty };
+                yield return new object[] { engine, @"[ab\-\[cd-[-[]]]]", "e]]", RegexOptions.None, 0, 3, false, string.Empty };
+                yield return new object[] { engine, @"[ab\-\[cd-[[]]]]", "']]", RegexOptions.None, 0, 3, false, string.Empty };
+                yield return new object[] { engine, @"[ab\-\[cd-[[]]]]", "e]]", RegexOptions.None, 0, 3, false, string.Empty };
+
+                if (!RegexHelpers.IsNonBacktracking(engine))
+                {
+                    // TODO-NONBACKTRACKING: Triggering an assertion on NonBacktracking
+                    yield return new object[] { engine, @"[a-[a-f]]", "abcdefghijklmnopqrstuvwxyz", RegexOptions.None, 0, 26, false, string.Empty };
```
stephentoub, line 439: @veanes, @olsaarik, this test triggers this assertion in BDD.cs:
https://github.com/dotnet/runtimelab/blob/59ba37522b0f3922c4d899e29710d72fecafde45/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDD.cs#L151
so I've disabled it for now.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -1144,6 +1244,12 @@ public async Task Match_SpecialUnicodeCharacters_Invariant(RegexEngine engine)
         [MemberData(nameof(RegexHelpers.AvailableEngines_MemberData), MemberType = typeof(RegexHelpers))]
         public void Match_ExcessPrefix(RegexEngine engine)
         {
+            if (RegexHelpers.IsNonBacktracking(engine))
+            {
+                // TODO-NONBACKTRACKING: Takes too long and sometimes times out with NonBacktracking
+                throw new SkipTestException("Takes too long and sometimes times out with NonBacktracking");
```
stephentoub, line 1250: @veanes, @olsaarik, this test takes a very long time with NonBacktracking, so I've disabled it for now.

src/libraries/System.Text.RegularExpressions/tests/Regex.Tests.Common.cs
```diff
@@ -103,16 +81,23 @@ public static IEnumerable<RegexEngine> AvailableEngines
...
                     yield return RegexEngine.NonBacktracking;
+
                     if (PlatformDetection.IsReflectionEmitSupported && // the source generator doesn't use reflection emit, but it does use Roslyn for the equivalent
                         PlatformDetection.IsNotMobile &&
                         PlatformDetection.IsNotBrowser)
                     {
                         yield return RegexEngine.SourceGenerated;
+
+                        // TODO-NONBACKTRACKING:
+                        // yield return RegexEngine.NonBacktrackingSourceGenerated;
```
stephentoub, line 92: @olsaarik, I've added in the test hooks for when we augment the regex source generator with NonBacktracking.

src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
veanes: In `NonBacktracking` mode replacement strings like `${day}` are not supported, because  named captures are not supported, but in order to _know_ that, say `${day}` refers to an actual named capture one has to know that  `(?<day> ...)` exists in the regex, else the replacement is simply `${day}` . At some point I added a test like that but that change is perhaps inactive --- it is based on using the property `HasGroups` in the `RegexReplacement` class. If the property is true then `NotSupportedException` should be thrown in `NonBacktracking` mode. It should perhaps be simpler do just throw if `${..}` appears in the replacement string regardless of whether `${..}` is actually valid or not because the validity requires applying a compilation phase to the regex we want to avoid in `NonBacktracking` mode.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: I notice that I've missed to lift some unit tests earlier to `NonBacktracking`. In some cases the answer is different _due to order of alternations_ (here implicit in the loop). Usually I handled the cases individually. Will devote some more time on `Regex.Match.Tests` after merge. In the `NonBacktracking` case it tries to maximize both loops so the outer match in this particular case gives the `"1234"` match. (As opposed to first maximizing the inner loop of 3 steps and then not being able to do another iteration of the outer loop)

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: Oooh, this is **balancing group**, didn't even know this existed until now. **Indeed**, we need to throw `NotSupportedException` for this case, it is a nonregular feature (can be used e.g. to match some context-free expressions like balanced parenthesis)

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: Will investigate after merge.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: Right, here `[a-[a-f]]` will be the empty set, essentially `BDD.False` and some invariant is being violated. My guess is that it is a missed case in minterm generation or similar.

veanes:
Seems I missed some unit test in `Regex.Match.Tests`. Overall really good discoveries, definitely some work cut out for us, but at least the errors seem clear to fix.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCharClass.MappingTable.cs
```diff
@@ -134,7 +132,7 @@ internal sealed partial class RegexCharClass
             new LowerCaseMapping('\u1FFA', '\u1FFB', LowercaseAdd, -126),
             new LowerCaseMapping('\u1FFC', '\u1FFC', LowercaseSet, 0x1FF3),
             new LowerCaseMapping('\u2160', '\u216F', LowercaseAdd, 16),
-            new LowerCaseMapping('\u24B6', '\u24CF', LowercaseAdd, 26),
+            new LowerCaseMapping('\u24B6', '\u24D0', LowercaseAdd, 26),
```
veanes, line 135: It seems these changes do not affect IgnoringCase semantics or do they? If they do then the IgnoreCase BDD table needs regeneration, but this should be detected then by a failing unit test also.

veanes (approved):
My approval applies to all the parts related to Regex files related to `NonBacktracking`. I added some comments to earlier questions and will be looking into them after merge.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCharClass.MappingTable.cs
stephentoub: This is https://github.com/dotnet/runtime/pull/59425, in response to your https://github.com/dotnet/runtime/issues/58958.

src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
stephentoub: > It should perhaps be simpler do just throw if ${..} appears in the replacement string regardless of whether ${..} is actually valid or not

Yeah, I think that will clean up a few things.  I'm not sure why regex goes to all that trouble to treat ones that don't match as regular text.  I can look at some of our data sets to see if this shows up meaningfully.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
olsaarik: This will be fixed in https://github.com/dotnet/runtimelab/pull/1615

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: fixed in #1613

src/libraries/System.Text.RegularExpressions/tests/Regex.KnownPattern.Tests.cs
veanes: Fix in #1616

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: fixed in #1626

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: fixed in #1630

PR #1604, opened 2021-09-28
Fix stack overflow in SymbolicRegexNode.GetFixedLength

stephentoub:
Also removed remaining uses of Array.ConvertAll, which were resulting in unnecessary array allocations and closures.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BVAlgebra.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BVAlgebra.cs
@@ -57,7 +57,7 @@ namespace System.Text.RegularExpressions.Symbolic
         }
 
         public BVAlgebra(CharSetSolver solver, BDD[] minterms) :
-            base(PartitionClassifier.Create(solver, minterms), Array.ConvertAll(minterms, solver.ComputeDomainSize), minterms)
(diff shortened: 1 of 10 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BVAlgebra.cs
```diff
@@ -57,7 +57,7 @@ public ulong ComputeDomainSize(BV set)
         }
 
         public BVAlgebra(CharSetSolver solver, BDD[] minterms) :
-            base(PartitionClassifier.Create(solver, minterms), Array.ConvertAll(minterms, solver.ComputeDomainSize), minterms)
```
veanes, line 60: I did not know that `Array.ConvertAll` should be avoided in general. Now I do.

stephentoub: Everything is relative.  There's nothing inherently wrong with it, but for example in this case when the BVAlgebra is constructed, we're also then allocating a delegate for solver.ComputeDomainSize, and we're paying for minterms.Length delegate invocations in order to invoke ComputeDomainSize for each item.

In other cases in the code, we've got a `List<T>` and are calling `ToArray()` on it just to be able to use `ConvertAll`, when we could instead skip that intermediate array altogether.

veanes (approved):

PR #1605, opened 2021-09-28
Add back "quick" optimization to SymbolicRegexRunner

stephentoub:
Contributes to https://github.com/dotnet/runtime/issues/59629

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexRunner.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexRunner.cs
@@ -57,6 +57,9 @@ namespace System.Text.RegularExpressions
         protected internal Match? runmatch;        // result object
         protected internal Regex? runregex;        // regex object
 
+        // TODO: Expose something as protected internal: https://github.com/dotnet/runtime/issues/59629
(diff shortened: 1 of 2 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexRunner.cs
```diff
@@ -57,6 +57,9 @@ public abstract class RegexRunner
         protected internal Match? runmatch;        // result object
         protected internal Regex? runregex;        // regex object
 
+        // TODO: Expose something as protected internal: https://github.com/dotnet/runtime/issues/59629
+        private protected bool quick;              // false if match details matter, true if only the fact that match occurred matters
```
danmoseley, line 61: I was going to suggest we take the opportunity to name it isMatch or something, but I see it's already in public API.

danmoseley (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexRunner.cs
stephentoub: Yeah. We can still choose to as part of https://github.com/dotnet/runtime/issues/59629, but for now I've kept it consistent with the existing exposed surface area (though I don't know why those methods exist or who would ever want to use them).

PR #1607, opened 2021-09-30
FreeBSD build

wfurt:
replacement for #136.


wfurt:
I'm not sure how to consume & test the outcome yet but at least it builds.

eng/pipelines/runtimelab.yml
```diff
@@ -95,128 +76,6 @@ stages:
         ${{ if eq(variables.isOfficialBuild, true) }}:
           buildArgs: -s clr+libs -c $(_BuildConfig)
```
jkotas: Do you plan to depend on the official builds that get pushed to the runtimelab nuget feed? If yes, I think you will want to build `host` and `installer` as well here.

wfurt: probably yes, but I'm ok taking incremental steps.

wfurt: updated. I'll check how to consume NuGet packages once the official build is done.

eng/pipelines/runtimelab.yml
```diff
@@ -81,9 +62,9 @@ stages:
       buildConfig: Release
       platforms:
       - Linux_x64
```
safern, line 64: Why are we keeping the Linux job?

eng/pipelines/runtimelab.yml
```diff
@@ -81,9 +62,9 @@ stages:
...
-      - windows_x64
+      - FreeBSD_x64
       jobParameters:
-        timeoutInMinutes: 100
+        timeoutInMinutes: 120
         isOfficialBuild: ${{ variables.isOfficialBuild }}
         testGroup: innerloop
         extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```
safern, line 70: Since you are not building tests, you probably want to set the `uploadLibrariesTests` parameter under `extraStepsParameters` to false always. 

You also need to set `uploadIntermediateArtifacts` to `true` for the official build to work.

eng/pipelines/runtimelab.yml
wfurt: We don't have to. For now I wanted to have it as baseline for comparisons and A/B testing (if needed)

eng/pipelines/runtimelab.yml
wfurt: That one I don't know. I would like to get to point where we _can_  run tests. It seems like we now have some test_assets but I did not have time to figure out if that is sufficient to just download and run on FreeBSD. That may be useful even if there is no Helix (yet?).

safern: Ok, I don't mind leaving that as is. However you definitely need to set `uploadIntermediateArtifacts` to `true`.

wfurt: should I set the `uploadIntermediateArtifacts` now? I certainly don't really know what exactly those settings do. 
For me the priority is:
1) get build rolling
2) publish and consume product Nuget package
3) get test running somehow

safern: In order to publish and consume NuGet package you need that setting to `true`. That is what publishes the NuGet packages into an artifact and then we download that artifact and send them to sign/publish.

wfurt: updated. BTW to what feed the packages would go?

wfurt: The intermediates did show up https://dev.azure.com/dnceng/_apis/resources/Containers/8362254/IntermediateArtifacts?itemPath=IntermediateArtifacts%2FFreeBSD_x64%2FShipping%2Fdotnet-runtime-6.0.0-ci-freebsd-x64.tar.gz

jkotas: > BTW to what feed the packages would go?

runtimelab official builds are publishing to https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-experimental

wfurt:
anything else you want me to do @jkotas @safern ???
I'm going to look into running tests as separate change.

jkotas (approved):
Looks fine to me

safern (approved):
LGTM

PR #1613, opened 2021-09-30
detect and throw NotSupported for balancing capture group

veanes:
Fixed a TODO for `NonBacktracking`. 
- Detecting the case when a balancing group` (?<pat1-pat2>...)` is being used and throwing `NotSupportedException`.
- Updated unit tests (removed the related TODO comment and added extra unit test for constructor, to exercise the new code paths)
- Updated the resource file with corresponding error messages

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
@@ -228,7 +228,14 @@ namespace System.Text.RegularExpressions.Symbolic
                     return _builder._bolAnchor;
 
                 case RegexNode.Capture: // treat as non-capturing group (...)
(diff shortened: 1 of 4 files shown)


src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/RegexNodeToSymbolicConverter.cs
```diff
@@ -228,7 +228,14 @@ public SymbolicRegexNode<BDD> Convert(RegexNode node, bool topLevel)
                     return _builder._bolAnchor;
 
                 case RegexNode.Capture: // treat as non-capturing group (...)
-                    return Convert(node.Child(0), topLevel);
+                    if (node.N == -1)
```
stephentoub: This can be:
```C#
case RegexNode.Capture when node.N == -1:
    return Convert(node.Child(0), topLevel); // treat as non-capturing group (...)
```

stephentoub (approved):

PR #1614, opened 2021-10-01
Use newly added rooting mechanism in CoreCLR tests

MichalStrehovsky:
This is better because it will also root any supporting libraries.


jkotas:
Looks like it might have uncovered a new bug.

MichalStrehovsky:
> Looks like it might have uncovered a new bug.

Yeah, I'll need to make another pass at this. I've also been unblocking test without actually validating they work because running tests is annoying to do locally and the CI had too much noise.

PR #1615, opened 2021-10-01
Fix issue with empty character sets

olsaarik:
`GetFixedPrefix` was not working for empty predicates (probably since my iterative rewrite of it). This also enables a test with empty sets that was disabled for NonBackTracking.


veanes (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexNode.cs
```diff
@@ -1504,6 +1504,10 @@ private IEnumerable<S> GetPrefixSequence()
                     // A path didn't have a next set as supported by this algorithm
                     yield break;
                 }
+                if (!_builder._solver.IsSatisfiable(next))
+                {
+                        yield break;
```
stephentoub: Nit:
```suggestion
                    yield break;
```

stephentoub (approved):

PR #1616, opened 2021-10-02
Fix and update replace for NonBackTracking

olsaarik:
This fixes the problem we had of replace for `NonBackTracking` treating all `${...}` patterns as literal strings. They are instead unsupported, regardless of whether they appear in the pattern or not.

I also moved the detection logic to `RegexParser` and made the check otherwise more lenient to allow the following substitutions:

- `$&` and `$0` for the whole match
- ``$` `` for the string before the match
- `$'` for the string after the match
- `$_` for the whole input string

The tests are updated to include tests for these with `NonBackTracking`.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexParser.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexParser.cs
@@ -1443,7 +1443,7 @@ namespace System.Text.RegularExpressions
                 string capname = ScanCapname();
                 if (CharsRight() > 0 && RightCharMoveRight() == '}')
                 {
-                    // Throw unconditionally for backtracking, even if not a valid capture name, as those aren't tracked correctly
+                    // Throw unconditionally for non-backtracking, even if not a valid capture name, as information to determine whether a name is valid or not isn't tracked
                     if ((_options & RegexOptions.NonBacktracking) != 0)
                     {
                         throw new NotSupportedException(SR.NotSupported_NonBacktrackingAndReplacementsWithSubstitutionsOfGroups);
--- a/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexParser.cs
+++ b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexParser.cs
@@ -533,6 +533,9 @@ namespace System.Text.RegularExpressions
                     if (RightCharMoveRight() == '$')
                     {
                         RegexNode node = ScanDollar();
+                        // NonBacktracking does not support capture groups, so any replacement patterns that refer to
+                        // groups are unsupported. The replacement patterns that refer to the left/right portion or all
+                        // of the input as well as referring to group 0, i.e. the whole match, are supported.
                         if ((_options & RegexOptions.NonBacktracking) != 0 && node.Type == RegexNode.Ref &&
                             !(node.M == 0 ||
                               node.M == RegexReplacement.LeftPortion ||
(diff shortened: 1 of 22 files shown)


veanes (approved):

stephentoub:
Can you also delete:
https://github.com/dotnet/runtimelab/blob/58afff25d82dc1d44bb74a763f1d95b102b61347/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L32
https://github.com/dotnet/runtimelab/blob/58afff25d82dc1d44bb74a763f1d95b102b61347/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L127
and change the uses of it in places like:
https://github.com/dotnet/runtimelab/blob/58afff25d82dc1d44bb74a763f1d95b102b61347/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Replace.cs#L80
back to just using capsize?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexParser.cs
```diff
@@ -532,7 +532,16 @@ private RegexNode ScanReplacement()
                 {
                     if (RightCharMoveRight() == '$')
                     {
-                        AddUnitNode(ScanDollar());
+                        RegexNode node = ScanDollar();
+                        if ((_options & RegexOptions.NonBacktracking) != 0 && node.Type == RegexNode.Ref &&
+                            !(node.M == 0 ||
+                              node.M == RegexReplacement.LeftPortion ||
+                              node.M == RegexReplacement.RightPortion ||
+                              node.M == RegexReplacement.WholeString))
```
stephentoub, line 543: Can you add a comment as to why these are special-cased?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexParser.cs
```diff
@@ -1432,9 +1441,17 @@ private RegexNode ScanDollar()
             else if (angled && RegexCharClass.IsWordChar(ch))
             {
                 string capname = ScanCapname();
-                if (CharsRight() > 0 && RightCharMoveRight() == '}' && IsCaptureName(capname))
+                if (CharsRight() > 0 && RightCharMoveRight() == '}')
                 {
-                    return new RegexNode(RegexNode.Ref, _options, CaptureSlotFromName(capname));
+                    // Throw unconditionally for backtracking, even if not a valid capture name, as those aren't tracked correctly
```
stephentoub:
```suggestion
                    // Throw unconditionally for non-backtracking, even if not a valid capture name, as information to determine whether a name is valid or not isn't tracked
```

src/libraries/System.Text.RegularExpressions/tests/Regex.Replace.Tests.cs
```diff
@@ -333,7 +340,7 @@ public void Replace_Invalid()
             if (PlatformDetection.IsNetCore)
             {
                 // Substitutions not supported in DFA mode
-                Assert.Throws<NotSupportedException>(() => new Regex("pattern", RegexHelpers.RegexOptionNonBacktracking).Replace("input", "$0", -1));
+                Assert.Throws<NotSupportedException>(() => new Regex("pattern", RegexHelpers.RegexOptionNonBacktracking).Replace("input", "$1", -1));
```
stephentoub, line 342: This changed because $0 is now supported?  Are there any tests that validate that?

olsaarik: Yes, that's the reason for the change. I'll add a test for $0 directly ($& does parse to the same thing internally, so it does work, but a test will be good to have.)

PR #1620, opened 2021-10-04
Swallow any exceptions out of module initializer

MichalStrehovsky:


jkotas:
Is it what CoreCLR does?

MichalStrehovsky:
CoreCLR will keep throwing TypeInitializationException for anything protected by the cctor. https://github.com/dotnet/runtime/tree/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/tests/Loader/classloader/regressions/vsw529206

jkotas:
It think we should rather throw than to have hard-to-debug silently breakages.

MichalStrehovsky:
> It think we should rather throw than to have hard-to-debug silently breakages.

Works for me. It means that if there's any failing module cctor in the app, we won't even reach `Main()`. But we're initialized enough to at least print the stack trace and blame user code.

PR #1626, opened 2021-10-04
Fixed omitted NonBacktracking option in Match test code

veanes:
In the `Match` test method, in one place the  `NonBacktracking` option was not passed to the static `Regex.Match` and `Regex.IsMatch` methods resulting in the wrong engine being used in the test.

Fixed a TODO, where resulting match value is different in the `NonBacktracking`  engine where nested eager loops allow multiple valid eager matches.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
+++ b/src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
@@ -176,9 +176,14 @@ namespace System.Text.RegularExpressions.Tests
                 yield return new object[] { engine, @"((\d{2,3})){2}", "1234", RegexOptions.None, 0, 4, true, "1234" };
                 if (!RegexHelpers.IsNonBacktracking(engine))
                 {
-                    // TODO-NONBACKTRACKING: Produce "1234" rather than "123", which could be considered an alternative valid result given greedy semantics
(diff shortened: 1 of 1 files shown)


src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -550,10 +555,13 @@ public async Task Match(RegexEngine engine, string pattern, string input, RegexO
...
                 VerifyMatch(r.Match(input));
-                VerifyMatch(Regex.Match(input, pattern, options));
+
+                // Make sure to pass the NonBacktracking option to the static Match method
+                RegexOptions engine_with_options = engine == RegexEngine.NonBacktracking ? options | RegexHelpers.RegexOptionNonBacktracking : options;
+                VerifyMatch(Regex.Match(input, pattern, engine_with_options));
 
                 Assert.Equal(expectedSuccess, r.IsMatch(input));
-                Assert.Equal(expectedSuccess, Regex.IsMatch(input, pattern, options));
+                Assert.Equal(expectedSuccess, Regex.IsMatch(input, pattern, engine_with_options));
```
stephentoub, line 564: This isn't just an issue with NonBacktracking; RegexOptions.Compiled also isn't being factored in.  And these statics don't make sense for the source generator(s).

I think we should address it instead like:
https://github.com/dotnet/runtimelab/pull/1628

PR #1627, opened 2021-10-04
NativeAOT-LLVM: add support for returning ref types (in ShadowStack spill slot)

yowl:
This PR add support for functions that return `REF` types, using the same convention as the IL compiler, i.e return value in a shadow stack slot addressed by the second argument.

Also add string constant symbols (`FrozenStringNode`) as if they were not getting filled if not referenced from the IL->LLVM generation.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
@@ -46,13 +46,23 @@ namespace Internal.JitInterface
                         break;
                     case ReadyToRunHelperId.GetThreadStaticBase:
                         _this._codeRelocs.Add(new Relocation(RelocType.IMAGE_REL_BASED_REL32, 0,
-                            _this._compilation.NodeFactory.TypeThreadStaticsSymbol(target)));
+                            _this._compilation.NodeFactory.TypeThreadStaticIndex(target)));
+                        _this._codeRelocs.Add(new Relocation(RelocType.IMAGE_REL_BASED_REL32, 0,
+                            _this._compilation.NodeFactory.TypeNonGCStaticsSymbol(target)));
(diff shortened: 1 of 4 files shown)


yowl:
```
D:\workspace\_work\1\s\.dotnet\sdk\6.0.100-rc.1.21411.28\Microsoft.Common.CurrentVersion.targets(4606,5): error MSB3026: Could not copy "D:\workspace\_work\1\s\artifacts\obj\System.Security.Principal.Windows\net6.0-windows-Release\System.Security.Principal.Windows.dll" to "D:\workspace\_work\1\s\artifacts\bin\System.Security.Principal.Windows\net6.0-windows-Release\System.Security.Principal.Windows.dll". Beginning retry 1 in 1000ms. The requested operation cannot be performed on a file with a user-mapped section open. : 'D:\workspace\_work\1\s\artifacts\bin\System.Security.Principal.Windows\net6.0-windows-Release\System.Security.Principal.Windows.dll'  [D:\workspace\_work\1\s\src\libraries\System.Security.Principal.Windows\src\System.Security.Principal.Windows.csproj]
```
Should we try to use `handle` from sysinternals to try to find out what causes this problem?

jkotas:
> Should we try to use handle from sysinternals to try to find out what causes this problem?

@MichalStrehovsky worked around it in #1495.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -46,13 +46,23 @@ public static void addCodeReloc(IntPtr thisHandle, void* handle)
...
                         break;
                     default:
                         throw new NotImplementedException();
                 }
+
+                return;
+            }
+
+            var frozenStringNode = node as FrozenStringNode;
+            if (frozenStringNode != null)
```
jkotas:
```suggestion
            if (node is FrozenStringNode frozenStringNode)
```
Nit: Newer C# style

jkotas (approved):
Thanks

PR #1630, opened 2021-10-06
Clean up remaining RegexSRMTests and fixed some todo's

veanes:
- Took care of two leftover TODO's from #1603 
- Cleaned up/generalized and moved remaining SRM unit tests to `Regex.Match.Tests` and all tests that are only run in DEBUG mode reside now in the `RegexExperiment` class.
- Deleted the temporary `RegexSRMTests.cs` file


src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -1371,13 +1377,44 @@ public void Synchronized()
...
+            {
+                yield return new object[] { engine, @"\b\w+\b", "one two three", 3 };
+                yield return new object[] { engine, @"\b\w+\b", "on\u200ce two three", 2 };
+                yield return new object[] { engine, @"\b\w+\b", "one tw\u200do three", 2 };
+            }
+
+            string b1 = @"((?<=\w)(?!\w)|(?<!\w)(?=\w))";
+            string b2 = @"((?<=\w)(?=\W)|(?<=\W)(?=\w))";
+            // Lookarounds are currently not supported in the NonBacktracking engine
+            foreach (RegexEngine engine in new RegexEngine[] { RegexEngine.Compiled, RegexEngine.Interpreter })
```
stephentoub: This will leave out the source generator.  Can you change it to:
```C#
foreach (RegexEngine engine in RegexHelpers.AvailableEngines)
{
    if (engine == RegexEngine.NonBacktracking) continue;
    ...
}
```
or something like that, so that we're only opting-out the engine we know doesn't work?

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -1422,6 +1459,32 @@ public async Task StressTestDeepNestingOfLoops(RegexEngine engine, string begin,
             Assert.True(re.Match(fullinput).Success);
         }
 
+        public static IEnumerable<object[]> StressTestAntimirovMode_TestData()
+        {
+            yield return new object[] { "a.{20}$", "a01234567890123456789", 21 };
+            yield return new object[] { "(a.{20}|a.{10})bc$", "a01234567890123456789bc", 23 };
+        }
```
stephentoub, line 1468:
```suggestion
        }

```

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
```diff
@@ -1621,6 +1684,47 @@ public async Task Match_DisjunctionOverCounting(RegexEngine engine, string disju
...
 
+        public static IEnumerable<object[]> MatchAmbiguousRegexes_TestData()
+        {
+            // Different results in NonBacktracking vs Compiled or Interpreter
+            yield return new object[] { RegexEngine.NonBacktracking, "(a|ab|c|bcd){0,}d*", "ababcd", (0, 6) };
+            yield return new object[] { RegexEngine.NonBacktracking, "(a|ab|c|bcd){0,10}d*", "ababcd", (0, 6) };
+            yield return new object[] { RegexEngine.NonBacktracking, "(a|ab|c|bcd)*d*", "ababcd", (0, 6) };
+            yield return new object[] { RegexEngine.NonBacktracking, @"(the)\s*([12][0-9]|3[01]|0?[1-9])", "it is the 10:00 time", (6, 5) };
+
+            foreach (RegexEngine engine in new RegexEngine[] { RegexEngine.Compiled, RegexEngine.Interpreter })
```
stephentoub: Same comment as earlier; I'd prefer to include all engines and then opt-out the ones to be excluded rather than the other way around so that, for example, we don't accidentally exclude the source generators.

stephentoub (approved):

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: sure, will update.

src/libraries/System.Text.RegularExpressions/tests/Regex.Match.Tests.cs
veanes: will take care of, as above, before merging

PR #1633, opened 2021-10-06
Add asserts for maxBit in BDD range conversions

olsaarik:
The ranges are 32-bit uints and this PR adds Debug.Asserts to check that the given maxBit is in the range from 0 to 32 (inclusive).


veanes (approved):
Maxbit limit should be 31 (inclusive), not 32 (inclusive), i.e., 32 bits: from 0 to 31

olsaarik:
> Maxbit limit should be 31 (inclusive), not 32 (inclusive), i.e., 32 bits: from 0 to 31

Argh, of course. Thank you for pointing that out.

PR #1637, opened 2021-10-07
NativeAOT-LLVM: add some support for floats/doubles/uint/long/byte/ubyte/short/ushort and GT_LE/LT/GE/GT

yowl:
This PR adds some support in the RyuJIT backend for more types and some binary arithmetic comparisons.  

Increases coverage by about 1%.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -258,11 +258,19 @@ llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
             return Type::getInt8Ty(_llvmContext);
 
         case CorInfoType::CORINFO_TYPE_INT:
+        case CorInfoType::CORINFO_TYPE_UINT:
         case CorInfoType::CORINFO_TYPE_NATIVEINT:  // TODO: Wasm64 - what does NativeInt mean for Wasm64
             return Type::getInt32Ty(_llvmContext);
 
+        case CorInfoType::CORINFO_TYPE_LONG:
         case CorInfoType::CORINFO_TYPE_ULONG:
             return Type::getInt64Ty(_llvmContext);
+
+        case CorInfoType::CORINFO_TYPE_FLOAT:
+            return Type::getFloatTy(_llvmContext);
+
+        case CorInfoType::CORINFO_TYPE_DOUBLE:
+            return Type::getDoubleTy(_llvmContext);
 
         case CorInfoType::CORINFO_TYPE_BYREF:
         case CorInfoType::CORINFO_TYPE_CLASS:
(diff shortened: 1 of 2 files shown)


yowl:
Hmm, it "Just worked".  I'll leave the debug in, in case it happens again.

src/coreclr/jit/llvm.cpp
```diff
@@ -853,13 +883,34 @@ void buildCast(llvm::IRBuilder<>& builder, GenTreeCast* cast)
...
 }
 
+void buildCnsDouble(llvm::IRBuilder<>& builder, GenTree* node)
+{
+    if (node->gtType == var_types::TYP_DOUBLE)
+    {
+        mapGenTreeToValue(node, llvm::ConstantFP::get(Type::getDoubleTy(_llvmContext), node->AsDblCon()->gtDconVal));
+        return;
+    }
+    failFunctionCompilation(); // TODO-LLVM: how are floats and decimals done?
```
SingleAccretion: > how are floats and decimals done?

Floats are also stored in `GenTreeCnsDbl`, you just have to check for `TypeIs(TYP_FLOAT)` and downcast `gtDconVal`. Decimals do not exist at the IL level...

src/coreclr/jit/llvm.cpp
```diff
@@ -853,13 +883,34 @@ void buildCast(llvm::IRBuilder<>& builder, GenTreeCast* cast)
...
+                : builder.CreateFPToSI(getGenTreeValue(cast->CastOp()).getValue(builder), getLLVMTypeForVarType(TYP_INT)));
+    }
     else
     {
         // TODO: other casts
         failFunctionCompilation();
     }
 }
 
+void buildCnsDouble(llvm::IRBuilder<>& builder, GenTree* node)
```
SingleAccretion: Use the strongly-typed `GenTreeCnsDbl*`?

src/coreclr/jit/llvm.cpp
```diff
@@ -853,13 +883,34 @@ void buildCast(llvm::IRBuilder<>& builder, GenTreeCast* cast)
...
+{
+    if (node->gtType == var_types::TYP_DOUBLE)
```
```suggestion
    if (node->TypeIs(TYP_DOUBLE))
```

src/coreclr/jit/llvm.cpp

src/coreclr/jit/llvm.cpp
yowl: `ConstantFP::get` handles the downcast, so I left that out.  I did check and the new `TYP_FLOAT` path is exercised.

SingleAccretion: I think you can also remove the `failFunctionCompilation` at this point, and do something like `if (node->TypeIs(TYP_DOUBLE)) else { assert(node->TypeIs(TYP_FLOAT); }`, we will never see a constant of some other type here.

yowl: Thanks, done.

SingleAccretion (approved):

jkotas (approved):
Thanks!

PR #1643, opened 2021-10-09
Merge from dotnet/runtime

jkotas:


jkotas:
@MichalStrehovsky The last two commits are of interest. I think we may see more fallout from the test script refactoring.

MichalStrehovsky:
> @MichalStrehovsky The last two commits are of interest. I think we may see more fallout from the test script refactoring.

Thank you! I've been meaning to switch SmokeTestsOnly to a test filter.

PR #1645, opened 2021-10-09
NativeAOT-LLVM: Add more helper calls (converstions and cast)

yowl:
This PR adds support for more helpers in the clrjit/llvm backend.

There are 3 types of helpers implemented here

1. CORINFO_HELP_DBL2UINT and friends - implemented directly
2. CORINFO_HELP_READYTORUN_STATIC_BASE - a managed call (with shadow stack)
3. CORINFO_HELP_CHKCASTCLASS - a call to the RuntimeExport (which is a native method for LLVM and has no shadow stack)

I'd like to fill these out with the other variant, but before doing that I wanted to ask if there was some way I can determine which helpers are which to avoid  a large `if` statement with lots of `||`s ?  At least for points 2 and 3.  For the conversions it will be necessary to identify them all individually I suspect.

For the `RuntimeExport` calls, the LLVM emits a native to managed thunk which sets up a new shadow stack if it can't find one.  E.g. it starts:
```
define i8* @RhTypeCast_CheckCastClass(%"[S.P.CoreLib]Internal.Runtime.MethodTable"* %0, i8* %1) {
ShadowStackSetupBlock:
  %ShadowStackPtr = alloca i8*, align 4
  %SavedShadowStack = load i8*, i8** @t_pShadowStackTop, align 4
  store i8* %SavedShadowStack, i8** %ShadowStackPtr, align 4
  %ShadowStackNull = icmp eq i8* %SavedShadowStack, null
  br i1 %ShadowStackNull, label %allocateShadowStackBlock, label %ManagedCallBlock

allocateShadowStackBlock:                         ; preds = %ShadowStackSetupBlock
  %NewShadowStack = tail call i8* @malloc(i32 mul (i32 ptrtoint (i8* getelementptr (i8, i8* null, i32 1) to i32), i32 1000000))
  store i8* %NewShadowStack, i8** %ShadowStackPtr, align 4
  br label %ManagedCallBlock
```

The IL generation didn't do this as it had a hard coded mapping to the managed functions in `TypeCast`:
https://github.com/dotnet/runtimelab/blob/d56d19272484f803b4c8739e0ca19482735ca15b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs#L1729-L1741

As it is it works, but I think it would benefit from setting `t_pShadowStackTop` to avoid the `malloc`

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -246,7 +246,8 @@ LocatedLlvmValue getSsaLocalForPhi(unsigned lclNum, unsigned ssaNum)
 }
 
 // Copy of logic from ILImporter.GetLLVMTypeForTypeDesc
-llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType) {
+llvm::Type* getLlvmTypeForCorInfoType(CorInfoType corInfoType)
+{
     switch (corInfoType)
     {
         case CorInfoType::CORINFO_TYPE_VOID:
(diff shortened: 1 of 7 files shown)


yowl:
Maybe its better to fill it out more first so putting to draft

yowl:
Going to close for now as there some stuff to do first to enable some of the helpers, which likely renders some of this obsolete.

PR #1646, opened 2021-10-10
Don't run trimming logic unless we're AOT compiling

MichalStrehovsky:
Also allow opting out of PublishTrimmed if someone really wants to in case it's needed now.

But I think longer term we would want `<Error Condition="PublishTrimmed != true">` instead of setting a default like this.


jkotas (approved):
Thanks!

davidfowl:
Question, is the entry assembly always rooted with native AOT?

MichalStrehovsky:
> Question, is the entry assembly always rooted with native AOT?

We root it, but IL Linker doesn't. I finally sat down to figure out why.

There's this line in the SDK that makes us root it.

https://github.com/dotnet/sdk/blob/8bbcc9a1cc51644e2b784623fd6c28ad98cbae12/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L277-L280

Then there's this in the linker:

https://github.com/dotnet/linker/blob/e47d057dbf17fd92763fd70473aae7cf5287ada9/src/ILLink.Tasks/LinkTask.cs#L40-L50

So TrimmerRootAssembly on something that has entrypoint means "root entrypoint". Which is why linker doesn't really root things.

Sigh.

PR #1649, opened 2021-10-13
NativeAOT-LLVM: Add support for struct types in signatures.

yowl:
This PR replicates some of the logic from IlToLLVMImporter for the creation of LLVM function signatures where a struct is passed.
Its not well tested as most of these methods seem to fail compilation (missing implementation for loading/storing struct fields maybe), but it does look ok as far as it goes:
```
define i32 @S_P_CoreLib_System_Globalization_HebrewCalendar__GetEra(i8* %0, %System.DateTime %1) !dbg !4822 {
Prolog:
  br label %2
```
Also uncovered the following bugs and their fixes are here:

- Fix stack offset bug for locals - the incorrect local was tested to determine how many locals were on the shadow stack
- Fix lcl->llvm arg index when the return value is a parameter.  clrjit compilation is attempting more of some methods that return through `compRetBuffArg`.  These need an adjustment when mapping `lclNum`s to llvm args as the return occupies a slot before the parameters.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -575,7 +575,8 @@ unsigned int padNextOffset(CorInfoType corInfoType, unsigned int atOffset)
 
 /// <summary>
 /// Returns true if the type can be stored on the LLVM stack
-/// instead of the shadow stack in this method.
+/// instead of the shadow stack in this method. This is the case
+/// if it is a non-ref primitive or a struct without GC fields.
 /// </summary>
 bool canStoreLocalOnLlvmStack(LclVarDsc* varDsc)
 {
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -272,39 +272,9 @@ void addPaddingFields(unsigned paddingSize, std::vector<Type*> llvmFields)
     }
 }
 
-// TODO-LLVM: this is a duplicate of GetWellKnownTypeSize in TargetDetails.  If we could get the CORINFO_CLASS_HANDLE for these types, we could probably delete this and rely on compCompHnd->getClassSize
 unsigned getWellKnownTypeSize(CorInfoType corInfoType)
 {
-    switch (corInfoType)
-    {
-        case CorInfoType::CORINFO_TYPE_BOOL:
-            return 1;
-        case CorInfoType::CORINFO_TYPE_CHAR:
-            return 2;
-        case CorInfoType::CORINFO_TYPE_BYTE:
-        case CorInfoType::CORINFO_TYPE_UBYTE:
-            return 1;
-        case CorInfoType::CORINFO_TYPE_USHORT:
-        case CorInfoType::CORINFO_TYPE_SHORT:
-            return 2;
-        case CorInfoType::CORINFO_TYPE_UINT:
-        case CorInfoType::CORINFO_TYPE_INT:
-            return 4;
-        case CorInfoType::CORINFO_TYPE_ULONG:
-        case CorInfoType::CORINFO_TYPE_LONG:
-            return 8;
-        case CorInfoType::CORINFO_TYPE_FLOAT:
-            return 4;
-        case CorInfoType::CORINFO_TYPE_DOUBLE:
-            return 8;
-        case CorInfoType::CORINFO_TYPE_PTR:
-        case CorInfoType::CORINFO_TYPE_NATIVEINT:
-        case CorInfoType::CORINFO_TYPE_NATIVEUINT:
-            return TARGET_POINTER_SIZE;
-    }
-
-    assert(corInfoType == CorInfoType::CORINFO_TYPE_VOID);
-    return TARGET_POINTER_SIZE;
+    return genTypeSize(JITtype2varType(corInfoType));
 }
 
 unsigned getElementSize(CORINFO_CLASS_HANDLE fieldClassHandle, CorInfoType corInfoType)
@@ -378,16 +348,16 @@ llvm::Type* getLlvmTypeForStruct(CORINFO_CLASS_HANDLE structHandle)
 
                 for (unsigned i = 0; i < structSize; i++) sparseFields[i] = nullptr;
 
-                // TODO-LLVM: Are fields already in offset order?  If so we could probably make this more efficient.
                 for (unsigned i = 0; i < fieldCnt; i++)
                 {
                     CORINFO_FIELD_HANDLE fieldHandle = _info.compCompHnd->getFieldInClass(structHandle, i);
-                    unsigned             fldOffset = _info.compCompHnd->getFieldOffset(fieldHandle);
+                    unsigned fldOffset = _info.compCompHnd->getFieldOffset(fieldHandle);
 
                     assert(fldOffset < structSize);
 
                     // store the biggest field at the offset for unions
-                    if (sparseFields[fldOffset] == nullptr || _info.compCompHnd->getClassSize(_info.compCompHnd->getFieldClass(fieldHandle)) > _info.compCompHnd->getClassSize(_info.compCompHnd->getFieldClass(sparseFields[fldOffset])))
+                    if (sparseFields[fldOffset] == nullptr ||
+                        _info.compCompHnd->getClassSize(_info.compCompHnd->getFieldClass(fieldHandle)) > _info.compCompHnd->getClassSize(_info.compCompHnd->getFieldClass(sparseFields[fldOffset])))
                     {
                         sparseFields[fldOffset] = fieldHandle;
                     }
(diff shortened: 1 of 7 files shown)


jkotas:
cc @SingleAccretion

src/coreclr/jit/llvm.cpp
```diff
@@ -245,8 +252,210 @@ LocatedLlvmValue getSsaLocalForPhi(unsigned lclNum, unsigned ssaNum)
...
+    {
+        llvmFields.push_back(Type::getInt32Ty(_llvmContext));
+    }
+    for (unsigned i = 0; i < numBytes; i++)
+    {
+        llvmFields.push_back(Type::getInt8Ty(_llvmContext));
+    }
+}
+
+// TODO-LLVM: this is a duplicate of GetWellKnownTypeSize in TargetDetails.  If we could get the CORINFO_CLASS_HANDLE for these types, we could probably delete this and rely on compCompHnd->getClassSize
```
SingleAccretion: I think this method can simply be `genTypeSize(JITtype2varType(corInfoType))` (I guess with a special case for `VOID` - though it is not clear to me why we need to handle it at all).

src/coreclr/jit/llvm.cpp
```diff
@@ -245,8 +252,210 @@ LocatedLlvmValue getSsaLocalForPhi(unsigned lclNum, unsigned ssaNum)
...
+                    {
+                        llvmType = Type::getInt64Ty(_llvmContext);
+                    }
+                    break;
+                }
+
+            default:
+                // Forward-declare the struct in case there's a reference to it in the fields.
+                // This must be a named struct or LLVM hits a stack overflow
+                const char* name = _info.compCompHnd->getClassName(structHandle);
```
SingleAccretion, line 341: Obligatory note that names of types are not unique, but fine to leave mangling them for later I suppose.

src/coreclr/jit/llvm.cpp
```diff
@@ -245,8 +252,210 @@ LocatedLlvmValue getSsaLocalForPhi(unsigned lclNum, unsigned ssaNum)
...
+                std::vector<CORINFO_FIELD_HANDLE> sparseFields = std::vector<CORINFO_FIELD_HANDLE>(structSize);
+                std::vector<Type*> llvmFields = std::vector<Type*>();
+
+                for (unsigned i = 0; i < structSize; i++) sparseFields[i] = nullptr;
+
+                // TODO-LLVM: Are fields already in offset order?  If so we could probably make this more efficient.
+                for (unsigned i = 0; i < fieldCnt; i++)
+                {
+                    CORINFO_FIELD_HANDLE fieldHandle = _info.compCompHnd->getFieldInClass(structHandle, i);
+                    unsigned             fldOffset = _info.compCompHnd->getFieldOffset(fieldHandle);
```
SingleAccretion: Probably we should give up this alignment style thing, shouldn't we, it is very hard to maintain it without CI constantly reminding one to fix formatting.

src/coreclr/jit/llvm.cpp
```diff
@@ -245,8 +252,210 @@ LocatedLlvmValue getSsaLocalForPhi(unsigned lclNum, unsigned ssaNum)
...
+                llvm::StructType* llvmStructType = llvm::StructType::create(_llvmContext, _info.compCompHnd->getClassName(structHandle));
+                llvmType = llvmStructType;
+                unsigned fieldCnt = _info.compCompHnd->getClassNumInstanceFields(structHandle);
+
+                std::vector<CORINFO_FIELD_HANDLE> sparseFields = std::vector<CORINFO_FIELD_HANDLE>(structSize);
+                std::vector<Type*> llvmFields = std::vector<Type*>();
+
+                for (unsigned i = 0; i < structSize; i++) sparseFields[i] = nullptr;
+
+                // TODO-LLVM: Are fields already in offset order?  If so we could probably make this more efficient.
```
SingleAccretion: (I do not know whether that's the case, presumably it should be easy enough to check it in the managed VM's source)

src/coreclr/jit/llvm.cpp
```diff
@@ -365,20 +577,33 @@ unsigned int padNextOffset(CorInfoType corInfoType, unsigned int atOffset)
 /// Returns true if the type can be stored on the LLVM stack
 /// instead of the shadow stack in this method.
 /// </summary>
```
```suggestion
/// Returns true if the type can be stored on the LLVM stack
/// instead of the shadow stack in this method. This is the case
/// if it is a non-ref primitive or a struct without GC fields.
/// </summary>
```

src/coreclr/jit/llvm.cpp
```diff
@@ -365,20 +577,33 @@ unsigned int padNextOffset(CorInfoType corInfoType, unsigned int atOffset)
...
-bool canStoreTypeOnLlvmStack(CorInfoType corInfoType)
+bool canStoreLocalOnLlvmStack(LclVarDsc* varDsc)
 {
+    CorInfoType corInfoType = toCorInfoType(varDsc->TypeGet());
```
```suggestion
    return !varDsc->HasGCPtr();
```

SingleAccretion: (And delete the rest of the method)

src/coreclr/jit/llvm.cpp
```diff
@@ -414,14 +638,14 @@ FunctionType* getFunctionTypeForSigInfo(CORINFO_SIG_INFO& sigInfo)
     std::vector<llvm::Type*> argVec{Type::getInt8PtrTy(_llvmContext)};
     llvm::Type*              retLlvmType;
 
-    if (needsReturnStackSlot(sigInfo.retType))
+    if (needsReturnStackSlot(sigInfo.retType, sigInfo.retTypeClass)) // TODO-LLVM: retTypeClass or retTypeSigClass?
```
SingleAccretion: The documentation says:
```c
retTypeClass;   // if the return type is a value class, this is its handle (enums are normalized)
retTypeSigClass;// returns the value class as it is in the sig (enums are not converted to primitives)
```
So probably `retTypeClass` is what we want here.

src/coreclr/jit/llvm.cpp
```diff
@@ -245,8 +252,210 @@ LocatedLlvmValue getSsaLocalForPhi(unsigned lclNum, unsigned ssaNum)
...
+                // TODO-LLVM: Are fields already in offset order?  If so we could probably make this more efficient.
+                for (unsigned i = 0; i < fieldCnt; i++)
+                {
+                    CORINFO_FIELD_HANDLE fieldHandle = _info.compCompHnd->getFieldInClass(structHandle, i);
+                    unsigned             fldOffset = _info.compCompHnd->getFieldOffset(fieldHandle);
+
+                    assert(fldOffset < structSize);
+
+                    // store the biggest field at the offset for unions
+                    if (sparseFields[fldOffset] == nullptr || _info.compCompHnd->getClassSize(_info.compCompHnd->getFieldClass(fieldHandle)) > _info.compCompHnd->getClassSize(_info.compCompHnd->getFieldClass(sparseFields[fldOffset])))
```
SingleAccretion: This line is very long, but saving the sizes into locals will cost us 4 VM calls in the common case (which are also slow because of RPI and stuff...), so probably the best thing to do is to move the comparison down a line at least.

src/coreclr/jit/llvm.cpp
```diff
@@ -1228,17 +1463,17 @@ int getLocalOffsetAtIndex(GenTreeLclVar* lclVar)
...
-                if (!canStoreTypeOnLlvmStack(corInfoType))
+                if (!canStoreLocalOnLlvmStack(varDsc))
                 {
                     offset = padNextOffset(corInfoType, offset);
                 }
             }
         }
-        offset = padOffset(localCorInfoType, offset);
+        offset = padOffset(toCorInfoType(lclVar->TypeGet()), offset);
     }
```
SingleAccretion, line 1434:
```suggestion
    }

```

src/coreclr/jit/llvm.cpp
```diff
@@ -1228,17 +1463,17 @@ int getLocalOffsetAtIndex(GenTreeLclVar* lclVar)
...
                 CorInfoType corInfoType = toCorInfoType(varDsc->TypeGet());
-                if (!canStoreTypeOnLlvmStack(corInfoType))
+                if (!canStoreLocalOnLlvmStack(varDsc))
                 {
                     offset = padNextOffset(corInfoType, offset);
                 }
             }
         }
-        offset = padOffset(localCorInfoType, offset);
+        offset = padOffset(toCorInfoType(lclVar->TypeGet()), offset);
```
SingleAccretion, line 1433: I think I've said it somewhere, but will repeat myself that I think the "righter" way to have this is to:
1) Transform all locals that will live on the shadow stack into `IND/OBJ(SHADOW_STACK_ADDRESS_LCL_VAR + OFFSET)` (in a pre-SSA walk).
2) Then perform a layout pass that sets `lvStkOffs`, which the codegen will then ultimately use.

(Not in this change of course)

SingleAccretion (approved):
LGTM modulo notes.

src/coreclr/jit/llvm.cpp
yowl: In other places LLVM will add a suffix when an attempt is made to use the same name, e.g. for locals:
```
%CastPtr47
```
`47` is added.  If we're lucky it will do the same for struct names,

src/coreclr/jit/llvm.cpp
yowl: I'm all for that.  I'll remove it as a I go, starting with this one

src/coreclr/jit/llvm.cpp
yowl: ok, this sorting came from there.  I didn't appreciate, although its obvious now I look at it, that it was the same order used here.

src/coreclr/jit/llvm.cpp
yowl: Yes, you have.  I'll look at this next as I've got similar problems coming up with storing structs with gc refs.

src/coreclr/jit/llvm.cpp
yowl: Thanks, done.

src/coreclr/jit/llvm.cpp
yowl: Thanks, done.

src/coreclr/jit/llvm.cpp
yowl: Added the new line.  Don't know why its not showing as "outdated"

yowl:
Thanks for the review.

yowl:
emscripten cache lock.

jkotas (approved):
Nice

PR #1651, opened 2021-10-13
build platform package without BuildAllConfigurations

wfurt:
creates artifacts/packages/Release/Shipping/Microsoft.NETCore.Platforms.6.0.0-dev.nupkg


safern (approved):
LGTM. Did you diff the package vs the one we generate agains all configurations and made sure there is no diff?

wfurt:
> LGTM. Did you diff the package vs the one we generate agains all configurations and made sure there is no diff?

no. but I can take a look.

PR #1654, opened 2021-10-15
More cleanup in Regex SRM

stephentoub:
- Added comments to classifiers
- Renamed several types and members
- Removed some closure allocations from BDDAlgebra


src/libraries/System.Text.RegularExpressions/src/System.Text.RegularExpressions.csproj
```diff
@@ -59,7 +59,7 @@
     <Compile Include="System\Text\RegularExpressions\Symbolic\BooleanClassifier.cs" />
     <Compile Include="System\Text\RegularExpressions\Symbolic\CharKind.cs" />
     <Compile Include="System\Text\RegularExpressions\Symbolic\DfaMatchingState.cs" />
-    <Compile Include="System\Text\RegularExpressions\Symbolic\PartitionClassifier.cs" />
+    <Compile Include="System\Text\RegularExpressions\Symbolic\MintermClassifier.cs" />
```
danmoseley, line 62: non sequitur - I noticed UnicodeCategoryPredicateName is also dead. or maybe should be #if debug.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
```diff
@@ -59,7 +54,7 @@ internal abstract class BDDAlgebra : IBooleanAlgebra<BDD>
         /// Returns the BDD from the cache if it already exists.
         /// </summary>
         public BDD GetOrCreateBDD(int ordinal, BDD? one, BDD? zero) =>
-            _bddCache.GetOrAdd(new BDDKey(ordinal, one, zero), (key, state) => new BDD(state.ordinal, state.one, state.zero), (ordinal, one, zero));
+            _bddCache.GetOrAdd((ordinal, one, zero), static key => new BDD(key.ordinal, key.one, key.zero));
```
danmoseley, line 57: static lambdas -- TIL. do we need to do a pass over our own tree to see whether any lambdas could be static? (analyzer?)

stephentoub: For reference, `static` doesn't change anything about code generation: with or without the static keyword this code will produce the exact same output.  What it impacts, however, is whether you're allowed to close over anything, so by adding `static`, any references from the lambda to non-statics outside of it will fail to compile.  I like to use `static` then (for anything other than trivial lambdas like `x => x`) as a way of ensuring that we don't regress / that I've successfully removed the closure.

veanes (approved):
I read through all the comments and edits. 
All the explanations are quite detailed and intuitive and correct.

stephentoub:
Excellent, thanks

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
danmoseley: right, that seems to me potentially a good reason to apply it everywhere that currently we're not closing over anything, so that we don't do it accidentally in future. (similar to const?) Worth an analyzer suggestion?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/BooleanClassifier.cs
```diff
@@ -1,62 +1,52 @@
...
                 ascii[i] = domain.Contains(i);
             }
 
-            // Remove the ASCII characters from the domain if the domain is not everything
-            BDD bdd = domain.IsFull ?
-                domain :
-                solver.And(solver._nonascii, domain);
+            // Now, as an optimization, we can remove the ASCII characters from the BDD
+            // as they'll never be used.
+            domain = solver.And(solver._nonAscii, domain);
```
danmoseley: trying to follow.. why remove IsFull check here?

also, should we be renaming domain (to bdd?)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/MintermClassifier.cs
```diff
@@ -0,0 +1,100 @@
...
+            //    patterns don't distinguish between any non-ASCII characters (e.g. "[0-9]*").  If every character
+            //    in the BDD now maps to the same minterm, we can replace the BDD with a much simpler/faster/smaller one.
+            BDD nonAsciiBDD = solver.And(anyCharacterToMintermId, solver._nonAscii);
+            nonAsciiBDD = nonAsciiBDD.IsEssentiallyBoolean(out BDD? singleTerminalBDD) ? singleTerminalBDD : nonAsciiBDD;
+            _nonAscii = nonAsciiBDD;
+        }
+
+        /// <summary>Gets the ID of the minterm associated with the specified character.</summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public int GetMintermID(int c)
```
danmoseley, line 94: You're using `ID` in the method name and `Id` in variable names. Is that intentional? We're completely inconsistent in our own API, but it seems we should be consistent in this library at least.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/BooleanClassifier.cs
stephentoub: Margus and Olli can correct me if I'm wrong, but I believe it's superfluous.  And immediately calls a helper to apply the And operation, and that helper immediately checks to see if either input is equal to True, returning the other if it is.  IsFull is the same as checking for equality to True.  It seemed strange then to just special-case this one condition.

As for renaming domain to bdd, I considered it and didn't end up changing it; happy to change it to bdd or something else if folks prefer.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/MintermClassifier.cs
stephentoub: I'm trying to draw a distinction between the numerical ID for the minterm and the minterm predicate function/bdd itself.  I haven't fully gone through all code doing so, though.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexBuilder.cs
```diff
@@ -67,9 +65,9 @@ internal sealed class SymbolicRegexBuilder<TElement> where TElement : notnull
         private const int InitialStateLimit = 1024;
 
         /// <summary>
-        /// K is the smallest k s.t. 2^k >= atoms.Length + 1
+        /// <see cref="_mintermsCount"/> is the smallest k s.t. 2^k >= minterms.Length + 1
         /// </summary>
-        internal int _K;
+        internal int _mintermsCount;
```
danmoseley, line 70: should we change to properties?

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/BooleanClassifier.cs
veanes: Oh, Dan is right here, I missed this edit, ...good to have multiple people look at stuff. 
We should keep the Full case because it is the trivial True BDD instead of replacing it with the more complicated one here. Usually, unless domain is just True, the replacement after cutting of the ASCII portion gives a smaller BDD, but not when it is just True, the the domai just becomes nonASCII, but we already know it is nonASCII so it would check twice

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Dgml/RegexAutomaton.cs
```diff
@@ -47,7 +47,7 @@ internal RegexAutomaton(SymbolicRegexMatcher<T> srm, int bound, bool addDotStar,
                 _states.Add(_q0.Id);
                 _stateSet.Add(_q0.Id);
 
-                T[]? partition = _builder._solver.GetPartition();
+                T[]? partition = _builder._solver.GetMinterms();
```
danmoseley, line 50: are we still using the term partition?

danmoseley (approved):

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Dgml/RegexAutomaton.cs
stephentoub: I'd like to change them, but I didn't manually edit this file.  I only renamed the functions in the file I was editing, which then renamed all the call sites.

I think you should submit the next comment/renaming PR ;-)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/SymbolicRegexBuilder.cs
stephentoub: No opinion.  We use internal fields all the time.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/BooleanClassifier.cs
veanes: Just to reiterate a bit, after knowing that a character is not ASCII the BDD portion is being used. 
At that point it is typically cheaper to use the solver.And(solver._nonAscii, domain) BDD as the updated domain, 
as opposed to the original domain ... unless the whole domain is actually True (can happen, although rarely)

stephentoub: I'm not following... 

```C#
solver.And(solver._nonAscii, domain)
```
just ends up being a call to:
https://github.com/dotnet/runtimelab/blob/982544a8d226c4edc74fdd38e91778c0d496978f/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs#L74
```C#
ApplyBinaryOp(Op.And, solver._nonAscii, domain)
```
which in turn does:
https://github.com/dotnet/runtimelab/blob/982544a8d226c4edc74fdd38e91778c0d496978f/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs#L94-L120
and specifically:
```C#
                    if (b == True)
                        return a;
```
meaning "if domain == True return _nonAscii".  And `IsFull` is just a comparison against `True` as well:
https://github.com/dotnet/runtimelab/blob/982544a8d226c4edc74fdd38e91778c0d496978f/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDD.cs#L108

So the optimization is still there, no?

veanes: Stephen, regarding 
>It seemed strange then to just special-case this one condition.

note: the operation there was not special casing conjunction (indeed that would not make much sense)
but rather, if the domain == True then keep it "as is" , do not cut it to nonASCII

stephentoub: Thanks, I understand.  My point was, why would the call site only special case IsFull and not every other special case that's in the case BoolOp.And section of ApplyBinaryOp... we can just always call And/ApplyBinaryOp and let any or all of its relevant optimizations kick in.

Am I missing something?

veanes: the classifier is used this way:
if input character is ascii look it up in the array else use the BDD domain.
If the BDD doin is just True (everything, you dont wnat to replace it with non-ASCII that would be 
the result of AND(domain, nonASCII) because AND(True, X) = X, but here we reall just want to keep TRue

stephentoub: > but here we reall just want to keep TRue

I understand. And with my edit we still do.  Am I going crazy?

veanes: In other words: 
`domain = (domain == True ? True : And(domain, nonASCII))`
is **not** the same as 
`domain = And(domain, nonASCII)`

stephentoub: Ok, I've inverted the condition in my head.  Thanks for setting me straight.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Dgml/RegexAutomaton.cs
danmoseley: I figured it's tough to do this concurrently without causing you horrible conflicts - I was just looking out for you 😏
OK , me next then.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Symbolic/Algebras/BDDAlgebra.cs
stephentoub: > similar to const

const does actually affect codegen, e.g. there's a difference between:
```C#
const RegexOptions options = ...;
```
and
```C#
RegexOptions options = ...;
```
The latter is a local.  The former evaporates, with the constant value propagated to call sites.

> Worth an analyzer suggestion?

I don't have major qualms with adding one, and if one existed, I'd probably be fine with us turning it on in dotnet/runtime once we were clean.  It's closer to a style rule, though.  And it does make trivial lambdas more complicated, e.g. presumably an analyzer would flag things like `.OrderBy(x => x)` and say it must be `.OrderBy(static x => x)`, which isn't really helping anything.

PR #1658, opened 2021-10-16
NativeAOT-LLVM: rewrite local vars to shadow stack

yowl:
This PR addresses the comment from https://github.com/dotnet/runtimelab/pull/1649#discussion_r728334303.  It replaces local vars `GT_STORE_LCL_VAR` and `GT_LCL_VAR` with `GT_STORE_OBJ/GT_STOREIND` and `GT_OBJ/GT_IND` where the local is, or contains, gc references.  The replacement stores and load place the local on the shadow stack.  

Coverage is down slightly but not too worried about that as there are a couple of fixes for `TYP_LCLBLK` and `_info.compRetBuffArg != BAD_VAR_NUM` that probably account for it.

If optimisation is on then there is an additional call to `DoPhase(this, PHASE_MARK_LOCAL_VARS, &Compiler::lvaMarkLocalVars);` to get the local weights up to date, then a sort so that higher weighted locals are at the top of the stack.


src/coreclr/jit/compiler.cpp
```diff
@@ -5237,10 +5233,19 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
 #if defined(TARGET_WASM)
     lvaMarkLocalVars();
 
+    if (opts.OptimizationEnabled())
+    {
+        DoPhase(this, PHASE_MARK_LOCAL_VARS, &Compiler::lvaMarkLocalVars);
+    }
```
SingleAccretion: This doesn't look right - we're doing the refcounting twice. I would expect something more like this:
```c
if (opts.OptimizationEnabled())
    lvaMarkLocalVars(); // For optimization only
PlaceAndConvertShadowStackLocals();

lvaMarkLocalVars(); // For SSA i. e. correctness
fgSsaBuild();
```

src/coreclr/jit/compiler.cpp
```diff
@@ -5237,10 +5233,19 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
...
+
+    Llvm* llvm = new Llvm(this);
+    llvm->PlaceAndConvertShadowStackLocals();
```
SingleAccretion: It would be nice to plug `PlaceAndConvertShadowStackLocals` (and any new phases that come up) into the standard `DoPhase` phase mechanism, so that we get dumps at the end automatically.

The simple way to do that would be to use lambdas:
```cpp
auto placeAndConvertShadowStackLocalsPhase = [llvm]() {
    llvm->PlaceAndConvertShadowStackLocals();
}

DoPhase(this, PHASE_SHDWSTK_SETUP, placeAndConvertShadowStackLocalsPhase);
```
(You will also need to add `PHASE_SHDWSTK_SETUP` in `compphases.h`)

This is the "raw" way, without `PhaseStatus`, because there is `ActionPhase` which supports `PhaseStatus`, but it seems fine to have it be that way for the time being.

src/coreclr/jit/llvm.cpp
```diff
@@ -558,9 +537,10 @@ bool canStoreArgOnLlvmStack(CorInfoType corInfoType, CORINFO_CLASS_HANDLE classH
     // structs with no GC pointers can go on LLVM stack.
     if (corInfoType == CorInfoType::CORINFO_TYPE_VALUECLASS)
     {
-        ClassLayout* classLayout = _compiler->typGetObjLayout(classHnd);
+        // Use getClassAtribs over typGetObjLayout because EETypePtr has CORINFO_FLG_GENERIC_TYPE_VARIABLE? which fails with typGetObjLayout
```
SingleAccretion, line 433: Interesting. I wonder how that woks for the other targets (perhaps `EETypePtr` is treated as a "normed struct" there?).

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
     }
 }
 
+
+Llvm::Llvm(Compiler* pCompiler)
```
SingleAccretion, line 1401: Nit: we prefer member initializer lists for simple member initialization:
```cpp
Llvm::Llvm(Compiler* pCompiler)
    : _compiler(pCompiler)
    , _info(pCompiler->info)
    , /* and so on for other fields */
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    _sigInfoIsValid = false;
+}
+
+void Llvm::llvmShutdown()
+{
+    if (_diBuilder != nullptr)
+    {
+        emitDebugMetadata(_llvmContext);
+    }
+#if DEBUG
```
```suggestion
#ifdef DEBUG
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    _module->print(textOutputStream, (llvm::AssemblyAnnotationWriter*)NULL);
+    free(txtFileName);
+#endif //DEBUG
+    llvm::raw_fd_ostream OS(_outputFileName, ec);
+    llvm::WriteBitcodeToFile(*_module, OS);
+    delete _module;
+    //    Module.Verify(LLVMVerifierFailureAction.LLVMAbortProcessAction);
+}
+
+GenTree* createAddNodeForShadowStackLocal(LclVarDsc* varDsc, unsigned shadowStackLclNum, GenTreeIntCon*& offset, GenTreeLclVar*& shadowStackVar)
```
SingleAccretion: I think it would be better to name this function `createShadowStackLocalAddress`, since that's what it does, and also move the insertion code here as well, it seem error-prone to create operands without inserting them into the range (the usual pattern for ranges, FWIW, it to have a "current range" field somewhere and an accessor function for it, so you don't have to pass it everywhere as an argument).

This would also get rid of the `out`-parameters through references, which are forbidden by out code convention (`out` parameters should use pointers).

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+{
+    // TODO-LLVM: if the offset == 0, just GT_STOREIND at the shadowStack
+    offset = _compiler->gtNewOneConNode(var_types::TYP_INT)->AsIntCon();
+    offset->SetIconValue(varDsc->GetStackOffset());
```
SingleAccretion: Why not just:
```cpp
offset = _compiler->gtNewIconNode(varDsc->GetStackOffset(), TYP_I_IMPL);
// TYP_I_IMPL because it is a pointer offset.
```
?

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    shadowStackVar = _compiler->gtNewLclvNode(shadowStackLclNum, var_types::TYP_REF);
```
```suggestion
    shadowStackVar = _compiler->gtNewLclvNode(shadowStackLclNum, TYP_I_IMPL);
```
The shadow stack base itself is not an object reference...

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    return _compiler->gtNewOperNode(genTreeOps::GT_ADD, var_types::TYP_REF, shadowStackVar, offset);
```
```suggestion
    return _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+void insertAddWithOperands(LIR::Range& lirRange, GenTree* node, GenTree* addNode, GenTreeIntCon* offset, GenTreeLclVar* shadowStackVar)
+{
+    lirRange.InsertBefore(node, offset);
+    lirRange.InsertAfter(offset, shadowStackVar);
+    lirRange.InsertAfter(shadowStackVar, addNode);
+}
+
+void Llvm::ConvertShadowStackLocals()
+{
+    unsigned shadowStackLclNum = _compiler->lvaGrabTemp(false, "shadowstack"); // TODO-LLVM: create a new local as a temp, is this right?  Copied from lower.cpp
```
SingleAccretion: > is this right

Yep.
```suggestion
    unsigned shadowStackLclNum = _compiler->lvaGrabTemp(true DEBUGARG("shadowstack"));
```
`true` because it "has long lifetime", though it doesn't actually matter (this was used a very long time ago by Pre-RyuJit "legacy" backend as a register allocator hint).

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+
+    GenTreeIntCon* shadowStackOffset = _compiler->gtNewOneConNode(var_types::TYP_UINT)->AsIntCon(); // LLVM-TODO: TYP_LONG for Wasm64?
+
+    _compiler->eeGetMethodSig(_compiler->info.compMethodHnd, &_sigInfo);
+    _sigInfoIsValid = true;
+
+    shadowStackOffset->SetIconValue(getTotalParameterOffset(_sigInfo));
+
+    // This is a placeholder that we'll replace at LLVM generation as there is no IR Node or Arg
+    GenTreeIntCon* shadowStackArg = _compiler->gtNewOneConNode(var_types::TYP_UINT)->AsIntCon();
```
SingleAccretion: I would like to understand this.

What prevents us from making the shadow stack local refer to the base of the stack for all locals, including arguments? That is sort of the whole point of this, to get rid of handling this in codegen.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+
+    // TODO-LLVM : is GT_ADD the right way to do pointer arithmetic?
+    GenTree* localShadowStack = _compiler->gtNewOperNode(genTreeOps::GT_ADD, var_types::TYP_REF, shadowStackArg, shadowStackOffset);
+
+    GenTreeLclVar* shadowStack = _compiler->gtNewStoreLclVar(shadowStackLclNum, localShadowStack);
+    LclVarDsc* shadowStackVarDsc = _compiler->lvaGetDesc(shadowStack->GetLclNum());
+    shadowStackVarDsc->lvIsParam = 1;
+    shadowStackVarDsc->lvType = var_types::TYP_REF;
```
SingleAccretion: Ditto: the shadow stack should have `TYP_I_IMPL`.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    BasicBlock* firstBlock = _compiler->fgFirstBB;
+    LIR::Range& firstRange = LIR::AsRange(firstBlock);
+    firstRange.InsertAtBeginning(shadowStackOffset);
+    firstRange.InsertAfter(shadowStackOffset, shadowStackArg);
+    firstRange.InsertAfter(shadowStackArg, localShadowStack);
+    firstRange.InsertAfter(localShadowStack, shadowStack);
+
+    for (BasicBlock* const block : _compiler->Blocks())
+    {
+        LIR::Range& lirRange = LIR::AsRange(block);
```
SingleAccretion: Probably the shorter `range` is enough, ranges are a LIR-only concept.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+        for (GenTree* node : LIR::AsRange(block))
```
```suggestion
        for (GenTree* node : range)
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+            {
+                GenTreeLclVarCommon* lclVar = node->AsLclVarCommon();
+                LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar->GetLclNum());
+                if (!varDsc->lvIsParam && !canStoreLocalOnLlvmStack(varDsc))
+                {
+                    GenTreeIntCon* offset;
+                    GenTreeLclVar* shadowStackVar;
+                    GenTree* addNode = createAddNodeForShadowStackLocal(varDsc, shadowStackLclNum, offset, shadowStackVar);
+
+                    genTreeOps oper = lclVar->TypeGet() == var_types::TYP_STRUCT ? GT_STORE_OBJ : GT_STOREIND;
```
```suggestion
                    genTreeOps oper = lclVar->TypeIs(TYP_STRUCT) ? GT_STORE_OBJ : GT_STOREIND;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+            genTreeOps oper = node->OperGet();
+            if (oper == GT_STORE_LCL_VAR)
+            {
+                GenTreeLclVarCommon* lclVar = node->AsLclVarCommon();
+                LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar->GetLclNum());
+                if (!varDsc->lvIsParam && !canStoreLocalOnLlvmStack(varDsc))
+                {
+                    GenTreeIntCon* offset;
+                    GenTreeLclVar* shadowStackVar;
+                    GenTree* addNode = createAddNodeForShadowStackLocal(varDsc, shadowStackLclNum, offset, shadowStackVar);
```
SingleAccretion: (See the comments above about combining `createAddNodeForShadowStackLocal` with `insertAddWithOperands`, also, I would rename `addNode` to `lclAddress` (or something similar)).

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+                    node->ChangeOper(oper);
+                    node->gtFlags |= GTF_IND_TGT_NOT_HEAP;
+                    GenTreeOp* opNode = node->AsOp();
+                    opNode->gtOp2 = opNode->gtGetOp1();
+                    opNode->gtOp1 = addNode;
+
+                    insertAddWithOperands(lirRange, node, addNode, offset, shadowStackVar);
+                }
+            }
+            else if (oper == GT_LCL_VAR)
```
SingleAccretion: This code looks a bit repetitive, I wonder if we could deduplicate it with something like this:
```cpp
if (!node->OperIs(GT_LCL_VAR, GT_STORE_VAR))
{
    continue;
}

GenTreeLvlVarCommon* lclVar = ...;
// The varDsc, the createAddress call, etc

genTreeOps indirOper = GT_NONE:
switch (node->OperGet() { /* choose the new oper */ }

node->ChangeOper(indirOper);
node->AsIndir()->SetAddr(address);

if (GenTree::OperIsStore(indiOper)
{
   node->AsOp()->gtOp2 = storedValue;
}
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+                    insertAddWithOperands(lirRange, node, addNode, offset, shadowStackVar);
+                }
+            }
+        }
+    }
+
+    if (_compiler->verboseTrees)
+    {
+        _compiler->fgDispBasicBlocks(true);
+    }
```
SingleAccretion: This would go away with the proper phase integration.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    _shadowStackLocalsSize = 0;
+    if (_compiler->lvaCount == 0) return;
+
+    std::vector<LclVarDsc*> locals;
+
+    unsigned lclNum;
+    LclVarDsc* varDsc;
+    for (lclNum = 0, varDsc = _compiler->lvaTable; lclNum < _compiler->lvaCount; lclNum++, varDsc++)
+    {
+        if (!varDsc->lvIsParam && !canStoreLocalOnLlvmStack(varDsc))
```
SingleAccretion: Ditto: what prevents us from laying out parameters as well?

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+
+//------------------------------------------------------------------------
+// Convert GT_STORE_LCL_VAR and GT_LCL_VAR to use the shadow stack when the local needs to be GC tracked
+//
+void Llvm::PlaceAndConvertShadowStackLocals()
+{
+    _shadowStackLocalsSize = 0;
+    if (_compiler->lvaCount == 0) return;
+
+    std::vector<LclVarDsc*> locals;
```
SingleAccretion, line 1776: Seems like `locals` are uninitialized here?

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    }
+}
+
+//------------------------------------------------------------------------
+// Convert GT_STORE_LCL_VAR and GT_LCL_VAR to use the shadow stack when the local needs to be GC tracked
+//
+void Llvm::PlaceAndConvertShadowStackLocals()
+{
+    _shadowStackLocalsSize = 0;
+    if (_compiler->lvaCount == 0) return;
```
```suggestion
    if (_compiler->lvaCount == 0)
    {
        return;
    }
```
A rare condition, but presumably can happen.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    LclVarDsc* varDsc;
+    for (lclNum = 0, varDsc = _compiler->lvaTable; lclNum < _compiler->lvaCount; lclNum++, varDsc++)
+    {
+        if (!varDsc->lvIsParam && !canStoreLocalOnLlvmStack(varDsc))
+        {
+            locals.push_back(varDsc);
+        }
+    }
+
+    if (locals.size() == 0) return;
```
```suggestion
    if (locals.size() == 0)
    {
        return;
    }
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1630,190 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+
+    _requiresShadowStackAddSubsitution = true;
+    if (_compiler->opts.OptimizationEnabled())
+    {
+        auto cmpLambda = [](const LclVarDsc* lhs, const LclVarDsc* rhs) { return lhs->lvRefCntWtd() > rhs->lvRefCntWtd(); };
```
SingleAccretion: You could inline this into the call if you want.

src/coreclr/jit/llvm.cpp
```diff
@@ -1487,12 +1452,23 @@ void storeLocalVar(llvm::IRBuilder<>& builder, GenTreeLclVar* lclVar)
     }
 }
 
+void replaceMappedLlvmValueWithShadowStack(GenTree* node)
+{
+    _sdsuMap->at(node) = LocatedLlvmValue(ValueLocation::LlvmStack, _function->getArg(0));
+}
```
SingleAccretion: Ditto: the arguments & shadow stack.

What prevents us from, when seeing the shadow stack local node in the IR, map it to the shadow stack arg?

src/coreclr/jit/llvm.cpp
```diff
@@ -1466,17 +1431,17 @@ void storeLocalVar(llvm::IRBuilder<>& builder, GenTreeLclVar* lclVar)
         LocatedLlvmValue locatedValue{ValueLocation::LlvmStack, nullptr}; // unused
 
         LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar);
-        if (canStoreLocalOnLlvmStack(varDsc))
-        {
-            locatedValue = LocatedLlvmValue(ValueLocation::LlvmStack, valueRef);
-        }
-        else
+        if (varDsc->lvIsParam && !canStoreLocalOnLlvmStack(varDsc))
```
SingleAccretion: Ditto: the arguments & shadow stack.

SingleAccretion (changes requested):
I think we need to figure out how to have the arguments be treated the same as locals (and temps), after all, the whole purpose of this is to remove this handling from codegen (i. e. get rid of `ValueLocation`).

yowl:
Thanks for the review so far.  I've included arguments now and removed the shadow stack setup logic as it's just arg 0 now.  I also moved most of the functions to class member.

I haven't removed `ValueLocation` yet as it's still on use for phi args and for return values.  Phi args can be done some other way, that should be easy.  Return values I'm not sure .  Do you have something I mind?

SingleAccretion:
>  I also moved most of the functions to class member.

(Very nice! Will also allow for getting rid much of the `IRBuilder` passing around)

>  Phi args can be done some other way, that should be easy.

(I've come to think that it will be easiest to not play tricks with them and just record every instance (`GT_PHI` node + the LLVM's `PHINode*`) in an array and then at the end of code generation "fill them in", since the locals map should have all the defs needed at that point)

>  Return values I'm not sure . Do you have something I mind?

Could you elaborate? I see we now return GC types by callee storing them to the shadow stack to an address passed by the caller. Presumably it should be easy enough to transform the following IR:
```scala
SDSU = CALL_REF(SS, ...)
```
To something we are actually going to generate:
```scala
// SS - the shadow stack local
// OFFSET is some reserved address on the shadow stack
CALL_REF(SS, SS + OFFSET, ...)
SDSU = IND(SS + OFFSET)
```
(I also see we're incrementing `OFFSET` currently by maintaining a stack of them - why is that necessary? Presumably callee returning frees the `OFFSET` slot on the shadow stack)
(Also we should transform all arguments passed via the shadow stack to explicit stores to the said stack)

And for the method itself returning, that should also presumably be straightforward:
```scala
RETURN(ref)
```
=>
```scala
STOREIND(RetSpillAddress aka Arg(1), ref)
RETURN(void)
```

src/coreclr/jit/compiler.cpp
```diff
@@ -5235,12 +5231,26 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
     rat.Run();
 
 #if defined(TARGET_WASM)
-    lvaMarkLocalVars();
+    if (opts.OptimizationEnabled())
+    {
+        lvaMarkLocalVars(); // For optimization only
```
```suggestion
        // When optimizing, we'll sort the locals on the shadow stack by ref count.
        lvaMarkLocalVars();
```

src/coreclr/jit/compiler.cpp
```diff
@@ -5235,12 +5231,26 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
...
+    }
+
+    Llvm* llvm = new Llvm(this);
+    auto placeAndConvertShadowStackLocalsPhase = [llvm]() {
+        llvm->PlaceAndConvertShadowStackLocals();
+    };
+    DoPhase(this, PHASE_SHDWSTK_SETUP, placeAndConvertShadowStackLocalsPhase);
 
+    lvaMarkLocalVars();  // For SSA i. e. correctness
```
```suggestion
    lvaMarkLocalVars();  // For SSA.
```
Don't think there is a need to elaborate, the lack of `if (OptimizationEnabled())` implies that already.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+void Llvm::CreateShadowStackLocalAddress(GenTree* node)
+{
+    GenTreeLclVarCommon* lclVar = node->AsLclVarCommon();
+    LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar->GetLclNum());
+    genTreeOps oper = node->OperGet();
+    if (!canStoreLocalOnLlvmStack(varDsc))
+    {
+        // TODO-LLVM: if the offset == 0, just GT_STOREIND at the shadowStack
+        GenTreeIntCon* offset = _compiler->gtNewIconNode(varDsc->GetStackOffset(), TYP_I_IMPL);
+        GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, var_types::TYP_I_IMPL);
```
```suggestion
        GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+        GenTree* lclAddress = _compiler->gtNewOperNode(genTreeOps::GT_ADD, var_types::TYP_I_IMPL, shadowStackVar, offset);
```
```suggestion
        GenTree* lclAddress = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    _module->print(textOutputStream, (llvm::AssemblyAnnotationWriter*)NULL);
+    free(txtFileName);
+#endif //DEBUG
+    llvm::raw_fd_ostream OS(_outputFileName, ec);
+    llvm::WriteBitcodeToFile(*_module, OS);
+    delete _module;
+    //    Module.Verify(LLVMVerifierFailureAction.LLVMAbortProcessAction);
+}
+
+void Llvm::CreateShadowStackLocalAddress(GenTree* node)
```
SingleAccretion: Hmm I think the name has now become a misnomer as this both creates the address and does the rewriting. Proposed (another...) new name: `ConvertShadowStackLocalNode`.

I think this can also take `GenTreeLclVarCommon*` as the argument.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+        GenTree* storedValue = nullptr;
+        switch (node->OperGet())
+        {
+        case GT_STORE_LCL_VAR:
+            indirOper = lclVar->TypeIs(TYP_STRUCT) ? GT_STORE_OBJ : GT_STOREIND;
+            storedValue = node->AsOp()->gtGetOp1();
+            break;
+        case GT_LCL_VAR:
+            indirOper = lclVar->TypeIs(TYP_STRUCT) ? GT_OBJ : GT_IND;
+            break;
```
SingleAccretion: Nit: indentation needs to move one level to the right.

Nit:
```c
default:
    unreached();
```
needs to be added (some Unix compilers warn without it, so we always add it).

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+    LclVarDsc* shadowStackVarDsc = _compiler->lvaGetDesc(_shadowStackLclNum);
+    shadowStackVarDsc->lvIsParam = 1;
+    shadowStackVarDsc->lvType = var_types::TYP_I_IMPL;
+
+    _compiler->eeGetMethodSig(_compiler->info.compMethodHnd, &_sigInfo);
+    _sigInfoIsValid = true;
+
+    for (BasicBlock* const block : _compiler->Blocks())
+    {
+        LIR::Range& currentRange = LIR::AsRange(block);
```
```suggestion
```
Unused.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+        _currentRange = &LIR::AsRange(block);
```
SingleAccretion: I think we can use the pre-existing `_currentBlock` here, it seems unlikely that we'll need to use scratch ranges as "current".

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+
+    unsigned lclNum;
+    LclVarDsc* varDsc;
+    for (lclNum = 0, varDsc = _compiler->lvaTable; lclNum < _compiler->lvaCount; lclNum++, varDsc++)
+    {
+        if (!canStoreLocalOnLlvmStack(varDsc))
+        {
+            locals.push_back(varDsc);
+        }
+    }
```
```suggestion
    for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
    {
        LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
        if (!canStoreLocalOnLlvmStack(varDsc))
        {
            locals.push_back(varDsc);
        }
    }
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+
+void Llvm::ConvertShadowStackLocals()
+{
+    _shadowStackLclNum = _compiler->lvaGrabTemp(true DEBUGARG("shadowstack"));
+    LclVarDsc* shadowStackVarDsc = _compiler->lvaGetDesc(_shadowStackLclNum);
+    shadowStackVarDsc->lvIsParam = 1;
+    shadowStackVarDsc->lvType = var_types::TYP_I_IMPL;
+
+    _compiler->eeGetMethodSig(_compiler->info.compMethodHnd, &_sigInfo);
+    _sigInfoIsValid = true;
```
SingleAccretion: Get the sig info in the constructor? We'll need it anyway, so doesn't seem like lazy initialization is worth it.

src/coreclr/jit/llvm.cpp
```diff
@@ -1654,14 +1490,153 @@ void startImportingNode(llvm::IRBuilder<>& builder)
...
+
+    if (locals.size() == 0)
+    {
+        return;
+    }
+
+    if (_compiler->opts.OptimizationEnabled())
+    {
+        std::sort(locals.begin(), locals.end(), [](const LclVarDsc* lhs, const LclVarDsc* rhs) { return lhs->lvRefCntWtd() > rhs->lvRefCntWtd(); });
+    }
```
SingleAccretion: We'll be sorting all variables... but don't arguments need to be at fixed offsets?

yowl:
> Could you elaborate? I see we now return GC types by callee storing them to the shadow stack to an address passed by the caller. Presumably it should be easy enough to transform the following IR:
> 

> (I also see we're incrementing `OFFSET` currently by maintaining a stack of them - why is that necessary? Presumably callee returning frees the `OFFSET` slot on the shadow stack) (Also we should transform all arguments passed via the shadow stack to explicit stores to the said stack)
> 

In the IL->LLVM generation, the result of the call is placed in a spilled expression entry which is on the expressions stack.  I don't think its guaranteed that that will be popped before the next call.  The spilled expression stack is also used in a few other places, new arrays, exception slots, AddressOf (boxing),.  As there is no expression stack here, and I imagine exception/filters/finallys will be done differently I guess we can just delete the spilled slot stack in `llvm.cpp`

yowl:
> 
> ```scala
> // SS - the shadow stack local
> // OFFSET is some reserved address on the shadow stack
> CALL_REF(SS, SS + OFFSET, ...)
> SDSU = IND(SS + OFFSET)
> ```

I've a problem with generating this, obviously I'm not using the right methods.  I've tried to clone the call, and bash the original 
```scala
            else if (node->IsCall() && needsReturnStackSlot(_sigInfo.retType, _sigInfo.retTypeClass))
            {
                GenTreeCall* callClone = _compiler->gtCloneExpr(node)->AsCall();
                GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
                GenTreeIntCon* offset = _compiler->gtNewIconNode(_shadowStackLocalsSize, TYP_I_IMPL);
                GenTree* retSlot = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
                _compiler->gtPrependNewCallArg(retSlot, callClone->gtCallArgs);

                var_types originalReturnType = callClone->TypeGet();
                callClone->ResetReturnType();
                callClone->gtReturnType = var_types::TYP_VOID;

                CurrentRange().InsertBefore(node, shadowStackVar);
                CurrentRange().InsertAfter(shadowStackVar, offset);
                CurrentRange().InsertAfter(offset, retSlot);
                CurrentRange().InsertAfter(retSlot, callClone);

                node->ChangeOper(originalReturnType == var_types::TYP_STRUCT ? GT_OBJ : GT_IND);
                node->AsOp()->gtOp1 = retSlot;
            }
```
But the resulting IR is not right: the return type has not changed to void, and the extra call arg is missing:
```scala
               [000211] ------------       t211 =    LCL_VAR   int    V14 tmp8         u:1
               [000212] ------------       t212 =    CNS_INT   int    24
                                                  /--*  t211   int
                                                  +--*  t212   int
               [000213] ------------       t213 = *  ADD       int
                                                  /--*  t206   int    arg0 in rcx
                                                  +--*  t207   ref    arg1 in rdx
     ( 20,  8) [000203] --CXG-------       t203 = *  CALL      ref    System.Runtime.TypeCast.CheckCastClass
                                                  /--*  t213   int
N005 ( 20,  8) [000040] --CXG-------        t40 = *  IND       ref
```
Where am I going wrong?

SingleAccretion:
I think in this case `LIR::TryGetUse` will be a better choice - cloning arbitrary calls is a badly tested path in the compiler. Something like:
```scala
// Set up the `retSlot`
// Prepend it to the existing call.
LIR::Use callUse;
if (CurrentRange().TryGetUse(call, &callUse))
{
    // Set up the `retSlot`, again (maybe a better name is returnValueAddress)...
    GenTree* returnedValue = gtNewIndir(retSlot, call->TypeGet());
    callUse.ReplaceWith(_compiler, returnedValue);
}
else
{
    call->ClearUnusedValue();
}

// Calls have *3* "return types" - the descriptor, which is irrelevant for us,
// the node's type and the gtReturnType, which we'll set.
call->gtReturnType = TYP_VOID;
call->ChangeType(TYP_VOID);
```

As to the original snippet, there were a few issues:

> the return type has not changed to void

The `gtType` was not changed.

> extra call arg is missing

Note that you're creating an SDSU, but using it in two places: the call's arg and the indirection. The dump is just showing that there is only one node inserted in the range.

yowl:
Thanks, I'm afraid that I've still not got the adding of the new call arg right.  I have now (created a temp for the 2 references to the return address)
```scala
                // replace the "CALL ref" with a "CALL void" that takes a return address as the first argument
                GenTreeCall* callNode = node->AsCall();
                var_types callReturnType = callNode->TypeGet();
                GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
                GenTreeIntCon* offset = _compiler->gtNewIconNode(_shadowStackLocalsSize, TYP_I_IMPL);
                GenTree* returnValueAddress = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);

                // create temp for the return address
                unsigned returnTempNum = _compiler->lvaGrabTemp(true, "return value address");
                LclVarDsc* returnAddrVarDsc = _compiler->lvaGetDesc(returnTempNum);
                returnAddrVarDsc->lvIsParam = 0;
                returnAddrVarDsc->lvType = var_types::TYP_I_IMPL;

                GenTree* addrStore = _compiler->gtNewStoreLclVar(returnTempNum, returnValueAddress);
                GenTree* returnAddrLcl = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
                
                callNode->gtCallArgs = _compiler->gtPrependNewCallArg(returnAddrLcl, callNode->gtCallArgs);
```
However this hits the assert https://github.com/dotnet/runtimelab/blob/1f26ea2d2d9514f3d9bb4d0b33a123ab2a63b055/src/coreclr/jit/gentree.cpp#L7204
If I add a breakpoint on the `gtPrependNewCallArg` line, there's a couple of things I dont understand about what I'm doing.  There's 2 args to this call `public static unsafe object CheckCastArray(MethodTable* pTargetEEType, object obj)`  an int and the ref.  

`callNode->gtCallArgs` is a nullptr, that sounds odd
image

But there is something in `gtCallLateArgs`.

`fgArgInfo.argCount` == 1:
image

- When, if ever, should I be trying to update `gtCallLateArgs` and not `gtCallArgs`?
- Why is that not 2?
- Do I have to call something to rebuild `fgArgInfo` ?

The IR before I modify it is
```scala
N003 (  3,  2) [000058] ------------        t58 =    LCL_VAR   int    V00 arg0
N004 (  3,  2) [000059] ------------        t59 =    LCL_VAR   ref    V01 arg1
                                                  /--*  t58    int    arg0 in rcx
                                                  +--*  t59    ref    arg1 in rdx
N005 ( 20,  8) [000060] --CXG-------        t60 = *  CALL      ref    System.Runtime.TypeCast.CheckCastInterface
```
Thanks for all the help so far.

SingleAccretion:
I should have seen this coming...

> When, if ever, should I be trying to update gtCallLateArgs and not gtCallArgs?

It's messy. The short answer is updating either beyond morphing is not that well supported a scenario.

> Why is that not 2?

Honestly, I do not know, seems unexpected.

> Do I have to call something to rebuild `fgArgInfo` ?

In theory, you can call `ResetArgInfo` and then `fgInitArgInfo`. But the latter is likely not work properly for us (it does not know about the shadow stack and the spill slot). I suggest you try to it anyway for now, it is not in this PR that calls will be unraveled. If turns out not to work at all, I suggest punting the calls in this PR (and leaving the `LlvmLocation` in for now).

The proper fix for these problems is to rewrite all calls to:
1. Move all "late args" to the normal arg list. The "late args" list is an invention meant to control evaluation order of arguments in HIR, where use order _is_ the evaluation order. LIR has no such constraints. For the LLVM backend, it is doubly useless because we have no complicated register-not-register ABI requirements to abide by (for now at least).
2. Rebuild the arg info, aka just recreate it with the minimal necessary information. It will not be terribly useful anyway once we get rid of the late args.

yowl:
Your suggestion has worked, thanks!.  I've still to try to remove `LLVMLocation` but looks promising.  I just fail the compilation if there are non -late args which seems to occur just for out args (which it doesn't do anyway yet).

yowl:
This is ready now, except for an issue with how to create `GT_STORE_OBJ ` nodes.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+}
+
+
+Llvm::Llvm(Compiler* pCompiler)
+    : _compiler(pCompiler),
+      _info(pCompiler->info),
+      _function(nullptr),
+      _builder(_llvmContext),
+      _prologBuilder(_llvmContext),
+      _shadowStackLclNum(-1),
```
```suggestion
      _shadowStackLclNum(BAD_VAR_NUM),
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+      _retAddressLclNum(-1)
```
```suggestion
      _retAddressLclNum(BAD_VAR_NUM)
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+        CurrentRange().InsertAfter(shadowStackVar, lclAddress);
+    }
+}
+
+void Llvm::ConvertShadowStackLocals()
+{
+    _shadowStackLclNum = _compiler->lvaGrabTemp(true DEBUGARG("shadowstack"));
+    LclVarDsc* shadowStackVarDsc = _compiler->lvaGetDesc(_shadowStackLclNum);
+    shadowStackVarDsc->lvIsParam = 1;
+    shadowStackVarDsc->lvType = var_types::TYP_I_IMPL;
```
```suggestion
    shadowStackVarDsc->lvType = TYP_I_IMPL;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+        {
+            if (node->OperIs(GT_STORE_LCL_VAR, GT_LCL_VAR))
+            {
+                ConvertShadowStackLocalNode(node->AsLclVarCommon());
+            }
+            else if (node->IsCall())
+            {
+                GenTreeCall* callNode = node->AsCall();
+
+                if (callNode->gtCallType == CT_HELPER || callNode->gtReturnType == var_types::TYP_VOID)
```
```suggestion
                if (callNode->IsHelperCall() || callNode->TypeIs(TYP_VOID))
```
`gtReturnType` is only really different for small typed calls.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                    // replace the "CALL ref" with a "CALL void" that takes a return address as the first argument
+                    var_types callReturnType = callNode->TypeGet();
+                    GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
+                    GenTreeIntCon* offset = _compiler->gtNewIconNode(_shadowStackLocalsSize, TYP_I_IMPL);
+                    GenTree* returnValueAddress = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
+
+                    // create temp for the return address
+                    unsigned returnTempNum = _compiler->lvaGrabTemp(true, "return value address");
+                    LclVarDsc* returnAddrVarDsc = _compiler->lvaGetDesc(returnTempNum);
+                    returnAddrVarDsc->lvIsParam = 0;
```
```suggestion
```
No need to set it explicitly as that's the default.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                    returnAddrVarDsc->lvType = var_types::TYP_I_IMPL;
```
```suggestion
                    returnAddrVarDsc->lvType = TYP_I_IMPL;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                        failFunctionCompilation();
+                    }
+                    GenTreeCall::Use* oldArgs = callNode->gtCallLateArgs;
+                    callNode->ResetArgInfo();
+                    callNode->gtCallArgs = _compiler->gtPrependNewCallArg(returnAddrLcl, oldArgs);
+                    callNode->gtCallLateArgs = nullptr;
+                    _compiler->fgInitArgInfo(callNode);
+
+                    GenTree* returnAddrLcl2 = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
+                    GenTree* indirNode = _compiler->gtNewOperNode(callReturnType == var_types::TYP_STRUCT ? GT_OBJ : GT_IND, callReturnType, returnAddrLcl2);
```
```suggestion
                    GenTree* indirNode = _compiler->gtNewOperNode(callReturnType == TYP_STRUCT ? GT_OBJ : GT_IND, callReturnType, returnAddrLcl2);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                if (needsReturnStackSlot(calleeSigInfo.retType, calleeSigInfo.retTypeClass))
+                {
+                    // replace the "CALL ref" with a "CALL void" that takes a return address as the first argument
+                    var_types callReturnType = callNode->TypeGet();
+                    GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
+                    GenTreeIntCon* offset = _compiler->gtNewIconNode(_shadowStackLocalsSize, TYP_I_IMPL);
+                    GenTree* returnValueAddress = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
+
+                    // create temp for the return address
+                    unsigned returnTempNum = _compiler->lvaGrabTemp(true, "return value address");
```
```suggestion
                    unsigned returnTempNum = _compiler->lvaGrabTemp(false DEBUGARG("return value address"));
```
The heuristic here is "do we use the local in multiple blocks?". But actually it does not matter.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                    if (CurrentRange().TryGetUse(callNode, &callUse))
+                    {
+                        callUse.ReplaceWith(_compiler, indirNode);
+                    }
+                    else
+                    {
+                        callNode->ClearUnusedValue();
+                    }
+
+                    callNode->gtReturnType = var_types::TYP_VOID;
```
```suggestion
                    callNode->gtReturnType = TYP_VOID;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                        //    +--*  t393   ref    arg1 in rdx
+                        failFunctionCompilation();
+                    }
+                    GenTreeCall::Use* oldArgs = callNode->gtCallLateArgs;
+                    callNode->ResetArgInfo();
+                    callNode->gtCallArgs = _compiler->gtPrependNewCallArg(returnAddrLcl, oldArgs);
+                    callNode->gtCallLateArgs = nullptr;
+                    _compiler->fgInitArgInfo(callNode);
+
+                    GenTree* returnAddrLcl2 = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
```
SingleAccretion: Not a big fan of numbered locals... `returnAddrLclAfterCall` perhaps?

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                    callNode->gtReturnType = var_types::TYP_VOID;
+                    callNode->ChangeType(TYP_VOID);
+
+                    CurrentRange().InsertBefore(callNode, shadowStackVar);
+                    CurrentRange().InsertAfter(shadowStackVar, offset);
+                    CurrentRange().InsertAfter(offset, returnValueAddress);
+                    CurrentRange().InsertAfter(returnValueAddress, addrStore);
+                    CurrentRange().InsertAfter(addrStore, returnAddrLcl);
+                    CurrentRange().InsertAfter(callNode, returnAddrLcl2);
+                    CurrentRange().InsertAfter(returnAddrLcl2, indirNode);
```
SingleAccretion: We have overloads for the insertion functions going up to 4 arguments, so you could shorten this (applies to other places too):
```
CurrentRange().InsertBefore(callNode, shadowStackVar, offset, returnValueAddress, addrStore);
CurrentRange().InsertAfter(callNode, returnAddrLcl2, indirNode);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                    /* TODO-LLVM: retbuf .   compHasRetBuffArg doesn't seem to have an implementation */
+                    failFunctionCompilation();
+                }
+                //TODO-LLVM: how to create a GT_STORE_OBJ node?
+                if (originalReturnType == TYP_STRUCT)
+                {
+                    failFunctionCompilation();
+                }
+
+                assert(_retAddressLclNum == -1);
```
```suggestion
                assert(_retAddressLclNum == BAD_VAR_NUM);
```
Or probably just have a step before the LLVM phases that initializes the special locals.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+            }
+            else if (node->OperIs(GT_RETURN) && _functionSigHasReturnAddress)
+            {
+                var_types originalReturnType = node->TypeGet();
+                if(node->TypeIs(TYP_VOID))
+                {
+                    /* TODO-LLVM: retbuf .   compHasRetBuffArg doesn't seem to have an implementation */
+                    failFunctionCompilation();
+                }
+                //TODO-LLVM: how to create a GT_STORE_OBJ node?
```
SingleAccretion: It seems we do not have a general helper, however, in this case `gtNewTempAssign` should be sufficient.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                if (originalReturnType == TYP_STRUCT)
+                {
+                    failFunctionCompilation();
+                }
+
+                assert(_retAddressLclNum == -1);
+                _retAddressLclNum = _compiler->lvaGrabTemp(true DEBUGARG("shadowstack"));
+                LclVarDsc* retAddressVarDsc = _compiler->lvaGetDesc(_retAddressLclNum);
+                retAddressVarDsc->lvIsParam = 1;
+                retAddressVarDsc->lvType = var_types::TYP_I_IMPL;
```
```suggestion
                retAddressVarDsc->lvType = TYP_I_IMPL;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+
+                GenTree* storeNode = _compiler->gtNewOperNode(originalReturnType == TYP_STRUCT ? GT_STORE_OBJ : GT_STOREIND, originalReturnType, retAddressLocal, node->AsOp()->gtOp1);
+                storeNode->gtFlags |= GTF_IND_TGT_NOT_HEAP; // No RhpAssignRef required
+
+                GenTreeOp* retNode = node->AsOp();
+                retNode->gtOp1 = nullptr;
+                node->ChangeType(TYP_VOID);
+
+                CurrentRange().InsertBefore(node, retAddressLocal);
+                CurrentRange().InsertAfter(retAddressLocal, storeNode);
```
SingleAccretion: Succinct insert overloads apply.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+//
+void Llvm::PlaceAndConvertShadowStackLocals()
+{
+    _shadowStackLocalsSize = 0;
+    if (_compiler->lvaCount == 0)
+    {
+        return;
+    }
+
+    std::vector<LclVarDsc*> locals;
```
SingleAccretion, line 1611:
```suggestion
    std::vector<LclVarDsc*> locals();
```
Initializing the `locals` local...

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                CurrentRange().InsertBefore(node, retAddressLocal);
+                CurrentRange().InsertAfter(retAddressLocal, storeNode);
+            }
+        }
     }
 }
 
+//------------------------------------------------------------------------
+// Convert GT_STORE_LCL_VAR and GT_LCL_VAR to use the shadow stack when the local needs to be GC tracked
+//
```
```suggestion
//------------------------------------------------------------------------
// Convert GT_STORE_LCL_VAR and GT_LCL_VAR to use the shadow stack when the local needs to be GC tracked,
// rewrite calls that returns GC types to do so via a store to a passed in address on the shadow stack.
// Likewise, store the returned value there if required.
//
```

src/coreclr/jit/compiler.h
```diff
@@ -6098,7 +6098,9 @@ class Compiler
     GenTree* fgMorphArrayIndex(GenTree* tree);
     GenTree* fgMorphCast(GenTree* tree);
     GenTreeFieldList* fgMorphLclArgToFieldlist(GenTreeLclVarCommon* lcl);
+#ifndef TARGET_WASM
```
SingleAccretion: I think it is better to `#ifdef` `public:` here, it is very unusual to see the declarations split like that.

src/coreclr/jit/llvm.cpp
```diff
@@ -1178,113 +1003,67 @@ void buildCmp(llvm::IRBuilder<>& builder, genTreeOps op, GenTree* node, Value* o
...
     }
-
-    // reposition builder at end of block
-    builder.SetInsertPoint(block);
-
-    mapGenTreeToValue(phi, requiresLoad ? ValueLocation::ShadowStack : ValueLocation::LlvmStack, llvmPhiNode);
 }
 
-void addForwardPhiArg(SsaPair ssaPair, llvm::Value* phiArg)
+void Llvm::addForwardPhiArg(SsaPair ssaPair, llvm::Value* phiArg)
```
SingleAccretion: Now unused along with `_forwardReferencingPhis`.

src/coreclr/jit/llvm.cpp
```diff
@@ -1293,84 +1072,83 @@ void addForwardPhiArg(SsaPair ssaPair, llvm::Value* phiArg)
...
-void buildReturn(llvm::IRBuilder<>& builder, GenTree* node)
+void Llvm::buildReturn(GenTree* node)
 {
     switch (node->gtType)
     {
         case TYP_INT:
-            builder.CreateRet(castIfNecessary(builder, getGenTreeValue(node->gtGetOp1()).getValue(builder), getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass)));
+            _builder.CreateRet(castIfNecessary(getGenTreeValue(node->gtGetOp1()), getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass)));
             return;
         case TYP_REF:
```
SingleAccretion: Presumably this is now `unreachable`?

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+        node->AsIndir()->SetAddr(lclAddress);
+        if (GenTree::OperIsStore(indirOper))
+        {
+            node->gtFlags |= GTF_IND_TGT_NOT_HEAP;
+            node->AsOp()->gtOp2 = storedValue;
+        }
+
+        CurrentRange().InsertBefore(node, offset);
+        CurrentRange().InsertAfter(offset, shadowStackVar);
+        CurrentRange().InsertAfter(shadowStackVar, lclAddress);
```
SingleAccretion: Succinct insertion overloads (see comment below).

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1405,298 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                    LclVarDsc* returnAddrVarDsc = _compiler->lvaGetDesc(returnTempNum);
+                    returnAddrVarDsc->lvIsParam = 0;
+                    returnAddrVarDsc->lvType = var_types::TYP_I_IMPL;
+
+                    GenTree* addrStore = _compiler->gtNewStoreLclVar(returnTempNum, returnValueAddress);
+                    GenTree* returnAddrLcl = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
+
+                    if (callNode->gtCallArgs != nullptr)
+                    {
+                        // TODO-LLVM: out args?  E.g.
```
SingleAccretion, line 1527: FWIW, we'll not have them for arguments that LLVM will handle. For arguments that it does not handle (i. e. those that need to be on the shadow stack), we'll `STOREIND/OBJ` them to the said stack before the call.

yowl:
I've removed an assert to enable the creation of `GT_STORE_OBJ`.  Maybe I missed something obvious.  For the vector creation I left off the `()` as it didn't compile.  I left a picture in the review.  Again maybe I've just not understood c++'s intricacies here.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1386,288 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                GenTreeLclVar* retAddressLocal = _compiler->gtNewLclvNode(_retAddressLclNum, TYP_I_IMPL);
+                GenTree* storeNode;
+                if (originalReturnType == TYP_STRUCT)
+                {
+                    storeNode = _compiler->gtNewOperNode(GT_STORE_OBJ, originalReturnType, retAddressLocal, node->AsOp()->gtOp1);
+                    storeNode->AsBlk()->SetLayout(_compiler->typGetObjLayout(_sigInfo.retTypeClass));
+                }
+                else
+                {
+                    storeNode = originalReturnType == TYP_STRUCT
```
SingleAccretion: Hmm, I think you can leave only the `STOREIND` case here, since this is an unreachable branch: `if (originalReturnType == TYP_STRUCT) { if (originalReturnType == TYP_STRUCT) ... } }`.

src/coreclr/jit/llvm.cpp
```diff
@@ -1643,47 +1386,288 @@ llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMet
...
+                _retAddressLclNum = _compiler->lvaGrabTemp(true DEBUGARG("shadowstack"));
+                LclVarDsc* retAddressVarDsc = _compiler->lvaGetDesc(_retAddressLclNum);
+                retAddressVarDsc->lvIsParam = 1;
+                retAddressVarDsc->lvType = TYP_I_IMPL;
+
+                GenTreeLclVar* retAddressLocal = _compiler->gtNewLclvNode(_retAddressLclNum, TYP_I_IMPL);
+                GenTree* storeNode;
+                if (originalReturnType == TYP_STRUCT)
+                {
+                    storeNode = _compiler->gtNewOperNode(GT_STORE_OBJ, originalReturnType, retAddressLocal, node->AsOp()->gtOp1);
```
SingleAccretion: Sorry for drifting in some unclear directions with the `gtNewTempAssign`...

The "normal" way to get a `ExOp` node is to use the construstor, i. e. do `new (_compiler, GT_STORE_OBJ) GenTreeObj(originalReturnType, retAddressLocal, node->AsOp()->gtGetOp1(), layout)` (may necessitate some massaging of the compiler with `::new` or `GenTree::new`.

SingleAccretion (approved):
Just two comments, otherwise looks **great**!

yowl:
Thanks, these 2 bits corrected and the assert reinstated

jkotas:
@SingleAccretion Thank you for reviewing these changes!

PR #1663, opened 2021-10-18
Set RegexOptions.ExplicitCapture when RegexOptions.NonBacktracking is set

stephentoub:
There's also no need to remove the NonBacktracking flag.  Might as well keep for debugging purposes.


stephentoub:
I will close this and re-make it against dotnet/runtime once things are merged there.

PR #1672, opened 2021-10-22
[NativeAOT-LLVM] Improve the build setup a little

SingleAccretion:
1. Delete unused cmake defines from various places.
2. Move the other defines in places where they logically should be.
3. When building the CLR for Wasm, exclude everything but the nativeaot component.
4. Delete the special handling of the Jit targeting Wasm, use the standard altjit configuration. Add a new subset that builds only this Jit, speeding up the innerloop.

The scripts have only been wired up on Windows.


jkotas:
cc @yowl

yowl:
This is great, thanks!

SingleAccretion:
@jkotas, @yowl PTAL

The most controversial change is the sharding of `nativaot.tools` into `nativeaot.ilc` and `nativeaot.build`. I needed it because the ILC is no longer built when you `./build -a wasm`, but instead is built when you build for the host architecture (but we still need the `build` part to run tests and for packaging). A possible workaround would have been to silently exclude ILC from the `tools` subset when targeting Wasm; I felt it was too much of a hack and an unintuitive solution.

@yowl I would like to ask you to pull down these changes and test locally, the very last thing I want is to inadvertently break your workflow in some not-obvious-to-me-from-reading-the-build-infra-code way.

yowl:
This looks good to me.  I cloned, built, and ran the tests, including debugging .  Thanks!

image

image

PR #1677, opened 2021-10-24
[NativeAOT-LLVM]: Struct padding on the shadow stack

yowl:
This PR adds a call back to calculate the padding required for structs on the shadowstack, and removes the forced failures for this scenario.  Also 

- Allows multiple returns
- Changes `OperIsBlkOp ` to explicitly check the operands instead of `Data()  != nullptr` <-  Shall I create a PR to push that upstream?
- For Blk operations sets the layout and `gtBlkOpKind`. I fixed it to  `BlkOpKindHelper`  which is just a guess.

### DIFF
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -7181,7 +7181,7 @@ struct GenTreeCC final : public GenTree
 
 inline bool GenTree::OperIsBlkOp()
 {
-    return ((gtOper == GT_ASG) && varTypeIsStruct(AsOp()->gtOp1)) || (OperIsBlk() && (AsBlk()->Data() != nullptr));
+    return ((gtOper == GT_ASG) && varTypeIsStruct(AsOp()->gtOp1)) || (OperIsBlk() && (OperIs(GT_STORE_DYN_BLK, GT_STORE_BLK, GT_STORE_OBJ)));
 }
 
 inline bool GenTree::OperIsDynBlkOp()
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -7181,7 +7181,7 @@ struct GenTreeCC final : public GenTree
 
 inline bool GenTree::OperIsBlkOp()
 {
-    return ((gtOper == GT_ASG) && varTypeIsStruct(AsOp()->gtOp1)) || (OperIsBlk() && (OperIs(GT_STORE_DYN_BLK, GT_STORE_BLK, GT_STORE_OBJ)));
+    return ((gtOper == GT_ASG) && varTypeIsStruct(AsOp()->gtOp1)) || OperIs(GT_STORE_DYN_BLK, GT_STORE_BLK, GT_STORE_OBJ);
 }
 
 inline bool GenTree::OperIsDynBlkOp()
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -390,11 +390,8 @@ unsigned int Llvm::padOffset(CorInfoType corInfoType, CORINFO_CLASS_HANDLE class
     }
     else
     {
-        if (corInfoType == CorInfoType::CORINFO_TYPE_VALUECLASS)
-        {
-            return _padOffset(_thisPtr, classHandle, atOffset);
-        }
-        failFunctionCompilation();
+        assert(corInfoType == CorInfoType::CORINFO_TYPE_VALUECLASS);
+        return _padOffset(_thisPtr, classHandle, atOffset);
     }
     return roundUp(atOffset, alignment);
 }
@@ -409,20 +406,14 @@ unsigned int Llvm::padNextOffset(CorInfoType corInfoType, CORINFO_CLASS_HANDLE c
     }
     else
     {
-        // TODO-LLvm: LCLBLK size
-        if (corInfoType == CorInfoType::CORINFO_TYPE_VALUECLASS)
-        {
-            size = getElementSize(classHandle, corInfoType);
-        }
-        else
-        {
-            size = TARGET_POINTER_SIZE;
-        }
+        // TODO-LLvm: LCLBLK size?
+        assert(corInfoType == CorInfoType::CORINFO_TYPE_VALUECLASS);
+        size = getElementSize(classHandle, corInfoType);
     }
     return padOffset(corInfoType, classHandle, atOffset) + size;
 }
 
-/// <summary>value type field alignmentvalue type field alignment
+/// <summary>
 /// Returns true if the type can be stored on the LLVM stack
 /// instead of the shadow stack in this method. This is the case
 /// if it is a non-ref primitive or a struct without GC fields.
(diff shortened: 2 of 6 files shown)


src/coreclr/jit/gentree.h
```diff
@@ -7181,7 +7181,7 @@ struct GenTreeCC final : public GenTree
 
 inline bool GenTree::OperIsBlkOp()
 {
-    return ((gtOper == GT_ASG) && varTypeIsStruct(AsOp()->gtOp1)) || (OperIsBlk() && (AsBlk()->Data() != nullptr));
+    return ((gtOper == GT_ASG) && varTypeIsStruct(AsOp()->gtOp1)) || (OperIsBlk() && (OperIs(GT_STORE_DYN_BLK, GT_STORE_BLK, GT_STORE_OBJ)));
```
SingleAccretion:
```suggestion
    return ((gtOper == GT_ASG) && varTypeIsStruct(AsOp()->gtOp1)) || OperIs(GT_STORE_DYN_BLK, GT_STORE_BLK, GT_STORE_OBJ);
```
Since `OperIsBlk` just checks for these same opers.

>  Shall I create a PR to push that upstream?

Not-a-maintainer's-opinion: yes.

src/coreclr/jit/llvm.cpp
```diff
@@ -407,12 +406,11 @@ unsigned int Llvm::padNextOffset(CorInfoType corInfoType, unsigned int atOffset)
     }
     else
     {
-        // TODO: value type field size - this is the ILToLLVMImporter logic:
-        // var size = type is DefType && type.IsValueType ? ((DefType)type).InstanceFieldSize
-        //                                               : type.Context.Target.LayoutPointerSize;
-        failFunctionCompilation(); // TODO value type sizes and alignment
+        // TODO-LLvm: LCLBLK size?
```
SingleAccretion: ...will always be zero as the outgoing args size is calculated during simple lowering.

I think we can just ignore it.

src/coreclr/jit/llvm.cpp
```diff
@@ -1185,11 +1183,11 @@ int Llvm::getLocalOffsetAtIndex(GenTreeLclVar* lclVar)
                 CorInfoType corInfoType = toCorInfoType(varDsc->TypeGet());
                 if (!canStoreLocalOnLlvmStack(varDsc))
                 {
-                    offset = padNextOffset(corInfoType, offset);
+                    offset = padNextOffset(corInfoType, varDsc->lvClassHnd, offset);
```
SingleAccretion: What is the reason for using `lvClassHnd`? It is only used for type optimizations in the compiler and I suspect is not available in a number of cases. I would have expected we would use `GetStructHnd` here.

src/coreclr/jit/llvm.cpp
```diff
@@ -1466,7 +1464,13 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
             node->gtFlags |= GTF_IND_TGT_NOT_HEAP;
             node->AsOp()->gtOp2 = storedValue;
         }
-
+        if (GenTree::OperIsBlk(indirOper))
+        {
+            GenTreeBlk* blk = node->AsBlk();
+            CORINFO_CLASS_HANDLE handle = varDsc->GetStructHnd();
+            blk->SetLayout(_compiler->typGetObjLayout(handle));
+            blk->gtBlkOpKind = GenTreeBlk::BlkOpKindHelper; // TODO-LLVM: dumping the tree requires a valid value, is this ok?
```
SingleAccretion: I would say leave this as `BlkOpKindInvalid` - we won't be using it.

Normally it is set during lowering so that the register allocator knows how many temporary registers are needed for the copy, but we'll have LLVM figure this out by itself.

src/coreclr/jit/llvm.cpp
```diff
@@ -1634,9 +1651,10 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     {
         LclVarDsc* varDsc = locals.at(i);
         CorInfoType corInfoType = toCorInfoType(varDsc->TypeGet());
-        offset = padOffset(corInfoType, offset);
+        CORINFO_CLASS_HANDLE classHandle = corInfoType == CORINFO_TYPE_VALUECLASS ? varDsc->GetStructHnd() : varDsc->lvClassHnd;
```
SingleAccretion: Ditto: `lvClassHnd`. Is there a specific need to pass this down?

src/coreclr/jit/llvm.h
```diff
@@ -174,8 +176,8 @@ class Llvm
     Value* localVar(GenTreeLclVar* lclVar);
     Value* mapGenTreeToValue(GenTree* genTree, Value* valueRef);
     bool needsReturnStackSlot(CorInfoType corInfoType, CORINFO_CLASS_HANDLE classHnd);
-    unsigned int padNextOffset(CorInfoType corInfoType, unsigned int atOffset);
-    unsigned int padOffset(CorInfoType corInfoType, unsigned int atOffset);
+    unsigned int padNextOffset(CorInfoType corInfoType, CORINFO_CLASS_HANDLE classHandle, unsigned int atOffset);
+    unsigned int padOffset(CorInfoType corInfoType, CORINFO_CLASS_HANDLE classHandle, unsigned int atOffset);
```
SingleAccretion, line 172: General question: do you expect we'll leave this "in the VM" in the long term or is this a temporary solution to have the IL/RyuJit - codegens agree with each other? I ask because if we think we should leave this logic to the VM, we should probably think of ways to make transmitting the results cheaper (RPI transitions in loops seem expensive).

src/coreclr/jit/llvm.cpp
yowl: thanks

src/coreclr/jit/llvm.cpp
yowl: Right, thanks

src/coreclr/jit/llvm.cpp
yowl: I'm changing this to be clearer I hope, it will just have the struct class handle.

src/coreclr/jit/llvm.h
yowl: We'll need it long term I think.  The VM defines the layout I believe and we need the alignment and packing size from that.  If we dont use the same layout then I think structs in structs for example, would be a problem.

yowl:
I tidied this up addressing the feedback.  I removed the `spilledExpression` vector, there was nothing in it at present.  It can be put back when/if  required.

src/coreclr/jit/llvm.cpp
```diff
@@ -375,8 +378,12 @@ CorInfoType Llvm::toCorInfoType(var_types varType)
     }
 }
 
+CORINFO_CLASS_HANDLE Llvm::tryGetStructClassHandle(LclVarDsc* varDsc)
+{
+    return toCorInfoType(varDsc->TypeGet()) == CorInfoType::CORINFO_TYPE_VALUECLASS ? varDsc->GetStructHnd() : nullptr;
```
SingleAccretion, line 383:
```suggestion
    return varTypeIsStruct(varDsc) ? varDsc->GetStructHnd() : NO_CLASS_HANDLE;
```

SingleAccretion (approved):
LGTM, modulo one nit - feel free to address separately (to not spin up the CI again).

src/coreclr/jit/llvm.cpp
yowl: Thanks, I'll get that in the next PR.

yowl:
@jkotas   I'll get the last suggestion in the next PR.  Pending comments from yourself and others this is ok to go.

jkotas (approved):
@yowl @SingleAccretion It is a pleasure to see you working together on this!

Issue #776, opened 2021-03-05
Detect infinite generic expansion


hez2010:
*Previous discussion: https://github.com/dotnet/corert/issues/363*

We need to detect infinite generic expansion and terminate type loading.

Affected test:
```
JIT\Methodical\inlining\bug505642\test\
```

.NET Native for UWP apps uses [Tarjan's algorithm](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm) to find strongly connected components in the graph of constructed types.

The task is to:

- Figure out how to run the algorithm in parallel with code generation (in .NET Native for UWP apps, this is a global pre-pass, but we don't have global pre-passes here)
- Convert static generic lookups to lazy lookups to break the connected component. Lazy lookups are not wired up yet, so maybe that should be the first thing to do.

Here's a couple test cases: [recursion.zip](https://github.com/dotnet/runtimelab/files/6091566/recursion.zip). We would be able to solve the ones where the recursion happens over reference types. If the recursion happens over value types, this is not solvable without universal shared code.

MichalStrehovsky:
This is mostly done but I would like to avoid the spurious `NullReferenceException`s when cutoff happens, and I would like the compiler to print some sort of useful feedback when the cutoff happens so that people can ideally remove the generic cycle from the source.

MichalStrehovsky:
Fixed in #1681, #1699, #1709, #1710.

PR #1681, opened 2021-10-27
Introduce detection of generic cycles

MichalStrehovsky:
Motivated by https://github.com/npgsql/npgsql/issues/4057. The generic recursion that was in Npgsql made its way back into Npgsql. We've seen these show up too often. Not because it's common to write recursions in C#, but because popular NuGets end up having these and then people hit it too often.

This brings over the generic cycle detection from .NET Native and hooks it up in a rudimentary way. The pull request is split into logical commits for better reviewability.

We detect generic recursion at two spots:
* In the code generation phase. This detects all the cases where shared generics are not involved.
* In generic dictionaries. This is the more typical case where recursion results in never-ending expansion in terms of types and generic dictionaries.

If we detect a recursion, we unceremoniously cut it off after a couple levels of nesting. The cutoff point is fairly low because in the cases I’ve seen, the recursion also causes an expansion in breadth, not just depth, and allowing that to go unchecked was resulting in sadness. If the code goes beyond the cutoff point at runtime, it will not work.

The recursion detector is essentially a prepass – it goes over everything in the assembly, trying to look for recursions and generating a list of types/generic method involved in a cycle. I’m fairly convinced it cannot be done in other ways than with a prepass.

Looking at the .NET Native code, we had quite a few limitations where generic recursion would not be detected. Given a failure to detect a cycle results in a compiler failure, and we’ve not heard of this failure over the years, the .NET Native approach is likely sufficient.

## What’s missing

If we cutoff the generic recursion within a generic dictionary, the failure mode at runtime is a NullRef shortly after a dictionary lookup (because the cut off slot will be null). It’s not great. I think we can update the lookup helpers to check for null while restricting the check to only the helpers where we’ve seen a damaged slot. We also need to make sure such lookups never get inlined by RyuJIT and always go through our helpers. I think it’s doable and not hard. I just don’t want to put more in this pull request than what’s strictly necessary.

Lazy generics. If the recursion happens over reference types, we could potentially generate fully working code by failing back to lazy generics. The cases I was looking at were not over reference types.

We might want to allow specifying the cutoff point at compile time (also allow disabling all of this).

## Perf

I've seen a 2% regression in compilation throughput for WebApi template. We can make this more efficient by e.g. not hydrating type system objects for everything when we're scanning for cycles and staying with S.R.Metadata for longer.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/LazyGenerics/Graph.Cycles.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/LazyGenerics/Graph.Cycles.cs
@@ -1,0 +1,390 @@
+namespace Microsoft.Build.ILTasks.Transforms
+{
+    using System;
+    using System.IO;
+    using System.Text;
+    using System.Linq;
+    using System.Collections;
+    using System.Diagnostics;
+    using System.Collections.Generic;
+
+    internal static partial class LazyGenericsSupport
+    {
+        private const long s_previousAlgorithmTimeout = 10000;
+
+        private sealed partial class Graph<P>
+        {
+            class TarjanWorkerClass
+            {
+                Stack<Vertex> _inProgress = new Stack<Vertex>();
+                int _currentComponentIndex = 0;
+                Vertex[] _vertices;
+                List<List<Vertex>> _result = new List<List<Vertex>>();
+
+                public TarjanWorkerClass(Vertex[] vertices, bool iterative)
+                {
+                    this._vertices = vertices;
+
+                    foreach (Vertex vertex in vertices)
+                    {
+                        vertex.Index = -1;
+                        vertex.LowLink = -1;
+                        vertex.OnStack = false;
+                    }
+
+                    foreach (Vertex vertex in vertices)
+                    {
+                        if (vertex.Index == -1)
+                        {
+                            if (iterative)
+                            {
+                                StrongConnectIterative(vertex);
+                            }
+                            else
+                            {
+                                StrongConnectRecursive(vertex);
+                            }
+                        }
+                    }
+                }
+
+                void StrongConnectRecursive(Vertex vertex)
+                {
+                    vertex.Index = _currentComponentIndex;
+                    vertex.LowLink = _currentComponentIndex;
+
+                    _currentComponentIndex++;
+
+                    _inProgress.Push(vertex);
+                    vertex.OnStack = true;
+
+                    foreach (Edge edge in vertex.Edges)
+                    {
+                        if (edge.Destination.Index == -1)
+                        {
+                            // Recurse if the destination has not begun processing yet
+                            StrongConnectRecursive(edge.Destination);
+                            vertex.LowLink = Math.Min(vertex.LowLink, edge.Destination.LowLink);
+                        }
+                        else if (edge.Destination.OnStack)
+                        {
+                            vertex.LowLink = Math.Min(vertex.LowLink, edge.Destination.Index);
+                        }
+                    }
+
+                    // If v is a root node, pop the stack and generate an SCC
+                    if (vertex.LowLink == vertex.Index)
+                    {
+                        List<Vertex> newStronglyConnectedComponent = new List<Vertex>();
+                        Vertex poppedVertex = null;
+                        while (poppedVertex != vertex)
+                        {
+                            poppedVertex = _inProgress.Pop();
+                            poppedVertex.OnStack = false;
+
+                            newStronglyConnectedComponent.Add(poppedVertex);
+                        }
+
+                        _result.Add(newStronglyConnectedComponent);
+                    }
+                }
+
+                struct StrongConnectStackElement
+                {
+                    public Vertex Vertex;
+                    public IEnumerator<Edge> EdgeEnumeratorPosition;
+                }
+
+                private Stack<StrongConnectStackElement> IterativeStrongConnectStack = new Stack<StrongConnectStackElement>();
+
+                void StrongConnectIterative(Vertex vertex)
+                {
+                    IEnumerator<Edge> currentEdgeEnumerator = null;
+
+StartOfFunctionWithLogicalRecursion:
+                    vertex.Index = _currentComponentIndex;
+                    vertex.LowLink = _currentComponentIndex;
+
+                    _currentComponentIndex++;
+
+                    _inProgress.Push(vertex);
+                    vertex.OnStack = true;
+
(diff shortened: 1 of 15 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/LazyGenerics/Graph.Cycles.cs
```diff
@@ -0,0 +1,390 @@
...
+                                FindCyclesWorker(startingEdge, new List<Edge>(), ref operationCount, previousAlgorithmTimeoutWatch);
+                            }
+                        }
+                    }
+
+                    if (previousAlgorithmTimeoutWatch.ElapsedMilliseconds > s_previousAlgorithmTimeout)
+                    {
+                        abortedDueToTimeout = true;
+                        break;
+                    }
```
MichalStrehovsky, line 287: We might not want this for determinism. I've not seen things go anywhere near the timeout. Even in unoptimized debug builds most assemblies are analyzed within 2 seconds.

MichalStrehovsky:
With the latest commit, the impact on compiler throughput is within noise range. The largest assembly in WebApi gets analyzed in 88 ms. Most assemblies are in single-digit ms range, a lot even at 0. We do pretty much all of this work in the multithreaded phase.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/LazyGenerics/Graph.Cycles.cs
jkotas: So delete it for now and see whether it is going to show up as actual problem?

jkotas: We may want to just print a message to verbose logger, so that it is easy to tell that the compilation is struck in this recursion detector.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/LazyGenerics/ModuleCycleInfo.cs
```diff
@@ -0,0 +1,189 @@
...
+                ModuleCycleInfo cycleInfo = _hashtable.GetOrCreateValue(ownerModule);
+                if (cycleInfo.FormsCycle(ownerDefinition, referentDefinition))
+                {
+                    // Just the presence of a cycle is not a problem, but once we start getting too deep,
+                    // we need to cut our losses.
+                    if (cycleInfo.IsDeepPossiblyCyclicInstantiation(referent))
+                    {
+                        if (referentType != null)
+                        {
+                            // TODO: better exception string ID?
```
jkotas, line 176: This is more part of the compiler, not part of the core type system. I think we can just use regular logger to communicate the problem here, and also log the types and method involved in the cycle here.

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/LazyGenerics/Graph.Cycles.cs
MichalStrehovsky: I looked at this a bit more and the "previous algorithm" this is referring to is the original algorithm .NET Native used. .NET Native originally had a naive implementation of this that would take a really long time. Then David changed it to use Tarjan's algorithm. He kept the original algorithm and in DEBUG builds we run both and compare the results. So all of the timeouts are under `#if DEBUG`. It's probably fine to just keep this or delete it all. I don't have a strong opinion either way. It's probably easier to see why the cycle formed in the naive algorithm.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/LazyGenerics/ModuleCycleInfo.cs
MichalStrehovsky: This exception will still be rethown at runtime when the cutoff code is actually reached, so we need a string ID because we might end up generating throwing code out of this. Throwing a type system exception makes all of this fall out of it nicely because we have all the infrastructure to do that.

I'll revisit this when I work on a nicer experience around this.

MichalStrehovsky:
Cc @yowl - the change in CorInfoImpl.RyuJit.cs (or the equivalent change in ILImporter.Scanner.cs) is relevant to WASM as well - without the extra call to check if we're forming a cycle, the new tests in Generics.cs will run out of memory at compile time trying to compile an endless generic recursion.

hez2010:
With above changes there're still some tests failing, see https://github.com/dotnet/runtimelab/issues/776 for test cases.

- `GenericExpansionWithGenericMethod`: Arg_NullReferenceException (if I reduce the expansion level, it will yield NotSupported_SubclassOverride)
- `GenericExpansionWithInterfaces`: NotSupported_SubclassOverride
- `GenericExpansionWithStructsViaDirectMethodCall`: compilation never finish

MichalStrehovsky:
> With above changes there're still some tests failing, see #776 for test cases.

Thanks for checking! Yeah, I intentionally didn't make this resolve #776 - this is just the bare minimum. I expect 2-3 more pull requests before it gets in a shape that I'll be happy with.

MichalStrehovsky:
Are you running the tests in reflection disabled mode? The tests do use reflection and reflection disabled mode didn't bother overriding non-abstract methods on System.Type, so that would explain NotSupported_SubclassOverride.

The NullReferenceException is expected because _"If we cutoff the generic recursion within a generic dictionary, the failure mode at runtime is a NullRef shortly after a dictionary lookup (because the cut off slot will be null). It’s not great. I think we can update the lookup helpers to check for null while restricting the check to only the helpers where we’ve seen a damaged slot. We also need to make sure such lookups never get inlined by RyuJIT and always go through our helpers. I think it’s doable and not hard. I just don’t want to put more in this pull request than what’s strictly necessary."_).

hez2010:
> Are you running the tests in reflection disabled mode?

Yes. I forgot to enable reflection while testing :D

yowl:
> 
> 
> Cc @yowl - the change in CorInfoImpl.RyuJit.cs (or the equivalent change in ILImporter.Scanner.cs) is relevant to WASM as well - 

Alternatively I suppose if the methods in questions were compiled with the new RyuJit process they would get the benefit of your changes in `CorInfoImpl.RyuJit.cs`?

MichalStrehovsky:
> Alternatively I suppose if the methods in questions were compiled with the new RyuJit process they would get the benefit of your changes in CorInfoImpl.RyuJit.cs?

Yup, that should just fall out there!

PR #1682, opened 2021-10-27
[NativeAOT-LLVM] Fix packaging

SingleAccretion:
It appears that the packaging code has bit-rotted a little, so fix that, by deleting code duplication.


jkotas:
@yowl PTLA

yowl:
Looks good, thanks!

Issue #1640, opened 2021-10-08
NativeAOT-LLVM: Question: CT_HELPER calls and getMethodSig


yowl:
Is there something about helper methods, e.g. `CORINFO_HELP_READYTORUN_STATIC_BASE` that means that I can't call 
```
_compiler->eeGetMethodSig(call->gtCallMethHnd, &sigInfo);
```
 for them?

I want to generate the function signature for LLVM.  Maybe I should just use the args on the `GenTreeCall`?

yowl:
I tried and the `handle` is much lower than the `handleBase` for
```
        private Object HandleToObject(IntPtr handle)
        {
#if DEBUG
            handle = new IntPtr(~(long)s_handleHighBitSet & (long) handle);
#endif
            int index = ((int)handle - handleBase) / handleMultipler;
            return _handleToObject[index];
        }
```
In `CorInfoImpl.cs`

jkotas:
JIT helpers have fake `METHOD_HANDLEs` inside the JIT. These fake method handles do not work when passed to JIT/EE interface.

yowl:
Thanks

Issue #839, opened 2021-03-21
NativeAOT-LLVM: IR: CALL help byref  HELPER.CORINFO_HELP_READYTORUN_STATIC_BASE


yowl:
I'm looking to extend the IR support to cover 
https://github.com/dotnet/runtimelab/blob/0821949839aefc3cb69752d2cbbdaf64d33c3863/src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs#L646-L649
This is pretty trivial IL, although expands to cover a few hidden extras as can be seen from the current LLVM:
```
define void @HelloWasm_Program__KillParent(i8* %0) !dbg !483989 {
Prolog:
  br label %Block0

Block0:                                           ; preds = %Prolog
  call void @llvm.donothing(), !dbg !483990
  %LoadAddressOfSymbolNode = load i32*, i32** @__GCStaticBase_HelloWasm_Program___SYMBOL, align 4, !dbg !483991
  %castBasePtrPtr = bitcast i32* %LoadAddressOfSymbolNode to i8***, !dbg !483991
  %basePtr = load i8**, i8*** %castBasePtrPtr, align 4, !dbg !483991
  %base = load i8*, i8** %basePtr, align 4, !dbg !483991
  %LoadAddressOfSymbolNode1 = load i32*, i32** @__NonGCStaticBase_HelloWasm_Program___SYMBOL, align 4, !dbg !483991
  %classConstructionContext = getelementptr i32, i32* %LoadAddressOfSymbolNode1, i32 -2, !dbg !483991
  %1 = getelementptr i8, i8* %0, i32 0, !dbg !483991
  %CastPtrclassConstructionContext = bitcast i32* %classConstructionContext to %"[S.P.CoreLib]System.Runtime.CompilerServices.StaticClassConstructionContext"*, !dbg !483991
  call void @S_P_CoreLib_System_Runtime_CompilerServices_ClassConstructorRunner__EnsureClassConstructorRun(i8* %1, %"[S.P.CoreLib]System.Runtime.CompilerServices.StaticClassConstructionContext"* %CastPtrclassConstructionContext), !dbg !483991
  %aParent_addr = getelementptr i8, i8* %base, i32 4, !dbg !483991
  %shadowStackTop = getelementptr i8, i8* %0, i32 0, !dbg !483991
  store i8* %shadowStackTop, i8** @t_pShadowStackTop, align 4, !dbg !483991
[...]

PR #1683, opened 2021-10-28
[NativeAOT-LLVM] : Add support for more helpers

yowl:
This PR removes the restriction on which helper calls are attempted.  There are still 3 excluded, `CORINFO_HELP_READYTORUN_DELEGATE_CTOR`, `CORINFO_HELP_READYTORUN_GENERIC_HANDLE` and `CORINFO_HELP_READYTORUN_GENERIC_STATIC_BASE`. They are not handled in https://github.com/dotnet/runtimelab/blob/cc4ed47fd95047921b2703286526926e8a1f8b54/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/JitHelper.cs#L18

It adds a truncate when creating an `int` local from a `long`.

Include all the helpers from `MathHelpers.cpp`  is that ok?

Adds a few more conversions and floating point casts.  

I changed `LLVM` to `Llvm` to match the class name, sorry for that churn.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1035,10 +1035,13 @@ void Llvm::buildCast(GenTreeCast* cast)
     }
     else if (cast->TypeIs(TYP_LONG) && genActualTypeIsInt(cast->CastOp()))
     {
+        // Cast pointer to int if necessary.  TODO-LLVM: candidate for lowering?
+        Value* sourceValue = castIfNecessary(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(cast->CastOp()->TypeGet()));
+
         mapGenTreeToValue(cast,
             cast->IsUnsigned()
-            ? _builder.CreateZExt(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(cast->CastToType()))
-            : _builder.CreateSExt(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(cast->CastToType())));
+            ? _builder.CreateZExt(sourceValue, getLlvmTypeForVarType(cast->CastToType()))
+            : _builder.CreateSExt(sourceValue, getLlvmTypeForVarType(cast->CastToType())));
     }
     else if (cast->TypeIs(TYP_INT, TYP_LONG) && cast->CastOp()->TypeIs(TYP_FLOAT, TYP_DOUBLE))
     {
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -852,7 +852,7 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
     return mapGenTreeToValue(call, llvmCall);
 }
 
-llvm::FunctionType* Llvm::buildHelperLlvmFunctionType(GenTreeCall* call, bool withShadowStack)
+FunctionType* Llvm::buildHelperLlvmFunctionType(GenTreeCall* call, bool withShadowStack)
 {
     Type* retLlvmType = getLlvmTypeForVarType(call->TypeGet());
     std::vector<llvm::Type*> argVec;
@@ -910,6 +910,7 @@ void Llvm::buildHelperFuncCall(GenTreeCall* call)
         OperandArgNum* sortedData = sortedArgs.data();
         bool requiresShadowStack = helperRequiresShadowStack(call->gtCallMethHnd);
 
+        //TODO-LLVM: refactor calling code with user calls.
         for (unsigned i = 0; i < argCount; i++)
         {
             fgArgTabEntry* curArgTabEntry = argTable[i];
@@ -976,14 +977,15 @@ void Llvm::buildCall(GenTree* node)
     if (call->gtCallType == CT_HELPER)
     {
         buildHelperFuncCall(call);
-        return;
     }
     else if (call->gtCallType == CT_USER_FUNC && !call->IsVirtualStub() /* TODO: Virtual stub not implemented */)
     {
         buildUserFuncCall(call);
-        return;
     }
-    failFunctionCompilation();
+    else
+    {
+        failFunctionCompilation();
+    }
 }
 
 void Llvm::buildCast(GenTreeCast* cast)
@@ -998,14 +1000,14 @@ void Llvm::buildCast(GenTreeCast* cast)
     {
         mapGenTreeToValue(cast, _builder.CreateFPCast(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(TYP_DOUBLE)));
     }
-    else if ((castToType == TYP_LONG || castToType == TYP_ULONG) && cast->CastOp()->TypeIs(TYP_INT, TYP_UINT))
+    else if (cast->TypeIs(TYP_LONG) && genActualTypeIsInt(cast->CastOp()))
     {
         mapGenTreeToValue(cast,
             cast->IsUnsigned()
             ? _builder.CreateZExt(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(cast->CastToType()))
             : _builder.CreateSExt(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(cast->CastToType())));
     }
-    else if ((castToType == TYP_INT || castToType == TYP_LONG || castToType == TYP_UINT || castToType == TYP_ULONG) && cast->CastOp()->TypeIs(TYP_FLOAT, TYP_DOUBLE))
+    else if (cast->TypeIs(TYP_INT, TYP_LONG) && cast->CastOp()->TypeIs(TYP_FLOAT, TYP_DOUBLE))
     {
         mapGenTreeToValue(cast,
             cast->IsUnsigned()
@@ -1174,7 +1176,7 @@ void Llvm::fillPhis()
 
 void Llvm::buildReturn(GenTree* node)
 {
-    switch (node->gtType)
+    switch (node->TypeGet())
     {
         case TYP_BOOL:
         case TYP_BYTE:
@@ -1260,7 +1262,7 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
     }
 
     // implicit truncating from long to int
-    if (llvmRef->getType() == Type::getInt64Ty(_llvmContext) && getLlvmTypeForVarType(lclVar->TypeGet()) == Type::getInt32Ty(_llvmContext))
+    if (llvmRef->getType() == Type::getInt64Ty(_llvmContext) && lclVar->TypeIs(TYP_INT))
     {
         llvmRef = _builder.CreateTrunc(llvmRef, Type::getInt32Ty(_llvmContext));
     }
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -380,7 +380,7 @@ CorInfoType Llvm::toCorInfoType(var_types varType)
 
 CORINFO_CLASS_HANDLE Llvm::tryGetStructClassHandle(LclVarDsc* varDsc)
 {
-    return toCorInfoType(varDsc->TypeGet()) == CorInfoType::CORINFO_TYPE_VALUECLASS ? varDsc->GetStructHnd() : nullptr;
+    return varTypeIsStruct(varDsc) ? varDsc->GetStructHnd() : NO_CLASS_HANDLE;;
 }
 
 unsigned int Llvm::padOffset(CorInfoType corInfoType, CORINFO_CLASS_HANDLE structClassHandle, unsigned int atOffset)
@@ -501,12 +501,17 @@ FunctionType* Llvm::getFunctionTypeForSigInfo(CORINFO_SIG_INFO& sigInfo)
     return FunctionType::get(retLlvmType, ArrayRef<Type*>(argVec), false);
 }
 
-Value* getOrCreateExternalSymbol(const char* symbolName)
+Value* getOrCreateExternalSymbol(const char* symbolName, Type* symbolType = nullptr)
 {
+    if (symbolType == nullptr)
+    {
+        symbolType = Type::getInt32PtrTy(_llvmContext);
+    }
+
     Value* symbol = _module->getGlobalVariable(symbolName);
     if (symbol == nullptr)
     {
-        symbol = new llvm::GlobalVariable(*_module, Type::getInt32PtrTy(_llvmContext), true, llvm::GlobalValue::LinkageTypes::ExternalLinkage, (llvm::Constant*)nullptr, symbolName);
+        symbol = new llvm::GlobalVariable(*_module, symbolType, true, llvm::GlobalValue::LinkageTypes::ExternalLinkage, (llvm::Constant*)nullptr, symbolName);
     }
     return symbol;
 }
@@ -521,7 +526,7 @@ Function* getOrCreateRhpAssignRef()
     return llvmFunc;
 }
 
-Type* Llvm::getLLVMTypeForVarType(var_types type)
+Type* Llvm::getLlvmTypeForVarType(var_types type)
 {
     // TODO: Fill out with missing type mappings and when all code done via clrjit, default should fail with useful
     // message
@@ -535,7 +540,11 @@ Type* Llvm::getLLVMTypeForVarType(var_types type)
         case TYP_USHORT:
             return Type::getInt16Ty(_llvmContext);
         case TYP_INT:
+        case TYP_UINT:
             return Type::getInt32Ty(_llvmContext);
+        case TYP_LONG:
+        case TYP_ULONG:
+            return Type::getInt64Ty(_llvmContext);
         case var_types::TYP_FLOAT:
             return Type::getFloatTy(_llvmContext);
         case var_types::TYP_DOUBLE:
@@ -572,6 +581,8 @@ llvm::Instruction* Llvm::getCast(llvm::Value* source, Type* targetType)
     {
         switch (sourceTypeID)
         {
+            case Type::TypeID::PointerTyID:
+                return new llvm::PtrToIntInst(source, targetType, "CastPtrToInt");
             case Type::TypeID::IntegerTyID:
                 if (sourceType->getPrimitiveSizeInBits() > targetType->getPrimitiveSizeInBits())
                 {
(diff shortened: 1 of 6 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -378,9 +378,45 @@ CorInfoType Llvm::toCorInfoType(var_types varType)
...
+
+        case TYP_DOUBLE:
+            return Type::getDoubleTy(_llvmContext);
+
+        case TYP_BYREF:
+        case TYP_REF:
+            return Type::getInt8PtrTy(_llvmContext);
+
+        default:
+            failFunctionCompilation();
```
yowl: This is not great.  the value 27, 0x1b is seen and fails, I'm not sure what that means 0x10 & 0x0b ?  E..g for 
```
        public static long Flt2LngOvf(float val)
```

yowl: I think this happens in `fgMorphIntoHelperCall` where a cast is changed to the helper call and `gtCallType` is not set.  Maybe I shouldn't rely on `gtCallType` for helper calls.

yowl:
Converting to draft while I redo the return type and add some casts

jkotas:
cc @SingleAccretion

SingleAccretion:
Will review later tomorrow.

src/coreclr/jit/llvm.cpp
```diff
@@ -841,26 +852,131 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
...
+        }
+        else
+        {
+            // we may come back into managed from the unmanaged call so store the shadowstack
+            _builder.CreateStore(shadowStackForCallee, getOrCreateExternalSymbol("t_pShadowStackTop", Type::getInt8PtrTy(_llvmContext)));
+        }
+
+        for (OperandArgNum opAndArg : sortedArgs)
+        {
+            if ((opAndArg.operand->gtOper == GT_CNS_INT) && opAndArg.operand->IsIconHandle())
```
SingleAccretion, line 985:
```suggestion
            if (opAndArg.operand->IsCnsIntOrI() && opAndArg.operand->IsIconHandle())
```
It seems that we should check for what kind (`GTF_ICON_*`) of handle this is, not assume that it represents a type, though I do not know which flag this one will have.

It also seems that we should register the relocations when building the constant, not when processing the call. It is possible for the constant to get CSEd away, technically.

src/coreclr/jit/llvm.cpp
```diff
@@ -841,26 +852,131 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
...
+        // TODO-LLVM: If the block has a handler, this will need to be an invoke.  E.g. create a CallOrInvoke as per ILToLLVMImporter
+        mapGenTreeToValue(call, _builder.CreateCall(llvmFunc, llvm::ArrayRef<Value*>(argVec)));
+        if (call->gtCallMethHnd == _compiler->eeFindHelper(CORINFO_HELP_THROW))
+        {
+            _builder.CreateUnreachable();
+        }
+    }
+}
+
+void Llvm::buildCall(GenTree* node)
```
SingleAccretion, line 1007: Nit: I would drop the `return`s from the `if` branches and instead add the `else` for the failure case:
```c++
if (call->gtCallType == CT_HELPER) { ... }
else if (call->gtCallType == CT_USER_FUNC) { ... }
else { failFunctionCompilation(); }
```
It is unusual to combine "early return" style with an `if/else` style.

src/coreclr/jit/llvm.cpp
```diff
@@ -841,26 +852,131 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
...
+{
+    if (call->gtCallMethHnd == _compiler->eeFindHelper(CORINFO_HELP_READYTORUN_GENERIC_HANDLE) ||
+        call->gtCallMethHnd == _compiler->eeFindHelper(CORINFO_HELP_READYTORUN_GENERIC_STATIC_BASE) ||
+        call->gtCallMethHnd == _compiler->eeFindHelper(CORINFO_HELP_READYTORUN_DELEGATE_CTOR))
+    {
+        // TODO-LLVM
+        failFunctionCompilation();
+    }
+
+    if (call->gtCallMethHnd == _compiler->eeFindHelper(CORINFO_HELP_READYTORUN_STATIC_BASE))
```
SingleAccretion, line 922: For my education: what makes this case special?

src/coreclr/jit/llvm.cpp
```diff
@@ -841,26 +852,131 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
...
+        Value* shadowStackForCallee = getShadowStackForCallee();
+        if (requiresShadowStack)
+        {
+            argVec.push_back(shadowStackForCallee);
+            argIx++;
+        }
+        else
+        {
+            // we may come back into managed from the unmanaged call so store the shadowstack
+            _builder.CreateStore(shadowStackForCallee, getOrCreateExternalSymbol("t_pShadowStackTop", Type::getInt8PtrTy(_llvmContext)));
```
SingleAccretion, line 980: If I understand correctly - we need this only when calling unmanaged methods?

Seems like something we should handle in lowering instead of codegen.

src/coreclr/jit/llvm.cpp
```diff
@@ -872,21 +988,29 @@ void Llvm::buildCall(GenTree* node)
...
         mapGenTreeToValue(cast, intValue); // nothing to do except map the source value to the destination GenTree
     }
-    else if (cast->CastToType() == TYP_DOUBLE && cast->CastOp()->TypeIs(TYP_FLOAT))
+    else if (castToType == TYP_DOUBLE && cast->CastOp()->TypeIs(TYP_FLOAT))
     {
-        mapGenTreeToValue(cast, _builder.CreateFPCast(getGenTreeValue(cast->CastOp()), getLLVMTypeForVarType(TYP_DOUBLE)));
+        mapGenTreeToValue(cast, _builder.CreateFPCast(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(TYP_DOUBLE)));
     }
-    else if (cast->CastToType() == TYP_INT && cast->CastOp()->TypeIs(TYP_FLOAT, TYP_DOUBLE))
+    else if ((castToType == TYP_LONG || castToType == TYP_ULONG) && cast->CastOp()->TypeIs(TYP_INT, TYP_UINT))
```
```suggestion
    else if (cast->TypeIs(TYP_LONG) && cast->CastOp()->TypeIs(TYP_INT))
```
Just a simpler condition that expresses the same thing (because we don't handle check arithmetic right now).

You could also make it `genActualTypeIsInt(cast->CastOp())`, it should handle more cases.

src/coreclr/jit/llvm.cpp
```diff
@@ -872,21 +988,29 @@ void Llvm::buildCall(GenTree* node)
...
     {
         mapGenTreeToValue(cast,
             cast->IsUnsigned()
-                ? _builder.CreateFPToUI(getGenTreeValue(cast->CastOp()), getLLVMTypeForVarType(TYP_INT))
-                : _builder.CreateFPToSI(getGenTreeValue(cast->CastOp()), getLLVMTypeForVarType(TYP_INT)));
+            ? _builder.CreateZExt(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(cast->CastToType()))
+            : _builder.CreateSExt(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(cast->CastToType())));
+    }
+    else if ((castToType == TYP_INT || castToType == TYP_LONG || castToType == TYP_UINT || castToType == TYP_ULONG) && cast->CastOp()->TypeIs(TYP_FLOAT, TYP_DOUBLE))
```
```suggestion
    else if (cast->TypeIs(TYP_INT, TYP_LONG) && cast->CastOp()->TypeIs(TYP_FLOAT, TYP_DOUBLE))
```

src/coreclr/jit/llvm.cpp
```diff
@@ -841,26 +852,131 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
     return mapGenTreeToValue(call, llvmCall);
 }
 
-void Llvm::buildCall(GenTree* node)
+llvm::FunctionType* Llvm::buildHelperLlvmFunctionType(GenTreeCall* call, bool withShadowStack)
```
```suggestion
FunctionType* Llvm::buildHelperLlvmFunctionType(GenTreeCall* call, bool withShadowStack)
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1052,7 +1176,15 @@ void Llvm::buildReturn(GenTree* node)
 {
     switch (node->gtType)
```
```suggestion
    switch (node->TypeGet())
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1127,6 +1259,12 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
         llvmRef = _localsMap->at({lclNum, ssaNum});
     }
 
+    // implicit truncating from long to int
+    if (llvmRef->getType() == Type::getInt64Ty(_llvmContext) && getLlvmTypeForVarType(lclVar->TypeGet()) == Type::getInt32Ty(_llvmContext))
```
```suggestion
    if (llvmRef->getType() == Type::getInt64Ty(_llvmContext) && lclVar->TypeIs(TYP_INT))
```
Seems simpler to use the tree type directly.

I think this should also be handled in lowering, though it does not matter much I guess.

src/coreclr/jit/llvm.cpp
```diff
@@ -841,26 +852,131 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
...
+        OperandArgNum* sortedData = sortedArgs.data();
+        bool requiresShadowStack = helperRequiresShadowStack(call->gtCallMethHnd);
+
+        for (unsigned i = 0; i < argCount; i++)
+        {
+            fgArgTabEntry* curArgTabEntry = argTable[i];
+            unsigned int   argNum = curArgTabEntry->argNum;
+            OperandArgNum  opAndArg = { argNum, curArgTabEntry->GetNode() };
+            sortedData[argNum] = opAndArg;
+        }
```
SingleAccretion: It would be nice to share the logic between user and helper calls; it seems they have no fundamental reason to be handled differently. I understand currently the sharing would not achieve a lot (there's not a lot of code), so this is a suggestion for a `TODO-LLVM`.

src/coreclr/jit/llvm.cpp
```diff
@@ -841,26 +852,131 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
...
 
-void Llvm::buildCall(GenTree* node)
+llvm::FunctionType* Llvm::buildHelperLlvmFunctionType(GenTreeCall* call, bool withShadowStack)
 {
-    GenTreeCall* call = node->AsCall();
-    if (call->gtCallType == CT_HELPER)
+    Type* retLlvmType = getLlvmTypeForVarType(call->TypeGet());
+    std::vector<llvm::Type*> argVec;
+
+    if (withShadowStack)
```
SingleAccretion, line 893: We insert the shadow stack in IR for user call, seems like we should do that for the helper calls as well?

SingleAccretion (commented):
Generally LGTM, but left some comments/questions.

src/coreclr/jit/llvm.cpp
yowl: Probably the best thing I can do is point you to https://github.com/dotnet/runtimelab/issues/839 and https://github.com/dotnet/runtimelab/issues/1640.  

What this means in practice is that for most helpers,  `gtEntryPoint.handle` is a `nullptr` and for `CORINFO_HELP_READYTORUN_STATIC_BASE` there is no implementation in https://github.com/dotnet/runtimelab/blob/d623b2be57be73364ecf853a0fa31e7e9582b4ee/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs#L240

src/coreclr/jit/llvm.cpp
yowl: Yes, anything unmanaged that might callback, anything that invokes a GC for example, then to avoid the generation of a new shadow stack via `builder.BuildArrayMalloc(LLVMTypeRef.Int8, BuildConstInt32(1000000), "NewShadowStack");`) we store the stack and it's picked up in `EmitNativeToManagedThunk` in https://github.com/dotnet/runtimelab/blob/d623b2be57be73364ecf853a0fa31e7e9582b4ee/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter_Statics.cs#L261-L266.  

If I move it to lowering, would I need some special `IconValue` to represent the thread static `t_pShadowStackTop`?

src/coreclr/jit/llvm.cpp
yowl: Yes, ok.  Especially when we get to exceptions.

src/coreclr/jit/llvm.cpp
yowl: Its been talked about, but currently we only add the return slot to the IR for user calls.  The shadow stack arg itself has not been done yet.  Its added for user calls here https://github.com/dotnet/runtimelab/blob/d623b2be57be73364ecf853a0fa31e7e9582b4ee/src/coreclr/jit/llvm.cpp#L797-L801  Agree though that it would be good to do.  I've got a couple of PRs queued up behind this one, then I can look at moving more things to lowering, if that's ok?

src/coreclr/jit/llvm.cpp
yowl: Its not clear what `GTF_ICON_` should be here.  `GTF_ICON_CLASS_HDL` words for some, but for `RhpNewFast` it takes a `MethodTable` and that has `GTF_ICON_METHOD_HDL` . I've left it without a check for now.

yowl: `((opAndArg.operand->GetIconHandleFlag() & GTF_ICON_CLASS_HDL) == GTF_ICON_CLASS_HDL)`  "works", but I don't see any indication that `GTF_ICON_*` are bit flags, so that's probably just luck.  I

src/coreclr/jit/llvm.cpp
yowl: thanks

src/coreclr/jit/llvm.cpp
yowl: done

src/coreclr/jit/llvm.cpp
yowl: done.

src/coreclr/jit/llvm.cpp
yowl: Done

src/coreclr/jit/llvm.cpp
yowl: thanks,

src/coreclr/jit/llvm.cpp
SingleAccretion: > then I can look at moving more things to lowering, if that's ok?

Of course.

src/coreclr/jit/llvm.cpp
SingleAccretion: > If I move it to lowering, would I need some special `IconValue` to represent the thread static `t_pShadowStackTop`?

Yes. Alternatively, and I think it'll be easier/more straightforward, you could use `GT_CLS_VAR_ADDR` with some distinguished handle (say an address of an entry in some static array), which is what the mainline lowering does in similar situations. The codegen would then pick the handles up.

src/coreclr/jit/llvm.cpp
SingleAccretion: I will dig deeper into this some other day;  let's just leave a `TODO-LLVM` for now.

yowl:
Going to update emscripten.  There is this
```
Fix race condition when running many emcc processes after clearing the cache.
  The processes would race to run the sanity checks and could interfere with
  each other (#13299).
```
https://emscripten.org/docs/introducing_emscripten/release_notes.html

which we don't have, maybe it will help ...

yowl:
It timed out.  I'll try tomorrow.

jkotas:
All tests are timing out...

yowl:
I see, `(ulong)pointer` is failing now, coverage must be increasing.

src/coreclr/jit/llvm.cpp
```diff
@@ -874,52 +885,170 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
...
-    else if (cast->CastToType() == TYP_DOUBLE && cast->CastOp()->TypeIs(TYP_FLOAT))
+    else if (castToType == TYP_DOUBLE && cast->CastOp()->TypeIs(TYP_FLOAT))
     {
-        mapGenTreeToValue(cast, _builder.CreateFPCast(getGenTreeValue(cast->CastOp()), getLLVMTypeForVarType(TYP_DOUBLE)));
+        mapGenTreeToValue(cast, _builder.CreateFPCast(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(TYP_DOUBLE)));
+    }
+    else if (cast->TypeIs(TYP_LONG) && genActualTypeIsInt(cast->CastOp()))
+    {
+        // Cast pointer to int if necessary.  TODO-LLVM: candidate for lowering?
+        Value* sourceValue = castIfNecessary(getGenTreeValue(cast->CastOp()), getLlvmTypeForVarType(cast->CastOp()->TypeGet()));
```
SingleAccretion, line 1039: A bit curious what did this code encounter that the source was lying about the type. There was one case fixed somewhat recently in the upstream that looked sort of similar.

Definitely a candidate for lowering.

yowl: The method is https://github.com/dotnet/runtimelab/blob/9d54885c319c09817e90928e96bd88382726907d/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L1244-L1249 for which the IR starts (node 4 is the problem):

```

------------ BB02 [000..018) (return), preds={BB01} succs={}
               [000020] ------------                 IL_OFFSET void   IL offset: 0x0
               [000021] ------------                 IL_OFFSET void   IL offset: 0x1
N001 (  3,  2) [000003] ------------         t3 =    LCL_VAR   int    V01 arg1         u:1 (last use)
                                                  /--*  t3     int
N002 (  4,  4) [000004] ---------U--         t4 = *  CAST      long <- ulong <- uint
```

yowl: `t3` maps to arg 1 which in LLVM is a pointer type, and the cast then just tries to zext that to i64 which LLVM doesn't like

SingleAccretion: I see, these are actual pointers, not `TYP_BYREF/TYP_REF`. Unfortunate that we have to insert these casts that have no semantic meaning, though as I understand we have no choice. We'll leave this to codegen then, there is not way to even represent "typed pointers" in IR.

yowl:
@jkotas I think this is ready now, unless you have any concerns.  Thanks.

SingleAccretion (approved):

jkotas:
Thanks!

PR #1037, opened 2021-04-29
Scan throw helper for unsupported EH

MichalStrehovsky:
Fixed #991.

PR #1685, opened 2021-10-29
Change how we deal with scanning failures

MichalStrehovsky:
When we import a method that has a problem in it (e.g. read a non-existent field because the user provided mismatched assemblies), we generate IL body for a throwing method that throws the appropriate exception matching JIT-based behavior. We then import that IL body instead of the original method.

We do this in both scanner and codegen. This worked for us so far, but I just hit a case where this doesn't work.

Before RyuJIT hits the problematic part of the method that makes us discard it, it might attempt to import stuff we didn't scan (e.g. try to do dictionary lookups, ask about virtual slots, etc.). This makes us take the "there's a compiler bug" codepaths before we find out it doesn't actually matter. We don't keep the partial dependencies from the scanning phase and even if we did, there's no guarantees RyuJIT imports in the same order as scanner - all we have is the dependencies of the throwing stub.

So instead, remember that a method had a problem during scanning and don't even attempt to give such method body to RyuJIT - just pass the throwing body right away.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
@@ -204,7 +204,7 @@ namespace Internal.IL
                     {
                         TypeDesc catchType = (TypeDesc)_methodIL.GetObject(region.ClassToken);
                         if (catchType.IsRuntimeDeterminedSubtype)
-                            _dependencies.Add(_factory.MethodEntrypoint(_factory.TypeSystemContext.GetHelperEntryPoint("ThrowHelpers", "ThrowInvalidProgramException")), "Unsupported EH");
+                            ThrowHelper.ThrowInvalidProgramException();
                     }
                 }
             }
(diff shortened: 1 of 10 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
```diff
@@ -204,7 +204,7 @@ private void StartImportingBasicBlock(BasicBlock basicBlock)
                     {
                         TypeDesc catchType = (TypeDesc)_methodIL.GetObject(region.ClassToken);
                         if (catchType.IsRuntimeDeterminedSubtype)
-                            _dependencies.Add(_factory.MethodEntrypoint(_factory.TypeSystemContext.GetHelperEntryPoint("ThrowHelpers", "ThrowInvalidProgramException")), "Unsupported EH");
+                            ThrowHelper.ThrowInvalidProgramException();
```
MichalStrehovsky, line 207: There was a back and forth with this in 314ebe3123302c4f9728dd5dd2dd81607a9c8920 and #1037. We can now do the thing that was originally there (and matches what CorInfoImpl does as well).

jkotas (approved):
Nice

PR #1686, opened 2021-10-29
Add generic cycle detection to generic virtual methods

MichalStrehovsky:


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/LazyGenerics/ModuleCycleInfo.cs
```diff
@@ -22,13 +24,11 @@ public ModuleCycleInfo(EcmaModule module, HashSet<TypeSystemEntity> entitiesInCy
                 _entitiesInCycles = entitiesInCycles;
             }
 
-            public bool FormsCycle(TypeSystemEntity owner, TypeSystemEntity referent)
+            public bool FormsCycle(TypeSystemEntity owner)
```
MichalStrehovsky, line 27: The things that come out of the algorithm are places where we need to enable lazy generics. What was here originally wouldn't identify the Make method in the test below as problematic, and we wouldn't run the cutoff logic.

jkotas (approved):
Thanks

PR #1698, opened 2021-11-02
[NativeAOT-LLVM] Add support for *struct args

yowl:
This PR adds support for LLVM signatures where an argument is a pointer to a struct (or any level of **..struct)

I've copied the algorithm at https://github.com/dotnet/runtimelab/blob/d623b2be57be73364ecf853a0fa31e7e9582b4ee/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs#L1242-L1269 where it recurses to get the type so I added 2 callbacks in order to do that.  The existing `getArgType` in `CorInfoImpl` does not return the class handle for parameterized types so rather than modify it I created another method.  I did that because up till now we've not modified
`CorInfoImpl` and adding callbacks is one way to see the changes required for LLVM.

This also highlighted a deficiency in the phi args, when the arg is an argument passed on the LLVM stack (and there is no `d:1` local for it), so there's a fix for that.

Also copied the workaround for `GetRandomBytes`  .  I don't really want to bake into the build a dependency on javascript libraries (which is how mono does this) as it would stop the compiler being able to generate pure Wasm builds, for other hosts, wasmer/wasmtime etc.  Long term this problem could perhaps be solved with an interface using the WebAssembly component ABI which could be fulfilled by either javascript, or a Wasm module.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -302,6 +302,11 @@ llvm::Type* Llvm::getLlvmTypeForStruct(CORINFO_CLASS_HANDLE structHandle)
     return _llvmStructs->at(structHandle);
 }
 
+//------------------------------------------------------------------------
+// Returns the VM defined TypeDesc.GetParameterType() for the given type
+// Intended for pointers to generate the appropriate LLVM pointer type
+// E.g. "[S.P.CoreLib]Internal.Runtime.MethodTable"*
+// 
 llvm::Type* Llvm::getLlvmTypeForParameterType(CORINFO_CLASS_HANDLE classHnd)
 {
     CORINFO_CLASS_HANDLE innerParameterHandle;
@@ -1055,9 +1060,8 @@ void Llvm::fillPhis()
             auto iter = _localsMap->find({ lclNum, ssaNum });
             if (iter == _localsMap->end())
             {
-                //// Function args can be used as phi args without having seen them, and added them to _localsMap via a local.
-                //// But only for non shadow stack args with u:1
-                assert(_compiler->lvaIsParameter(lclNum) && ssaNum == 1);
+                // Arguments are implicitly defined on entry to the method.
+                assert(_compiler->lvaIsParameter(lclNum) && ssaNum == SsaConfig::FIRST_SSA_NUM);
                 LlvmArgInfo  llvmArgInfo = getLlvmArgInfoForArgIx(_sigInfo, lclNum);
                 localPhiArg = _function->getArg(llvmArgInfo.m_argIx);
             }
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -35,7 +35,9 @@ static const char* (*_getDocumentFileName)(void*);
 static const uint32_t (*_firstSequencePointLineNumber)(void*);
 static const uint32_t (*_getOffsetLineNumber)(void*, unsigned int ilOffset);
 static const uint32_t(*_structIsWrappedPrimitive)(void*, CORINFO_CLASS_STRUCT_*, CorInfoType);
-static const uint32_t(*_padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned atOffset);
+static const uint32_t(*_padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned);
+static const CorInfoTypeWithMod(*_getArgTypeIncludingParameterized)(void*, CORINFO_SIG_INFO*, CORINFO_ARG_LIST_HANDLE, CORINFO_CLASS_HANDLE*);
+static const CorInfoTypeWithMod(*_getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*);
 
 static char*                              _outputFileName;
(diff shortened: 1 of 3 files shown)


yowl:
cc @SingleAccretion

src/coreclr/jit/llvm.cpp
```diff
@@ -323,6 +341,16 @@ llvm::Type* Llvm::getLlvmTypeForCorInfoType(CorInfoType corInfoType, CORINFO_CLA
...
+        case CorInfoType::CORINFO_TYPE_PTR:
+        {
+            if (classHnd == NO_CLASS_HANDLE)
+            {
+                return Type::getInt8Ty(_llvmContext)->getPointerTo();
+            }
+
+            return getLlvmTypeForParameterType(classHnd)->getPointerTo();
+        }
+
```
SingleAccretion, line 358: I find this confusing: the method (`getLlvmTypeForCorInfoType`) claims to work on arbitrary types, but here it depends on the type being that of a parameter. Is the presumption here that we'll never see/need `CORINFO_TYPE_PTR` outside of parameters?

src/coreclr/jit/llvm.cpp
```diff
@@ -1023,7 +1051,21 @@ void Llvm::fillPhis()
...
             unsigned       ssaNum = phiArg->GetSsaNum();
 
-            Value* localPhiArg = _localsMap->at({ lclNum, ssaNum });
+            Value* localPhiArg = nullptr;
+            auto iter = _localsMap->find({ lclNum, ssaNum });
+            if (iter == _localsMap->end())
+            {
+                //// Function args can be used as phi args without having seen them, and added them to _localsMap via a local.
+                //// But only for non shadow stack args with u:1
+                assert(_compiler->lvaIsParameter(lclNum) && ssaNum == 1);
```
```suggestion
                assert(_compiler->lvaIsParameter(lclNum) && ssaNum == SsaConfig::FIRST_SSA_NUM);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1023,7 +1051,21 @@ void Llvm::fillPhis()
             unsigned       lclNum = phiArg->GetLclNum();
             unsigned       ssaNum = phiArg->GetSsaNum();
 
-            Value* localPhiArg = _localsMap->at({ lclNum, ssaNum });
+            Value* localPhiArg = nullptr;
+            auto iter = _localsMap->find({ lclNum, ssaNum });
+            if (iter == _localsMap->end())
+            {
+                //// Function args can be used as phi args without having seen them, and added them to _localsMap via a local.
+                //// But only for non shadow stack args with u:1
```
```suggestion
                // Arguments are implicitly defined on entry to the method.
```
Don't think we need to be more elaborate than that.

src/coreclr/jit/llvm.cpp
```diff
@@ -1671,6 +1713,17 @@ void Llvm::Compile()
             _module); // TODO: ExternalLinkage forced as linked from old module
     }
 
+    // mono does this via Javascript (pal_random.js), but prefer not to introduce that dependency as it limits the ability to run out of the browser.
+    // Copy the temporary workaround from the IL->LLVM generator for now.
+    if (!strcmp(mangledName, "S_P_CoreLib_Interop__GetRandomBytes"))
```
SingleAccretion, line 1722: Do you expect we'll have more workarounds like this in the future? If so, it may be worthwhile to put this into its own method, to reduce the noise when reading the mainline code.

src/coreclr/jit/llvm.cpp
yowl: You can have `void*` which is caught by `classHnd == NO_CLASS_HANDLE`.  Am I confusing things by the use of `parameter`?  Here `parameter` means a type parameter, as in https://github.com/dotnet/runtimelab/blob/d623b2be57be73364ecf853a0fa31e7e9582b4ee/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs#L1269.  I suppose `*int` and similar are also possible, they might be a problem I suppose in that there might be some unnecessary bitcasting going on.

SingleAccretion: Aha, I see what is going on. The terminology is a bit overloaded, but it is consistent with the VM, so I guess we should leave it as is. I suggest adding a header comment to `getLlvmTypeForParameterType` on what it is meant to do though.

src/coreclr/jit/llvm.cpp
yowl: The only other one we have is 
```
            else if ((method.OwningType as EcmaType)?.Name == "CalendarData" && method.Name == "EnumCalendarInfo")
            {
                // just return false 
                methodIL = new ILStubMethodIL(method, new byte[] { (byte)ILOpcode.ldc_i4_0, (byte)ILOpcode.ret }, Array.Empty<LocalVariableDefinition>(), null);
            }
```
And that may be a hang over from before mono implemented a lot of globalization functions.  Before moving that one over, we should check if its implemented.

yowl: As far as random bytes are concerned for WebAssembly, this proposal looks like the way to go in the long term.
https://github.com/sunfishcode/wasi-random-presentation

SingleAccretion (approved):

PR #1700, opened 2021-11-04
Prevent devirtualizing into abstract classes never derived from

MichalStrehovsky:
We defer looking at virtual methods on abstract classes and instead only look at them once we have a non-abstract class deriving from them. This lets us optimize away method bodies on abstract classes that are overriden by all descendants and therefore unreachable.

If there's no non-abstract descendant, we didn't look at any virtuals. But RyuJIT might still attempt to devirtualize into such classes. We need to prevent that to avoid compilation failures (and bloat).


jkotas (approved):
Thanks

PR #1702, opened 2021-11-05
[NativeAOT-LLVM] Upgrade emscripten and workaround cache lock problem

yowl:
This PR upgrades the Emscripten version and puts a workaround for the cache lock problem we've been experiencing.  I've not got to the bottom of the underlying reason, but the workaround seems ok locally (and it was failing locally before).  

Workaround is to build the Emscripten ports for ICU and zlib upfront.  These builds are cached, so it does not incur extra build work.

https://github.com/emscripten-core/emscripten/issues/15440

Also suppress warning for __x variables as latest clang defines them as reserved identifiers.

### DIFF
--- a/eng/native/gen-buildsys.cmd
+++ b/eng/native/gen-buildsys.cmd
@@ -84,6 +84,10 @@ if not "%__ConfigureOnly%" == "1" (
 )
 
 if /i "%__UseEmcmake%" == "1" (
+    REM workaround for https://github.com/emscripten-core/emscripten/issues/15440 - emscripten cache lock problems
+    REM build the ports for ICU and ZLIB upfront
+    embuilder build icu zlib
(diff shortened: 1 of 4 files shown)


SingleAccretion:
I think we need to add `TODO`s to remove the supressions once https://github.com/dotnet/runtime/commit/2265a61565e20fc26ca8ec0632eb14268b99158e is merged (which also suppresses the warnings, but is present in the upstream).

SingleAccretion (approved):

yowl:
Linux thread test failed with
```
native/BasicThreading ''
FinalizeTest failed
Expected: 100
Actual: 255
```
Seems unrelated.

jkotas (approved):
Thanks

PR #1705, opened 2021-11-06
[NativeAOT-LLVM] Support for more unary and binary ops

### LINKED ISSUE #1731: [NativeAOT-LLVM] Question: Is it valid IR for phi args with common dominator to share a phi arg?
Looking at this IR
```
------------ BB09 [05B..060) -> BB10 (always), preds={BB03,BB05,BB08} succs={BB10}
     (  0,  0) [000140] ------------       t140 =    PHI_ARG   int    V03 loc0         u:1
     (  0,  0) [000139] ------------       t139 =    PHI_ARG   int    V03 loc0         u:3
                                                  /--*  t140   int
                                                  +--*  t139   int
               [000137] ------------       t137 = *  PHI       int
```
There are 3 predecessors, `BB03,BB05,BB08` and 2 phi args.  Both `BB03` and `BB05` are dominated by `BB02` which is where the definition is for `u:1`:

```
After computing the Dominance Tree:
BB01 : BB02
BB02 : BB09 BB04 BB03
BB04 : BB06 BB05
BB06 : BB08 BB07
BB09 : BB10
```

```
------------ BB02 [000..00F) -> BB04 (cond), preds={BB01} succs={BB03,BB04}
               [000088] ------------                 IL_OFFSET void   IL offset: 0x0
               [000089] ------------                 IL_OFFSET void   IL offset: 0x1
N001 (  1,  1) [000002] ------------         t2 =    CNS_INT   int    0
                                                  /--*  t2     int
N003 (  5,  4) [000004] DA----------              *  STORE_LCL_VAR int    V03 loc0         d:1
```

So logically that seems fine, LLVM however does not allow that:
for
```llvm
32:                                               ; preds = %60, %34, %11
  %33 = phi i32 [ 0, %34 ], [ %48, %60 ], !dbg !17513
```
you get:
```
PHINode should have one entry for each predecessor of its parent basic block!
  %33 = phi i32 [ 0, %34 ], [ %48, %60 ], !dbg !17513
```
Is the IR valid?

yowl:
This PR adds support in the RyuJIT ->LLVM backend for some more unary ops (NEG and NOT) and some binary ops, shifts and logical ops.

Closes #1731

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1213,7 +1213,6 @@ void Llvm::fillPhis()
                 localPhiArg = iter->second;
             }
 
-
             Value* phiRealArgValue;
             llvm::Instruction* castRequired = getCast(localPhiArg, llvmPhiNode->getType());
             if (castRequired != nullptr)
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -742,36 +742,6 @@ void Llvm::buildAdd(GenTree* node, Value* op1, Value* op2)
     }
 }
 
-Value* Llvm::genTreeAsLlvmType(GenTree* tree, Type* type)
-{
-    Value* v = getGenTreeValue(tree);
-    if (v->getType() == type)
-        return v;
-
-    if (tree->IsIntegralConst() && tree->TypeIs(TYP_INT))
-    {
-        if (type->isPointerTy())
-        {
-            return _builder.CreateIntToPtr(v, type);
-        }
-        return _builder.getInt({(unsigned int)type->getPrimitiveSizeInBits().getFixedSize(), (uint64_t)tree->AsIntCon()->IconValue(), true});
-    }
-    //// from i1 (e.g. from GT_EQ) to wider int
-    if (v->getType()->isIntegerTy() && type->isIntegerTy() &&
-        v->getType()->getPrimitiveSizeInBits().getFixedSize() < type->getPrimitiveSizeInBits().getFixedSize())
-    {
-        return _builder.CreateZExt(v, type);
-    }
-
-    // i32* e.g symbols, to i8*
-    if (v->getType()->isPointerTy() && type->isPointerTy())
-    {
-        return _builder.CreateBitCast(v, type);
-    }
-
-    failFunctionCompilation();
-}
-
 unsigned int Llvm::getTotalRealLocalOffset()
 {
     return _shadowStackLocalsSize;
@@ -816,7 +786,7 @@ bool Llvm::isThisArg(GenTreeCall* call, GenTree* operand)
 
 void Llvm::storeOnShadowStack(GenTree* operand, Value* shadowStackForCallee, unsigned int offset)
 {
-    castingStore(genTreeAsLlvmType(operand, Type::getInt8PtrTy(_llvmContext)),
+    castingStore(consumeValue(operand, Type::getInt8PtrTy(_llvmContext)),
                  getShadowStackOffest(shadowStackForCallee, offset), Type::getInt8PtrTy(_llvmContext));
 }
 
@@ -835,12 +805,31 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
 
     if (nodeValue->getType() != targetLlvmType)
     {
+        // int to pointer type
+        if (node->TypeIs(TYP_INT) && targetLlvmType->isPointerTy())
+        {
+            return _builder.CreateIntToPtr(nodeValue, targetLlvmType);
+        }
+
+        // pointer to ints
+        if (nodeValue->getType()->isPointerTy() && targetLlvmType == Type::getInt32Ty(_llvmContext))
+        {
+            return _builder.CreatePtrToInt(nodeValue, Type::getInt32Ty(_llvmContext));
+        }
+
+        // i32* e.g symbols, to i8*
+        if (nodeValue->getType()->isPointerTy() && targetLlvmType->isPointerTy())
+        {
+            return _builder.CreateBitCast(nodeValue, targetLlvmType);
+        }
+
+        // int and smaller int conversions
         assert(targetLlvmType->isIntegerTy() && nodeValue->getType()->isIntegerTy() &&
-               nodeValue->getType()->getPrimitiveSizeInBits() <= 64 && targetLlvmType->getPrimitiveSizeInBits() <= 64);
+               nodeValue->getType()->getPrimitiveSizeInBits() <= 32 && targetLlvmType->getPrimitiveSizeInBits() <= 32);
         if (nodeValue->getType()->getPrimitiveSizeInBits() < targetLlvmType->getPrimitiveSizeInBits())
         {
             // Upcast.
-            if (varTypeIsSmall(node) || targetLlvmType == Type::getInt64Ty(_llvmContext))
+            if (varTypeIsSmall(node))
             {
                 finalValue = varTypeIsSigned(node) ? _builder.CreateSExt(nodeValue, targetLlvmType)
                                                    : _builder.CreateZExt(nodeValue, targetLlvmType);
@@ -849,7 +838,7 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
             {
                 // This is the special case for relops. Ordinary codegen "just knows" they need zero-extension.
                 assert(nodeValue->getType() == Type::getInt1Ty(_llvmContext));
-                finalValue = _builder.CreateZExt(nodeValue, Type::getInt32Ty(_llvmContext));
+                finalValue = _builder.CreateZExt(nodeValue, targetLlvmType);
             }
         }
         else
@@ -900,6 +889,17 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
     fgArgTabEntry**                   argTable             = argInfo->ArgTable();
     std::vector<OperandArgNum>        sortedArgs           = std::vector<OperandArgNum>(argCount);
     OperandArgNum*                    sortedData           = sortedArgs.data();
+    unsigned int                      thisArgNum;
+    unsigned int                      returnAddressArgNum  = -1;
+
+    // TODO-LLVM: delete this when PUT_ARG inserted for the return address
+    thisArgNum          = 0;
+    if (needsReturnStackSlot(sigInfo.retType, sigInfo.retTypeClass))
+    {
+        // return slot before this
+        thisArgNum++;
+        returnAddressArgNum = 0;
+    }
 
     for (unsigned i = 0; i < argCount; i++)
     {
@@ -913,13 +913,32 @@ llvm::Value* Llvm::buildUserFuncCall(GenTreeCall* call)
     {
         if (opAndArg.operand->IsArgPlaceHolderNode())
         {
+            // TODO-LLVM: delete place holder args when rewriting the signature
+            thisArgNum++;
+            returnAddressArgNum++;
             continue;
         }
+
+        // "this" not in sigInfo arg list, dont increment argIx.  TODO-LLVM look to remove when PUTARG is inserted
+        if (opAndArg.argNum == thisArgNum && sigInfo.hasThis())
+        {
+            storeOnShadowStack(opAndArg.operand, shadowStackForCallee, shadowStackUseOffest);
+            shadowStackUseOffest += TARGET_POINTER_SIZE;
+            continue;
+        }
+
+        // Return address arg not in sigInfo, but needs to be added to LLVM args. TODO-LLVM delete when PUT_ARG is inserted
+        if (returnAddressArgNum == opAndArg.argNum)
+        {
+            argVec.push_back(consumeValue(opAndArg.operand, Type::getInt8PtrTy(_llvmContext)));
+            continue;
+        }
+
         LlvmArgInfo llvmArgInfo = getLlvmArgInfoForArgIx(sigInfo, argIx);
         if (llvmArgInfo.m_argIx >= 0)
         {
             // pass the parameter on the LLVM stack
-            argVec.push_back(genTreeAsLlvmType(opAndArg.operand, llvmFunc->getArg(llvmArgInfo.m_argIx)->getType()));
+            argVec.push_back(consumeValue(opAndArg.operand, llvmFunc->getArg(llvmArgInfo.m_argIx)->getType()));
         }
         else
         {
@@ -954,7 +973,12 @@ FunctionType* Llvm::buildHelperLlvmFunctionType(GenTreeCall* call, bool withShad
 
 bool Llvm::helperRequiresShadowStack(CORINFO_METHOD_HANDLE corinfoMethodHnd)
 {
-    return corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPEHANDLE);
+    //TODO-LLVM: is there a better way to identify managed helpers?
+    return corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPEHANDLE) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2INT_OVF) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2LNG_OVF) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2UINT_OVF) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2ULNG_OVF);
 }
 
 void Llvm::buildHelperFuncCall(GenTreeCall* call)
@@ -1039,7 +1063,7 @@ void Llvm::buildHelperFuncCall(GenTreeCall* call)
             }
             else
             {
-                argVec.push_back(genTreeAsLlvmType(opAndArg.operand, llvmFunc->getArg(argIx)->getType()));
+                argVec.push_back(consumeValue(opAndArg.operand, llvmFunc->getArg(argIx)->getType()));
             }
             argIx++;
         }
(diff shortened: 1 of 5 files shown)


jkotas:
cc @SingleAccretion

src/coreclr/jit/llvm.cpp
```diff
@@ -1223,6 +1237,83 @@ void Llvm::fillPhis()
...
+    }
+    mapGenTreeToValue(node, result);
+}
+
+void Llvm::buildBinaryOperation(genTreeOps oper, GenTree* node, Value* op1, Value* op2)
+{
+    Value* result;
+    // widen short ints as required
+    op1 = zextIntIfNecessary(op1);
+    op2 = zextIntIfNecessary(op2);
```
SingleAccretion: > genTreeOps oper, GenTree* node, Value* op1, Value* op2

I think we should reconsider this pattern, it loses information about small types (also `oper` is not necessary, since the node carries it).

There are basically two possibilities here:
1) Produce "normal" (upcast to `TYP_INT`, converted from pointers to integers) values when building the node.
2) Consume "normal" values when building the user.

The first is simpler, but will result in a more voluminous LLVM IR in case the user immediately downcasts, which may be confusing to read in the LLVM IR dumps. E. g. for this:
`CALL(short V00)` (where `short` is the signature type) it will produce:
```llvm
%argNorm = sext i16 %V00 to i32
%argVal  = trunc i32 %argNorm to i16
call %argVal
```
The code for the first approach would basically see us replace all `mapGenTreeToValue` with something like `produceValue`, which does the normalization.

The second approach would be to "normalize on demand", where the user nodes would pass in the types they "desire" (e. g. in the example above it'd be `short` for that call argument), this is basically what is going on here, only we would also pass the source node and the target type so as to not lose the information about small types:
```c++
Value* consumeValue(GenTree* node, var_types targetType);
Value* consumeValue(GenTree* node) => consumeValue(node, genActualType(node));
```
The primary benefit is it'll create less LLVM IR for calls and less confusing dumps, the primary downside is it is more complex (i. e. we'll have to to handle cases like `node->TypeGet()` is `TYP_BYTE`, `targetType` is `TYP_USHORT`, what does that mean - the answer being that it means we need one `sext i8 to i16` followed by `zext i16 to i32`).

SingleAccretion (commented):
We should decide on the normalization strategy here.

src/coreclr/jit/llvm.cpp
yowl: > `node->TypeGet()` is `TYP_BYTE`, `targetType` is `TYP_USHORT`, what does that mean - the answer being that it means we need one `sext i8 to i16`

Why is that `sext`?  `0xff TYP_BYTE || 0x0 TYP_BYTE`  with `targetType` is `TYP_USHORT`would then result in `0x0000ffff` wouldn't it ?  Is that right?

SingleAccretion: > Why is that `sext`?

`TYP_BYTE` is signed, so it is `sext`. I guess my example is not the best because in the end we can just do `sext i8 to i16`, seeing as we're targeting `i16`, not `i32` - I got too caught up in the IR semantics for small types.

(The IR semantic of `targetType == TYP_USHORT` is indeed that the true target type is `TYP_INT`)

So probably it will actually be simple enough after all - if we're truncating, truncate, if we're extending, look at the signedness for the small type.

yowl: Ok, thanks.  Let me redo it for the second approach.  Keeping the LLVM small and efficient is a goal for me as Mono already have a LLVM story, so the point here is to see if going this RyuJIT approach can match it or do better.  Anything that improves the chances of the LLVM being efficient is good for me.

yowl: Well, I'm not sure now.  Having run the compiler again, I realize that a common cause of having to extend is the result of a comparison, e.g. `GT_EQ`.  In IR thats an `INT` type, but in LLVM its an `i1`.  In LLVM this is valid:

```
 %beq = icmp eq i32 %Loadloc0_, 23
  br i1 %beq, label %Block14, label %BlockD
```
(and is emitted by the IL backend)
In LLVM `i1` is the only type permitted by a conditional branch https://llvm.org/docs/LangRef.html#br-instruction so I haven't normalised the results of comparison operators in case they would just have to be `trunc`ed for the next instruction.  I can still do the suggestion, more or less, but it would be based on I suppose IR type, and then if that's the same, then the LLVM type of the operators.

SingleAccretion: Would changing `targetType` to be an LLVM type solve the issue?

yowl: I don't think, but then I dont think there is much of a problem.  I changed to 
```

void Llvm::buildBinaryOperation(GenTree* node)
{
    Value* result;

    Value* op1 = consumeValue(node->gtGetOp1(), node->TypeGet());
    Value* op2 = consumeValue(node->gtGetOp2(), node->TypeGet());
```

and 

```
Value* Llvm::consumeValue(GenTree* node, var_types targetType)
{
    Value* nodeValue = getGenTreeValue(node);
    if (node->TypeGet() != targetType)
    {
        // this is never hit
        nodeValue = nodeValue;
    }

    Type* targetLlvmType = getLlvmTypeForVarType(targetType);

    llvm::TypeSize intSize = nodeValue->getType()->getPrimitiveSizeInBits();
    if (targetLlvmType->isIntegerTy() && intSize == 1)
    {
        // this is hit, so just special case i1 as that is the LLVM result type of a logical operation?
        nodeValue = _builder.CreateIntCast(nodeValue, Type::getInt32Ty(_llvmContext), false);
    }
    return nodeValue;
}
```
The only possible widening that I can make happen is from `i1` to `i32` which occurs due to the difference between IR and LLVM for the result of a logical comparison.  E.g. For this
```
    private static void TestDifferentSizeIntOperator(short s)
    {
        StartTest("Logical and short and int");

        int i = 4;

        EndTest((s & i) != 0);
    }
```
You get (snipped)
```
N004 (  4,  4) [000020] ------------        t20 = *  CAST      int <- short <- int
N005 (  3,  2) [000009] ------------         t9 =    LCL_VAR   int    V01 loc0         u:2 (last use)
                                                  /--*  t20    int
                                                  +--*  t9     int
N006 (  8,  7) [000010] ------------        t10 = *  AND       int#
```
There is a `CAST` anyway to handle the difference size.  If you try with a local than it starts with an IL `ldc.i4` so its an int32 anyway.

Would it be better just to special case the LLVM/IR difference in logical op return values, ie `i1` and fail on anything else?

SingleAccretion: >     if (node->TypeGet() != targetType)
>     {
>        // this is never hit
>        nodeValue = nodeValue;
>     }

Hmm, I don't think this is true. `ADD(IND short, IND byte)` would be the canonical example.

Of course now it is not hit because loads use the "normalize on produce" semantic:
https://github.com/dotnet/runtimelab/blob/be8db1725c4c3c5a35edd93f718193ace43c7b73/src/coreclr/jit/llvm.cpp#L1116-L1122

If we go with the "normalize on load" semantic (aka option 2), the `buildInd` code will be changed to produce `int16`s and `int8`s in terms of LLVM types. I do not think it would be a good idea to have a mix of normalize on load and normalize on produce, it seems be hard to reason about.

> Would it be better just to special case the LLVM/IR difference in logical op return values, ie i1 and fail on anything else?

Do we need to? In `consumeValue` we have, essentially, 3 types: the LLVM type of the value the node produces, the IR type and the target LLVM type.

Essentially, I think, the logic should be:
```c++
if (valueType != targetType) // Both are LLVM types
{
    assert(/* both are integral types less than or 32 bits in size */);
    if (sizeof(valueType) < sizeof(targetType))
    {
       // Upcast.
       if (varTypeIsSmall(node))
       {
           finalValue = varTypeIsSigned(node) ? sext(valueType to targetType) : zext(valueType to targetType);
       }
       else
       {
           // This is the special case for relops. Ordinary codegen "just knows" they need zero-extension.
           assert(valueType == i1);
           finalValue = zext(i1 to i32);
       }
    }
    else
    {
        // Truncate.
        finalValue = trunc(valueType to targetType);
    }
}
```
This should then handle all scenarios we care about: `JTRUE(RELOP)`, `CALL(small arg SMALL_NODE`, etc.

yowl: Thanks for the code, I allowed int64 types so that I could use it for shifts also.

SingleAccretion: Not sure I'd put the int32 -> in64 upcast there, seems prone to hiding type mismatches, but if you think that's better I have no objections.

src/coreclr/jit/llvm.cpp
```diff
@@ -755,6 +756,19 @@ Value* Llvm::genTreeAsLlvmType(GenTree* tree, Type* type)
         }
         return _builder.getInt({(unsigned int)type->getPrimitiveSizeInBits().getFixedSize(), (uint64_t)tree->AsIntCon()->IconValue(), true});
     }
+    //// from i1 (e.g. from GT_EQ) to wider int
+    if (v->getType()->isIntegerTy() && type->isIntegerTy() &&
+        v->getType()->getPrimitiveSizeInBits().getFixedSize() < type->getPrimitiveSizeInBits().getFixedSize())
+    {
+        return _builder.CreateZExt(v, type);
+    }
```
SingleAccretion: Still needed? It seems like `genTreeAsLlvmType` is similar to `consumeValue`, but different. Put the logic for bitcasting pointers to integers and back into `consumeValue` as well?

src/coreclr/jit/llvm.cpp
```diff
@@ -1223,6 +1271,81 @@ void Llvm::fillPhis()
     }
 }
 
+void Llvm::buildUnaryOperation(GenTree* node, Value* op1)
```
SingleAccretion: This also needs to use `consumeValue` (and `op1` moved into the body like the others).

src/coreclr/jit/llvm.cpp
```diff
@@ -814,6 +828,40 @@ llvm::Value* Llvm::getShadowStackForCallee()
     return offset == 0 ? _function->getArg(0) : _builder.CreateGEP(_function->getArg(0), _builder.getInt32(offset));
 }
 
+Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
```
SingleAccretion, line 822:
```suggestion
//------------------------------------------------------------------------
// consumeValue: Get the Value* "node" produces when consumed as "targetLlvmType".
//
// During codegen, we follow the "normalize on demand" convention, i. e.
// the IR nodes produce "raw" values that have exactly the types of nodes,
// preserving small types, pointers, etc. However, the user in the IR
// consumes "actual" types, and this is the method where we normalize
// to those types. We could have followed the reverse convention and
// normalized on production of "Value*"s, but we presume the "on demand"
// convention is more efficient LLVM-IR-size-wise. It allows us to avoid
// situations where we'd be upcasting only to immediately truncate, which
// would be the case for small typed arguments and relops feeding jumps,
// to name a few examples.
//
// Arguments:
//    node           - the node for which to obtain the normalized value of
//    targetLlvmType - the LLVM type through which the user uses "node"
//
// Return Value:
//    The normalized value, of "targetLlvmType" type.
//
Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
```
To document the rationale here. Feel free not to include if you feel it is too verbose / noisy, etc.

src/coreclr/jit/llvm.cpp
```diff
@@ -1316,6 +1439,12 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
         llvmRef = _builder.CreateTrunc(llvmRef, Type::getInt32Ty(_llvmContext));
     }
 
+    // implicit casting from * to int
+    else if (llvmRef->getType()->isPointerTy() && getLlvmTypeForVarType(lclVar->TypeGet()) == Type::getInt32Ty(_llvmContext))
+    {
+        llvmRef = _builder.CreatePtrToInt(llvmRef, Type::getInt32Ty(_llvmContext));
+    }
+
```
SingleAccretion: Should this too be the job of `consumeValue`?

src/coreclr/jit/llvm.cpp
yowl: Good if you ask me.

yowl:
The scope crept with the latest commit because removing genTreeAsLlvmType meant that some situations that failed compilation where now being attempting and some bugs were uncovered.

The math overflow helpers are all managed so they needed to be identified as such

The return arg and this arg need to be identifed when creating the call as the LLVM signature is built from the `CORINFO_SIG_INFO`.  I expect this will get cleared up when, or after, the PUTARG work is done.

Ctors returning `this` need to be excluded currently for the same reason: the CORINFO_SIG_INFO shows those methods as `void`.

There seems to be a difference in phi rules between IR and LLVM (https://github.com/dotnet/runtimelab/issues/1731) so I removed the restriction on the local only appearing once in the phi args.

src/coreclr/jit/llvm.cpp
yowl: Agree, best not to conflate the issues.

yowl:
On the plus side, coverage is up to around 22%

src/coreclr/jit/llvm.cpp
```diff
@@ -906,7 +994,12 @@ FunctionType* Llvm::buildHelperLlvmFunctionType(GenTreeCall* call, bool withShad
 
 bool Llvm::helperRequiresShadowStack(CORINFO_METHOD_HANDLE corinfoMethodHnd)
 {
-    return corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPEHANDLE);
+    //TODO-LLVM: is there a better way to identify managed helpers?
+    return corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPEHANDLE) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2INT_OVF) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2LNG_OVF) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2UINT_OVF) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2ULNG_OVF);
```
SingleAccretion, line 1004: A note could be left that we'll likely want to lower these specifically into "ordinary" casts (and let LLVM deal with them as it sees fit), to not have to deal with the call setup overheads.

src/coreclr/jit/llvm.cpp
```diff
@@ -1223,6 +1316,87 @@ void Llvm::fillPhis()
...
+void Llvm::buildShift(GenTreeOp* node)
+{
+    Type*  llvmTargetType = getLlvmTypeForVarType(node->TypeGet());
+    Value* numBitsToShift = consumeValue(node->gtOp2, getLlvmTypeForVarType(node->gtOp2->TypeGet()));
+
+    // the LLVM docs say that both operands must be the same type and a compilation failure results if this is not the case.
+    if (numBitsToShift->getType()->getPrimitiveSizeInBits() < llvmTargetType->getPrimitiveSizeInBits())
+    {
+        numBitsToShift = _builder.CreateZExt(numBitsToShift, llvmTargetType);
+    }
```
```suggestion
    // LLVM requires the operands be the same type as the shift itself.
    // Shift counts are assumed to never be negative, so we zero extend.
    if (numBitsToShift->getType()->getPrimitiveSizeInBits() < llvmTargetType->getPrimitiveSizeInBits())
    {
        numBitsToShift = _builder.CreateZExt(numBitsToShift, llvmTargetType);
    }
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1686,10 +1874,29 @@ void Llvm::ConvertShadowStackLocals()
...
+                    // We come here also for ctors like the following
+                    //                                         /--*  t18    int    arg0 in rcx
+                    // N003(17, 5)[000019]-- CXG-- -----t19 = *CALL ref System.String..ctor
+                    failFunctionCompilation();
+                }
+
+                var_types callReturnType = callNode->TypeGet();
+                // ctors in CORINFO_SIG_INFO are void, but in IR are TYP_REF.  ILtoLLVMImporter does not have this problem as it uses MethodSignature
+                // which has the return type.  TODO-LLVM: is it better to create a new EE callback that matches IR/MethodSignature regards return types for ctors?
+                if (callReturnType == TYP_REF || needsReturnStackSlot(calleeSigInfo.retType, calleeSigInfo.retTypeClass))
```
SingleAccretion: Hmm, I looked around and I think this very special case (non-`VOID` ctors) only applies to strings, i. e.  structs for which we'd need to fetch the handle from somewhere always have "proper" constructors. So I am inclined to say this is fine as-is (possible simplification is to always pass the the first parameter to `needsReturnStackSlot` as `toCorInfoType(callReturnType)`).

SingleAccretion: (And add a comment)

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -619,8 +619,9 @@ void SsaBuilder::InsertPhiToRationalIRForm(BasicBlock* block, unsigned lclNum)
 void SsaBuilder::AddPhiArg(
     BasicBlock* block, Statement* stmt, GenTreePhi* phi, unsigned lclNum, unsigned ssaNum, BasicBlock* pred)
 {
-#ifdef DEBUG
+#if defined(DEBUG) && !defined(TARGET_WASM)
     // Make sure it isn't already present: we should only add each definition once.
+    // Except for Wasm/LLVM where every predecessor must have its own phi arg
```
SingleAccretion: It would be cleaner to exclude this only "for the backed", i. e. something like `defined(TARGET_WASM)` + `block->IsLIR()`, since it is still an always true invariant for the front-end SSA.

src/coreclr/jit/llvm.cpp
yowl: Thanks, I'll do this one and leave the other calls to `needsReturnStackSlot` as is because I think some or all of them will go when the args are lowered.

src/coreclr/jit/ssabuilder.cpp
yowl: Thanks, will do.

src/coreclr/jit/llvm.cpp
yowl: ok, added a comment, assume that we are talking about lowering to `GT_CAST`s

SingleAccretion (approved):
Looks great!

jkotas (approved):
Thank you @yowl and @SingleAccretion !

PR #1706, opened 2021-11-06
Use relative pointers for module initializers section

### LINKED ISSUE #1704: [NativeAOT] cannot preempt symbol: Program__Module__cctor
I'm getting the following lld linker error when attempting to build any .NET assembly with module initializers for the Switch platform (linux-arm64):
`cannot preempt symbol: Program__Module____cctor`

These errors appear for both the main assembly and every referenced assembly that contains a module initializer. (`FNA__Module__cctor`, for example.) If I remove the module initializers, then the build has no errors.

The same project built correctly as of early September, so this seems to be either a regression or a side effect of the new module initializer logic implemented in NativeAOT.

I'm aware I'm in unsupported territory targeting an exotic platform, but it's possible this might be a bug on linux-arm64 in general. I don't have the infrastructure to test though, so I'm not certain. If this is something potentially specific to the toolchain I'm using, is there any clear way to start debugging this?

jkotas:
Fixes #1704


src/coreclr/nativeaot/Common/src/Internal/Runtime/CompilerHelpers/StartupCodeHelpers.cs
```diff
@@ -218,12 +218,25 @@ internal static unsafe void RunModuleInitializers()
...
-                    var cctorsEnd = (void*)(moduleInitializerSection + length);
-                    while (currentCctor < cctorsEnd)
+                    var current = (byte*)moduleInitializerSection;
+                    var end = current + length;
+                    while (current < end)
                     {
-                        (*currentCctor++)();
+                        void* cctor;
+
+                        if (Internal.Runtime.MethodTable.SupportsRelativePointers)
```
jkotas: An alternative solution would be to make the section writeable on non-Windows. (The sections with absolute pointers have to be writeable on non-Windows.)

MichalStrehovsky (approved):
It would be nice if this was consistent with how we do eager constructors (they're all pointer sized) but since we already have this code, LGTM.

MichalStrehovsky:
> It would be nice if this was consistent with how we do eager constructors (they're all pointer sized) but since we already have this code, LGTM.

Oh, looks like we just put those into Data section no matter what. It probably doesn't matter much because they're never looked at once we're done starting up.

PR #1710, opened 2021-11-09
Improve failure mode of recursive generic lookups

MichalStrehovsky:
Throw a `TypeLoadException` instead of a wild null dereference that may or may not translate to `NullReferenceException` when generic recursion cutoff gets hit at runtime.

* Remember there was a dictionary associated with a ReadyToRun generic lookup helper that has a null slot.
* Add an extra null check to affected helpers that tailcall a throw helper.
* Remember we had to do this during scanning and make sure the same thing happens in a compilation after scanning (i.e. RyuJIT doesn't need to be aware of this because we'll just keep piping these through the ReadyToRun helper even in optimized builds).

I didn't do ARM64 because I can't test this. We'll see a failing test once we enable ARM64 testing. This is a corner situation that I'm not too worried about.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Resources/Strings.resx
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Resources/Strings.resx
@@ -198,6 +198,9 @@
   <data name="Arg_DecBitCtor" xml:space="preserve">
     <value>Decimal byte array constructor requires an array of length four containing valid decimal bytes.</value>
   </data>
+  <data name="Arg_UnavailableTypeLoadException" xml:space="preserve">
+    <value>Attempted to load a type that couldn't be loaded during ahead of time compilation.</value>
+  </data>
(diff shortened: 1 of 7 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/Resources/Strings.resx
```diff
@@ -198,6 +198,9 @@
   <data name="Arg_DecBitCtor" xml:space="preserve">
     <value>Decimal byte array constructor requires an array of length four containing valid decimal bytes.</value>
   </data>
+  <data name="Arg_UnavailableTypeLoadException" xml:space="preserve">
+    <value>Attempted to load a type that couldn't be loaded during ahead of time compilation.</value>
```
jkotas:
```suggestion
    <value>Attempted to load a type that was not created during ahead of time compilation.</value>
```

jkotas (approved):
Thanks!

PR #1714, opened 2021-11-10
Speed up IsInstanceOfClass

MichalStrehovsky:
Gets rid of building a frame around the method, nicely speeding it up.

We'll have to rearrange these to be more like in CoreCLR proper. E.g. CoreCLR doesn't seem to be handling variance in this codepath.


jkotas (approved):

jkotas:
> We'll have to rearrange these to be more like in CoreCLR proper. E.g. CoreCLR doesn't seem to be handling variance in this codepath.

Yes. Also, CoreCLR has fine-tuned cache for the slow path.

PR #1716, opened 2021-11-11
[NativeAOT-LLVM] Add field to LclVarDsc to represent LLVM arg index

yowl:
WIP

This PR adds a node to represent LLVM args.  All arguments that are passed on the LLVM stack: e.g the shadowstack, non gc tracked IL args, will be have an `ARGLLVM` node that will be created and replaced into the IR as part of lowering to LLVM.  This will simplify the codegen phase.

For https://github.com/dotnet/runtimelab/blob/a87e9dd7a2d7f50cf9f55bbb346cc52613c0cf0e/src/coreclr/nativeaot/Runtime.Base/src/System/AttributeUsageAttribute.cs#L36 it produces
```
               [000012] ------------        t12 =    ARGLLVM   int    index: 0
                                                  /--*  t12    int
               [000013] DA----------              *  STORE_LCL_VAR int    V03 tmp1         d:2

------------ BB02 [000..008) (return), preds={BB01} succs={}
               [000009] ------------                 IL_OFFSET void   IL offset: 0x0
               [000014] ------------        t14 =    CNS_INT   int    0
               [000015] ------------        t15 =    LCL_VAR   int    V03 tmp1         u:2 (last use)
                                                  /--*  t15    int
                                                  +--*  t14    int
               [000016] ------------        t16 = *  ADD       int
                                                  /--*  t16    int
N001 (  3,  2) [000001] ------------         t1 = *  IND       ref
N002 (  1,  1) [000006] ------------         t6 =    CNS_INT   int    8 field offset Fseq[_allowMultiple]
                                                  /--*  t1     ref
                                                  +--*  t6     int
N003 (  4,  3) [000007] -------N----         t7 = *  ADD       byref
N005 (  3,  2) [000002] ------------         t2 =    LCL_VAR   int    V01 arg1         u:1 (last use)
                                                  /--*  t7     byref
                                                  +--*  t2     int
               [000010] -A-XG-------              *  STOREIND  bool
               [000011] ------------                 IL_OFFSET void   IL offset: 0x7
N001 (  0,  0) [000005] ------------                 RETURN    void
```

In `BB01` the arg node is created and stored into a local, `V03`.  Then for the `ADD` node at `t16` the local is used to generate the shadow stack offset for the `this`  arg.

WIP as lots to do, but @SingleAccretion, does this look like the IR you anticipated?

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -459,6 +459,7 @@ bool canStoreLocalOnLlvmStack(LclVarDsc* varDsc)
     return !varDsc->HasGCPtr();
 }
 
+//TODO-LLVM: delete this function when we lower the args, its confusing with canStoreLocalOnLlvmStack regards: is the logic exactly the same?
 bool Llvm::canStoreArgOnLlvmStack(CorInfoType corInfoType, CORINFO_CLASS_HANDLE classHnd)
 {
     // structs with no GC pointers can go on LLVM stack.
@@ -467,7 +468,7 @@ bool Llvm::canStoreArgOnLlvmStack(CorInfoType corInfoType, CORINFO_CLASS_HANDLE 
         // Use getClassAtribs over typGetObjLayout because EETypePtr has CORINFO_FLG_GENERIC_TYPE_VARIABLE? which fails with typGetObjLayout
         uint32_t classAttribs = _info.compCompHnd->getClassAttribs(classHnd);
 
-        return (classAttribs & CORINFO_FLG_CONTAINS_GC_PTR) == 0;
+        return (classAttribs & (CORINFO_FLG_CONTAINS_GC_PTR | CORINFO_FLG_CONTAINS_STACK_PTR)) == 0;
     }
 
     if (corInfoType == CorInfoType::CORINFO_TYPE_BYREF || corInfoType == CorInfoType::CORINFO_TYPE_CLASS ||
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -115,18 +115,6 @@ Value* Llvm::mapGenTreeToValue(GenTree* genTree, Value* valueRef)
 
 Value* Llvm::getGenTreeValue(GenTree* op)
 {
-    if (op->IsLocal())
-    {
-        unsigned lclNum = op->AsLclVar()->GetLclNum();
-        if (lclNum == _shadowStackLclNum)
-        {
-            return _function->getArg(0);
-        }
-        else if (lclNum == _retAddressLclNum)
-        {
-            return _function->getArg(1);
-        }
-    }
     return _sdsuMap->at(op);
 }
 
@@ -491,6 +479,35 @@ CorInfoType Llvm::getCorInfoTypeForArg(CORINFO_SIG_INFO& sigInfo, CORINFO_ARG_LI
 {
     CorInfoTypeWithMod corTypeWithMod = _getArgTypeIncludingParameterized(_thisPtr, &sigInfo, arg, clsHnd);
     return strip(corTypeWithMod);
+}
+
+FunctionType* Llvm::getFunctionType()
+{
+    // TODO-LLVM: delete this when these signatures implemented
+    if (_sigInfo.hasExplicitThis() || _sigInfo.hasTypeArg())
+        failFunctionCompilation();
+
+    std::vector<llvm::Type*> argVec(_llvmArgCount);
+    llvm::Type*              retLlvmType;
+
+    for (unsigned i = 0; i < _compiler->lvaTableCnt; i++)
+    {
+        LclVarDsc varDsc = _compiler->lvaTable[i];
+        if (varDsc.lvIsParam)
+        {
+            assert(varDsc.lvLlvmArgNum != BAD_LLVM_ARG_NUM);
+
+            CorInfoType          corInfoType = toCorInfoType(varDsc.TypeGet());
+            CORINFO_CLASS_HANDLE classHandle = tryGetStructClassHandle(&varDsc);
+            argVec[varDsc.lvLlvmArgNum]      = getLlvmTypeForCorInfoType(corInfoType, classHandle);
+        }
+    }
+
+    retLlvmType = _retAddressLclNum == BAD_VAR_NUM
+        ? getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass)
+        : Type::getVoidTy(_llvmContext);
+
+    return FunctionType::get(retLlvmType, ArrayRef<Type*>(argVec), false);
 }
 
 FunctionType* Llvm::getFunctionTypeForSigInfo(CORINFO_SIG_INFO& sigInfo)
@@ -1445,35 +1462,17 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
     unsigned int ssaNum = lclVar->GetSsaNum();
     if (_localsMap->find({lclNum, ssaNum}) == _localsMap->end())
     {
-        if (_compiler->lvaIsParameter(lclNum))
+        LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar);
+
+        if (varDsc->lvLlvmArgNum != BAD_LLVM_ARG_NUM)
         {
-            if (lclNum == _shadowStackLclNum)
+            // TODO-LLVM: we dont handle this yet
+            if (_info.compRetBuffArg != BAD_VAR_NUM)
             {
-                llvmRef = _function->getArg(0);
-            }
-            else if (lclNum == _retAddressLclNum)
-            {
-                llvmRef = _function->getArg(1);
-            }
-            else
-            {
-                // adjust for "this" (on shadowstack, not an arg)
-                unsigned int argIx = _info.compIsStatic ? lclNum : lclNum - 1;
-
-                // adjust for return address arg, not in siginfo
-                if (_functionSigHasReturnAddress)
-                {
-                    argIx++;
-                }
-
-                if (_info.compRetBuffArg != BAD_VAR_NUM)
-                {
-                    failFunctionCompilation();
-                }
-                LlvmArgInfo  llvmArgInfo = getLlvmArgInfoForArgIx(_sigInfo, argIx);
-                llvmRef = _function->getArg(llvmArgInfo.m_argIx);
+                failFunctionCompilation();
             }
 
+            llvmRef = _function->getArg(varDsc->lvLlvmArgNum);
             _localsMap->insert({{lclNum, ssaNum}, llvmRef});
         }
         else
@@ -1773,7 +1772,6 @@ Llvm::Llvm(Compiler* pCompiler)
       _retAddressLclNum(BAD_VAR_NUM)
 {
     _compiler->eeGetMethodSig(_info.compMethodHnd, &_sigInfo);
-    _functionSigHasReturnAddress = needsReturnStackSlot(_sigInfo.retType, _sigInfo.retTypeClass);
 }
 
 void Llvm::llvmShutdown()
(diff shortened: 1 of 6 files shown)


SingleAccretion:
> does this look like the IR you anticipated?

Hmm, I will admit that is now what I had in mind, since I was thinking about how the handling of callees should be done. Basically the problem is that for this C#:
```cs
static void Method(SmallStruct s1, StructWithLong s2, object obj) { }

SmallStruct { short SmallInteger; }
StructWithLong { long Long; }
```
You can end up with something like this in IR:
```
CALL(int, long, ref);
```
And for LLVM we want this:
```
CALL(ShadowStack, SmallStruct, StructWithLong);
```
So the solution was to
1) wrap the argument nodes into `PUTARG_TYPE` nodes that would carry their true types:
```
CALL(PUTARG_TYPE(SmallStruct (int)), PUTARG_TYPE(StructWithLong (long)), PUTARG_TYPE(object (object)));
```
2) Rewrite all arguments that would be needed to be passed on the shadow stack:
```
STOREIND(VSS + OFFSET, object)

CALL(PUTARG_TYPE(nint (VSS + OFFSET)), PUTARG_TYPE(SmallStruct (int)), PUTARG_TYPE(StructWithLong (long)));
```
So by the time codegen comes along, it could simply loop all over the `PUTARG_TYPE`s (each node would also carry an argument number so that they can be built in the right order) to build the signature and emit he necessary bitcasts/upcasts/downcasts.

What is being solved here though is a somewhat different problem: how to represent things not in the signature of _the caller_. The crux of the issue is basically in this logic:
https://github.com/dotnet/runtimelab/blob/a80480e085e61376ec3d3865d49c5a9592245710/src/coreclr/jit/llvm.cpp#L492-L531

I. e. we would like to map between the LLVM's arguments and the IR ones. Ideally this should be shareable with callee logic, but that is probably not something I would worry about for now. What about this then: add a field to `LclVarDsc`, say `lvLlvmArgNum`, that would identify its position, and fill those in when building the signature (before lowering say, so that it also adds locals for the shadow stack and others)? Then we may yet evade a need for a custom node.

Codegen would then also be able to walk locals that are `lvParam` and build the signature easily.

yowl:
Thanks, regards the callee and `PUTARG_TYPE`, sounds good, I'll get to it after doing the "caller", thanks.

I've removed the new node and added `lvLlvmArgNum` with
```cs
void Llvm::populateLlvmArgNums()
{
    _shadowStackLclNum = _compiler->lvaGrabTemp(true DEBUGARG("shadowstack"));
    LclVarDsc* shadowStackVarDsc = _compiler->lvaGetDesc(_shadowStackLclNum);
    unsigned   nextLlvmArgNum    = 0;

    shadowStackVarDsc->lvLlvmArgNum = nextLlvmArgNum++;
    shadowStackVarDsc->lvType    = TYP_I_IMPL;
    shadowStackVarDsc->lvIsParam = true;

    if (needsReturnStackSlot(_sigInfo.retType, _sigInfo.retTypeClass))
    {
        _retAddressLclNum = _compiler->lvaGrabTemp(true DEBUGARG("returnslot"));
        LclVarDsc* retAddressVarDsc  = _compiler->lvaGetDesc(_retAddressLclNum);
        retAddressVarDsc->lvLlvmArgNum = nextLlvmArgNum++;
        retAddressVarDsc->lvType       = TYP_I_IMPL;
        retAddressVarDsc->lvIsParam    = true;
    }

    // TODO-LLVM: other non-standard args, generic context, outs

    CORINFO_ARG_LIST_HANDLE sigArgs = _sigInfo.args;

    for (unsigned int i = 0; i < _sigInfo.numArgs; i++, sigArgs = _info.compCompHnd->getArgNext(sigArgs))
    {
        CORINFO_CLASS_HANDLE classHnd;
        CorInfoType          corInfoType = getCorInfoTypeForArg(_sigInfo, sigArgs, &classHnd);
        LclVarDsc*           varDsc      = _compiler->lvaGetDesc(i);
        if (canStoreArgOnLlvmStack(corInfoType, classHnd))
        {
            varDsc->lvLlvmArgNum = nextLlvmArgNum++;
        }
    }

    _llvmArgCount = nextLlvmArgNum;
}
```
Generating the function signature for the caller then looks like
```cs
FunctionType* Llvm::getFunctionType()
{
    // TODO-LLVM: delete this when these signatures implemented
    if (_sigInfo.hasExplicitThis() || _sigInfo.hasTypeArg())
        failFunctionCompilation();

    std::vector<llvm::Type*> argVec(_llvmArgCount);
    llvm::Type*              retLlvmType;

    for (unsigned i = 0; i < _compiler->lvaTableCnt; i++)
    {
        LclVarDsc varDsc = _compiler->lvaTable[i];
        if (varDsc.lvIsParam)
        {
            assert(varDsc.lvLlvmArgNum != BAD_LLVM_ARG_NUM);

            CorInfoType          corInfoType = toCorInfoType(varDsc.TypeGet());
            CORINFO_CLASS_HANDLE classHandle = tryGetStructClassHandle(&varDsc);
            argVec[varDsc.lvLlvmArgNum]      = getLlvmTypeForCorInfoType(corInfoType, classHandle);
        }
    }

    retLlvmType = _retAddressLclNum == BAD_VAR_NUM
        ? getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass)
        : Type::getVoidTy(_llvmContext);

    return FunctionType::get(retLlvmType, ArrayRef<Type*>(argVec), false);
}
```

yowl:
I've squashed the commits and removed what I think we can now we have `lvLlvmArgNum`.

src/coreclr/jit/llvm.cpp
```diff
@@ -493,6 +481,35 @@ CorInfoType Llvm::getCorInfoTypeForArg(CORINFO_SIG_INFO& sigInfo, CORINFO_ARG_LI
...
+    // TODO-LLVM: delete this when these signatures implemented
+    if (_sigInfo.hasExplicitThis() || _sigInfo.hasTypeArg())
+        failFunctionCompilation();
+
+    std::vector<llvm::Type*> argVec(_llvmArgCount);
+    llvm::Type*              retLlvmType;
+
+    for (unsigned i = 0; i < _compiler->lvaTableCnt; i++)
+    {
+        LclVarDsc varDsc = _compiler->lvaTable[i];
```
```suggestion
        LclVarDsc* varDsc = _compiler->lvaGetDesc(i);
```
`LclVarDsc`s are always addressed though pointers as they are rather large and expensive to copy.

src/coreclr/jit/llvm.cpp
```diff
@@ -1798,6 +1796,51 @@ void Llvm::llvmShutdown()
...
+    unsigned   nextLlvmArgNum    = 0;
+
+    shadowStackVarDsc->lvLlvmArgNum = nextLlvmArgNum++;
+    shadowStackVarDsc->lvType    = TYP_I_IMPL;
+    shadowStackVarDsc->lvIsParam = true;
+
+    if (_sigInfo.retType == CorInfoType::CORINFO_TYPE_VOID && _info.compRetType != TYP_VOID)
+    {
+        needsReturnStackSlot(_sigInfo.retType, _sigInfo.retTypeClass);
+    }
```
SingleAccretion: Hmm, `needsReturnStackSlot` is a pure function - what does this do then?

src/coreclr/jit/llvm.cpp
```diff
@@ -493,6 +481,35 @@ CorInfoType Llvm::getCorInfoTypeForArg(CORINFO_SIG_INFO& sigInfo, CORINFO_ARG_LI
     return strip(corTypeWithMod);
 }
 
+FunctionType* Llvm::getFunctionType()
```
SingleAccretion, line 485: Hmm, shouldn't we call it in `Llvm::Compile` (right now it is unused)?

Nit: I would name it something like `createFunctionTypeForCompilation`, to signify it is to be called only once (and is meant to be used for the function being compiled).

src/coreclr/jit/llvm.cpp
```diff
@@ -493,6 +481,35 @@ CorInfoType Llvm::getCorInfoTypeForArg(CORINFO_SIG_INFO& sigInfo, CORINFO_ARG_LI
...
+{
+    // TODO-LLVM: delete this when these signatures implemented
+    if (_sigInfo.hasExplicitThis() || _sigInfo.hasTypeArg())
+        failFunctionCompilation();
+
+    std::vector<llvm::Type*> argVec(_llvmArgCount);
+    llvm::Type*              retLlvmType;
+
+    for (unsigned i = 0; i < _compiler->lvaTableCnt; i++)
```
```suggestion
    for (unsigned i = 0; i < _compiler->lvaCount; i++)
```

src/coreclr/jit/llvm.cpp
yowl: Lol, yes, I put that in just to have some code to put the breakpoint, I wanted to see if that condition ever happened and forgot to delete it.

src/coreclr/jit/llvm.cpp
yowl: Ah sorry, I now realise there is a problem with this method in that it can't distinguish between `TYP_INT` and `TYP_I_IMPL` (being a macro to `TYP_INT`, and my `i8*` args end up being `i32` ,  `toCorInfoType` maps them both to `CORINFO_TYPE_INT`.  Is it ok to use `lvIsPtr` to mark the actual pointers ?

SingleAccretion: No... it is only used by the GS security code for some very specific purposes.

What you can do is add some more metadata under `#ifdef TARGET_WASM`, that preserves "the original type". Look for `lvaInitVarDsc` in `lclvars.cpp`.

SingleAccretion: I was writing this assuming we need the actual full type, not just `i8*` for pointers. If we don't, then `lvIsPtr` should be enough.

yowl: Ultimately, we are going to need to build `IDynamicCastableGetInterfaceImplementation(i8* %0, %"[S.P.CoreLib]Internal.Runtime.MethodTable"* %1, i16 %2)`  so I think we'll need the full type.  I don't know that we are currently covering any of these signatures.

yowl:
As usual, another problem is uncovered: `ref structs`  Take this c#
```cs
        public ValueStringBuilder(Span<char> initialBuffer)
        {
```

Currently this generates in the IL -> LLVM backend
```llvm
define void @S_P_CoreLib_System_Text_ValueStringBuilder___ctor(i8* %0, %"[S.P.CoreLib]System.Span`1<char>" %1)
```
Which I think is wrong in that it's a copy, but it's right in placing it on the LLVM stack.  However in the IR backend, to test for LLVM stack capability we just have
```cpp
bool canStoreLocalOnLlvmStack(LclVarDsc* varDsc)
{
    return !varDsc->HasGCPtr();
}
```
which in this case, it has gc pointers, so its false.  My understanding is ref structs are never on the managed heap, so never need to be GC tracked.  Can I identify ref structs somehow?

SingleAccretion:
> Which I think is wrong in that it's a copy

I think it is correct. Structs, regardless of being `ref` are passed by copying.

>  My understanding is ref structs are never on the managed heap, so never need to be GC tracked. Can I identify ref structs somehow?

The struct itself is never on the heap, but its constituent fields can be still be GC pointers that need tracking. In essense, `ref struct`s are just normal structs with the exception that you will never see them as the results of loads from pointers in the GC heap (they can still be loaded from "other" global or local pointers though, such as if you pass them via `ref`).

SingleAccretion:
So, to summarize, the span is to be passed on the shadow stack, just as something like `struct { object, int }`.

yowl:
Thanks, so the IL backend is wrong, I guess the best way forward is to fix it.

yowl:
The `Span<T>` problem is being fixed in the IL in #1742   The latest commit merges that change, so best wait on that PR before moving this one forward again.  Also added a workaround for #1729 as that's getting hit now.  Will look at lowering these locals and `LCL_VAR_ADDR` after PUTARG work.

yowl:
I've put the alternative for checking for structs with byref like test in here due to the problem in https://github.com/dotnet/runtimelab/pull/1742#discussion_r753861338

SingleAccretion (approved):
Looks good!

jkotas (approved):
Thanks!

PR #1719, opened 2021-11-12
Speed up access to threadstatic fields

MichalStrehovsky:
Accessing thread statics requires calling a method in C# (because we might need to do an allocation) followed by calling a method in C++ (because that's where we keep thread-related datastructures). Ideally we would want to do the inverse - call into C++ that has access to everything and only call the C# part on the slow path. But for that we need C++ to support guaranteed tailcalls and that's not going to happen anytime soon. JIT-based CoreCLR has the luxury (and curse) of being able to switch to managed world in C++ and gets away with a single call. Our only option seem to be inline assembly.

Speeds up accessing threadstatics by 50%.

Opening as a draft only to get @jkotas opinion - looks like this is a bit more complex on Unix because INLINE_GETTHREAD is an actual method call there and I'm not sure it's worth the hassle. Eventually it would be nice if threadstatics could be handled the same how they're done in C++. Unclear if we need this middle ground. I guess it would help Techempower Json benchmark.


jkotas:
I do not have a strong opinion. I agree with what you said. It would be nice to get onto the C/C++ plan, but that would require teaching the JIT to emit the OS-specific TLS access sequences.

jkotas:
> this is a bit more complex on Unix because INLINE_GETTHREAD is an actual method call there

Note that the call gets replaces by inline assembly by the ELF loader if you emit the right fixups and the TLS layout is known.

MichalStrehovsky:
> > this is a bit more complex on Unix because INLINE_GETTHREAD is an actual method call there
> 
> Note that the call gets replaces by inline assembly by the ELF loader if you emit the right fixups and the TLS layout is known.

Yup, looks like it's the case:

```asm
Dump of assembler code for function RhpGetThreadStaticBaseForType:
=> 0x00000000004b9ff7 <+0>:     push   %rbx
   0x00000000004b9ff8 <+1>:     push   %r12
   0x00000000004b9ffa <+3>:     mov    %rdi,%rbx
   0x00000000004b9ffd <+6>:     mov    %rsi,%r12
   0x00000000004ba000 <+9>:     data16 data16 data16 mov %fs:0x0,%rax
   0x00000000004ba00c <+21>:    add    $0xfffffffffffffee0,%rax
   0x00000000004ba013 <+28>:    mov    0x8(%rbx),%r8d
   0x00000000004ba017 <+32>:    cmp    0x98(%rax),%r8d
   0x00000000004ba01e <+39>:    jae    0x4ba047 <RhpGetThreadStaticBaseForType+80>
   0x00000000004ba020 <+41>:    mov    0x90(%rax),%r9
   0x00000000004ba027 <+48>:    mov    (%r9,%r8,8),%rax
   0x00000000004ba02b <+52>:    test   %rax,%rax
   0x00000000004ba02e <+55>:    je     0x4ba047 <RhpGetThreadStaticBaseForType+80>
   0x00000000004ba030 <+57>:    mov    (%rax),%r8
   0x00000000004ba033 <+60>:    cmp    0x8(%r8),%r12d
   0x00000000004ba037 <+64>:    jae    0x4ba047 <RhpGetThreadStaticBaseForType+80>
   0x00000000004ba039 <+66>:    mov    0x10(%r8,%r12,8),%rax
   0x00000000004ba03e <+71>:    test   %rax,%rax
   0x00000000004ba041 <+74>:    je     0x4ba047 <RhpGetThreadStaticBaseForType+80>
   0x00000000004ba043 <+76>:    pop    %r12
   0x00000000004ba045 <+78>:    pop    %rbx
   0x00000000004ba046 <+79>:    retq
   0x00000000004ba047 <+80>:    mov    %rbx,%rdi
   0x00000000004ba04a <+83>:    mov    %r12,%rsi
   0x00000000004ba04d <+86>:    pop    %r12
   0x00000000004ba04f <+88>:    pop    %rbx
   0x00000000004ba050 <+89>:    jmpq   0x6f6fc0 <S_P_CoreLib_Internal_Runtime_ThreadStatics__GetThreadStaticBaseForTypeSlow>
```

MichalStrehovsky:
Oh, interesting. I was reading up about it in https://maskray.me/blog/2021-02-14-all-about-thread-local-storage. In the "we're producing an executable" case, we could probably also generate this as `movl %fs:tls_CurrentThread@TPOFF, %eax` instead of the `call __tls_get_addr@PLT` and then we don't need to do the push/pop dance and be afraid that a potential call could stomp over non-volatile registers. We need to do the dance because when producing a shared/static library, a call might actually be needed.

This would shave a couple instructions off this and all the managed object allocation helpers (`RhpNewFast` and friends) as well. Maybe not worth the trouble now, but nice to know about.

jkotas (approved):
It would be nice to do ARM64 ones as well

MichalStrehovsky:
> It would be nice to do ARM64 ones as well

Filed dotnet/runtime#73997. I didn't take the raspberry pi with me and the CI wouldn't test any of that.

PR #1723, opened 2021-11-13
Custom marshaller work.

### LINKED ISSUE #160: Implement ICustomMarshaler
Related tests:

```
Interop/ICustomMarshaler/ConflictingNames/SameNameDifferentAssembly
Interop/ICustomMarshaler/Primitives/ICustomMarshaler_TargetWindows
Interop/PInvoke/CustomMarshalers/CustomMarshalersTest
```

kant2002:
Related to #160


src/coreclr/tools/Common/TypeSystem/Ecma/EcmaSignatureParser.cs
```diff
@@ -616,7 +617,7 @@ public MarshalAsDescriptor ParseMarshalAsDescriptor()
                         _reader.ReadSerializedString();
 
                         // Read managed marshaler name
-                        _reader.ReadSerializedString();
+                        managedMarshallerTypeName = _reader.ReadSerializedString();
 
                         // Read cookie
                         _reader.ReadSerializedString();
```
jkotas: You will need to read the cookie as well. It is argument for the CustomMarshaler GetInstance method.

src/coreclr/tools/Common/TypeSystem/Ecma/EcmaSignatureParser.cs
```diff
@@ -616,7 +617,7 @@ public MarshalAsDescriptor ParseMarshalAsDescriptor()
...
 
                         // Read managed marshaler name
-                        _reader.ReadSerializedString();
+                        managedMarshallerTypeName = _reader.ReadSerializedString();
```
jkotas: I think you should be able to use `_ecmaModule` and `CustomAttributeTypeNameParser.GetTypeByCustomAttributeTypeName` here to resolve the type.

kant2002:
I'm surprised that this is green, since I did not implement cleanup. AAh, now I get it, I run only smoke tests. Have to run unblocked tests locally then :disappointed:

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,67 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+        {
+            throw new NotSupportedException();
+        }
+
+        private void EmitMarshallerCreation(ILCodeStream codeStream, TypeDesc marshallerType)
+        {
+            TypeDesc customMarshallerType = null;
+            foreach (var interfaceType in marshallerType.RuntimeInterfaces)
+            {
+                if (interfaceType.GetFullName() == "System.Runtime.InteropServices.ICustomMarshaler")
```
jkotas: This should check that the type is castable to ICustomMarshaler from the system library instead of comparing by name.

(It is what CoreCLR does.)

jkotas: Also, the methods that actually do the marshaling should be called on the ICustomMarshaler  interface; not by name.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,66 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+        protected override void TransformManagedToNative(ILCodeStream codeStream)
+        {
+            var marshallerType = MarshalAsDescriptor.CustomMarshallerType;
+            TypeDesc customMarshallerType = Context.SystemModule.GetKnownType("System.Runtime.InteropServices", "ICustomMarshaler");
+            if (!marshallerType.CanCastTo(customMarshallerType))
+            {
+                throw new InvalidOperationException($"Custom marshaller type {marshallerType.ToString()} does not implements System.Runtime.InteropServices.ICustomMarshaler");
+            }
+
+            var getInstanceMethod = marshallerType.GetKnownMethod(
```
MichalStrehovsky: Could you switch to `GetMethod`? `GetKnownMethod` will crash the compilation if the method is missing.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,66 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
             codeStream.Emit(ILOpcode.call, emitter.NewToken(helper));
         }
     }
+
+    class CustomTypeMarshaller : Marshaller
+    {
+        protected override void TransformManagedToNative(ILCodeStream codeStream)
+        {
+            var marshallerType = MarshalAsDescriptor.CustomMarshallerType;
```
MichalStrehovsky: We should also abort if the type `IsGenericDefinition`. CoreCLR does something with that. I don't know if it's tested anywhere. It would crash badly if we let that through.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,66 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+            TypeDesc customMarshallerType = Context.SystemModule.GetKnownType("System.Runtime.InteropServices", "ICustomMarshaler");
+            if (!marshallerType.CanCastTo(customMarshallerType))
+            {
+                throw new InvalidOperationException($"Custom marshaller type {marshallerType.ToString()} does not implements System.Runtime.InteropServices.ICustomMarshaler");
```
MichalStrehovsky: This will crash the compiler. Maybe `ThrowHelper.ThrowMarshalDirectiveException();` for now?

src/coreclr/tools/Common/TypeSystem/Interop/MarshalAsDescriptor.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Runtime.CompilerServices;
 using System;
+#nullable enable
```
MichalStrehovsky: Nit: We don't use nullable in the type system. It's fine like that.

It would be of more value to assert in the `CustomMarshallerType`/`Cookie` getters that `Type == NativeTypeKind.CustomMarshaler` because it doesn't make sense to ask for it outside of that. But maybe just deleting this line is enough.

src/coreclr/tools/Common/TypeSystem/Ecma/EcmaSignatureParser.cs
```diff
@@ -616,10 +618,11 @@ public MarshalAsDescriptor ParseMarshalAsDescriptor()
                         _reader.ReadSerializedString();
 
                         // Read managed marshaler name
-                        _reader.ReadSerializedString();
+                        var managedMarshallerTypeName = _reader.ReadSerializedString();
```
MichalStrehovsky, line 618: Could you delete this comment above:

```
                        // There's nobody to consume CustomMarshaller, so let's just parse the data
                        // to avoid asserting later.
```

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
```diff
@@ -569,6 +572,9 @@ public static partial class MarshalHelpers
                     case NativeTypeKind.AnsiBStr:
                         return MarshallerKind.AnsiBSTRString;
 
+                    case NativeTypeKind.CustomMarshaler:
```
MichalStrehovsky: According to CoreCLR sources, CustomMarshaller is also allowed on singledim+multidim arrays and classes (not just System.Object).

It's disallowed for field marshalling (`isField` here).

kant2002:
Okey. Based on tests seems to be GetInstance called just once per single cookie, and persisted across calls. I would assume that  customtype+cookie pair created just once during application lifetime, and then live forever. 

I plan to use `ConditionalWeakTable<TMarshaller, Dictionary<sting, object>>` and additional method `GetInstance<T>(string)` somewhere appropriate which will hide concurrent access to the table there. At least I will proceed with this plan for now. Just want to clarify why I add a bit more new code and ask if this is appropriate in case you are around.

src/coreclr/tools/Common/TypeSystem/Interop/MarshalAsDescriptor.cs
```diff
@@ -48,17 +48,24 @@ public enum NativeTypeKind : byte
 
     public class MarshalAsDescriptor
     {
+        private TypeDesc customMarshallerType;
```
MichalStrehovsky: Instance fields get prefixed with `_`

src/coreclr/tools/Common/TypeSystem/Interop/MarshalAsDescriptor.cs
```diff
@@ -48,17 +48,24 @@ public enum NativeTypeKind : byte
...
+        private string cookie;
+
         public NativeTypeKind Type { get; }
         public NativeTypeKind ArraySubType { get; }
         public uint? SizeParamIndex { get; }
         public uint? SizeConst { get; }
+        public TypeDesc CustomMarshallerType => Type == NativeTypeKind.CustomMarshaler ? customMarshallerType : throw new InvalidOperationException("Custom marshaller type can be set only when using for CustomMarshaller");
```
MichalStrehovsky: We don't do throws like this because then it just raises questions whether the message needs to be localized, etc. This should be an assert, but I'm not firm on the validation (we don't validate the array stuff above either).

src/coreclr/tools/Common/TypeSystem/Interop/MarshalAsDescriptor.cs
```diff
@@ -48,17 +48,24 @@ public enum NativeTypeKind : byte
...
+        public string Cookie => Type == NativeTypeKind.CustomMarshaler ? cookie : throw new InvalidOperationException("Cookie can be set only when using for CustomMarshaller");
 
-        public MarshalAsDescriptor(NativeTypeKind type, NativeTypeKind arraySubType, uint? sizeParamIndex, uint? sizeConst)
+        public MarshalAsDescriptor(NativeTypeKind type, NativeTypeKind arraySubType, uint? sizeParamIndex, uint? sizeConst, TypeDesc customMarshallerType, string cookie)
         {
             Type = type;
             ArraySubType = arraySubType;
             SizeParamIndex = sizeParamIndex;
             SizeConst = sizeConst;
+            this.customMarshallerType = customMarshallerType;
```
MichalStrehovsky: No need to prefix with `this` if the field has underscore because that's the convention.

MichalStrehovsky:
If I'm reading the CoreCLR sources correctly, this is cached per `TypeOfTheParameterBeingMarshalled+CustomMarshallerType+cookie` - https://github.com/dotnet/runtime/blob/948770a23b1e1b5d128d7bbc06f172fbd01d4ba5/src/coreclr/vm/mlinfo.cpp#L908-L910.

An approach that would probably work is to add a helper method to InteropHelpers, e.g. `object GetCustomMarshallerInstance(MethodTable* pParameterType, MethodTable* pMarshallerType, string cookie, delegate*<object, string> getInstanceMethod)`. The marshalling stub would call this and the helper would consult the cache and decide whether getInstanceMethod function pointer should be called and cached or not (and the request can be satisfied from the cache).

I think we can use ConcurrentUnifier from CoreLib to do the value unification. GetInstance will be called twice if there's a race but that will happen in CoreCLR too.

kant2002:
I think key created using `MethodSignatureType+CustomMarshallerType+cookie`
I take this guess using this line
https://github.com/dotnet/runtime/blob/948770a23b1e1b5d128d7bbc06f172fbd01d4ba5/src/coreclr/vm/mlinfo.cpp#L1297

kant2002:
I think I make something more or less stable. At least half of checks from `Interop/ICustomMarshaler/Primitives/ICustomMarshaler_TargetWindows` works. Not works only different exception types.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/PInvokeMarshal.Unix.cs
```diff
@@ -60,5 +60,10 @@ public static unsafe int GetCharCount(byte* multiByteStr, int multiByteLen)
             return System.Text.Encoding.UTF8.GetCharCount(multiByteStr, multiByteLen);
         }
         #endregion
+
+        public static unsafe object InitializeCustomMarshaller(RuntimeTypeHandle pParameterType, RuntimeTypeHandle pMarshallerType, string cookie, delegate*<string, object> getInstanceMethod)
```
jkotas: Custom marshallers are not Windows-specific. This should be in platform neutral file.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/PInvokeMarshal.Windows.cs
```diff
@@ -90,5 +91,68 @@ public static unsafe int GetCharCount(byte* multiByteStr, int multiByteLen)
...
+            var marshaller = s_customMarshallersTable.GetOrAdd(new CustomMarshallerKey(pParameterType, pMarshallerType, cookie, getInstanceMethod));
+            if (marshaller == null)
+            {
+                throw new ApplicationException();
+            }
+
+            return marshaller;
+        }
+
+        private static CustomMarshallerTable s_customMarshallersTable = new CustomMarshallerTable();
```
jkotas: This static will pull in the `CustomMarshallerTable` into the typical all (static fields in the given type cannot be trimmed independently). The static should be on `CustomMarshallerTable` to keep it separate.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/PInvokeMarshal.cs
```diff
@@ -733,5 +734,68 @@ private static unsafe bool CalculateStringLength(byte* pchBuffer, out int ansiBu
...
+        {
+            if (getInstanceMethod == null)
+            {
+                throw new ApplicationException();
+            }
+
+            var marshaller = CustomMarshallerTable.s_customMarshallersTable.GetOrAdd(new CustomMarshallerKey(pParameterType, pMarshallerType, cookie, getInstanceMethod));
+            if (marshaller == null)
+            {
+                throw new ApplicationException();
```
jkotas: What would cause this exception to be thrown?

kant2002: This case is checked here https://github.com/dotnet/runtimelab/blob/861707ef38233883151921bfac679353be4836f3/src/tests/Interop/ICustomMarshaler/Primitives/ICustomMarshaler.cs#L470

Check above in this method is attempt to cover these tests
https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/tests/Interop/ICustomMarshaler/Primitives/ICustomMarshaler.cs#L637-L641

But I cannot pass `null` as pointer value using `Emitter`.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,172 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+            ILEmitter emitter = _ilCodeStreams.Emitter;
+            var lMarshaller = emitter.NewLocal(customMarshallerType);
+            var cookie = MarshalAsDescriptor.Cookie;
+
+            // Custom marshaller initialization should not be catched, so initialize early
+            ILCodeStream fnptrLoadStream = _ilCodeStreams.FunctionPointerLoadStream;
+            fnptrLoadStream.Emit(ILOpcode.ldtoken, emitter.NewToken(ManagedParameterType));
+            fnptrLoadStream.Emit(ILOpcode.ldtoken, emitter.NewToken(marshallerType));
+            fnptrLoadStream.Emit(ILOpcode.ldstr, emitter.NewToken(cookie));
+            fnptrLoadStream.Emit(ILOpcode.ldftn, emitter.NewToken(getInstanceMethod));
```
MichalStrehovsky: If you drop the `if (getInstanceMethod == null)` check a couple lines above, you can do this:

```suggestion
            if (getInstanceMethod != null)
            {
                fnptrLoadStream.Emit(ILOpcode.ldftn, emitter.NewToken(getInstanceMethod));
            }
            else
            {
                fnptrLoadStream.EmitLdc(0);
                fnptrLoadStream.Emit(ILOpcode.conv_i);
            }
```

Then this will go into the `throw new ApplicationException();` codepath you have in the classlib.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,172 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+                    new TypeDesc[] { Context.GetWellKnownType(WellKnownType.String) }));
+            var initializeCustomMarshallerMethod = InteropTypes.GetPInvokeMarshal(Context).GetKnownMethod("InitializeCustomMarshaller",
+                new MethodSignature(MethodSignatureFlags.Static, 0, Context.GetWellKnownType(WellKnownType.Object),
+                    new TypeDesc[] {
+                        Context.GetWellKnownType(WellKnownType.RuntimeTypeHandle),
+                        Context.GetWellKnownType(WellKnownType.RuntimeTypeHandle),
+                        Context.GetWellKnownType(WellKnownType.String),
+                        getInstanceMethodType,
+                    }
+                ));
```
MichalStrehovsky: This can just be:

```suggestion
            var initializeCustomMarshallerMethod = InteropTypes.GetPInvokeMarshal(Context).GetKnownMethod("InitializeCustomMarshaller", null);
```

We don't generally worry about overloads in our helpers. Helpers are under our control and this is less code.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/PInvokeMarshal.cs
```diff
@@ -733,5 +734,68 @@ private static unsafe bool CalculateStringLength(byte* pchBuffer, out int ansiBu
         }
 
         #endregion
+
+        public static unsafe object InitializeCustomMarshaller(RuntimeTypeHandle pParameterType, RuntimeTypeHandle pMarshallerType, string cookie, delegate*<string, object> getInstanceMethod)
```
MichalStrehovsky: Could you move this helper to src\coreclr\nativeaot\System.Private.CoreLib\src\Internal\Runtime\CompilerHelpers\InteropHelpers.cs? It's where we keep all the other helpers for marshalling stubs.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,172 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+            // Custom marshaller initialization should not be catched, so initialize early
+            ILCodeStream fnptrLoadStream = _ilCodeStreams.FunctionPointerLoadStream;
+            fnptrLoadStream.Emit(ILOpcode.ldtoken, emitter.NewToken(ManagedParameterType));
+            fnptrLoadStream.Emit(ILOpcode.ldtoken, emitter.NewToken(marshallerType));
+            fnptrLoadStream.Emit(ILOpcode.ldstr, emitter.NewToken(cookie));
+            fnptrLoadStream.Emit(ILOpcode.ldftn, emitter.NewToken(getInstanceMethod));
+            fnptrLoadStream.Emit(ILOpcode.call, emitter.NewToken(initializeCustomMarshallerMethod));
+            fnptrLoadStream.EmitStLoc(lMarshaller);
+
+            var marshalNativeToManagedMethod = marshallerType.GetMethod(
```
MichalStrehovsky:
```suggestion
            var marshalNativeToManagedMethod = customMarshallerType.GetMethod(
```

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,172 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+            fnptrLoadStream.Emit(ILOpcode.ldtoken, emitter.NewToken(marshallerType));
+            fnptrLoadStream.Emit(ILOpcode.ldstr, emitter.NewToken(cookie));
+            fnptrLoadStream.Emit(ILOpcode.ldftn, emitter.NewToken(getInstanceMethod));
+            fnptrLoadStream.Emit(ILOpcode.call, emitter.NewToken(initializeCustomMarshallerMethod));
+            fnptrLoadStream.EmitStLoc(lMarshaller);
+
+            var manageToNativeMethod = customMarshallerType.GetMethod(
+                "MarshalManagedToNative",
+                new MethodSignature(MethodSignatureFlags.None, 0, Context.GetWellKnownType(WellKnownType.IntPtr), new[] { Context.GetWellKnownType(WellKnownType.Object) }));
+            if (manageToNativeMethod == null)
```
MichalStrehovsky: This would be a classlib bug. It can be `GetKnownMethod`.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,172 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+            codeStream.EmitLdLoc(lMarshaller);
+            LoadManagedValue(codeStream);
+            codeStream.Emit(ILOpcode.callvirt, emitter.NewToken(manageToNativeMethod));
+            StoreNativeValue(codeStream);
+
+            // Call CleanUpNativeData on cleanup code stream.
+            var cleanupNativeDataMethod = customMarshallerType.GetMethod(
+                "CleanUpNativeData",
+                new MethodSignature(MethodSignatureFlags.None, 0, Context.GetWellKnownType(WellKnownType.Void), new[] { Context.GetWellKnownType(WellKnownType.IntPtr) }));
+            if (cleanupNativeDataMethod == null)
```
MichalStrehovsky: Same - we can use `GetKnownMethod` and no need to handle null.

kant2002 (commented):
Apply PR feedback

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,146 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+                ThrowHelper.ThrowMarshalDirectiveException();
+            }
+
+            if (!marshallerType.CanCastTo(customMarshallerType))
+            {
+                // ApplicationException throw here.
+                ThrowHelper.ThrowMarshalDirectiveException();
+            }
+
+            var getInstanceMethod = marshallerType?.GetMethod(
```
jkotas: Can `marshallerType` be null here? We are using the variable number of times above.

kant2002: I working on this right now. My idea is to pass Type to InitializeCustomMarshaller and let it decide what exceptions to throw. That allow me to cover all cases except one.

kant2002:
I do not know how to make this test pass
https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/tests/Interop/ICustomMarshaler/Primitives/ICustomMarshaler.cs#L646

Also I need to add `System.Runtime.InteropServices.CustomMarshalers.XXXX` types to NativeAOT version of `System.Private.CoreLib`. It's not clear for me proper way how to share files.
Aha, after some investigation I think I should move files from `src\coreclr\System.Private.CoreLib\src\System\Runtime\InteropServices\CustomMarshalers` to `src\libraries\Common\src\System\Runtime\InteropServices\CustomMarshalers` and include them in the NativeAOT version of `System.Private.CoreLib` from that location.

jkotas:
> I need to add System.Runtime.InteropServices.CustomMarshalers.XXXX types to NativeAOT version of System.Private.CoreLib

Why do you need these? These types are part of built-in COM interop support that we do not have in NativeAOT.

kant2002:
> Why do you need these? These types are part of built-in COM interop support that we do not have in NativeAOT.

Purity, I think. That's from `Interop/PInvoke/CustomMarshalers/CustomMarshalersTest` these tests. Instead of fixing tests, I can change rationale for this test exclusion to `<Issue>Specific to CoreCLR</Issue>` for example.

jkotas:
Yes, I think that would be better.

kant2002:
Is it possible create something like that
```csharp
#if READYTORUN
Field_ParentIsStruct_ThrowsTypeLoadException();
#endif

So I remove `Interop/ICustomMarshaler/Primitives` tests from exclusion? Other checks in these tests are passing now.
```

jkotas:
> Is it possible create something like that

We do not have a good fine-grained way to disable tests under src/tests today. It is being worked on in dotnet/runtime - we will use the same scheme as libraries tests.

> Field_ParentIsStruct_ThrowsTypeLoadException

Have you tried to check `isField` based on https://github.com/dotnet/runtimelab/pull/1723#discussion_r751860805  to make this test work?

kant2002:
> Have you tried to check `isField` based on [#1723 (comment)](https://github.com/dotnet/runtimelab/pull/1723#discussion_r751860805) to make this test work?

No, because that's mean that I completely disable marshaling on fields, but that's should be disabled only for fields within structs. I probably have to raise my concerns there in comment. But I was busy make that work, and forget about it.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -598,6 +599,43 @@ internal static unsafe void CleanupVariant(IntPtr pDstNativeVariant)
...
+            {
+                throw new TypeLoadException();
+            }
+
+            if (pMarshallerType.IsGenericTypeDefinition)
+            {
+                throw new TypeLoadException();
+            }
+
+            if (!pMarshallerType.IsAssignableTo(typeof(ICustomMarshaler)))
```
MichalStrehovsky: This is making interop a lot slower just to throw a compatible exception. It might not be the right compromise.

(Also grabbing a System.Type is not free either.)

We have been compromising on throwing the right exception type if it's more than a couple lines of code, or comes with size/perf issues.

MichalStrehovsky:
> No, because that's mean that I completely disable marshaling on fields, but that's should be disabled only for fields within structs. I probably have to raise my concerns there in comment. But I was busy make that work, and forget about it.

When would an `!isField` check do the wrong thing for this?

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
kant2002: Okay, I make that change to be pedantic about passing tests without changing `ExceptionStringID` enum, so I will revert `Type` usage, and will use `RuntimeTypeHandle` again with wrong exception temporary.

Regarding field. Let's take `String` as type which would be marshalled.
There 3 cases
1. String as parameter. This works now and CoreCLR
https://github.com/dotnet/runtime/blob/c2b7638a5b6f3a6a52540244f3f85fba02284459/src/tests/Interop/ICustomMarshaler/Primitives/ICustomMarshaler.cs#L25

2. String as field inside class. This works now and in CoreCLR
https://github.com/dotnet/runtime/blob/c2b7638a5b6f3a6a52540244f3f85fba02284459/src/tests/Interop/ICustomMarshaler/Primitives/ICustomMarshaler.cs#L96

3. String as field inside structure. This presumable works now, but should not in CoreCLR
https://github.com/dotnet/runtime/blob/c2b7638a5b6f3a6a52540244f3f85fba02284459/src/tests/Interop/ICustomMarshaler/Primitives/ICustomMarshaler.cs#L554

Cases 2 and 3 identical except that field inside class or struct. I need something more then `isField`. Something like `isStruct`. I maybe miss something obvious?

MichalStrehovsky: Maybe I'm missing something. Case 3 attempts to marshal following structure:

```csharp
        public struct StructWithCustomMarshalerField
        {
            [MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof(StringForwardingCustomMarshaler))]
            public string Field;
        }

// With this P.invoke:

public static extern int StructWithCustomMarshalerFieldMethod(StructWithCustomMarshalerField c);
```

This fails because the field is decorated as CustomMarshaler. It's what `isField` covers.

Case 2 marshals this class:

```csharp

        public class StringContainer
        {
            public string Value { get; set; }
        }

// With this pinvoke

        public static extern StringContainer MarshalerOnClassTypeMethod([MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof(ClassForwardingCustomMarshaler))] StringContainer str);
```

This works because CustomMarshaller is on a parameter, not on a field. The `isField` check would not be hit for this case because the MarshalAs is not on the field.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -598,6 +599,33 @@ internal static unsafe void CleanupVariant(IntPtr pDstNativeVariant)
...
+
+            if (pParameterType.ToEETypePtr().IsPointer || pParameterType.IsValueType())
+            {
+                throw new MarshalDirectiveException();
+            }
+
+            if (getInstanceMethod == null)
+            {
+                throw new ApplicationException();
+            }
```
jkotas: We have some error handling in the compiler and some here. I would move all of it to the compiler. It is ok that the errors won't match CoreCLR exactly.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,135 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+            if (marshallerType == null)
+            {
+                ThrowHelper.ThrowMarshalDirectiveException();
+            }
+
+            var customMarshallerType = Context.SystemModule.GetKnownType("System.Runtime.InteropServices", "ICustomMarshaler");
+            if (customMarshallerType == null)
+            {
+                ThrowHelper.ThrowMarshalDirectiveException();
+            }
```
jkotas:
```suggestion
```
GetKnownType does not return null.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -598,6 +599,33 @@ internal static unsafe void CleanupVariant(IntPtr pDstNativeVariant)
...
         }
 
+        public static unsafe ICustomMarshaler InitializeCustomMarshaller(RuntimeTypeHandle pParameterType, RuntimeTypeHandle pMarshallerType, string cookie, delegate*<string, object> getInstanceMethod)
+        {
+            if (pMarshallerType.IsGenericTypeDefinition())
+            {
+                throw new TypeLoadException();
+            }
+
+            if (pParameterType.ToEETypePtr().IsPointer || pParameterType.IsValueType())
```
jkotas: This condition is checked in the compiler.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
kant2002: Aah, Yes. I definitely overlook tha. Thanks.

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
kant2002: > is also allowed on singledim+multidim arrays and classes

I think I handle that now.

> isField here

Finally make it thank's to your suggestion below.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -598,6 +599,18 @@ internal static unsafe void CleanupVariant(IntPtr pDstNativeVariant)
...
         }
 
+        public static unsafe ICustomMarshaler InitializeCustomMarshaller(RuntimeTypeHandle pParameterType, RuntimeTypeHandle pMarshallerType, string cookie, delegate*<string, object> getInstanceMethod)
+        {
+            var marshaller = CustomMarshallerTable.s_customMarshallersTable.GetOrAdd(
+                new CustomMarshallerKey(pParameterType, pMarshallerType, cookie, getInstanceMethod));
+            if (marshaller == null)
+            {
+                throw new ApplicationException();
+            }
```
jkotas, line 609: This cannot happen anymore now that this condition is checked in the compiler. Is that right?

kant2002: No, because this is can be only runtime based check. This is checking created marshaller instance. I may express that in IL, but what's the difference?

jkotas: Ah ok

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/InteropHelpers.cs
```diff
@@ -615,5 +628,52 @@ internal unsafe struct MethodFixupCell
...
+            {
+                return ParameterType.GetHashCode()
+                    ^ MarshallerType.GetHashCode()
+                    ^ Cookie.GetHashCode();
+            }
+        }
+
+        internal sealed class CustomMarshallerTable : ConcurrentUnifier<CustomMarshallerKey, object>
+        {
+            internal static CustomMarshallerTable s_customMarshallersTable = new CustomMarshallerTable();
```
jkotas:
```suggestion
            internal static readonly CustomMarshallerTable s_customMarshallersTable = new CustomMarshallerTable();
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/PInvokeMarshal.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Security;
 using Debug = System.Diagnostics.Debug;
+using System.Collections.Concurrent;
```
jkotas:
```suggestion
```

src/coreclr/tools/Common/TypeSystem/Interop/IL/MarshalHelpers.cs
```diff
@@ -244,11 +247,17 @@ public static partial class MarshalHelpers
...
 
+            if (isField && nativeType == NativeTypeKind.CustomMarshaler)
+                return MarshallerKind.Invalid;
+
             //
             // Determine MarshalerKind
             //
             if (type.IsPrimitive)
             {
+                if (nativeType == NativeTypeKind.CustomMarshaler)
```
jkotas: How does this case work in CoreCLR? I do not think it works in this implementation.

The `ICustomMarshaler` takes object. The primitive type would have to be boxed to get converted to object, but I do not see any boxing in the generated IL.

MichalStrehovsky: Primitive types are going to go into the throwing path with rest of IsValueType types.

Instead of putting all these `if (nativeType == NativeTypeKind.CustomMarshaler)` checks everywhere, can we do this like CoreCLR before we start looking in detail at `type`:

https://github.com/dotnet/runtimelab/blob/0aeac63238f2b7cd6b788b66b5fd9655cc183a60/src/coreclr/vm/mlinfo.cpp#L1264-L1327

I.e. do a single check for nativeType == NativeTypeKind.CustomMarshaler 3 lines above this one.

We're still missing handling of CustomMarshaller on e.g. string and delegate below, which I assume is possible with CoreCLR. A single check should accomplish what we need.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,132 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+
+            if (!marshallerType.CanCastTo(customMarshallerType))
+            {
+                ThrowHelper.ThrowMarshalDirectiveException();
+            }
+
+            if (ManagedParameterType.IsPointer || ManagedParameterType.IsValueType)
+            {
+                ThrowHelper.ThrowMarshalDirectiveException();
+            }
```
MichalStrehovsky: Would be nice to deduplicate as much of this as possible with TransformNativeToManaged.

src/coreclr/tools/Common/TypeSystem/Interop/IL/Marshaller.Aot.cs
```diff
@@ -1125,4 +1127,95 @@ protected override void EmitCleanupManaged(ILCodeStream codeStream)
...
+
+            var marshalNativeToManagedMethod = customMarshallerType.GetKnownMethod(
+                "MarshalNativeToManaged",
+                new MethodSignature(MethodSignatureFlags.None, 0, Context.GetWellKnownType(WellKnownType.Object), new[] { Context.GetWellKnownType(WellKnownType.IntPtr) }));
+
+            ILEmitter emitter = _ilCodeStreams.Emitter;
+            codeStream.EmitLdLoc(lMarshaller);
+            LoadNativeValue(codeStream);
+            codeStream.Emit(ILOpcode.callvirt, emitter.NewToken(marshalNativeToManagedMethod));
+            StoreManagedValue(codeStream);
```
jkotas, line 1171: Do you need to call `ICustomMarshaler.CleanUpManagedData` here?

kant2002: Honestly, if not Michal help, I would never found these places. Even when I'm looking from them, I cannot find who emit calls to `CleanUpManagedData` so I cannot follow existing logic and rely only on my gut feling and tests.

If you think, that I should place that in `CleanUpManagedData` in `Marshaller.EmitCleanupManaged`, then I would 
- have to exclude call that for `TransformNativeToManaged` or it will fails.
- I would not have easy access to `lMarshaller` variable. This is not so problematic, just convinient.

Or did you mean that I select incorrect code stream?

jkotas: The managed to native, and native to managed directions are symmetric. 

Managed to native should call `ICustomMarshaler.MarshalManagedToNative` and `ICustomMarshaler.CleanUpNativeData`. The code in this PR seems to be doing it correctly.

Native to managed should call `ICustomMarshaler.MarshalNativeToManaged` and  `ICustomMarshaler.CleanUpManagedData`.  The code in this PR seems to be missing the `CleanUpManagedData` part.

kant2002: Yeah. I agree. Still I did not find original code, maybe I should look more. And that case did not covered by tests.

https://github.com/dotnet/runtimelab/blob/1e626fccf9cd223502a3c5fc211e4f7a91c03591/src/tests/Interop/ICustomMarshaler/Primitives/ICustomMarshaler.cs#L141

I suspect that `ICustomMarshaler.CleanUpManagedData` probably related to delegates marshalling, which currently did not tested. I can go to runtime repo and add additional tests which highlight that behaviour. Then my life would be slightly easier, since I may rely on `runtime` tests, and your experience.

jkotas: The CoreCLR code that calls the cleanup method is: https://github.com/dotnet/runtime/search?q=InvokeCleanUpManagedMeth

I am not able to find any tests for custom marshaler and delegates. It would be nice to add a few.

kant2002: https://github.com/dotnet/runtime/pull/62691 this is what I plan to add. I hope this would be enough.

PR #1724, opened 2021-11-14
Update version of ILC

kant2002:
This bump to prevent errors during local compilation which happens after https://github.com/dotnet/runtime/pull/61333

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
+++ b/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
@@ -31,7 +31,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


MichalStrehovsky:
> This bump to prevent errors during local compilation which happens after dotnet/runtime#61333

Bumping this is fine, but I would like to understand what errors you're seeing - this should only be used so that we can extract objwriter.dll out of the package; the rest is unused. We just don't have a standalone package for the object writer anymore.

kant2002:
I was having `{"Missing method 'System.Security.Cryptography.Oid System.Security.Cryptography.X509Certificates.PublicKey.get_Oid()'"}` during debugging in `ILScanner.CompileSingleMethod` but it was catched and handled. Work fine without bumping. I was misinterpret this as related.

PR #1727, opened 2021-11-15
Reduce diffs with runtime repo in src/tests

MichalStrehovsky:
* Use `TestBuildMode` passed as an env variable, same as everyone else.
* Skip CoreCLR-JIT specific tests by using the same mechanism as Mono
* Compile Microsoft.CodeAnalysis junk. We no longer choke on recursion. This part I'm not too happy about. We should ideally have a scheme where we can test high-value external things outside the CoreCLR test infra.


jkotas:
Tests are failing...

MichalStrehovsky:
> Tests are failing...

This was an assert when compiling Microsoft.CodeAnalysis.VisualBasic.

jkotas (approved):
Thanks

PR #1736, opened 2021-11-19
Make MetadataFieldLayoutAlgorithm shareable

MichalStrehovsky:
When and how `StructLayout.Sequential` applies is a difficult question and shouldn't be in the core algorithm.

Making `MetadataFieldLayoutAlgorithm` abstract caused trouble in the type loader; for no good reason since we shouldn't be loading things from metadata in NativeAOT right now. Put the non-standard cases (that didn't make it out of the CoreRT repo for now) under ifdefs. If we ever need this back, the solution is a middle man between `MetadataFieldLayoutAlgorithm` and `CompilerFieldLayoutAlgorithm` that can be shared.


jkotas (approved):
Thank you!

PR #1737, opened 2021-11-19
Merge from dotnet/runtime

jkotas:


jkotas:
@MichalStrehovsky Is there any known adjustments that has to made after the integration of the shared files to fix the failing tests?

It looks like we are not marking methods as reflectable in some cases, e.g. "Unhandled Exception: System.Exception: Method 'Frob' not found on type ReflectionTest+TestInterfaceMethod+IFoo".

MichalStrehovsky:
> It looks like we are not marking methods as reflectable in some cases, e.g. "Unhandled Exception: System.Exception: Method 'Frob' not found on type ReflectionTest+TestInterfaceMethod+IFoo".

RyuJIT apparently devirtualized the interface method call in Debug builds and we no longer see the interface method as used.

This is also why it work in Release builds (where the scanner makes sure the interface method is seen as used).

I'll look into this later. Possibly tomorrow or day after. We could also just disable the Reflection test for now to unblock integration. I don't think this is super critical. We have coverage through the Reflection_ReflectedOnly test.

But it's surprising to see optimizations kicking in with optimizations disabled.

```
00007FF76D25418B 48 8D 0D A6 D3 42 00 lea         rcx,[repro_ReflectionTest_TestInterfaceMethod_Foo::`vftable' (07FF76D681538h)]  
00007FF76D254192 E8 19 4B E0 FF       call        RhpNewFast (07FF76D058CB0h)  
00007FF76D254197 48 89 85 58 FF FF FF mov         qword ptr [rbp-0A8h],rax  
00007FF76D25419E 48 8B 8D 58 FF FF FF mov         rcx,qword ptr [rbp-0A8h]  
00007FF76D2541A5 E8 B6 37 00 00       call        repro_ReflectionTest_TestInterfaceMethod_Foo___ctor (07FF76D257960h)  
00007FF76D2541AA 48 8B 8D 58 FF FF FF mov         rcx,qword ptr [rbp-0A8h]  
00007FF76D2541B1 BA 01 00 00 00       mov         edx,1  
00007FF76D2541B6 39 09                cmp         dword ptr [rcx],ecx  
00007FF76D2541B8 E8 63 37 00 00       call        repro_ReflectionTest_TestInterfaceMethod_Foo__Frob (07FF76D257920h)  
00007FF76D2541BD 48 89 85 50 FF FF FF mov         qword ptr [rbp-0B0h],rax  
```

jkotas:
> I'll look into this later. Possibly tomorrow or day after.

Thanks! 

I think we should wait with the merge until the problem is understood.

MichalStrehovsky:
> I think we should wait with the merge until the problem is understood.

https://github.com/dotnet/runtime/pull/61868 should fix this.

jkotas:
There are still a few reflection tests failing in release builds. @MichalStrehovsky Could you please take a look when you get a chance?

MichalStrehovsky:
> There are still a few reflection tests failing in release builds. @MichalStrehovsky Could you please take a look when you get a chance?

Based on what the test says, I'm pretty sure this is https://github.com/dotnet/runtime/pull/54056.

I wonder if we should just delete the attribute stripping from the AOT compiler. Attribute stripping currently removes attributes from the assembly, no questions asked, breaking user code in the process. But I believe pretty strongly that attribute stripping should only happen to things not visible from reflection (https://github.com/dotnet/runtime/issues/103934). If we go that route, it's basically what the compiler does in `--reflectedonly` mode and it's not limited to just attributes.

jkotas:
I am disabling the tests for now. I agree that the current behavior does not seem to be ideal. At the same time, I would like to avoid introducing differences between trimming done by the IL linker and trimming done by the NativeAOT compiler.

PR #1739, opened 2021-11-20
Enable -pie and -z,relro,-z,now for NativeAOT binaries

jkotas:
Fixes binskim warning BA3001 and BA3011

Contributes to dotnet/runtime#96848

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.props
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.props
@@ -89,6 +89,12 @@ The .NET Foundation licenses this file to you under the MIT license.
       <LinkerArg Include="-licucore" Condition="'$(TargetOS)' == 'OSX'" />
       <LinkerArg Include="-dynamiclib" Condition="'$(TargetOS)' == 'OSX' and '$(NativeLib)' == 'Shared'" />
       <LinkerArg Include="-shared" Condition="'$(TargetOS)' != 'OSX' and '$(NativeLib)' == 'Shared'" />
+      <!-- binskim warning BA3001 PIE disabled on executable -->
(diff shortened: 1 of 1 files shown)


MichalStrehovsky (approved):
Thank you!

StephanDollberg:
Hi, any way to disable these options? 

Other things might not work with pie binaries so wondering whether I can disable those linker options?

jkotas:
I do think there is an easy way to suppress this option today. We would be happy to accept a PR to add `PositionIndependentExecutable` msbuild property that you can set to false to suppress this option. (This PR should be submitted to dotnet/runtime repo.)

StephanDollberg:
> I do think there

I assume you are missing a "not" there?

jkotas:
Yes, of course...

StephanDollberg:
https://github.com/dotnet/runtime/compare/main...StephanDollberg:stephandollberg/PositionIndependentExecutable-flag?expand=1 

Does this match what you were thinking of?

Unfortunately I am failing to make my project work with the local build of nativeaot so can't really test it.

jkotas:
I think this can be just one-line change for the `-pie` command line option, like:

`<LinkerArg Include="-pie" Condition="'$(TargetOS)' != 'OSX' and '$(NativeLib)' == '' and '$(PositionIndependentExecutable)' != 'false'" />`

Do you really need to control the other command line options too? If yes, the msbuild property should have different name.

StephanDollberg:
I guess not but I feel like they all belong together hence why I included them all.

Happy to only do pie if you prefer that.

jkotas:
All of these options are related to security hardening of the final binary. We strongly prefer .NET (and NativeAOT in particular) to have security hardening enabled if possible. So it is preferable to just omit the one option that is causing problems.

StephanDollberg:
Yeah sure, have created https://github.com/dotnet/runtime/pull/64580

Issue #1740, opened 2021-11-21
Question: Difference between CORINFO_FLG_CONTAINS_GC_PTR and varDsc->HasGCPtr()


yowl:
A while back there was some discussion on a means to decide if a struct should be GC tracked for the LLVM/Wasm backend.  
https://github.com/dotnet/runtimelab/issues/1647#issuecomment-939528154
and
https://github.com/dotnet/runtimelab/issues/1647#issuecomment-939538732

There's  a different between these 2 for `Span<char>` as one method includes `type.IsByReferenceOfT` and one doesn't.  Why is that and should `type.IsByReferenceOfT` be considered when passing `Span<char>` as an argument?

Currently the IL->LLVM backed passes `Span<char>` on the LLVM stack, so is that correct?

jkotas:
`Span<char>` has to be GC tracked, so it means `varDsc->HasGCPtr()` is what you need.

There was a discussion about `CORINFO_FLG_CONTAINS_GC_PTR` back in the day when `Span<T>` was introduced, and we agreed that it should not be set for `Span`: https://github.com/dotnet/runtime/issues/6909#issuecomment-277065925

PR #1742, opened 2021-11-21
[NativeAOT-LLVM]: Move structs with ByRefs to the shadow stack

yowl:
This PR fixes a problem with the IL backend where it was passing `Span<char>` (and probably all `Span<T>`) on the LLVM stack, preventing them from being considered for GC.

Follows https://github.com/dotnet/runtimelab/issues/1740

It copies some of the logic from `GatherClassGCLayout` in  `CorInfoImpl.cs` as I didn't see it exposed.  Probably ok as eventually this will be deleted when all the LLVM is generated from clrjit.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -1481,36 +1481,6 @@ namespace Internal.IL
             return false;
         }
 
-        // TODO-LLVM: this is a copy of some of the logic in GatherClassGCLayout so we get the same logic as clrjit for
-        // determining if a struct should be passed on the shadow stack.  Span<char/T> is an example.
-        private static bool ContainsIsByReferenceOfT(TypeDesc type)
-        {
-            if (type.IsByReferenceOfT)
-            {
-                return true;
-            }
-
-            foreach (var field in type.GetFields())
-            {
-                if (field.IsStatic)
-                    continue;
-
-                var fieldType = field.FieldType;
-                if (fieldType.IsValueType)
-                {
-                    var fieldDefType = (DefType)fieldType;
-                    if (!fieldDefType.ContainsGCPointers && !fieldDefType.IsByRefLike)
-                        continue;
-
-                    if (ContainsIsByReferenceOfT(fieldType))
-                    {
-                        return true;
-                    }
-                }
-            }
-            return false;
-        }
-
         /// <summary>
         /// Returns true if the type can be stored on the local stack
         /// instead of the shadow stack in this method.
@@ -1519,7 +1489,7 @@ namespace Internal.IL
         {
             if (type is DefType defType)
             {
-                if (!defType.IsGCPointer && !defType.ContainsGCPointers && !ContainsIsByReferenceOfT(type))
+                if (!defType.IsGCPointer && !defType.ContainsGCPointers && !type.IsByRefLike)
                 {
                     return true;
                 }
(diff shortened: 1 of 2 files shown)


jkotas:
cc @SingleAccretion

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
```diff
@@ -1489,7 +1519,7 @@ internal static bool CanStoreTypeOnStack(TypeDesc type)
         {
             if (type is DefType defType)
             {
-                if (!defType.IsGCPointer && !defType.ContainsGCPointers)
+                if (!defType.IsGCPointer && !defType.ContainsGCPointers && !ContainsIsByReferenceOfT(type))
```
jkotas: `CORINFO_FLG_CONTAINS_STACK_PTR` is equivalent to `type.IsByRefLike`. Should this just call `type.IsByRefLike`?

yowl: It is correct with `type.IsByRefLike`  so I guess so, I just couldn't see where it did the recursion, i.e if a field is `IsByRefLike`, then so is it's struct.

SingleAccretion: > `CORINFO_FLG_CONTAINS_STACK_PTR` is equivalent to `type.IsByRefLike`

For my education: this means that reflection's `IsByRefLike` is not equivalent to type system's `IsByRefLike`, correct?

jkotas: Reflection IsByRefLike is same as type system IsByRefLike. And the type system IsByRefLike is same as `CORINFO_FLG_CONTAINS_STACK_PTR`.

Using `IsByRefLike` / `CORINFO_FLG_CONTAINS_STACK_PTR` here is a simplification that is leaving some performance on the table for byref-like types that do not need GC tracking.

jkotas (approved):
LGTM. Thanks

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
yowl: Edited...

> CORINFO_FLG_CONTAINS_STACK_PTR is equivalent to type.IsByRefLike. Should this just call type.IsByRefLike?

I'm finding I've still got a problem with

```cs
    public ref struct GenStructWithImplicitOp<TKey>
    {
        private int length;
        private int length2; // just one int field will not create an LLVM struct type, so put another field

        public GenStructWithImplicitOp(TKey[] key)
        {
            length = key.Length;
            length2 = length;
        }

        public static implicit operator GenStructWithImplicitOp<TKey>(TKey[] array) => new GenStructWithImplicitOp<TKey>(array);
    }
```
With `GenStructWithImplicitOp<string>`  .   This struct is `IsByRefLike` but `IsByReferenceOfT` is not set for any field.  Is `IsByRefLike` being set because its a  `ref struct`?

jkotas: Yes, both `IsByRefLike` and `CORINFO_FLG_CONTAINS_STACK_PTR` means `ref struct`. It means that these can be set even for structs that do not require any GC tracking, like the struct from your example.

If you need precise answer for whether the struct requires GC tracking, algorithm like what you have in `ContainsIsByReferenceOfT` or the `HasGCPtr` property in the JIT is way to go.

PR #1744, opened 2021-11-22
[NativeAOT-LLVM] lower args using PUTARG_TYPE

yowl:
This PR lowers call arguments to the shadow stack, simplifying the process of creating the LLVM call.

The original idea is below,

There's a bit of refactoring here making the diff a bit hard to read, sorry about that.

Points of interest:

- All struct types are forced to `SPK_ByValue` as the lowering will decide where to place them
- `GenTreeCall` has a new field to hold the original call type.  I did this so that I could create the LLVM signature accurately for string constructors which have a return value, but that type was not available at the point of creating the signature.
- When calling back to Ilc to get the arg type, `CORINFO_TYPE_NATIVEINT` are changed back to `CORINFO_TYPE_PTR` to make them generate as i8*, this will be a problem when we hit a real `nativeint`, I suppose.  I don't know why `TypeFlags` and `CorInfoTypes` aren't exactly inline.
- Coverage up to around 34%

Lower arguments to `CALL`s to match the intended LLVM call convention, adding shadowstack arg and moving GC tracked args out of the signature and onto the shadowstack.

Copied from #1716 

Basically the problem is that for this C#:

static void Method(SmallStruct s1, StructWithLong s2, object obj) { }

SmallStruct { short SmallInteger; }
StructWithLong { long Long; }

You can end up with something like this in IR:

```
CALL(int, long, ref);
```
And for LLVM we want this:
```
CALL(ShadowStack, SmallStruct, StructWithLong);
```
So the solution was to wrap the argument nodes into PUTARG_TYPE nodes that would carry their true types:
```
CALL(PUTARG_TYPE(SmallStruct (int)), PUTARG_TYPE(StructWithLong (long)), PUTARG_TYPE(object (object)));
```
    Rewrite all arguments that would be needed to be passed on the shadow stack:
```
STOREIND(VSS + OFFSET, object)

CALL(PUTARG_TYPE(nint (VSS + OFFSET)), PUTARG_TYPE(SmallStruct (int)), PUTARG_TYPE(StructWithLong (long)));
```
So by the time codegen comes along, it could simply loop all over the PUTARG_TYPEs (each node would also carry an argument number so that they can be built in the right order) to build the signature and emit he necessary bitcasts/upcasts/downcasts.

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -939,7 +939,7 @@ var_types Compiler::getArgTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
     }
 
     return useType;
-#endif // TARGET_WASM
+#endif // !TARGET_WASM
 }
 
 #ifdef TARGET_WASM
@@ -1276,7 +1276,7 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE     clsHnd,
     }
 
     return useType;
-#endif  // TARGET_WASM
+#endif  // !TARGET_WASM
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -756,7 +756,7 @@ var_types Compiler::getArgTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
 #if defined(TARGET_WASM)
     *wbPassStruct = SPK_ByValue;
     return TYP_STRUCT;
-#else
+#else // !TARGET_WASM
     var_types         useType         = TYP_UNKNOWN;
     structPassingKind howToPassStruct = SPK_Unknown; // We must change this before we return
 
@@ -1055,7 +1055,7 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE     clsHnd,
 #if defined(TARGET_WASM)
     *wbReturnStruct = SPK_ByValue;
     return TYP_STRUCT;
-#else // TARGET_WASM
+#else // !TARGET_WASM
     var_types         useType             = TYP_UNKNOWN;
     structPassingKind howToReturnStruct   = SPK_Unknown; // We must change this before we return
     bool              canReturnInRegister = true;
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -752,6 +752,11 @@ var_types Compiler::getArgTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
                                         bool                 isVarArg,
                                         unsigned             structSize)
 {
+// For LLVM/Wasm, always pass by value and let the lowering decide
+#if defined(TARGET_WASM)
+    *wbPassStruct = SPK_ByValue;
+    return TYP_STRUCT;
+#else
     var_types         useType         = TYP_UNKNOWN;
     structPassingKind howToPassStruct = SPK_Unknown; // We must change this before we return
 
@@ -934,6 +939,7 @@ var_types Compiler::getArgTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
     }
 
     return useType;
+#endif // TARGET_WASM
 }
 
 #ifdef TARGET_WASM
@@ -1046,6 +1052,10 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE     clsHnd,
                                            structPassingKind*       wbReturnStruct /* = nullptr */,
                                            unsigned                 structSize /* = 0 */)
 {
+#if defined(TARGET_WASM)
+    *wbReturnStruct = SPK_ByValue;
+    return TYP_STRUCT;
+#else // TARGET_WASM
     var_types         useType             = TYP_UNKNOWN;
     structPassingKind howToReturnStruct   = SPK_Unknown; // We must change this before we return
     bool              canReturnInRegister = true;
@@ -1266,6 +1276,7 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE     clsHnd,
     }
 
     return useType;
+#endif  // TARGET_WASM
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -6875,6 +6875,13 @@ GenTreeLclVar* Compiler::gtNewStoreLclVar(unsigned dstLclNum, GenTree* src)
     return store;
 }
 
+#ifdef TARGET_WASM
+GenTreePutArgType* Compiler::gtNewPutArgType(var_types type, GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
+{
+    return new (this, GT_PUTARG_TYPE) GenTreePutArgType(type, op, corInfoType, clsHnd);
+}
+#endif
+
 #ifdef FEATURE_SIMD
 //---------------------------------------------------------------------
 // gtNewSIMDVectorZero: create a GT_SIMD node for Vector<T>.Zero
@@ -13082,6 +13089,9 @@ void Compiler::gtDispLIRNode(GenTree* node, const char* prefixMsg /* = nullptr *
             }
             else
             {
+    #ifdef TARGET_WASM
+                displayOperand(operand, "wasm arg" /* TODO-LLVM */, operandArc, indentStack, prefixIndent);
+    #else
                 fgArgTabEntry* curArgTabEntry = gtArgEntryByNode(call, operand);
                 assert(curArgTabEntry);
                 assert(operand->OperGet() != GT_LIST);
@@ -13096,6 +13106,7 @@ void Compiler::gtDispLIRNode(GenTree* node, const char* prefixMsg /* = nullptr *
                 }
 
                 displayOperand(operand, buf, operandArc, indentStack, prefixIndent);
+#endif
             }
         }
         else if (node->OperIsDynBlkOp())
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -284,6 +284,10 @@ void GenTree::InitNodeSize()
 #endif // FEATURE_ARG_SPLIT
 #endif // FEATURE_PUT_STRUCT_ARG_STK
 
+#if defined(TARGET_WASM) && defined(DEBUG)
+    GenTree::s_gtNodeSizes[GT_PUTARG_TYPE]      = TREE_NODE_SZ_LARGE;
+#endif
+
     assert(GenTree::s_gtNodeSizes[GT_RETURN] == GenTree::s_gtNodeSizes[GT_ASG]);
 
     // This list of assertions should come to contain all GenTree subtypes that are declared
@@ -347,8 +351,12 @@ void GenTree::InitNodeSize()
 #endif // FEATURE_PUT_STRUCT_ARG_STK
 
 #if defined TARGET_WASM
+#ifdef DEBUG
+    static_assert_no_msg(sizeof(GenTreePutArgType)   <= TREE_NODE_SZ_LARGE);
+#else
     static_assert_no_msg(sizeof(GenTreePutArgType)   <= TREE_NODE_SZ_SMALL);
-#endif
+#endif // DEBUG
+#endif // TARGET_WASM
 
 #ifdef FEATURE_SIMD
     static_assert_no_msg(sizeof(GenTreeSIMD)         <= TREE_NODE_SZ_SMALL);
@@ -6876,9 +6884,9 @@ GenTreeLclVar* Compiler::gtNewStoreLclVar(unsigned dstLclNum, GenTree* src)
 }
 
 #ifdef TARGET_WASM
-GenTreePutArgType* Compiler::gtNewPutArgType(var_types type, GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
+GenTreePutArgType* Compiler::gtNewPutArgType(GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
 {
-    return new (this, GT_PUTARG_TYPE) GenTreePutArgType(type, op, corInfoType, clsHnd);
+    return new (this, GT_PUTARG_TYPE) GenTreePutArgType(op, corInfoType, clsHnd);
 }
 #endif
 
@@ -13089,24 +13097,32 @@ void Compiler::gtDispLIRNode(GenTree* node, const char* prefixMsg /* = nullptr *
             }
             else
             {
-    #ifdef TARGET_WASM
-                displayOperand(operand, "wasm arg" /* TODO-LLVM */, operandArc, indentStack, prefixIndent);
-    #else
-                fgArgTabEntry* curArgTabEntry = gtArgEntryByNode(call, operand);
-                assert(curArgTabEntry);
-                assert(operand->OperGet() != GT_LIST);
-
-                if (!curArgTabEntry->isLateArg())
+#ifdef TARGET_WASM
+                // LLVM rewrites the call args, but does not reinitialise the arg infos
+                if(operand->OperIs(GT_PUTARG_TYPE))
                 {
-                    gtGetArgMsg(call, operand, curArgTabEntry->argNum, buf, sizeof(buf));
+                    GenTreePutArgType* putAarg = operand->AsPutArgType();
+                    gtGetArgMsg(call, operand, putAarg->GetArgNum(), buf, sizeof(buf));
                 }
                 else
                 {
-                    gtGetLateArgMsg(call, operand, curArgTabEntry->GetLateArgInx(), buf, sizeof(buf));
-                }
+#endif // TARGET_WASM
+                    fgArgTabEntry* curArgTabEntry = gtArgEntryByNode(call, operand);
+                    assert(curArgTabEntry);
+                    assert(operand->OperGet() != GT_LIST);
 
+                    if (!curArgTabEntry->isLateArg())
+                    {
+                        gtGetArgMsg(call, operand, curArgTabEntry->argNum, buf, sizeof(buf));
+                    }
+                    else
+                    {
+                        gtGetLateArgMsg(call, operand, curArgTabEntry->GetLateArgInx(), buf, sizeof(buf));
+                    }
+#ifdef TARGET_WASM
+                }
+#endif // TARGET_WASM
                 displayOperand(operand, buf, operandArc, indentStack, prefixIndent);
-#endif
             }
         }
         else if (node->OperIsDynBlkOp())
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -346,6 +346,10 @@ void GenTree::InitNodeSize()
 #endif // FEATURE_ARG_SPLIT
 #endif // FEATURE_PUT_STRUCT_ARG_STK
 
+#if defined TARGET_WASM
+    static_assert_no_msg(sizeof(GenTreePutArgType)   <= TREE_NODE_SZ_SMALL);
+#endif
+
 #ifdef FEATURE_SIMD
     static_assert_no_msg(sizeof(GenTreeSIMD)         <= TREE_NODE_SZ_SMALL);
 #endif // FEATURE_SIMD
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -4744,7 +4744,7 @@ struct GenTreeCall final : public GenTree
     unsigned char gtReturnType : 5; // exact return type
 
 #if defined(TARGET_WASM)
-    CorInfoType gtCallIrReturnType; // to help construct the LLVM signature for string ctors
+    CorInfoType gtCorInfoType; // original type from CORINFO_SIG_INFO used to construct signature
 #endif
 
     CORINFO_CLASS_HANDLE gtRetClsHnd; // The return type handle of the call if it is a struct; always available
@@ -6694,8 +6694,25 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
 #if defined(TARGET_WASM)
 struct GenTreePutArgType : public GenTreeOp
 {
-    CorInfoType          gtCorInfoType;
-    CORINFO_CLASS_HANDLE gtClsHnd;
+private:
+    CorInfoType          m_CorInfoType;
+    CORINFO_CLASS_HANDLE m_ClsHnd;
+
+public:
+    GenTreePutArgType(var_types type, GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
+        : GenTreeOp(GT_PUTARG_TYPE, type, op, nullptr), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
+    {
+    }
+
+    CorInfoType GetCorInfoType()
+    {
+        return m_CorInfoType;
+    }
+
+    CORINFO_CLASS_HANDLE GetClsHnd()
+    {
+        return m_ClsHnd;
+    }
 };
 #endif // TARGET_WASM
 
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -4743,6 +4743,10 @@ struct GenTreeCall final : public GenTree
     unsigned char gtCallType : 3;   // value from the gtCallTypes enumeration
     unsigned char gtReturnType : 5; // exact return type
 
+#if defined(TARGET_WASM)
+    CorInfoType gtCallIrReturnType; // to help construct the LLVM signature for string ctors
+#endif
+
     CORINFO_CLASS_HANDLE gtRetClsHnd; // The return type handle of the call if it is a struct; always available
 
     union {
@@ -6686,6 +6690,14 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
 #endif
 };
 #endif // FEATURE_ARG_SPLIT
+
+#if defined(TARGET_WASM)
+struct GenTreePutArgType : public GenTreeOp
+{
+    CorInfoType          gtCorInfoType;
+    CORINFO_CLASS_HANDLE gtClsHnd;
+};
+#endif // TARGET_WASM
 
 // Represents GT_COPY or GT_RELOAD node
 //
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -6707,7 +6707,9 @@ private:
 
 public:
     GenTreePutArgType(GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
-        : GenTreeOp(GT_PUTARG_TYPE, JITtype2varType(corInfoType), op, nullptr), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
+        : GenTreeUnOp(GT_PUTARG_TYPE, JITtype2varType(corInfoType), op)
+        , m_CorInfoType(corInfoType)
+        , m_ClsHnd(clsHnd)
     {
     }
 
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -6692,24 +6692,24 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
 #endif // FEATURE_ARG_SPLIT
 
 #if defined(TARGET_WASM)
-struct GenTreePutArgType : public GenTreeOp
+struct GenTreePutArgType : public GenTreeUnOp
 {
 private:
     CorInfoType          m_CorInfoType;
-    CORINFO_CLASS_HANDLE m_ClsHnd;
+    CORINFO_CLASS_HANDLE m_ClsHnd; // The struct handle if this is a struct, the pointee type if this is a pointer.
 
 public:
     GenTreePutArgType(var_types type, GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
-        : GenTreeOp(GT_PUTARG_TYPE, type, op, nullptr), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
+        : GenTreeUnOp(GT_PUTARG_TYPE, type, op), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
     {
     }
 
-    CorInfoType GetCorInfoType()
+    CorInfoType GetCorInfoType() const
     {
         return m_CorInfoType;
     }
 
-    CORINFO_CLASS_HANDLE GetClsHnd()
+    CORINFO_CLASS_HANDLE GetClsHnd() const
     {
         return m_ClsHnd;
     }
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -29,6 +29,10 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 #include "namedintrinsiclist.h"
 #include "layout.h"
 
+#ifdef TARGET_WASM
+extern var_types JITtype2varType(CorInfoType type);
+#endif // TARGET_WASM
+
 // Debugging GenTree is much easier if we add a magic virtual function to make the debugger able to figure out what type
 // it's got. This is enabled by default in DEBUG. To enable it in RET builds (temporarily!), you need to change the
 // build to define DEBUGGABLE_GENTREE=1, as well as pass /OPT:NOICF to the linker (or else all the vtables get merged,
@@ -4744,7 +4748,7 @@ struct GenTreeCall final : public GenTree
     unsigned char gtReturnType : 5; // exact return type
 
 #if defined(TARGET_WASM)
-    CorInfoType gtCorInfoType; // original type from CORINFO_SIG_INFO used to construct signature
+    CorInfoType gtCorInfoType; // the precise return type used to construct the signature
 #endif
 
     CORINFO_CLASS_HANDLE gtRetClsHnd; // The return type handle of the call if it is a struct; always available
@@ -6697,10 +6701,13 @@ struct GenTreePutArgType : public GenTreeOp
 private:
     CorInfoType          m_CorInfoType;
     CORINFO_CLASS_HANDLE m_ClsHnd;
+#if DEBUG
+    unsigned m_argNum;
+#endif 
 
 public:
-    GenTreePutArgType(var_types type, GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
-        : GenTreeOp(GT_PUTARG_TYPE, type, op, nullptr), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
+    GenTreePutArgType(GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
+        : GenTreeOp(GT_PUTARG_TYPE, JITtype2varType(corInfoType), op, nullptr), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
     {
     }
 
@@ -6713,6 +6720,18 @@ public:
     {
         return m_ClsHnd;
     }
+
+#if DEBUG
+    unsigned GetArgNum()
+    {
+        return m_argNum;
+    }
+
+    void SetArgNum(unsigned argNum)
+    {
+        m_argNum = argNum;
+    }
+#endif // DEBUG
 };
 #endif // TARGET_WASM
 
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -6696,11 +6696,11 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
 #endif // FEATURE_ARG_SPLIT
 
 #if defined(TARGET_WASM)
-struct GenTreePutArgType : public GenTreeOp
+struct GenTreePutArgType : public GenTreeUnOp
 {
 private:
     CorInfoType          m_CorInfoType;
-    CORINFO_CLASS_HANDLE m_ClsHnd;
+    CORINFO_CLASS_HANDLE m_ClsHnd; // The struct handle if this is a struct, the pointee type if this is a pointer.
 #if DEBUG
     unsigned m_argNum;
 #endif 
@@ -6711,12 +6711,12 @@ public:
     {
     }
 
-    CorInfoType GetCorInfoType()
+    CorInfoType GetCorInfoType() const
     {
         return m_CorInfoType;
     }
 
-    CORINFO_CLASS_HANDLE GetClsHnd()
+    CORINFO_CLASS_HANDLE GetClsHnd() const
     {
         return m_ClsHnd;
     }
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1923,7 +1923,6 @@ GenTree* Llvm::createStoreNode(var_types nodeType, GenTree* addr, GenTree* nodeT
     {
         storeNode = new (_compiler, GT_STORE_OBJ)
             GenTreeObj(nodeType, addr, nodeToStore, structClassLayout);
-        storeNode->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
     }
     else
     {
@@ -2030,7 +2029,7 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
             GenTree*       slotAddr = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, lclShadowStack, offset);
             GenTree*       storeNode =
                 createStoreNode(opAndArg.operand->TypeGet(), slotAddr, opAndArg.operand,
-                                corInfoType == CORINFO_TYPE_VALUECLASS ? _compiler->typGetObjLayout(clsHnd) : nullptr);
+                                corInfoType == CORINFO_TYPE_VALUECLASS ? _compiler->typGetObjLayout(clsHnd) : NO_CLASS_HANDLE);
 
             if (corInfoType == CORINFO_TYPE_VALUECLASS)
             {
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1972,24 +1972,14 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
                                                                   // _shadowStackLocalsSize == 0, then omit the GT_ADD
     GenTree* calleeShadowStack = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
 
-    // store in the temp
-    unsigned calleeShadowStackTemp = _compiler->lvaGrabTemp(
-        false DEBUGARG("callee shadowstack")); // used to create the store locations for args on shadow stack
-    LclVarDsc* calleeShadowStackVarDsc = _compiler->lvaGetDesc(calleeShadowStackTemp);
-    calleeShadowStackVarDsc->lvType    = TYP_I_IMPL;
-    GenTree* storeShadowStack          = _compiler->gtNewStoreLclVar(calleeShadowStackTemp, calleeShadowStack);
-
-    // create the PUTARG
-    GenTree*           lclCalleeShadowStack = _compiler->gtNewLclvNode(calleeShadowStackTemp, TYP_I_IMPL);
-    GenTreePutArgType* putArg = _compiler->gtNewPutArgType(TYP_I_IMPL, lclCalleeShadowStack, CORINFO_TYPE_PTR, nullptr);
+    GenTreePutArgType* putArg = _compiler->gtNewPutArgType(TYP_I_IMPL, calleeShadowStack, CORINFO_TYPE_PTR, NO_CLASS_HANDLE);
 
     callNode->gtCallArgs     = _compiler->gtNewCallArgs(putArg);
     lastArg                  = callNode->gtCallArgs;
     insertReturnAfter        = lastArg; // add the return slot after the shadow stack arg
     callNode->gtCallLateArgs = nullptr;
 
-    CurrentRange().InsertBefore(callNode, shadowStackVar, offset, calleeShadowStack, storeShadowStack);
-    CurrentRange().InsertAfter(storeShadowStack, lclCalleeShadowStack, putArg);
+    CurrentRange().InsertBefore(callNode, shadowStackVar, offset, calleeShadowStack, putArg);
 
     lastArg = lowerCallReturn(callNode, calleeSigInfo, insertReturnAfter);
 
@@ -2018,18 +2008,18 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
         bool argOnShadowStack = isThis || !canStoreArgOnLlvmStack(corInfoType, clsHnd);
         if (argOnShadowStack)
         {
-            GenTree* lclShadowStack = _compiler->gtNewLclvNode(calleeShadowStackTemp, TYP_I_IMPL);
+            GenTree* lclShadowStack = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
 
             if (corInfoType == CORINFO_TYPE_VALUECLASS)
             {
                 shadowStackUseOffest = padOffset(corInfoType, clsHnd, shadowStackUseOffest);
             }
 
-            GenTreeIntCon* offset   = _compiler->gtNewIconNode(shadowStackUseOffest, TYP_I_IMPL);
-            GenTree*       slotAddr = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, lclShadowStack, offset);
+            GenTreeIntCon* offset    = _compiler->gtNewIconNode(shadowStackUseOffest + _shadowStackLocalsSize, TYP_I_IMPL);
+            GenTree*       slotAddr  = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, lclShadowStack, offset);
             GenTree*       storeNode =
                 createStoreNode(opAndArg.operand->TypeGet(), slotAddr, opAndArg.operand,
-                                corInfoType == CORINFO_TYPE_VALUECLASS ? _compiler->typGetObjLayout(clsHnd) : NO_CLASS_HANDLE);
+                                corInfoType == CORINFO_TYPE_VALUECLASS ? _compiler->typGetObjLayout(clsHnd) : nullptr);
 
             if (corInfoType == CORINFO_TYPE_VALUECLASS)
             {
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1967,9 +1967,8 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
 
     // set up the callee shadowstack, creating a temp and the PUTARG
     GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
-    GenTreeIntCon* offset         = _compiler->gtNewIconNode(_shadowStackLocalsSize,
-                                                     TYP_I_IMPL); // TODO-LLVM: possible performance benefit: when
-                                                                  // _shadowStackLocalsSize == 0, then omit the GT_ADD
+    GenTreeIntCon* offset         = _compiler->gtNewIconNode(_shadowStackLocalsSize, TYP_I_IMPL);
+    // TODO-LLVM: possible performance benefit: when _shadowStackLocalsSize == 0, then omit the GT_ADD.
     GenTree* calleeShadowStack = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
 
     GenTreePutArgType* putArg = _compiler->gtNewPutArgType(TYP_I_IMPL, calleeShadowStack, CORINFO_TYPE_PTR, NO_CLASS_HANDLE);
@@ -2015,7 +2014,7 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
                 shadowStackUseOffest = padOffset(corInfoType, clsHnd, shadowStackUseOffest);
             }
 
-            GenTreeIntCon* offset    = _compiler->gtNewIconNode(shadowStackUseOffest + _shadowStackLocalsSize, TYP_I_IMPL);
+            GenTreeIntCon* offset    = _compiler->gtNewIconNode(_shadowStackLocalsSize + shadowStackUseOffest, TYP_I_IMPL);
             GenTree*       slotAddr  = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, lclShadowStack, offset);
             GenTree*       storeNode =
                 createStoreNode(opAndArg.operand->TypeGet(), slotAddr, opAndArg.operand,
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1833,7 +1833,6 @@ GenTreeCall::Use* Llvm::lowerCallReturn(GenTreeCall*      callNode,
         if (callReturnType == TYP_STRUCT)
         {
             indirNode    = _compiler->gtNewObjNode(calleeSigInfo.retTypeClass, returnAddrLclAfterCall);
-            indirNode->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
         }
         else
         {
@@ -1868,6 +1867,7 @@ GenTreeCall::Use* Llvm::lowerCallReturn(GenTreeCall*      callNode,
     {
         callNode->gtCorInfoType = calleeSigInfo.retType;
     }
+
     return lastArg;
 }
 
@@ -1967,9 +1967,8 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
 
     // set up the callee shadowstack, creating a temp and the PUTARG
     GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
-    GenTreeIntCon* offset         = _compiler->gtNewIconNode(_shadowStackLocalsSize,
-                                                     TYP_I_IMPL); // TODO-LLVM: possible performance benefit: when
-                                                                  // _shadowStackLocalsSize == 0, then omit the GT_ADD
+    GenTreeIntCon* offset         = _compiler->gtNewIconNode(_shadowStackLocalsSize, TYP_I_IMPL);
+    // TODO-LLVM: possible performance benefit: when _shadowStackLocalsSize == 0, then omit the GT_ADD.
     GenTree* calleeShadowStack = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
 
     GenTreePutArgType* calleeShadowStackPutArg =
@@ -2019,7 +2018,7 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
                 shadowStackUseOffest = padOffset(corInfoType, clsHnd, shadowStackUseOffest);
             }
 
-            GenTreeIntCon* offset    = _compiler->gtNewIconNode(shadowStackUseOffest + _shadowStackLocalsSize, TYP_I_IMPL);
+            GenTreeIntCon* offset    = _compiler->gtNewIconNode(_shadowStackLocalsSize + shadowStackUseOffest, TYP_I_IMPL);
             GenTree*       slotAddr  = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, lclShadowStack, offset);
             GenTree*       storeNode =
                 createStoreNode(opAndArg.operand->TypeGet(), slotAddr, opAndArg.operand,
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1865,6 +1865,7 @@ GenTreeCall::Use* Llvm::lowerCallReturn(GenTreeCall*      callNode,
     {
         callNode->gtCorInfoType = calleeSigInfo.retType;
     }
+
     return lastArg;
 }
 
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1833,7 +1833,6 @@ GenTreeCall::Use* Llvm::lowerCallReturn(GenTreeCall*      callNode,
         if (callReturnType == TYP_STRUCT)
         {
             indirNode    = _compiler->gtNewObjNode(calleeSigInfo.retTypeClass, returnAddrLclAfterCall);
-            indirNode->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
         }
         else
         {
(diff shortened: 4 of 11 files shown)


yowl:
cc @SingleAccretion

> So the solution was to wrap the argument nodes into PUTARG_TYPE nodes that would carry their true types:

Was there a place you had in mind where to store the "true type"?  I'm going to need the `CORINFO_TYPE` and class handle if exists, I think.

SingleAccretion:
> Was there a place you had in mind where to store the "true type"? I'm going to need the CORINFO_TYPE and class handle if exists, I think.

In the node itself, say `GenTreePutArgType`, derived from `GenTreeUnOp`. As we are using an existing oper, we should ideally avoid having to add support for it to the various IR traversal facilities (there are _a lot_ of them, unfortunately).

src/coreclr/jit/compiler.cpp
```diff
@@ -934,6 +939,7 @@ var_types Compiler::getArgTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
     }
 
     return useType;
+#endif // TARGET_WASM
```
```suggestion
#endif // !TARGET_WASM
```
The convention for the last `#ifdef`s is to signify what is above it.

src/coreclr/jit/compiler.cpp
```diff
@@ -1266,6 +1276,7 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE     clsHnd,
     }
 
     return useType;
+#endif  // TARGET_WASM
```
```suggestion
#endif  // !TARGET_WASM
```

src/coreclr/jit/gentree.h
```diff
@@ -4743,6 +4743,10 @@ struct GenTreeCall final : public GenTree
     unsigned char gtCallType : 3;   // value from the gtCallTypes enumeration
     unsigned char gtReturnType : 5; // exact return type
 
+#if defined(TARGET_WASM)
+    CorInfoType gtCallIrReturnType; // to help construct the LLVM signature for string ctors
```
SingleAccretion: In what scenarios do we see this and the node's actual type diverge?

src/coreclr/jit/gentree.h
```diff
@@ -6687,6 +6691,14 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
 };
 #endif // FEATURE_ARG_SPLIT
 
+#if defined(TARGET_WASM)
+struct GenTreePutArgType : public GenTreeOp
+{
+    CorInfoType          gtCorInfoType;
+    CORINFO_CLASS_HANDLE gtClsHnd;
+};
+#endif // TARGET_WASM
```
SingleAccretion: I know some other Jit code may not seem like it cares, but let's make this node "modern" :).

Make the fields private, prefix them with `m_` (`m_corInfoType`, `m_clsHnd`), add a constructor and `Get` accessor methods.

src/coreclr/jit/llvm.cpp
```diff
@@ -479,7 +480,23 @@ bool Llvm::needsReturnStackSlot(CorInfoType corInfoType, CORINFO_CLASS_HANDLE cl
...
+        // IntPtr = 0x0C,
+        // UIntPtr = 0x0D,
+
+        // CORINFO_TYPE_NATIVEINT = 0xc,
+        // CORINFO_TYPE_NATIVEUINT = 0xd,
+
+        // alternative in canStoreArgOnLlvmStack map nativeint to true
+
+        stripped = CORINFO_TYPE_PTR;
+    }
```
SingleAccretion: I am still not sure I understand this.

`IntPtr` aka `CORINFO_TYPE_NATIVEINT` aka `native int`, aka `I`, aka `nint` (in C#) are all the same type and are native integers, not pointers. What makes us want to map them to pointers - what fails if we do not?

src/coreclr/jit/llvm.cpp
```diff
@@ -892,100 +894,44 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
...
-    argVec.push_back(shadowStackForCallee); // TODO-LLVM: we are not moving this past any return address args, so the return address is clobbering the first slot in the shadow stack.
-                                            // As the return is the last thing executed this seems ok, but then the question becomes why do we need the return address arg at all
-                                            // apart from compatibility with IL->LLVM?
-
-    unsigned int                      shadowStackUseOffest = 0;
-    int                               argIx                = 0;
-    GenTree*                          thisArg              = nullptr;
     fgArgInfo*                        argInfo              = call->fgArgInfo;
     unsigned int                      argCount             = argInfo->ArgCount();
     fgArgTabEntry**                   argTable             = argInfo->ArgTable();
```
```

    fgArgInfo*                        argInfo              = call->fgArgInfo;
    unsigned int                      argCount             = argInfo->ArgCount();
    fgArgTabEntry**                   argTable             = argInfo->ArgTable();
```
Hmm, some of these whitespaces are definitely not necessary.

src/coreclr/jit/llvm.cpp
```diff
@@ -892,100 +894,44 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
...
-            shadowStackUseOffest += TARGET_POINTER_SIZE;
-        }
-        argIx++;
+    if (llvmFunc == nullptr)
+    {
+        CORINFO_ARG_LIST_HANDLE sigArgs = sigInfo.args;
+
+        // assume ExternalLinkage, if the function is defined in the clrjit module, then it is replaced and an extern
+        // added to the Ilc module
+        llvmFunc = Function::Create(getFunctionTypeForSigInfo(call->gtCallIrReturnType, sigInfo),
```
SingleAccretion: Hmm, the idea was that we'd determine the `FunctionType` from the arguments in IR and only them.

src/coreclr/jit/llvm.cpp
```diff
@@ -1301,6 +1247,15 @@ void Llvm::fillPhis()
...
             {
+                //TODO-LLVM: if a phi arg is an arg that is not in the LLVM args, then we may not have a local for it
+                // For now as this PR is already big, just fail these functions.
+                // For a fix, we could do another lowering pass after SSA, copying
+                // those args into locals.  We may have to do that anyway to get the address of locals, although that might
+                // be ok to do before SSA
+                if (!(_compiler->lvaIsParameter(lclNum) && ssaNum == SsaConfig::FIRST_SSA_NUM))
+                {
+                    failFunctionCompilation();
+                }
```
SingleAccretion: What are the scenarios that fail here?

One I can think of is uninitialized locals ("use before def locals"), it seems overkill to replace them "whole", inserting a "fake" def (say to zero for a start, though I believe LLVM provides more sophisticated mechanisms to indicate a local is uninitialized) in the first block seems enough.

src/coreclr/jit/llvm.h
```diff
@@ -175,8 +177,11 @@ class Llvm
     unsigned int getTotalLocalOffset();
     bool helperRequiresShadowStack(CORINFO_METHOD_HANDLE corinfoMethodHnd);
     void importStoreInd(GenTreeStoreInd* storeIndOp);
-    bool isThisArg(GenTreeCall* call, GenTree* operand);
     Value* localVar(GenTreeLclVar* lclVar);
+
+    GenTreeCall::Use* lowerCallReturn(GenTreeCall* callNode, CORINFO_SIG_INFO& calleeSigInfo, GenTreeCall::Use* lastArg);
+    void lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calleeSigInfo);
+    void lowerToShadowStack();
```
SingleAccretion, line 184: Nit: separate the `lowerToShadowStack` (logically lowering) and `mapGenTreeToValue` (logically codegen) with a space?

src/coreclr/jit/morph.cpp
```diff
@@ -3292,7 +3292,16 @@ void Compiler::fgInitArgInfo(GenTreeCall* call)
...
+#if defined TARGET_WASM
+            // LLVM/Wasm wraps all args in a PUTARG when lowering, but not in morphing so have to handle both ways
+            // .. or exclude the assert for LLVM 
+            if(args->GetNode()->OperIs(GT_PUTARG_TYPE))
+            {
+                assert(argx == args->GetNode()->AsPutArgType()->gtGetOp1());
+            }
+            else
+#endif
+                assert(argx == args->GetNode());
```
SingleAccretion: Honestly, doesn't seem terribly valuable to have this large `#ifdef` structure, can be just:
```cpp
#ifndef TARGET_WASM // We don't update the node in the arg info for LLVM.
assert(argx == args->GetNode());
#endif
```
We should target getting rid of the codegen's dependency on arg info and use the `PUTARG_TYPE`s to relay the relevant information.

src/coreclr/jit/morph.cpp
```diff
@@ -3331,7 +3340,15 @@ void Compiler::fgInitArgInfo(GenTreeCall* call)
...
-            byteSize = genTypeSize(argType);
+#if defined(TARGET_WASM)
+            if (isStructArg)
+            {
+                // byteSize is used when creating stk args, LLVM/Wasm does not use it so anything nonzero is ok?
+                byteSize = TARGET_POINTER_SIZE;
+            }
+            else
+#endif
+                byteSize = genTypeSize(argType);
```
SingleAccretion: This should have a `TODO-LLVM: stop calling "fgInitArgInfo" in the backend`.

(Applies to the assert about `argx == args->GetNode()` above too)

src/coreclr/jit/morph.cpp
```diff
@@ -3015,7 +3015,7 @@ void Compiler::fgInitArgInfo(GenTreeCall* call)
     SYSTEMV_AMD64_CORINFO_STRUCT_REG_PASSING_DESCRIPTOR structDesc;
 #endif // UNIX_AMD64_ABI
 
-#if defined(DEBUG)
+#if defined(DEBUG) && !defined(TARGET_WASM) /* For LLVM some args are lowered to the shadow stack so we can't compare arg counts */
```
SingleAccretion:
```suggestion
#if defined(DEBUG) && !defined(TARGET_WASM) // For LLVM some args are lowered to the shadow stack so we can't compare arg counts.
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+    //             EETypePtr(interfaceType)));
+    //
+    // / --*t2 ref this in rcx
+    // + --*t19 int arg2 in rdx
+    // +--*t137 int arg1 in r10                  N007(27, 18)[000020]-- CXG-- -----t20 =
+    // *CALLV stub struct System.Runtime.InteropServices.IDynamicInterfaceCastable.GetInterfaceImplementation
+    if (callNode->IsVirtual())
+    {
+        // TODO-LLVM: what is the 3rd argument in this example and how can we get its type?
+        // looks like an interface call that in IL would be done with a call to FindInterfaceMethodImplementationTarget
```
SingleAccretion: The `stub` of `CALLV stub struct` tells me this is a VSD dispatch and the `t137` is the address hidden argument VSD uses. So it is always `TYP_I_IMPL`.

SingleAccretion (commented):
Just nits and question for now, I am yet to go through the lowering itself.

src/coreclr/jit/gentree.h
yowl: This is string constructors that return `string`, so for example I think it was
```
        string Ctor(char[] value, int startIndex, int length)
        {
```
Want me to get the exact IR/method ?

SingleAccretion: > Want me to get the exact IR/method ?

No, I understand the string constructor situation, but for them - the IR node type would match this type (it would be `TYP_REF`). What I am asking is when `JITType2varType(gtCallIrReturnType) != call->TypeGet()`, i. e. when can we not use the node's type when the signature type disagrees with the node type.

yowl: Sorry, I think I was confused.   Let me address https://github.com/dotnet/runtimelab/pull/1744#discussion_r757692693 first and then this might go away.

src/coreclr/jit/llvm.cpp
yowl: right, I seem to constantly be fighting the clang-format file.

src/coreclr/jit/llvm.cpp
yowl: Thanks for bringing me back on track!   Yes, this was fairly straightforward.  I do still need something on the `GenTreeCall` to hold the true return type, but with that, this becomes `getFunctionTypeForCall(call)` and 
```
FunctionType* Llvm::getFunctionTypeForCall(GenTreeCall* call)
{
    llvm::Type* retLlvmType = getLlvmTypeForCorInfoType(call->gtCallIrReturnType, call->gtRetClsHnd);

    std::vector<llvm::Type*> argVec = std::vector<llvm::Type*>();

    for (GenTreeCall::Use& use : call->Args())
    {
        GenTreePutArgType* putArg = use.GetNode()->AsPutArgType();
        argVec.push_back(getLlvmTypeForCorInfoType(putArg->gtCorInfoType, putArg->gtClsHnd));
    }

    return FunctionType::get(retLlvmType, ArrayRef<Type*>(argVec), false);
}
```
I'll rename `gtCallIrReturnType`

src/coreclr/jit/llvm.cpp
yowl: `IntPtr__CompareTo` is the method.

`t194` is the problem here.  There is no `d` for 2:1 and all the predecessor are above it.  When I wrote the comment I thought it was a function argument, but its not, so I'm not sure what's gone wrong.
```

------------ BB02 [000..009) -> BB04 (cond), preds={BB01} succs={BB03,BB04}
               [000100] ------------                 IL_OFFSET void   IL offset: 0x0
               [000101] ------------                 IL_OFFSET void   IL offset: 0x1
               [000126] ------------       t126 =    CNS_INT   int    4
               [000127] ------------       t127 =    LCL_VAR   int    V12 tmp4         u:1
                                                  /--*  t127   int
                                                  +--*  t126   int
               [000128] ------------       t128 = *  ADD       int
                                                  /--*  t128   int
N001 (  3,  2) [000002] ------------         t2 = *  IND       ref
N002 (  1,  1) [000003] ------------         t3 =    CNS_INT   ref    null
                                                  /--*  t2     ref
                                                  +--*  t3     ref
N003 (  8,  4) [000004] ------------         t4 = *  EQ        int
                                                  /--*  t4     int
N005 ( 12,  7) [000006] DA----------              *  STORE_LCL_VAR int    V03 loc1         d:1
               [000102] ------------                 IL_OFFSET void   IL offset: 0x6
N001 (  3,  2) [000007] ------------         t7 =    LCL_VAR   int    V03 loc1         u:1 (last use)
N002 (  1,  1) [000008] ------------         t8 =    CNS_INT   int    0
                                                  /--*  t7     int
                                                  +--*  t8     int
N003 (  5,  4) [000009] J------N----         t9 = *  EQ        int
                                                  /--*  t9     int
N004 (  7,  6) [000010] ------------              *  JTRUE     void

------------ BB03 [009..00E) -> BB14 (always), preds={BB02} succs={BB14}
               [000103] ------------                 IL_OFFSET void   IL offset: 0x9
               [000104] ------------                 IL_OFFSET void   IL offset: 0xa
N001 (  1,  1) [000087] ------------        t87 =    CNS_INT   int    1
                                                  /--*  t87    int
N003 (  5,  4) [000089] DA----------              *  STORE_LCL_VAR int    V04 loc2         d:5
               [000105] ------------                 IL_OFFSET void   IL offset: 0xc

------------ BB04 [00E..016) -> BB06 (cond), preds={BB02} succs={BB05,BB06}
               [000106] ------------                 IL_OFFSET void   IL offset: 0xe
               [000129] ------------       t129 =    CNS_INT   int    4
               [000130] ------------       t130 =    LCL_VAR   int    V12 tmp4         u:1
                                                  /--*  t130   int
                                                  +--*  t129   int
               [000131] ------------       t131 = *  ADD       int
                                                  /--*  t131   int
N003 (  3,  2) [000011] ------------        t11 = *  IND       ref
N004 (  1,  4) [000012] H------N----        t12 =    CNS_INT(h) int    0x420050 class
                                                  /--*  t11    ref    wasm arg
                                                  +--*  t12    int    wasm arg
N005 ( 18, 10) [000013] --C-G-------        t13 = *  CALL help ref    HELPER.CORINFO_HELP_ISINSTANCEOFCLASS
N006 (  1,  1) [000014] ------------        t14 =    CNS_INT   ref    null
                                                  /--*  t13    ref
                                                  +--*  t14    ref
N007 ( 20, 12) [000015] J---G--N----        t15 = *  EQ        int
                                                  /--*  t15    int
N008 ( 22, 14) [000016] ----G-------              *  JTRUE     void

------------ BB05 [016..020) -> BB07 (always), preds={BB04} succs={BB07}
               [000107] ------------                 IL_OFFSET void   IL offset: 0x16
               [000132] ------------       t132 =    CNS_INT   int    4
               [000133] ------------       t133 =    LCL_VAR   int    V12 tmp4         u:1
                                                  /--*  t133   int
                                                  +--*  t132   int
               [000134] ------------       t134 = *  ADD       int
                                                  /--*  t134   int
N003 (  3,  2) [000075] ------------        t75 = *  IND       ref
N004 (  1,  4) [000076] H-----------        t76 =    CNS_INT(h) int    0x420050 class
                                                  /--*  t75    ref    wasm arg
                                                  +--*  t76    int    wasm arg
N005 ( 18, 10) [000077] --CXG-------        t77 = *  CALL help byref  HELPER.CORINFO_HELP_UNBOX
                                                  /--*  t77    byref
N006 ( 21, 12) [000078] *--XG-------        t78 = *  IND       int
                                                  /--*  t78    int
N008 ( 25, 15) [000080] DA-XG-------              *  STORE_LCL_VAR int    V02 loc0         d:3
               [000108] ------------                 IL_OFFSET void   IL offset: 0x1d
N001 (  1,  1) [000081] ------------        t81 =    CNS_INT   int    1
                                                  /--*  t81    int
N003 (  5,  4) [000084] DA----------              *  STORE_LCL_VAR int    V09 tmp1         d:3

------------ BB06 [020..021), preds={BB04} succs={BB07}
               [000109] ------------                 IL_OFFSET void   IL offset: 0x20
N001 (  1,  1) [000017] ------------        t17 =    CNS_INT   int    0
                                                  /--*  t17    int
N003 (  5,  4) [000019] DA----------              *  STORE_LCL_VAR int    V09 tmp1         d:2

------------ BB07 [021..025) -> BB13 (cond), preds={BB05,BB06} succs={BB08,BB13}
     (  0,  0) [000196] ------------       t196 =    PHI_ARG   int    V02 loc0         u:3
     (  0,  0) [000195] ------------       t195 =    PHI_ARG   int    V09 tmp1         u:3
     (  0,  0) [000194] ------------       t194 =    PHI_ARG   int    V02 loc0         u:1
     (  0,  0) [000193] ------------       t193 =    PHI_ARG   int    V09 tmp1         u:2
                                                  /--*  t195   int
                                                  +--*  t193   int
               [000188] ------------       t188 = *  PHI       int
                                                  /--*  t188   int
               [000189] DA----------              *  STORE_LCL_VAR int    V09 tmp1         d:1
                                                  /--*  t196   int
                                                  +--*  t194   int
               [000186] ------------       t186 = *  PHI       int
                                                  /--*  t186   int
               [000187] DA----------              *  STORE_LCL_VAR int    V02 loc0         d:2
N001 (  3,  2) [000021] ------------        t21 =    LCL_VAR   int    V09 tmp1         u:1 (last use)
                                                  /--*  t21    int
N002 (  4,  4) [000095] ------------        t95 = *  CAST      int <- bool <- int
                                                  /--*  t95    int
N004 (  8,  7) [000023] DA----------              *  STORE_LCL_VAR int    V05 loc3         d:1
               [000110] ------------                 IL_OFFSET void   IL offset: 0x22
N001 (  3,  2) [000024] ------------        t24 =    LCL_VAR   int    V05 loc3         u:1 (last use)
N002 (  1,  1) [000025] ------------        t25 =    CNS_INT   int    0
                                                  /--*  t24    int
                                                  +--*  t25    int
N003 (  5,  4) [000026] J------N----        t26 = *  EQ        int
                                                  /--*  t26    int
N004 (  7,  6) [000027] ------------              *  JTRUE     void

------------ BB08 [025..035) -> BB10 (cond), preds={BB07} succs={BB09,BB10}
               [000111] ------------                 IL_OFFSET void   IL offset: 0x25
               [000112] ------------                 IL_OFFSET void   IL offset: 0x26
               [000135] ------------       t135 =    CNS_INT   int    0
               [000136] ------------       t136 =    LCL_VAR   int    V12 tmp4         u:1
                                                  /--*  t136   int
                                                  +--*  t135   int
               [000137] ------------       t137 = *  ADD       int
                                                  /--*  t137   int
N001 (  3,  2) [000041] ------------        t41 = *  IND       byref  Zero Fseq[_value]
                                                  /--*  t41    byref
N002 (  6,  4) [000042] *--XG-------        t42 = *  IND       int
N003 (  3,  2) [000043] ------------        t43 =    LCL_VAR   int    V02 loc0         u:2
                                                  /--*  t42    int
                                                  +--*  t43    int
N004 ( 13,  7) [000044] ---XG-------        t44 = *  LT        int
                                                  /--*  t44    int
N006 ( 17, 10) [000046] DA-XG-------              *  STORE_LCL_VAR int    V06 loc4         d:1
               [000113] ------------                 IL_OFFSET void   IL offset: 0x31
N001 (  3,  2) [000047] ------------        t47 =    LCL_VAR   int    V06 loc4         u:1 (last use)
N002 (  1,  1) [000048] ------------        t48 =    CNS_INT   int    0
                                                  /--*  t47    int
                                                  +--*  t48    int
N003 (  5,  4) [000049] J------N----        t49 = *  EQ        int
                                                  /--*  t49    int
N004 (  7,  6) [000050] ------------              *  JTRUE     void

------------ BB09 [035..039) -> BB14 (always), preds={BB08} succs={BB14}
               [000114] ------------                 IL_OFFSET void   IL offset: 0x35
N001 (  1,  1) [000071] ------------        t71 =    CNS_INT   int    -1
                                                  /--*  t71    int
N003 (  5,  4) [000073] DA----------              *  STORE_LCL_VAR int    V04 loc2         d:4
               [000115] ------------                 IL_OFFSET void   IL offset: 0x37

------------ BB10 [039..048) -> BB12 (cond), preds={BB08} succs={BB11,BB12}
               [000116] ------------                 IL_OFFSET void   IL offset: 0x39
               [000138] ------------       t138 =    CNS_INT   int    0
               [000139] ------------       t139 =    LCL_VAR   int    V12 tmp4         u:1 (last use)
                                                  /--*  t139   int
                                                  +--*  t138   int
               [000140] ------------       t140 = *  ADD       int
                                                  /--*  t140   int
N001 (  3,  2) [000051] ------------        t51 = *  IND       byref  Zero Fseq[_value]
                                                  /--*  t51    byref
N002 (  6,  4) [000052] *--XG-------        t52 = *  IND       int
N003 (  3,  2) [000053] ------------        t53 =    LCL_VAR   int    V02 loc0         u:2 (last use)
                                                  /--*  t52    int
                                                  +--*  t53    int
N004 ( 13,  7) [000054] ---XG-------        t54 = *  GT        int
                                                  /--*  t54    int
N006 ( 17, 10) [000056] DA-XG-------              *  STORE_LCL_VAR int    V07 loc5         d:1
               [000117] ------------                 IL_OFFSET void   IL offset: 0x44
N001 (  3,  2) [000057] ------------        t57 =    LCL_VAR   int    V07 loc5         u:1 (last use)
N002 (  1,  1) [000058] ------------        t58 =    CNS_INT   int    0
                                                  /--*  t57    int
                                                  +--*  t58    int
N003 (  5,  4) [000059] J------N----        t59 = *  EQ        int
                                                  /--*  t59    int
N004 (  7,  6) [000060] ------------              *  JTRUE     void

------------ BB11 [048..04C) -> BB14 (always), preds={BB10} succs={BB14}
               [000118] ------------                 IL_OFFSET void   IL offset: 0x48
N001 (  1,  1) [000067] ------------        t67 =    CNS_INT   int    1
                                                  /--*  t67    int
N003 (  5,  4) [000069] DA----------              *  STORE_LCL_VAR int    V04 loc2         d:3
               [000119] ------------                 IL_OFFSET void   IL offset: 0x4a

------------ BB12 [04C..050) -> BB14 (always), preds={BB10} succs={BB14}
               [000120] ------------                 IL_OFFSET void   IL offset: 0x4c
N001 (  1,  1) [000061] ------------        t61 =    CNS_INT   int    0
                                                  /--*  t61    int
N003 (  5,  4) [000063] DA----------              *  STORE_LCL_VAR int    V04 loc2         d:2
               [000121] ------------                 IL_OFFSET void   IL offset: 0x4e

------------ BB13 [050..05B) (throw), preds={BB07} succs={}
               [000122] ------------                 IL_OFFSET void   IL offset: 0x50
               [000141] ------------       t141 =    LCL_VAR   int    V12 tmp4         u:1
               [000142] ------------       t142 =    CNS_INT   int    16
                                                  /--*  t141   int
                                                  +--*  t142   int
               [000143] ------------       t143 = *  ADD       int
                                                  /--*  t143   int
               [000144] DA----------              *  STORE_LCL_VAR int    V13 tmp5         d:1
               [000145] ------------       t145 =    LCL_VAR   int    V13 tmp5         u:1 (last use)
                                                  /--*  t145   int
               [000146] ------------       t146 = *  PUTARG_TYPE int
               [000147] ------------       t147 =    LCL_VAR   int    V12 tmp4         u:1
               [000148] ------------       t148 =    CNS_INT   int    16
                                                  /--*  t147   int
                                                  +--*  t148   int
               [000149] ------------       t149 = *  ADD       int
                                                  /--*  t149   int
               [000150] DA----------              *  STORE_LCL_VAR int    V14 tmp6         d:1
               [000151] ------------       t151 =    LCL_VAR   int    V14 tmp6         u:1
                                                  /--*  t151   int
               [000152] ------------       t152 = *  PUTARG_TYPE int
                                                  /--*  t146   int    wasm arg
                                                  +--*  t152   int    wasm arg
N001 ( 14,  2) [000028] --CXG-------              *  CALL      void   System.SR.get_Arg_MustBeIntPtr
               [000153] ------------       t153 =    LCL_VAR   int    V14 tmp6         u:1 (last use)
                                                  /--*  t153   int
               [000154] s-----------       t154 = *  IND       ref
               [000155] ------------       t155 =    CNS_INT   int    8
               [000156] ------------       t156 =    LCL_VAR   int    V12 tmp4         u:1
                                                  /--*  t156   int
                                                  +--*  t155   int
               [000157] ------------       t157 = *  ADD       int
                                                  /--*  t157   int
                                                  +--*  t154   ref
N003 ( 18,  5) [000030] -A-XG-------              *  STOREIND  ref
               [000123] ------------                 IL_OFFSET void   IL offset: 0x55
N002 (  1,  4) [000032] H-----------        t32 =    CNS_INT(h) int    0x420098 method
                                                  /--*  t32    int    wasm arg
N003 ( 15,  7) [000033] --C---------        t33 = *  CALL help ref    HELPER.CORINFO_HELP_NEWSFAST
               [000158] ------------       t158 =    CNS_INT   int    12
               [000159] ------------       t159 =    LCL_VAR   int    V12 tmp4         u:1
                                                  /--*  t159   int
                                                  +--*  t158   int
               [000160] ------------       t160 = *  ADD       int
                                                  /--*  t160   int
                                                  +--*  t33    ref
N005 ( 19, 10) [000035] -A----------              *  STOREIND  ref
               [000161] ------------       t161 =    CNS_INT   int    12
               [000162] ------------       t162 =    LCL_VAR   int    V12 tmp4         u:1
                                                  /--*  t162   int
                                                  +--*  t161   int
               [000163] ------------       t163 = *  ADD       int
                                                  /--*  t163   int
N003 (  3,  2) [000036] ------------        t36 = *  IND       ref
               [000164] ------------       t164 =    CNS_INT   int    8
               [000165] ------------       t165 =    LCL_VAR   int    V12 tmp4         u:1
                                                  /--*  t165   int
                                                  +--*  t164   int
               [000166] ------------       t166 = *  ADD       int
                                                  /--*  t166   int
N004 (  3,  2) [000031] ------------        t31 = *  IND       ref
               [000167] ------------       t167 =    LCL_VAR   int    V12 tmp4         u:1
               [000168] ------------       t168 =    CNS_INT   int    16
                                                  /--*  t167   int
                                                  +--*  t168   int
               [000169] ------------       t169 = *  ADD       int
                                                  /--*  t169   int
               [000170] DA----------              *  STORE_LCL_VAR int    V15 tmp7         d:1
               [000171] ------------       t171 =    LCL_VAR   int    V15 tmp7         u:1
                                                  /--*  t171   int
               [000172] ------------       t172 = *  PUTARG_TYPE int
               [000173] ------------       t173 =    LCL_VAR   int    V15 tmp7         u:1
               [000174] ------------       t174 =    CNS_INT   int    0
                                                  /--*  t173   int
                                                  +--*  t174   int
               [000175] ------------       t175 = *  ADD       int
                                                  /--*  t175   int
                                                  +--*  t36    ref
               [000176] ------------              *  STOREIND  ref
               [000177] ------------       t177 =    LCL_VAR   int    V15 tmp7         u:1 (last use)
               [000178] ------------       t178 =    CNS_INT   int    4
                                                  /--*  t177   int
                                                  +--*  t178   int
               [000179] ------------       t179 = *  ADD       int
                                                  /--*  t179   int
                                                  +--*  t31    ref
               [000180] ------------              *  STOREIND  ref
                                                  /--*  t172   int    wasm arg
N005 ( 20,  9) [000037] --CXG-------              *  CALL      void   System.ArgumentException..ctor
               [000124] ------------                 IL_OFFSET void   IL offset: 0x5a
               [000181] ------------       t181 =    CNS_INT   int    12
               [000182] ------------       t182 =    LCL_VAR   int    V12 tmp4         u:1 (last use)
                                                  /--*  t182   int
                                                  +--*  t181   int
               [000183] ------------       t183 = *  ADD       int
                                                  /--*  t183   int
N002 (  3,  2) [000038] ------------        t38 = *  IND       ref
                                                  /--*  t38    ref    wasm arg
N003 ( 17,  5) [000039] --CXG-------              *  CALL help void   HELPER.CORINFO_HELP_THROW

------------ BB14 [05B..05D) (return), preds={BB03,BB09,BB11,BB12} succs={}
     (  0,  0) [000197] ------------       t197 =    PHI_ARG   int    V04 loc2         u:5
     (  0,  0) [000192] ------------       t192 =    PHI_ARG   int    V04 loc2         u:4
     (  0,  0) [000191] ------------       t191 =    PHI_ARG   int    V04 loc2         u:3
     (  0,  0) [000190] ------------       t190 =    PHI_ARG   int    V04 loc2         u:2
                                                  /--*  t197   int
                                                  +--*  t192   int
                                                  +--*  t191   int
                                                  +--*  t190   int
               [000184] ------------       t184 = *  PHI       int
                                                  /--*  t184   int
               [000185] DA----------              *  STORE_LCL_VAR int    V04 loc2         d:1
               [000125] ------------                 IL_OFFSET void   IL offset: 0x5b
N001 (  3,  2) [000065] ------------        t65 =    LCL_VAR   int    V04 loc2         u:1 (last use)
                                                  /--*  t65    int
N002 (  4,  3) [000066] ------------              *  RETURN    int
```

SingleAccretion: Could you share the full dump? FWIW, it looks like an "uninitialized local" case to me looking at this IR (and in this case I think the PHI is also "dead", though that will not always be the case and it is something we should tolerate).

So, there are two cases: actual uninitialized locals and those that are must-init (with zero). The logic to decide which is which [is not straightforward](https://github.com/dotnet/runtime/blob/03e0187a1a912f38e718c00614201dd20fa1bff1/src/coreclr/jit/codegencommon.cpp#L4616)...

So I agree we should leave this as-is for this PR (maybe rephrase the comment though), and look at it separately. Likely we'll need to insert explicit zero-initalization (and copy (an approximation of) the logic above) after SSA (because SSA already accounts for these rules and does some optimizations based on them). Then for true uninitialized locals we can use something like LLVM's `freeze undef`.

src/coreclr/jit/morph.cpp
yowl: I think this is going to be easy so I'll just try and do it, removing these `#ifs`

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+        // create temp for the return address
+        unsigned   returnTempNum    = _compiler->lvaGrabTemp(false DEBUGARG("return value address"));
+        LclVarDsc* returnAddrVarDsc = _compiler->lvaGetDesc(returnTempNum);
+        returnAddrVarDsc->lvType    = TYP_I_IMPL;
+
+        GenTree*          addrStore     = _compiler->gtNewStoreLclVar(returnTempNum, returnValueAddress);
+        GenTree*          returnAddrLcl = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
+        GenTreePutArgType* putArg = _compiler->gtNewOperNode(GT_PUTARG_TYPE, TYP_I_IMPL, returnAddrLcl)->AsPutArgType();
+
+        putArg->gtCorInfoType = CORINFO_TYPE_PTR;
```
SingleAccretion: After `GenTreePutArgType` has a constructor, this should use it. Or, more likely, a helper, say `Llvm::newPutArgType()`, to avoid the boilerplate of the verbose `new` operator.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+        lastArg               = _compiler->gtInsertNewCallArgAfter(putArg, insertAfterArg);
+
+        GenTree* returnAddrLclAfterCall = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
+        GenTree* indirNode;
+        if (callReturnType == TYP_STRUCT)
+        {
+            putArg->gtClsHnd                = calleeSigInfo.retTypeClass;
+            indirNode                       = new (_compiler, GT_OBJ) GenTreeObj(callReturnType, returnAddrLclAfterCall,
+                                                           _compiler->typGetObjLayout(calleeSigInfo.retTypeClass));
+            indirNode->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
```
SingleAccretion: Any reason not to use `gtNewObjNode`?

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+        {
+            callUse.ReplaceWith(_compiler, indirNode);
+        }
+        else
+        {
+            callNode->ClearUnusedValue();
+        }
+
+        callNode->gtReturnType = TYP_VOID;
+        callNode->ChangeType(TYP_VOID);
```
SingleAccretion: Should `gtCallIrReturnType` stay the same as it was (presumably not)?

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+        GenTreePutArgType* putArg = _compiler->gtNewOperNode(GT_PUTARG_TYPE, TYP_I_IMPL, returnAddrLcl)->AsPutArgType();
+
+        putArg->gtCorInfoType = CORINFO_TYPE_PTR;
+        lastArg               = _compiler->gtInsertNewCallArgAfter(putArg, insertAfterArg);
+
+        GenTree* returnAddrLclAfterCall = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
+        GenTree* indirNode;
+        if (callReturnType == TYP_STRUCT)
+        {
+            putArg->gtClsHnd                = calleeSigInfo.retTypeClass;
```
SingleAccretion: Hmm, this is setting a handle for the address arg, which presumably should not have any since it is a simple `TYP_I_IMPL` node.

src/coreclr/jit/llvm.cpp
SingleAccretion: So the comment should be something like "not handling VTable-based virtuals or VSD, yet", I suppose.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+}
+
+GenTree* Llvm::createStoreNode(var_types nodeType, GenTree* addr, GenTree* nodeToStore, ClassLayout* structClassLayout)
+{
+    GenTree* storeNode;
+    if (nodeType == TYP_STRUCT)
+    {
+        storeNode = new (_compiler, GT_STORE_OBJ)
+            GenTreeObj(nodeType, addr, nodeToStore, structClassLayout);
+        storeNode->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
```
```suggestion
```
The constructor initializes it as such.

src/coreclr/jit/llvm.cpp
yowl: Thanks, have updated the comment with the full dump.  Call args appear as `wasm arg` as there's a minor problem with dumping the tree without the arg info.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+    }
+    else
+    {
+        storeNode = _compiler->gtNewOperNode(GT_STOREIND, nodeType, addr, nodeToStore);
```
SingleAccretion: I think it would be better to use the `GenTreeStoreInd` constructor here, though it technically does not matter a lot.

src/coreclr/jit/llvm.cpp
yowl: Thanks VSD - Virtual ... Dispatch ?

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+            thisArgNum++;
+            continue;
+        }
+
+        CORINFO_CLASS_HANDLE clsHnd      = NO_CLASS_HANDLE;
+        CorInfoType          corInfoType = CORINFO_TYPE_UNDEF;
+
+        // "this" not in sigInfo arg list - for compatbility with the IL->LLVM backend, this is always on the
+        // shadowstack A performance benefit could be to move it to the LLVM stack when it has no GC ptrs, but its not
+        // clear how to get the class handle if it is a struct.
```
SingleAccretion: I believe it is the case that `this` will always have to be on the shadow stack since it is either a `byref` or a `ref`, and in general we do not know where those come from (a `byref` to a struct that has no GC pointers itself can still be rooting a heap object of which the struct in question can be a field).

The only optimization legal here is to not GC-report things if we know they're not on the heap (say `this` is actually a local). It will be an important optimization for us, but it applies to all `byref`s, not just `this`, equally.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+                // from the IL->LLVM backed:
+                // The previous argument might have left this type unaligned, so pad if necessary
+                // shadowStackUseOffest = PadOffset(argType, shadowStackUseOffest);
+            }
+
+            GenTreeIntCon* offset   = _compiler->gtNewIconNode(shadowStackUseOffest, TYP_I_IMPL);
+            GenTree*       slotAddr = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, lclShadowStack, offset);
+            GenTree*       storeNode =
+                createStoreNode(opAndArg.operand->TypeGet(), slotAddr, opAndArg.operand,
+                                corInfoType == CORINFO_TYPE_VALUECLASS ? _compiler->typGetObjLayout(clsHnd) : nullptr);
```
SingleAccretion, line 2025:
```suggestion
                                corInfoType == CORINFO_TYPE_VALUECLASS ? _compiler->typGetObjLayout(clsHnd) : NO_CLASS_HANDLE);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+            }
+
+            CurrentRange().InsertBefore(callNode, lclShadowStack, offset, slotAddr, storeNode);
+        }
+        else
+        {
+            // arg on LLVM stack
+            // unsigned           putArgTempNum = _compiler->lvaGrabTemp(false DEBUGARG("putarg temp"));
+            // LclVarDsc*         argTempVarDsc = _compiler->lvaGetDesc(putArgTempNum);
+            // GenTree* argTemp = _compiler->gtNewLclvNode(putArgTempNum, opAndArg.operand->TypeGet());
```
SingleAccretion: Not sure what the commented code is for...?

src/coreclr/jit/llvm.cpp
```diff
@@ -2011,18 +2169,8 @@ void Llvm::ConvertShadowStackLocals()
...
-                        GenTreeObj(originalReturnType, retAddressLocal, node->AsOp()->gtGetOp1(), _compiler->typGetObjLayout(_sigInfo.retTypeClass));
-                    storeNode->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
-                }
-                else
-                {
-                    storeNode = _compiler->gtNewOperNode(GT_STOREIND, originalReturnType, retAddressLocal, node->AsOp()->gtOp1);
-                }
-                storeNode->gtFlags |= GTF_IND_TGT_NOT_HEAP; // No RhpAssignRef required
+                GenTree* storeNode = createStoreNode(originalReturnType, retAddressLocal, node->AsOp()->gtGetOp1(),
+                                    originalReturnType == TYP_STRUCT ? _compiler->typGetObjLayout(_sigInfo.retTypeClass) : nullptr);
```
SingleAccretion, line 2099:
```suggestion
                                    originalReturnType == TYP_STRUCT ? _compiler->typGetObjLayout(_sigInfo.retTypeClass) : NO_CLASS_HANDLE);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+    callNode->gtCallThisArg = nullptr;
+    argCount                = argInfo->ArgCount();
+    argTable                = argInfo->ArgTable();
+
+    // set up the callee shadowstack, creating a temp and the PUTARG
+    GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
+    GenTreeIntCon* offset         = _compiler->gtNewIconNode(_shadowStackLocalsSize,
+                                                     TYP_I_IMPL); // TODO-LLVM: possible performance benefit: when
+                                                                  // _shadowStackLocalsSize == 0, then omit the GT_ADD
+    GenTree* calleeShadowStack = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
```
```suggestion
    GenTreeIntCon* offset         = _compiler->gtNewIconNode(_shadowStackLocalsSize, TYP_I_IMPL);
    // TODO-LLVM: possible performance benefit: when _shadowStackLocalsSize == 0, then omit the GT_ADD.
    GenTree* calleeShadowStack = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+
+    // store in the temp
+    unsigned calleeShadowStackTemp = _compiler->lvaGrabTemp(
+        false DEBUGARG("callee shadowstack")); // used to create the store locations for args on shadow stack
```
SingleAccretion: Do we need this temp? It seem like we could use the `_shadowStackLclNum` as a base, just adding offsets to it with `_shadowStackLocalsSize + offset`.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+
+            CurrentRange().InsertBefore(callNode, putArg);
+        }
+        if (!isThis)
+        {
+            sigArgs = _info.compCompHnd->getArgNext(sigArgs);
+        }
+    }
+
+    _compiler->fgInitArgInfo(callNode);
```
SingleAccretion: Ditto: we should drop the arg info after lowering.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+        storeNode->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
+    }
+    else
+    {
+        storeNode = _compiler->gtNewOperNode(GT_STOREIND, nodeType, addr, nodeToStore);
+    }
+    return storeNode;
+}
+
+void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calleeSigInfo)
```
SingleAccretion, line 1950: We should have a comment above this method outlining our lowering strategy, something like:
```
// lowerCallToShadowStack: Lower the call, rewriting its arguments.
//
// This method has two primary objectives:
//  1) Transfer the information about the arguments from arg info to explicit
//     PutArgType nodes, to make it easy for codegen to consume it. Also, get
//     rid of the late/non-late argument distinction, by sorting the inserted nodes
//     in the original evaluation order, matching that of them in the signature.
//  2) Rewrite arguments and the return to be stored on the shadow stack. We take
//     the arguments which need to be on the shadow stack, remove them from the call
//     arguments list, store their values on the shadow stack, at offsets calculated
//     in a simple increasing order, matching the signature. We also rewrite returns
//     that must be on the shadow stack, see "lowerCallReturn".
//
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+    unsigned thisArgNum           = 0;
+    unsigned shadowStackUseOffest = 0;
+
+    fgArgInfo*                 argInfo    = callNode->fgArgInfo;
+    unsigned int               argCount   = argInfo->ArgCount();
+    fgArgTabEntry**            argTable   = argInfo->ArgTable();
+    std::vector<OperandArgNum> sortedArgs = std::vector<OperandArgNum>(argCount);
+    OperandArgNum*             sortedData = sortedArgs.data();
+
+    GenTreeCall::Use* oldArgs = callNode->gtCallLateArgs;
```
```suggestion
```
Unused.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1857,277 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+
+    CORINFO_ARG_LIST_HANDLE sigArgs = calleeSigInfo.args;
+
+    for (OperandArgNum opAndArg : sortedArgs)
+    {
+        if (opAndArg.operand->IsArgPlaceHolderNode())
+        {
+            thisArgNum++;
+            continue;
+        }
```
SingleAccretion: Hmm, not sure I understand this logic: `opAndArg.operand` should never be a placeholder (since it is obtained from `curArgTabEntry->GetNode()`).

src/coreclr/jit/llvm.cpp
yowl: I was going to put that in `gentree.cpp`: 
```
GenTreePutArgType* Compiler::gtNewPutArgType(var_types type DEBUGARG(IL_OFFSETX ILoffs), GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
{
    GenTreePutArgType* putArg = new (this, GT_PUTARG_TYPE) GenTreePutArgType(type, op, corInfoType, clsHnd);
    return putArg;
}
```
Is that ok?

yowl: I'll Remove `putArg` of course

yowl:
I've tidied a few things up and addressed some of the feedback, looks a lot better, thanks!

src/coreclr/jit/llvm.cpp
yowl: Its not needed, I'll delete it.

src/coreclr/jit/llvm.cpp
SingleAccretion: [Virtual Stub Dispatch](https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/botr/virtual-stub-dispatch.md).

src/coreclr/jit/llvm.cpp
yowl: Thanks, looks good.

src/coreclr/jit/llvm.cpp
yowl: yep, gone.

src/coreclr/jit/llvm.cpp
SingleAccretion: > Is that ok?

Sounds good to me (but do we need an IL offset?).

src/coreclr/jit/gentree.h
yowl: Thanks, done

src/coreclr/jit/llvm.cpp
yowl: I think putting the `CorInfoType` on the `GenTreeCall` node has eliminated the need for this.  I'll delete it.

src/coreclr/jit/morph.cpp
yowl: `morph.cpp` no longer appears in the list of changed files.

src/coreclr/jit/llvm.cpp
SingleAccretion: Thank you, looks like an initialized local case to me indeed. One can reproduce it with something like:
```cs
int Write(bool a)
{
    Unsafe.SkipInit(out int b);
    if (a) b = 1;

    return b;
}
```
(A bit curious how Roslyn produces such locals for pattern matching. It can pessimize register allocation and range check elimination)

src/coreclr/jit/llvm.cpp
yowl: I don't know....   Don't seem to.  I've deleted that Ctrl-C/V

src/coreclr/jit/gentree.h
```diff
@@ -6687,6 +6691,31 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
...
+    CorInfoType          m_CorInfoType;
+    CORINFO_CLASS_HANDLE m_ClsHnd;
+
+public:
+    GenTreePutArgType(var_types type, GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
+        : GenTreeOp(GT_PUTARG_TYPE, type, op, nullptr), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
+    {
+    }
+
+    CorInfoType GetCorInfoType()
```
SingleAccretion:
```suggestion
    CorInfoType GetCorInfoType() const
```

src/coreclr/jit/gentree.h
```diff
@@ -6687,6 +6691,31 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
...
+    {
+        return m_CorInfoType;
+    }
+
+    CORINFO_CLASS_HANDLE GetClsHnd()
```
SingleAccretion:
```suggestion
    CORINFO_CLASS_HANDLE GetClsHnd() const
```

src/coreclr/jit/gentree.h
```diff
@@ -6687,6 +6691,31 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
 };
 #endif // FEATURE_ARG_SPLIT
 
+#if defined(TARGET_WASM)
+struct GenTreePutArgType : public GenTreeOp
```
SingleAccretion:
```suggestion
struct GenTreePutArgType : public GenTreeUnOp
```

src/coreclr/jit/compiler.cpp
```diff
@@ -752,6 +752,11 @@ var_types Compiler::getArgTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
                                         bool                 isVarArg,
                                         unsigned             structSize)
 {
+// For LLVM/Wasm, always pass by value and let the lowering decide
+#if defined(TARGET_WASM)
+    *wbPassStruct = SPK_ByValue;
+    return TYP_STRUCT;
+#else
```
```suggestion
#else // !TARGET_WASM
```

src/coreclr/jit/compiler.cpp
```diff
@@ -1046,6 +1052,10 @@ var_types Compiler::getReturnTypeForStruct(CORINFO_CLASS_HANDLE     clsHnd,
                                            structPassingKind*       wbReturnStruct /* = nullptr */,
                                            unsigned                 structSize /* = 0 */)
 {
+#if defined(TARGET_WASM)
+    *wbReturnStruct = SPK_ByValue;
+    return TYP_STRUCT;
+#else // TARGET_WASM
```
```suggestion
#else // !TARGET_WASM
```

src/coreclr/jit/gentree.cpp
```diff
@@ -13078,6 +13089,9 @@ void Compiler::gtDispLIRNode(GenTree* node, const char* prefixMsg /* = nullptr *
             }
             else
             {
+    #ifdef TARGET_WASM
+                displayOperand(operand, "wasm arg" /* TODO-LLVM */, operandArc, indentStack, prefixIndent);
```
SingleAccretion: We can use info from the put arg here for display. Say stash an argument number there under `#ifdef DEBUG`, etc.

src/coreclr/jit/gentree.h
```diff
@@ -6687,6 +6691,31 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
...
+{
+private:
+    CorInfoType          m_CorInfoType;
+    CORINFO_CLASS_HANDLE m_ClsHnd;
+
+public:
+    GenTreePutArgType(var_types type, GenTree* op, CorInfoType corInfoType, CORINFO_CLASS_HANDLE clsHnd)
+        : GenTreeOp(GT_PUTARG_TYPE, type, op, nullptr), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
```
SingleAccretion:
```suggestion
        : GenTreeUnOp(GT_PUTARG_TYPE, type, op), m_CorInfoType(corInfoType), m_ClsHnd(clsHnd)
```
(Together with the inheritance change)

src/coreclr/jit/llvm.cpp
yowl: You are right, I've reset it to void

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1821,268 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+            {
+                shadowStackUseOffest += TARGET_POINTER_SIZE;
+            }
+
+            CurrentRange().InsertBefore(callNode, lclShadowStack, offset, slotAddr, storeNode);
+        }
+        else
+        {
+            // arg on LLVM stack
+            GenTreePutArgType* putArg = _compiler->gtNewPutArgType(opAndArg.operand->TypeGet(), opAndArg.operand, corInfoType, clsHnd);
```
SingleAccretion: So in case we "swizzled" (passing `struct { int }` as `int`, etc) operands, this will make a `PutArgType` with the "swizzled" type.

I think it would be better to have `PutArgType`s constructor take only the `corInfoType` and deduce the node type with `JITtype2varType`.

It doesn't matter a lot, but should make for less confusing dumps.

src/coreclr/jit/llvm.cpp
yowl: Right, I'll delete it.

src/coreclr/jit/llvm.h
```diff
@@ -146,16 +146,18 @@ class Llvm
     Value* castToPointerToLlvmType(Value* address, llvm::Type* llvmType);
     Value* consumeValue(GenTree* node, llvm::Type* targetLlvmType);
     llvm::DILocation* createDebugFunctionAndDiLocation(struct DebugMetadata debugMetadata, unsigned int lineNo);
+    GenTree* createStoreNode(var_types nodeType, GenTree* addr, GenTree* nodeToStore, ClassLayout* structClassLayout = nullptr);
```
SingleAccretion: Nit: the canonical name for `nodeToStore` is `data` (as in `AsBlk()->Data()`).

src/coreclr/jit/gentree.h
```diff
@@ -4743,6 +4743,10 @@ struct GenTreeCall final : public GenTree
     unsigned char gtCallType : 3;   // value from the gtCallTypes enumeration
     unsigned char gtReturnType : 5; // exact return type
 
+#if defined(TARGET_WASM)
+    CorInfoType gtCorInfoType; // original type from CORINFO_SIG_INFO used to construct signature
```
SingleAccretion: I think a more precise description for this would be something like `the precise return type used to construct the signature`, since it will not always match what was in the IL signature.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1821,267 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+
+        // create temp for the return address
+        unsigned   returnTempNum    = _compiler->lvaGrabTemp(false DEBUGARG("return value address"));
+        LclVarDsc* returnAddrVarDsc = _compiler->lvaGetDesc(returnTempNum);
+        returnAddrVarDsc->lvType    = TYP_I_IMPL;
+
+        GenTree*          addrStore     = _compiler->gtNewStoreLclVar(returnTempNum, returnValueAddress);
+        GenTree*          returnAddrLcl = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
+
+        CORINFO_CLASS_HANDLE putArgClsHnd = nullptr; // i8* for everything other than structs
```
```suggestion
        CORINFO_CLASS_HANDLE putArgClsHnd = NO_CLASS_HANDLE; // i8* for everything other than structs
```

src/coreclr/jit/llvm.cpp
yowl: Thanks, have added a TODO-LLVM, didn't really know where a good place for that is so just put it above `Compile`

src/coreclr/jit/llvm.cpp
yowl: Thanks, deleted

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1802,252 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+        if (argOnShadowStack)
+        {
+            GenTree* lclShadowStack = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
+
+            if (corInfoType == CORINFO_TYPE_VALUECLASS)
+            {
+                shadowStackUseOffest = padOffset(corInfoType, clsHnd, shadowStackUseOffest);
+            }
+
+            GenTreeIntCon* offset    = _compiler->gtNewIconNode(shadowStackUseOffest + _shadowStackLocalsSize, TYP_I_IMPL);
```
```suggestion
            GenTreeIntCon* offset    = _compiler->gtNewIconNode(_shadowStackLocalsSize + shadowStackUseOffest, TYP_I_IMPL);
```
Since that's logically how it is.

src/coreclr/jit/llvm.cpp
SingleAccretion: (There was also similar suggestion above:
```
corInfoType == CORINFO_TYPE_VALUECLASS ? _compiler->typGetObjLayout(clsHnd) : nullptr);
```
seems it got lost in a merge)

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1802,252 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+        for (GenTree* operand : callNode->Operands())
+        {
+            if (operand->IsArgPlaceHolderNode() || !operand->IsValue())
+            {
+                // Either of these situations may happen with calls.
+                continue;
+            }
+            fgArgTabEntry* curArgTabEntry = _compiler->gtArgEntryByNode(callNode, operand);
+            regNumber      argReg         = curArgTabEntry->GetRegNum();
+            if (argReg == REG_STK || curArgTabEntry->argType == TYP_BYREF) // TODO-LLVM: out and ref args
```
SingleAccretion, line 1914: I wonder if things will "just work" if you remove the `argReg == REG_STK` restriction.

src/coreclr/jit/gentree.h
```diff
@@ -6687,6 +6691,31 @@ struct GenTreePutArgSplit : public GenTreePutArgStk
 };
 #endif // FEATURE_ARG_SPLIT
 
+#if defined(TARGET_WASM)
+struct GenTreePutArgType : public GenTreeOp
+{
+private:
+    CorInfoType          m_CorInfoType;
+    CORINFO_CLASS_HANDLE m_ClsHnd;
```
SingleAccretion:
```suggestion
    CORINFO_CLASS_HANDLE m_ClsHnd; // The struct handle if this is a struct, the pointee type if this is a pointer.
```
To document we treat pointers somewhat specially here.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1802,252 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+
+        GenTree*          addrStore     = _compiler->gtNewStoreLclVar(returnTempNum, returnValueAddress);
+        GenTree*          returnAddrLcl = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
+
+        GenTree* returnAddrLclAfterCall = _compiler->gtNewLclvNode(returnTempNum, TYP_I_IMPL);
+        GenTree* indirNode;
+        if (callReturnType == TYP_STRUCT)
+        {
+            indirNode    = _compiler->gtNewObjNode(calleeSigInfo.retTypeClass, returnAddrLclAfterCall);
+            indirNode->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
```
```suggestion
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1802,252 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+
+        CurrentRange().InsertBefore(callNode, shadowStackVar, offset, returnValueAddress, addrStore);
+        CurrentRange().InsertAfter(addrStore, returnAddrLcl, putArg);
+        CurrentRange().InsertAfter(callNode, returnAddrLclAfterCall, indirNode);
+    }
+    else
+    {
+        callNode->gtCorInfoType = calleeSigInfo.retType;
+    }
+    return lastArg;
```
```suggestion
    }

    return lastArg;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1802,252 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+    OperandArgNum*             sortedData = sortedArgs.data();
+
+    GenTreeCall::Use* lastArg;
+    GenTreeCall::Use* insertReturnAfter;
+    GenTreeCall::Use* callThisArg = callNode->gtCallThisArg;
+
+    callNode->ResetArgInfo();
+    callNode->gtCallThisArg = nullptr;
+    argCount                = argInfo->ArgCount();
+    argTable                = argInfo->ArgTable();
```
SingleAccretion: This appears redundant with the initialization done at declaration.

src/coreclr/jit/llvm.cpp
```diff
@@ -1896,7 +1802,252 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+
+    // set up the callee shadowstack, creating a temp and the PUTARG
+    GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
+    GenTreeIntCon* offset         = _compiler->gtNewIconNode(_shadowStackLocalsSize,
+                                                     TYP_I_IMPL); // TODO-LLVM: possible performance benefit: when
+                                                                  // _shadowStackLocalsSize == 0, then omit the GT_ADD
+    GenTree* calleeShadowStack = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackVar, offset);
+
+    GenTreePutArgType* putArg = _compiler->gtNewPutArgType(TYP_I_IMPL, calleeShadowStack, CORINFO_TYPE_PTR, NO_CLASS_HANDLE);
```
SingleAccretion: Nit: let's rename `putArg` as something like `calleeShadowStackPutArg`, to not have redeclared locals (there is another `GenTreePutArgType* putArg` below).

src/coreclr/jit/gentree.cpp
yowl: I think the problem with that might be the node changes from small to large.  Let me try

src/coreclr/jit/llvm.cpp

src/coreclr/jit/llvm.cpp
yowl: Yep, ok

yowl:
This is actuallly green I think https://dev.azure.com/dnceng/public/_build/results?buildId=1487844&view=results  Probably the github outage killed the links.

SingleAccretion (approved):
Look great!

jkotas:
Nice!

PR #1746, opened 2021-11-23
Sync cgroup.cpp with dotnet/runtime

MichalStrehovsky:
We diverged here. Asking the CI if this was intentional.

### DIFF
--- a/src/coreclr/gc/unix/cgroup.cpp
+++ b/src/coreclr/gc/unix/cgroup.cpp
@@ -8,14 +8,13 @@ Module Name:
     cgroup.cpp
 
 Abstract:
-    Read memory and cpu limits for the current process
+    Read the memory limit for the current process
 --*/
 #ifdef __FreeBSD__
 #define _WITH_GETLINE
 #endif
 
 #include <cstdint>
-#include <cstddef>
 #include <cassert>
 #include <unistd.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


jkotas:
CoreCLR switched to reading the CPU limits in the thick CoreCLR PAL. The code in GC PAL was deleted as part of that change since it is no longer used for full CoreCLR.

The best fix would be to deduplicate the code to read the CPU limits and move it under src\native.

MichalStrehovsky:
I was afraid of that

MichalStrehovsky:
So I looked at is a bit. There's a lot of duplicated code for cgroup parsing because of the Unix philosophy of "why would you introduce proper strongly typed APIs if everyone can parse things out of text files instead because parsers are fun?".

GC still has its fork of the cgroup file format reader to read memory limits, and PAL has another fork of it to read CPU and memory limits. We would either:

1. Create yet another fork of the common parts of cgroup file format reading under src/native just to read CPU limits (and leave memory limits parsing duplicated the same way they are, but now we have 3 parsers of the cgroup format), or
2. Also move out memory limits parsing to src/native and have everyone use that.

PAL_GetCpuLimit is also used within the PAL itself to implement PAL. Somehow that makes me feel like we need to double the estimate of how much work this is going to be.

Restoring the deleted code under FEATURE_REDHAWK sounds pretty appealing at the moment to be honest.

PR #1754, opened 2021-11-26
Hide constructed/necessary type duality from RyuJIT

### LINKED ISSUE #1128: Fix VNs for method table fetched from newobj
NativeAOT compiler creates different type symbols for constructed (e.g. `new MyType`) and other uses (e.g. `typeof(MyType)`). It allows it to omit full vtable when the type is never constructed e.g. when the type is only used as `typeof(MyType)`. These symbols are folded together near final linking phase.

Optimization introduced in https://github.com/dotnet/runtime/pull/52476/ will make assertion prop compare the symbols, assuming that two different symbols will have two different addresses. It is not the case for constructed and unconstructed types since they are folded together later.

MichalStrehovsky:
If we have whole program view, use that to upgrade requests for "necessary" types to "constructed" types if we know a constructed was seen during scanning. This avoids RyuJIT seeing a constructed and necessary symbol for the same type.

If we don't have whole program view, forgo the optimization and use maximally loadable types instead of necessary types.

Fixes #1128.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs
@@ -405,14 +405,9 @@ namespace ILCompiler
                             //    didn't scan the virtual methods on them.
                             //
 
+                            _constructedTypes.Add(type);
+
                             TypeDesc canonType = type.ConvertToCanonForm(CanonicalFormKind.Specific);
-
-                            _constructedTypes.Add(canonType);
-
-                            // Since this is used for the purposes of devirtualization, it's really convenient
-                            // to also have Array<T> for each T[].
-                            if (canonType.IsArray)
-                                _constructedTypes.Add(canonType.GetClosestDefType());
 
                             bool hasNonAbstractTypeInHierarchy = canonType is not MetadataType mdType || !mdType.IsAbstract;
                             TypeDesc baseType = canonType.BaseType;
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/RyuJitCompilation.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/RyuJitCompilation.cs
@@ -72,9 +72,9 @@ namespace ILCompiler
             // information proving that it isn't, give RyuJIT the constructed symbol even
             // though we just need the unconstructed one.
             // https://github.com/dotnet/runtimelab/issues/1128
-            bool canConstructPerWholeProgramAnalysis = _devirtualizationManager == null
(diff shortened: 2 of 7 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs
```diff
@@ -405,14 +405,9 @@ public ScannedDevirtualizationManager(ImmutableArray<DependencyNodeCore<NodeFact
                             //    didn't scan the virtual methods on them.
                             //
 
-                            TypeDesc canonType = type.ConvertToCanonForm(CanonicalFormKind.Specific);
-
-                            _constructedTypes.Add(canonType);
+                            _constructedTypes.Add(type);
 
-                            // Since this is used for the purposes of devirtualization, it's really convenient
```
jkotas, line 412: Is the `Array<T>` case no longer needed; or are we giving up on de-virtualization for arrays?

src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/RyuJitCompilation.cs
```diff
@@ -63,6 +63,23 @@ public sealed class RyuJitCompilation : Compilation
...
+            // RyuJIT makes assumptions around the value of these symbols - in particular, it assumes
+            // that type handles and type symbols have a 1:1 relationship. We therefore need to
+            // make sure RyuJIT never sees a constructed and unconstructed type symbol for the
+            // same type. If the type is constructable and we don't have whole progam view
+            // information proving that it isn't, give RyuJIT the constructed symbol even
+            // though we just need the unconstructed one.
+            // https://github.com/dotnet/runtimelab/issues/1128
+            bool canConstructPerWholeProgramAnalysis = _devirtualizationManager == null
+                ? true : _devirtualizationManager.CanConstructType(type);
+            if (canConstructPerWholeProgramAnalysis)
```
jkotas:
```suggestion
            bool canPotentiallyConstruct = _devirtualizationManager == null
                ? true : _devirtualizationManager.CanConstructType(type);
            if (canPotentiallyConstruct)
```
`canConstructPerWholeProgramAnalysis` does not capture the meaning for the `_devirtualizationManager == null` case.

jkotas (approved):
Thank you!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs
MichalStrehovsky: `_constructedTypes` is no longer used for devirtualization. The comment was outdated.

PR #1758, opened 2021-11-29
[NativeAOT-LLVM] Initial support for OperIsLocalAddr

yowl:
This PR adds support for `GT_LCL_VAR_ADDR` 

For locals on the shadow stack, this is easy.  For other locals it saves them to LLVM `alloca` addresses.  

@SingleAccretion - I've not done parameters yet, as I wanted some feedback on the approach.  Currently for nongc locals, stored in address space allocated with LLVM `alloca` this is done in by adding logic to `storeLocalVar` and `localVar`.  I did think that I could rewrite the tree if I created a new GT_NODE to represent the `alloca`, and then replace `GT_LCL_VAR/GT_STORE_LCL_VAR` wtih `IND/STORE_IND`  so I wanted to ask if you think that would be better?

(I did also think to just put all locals where I need the address on the shadow stack, reusing a lot of the existing lowering code, but I didn't like that in the end because: 1) it creates extra work for the conservative GC scanning, and 2) for non-wasm targets, there may be some advantage in using `alloca` over explicitly using addresses on the shadow stack)

### DIFF
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -127,7 +127,7 @@ unsigned ReinterpretHexAsDecimal(unsigned in);
 
 const unsigned FLG_CCTOR = (CORINFO_FLG_CONSTRUCTOR | CORINFO_FLG_STATIC);
 
-#ifdef DEBUG
+#if defined(DEBUG) || defined(TARGET_WASM)
 const int BAD_STK_OFFS = 0xBAADF00D; // for LclVarDsc::lvStkOffs
 #endif
 
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -127,7 +127,7 @@ unsigned ReinterpretHexAsDecimal(unsigned in);
 
 const unsigned FLG_CCTOR = (CORINFO_FLG_CONSTRUCTOR | CORINFO_FLG_STATIC);
 
-#if defined(DEBUG) || defined(TARGET_WASM)
+#ifdef DEBUG
 const int BAD_STK_OFFS = 0xBAADF00D; // for LclVarDsc::lvStkOffs
 #endif
 
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -380,7 +380,8 @@ public:
         :
 #if defined(TARGET_WASM)
         lvLlvmArgNum(BAD_LLVM_ARG_NUM),
-        lvCorInfoType(CORINFO_TYPE_UNDEF)
+        lvCorInfoType(CORINFO_TYPE_UNDEF),
+        lvHasLocalAddr(0)
         ,
 #endif // TARGET_WASM
         _lvArgReg(REG_STK)
@@ -576,6 +577,7 @@ public:
 #if defined(TARGET_WASM)
     unsigned int lvLlvmArgNum;
     CorInfoType  lvCorInfoType;
+    unsigned char lvHasLocalAddr : 1;
 #endif
 
 #ifdef DEBUG
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -189,9 +189,9 @@ StructDesc* Llvm::getStructDesc(CORINFO_CLASS_HANDLE structHandle)
             assert(fldOffset < structSize);
 
             CORINFO_CLASS_HANDLE fieldClass;
-            _info.compCompHnd->getFieldType(fieldHandle, &fieldClass);
+            CorInfoType corInfoType = _info.compCompHnd->getFieldType(fieldHandle, &fieldClass);
 
-            unsigned fieldSize = _info.compCompHnd->getClassSize(fieldClass);
+            unsigned fieldSize = getElementSize(fieldClass, corInfoType);
 
             // store the biggest field at the offset for unions
             if (sparseFields[fldOffset] == nullptr || fieldSize > sparseFieldSizes[fldOffset])
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -2451,6 +2451,11 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     lowerToShadowStack();
 }
 
+static bool placeLocalInAlloca(LclVarDsc* varDsc)
+{
+    return !varDsc->lvTracked || varDsc->lvHasLocalAddr;
+}
+
 void Llvm::createAllocasForLocalsWithAddrOp()
 {
     m_allocas = new std::vector<Value*>(_compiler->lvaCount, nullptr);
@@ -2458,14 +2463,15 @@ void Llvm::createAllocasForLocalsWithAddrOp()
     for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
     {
         LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
-        if (varDsc->lvIsParam && varDsc->lvHasLocalAddr)
+        if (varDsc->lvIsParam && placeLocalInAlloca(varDsc))
         {
             // TODO-LLVM : copy param to alloca storage
             // see S_P_CoreLib_System_Diagnostics_Tracing_EventSource__InitializeIsSupported
             failFunctionCompilation();
         }
 
-        if (canStoreLocalOnLlvmStack(varDsc) && varDsc->lvHasLocalAddr)
+        // TODO-LLVM LCLBLK - outgoing arg space
+        if (varDsc->lvType != TYP_LCLBLK && canStoreLocalOnLlvmStack(varDsc) && placeLocalInAlloca(varDsc))
         {
             CORINFO_CLASS_HANDLE classHandle = NO_CLASS_HANDLE;
             if (varDsc->lvType == TYP_STRUCT)
@@ -2479,6 +2485,26 @@ void Llvm::createAllocasForLocalsWithAddrOp()
                 llvmType = Type::getInt32Ty(_llvmContext);
             }
             (*m_allocas)[lclNum] = _prologBuilder.CreateAlloca(llvmType);
+        }
+    }
+}
+
+void Llvm::DontTrackLocalsWithAddress()
+{
+    for (BasicBlock* _currentBlock : _compiler->Blocks())
+    {
+        _currentRange = &LIR::AsRange(_currentBlock);
+        for (GenTree* node : CurrentRange())
+        {
+            if (node->OperIsLocalAddr())
+            {
+                LclVarDsc* localVarDsc = _compiler->lvaGetDesc(node->AsLclVarCommon()->GetLclNum());
+                if (localVarDsc->lvTracked)
+                {
+                    assert(false);// TODO-LLVM this is never hit?
+                    // localVarDsc->lvTracked = 0;
+                }
+            }
         }
     }
 }
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1629,10 +1629,18 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
     return llvmRef;
 }
 
-void Llvm::buildLocalVarAddr(GenTreeLclVar* lclAddr)
+void Llvm::buildLocalVarAddr(GenTreeLclVarCommon* lclAddr)
 {
     unsigned int lclNum = lclAddr->GetLclNum();
-    mapGenTreeToValue(lclAddr, m_allocas[lclNum]);
+    if (lclAddr->isLclField())
+    {
+        GenTreeLclFld* lclFldAddr = lclAddr->AsLclFld();
+        mapGenTreeToValue(lclAddr, _builder.CreateGEP(m_allocas[lclNum], _builder.getInt16(lclFldAddr->GetLclOffs())));
+    }
+    else
+    {
+        mapGenTreeToValue(lclAddr, m_allocas[lclNum]);
+    }
 }
 
 // LLVM operations like ICmpNE return an i1, but in the IR it is expected to be an Int (i32).
@@ -1696,8 +1704,9 @@ Value* Llvm::getLocalVarAddress(GenTreeLclVar* lclVar) {
     return _builder.CreateGEP(_function->getArg(0), _builder.getInt32(varOffset), "lclVar");
 }
 
-static bool isLlvmFrameLocal(LclVarDsc* varDsc)
+bool Llvm::isLlvmFrameLocal(LclVarDsc* varDsc)
 {
+    assert(canStoreLocalOnLlvmStack(varDsc) && (_compiler->fgSsaPassesCompleted >= 1));
     return !varDsc->lvInSsa;
 }
 
@@ -1770,7 +1779,8 @@ void Llvm::visitNode(GenTree* node)
             localVar(node->AsLclVar());
             break;
         case GT_LCL_VAR_ADDR:
-            buildLocalVarAddr(node->AsLclVar());
+        case GT_LCL_FLD_ADDR:
+            buildLocalVarAddr(node->AsLclVarCommon());
             break;
         case GT_LSH:
         case GT_RSH:
@@ -2373,10 +2383,6 @@ void Llvm::lowerToShadowStack()
         _currentRange = &LIR::AsRange(_currentBlock);
         for (GenTree* node : CurrentRange())
         {
-            if (node->OperIs(GT_LCL_FLD_ADDR))
-            {
-                unsigned i = 0;
-            }
             if (node->OperIs(GT_STORE_LCL_VAR, GT_LCL_VAR, GT_LCL_VAR_ADDR))
             {
                 ConvertShadowStackLocalNode(node->AsLclVarCommon());
@@ -2481,7 +2487,6 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     lowerToShadowStack();
 }
 
-    assert(canStoreLocalOnLlvmStack(varDsc) && (_compiler->fgSsaPassesCompleted >= 1));
 void Llvm::createAllocasForLocalsWithAddrOp()
 {
     m_allocas = std::vector<Value*>(_compiler->lvaCount, nullptr);
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -188,7 +188,11 @@ StructDesc* Llvm::getStructDesc(CORINFO_CLASS_HANDLE structHandle)
 
             assert(fldOffset < structSize);
 
-            unsigned fieldSize = structTypeDescriptor.getFieldSize(i);
+            CORINFO_CLASS_HANDLE fieldClass;
+            _info.compCompHnd->getFieldType(fieldHandle, &fieldClass);
+
+            unsigned fieldSize = _info.compCompHnd->getClassSize(fieldClass);
+
             // store the biggest field at the offset for unions
             if (sparseFields[fldOffset] == nullptr || fieldSize > sparseFieldSizes[fldOffset])
             {
@@ -919,6 +923,16 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
             if (varTypeIsSmall(node))
             {
                 finalValue = varTypeIsSigned(node) ? _builder.CreateSExt(nodeValue, targetLlvmType)
+                                                   : _builder.CreateZExt(nodeValue, targetLlvmType);
+            }
+            else if (node->OperIs(GT_LCL_VAR))
+            {
+                // implicit upcast i8 -> i32
+                assert(nodeValue->getType() == Type::getInt8Ty(_llvmContext));
+
+                GenTreeLclVarCommon* lclNode = node->AsLclVarCommon();
+                LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNode->GetLclNum());
+                finalValue = varTypeIsSigned(varDsc->TypeGet()) ? _builder.CreateSExt(nodeValue, targetLlvmType)
                                                    : _builder.CreateZExt(nodeValue, targetLlvmType);
             }
             else
@@ -2042,7 +2056,9 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
     if (!canStoreLocalOnLlvmStack(varDsc))
     {
         // TODO-LLVM: if the offset == 0, just GT_STOREIND at the shadowStack
-        GenTreeIntCon* offset = _compiler->gtNewIconNode(varDsc->GetStackOffset(), TYP_I_IMPL);
+        unsigned offsetVal = varDsc->GetStackOffset() + node->GetLclOffs();
+        GenTreeIntCon* offset = _compiler->gtNewIconNode(offsetVal, TYP_I_IMPL);
+
         GenTreeLclVar* shadowStackLocal = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
         GenTree* lclAddress = _compiler->gtNewOperNode(GT_ADD, TYP_I_IMPL, shadowStackLocal, offset);
 
@@ -2062,7 +2078,6 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
                 break;
             case GT_LCL_FLD_ADDR:
                 indirOper = GT_NONE;
-                offset = _compiler->gtNewIconNode(varDsc->GetStackOffset() + node->AsLclFld()->GetLclOffs(), TYP_I_IMPL);
                 break;
             default:
                 unreached();
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -126,7 +126,11 @@ Value* Llvm::mapGenTreeToValue(GenTree* genTree, Value* valueRef)
 
 Value* Llvm::getGenTreeValue(GenTree* op)
 {
-    return _sdsuMap->at(op);
+    Value* mapValue = _sdsuMap->at(op);
+
+    assert(mapValue != nullptr);
+
+    return mapValue;
 }
 
 // maintains compatiblity with the IL->LLVM generation.  TODO-LLVM, when IL generation is no more, see if we can remove this unwrapping
@@ -1713,7 +1717,6 @@ bool Llvm::isLlvmFrameLocal(LclVarDsc* varDsc)
 void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
 {
     Value* localValue = getGenTreeValue(lclVar->gtGetOp1());
-    assert(localValue != nullptr);
 
     unsigned lclNum = lclVar->GetLclNum();
     Value* allocaValue = m_allocas[lclNum];
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -453,35 +453,41 @@ CORINFO_CLASS_HANDLE Llvm::tryGetStructClassHandle(LclVarDsc* varDsc)
     return varTypeIsStruct(varDsc) ? varDsc->GetStructHnd() : NO_CLASS_HANDLE;;
 }
 
+unsigned corInfoTypeAligment(CorInfoType corInfoType)
+{
+    unsigned size = TARGET_POINTER_SIZE; // TODO Wasm64 aligns pointers at 4 or 8?
+    switch (corInfoType)
+    {
+        case CORINFO_TYPE_LONG:
+        case CORINFO_TYPE_ULONG:
+        case CORINFO_TYPE_DOUBLE:
+            size = 8;
+    }
+    return size;
+}
+
 unsigned int Llvm::padOffset(CorInfoType corInfoType, CORINFO_CLASS_HANDLE structClassHandle, unsigned int atOffset)
 {
     unsigned int alignment;
-    if (corInfoType == CorInfoType::CORINFO_TYPE_BYREF || corInfoType == CorInfoType::CORINFO_TYPE_CLASS ||
-        corInfoType == CorInfoType::CORINFO_TYPE_REFANY)
+    if (corInfoType == CORINFO_TYPE_VALUECLASS)
     {
-        // simplified for just pointers
-        alignment = TARGET_POINTER_SIZE; // TODO Wasm64 aligns pointers at 4 or 8?
-    }
-    else
-    {
-        assert(corInfoType == CorInfoType::CORINFO_TYPE_VALUECLASS);
         return _padOffset(_thisPtr, structClassHandle, atOffset);
     }
+
+    alignment = corInfoTypeAligment(corInfoType);
     return roundUp(atOffset, alignment);
 }
 
 unsigned int Llvm::padNextOffset(CorInfoType corInfoType, CORINFO_CLASS_HANDLE structClassHandle, unsigned int atOffset)
 {
     unsigned int size;
-    if (corInfoType == CorInfoType::CORINFO_TYPE_BYREF || corInfoType == CorInfoType::CORINFO_TYPE_CLASS ||
-        corInfoType == CorInfoType::CORINFO_TYPE_REFANY)
+    if (corInfoType == CORINFO_TYPE_VALUECLASS)
     {
-        size = TARGET_POINTER_SIZE;
+        size = getElementSize(structClassHandle, corInfoType);
     }
     else
     {
-        assert(corInfoType == CorInfoType::CORINFO_TYPE_VALUECLASS);
-        size = getElementSize(structClassHandle, corInfoType);
+        size = corInfoTypeAligment(corInfoType);
     }
     return padOffset(corInfoType, structClassHandle, atOffset) + size;
 }
@@ -1555,7 +1561,11 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
     Value*       llvmRef;
     unsigned int lclNum = lclVar->GetLclNum();
     unsigned int ssaNum = lclVar->GetSsaNum();
-    if (_localsMap->find({lclNum, ssaNum}) == _localsMap->end())
+    if ((*m_allocas)[lclNum] != nullptr)
+    {
+        llvmRef = _builder.CreateLoad((*m_allocas)[lclNum]);
+    }
+    else if (_localsMap->find({lclNum, ssaNum}) == _localsMap->end())
     {
         LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar);
 
@@ -1589,6 +1599,18 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
 
     mapGenTreeToValue(lclVar, llvmRef);
     return llvmRef;
+}
+
+void Llvm::buildLocalVarAddr(GenTreeLclVar* lclVar)
+{
+    LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar->GetLclNum());
+    if (varDsc->lvIsParam)
+    {
+        failFunctionCompilation(); // TOOD-LLVM: args can have their address used when passing to calls
+    }
+
+    unsigned int lclNum = lclVar->GetLclNum();
+    mapGenTreeToValue(lclVar, (*m_allocas)[lclNum]);
 }
 
 // LLVM operations like ICmpNE return an i1, but in the IR it is expected to be an Int (i32).
@@ -1656,25 +1678,32 @@ void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
 {
     if (lclVar->gtFlags & GTF_VAR_DEF)
     {
-        Value* valueRef = getGenTreeValue(lclVar->gtGetOp1());
-        assert(valueRef != nullptr);
+        Value* localValue = getGenTreeValue(lclVar->gtGetOp1());
+        assert(localValue != nullptr);
+
+        unsigned lclNum = lclVar->GetLclNum();
+        Value* allocaValue = (*m_allocas)[lclNum];
+        if (allocaValue != nullptr)
+        {
+            _builder.CreateStore(localValue, castIfNecessary(allocaValue, localValue->getType()->getPointerTo()));
+        }
+
         // This could be done in the NE operator, but sometimes that would be needless, e.g. when followed by JTRUE
         // TODO-LLVM: As this is a zero extend widening operation, this is only valid if the small int is unsigned.  We don't know that here, so likely it would be better to
         // delete this and do the cast in the operator.  It seems likely that the cast will be a nop anyway, at least in Wasm, as Wasm does not have any number types smaller than i32
-        if (valueRef->getType()->isIntegerTy())
+        if (localValue->getType()->isIntegerTy())
         {
-            valueRef = zextIntIfNecessary(valueRef);
+            localValue = zextIntIfNecessary(localValue);
         }
 
         LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar);
-
-        if (varDsc->lvIsParam)
+        if (varDsc->lvIsParam) //TODO-LLVM: not doing params yet
         {
             failFunctionCompilation();
         }
 
-        SsaPair ssaPair = {lclVar->GetLclNum(), lclVar->GetSsaNum()};
-        _localsMap->insert({ssaPair, valueRef });
+        SsaPair ssaPair = {lclNum, lclVar->GetSsaNum()};
+        _localsMap->insert({ssaPair, localValue });
     }
     else
     {
@@ -1714,6 +1743,9 @@ void Llvm::visitNode(GenTree* node)
             break;
         case GT_LCL_VAR:
             localVar(node->AsLclVar());
+            break;
+        case GT_LCL_VAR_ADDR:
+            buildLocalVarAddr(node->AsLclVar());
             break;
         case GT_LSH:
         case GT_RSH:
@@ -1791,6 +1823,8 @@ void Llvm::generateProlog()
     // create a prolog block to store arguments passed on shadow stack, TODO: other things from ILToLLVMImporter to come
     llvm::BasicBlock* prologBlock = llvm::BasicBlock::Create(_llvmContext, "Prolog", _function);
     _prologBuilder.SetInsertPoint(prologBlock);
+
+    createAllocasForLocalsWithAddrOp();
 
     llvm::BasicBlock* block0 = getLLVMBasicBlockForBlock(_compiler->fgFirstBB);
     _prologBuilder.SetInsertPoint(_prologBuilder.CreateBr(block0)); // position _prologBuilder to add locals and arguments
@@ -2339,6 +2373,27 @@ void Llvm::lowerToShadowStack()
 
                 CurrentRange().InsertBefore(node, retAddressLocal, storeNode);
             }
+            else if (node->OperIsLocalAddr())
+            {
+                GenTreeLclVarCommon* localAddrNode  = node->AsLclVarCommon();
+
+                // replace with shadowstack + local offset, if on the shadow stack
+                LclVarDsc* localVarDsc = _compiler->lvaGetDesc(localAddrNode->GetLclNum());
+                int        stackOffset = localVarDsc->GetStackOffset();
+                localVarDsc->lvHasLocalAddr = 1; // TODO-LLVM: GT_LCL_FLD_ADDR will also get set to 1 here, is that a problem?
+                if (stackOffset != BAD_STK_OFFS)
+                {
+                    GenTreeLclVar* shadowStackVar = _compiler->gtNewLclvNode(_shadowStackLclNum, TYP_I_IMPL);
+                    GenTree*       localOffset    = _compiler->gtNewIconNode(stackOffset);
+
+                    CurrentRange().InsertBefore(node, shadowStackVar, localOffset);
+
+                    node->ChangeOper(GT_ADD);
+                    GenTreeOp* binOpNode = node->AsOp();
+                    binOpNode->gtOp1     = shadowStackVar;
+                    binOpNode->gtOp2     = localOffset;
+                }
+            }
         }
     }
 }
@@ -2360,7 +2415,11 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
     {
         LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
-        if (!canStoreLocalOnLlvmStack(varDsc))
+        if (canStoreLocalOnLlvmStack(varDsc))
+        {
+            varDsc->SetStackOffset(BAD_STK_OFFS);
+        }
+        else
         {
             locals.push_back(varDsc);
             if (varDsc->lvIsParam)
@@ -2390,6 +2449,38 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     _shadowStackLocalsSize = offset;
 
     lowerToShadowStack();
+}
+
+void Llvm::createAllocasForLocalsWithAddrOp()
+{
+    m_allocas = new std::vector<Value*>(_compiler->lvaCount, nullptr);
+
+    for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
+    {
+        LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
+        if (varDsc->lvIsParam && varDsc->lvHasLocalAddr)
+        {
+            // TODO-LLVM : copy param to alloca storage
+            // see S_P_CoreLib_System_Diagnostics_Tracing_EventSource__InitializeIsSupported
+            failFunctionCompilation();
+        }
+
+        if (canStoreLocalOnLlvmStack(varDsc) && varDsc->lvHasLocalAddr)
+        {
+            CORINFO_CLASS_HANDLE classHandle = NO_CLASS_HANDLE;
+            if (varDsc->lvType == TYP_STRUCT)
+            {
+                ClassLayout* layout = varDsc->GetLayout();
+                classHandle         = layout->GetClassHandle();
+            }
+            Type* llvmType = getLlvmTypeForCorInfoType(toCorInfoType(varDsc->lvType), classHandle);
+            if (llvmType->isIntegerTy() && llvmType->getIntegerBitWidth() < 32)
+            {
+                llvmType = Type::getInt32Ty(_llvmContext);
+            }
+            (*m_allocas)[lclNum] = _prologBuilder.CreateAlloca(llvmType);
+        }
+    }
 }
 
 //------------------------------------------------------------------------
@@ -2439,6 +2530,12 @@ void Llvm::Compile()
 
     for (BasicBlock* block = _compiler->fgFirstBB; block; block = block->bbNext)
     {
+        // TODO-LLVM: finret basic blocks
+        if (block->bbJumpKind == BBJ_EHFINALLYRET)
+        {
+            failFunctionCompilation();
+        }
+
         startImportingBasicBlock(block);
 
         llvm::BasicBlock* entry = getLLVMBasicBlockForBlock(block);
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -300,10 +300,6 @@ llvm::Type* Llvm::getLlvmTypeForStruct(CORINFO_CLASS_HANDLE structHandle)
                 // Forward-declare the struct in case there's a reference to it in the fields.
                 // This must be a named struct or LLVM hits a stack overflow
                 const char* name = _getTypeName(_thisPtr, structHandle);
-                if (strstr(name, "Register"))
-                {
-                    unsigned i = 0;
-                }
                 llvm::StructType* llvmStructType = llvm::StructType::create(_llvmContext, name);
                 llvmType = llvmStructType;
                 StructDesc* structDesc = getStructDesc(structHandle);
@@ -1722,12 +1718,13 @@ void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
     Value* localValue = consumeValue(lclVar->gtGetOp1(), destLlvmType);
 
     unsigned lclNum = lclVar->GetLclNum();
-    Value* allocaValue = m_allocas[lclNum];
-    LclVarDsc* varDsc  = _compiler->lvaGetDesc(lclVar);
+    LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar);
 
     if (isLlvmFrameLocal(varDsc))
     {
         _builder.CreateStore(localValue, castIfNecessary(allocaValue, destLlvmType->getPointerTo()));
+        Value* lclAddressValue = m_allocas[lclNum];
+        _builder.CreateStore(localValue, castIfNecessary(lclAddressValue, localValue->getType()->getPointerTo()));
     }
     else
     {
@@ -2087,17 +2084,15 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
             node->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
         }
 
+        CurrentRange().InsertBefore(node, shadowStackLocal, offset);
         if (indirOper == GT_NONE)
         {
-            // for GT_LCL_VAR_ADDR, the GT_ADD the result of the conversion,
-            // so replace node with the GT_ADD
+            // Local address nodes are directly replaced with the ADD.
             node->ReplaceWith(lclAddress, _compiler);
-
-            CurrentRange().InsertBefore(node, offset, shadowStackLocal);
         }
         else
         {
-            CurrentRange().InsertBefore(node, offset, shadowStackLocal, lclAddress);
+            CurrentRange().InsertBefore(node, lclAddress);
         }
     }
 }
@@ -2426,10 +2421,8 @@ void Llvm::lowerToShadowStack()
             }
             if (node->OperIsLocalAddr())
             {
-                GenTreeLclVarCommon* localAddrNode  = node->AsLclVarCommon();
-                LclVarDsc* localVarDsc = _compiler->lvaGetDesc(localAddrNode->GetLclNum());
-
-                localVarDsc->lvHasLocalAddr = 1; // TODO-LLVM: GT_LCL_FLD_ADDR will also get set to 1 here, is that a problem?
+                // Indicates that this local is to live on the LLVM frame, and will not participate in SSA.
+                _compiler->lvaGetDesc(node->AsLclVarCommon())->lvHasLocalAddr = 1;
             }
         }
     }
@@ -2507,7 +2500,7 @@ void Llvm::createAllocasForLocalsWithAddrOp()
             if (varDsc->lvIsParam)
             {
                 LlvmArgInfo argInfo = getLlvmArgInfoForArgIx(lclNum);
-                assert(argInfo.IsLlvmArg()); // check this arg is not on the shadow stack
+                assert(argInfo.IsLlvmArg());
                 Value* dataValue = _function->getArg(argInfo.m_argIx);
                 _prologBuilder.CreateStore(dataValue, castIfNecessary(allocaValue, dataValue->getType()->getPointerTo(),
                                                                       &_prologBuilder));
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -2481,6 +2481,7 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     lowerToShadowStack();
 }
 
+    assert(canStoreLocalOnLlvmStack(varDsc) && (_compiler->fgSsaPassesCompleted >= 1));
 void Llvm::createAllocasForLocalsWithAddrOp()
 {
     m_allocas = std::vector<Value*>(_compiler->lvaCount, nullptr);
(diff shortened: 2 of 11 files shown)


SingleAccretion (commented):
The `alloca` approach I agree is the right one. Also agree this should be done in codegen.

One thing to think about is how will this interact with the LLVM "registers" where we put in-SSA locals.

Broadly, there are a few categories of locals we can be dealing with:
1) Live on and used from the shadow stack. These are lowered into explicit loads/stores right now.
2) Home-less locals that only ever live in LLVM registers. We put them in SSA map during codegen and use from there.
3) Locals in SSA that also have a stack home. The current change puts them both into the SSA map on write (that also stores to the stack home, making it analogous to the "write-thru" locals that LSRA supports for other backends), but then uses them only from the stack. We could use them from the SSA map too, presumably, the question is do we want to.
4) Address-exposed locals - these must always be used from the stack.
5) Some locals will be neither address-exposed, nor in SSA, the canonical example being a pinned local. These should be analogous to address-exposed locals for our purposes.

I propose that for simplicity purposes, we collapse cases `3`, `4` and `5` into one: always live on and always used from the stack (`alloca`s). We can refine this later if we find there is a need.

Note: I wonder how much dumping code could we bring back if we define registers for LLVM as `REG_STK` and `REG_LLVM`, to denote the homes for locals.

yowl:
Thanks,  I've made the next iteration.

1. Introduced a `GT_ALLOCA` node to create the LLVM `alloca`
2. Lowered `GT_STORE_LCL_VAR`, `GT_LCL_VAR`, and `GT_LCL_VAR_ADDR` to use the `alloca` which I store in a long lived tmp
3. Added a new phase for this lowering.

I think that addresses your cases 3 and 4.  I don't know how to identify those locals that fall in point 5?

> Note: I wonder how much dumping code could we bring back if we define registers for LLVM as REG_STK and REG_LLVM, to denote the homes for locals.

You've lost me on this.  When you say dumping, are we talking about dumping the tree for debug purposes?

SingleAccretion:
I am sorry for probably not being clear :(, but I do think the original approach of doing this in codegen, instead of lowering, is better. The reason is that now there's a phase ordering problem where we need `lvTracked` (set by ref counting), but shadow stack setup created new locals that we'd like to be tracked. We also don't really want to create lots of locals (it can overflow the limit of tracked locals, which we can increase, but it doesn't seem there's great value in doing so).

The only problem was that we were putting `lvHasAddr` locals (BTW, I would suggest reusing the `DoNotEnregister` setting for this) in SSA. That should be fixable by unsetting `lvTracked` on them after the ref counting pass before SSA.

So the ordering would be:
```
<Lowering (now that's just the shadow stack setup, but it'll include e. g. GC safe point transform in the future ...>
<Ref counting - sets "lvTracked">
<Unset "lvTracked" on locals with their address taken>
<SSA> // This is sort of our "Register allocation"
// At this point no new tracked locals can be added
<Codegen>
```

> I think that addresses your cases 3 and 4. I don't know how to identify those locals that fall in point 5?

`varDsc->lvTracked && varDsc->IsAddressExposed` for that precise category, but a better test for all locals that would need to be used from the frame is `!varDsc->lvTracked || varDsc->lvHasAddr`

> When you say dumping, are we talking about dumping the tree for debug purposes?

Yes, but not the trees - the local variable table, i. e. `lvaTableDump`.

yowl:
Ah, no problem :-)

yowl:
I'm not sure I can use `!varDsc->lvTracked || varDsc->lvDoNotEnregister` as the test as to whether to create an alloca for the local, because `lvDoNotEnregister` is getting set for the shadowstack arg as part of `fgSsaBuild` with this stack
```
>	clrjit_browser_wasm32_x64.dll!Compiler::lvaSetVarDoNotEnregister(unsigned int varNum, Compiler::DoNotEnregisterReason reason) Line 2605	C++
 	clrjit_browser_wasm32_x64.dll!Compiler::lvaSortByRefCount() Line 3565	C++
 	clrjit_browser_wasm32_x64.dll!Compiler::fgLocalVarLivenessInit() Line 196	C++
 	clrjit_browser_wasm32_x64.dll!Compiler::fgLocalVarLiveness() Line 146	C++
 	clrjit_browser_wasm32_x64.dll!SsaBuilder::Build() Line 1759	C++
 	clrjit_browser_wasm32_x64.dll!Compiler::fgSsaBuild() Line 66	C++
```
So I've left `lvHasLocalAddr` for now.  Also the code to reset `lvTracked` doesn't seem to ever get executed, at least for HelloWasm, will see what the other tests do.

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2451,64 @@ void Llvm::PlaceAndConvertShadowStackLocals()
...
+        for (GenTree* node : CurrentRange())
+        {
+            if (node->OperIsLocalAddr())
+            {
+                LclVarDsc* localVarDsc = _compiler->lvaGetDesc(node->AsLclVarCommon()->GetLclNum());
+                if (localVarDsc->lvTracked)
+                {
+                    assert(false);// TODO-LLVM this is never hit?
+                    // localVarDsc->lvTracked = 0;
+                }
```
SingleAccretion: I would expect you can hit this with a struct argument:
```cs
void Caller(Struct s)
{
    Callee(s);
}
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2451,64 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     lowerToShadowStack();
 }
 
+static bool placeLocalInAlloca(LclVarDsc* varDsc)
+{
+    return !varDsc->lvTracked || varDsc->lvHasLocalAddr;
+}
```
SingleAccretion: I suppose with `DontTrackLocalsWithAddress` this becomes just `!varDsc->lvTracked`.

src/coreclr/jit/llvm.cpp
yowl: Thanks, I'll do the `lvIsParam` work and then that should get it.

yowl: I added this test:

```cs
    struct LlvmStruct
    {
        internal int i1, i2;
    }

    static int FuncWithStructArg(LlvmStruct s)
    {
        return s.i2;
    }

    private static void TestStructTracking(LlvmStruct s)
    {
        StartTest("Test passing struct arg");
        
        EndTest(FuncWithStructArg(s) == 2);
    }
```
The LIR ends up (up to the CALL):
```

------------ BB02 [000..01D) (return), preds={BB01} succs={}
               [000024] ------------                 IL_OFFSET void   IL offset: 0x0
               [000025] ------------                 IL_OFFSET void   IL offset: 0x1
N002 (  1,  4) [000019] H-----------        t19 =    CNS_INT(h) ref    0x420098 [ICON_STR_HDL]
               [000031] ------------        t31 =    LCL_VAR   int    V03 tmp2         u:1
               [000032] ------------        t32 =    CNS_INT   int    0
                                                  /--*  t31    int
                                                  +--*  t32    int
               [000033] ------------        t33 = *  ADD       int
                                                  /--*  t33    int
               [000034] ------------        t34 = *  PUTARG_TYPE int
               [000035] ------------        t35 =    LCL_VAR   int    V03 tmp2         u:1
               [000036] ------------        t36 =    CNS_INT   int    0
                                                  /--*  t35    int
                                                  +--*  t36    int
               [000037] ------------        t37 = *  ADD       int
                                                  /--*  t37    int
                                                  +--*  t19    ref
               [000038] ------------              *  STOREIND  ref
                                                  /--*  t34    int    arg-1
N003 ( 15,  7) [000003] --CXG-------              *  CALL      void   Program.StartTest
               [000026] ------------                 IL_OFFSET void   IL offset: 0xb
               [000027] ------------                 IL_OFFSET void   IL offset: 0xc
N002 (  3,  2) [000005] -------N----         t5 =    LCL_VAR_ADDR byref  V00 arg0          (last use)
                                                  /--*  t5     byref
N004 (  9,  7) [000008] n-----------         t8 = *  OBJ       struct<Program+LlvmStruct, 8>
               [000039] ------------        t39 =    LCL_VAR   int    V03 tmp2         u:1
               [000040] ------------        t40 =    CNS_INT   int    0
                                                  /--*  t39    int
                                                  +--*  t40    int
               [000041] ------------        t41 = *  ADD       int
                                                  /--*  t41    int
               [000042] ------------        t42 = *  PUTARG_TYPE int
                                                  /--*  t8     struct
               [000043] ------------        t43 = *  PUTARG_TYPE struct
                                                  /--*  t42    int    arg-1
                                                  +--*  t43    struct arg0
N005 ( 23, 10) [000006] --CXG-------         t6 = *  CALL      int    Program.FuncWithStructArg
```

Resulting in LLVM:
```
define void @HelloWasm_Program__TestStructTracking(i8* %0, %"[HelloWasm]Program+LlvmStruct" %1) !dbg !5 {
Prolog:
  %2 = alloca %"[HelloWasm]Program+LlvmStruct", align 8
  store %"[HelloWasm]Program+LlvmStruct" %1, %"[HelloWasm]Program+LlvmStruct"* %2, align 1
  br label %3

3:                                                ; preds = %Prolog
  br label %4

4:                                                ; preds = %3
  %5 = load i32*, i32** @__Str_Test_passing_struct_arg_402F9E1D2A2D7F7DD38FC1B0EB39F0651A66125099651830DE99655CE79A453E___SYMBOL, align 4, !dbg !7
  %6 = getelementptr i8, i8* %0, i32 0, !dbg !7
  %7 = getelementptr i8, i8* %0, i32 0, !dbg !7
  %8 = bitcast i8* %7 to i8**, !dbg !7
  %9 = bitcast i32* %5 to i8*, !dbg !7
  store i8* %9, i8** %8, align 4, !dbg !7
  call void @HelloWasm_Program__StartTest(i8* %6), !dbg !7
  %10 = load %"[HelloWasm]Program+LlvmStruct", %"[HelloWasm]Program+LlvmStruct"* %2, align 1, !dbg !8
  %11 = getelementptr i8, i8* %0, i32 0, !dbg !8
  %12 = call i32 @HelloWasm_Program__FuncWithStructArg(i8* %11, %"[HelloWasm]Program+LlvmStruct" %10), !dbg !8
  %13 = icmp eq i32 %12, 2, !dbg !9
  %14 = getelementptr i8, i8* %0, i32 0, !dbg !9
  %15 = getelementptr i8, i8* %0, i32 0, !dbg !9
  %16 = bitcast i8* %15 to i8**, !dbg !9
  store i8* null, i8** %16, align 4, !dbg !9
  %17 = zext i1 %13 to i8, !dbg !9
  call void @HelloWasm_Program__EndTest(i8* %14, i8 %17), !dbg !9
  ret void, !dbg !9
}
```
but that assert is not hit.  We do enter `DontTrackLocalsWithAddress` for `V00` but `lvTracked` is `0`

```
<!DOCTYPE html>
<HTML>
<head>
<title>Document</title></head>
<body>
<!--StartFragment-->


  | Name | Value | Type
-- | -- | -- | --
▶ | this | 0x000002164fe40420 {_compiler=0x000002164d44e058 [Program:TestStructTracking(Program+LlvmStruct) (0x2993ab77)] ...} | Llvm *
▶ | _currentBlock | 0x000002164d44f670 BB2; BBJ_RETURN | BasicBlock *
▶ | <begin>$L0 | {m_block=0x000002164d44f670 BB2; BBJ_RETURN } | BasicBlockIterator
▶ | <begin>$L1 | {m_node=0x000002164d44fe58 [GT_LCL_VAR_ADDR, TYP_BYREF V0] } | LIR::ReadOnlyRange::Iterator
▶ | <end>$L0 | {m_block=0x0000000000000000 <NULL> } | BasicBlockIterator
▶ | <end>$L1 | {m_node=0x0000000000000000 <NULL> } | LIR::ReadOnlyRange::Iterator
▶ | <range>$L0 | {m_begin=0x000002164d44f3d0 BB1; BBJ_NONE } | BasicBlockSimpleList &&
▶ | <range>$L1 | {...} | LIR::Range &
◢ | localVarDsc | 0x000002164d44ead0 [V0: TYP_STRUCT] | LclVarDsc *
  | lvType | TYP_STRUCT (15 '\xf') | var_types
  | lvIsParam | 1 '\x1' | unsigned char
  | lvIsRegArg | 1 '\x1' | unsigned char
  | lvFramePointerBased | 0 '\0' | unsigned char
  | lvOnFrame | 1 '\x1' | unsigned char
  | lvRegister | 0 '\0' | unsigned char
  | lvTracked | 0 '\0' | unsigned char
  | lvPinned | 0 '\0' | unsigned char
  | lvMustInit | 0 '\0' | unsigned char
  | lvAddrExposed | 0 '\0' | unsigned char
  | lvDoNotEnregister | 1 '\x1' | unsigned char
  | lvFieldAccessed | 0 '\0' | unsigned char
  | lvLiveInOutOfHndlr | 0 '\0' | unsigned char
  | lvInSsa | 0 '\0' | unsigned char
  | lvVMNeedsStackAddr | 0 '\0' | unsigned char
  | lvLclFieldExpr | 0 '\0' | unsigned char
  | lvLclBlockOpAddr | 0 '\0' | unsigned char
  | lvLiveAcrossUCall | 0 '\0' | unsigned char
  | lvIsCSE | 0 '\0' | unsigned char
  | lvHasLdAddrOp | 0 '\0' | unsigned char
  | lvStackByref | 0 '\0' | unsigned char
  | lvHasILStoreOp | 0 '\0' | unsigned char
  | lvHasMultipleILStoreOp | 0 '\0' | unsigned char
  | lvIsTemp | 0 '\0' | unsigned char
  | lvIsBoolean | 0 '\0' | unsigned char
  | lvSingleDef | 1 '\x1' | unsigned char
  | lvSingleDefRegCandidate | 1 '\x1' | unsigned char
  | lvDisqualifySingleDefRegCandidate | 0 '\0' | unsigned char
  | lvSpillAtSingleDef | 0 '\0' | unsigned char
  | lvDisqualify | 0 '\0' | unsigned char
  | lvVolatileHint | 0 '\0' | unsigned char
  | lvStructDoubleAlign | 0 '\0' | unsigned char
  | lvKeepType | 0 '\0' | unsigned char
  | lvNoLclFldStress | 0 '\0' | unsigned char
  | lvIsPtr | 0 '\0' | unsigned char
  | lvIsUnsafeBuffer | 0 '\0' | unsigned char
  | lvPromoted | 0 '\0' | unsigned char
  | lvIsStructField | 0 '\0' | unsigned char
  | lvOverlappingFields | 0 '\0' | unsigned char
  | lvContainsHoles | 0 '\0' | unsigned char
  | lvCustomLayout | 0 '\0' | unsigned char
  | lvIsMultiRegArg | 0 '\0' | unsigned char
  | lvIsMultiRegRet | 0 '\0' | unsigned char
  | lvUnusedStruct | 0 '\0' | unsigned char
  | lvUndoneStructPromotion | 0 '\0' | unsigned char
  | lvLRACandidate | 0 '\0' | unsigned char
  | lvRegStruct | 0 '\0' | unsigned char
  | lvClassIsExact | 0 '\0' | unsigned char
  | lvClassInfoUpdated | 0 '\0' | unsigned char
  | lvImplicitlyReferenced | 1 '\x1' | unsigned char
  | lvSuppressedZeroInit | 0 '\0' | unsigned char
  | lvHasExplicitInit | 0 '\0' | unsigned char
  | lvFieldLclStart | 0 | unsigned int
  | lvParentLcl | 0 | unsigned int
  | lvFieldCnt | 0 '\0' | unsigned char
  | lvFldOffset | 0 '\0' | unsigned char
  | lvFldOrdinal | 0 '\0' | unsigned char
  | lvLlvmArgNum | 1 | unsigned int
  | lvCorInfoType | CORINFO_TYPE_VALUECLASS (19) | CorInfoType
  | lvHasLocalAddr | 1 '\x1' | unsigned char
  | lvSingleDefDisqualifyReason | 72 'H' | unsigned char
  | _lvRegNum | 0 '\0' | unsigned char
  | _lvOtherReg | 0 '\0' | unsigned char
  | _lvArgReg | 1 '\x1' | unsigned char
  | _lvArgInitReg | 0 '\0' | unsigned char
  | lvVarIndex | 0 | unsigned short
  | m_lvRefCnt | 4 | unsigned short
  | m_lvRefCntWtd | 400.000000 | float
  | lvStkOffs | -1163005939 | int
  | lvExactSize | 8 | unsigned int
  | lvSlotNum | 0 | unsigned int
  | ▶ lvVerTypeInfo | {m_bits={type=TI_STRUCT (2) uninitobj=0 byref=0 ...} m_flags=2 m_cls=0x4000000000420028 {...} ...} | typeInfo
  | lvClassHnd | 0x4000000000420028 {...} | CORINFO_CLASS_STRUCT_ *
  | lvFieldHnd | 0x0000000000000000 <NULL> | CORINFO_FIELD_STRUCT_ *
  | ▶ m_layout | 0x000002164d44f370 {m_classHandle=0x4000000000420028 {...} m_size=8 m_isValueClass=1 ...} | ClassLayout *
  | ▶ lvRefBlks | 0x0000000000000000 {???} | unsigned __int64 *
  | ▶ lvDefStmt | 0x0000000000000000 <NULL> | Statement *
  | ▶ lvPerSsaData | {m_array=0x0000000000000000 <NULL> m_arraySize=0 m_count=0 } | SsaDefArray<LclSsaVarDsc>
  | ▶ lvReason | 0x0000000000000000 <NULL> | const char *
▶ | node | 0x000002164d44fe58 [GT_LCL_VAR_ADDR, TYP_BYREF V0] | GenTree * {GenTreeLclVar}


<!--EndFragment-->
</body>
</HTML>
```

yowl: Maybe with optimisations on, this would get hit.

SingleAccretion: Yes...

I realized it might not be that straightforward to clear `lvTracked` before SSA. The ordering here is:
```c
fgSsaBuild:
  lvaSortByRefCount();  // Sets lvTracked
  fgLocalVarLiveness(); // We'd like it to use the ordinary definition for "lvTracked"
  <SSA construction>    // We'd like it to not include address-taken variables here
```
What we can do (and I think is the right thing to do) is modify `IncludeInSsa` s.t. it ignores address-taken variables.

Then the test for whether a local should be in an `alloca` is `!lvaInSsa(lclNum)`.

yowl:
I've made a couple of changes here.  

1 Fixed a bug with aligning IR args with LLVM args.  I'm not sure why this didn't manifest earlier, but seems like the current logic was a bit confused with skipping the return buffer arg and `this`
2 I'd forgotten to add the logic in building a `TypeDescriptor` that handled unions.  The `Register` struct used in Vectors was coming out as a 16 `i32`s when we just want 2 `i64`s to match the IL module.

yowl:
Retry as some network error? "Unable to load the service index for source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json. "

yowl:
http error

yowl:
Randomly passed.  Oh well.

src/coreclr/jit/llvm.cpp
```diff
@@ -2339,6 +2380,27 @@ void Llvm::lowerToShadowStack()
 
                 CurrentRange().InsertBefore(node, retAddressLocal, storeNode);
             }
+            else if (node->OperIsLocalAddr())
```
SingleAccretion: I think it would be more consistent to fold this into `ConvertShadowStackLocalNode` (except for the setting of `lvHasLocalAddr` that should remain in this top-level method).

src/coreclr/jit/llvm.cpp
```diff
@@ -2339,6 +2380,27 @@ void Llvm::lowerToShadowStack()
...
+            {
+                GenTreeLclVarCommon* localAddrNode  = node->AsLclVarCommon();
+
+                // replace with shadowstack + local offset, if on the shadow stack
+                LclVarDsc* localVarDsc = _compiler->lvaGetDesc(localAddrNode->GetLclNum());
+                int        stackOffset = localVarDsc->GetStackOffset();
+                localVarDsc->lvHasLocalAddr = 1; // TODO-LLVM: GT_LCL_FLD_ADDR will also get set to 1 here, is that a problem?
+                if (stackOffset != BAD_STK_OFFS)
```
SingleAccretion: See also the comment about `ConvertShadowStackLocalNode`. I think it would be best to check for this via `canStoreLocalOnLlvmStack` everywhere.

src/coreclr/jit/llvm.cpp
```diff
@@ -2360,7 +2422,11 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
     {
         LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
-        if (!canStoreLocalOnLlvmStack(varDsc))
+        if (canStoreLocalOnLlvmStack(varDsc))
+        {
+            varDsc->SetStackOffset(BAD_STK_OFFS);
+        }
+        else
```
SingleAccretion: Can be deleted once the dependency on `SetStackOffset` for non-shadow-stack locals is removed.

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2458,48 @@ void Llvm::PlaceAndConvertShadowStackLocals()
...
+
+void Llvm::createAllocasForLocalsWithAddrOp()
+{
+    m_allocas = new std::vector<Value*>(_compiler->lvaCount, nullptr);
+
+    for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
+    {
+        LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
+
+        // TODO-LLVM LCLBLK - outgoing arg space
```
SingleAccretion: Side note: I wonder if we could/should turn off `FEATURE_FIXED_OUT_ARGS` and delete all these `LCLBLK` checks, they're just noise.

(Not in this change of course)

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2458,48 @@ void Llvm::PlaceAndConvertShadowStackLocals()
     lowerToShadowStack();
 }
 
+static bool placeLocalInAlloca(LclVarDsc* varDsc)
+{
+    return !varDsc->lvInSsa;
+}
```
```suggestion
static bool placeLocalInAlloca(LclVarDsc* varDsc)
{
    assert(canStoreLocalOnLlvmStack(varDsc) && (_compiler->fgSsaPassesCompleted >= 1));
    return !varDsc->lvInSsa;
}
```
We should not be asking this question for shadow stack locals, we also should not be asking it before codegen.

(Would be nice to start thinking about separating codegen into its own class)

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2458,48 @@ void Llvm::PlaceAndConvertShadowStackLocals()
...
+
+        // TODO-LLVM LCLBLK - outgoing arg space
+        if (varDsc->lvType != TYP_LCLBLK && canStoreLocalOnLlvmStack(varDsc) && placeLocalInAlloca(varDsc))
+        {
+            CORINFO_CLASS_HANDLE classHandle = NO_CLASS_HANDLE;
+            if (varDsc->lvType == TYP_STRUCT)
+            {
+                ClassLayout* layout = varDsc->GetLayout();
+                classHandle         = layout->GetClassHandle();
+            }
```
```suggestion
            CORINFO_CLASS_HANDLE classHandle = tryGetStructClassHandle(varDsc);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2458,48 @@ void Llvm::PlaceAndConvertShadowStackLocals()
...
+            Type* llvmType = getLlvmTypeForCorInfoType(toCorInfoType(varDsc->lvType), classHandle);
```
```suggestion
            Type* llvmType = getLlvmTypeForCorInfoType(toCorInfoType(varDsc->TypeGet()), classHandle);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2458,48 @@ void Llvm::PlaceAndConvertShadowStackLocals()
...
+            {
+                llvmType = Type::getInt32Ty(_llvmContext);
+            }
+            Value* allocaValue = _prologBuilder.CreateAlloca(llvmType);
+            (*m_allocas)[lclNum] = allocaValue;
+
+            if (varDsc->lvIsParam)
+            {
+                LlvmArgInfo argInfo = getLlvmArgInfoForArgIx(lclNum);
+                assert(argInfo.m_argIx >= 0); // check this arg is not on the shadow stack
```
SingleAccretion: A small helper could be added to `LlvmArgInfo` that checks for this, say `IsLlvmArg()`.

src/coreclr/jit/llvm.h
```diff
@@ -108,6 +108,7 @@ class Llvm
     std::unordered_map<GenTree*, Value*>* _sdsuMap;
     std::unordered_map<SsaPair, Value*, SsaPairHash>* _localsMap;
     std::vector<PhiPair> _phiPairs;
+    std::vector<Value*>* m_allocas;
```
SingleAccretion: `std::vector`s, AFAIK, are fairly small in size in most standard libraries, so I don't think there's a need for additional indirections.
```suggestion
    std::vector<Value*> m_allocas;
```
(Will also entail changing code which uses this)

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -316,10 +316,51 @@ public static TypeDescriptor getTypeDescriptor(IntPtr thisHandle, CORINFO_CLASS_
                 hasSignificantPadding = ecmaType.IsExplicitLayout || ecmaType.GetClassLayout().Size > 0;
             };
 
-            uint fieldCount = 0;
+            // Replicate the logic in \ILToLLVMImporter.cs GetLLVMTypeForTypeDesc but without Linq
+            // Build the TypeDescriptor with the largest fields at each offset,
+            // removing overlapping fields - i.e. union or explicit layouts
```
SingleAccretion: So...

Given we now have this logic in the VM instead of the Jit, should the same logic in the Jit be deleted?

src/coreclr/jit/llvm.cpp
```diff
@@ -1656,25 +1685,32 @@ void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
...
     if (lclVar->gtFlags & GTF_VAR_DEF)
     {
-        Value* valueRef = getGenTreeValue(lclVar->gtGetOp1());
-        assert(valueRef != nullptr);
+        Value* localValue = getGenTreeValue(lclVar->gtGetOp1());
+        assert(localValue != nullptr);
+
+        unsigned lclNum = lclVar->GetLclNum();
+        Value* allocaValue = (*m_allocas)[lclNum];
+        if (allocaValue != nullptr)
```
SingleAccretion: I think it would be better to use `placeLocalInAlloca` here (maybe rename it to, say, `isLlvmFrameLocal`) instead of the `!= nullptr` test, for consistency.

src/coreclr/jit/llvm.cpp
```diff
@@ -1656,25 +1685,32 @@ void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
 {
     if (lclVar->gtFlags & GTF_VAR_DEF)
```
SingleAccretion: Hmm, while you're at it, delete this `lclVar->gtFlags & GTF_VAR_DEF` check, it's not useful (`STORE_LCL_VAR` should always have this flag set).

src/coreclr/jit/llvm.cpp
```diff
@@ -1656,25 +1685,32 @@ void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
...
-
-        if (varDsc->lvIsParam)
+        if (varDsc->lvIsParam) //TODO-LLVM: not doing params yet
         {
             failFunctionCompilation();
         }
 
-        SsaPair ssaPair = {lclVar->GetLclNum(), lclVar->GetSsaNum()};
-        _localsMap->insert({ssaPair, valueRef });
+        SsaPair ssaPair = {lclNum, lclVar->GetSsaNum()};
```
SingleAccretion, line 1747: Will it store `RESERVED_SSA_NAME` for locals live on the frame?

That doesn't seem right, there should probably be an early return above, or an `if-else` structure (which could also be said to be better).

src/coreclr/jit/llvm.cpp
```diff
@@ -1591,6 +1613,13 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
     return llvmRef;
 }
 
+void Llvm::buildLocalVarAddr(GenTreeLclVar* lclVar)
```
SingleAccretion: Nit: `lclVar` -> `lclAddr`.

We should handle `GT_LCL_FLD_ADDR` here too via a GEP.

src/coreclr/jit/llvm.cpp
```diff
@@ -1656,25 +1685,32 @@ void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
 {
     if (lclVar->gtFlags & GTF_VAR_DEF)
     {
-        Value* valueRef = getGenTreeValue(lclVar->gtGetOp1());
-        assert(valueRef != nullptr);
+        Value* localValue = getGenTreeValue(lclVar->gtGetOp1());
+        assert(localValue != nullptr);
```
SingleAccretion: Should this assert be inside `getGenTreeValue`?

src/coreclr/jit/llvm.cpp
yowl: Is there a performance or size penalty in doing that?

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
yowl: Honestly, I forgot we had the same logic in `getStructDesc` .  I'll see why it didn't work for `Register` with a view to removing it from the VM.

src/coreclr/jit/llvm.cpp
yowl: Moved.

src/coreclr/jit/llvm.cpp
yowl: I added some code for `GT_LCL_FLD_ADDR ` but I'm not sure that it actually brings in any new methods to the compilation.

src/coreclr/jit/llvm.cpp
yowl: Thanks, changed to if/else

src/coreclr/jit/llvm.cpp
yowl: Done

src/coreclr/jit/llvm.cpp
yowl: Done.

src/coreclr/jit/llvm.cpp
yowl: Added.

src/coreclr/jit/llvm.cpp
yowl: Yep.

src/coreclr/jit/llvm.cpp
yowl: Done.

src/coreclr/jit/llvm.cpp
yowl: Yes, turned out better.  Thanks

src/coreclr/jit/llvm.cpp
yowl: Ok, when you say codegen, it refers to anything used once we get to the `Llvm::Compile` method?

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
yowl: I had 2 bugs in the original Jit code.   I was using the field's owning type for the class handle, not the actual field, and I was not clearing out the covered fields in unions.  Having addressed those, then the `Register` struct generated correctly.

src/coreclr/jit/llvm.cpp
```diff
@@ -2339,6 +2430,13 @@ void Llvm::lowerToShadowStack()
 
                 CurrentRange().InsertBefore(node, retAddressLocal, storeNode);
             }
+            if (node->OperIsLocalAddr())
+            {
+                GenTreeLclVarCommon* localAddrNode  = node->AsLclVarCommon();
+                LclVarDsc* localVarDsc = _compiler->lvaGetDesc(localAddrNode->GetLclNum());
+
+                localVarDsc->lvHasLocalAddr = 1; // TODO-LLVM: GT_LCL_FLD_ADDR will also get set to 1 here, is that a problem?
```
```suggestion
                // Indicates that this local is to live on the LLVM frame, and will not participate in SSA.
                _compiler->lvaGetDesc(node->AsLclVarCommon())->lvHasLocalAddr = 1;
```
(`LCL_FLD_ADDR` locals should also get it of course)

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2490,38 @@ void Llvm::PlaceAndConvertShadowStackLocals()
...
+            {
+                llvmType = Type::getInt32Ty(_llvmContext);
+            }
+            Value* allocaValue = _prologBuilder.CreateAlloca(llvmType);
+            m_allocas[lclNum] = allocaValue;
+
+            if (varDsc->lvIsParam)
+            {
+                LlvmArgInfo argInfo = getLlvmArgInfoForArgIx(lclNum);
+                assert(argInfo.IsLlvmArg()); // check this arg is not on the shadow stack
```
```suggestion
                assert(argInfo.IsLlvmArg());
```
(Self-evident now)

src/coreclr/jit/llvm.h
```diff
@@ -38,8 +38,13 @@ struct OperandArgNum
 
 struct LlvmArgInfo
 {
-    int m_argIx; // -1 indicates not in the LLVM arg list, but on the shadow stack
+    int          m_argIx; // -1 indicates not in the LLVM arg list, but on the shadow stack
     unsigned int m_shadowStackOffset;
+
+    bool         IsLlvmArg()
```
```suggestion
    bool IsLlvmArg()
```
(Fun fact: this is how alignment used to be in the Jit codebase before clang formatting)

src/coreclr/jit/llvm.h
```diff
@@ -194,6 +200,7 @@ class Llvm
     GenTreeCall::Use* lowerCallReturn(GenTreeCall* callNode, CORINFO_SIG_INFO* calleeSigInfo, GenTreeCall::Use* lastArg);
     void lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO* calleeSigInfo);
     void lowerToShadowStack();
+    void createAllocasForLocalsWithAddrOp();
```
SingleAccretion: This is codegen-related, so should be grouped with those functions?

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -444,6 +446,25 @@ private static unsafe void TestStructStoreWithSignificantPadding()
         EndTest(*ptrCopySized == 2);
     }
 
+    struct LlvmStruct
+    {
+        internal int i1, i2;
+    }
+
+    // dont want this inlined as it wi the `LCL_VAR_ADDR`
+    [MethodImpl(MethodImplOptions.NoInlining)]
```
SingleAccretion: Most codegen tests mark methods as noinline for similar reasons.
```suggestion
    [MethodImpl(MethodImplOptions.NoInlining)]
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1652,33 +1708,43 @@ Value* Llvm::getLocalVarAddress(GenTreeLclVar* lclVar) {
...
-    if (lclVar->gtFlags & GTF_VAR_DEF)
+    Value* localValue = getGenTreeValue(lclVar->gtGetOp1());
+
+    unsigned lclNum = lclVar->GetLclNum();
+    Value* allocaValue = m_allocas[lclNum];
+    LclVarDsc* varDsc  = _compiler->lvaGetDesc(lclVar);
+
+    if (isLlvmFrameLocal(varDsc))
+    {
+        _builder.CreateStore(localValue, castIfNecessary(allocaValue, localValue->getType()->getPointerTo()));
```
```suggestion
    LclVarDsc* varDsc = _compiler->lvaGetDesc(lclVar);

    if (isLlvmFrameLocal(varDsc))
    {
        Value* lclAddressValue = m_allocas[lclNum];
        _builder.CreateStore(localValue, castIfNecessary(lclAddressValue, localValue->getType()->getPointerTo()));
```
(Moved it to the scope where it is used)

src/coreclr/jit/llvm.cpp
```diff
@@ -1652,33 +1708,43 @@ Value* Llvm::getLocalVarAddress(GenTreeLclVar* lclVar) {
...
+bool Llvm::isLlvmFrameLocal(LclVarDsc* varDsc)
+{
+    assert(canStoreLocalOnLlvmStack(varDsc) && (_compiler->fgSsaPassesCompleted >= 1));
+    return !varDsc->lvInSsa;
+}
+
 void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
 {
-    if (lclVar->gtFlags & GTF_VAR_DEF)
+    Value* localValue = getGenTreeValue(lclVar->gtGetOp1());
```
SingleAccretion: So, one thing to consider here: what happens if our value is small (`IND byte`, say), and we're storing to a `TYP_INT` local.

I think as written, the code will only store the low byte, which is not correct.

This should call `consumeValue` with the actual type of the `lclVar->gtGetOp1()` node as the second argument.

src/coreclr/jit/llvm.cpp
```diff
@@ -1652,33 +1708,43 @@ Value* Llvm::getLocalVarAddress(GenTreeLclVar* lclVar) {
...
-        if (valueRef->getType()->isIntegerTy())
+        // TODO-LLVM: As this is a zero extend widening operation, this is only valid if the small int is unsigned.  We
+        // don't know that here, so likely it would be better to delete this and do the cast in the operator.  It seems
+        // likely that the cast will be a nop anyway, at least in Wasm, as Wasm does not have any number types smaller
+        // than i32
+        if (localValue->getType()->isIntegerTy())
         {
-            valueRef = zextIntIfNecessary(valueRef);
+            localValue = zextIntIfNecessary(localValue);
         }
```
SingleAccretion: Should be deleted once replaced with `consumeValue`.

src/coreclr/jit/llvm.cpp
```diff
@@ -2013,7 +2091,20 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
             blk->SetLayout(_compiler->typGetObjLayout(handle));
             blk->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
         }
-        CurrentRange().InsertBefore(node, offset, shadowStackLocal, lclAddress);
+
```
SingleAccretion, line 2096:
```suggestion
```
(Unnecessary space)

src/coreclr/jit/llvm.cpp
```diff
@@ -2013,7 +2091,20 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+            // for GT_LCL_VAR_ADDR, the GT_ADD the result of the conversion,
+            // so replace node with the GT_ADD
+            node->ReplaceWith(lclAddress, _compiler);
+
+            CurrentRange().InsertBefore(node, offset, shadowStackLocal);
+        }
+        else
+        {
+            CurrentRange().InsertBefore(node, offset, shadowStackLocal, lclAddress);
+        }
```
SingleAccretion: How about this:
```suggestion
        CurrentRange().InsertBefore(node, shadowStackLocal, offset);

        if (indirOper == GT_NONE)
        {
            // Local address nodes are directly replaced with the ADD.
            node->ReplaceWith(lclAddress, _compiler);
        }
        else
        {
            CurrentRange().InsertBefore(node, lclAddress);
        }
```
Reads a little cleaner to me.

src/coreclr/jit/llvm.cpp
```diff
@@ -1996,11 +2068,17 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
             case GT_LCL_VAR:
                 indirOper = lclVar->TypeIs(TYP_STRUCT) ? GT_OBJ : GT_IND;
                 break;
+            case GT_LCL_VAR_ADDR:
```
SingleAccretion, line 2074: We should enable `LCL_FLD_ADDR` here too, for that you need to add the offset to `offset`:
```diff
-        GenTreeIntCon* offset = _compiler->gtNewIconNode(varDsc->GetStackOffset(), TYP_I_IMPL);
+        GenTreeIntCon* offset = _compiler->gtNewIconNode(varDsc->GetStackOffset() + node->GetLclOffs(), TYP_I_IMPL);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2013,7 +2091,20 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
             blk->SetLayout(_compiler->typGetObjLayout(handle));
```
SingleAccretion: Not all code is shown here, but it can be simplified:
```diff
-            GenTreeBlk* blk = node->AsBlk();
-            CORINFO_CLASS_HANDLE handle = varDsc->GetStructHnd();
-            blk->SetLayout(_compiler->typGetObjLayout(handle));
-            blk->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
+            node->AsBlk()->SetLayout(varDsc->GetLayout());
+            node->AsBlk()->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1591,6 +1633,20 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
     return llvmRef;
 }
 
+void Llvm::buildLocalVarAddr(GenTreeLclVarCommon* lclAddr)
+{
+    unsigned int lclNum = lclAddr->GetLclNum();
+    if (lclAddr->isLclField())
+    {
+        GenTreeLclFld* lclFldAddr = lclAddr->AsLclFld();
+        mapGenTreeToValue(lclAddr, _builder.CreateGEP(m_allocas[lclNum], _builder.getInt16(lclFldAddr->GetLclOffs())));
```
SingleAccretion: Does `CreateGEP` use the bytewise offset here?

I remember we had a problem where it was using scaled addresses, so, for `i32*`, `GEP`ing `4` would be adding `16` bytes.

src/coreclr/jit/llvm.cpp
```diff
@@ -1555,7 +1593,11 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
     Value*       llvmRef;
     unsigned int lclNum = lclVar->GetLclNum();
     unsigned int ssaNum = lclVar->GetSsaNum();
-    if (_localsMap->find({lclNum, ssaNum}) == _localsMap->end())
+    if (m_allocas[lclNum] != nullptr)
```
```suggestion
    if (isLlvmFrameLocal(varDsc))
```
(`varDsc` definition will need to be moved to before the `if`)

src/coreclr/jit/compiler.h
```diff
@@ -127,7 +127,7 @@ unsigned ReinterpretHexAsDecimal(unsigned in);
 
 const unsigned FLG_CCTOR = (CORINFO_FLG_CONSTRUCTOR | CORINFO_FLG_STATIC);
 
-#ifdef DEBUG
+#if defined(DEBUG) || defined(TARGET_WASM)
```
```suggestion
#ifdef DEBUG
```
No longer needed.

src/coreclr/jit/llvm.cpp
```diff
@@ -171,6 +175,7 @@ StructDesc* Llvm::getStructDesc(CORINFO_CLASS_HANDLE structHandle)
         unsigned structSize                 = _info.compCompHnd->getClassSize(structHandle); // TODO-LLVM: add to TypeDescriptor?
 
         std::vector<CORINFO_FIELD_HANDLE> sparseFields = std::vector<CORINFO_FIELD_HANDLE>(structSize);
+        std::vector<unsigned> sparseFieldSizess = std::vector<unsigned>(structSize);
```
SingleAccretion: `sparseFieldSizess` -> `sparseFieldSizes`, presumably a typo.

src/coreclr/jit/llvm.cpp
```diff
@@ -179,26 +184,37 @@ StructDesc* Llvm::getStructDesc(CORINFO_CLASS_HANDLE structHandle)
...
+            else
             {
                 fieldCount++;
+                // clear out any fields that are covered by this field
+                for (unsigned j = 1; j < sparseFieldSizess[i]; j++)
+                {
+                    sparseFields[i + j] = nullptr;
+                }
+                i += sparseFieldSizess[i];
             }
```
SingleAccretion: Nit: "unbalanced" `if`s like this one are unusual. I would recommend reversing the condition.

src/coreclr/jit/llvm.cpp
```diff
@@ -284,6 +300,10 @@ llvm::Type* Llvm::getLlvmTypeForStruct(CORINFO_CLASS_HANDLE structHandle)
                 // Forward-declare the struct in case there's a reference to it in the fields.
                 // This must be a named struct or LLVM hits a stack overflow
                 const char* name = _getTypeName(_thisPtr, structHandle);
+                if (strstr(name, "Register"))
+                {
+                    unsigned i = 0;
+                }
```
```suggestion
```
Presumably not needed.

src/coreclr/jit/llvm_types.h
```diff
@@ -15,6 +15,7 @@ struct TypeDescriptor
 private:
     unsigned              m_fieldCount;
     CORINFO_FIELD_HANDLE* m_fields;
+    unsigned*             m_fieldSizes;
```
SingleAccretion: How are the field sized here different from those you get with `getClassSize`? Is the difference that they include alignment? This needs to be documented.

src/coreclr/jit/llvm_types.h
```diff
@@ -32,6 +33,11 @@ struct TypeDescriptor
         return m_fields[index];
     }
 
+    unsigned getFieldSize(unsigned index)
+    {
+        return m_fieldSizes[index];
+    }
```
```suggestion
    unsigned getFieldSize(unsigned index)
    {
        assert(index < m_fieldCount);
        return m_fieldSizes[index];
    }
```

src/coreclr/jit/llvm.cpp
SingleAccretion: As far as the Jit is concerned, it can be thought to be "better" because it allows nested calls (calls as arguments), thus fewer temps.

`FEATURE_FIXED_OUT_ARGS` is really about the low-level ABI detail of whether the stack pointer remains fixed throughout the method, or can change (via `push` instructions on x86). We don't need to worry about that when targeting LLVM (except for the shadow stack but there we decide ourselves and do not need to concern the rest of the Jit with those decisions).

src/coreclr/jit/llvm.cpp
yowl: ok, I normally do the positive condition first, but no problem.

src/coreclr/jit/llvm.cpp
yowl: right, just some debug

src/coreclr/jit/llvm.cpp
yowl: yes, you are right.

src/coreclr/jit/llvm.cpp
yowl: Yes, this looks better

src/coreclr/jit/llvm_types.h
yowl: I don't think they are, the problem is how to get the field class when in the JIT.  I did it to reduce the interop partly and also you could ask for the class size from the `CORINFO_FIELD_HANDLE` but we'd need to add a callback for it as there is nothing in the interface.

src/coreclr/jit/llvm.cpp
yowl: yes, thanks

src/coreclr/jit/llvm.h
yowl: done

src/coreclr/jit/llvm.cpp
yowl: Done.

src/coreclr/jit/llvm.cpp
yowl: Thanks, deleted.

src/coreclr/jit/llvm.cpp
yowl: ok, I just reassigned `offset`, discarding the original value.

src/coreclr/jit/llvm.cpp
yowl: I made your suggested change.  Seems fine, what code is missing?

SingleAccretion: Oh, I just meant that GH diff view doesn't show the this code fully:
```c
        if (GenTree::OperIsBlk(indirOper))
        {
            GenTreeBlk* blk = node->AsBlk();
            CORINFO_CLASS_HANDLE handle = varDsc->GetStructHnd();
            blk->SetLayout(_compiler->typGetObjLayout(handle));
            blk->gtBlkOpKind = GenTreeBlk::BlkOpKindInvalid;
        }
```

src/coreclr/jit/llvm.cpp
SingleAccretion: I think it would be better to create the right one from the start. This offset will be the right one for all cases, `LCL_VAR/LCL_FLD/STORE_LCL_VAR/STORE_LCL_FLD`, not just addresses.

src/coreclr/jit/llvm_types.h
SingleAccretion: > the problem is how to get the field class when in the JIT

Wouldn't `getFieldType` + `getClassSize` work here?

I think it is ok to take the interop cost hit for this building code, as it is run only once per type per process. Holding the (enlarged) type descriptor data structures in memory is not desirable (we want the type descriptor itself be as small as reasonable, as it is supposed to only be used for interop, and we already duplicate "the type system" on the on the Jit side with the `StructDesc*` map (that is supposed to be for "hot" data).

Another way to approach this is to build the `StructDesc*`s for field descriptors greedily, i. e. before filtering out the union-ed fields, and then save and get the size there (we may end up with some unused descriptors in the map though).

src/coreclr/jit/llvm.cpp
yowl: I can do that with a tertiary : ` + (node->isLclField() ? node->AsLclFld()->GetLclOffs() : 0)`  ?

SingleAccretion: `GetLclOffs` [is available](https://github.com/dotnet/runtimelab/blob/b135cf9b031395ffddb1b8479f1f9aef485e43ee/src/coreclr/jit/gentree.h#L3362) on `GenTreeLclVarCommon` nodes (and [does](https://github.com/dotnet/runtimelab/blob/b135cf9b031395ffddb1b8479f1f9aef485e43ee/src/coreclr/jit/gentree.cpp#L20300) that, yep).

src/coreclr/jit/llvm_types.h
yowl: I dont think so, `getFieldType` does not return the `CORINFO_CLASS_STRUCT_` for the field.
```
        private CorInfoType getFieldType(CORINFO_FIELD_STRUCT_* field, CORINFO_CLASS_STRUCT_** structType, CORINFO_CLASS_STRUCT_* memberParent)
        {
            FieldDesc fieldDesc = HandleToObject(field);
            TypeDesc fieldType = fieldDesc.FieldType;

            CorInfoType type;
            if (structType != null)
            {
                type = asCorInfoType(fieldType, structType);
            }
            else
            {
                type = asCorInfoType(fieldType);
            }

            Debug.Assert(!fieldDesc.OwningType.IsByReferenceOfT ||
                fieldDesc.OwningType.GetKnownField("_value").FieldType.Category == TypeFlags.IntPtr);
            if (type == CorInfoType.CORINFO_TYPE_NATIVEINT && fieldDesc.OwningType.IsByReferenceOfT)
            {
                Debug.Assert(structType == null || *structType == null);
                Debug.Assert(fieldDesc.Offset.AsInt == 0);
                type = CorInfoType.CORINFO_TYPE_BYREF;
            }

            return type;
        }
```

SingleAccretion: Isn't that what the `CORINFO_CLASS_STRUCT_** structType` out parameter does?

src/coreclr/jit/llvm.cpp
yowl: Ah, this is hitting an assert.  I'll need to see what that is about.

SingleAccretion: I think it needs to handle `TYP_STRUCT`s.

yowl: Maybe structs are also a problem, but this one is a phi bool:
```
------------ BB08 [099..09E) -> BB09 (always), preds={BB06,BB07} succs={BB09}
     (  0,  0) [000366] ------------       t366 =    PHI_ARG   bool   V00 loc0         u:8
     (  0,  0) [000365] ------------       t365 =    PHI_ARG   bool   V00 loc0         u:6
                                                  /--*  t366   bool
                                                  +--*  t365   bool
               [000357] ------------       t357 = *  PHI       bool
                                                  /--*  t357   bool
               [000358] DA----------              *  STORE_LCL_VAR bool   V00 loc0         d:7
               [000212] ------------                 IL_OFFSET void   IL offset: 0x99
N001 (  3,  2) [000095] ------------        t95 =    LCL_VAR   int    V00 loc0         u:7 (last use)
                                                  /--*  t95    int
N003 (  7,  5) [000097] DA----------              *  STORE_LCL_VAR int    V05 loc5         d:2
```
The node is `TYP_INT`, the LLVM type is `i32`, but the value stored in the `_sdsuMap` is an `i8`.  As the `node` is a `TYP_INT`, `varTypeIsSmall(node)` is false and it goes to the "special case for relops" which expects an `i1`

yowl: Should we `consumeValue` at 
https://github.com/yowl/runtimelab/blob/b135cf9b031395ffddb1b8479f1f9aef485e43ee/src/coreclr/jit/llvm.cpp#L1619 
?

SingleAccretion: I think the current place is right. We just need to follow some rules on what kinds of type we are to pass as the second parameter (writing them out right now).

SingleAccretion: So, first of all, we have two concepts here:

1) Location types, i. e. those we are to place in the SSA map. These can be small, and must match the local's exact type.
2) Raw tree types. Can be small, exist in the SDSU map, and also can be mismatched with the location types, but only if small.

To maintain the invariant under `1` (only exact types stored in locations), we are to pass the type of the local to `consumeValue` (and ignore the node type).

Now, we also need to decide what location type is proper for a local. This is a little dodgy because the Jit itself is not sure (there are known bugs that exist because of it).

For now, for simplicity, I propose we treat location types as simply `varDsc->TypeGet()`. It should also be used when building the `alloca`s. We could choose to use `TYP_INT` for small-typed locals as well, but it would require more thinking.

This won't solve the assert you observed, where we should have an `i8 -> i32` implicit upcast. This needs another special case (small locals in IR are special) in `consumeValue`: `if node->OperIs(GT_LCL_VAR) { /* use the local's type  to determine zext or sext */ }`.

src/coreclr/jit/llvm_types.h
yowl: missed that, will try.

src/coreclr/jit/llvm.cpp
```diff
@@ -887,6 +925,16 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
...
             }
+            else if (node->OperIs(GT_LCL_VAR))
+            {
+                // implicit upcast i8 -> i32
+                assert(nodeValue->getType() == Type::getInt8Ty(_llvmContext));
+
+                GenTreeLclVarCommon* lclNode = node->AsLclVarCommon();
+                LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNode->GetLclNum());
+                finalValue = varTypeIsSigned(varDsc->TypeGet()) ? _builder.CreateSExt(nodeValue, targetLlvmType)
+                                                   : _builder.CreateZExt(nodeValue, targetLlvmType);
```
```suggestion
                // In IR, small locals are usually typed as INTs, while in LLVM registers we store them
                // with their "true" type. So here we must re-extend them (as the user requests we do so).
                LclVarDsc* varDsc = _compiler->lvaGetDesc(node->AsLclVarCommon());
                assert(varTypeIsSmall(varDsc));
                
                finalValue = varTypeIsSigned(varDsc) ? _builder.CreateSExt(nodeValue, targetLlvmType)
                                                     : _builder.CreateZExt(nodeValue, targetLlvmType);
```
Any small type can be seen here (and any small type is legal).

yowl: Thanks!

src/coreclr/jit/llvm.cpp
yowl: Thanks, I removed the widening for small ints for the allocas

src/coreclr/tools/Common/TypeSystem/Common/MetadataType.cs
```diff
@@ -103,6 +103,8 @@ public struct FieldAndOffset
     {
         public static readonly LayoutInt InvalidOffset = new LayoutInt(int.MaxValue);
 
+        public static FieldAndOffset Invalid = new FieldAndOffset(null, InvalidOffset);
+
```
SingleAccretion: Dead code?
```suggestion
```

src/coreclr/tools/Common/TypeSystem/Common/MetadataType.cs
```diff
@@ -112,5 +114,10 @@ public FieldAndOffset(FieldDesc field, LayoutInt offset)
             Field = field;
             Offset = offset;
         }
+
+        public bool IsValid
+        {
+            get { return Offset != InvalidOffset; }
+        }
```
SingleAccretion:
```suggestion
```

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -327,7 +327,9 @@ public static TypeDescriptor getTypeDescriptor(IntPtr thisHandle, CORINFO_CLASS_
 
             //TODO-LLVM: change to NativeMemory.Alloc when upgraded to .net6
             IntPtr fieldArray = Marshal.AllocHGlobal((int)(sizeof(CORINFO_FIELD_STRUCT_*) * fieldCount));
+            IntPtr fieldSizes = Marshal.AllocHGlobal((int)(sizeof(uint) * fieldCount));
```
SingleAccretion:
```suggestion
```

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -327,7 +327,9 @@ public static TypeDescriptor getTypeDescriptor(IntPtr thisHandle, CORINFO_CLASS_
...
             _allocedMemory.Add(fieldArray);
+            _allocedMemory.Add(fieldSizes);
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1591,6 +1643,22 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
...
 
+void Llvm::buildLocalVarAddr(GenTreeLclVarCommon* lclAddr)
+{
+    unsigned int lclNum = lclAddr->GetLclNum();
+    if (lclAddr->isLclField())
+    {
+        GenTreeLclFld* lclFldAddr = lclAddr->AsLclFld();
+
+        Value* bytePtr = castIfNecessary(m_allocas[lclNum], Type::getInt8PtrTy(_llvmContext));
+        mapGenTreeToValue(lclAddr, _builder.CreateGEP(bytePtr, _builder.getInt16(lclFldAddr->GetLclOffs())));
```
```suggestion
        Value* bytePtr = castIfNecessary(m_allocas[lclNum], Type::getInt8PtrTy(_llvmContext));
        mapGenTreeToValue(lclAddr, _builder.CreateGEP(bytePtr, _builder.getInt16(lclAddr->GetLclOffs())));
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1996,24 +2073,41 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
             case GT_LCL_VAR:
                 indirOper = lclVar->TypeIs(TYP_STRUCT) ? GT_OBJ : GT_IND;
                 break;
+            case GT_LCL_VAR_ADDR:
```
SingleAccretion, line 2074:
```suggestion
            case GT_LCL_VAR_ADDR:
            case GT_LCL_FLD_ADDR:
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1996,24 +2073,41 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
+                indirOper = GT_NONE;
+                break;
+            case GT_LCL_FLD_ADDR:
+                indirOper = GT_NONE;
+                break;
```
SingleAccretion, line 2077:
```suggestion
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2295,7 +2389,7 @@ void Llvm::lowerToShadowStack()
         _currentRange = &LIR::AsRange(_currentBlock);
         for (GenTree* node : CurrentRange())
         {
-            if (node->OperIs(GT_STORE_LCL_VAR, GT_LCL_VAR))
+            if (node->OperIs(GT_STORE_LCL_VAR, GT_LCL_VAR, GT_LCL_VAR_ADDR))
```
```suggestion
            if (node->OperIs(GT_STORE_LCL_VAR, GT_LCL_VAR, GT_LCL_VAR_ADDR, GT_LCL_FLD_ADDR))
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2392,6 +2491,34 @@ void Llvm::PlaceAndConvertShadowStackLocals()
...
+void Llvm::createAllocasForLocalsWithAddrOp()
+{
+    m_allocas = std::vector<Value*>(_compiler->lvaCount, nullptr);
+
+    for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
+    {
+        LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
+
+        // TODO-LLVM LCLBLK - outgoing arg space.  Consider turning off FEATURE_FIXED_OUT_ARGS 
+        if (varDsc->lvType != TYP_LCLBLK && canStoreLocalOnLlvmStack(varDsc) && isLlvmFrameLocal(varDsc))
```
```suggestion
        if ((varDsc->TypeGet() != TYP_LCLBLK) && canStoreLocalOnLlvmStack(varDsc) && isLlvmFrameLocal(varDsc))
```

SingleAccretion (approved):
LGTM!

yowl:
@SingleAccretion THanks for reviewing and getting the quality up.   Think this one was more you than me!

jkotas:
@SingleAccretion @yowl Thank you!

PR #1759, opened 2021-11-30
Merge from dotnet/runtime

jkotas:


jkotas:
@MichalStrehovsky Could you please review the last commit?

src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
```diff
@@ -1185,14 +1185,18 @@ private void ReportInvalidInstruction(ILOpcode opcode)
...
-        private bool IsRuntimeHelpersInitializeArray(MethodDesc method)
+        private bool IsRuntimeHelpersInitializeArrayOrGetSpanData(MethodDesc method)
         {
-            if (method.IsIntrinsic && method.Name == "InitializeArray")
+            if (method.IsIntrinsic)
             {
-                MetadataType owningType = method.OwningType as MetadataType;
-                if (owningType != null)
+                string name = method.Name;
+                if (name == "InitializeArray" || name == "GetSpanData")
```
MichalStrehovsky: I think we need to look for CreateSpan instead - it's the one marked as intrinsic and the one where the ldtoken/call pattern will show up.

src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
```diff
@@ -899,7 +899,7 @@ private void ImportLdToken(int token)
                     {
                         int methodToken = ReadILTokenAt(_currentOffset + 1);
                         var method = (MethodDesc)_methodIL.GetObject(methodToken);
-                        if (IsRuntimeHelpersInitializeArray(method))
+                        if (IsRuntimeHelpersInitializeArrayOrGetSpanData(method))
```
MichalStrehovsky: Could you update the comment above `// First check if this is a ldtoken Field / InitializeArray sequence.`?

MichalStrehovsky (approved):
Looks good otherwise. Thank you!

MichalStrehovsky (approved):
Thank you!

I know it's a bit annoying that scanner needs to be aware. This is to make the asserts happy that check that for unoptimized builds (but with scanner enabled) we scan exactly the same methods as RyuJIT compiles. We don't run in such configuration often. But if we start scanning more stuff than RyuJIT needs to compile, it's a size on disk bug and it's good to have at least something that can check it without false positives.

PR #1760, opened 2021-12-01
Sync cgroup.cpp with runtime repo

MichalStrehovsky:
The first commit forks cgroup.cpp into NativeAOT.
Second commit syncs cgroup.cpp with runtime repo.
Third commit tweaks the forked file (it's the only interesting commit).


jkotas (approved):
Thanks

PR #1761, opened 2021-12-02
Merge from dotnet/runtime

MichalStrehovsky:


MichalStrehovsky:
Oh good, DllImportGenerator binplaces itself somewhere where we consider it part of the framework.

Maybe we can gracefully handle this one.

```
[D:\a\_work\1\s\artifacts\tests\coreclr\windows.x64.Release\Tests\Core_Root\nativeaot\build\BuildFrameworkNativeObjects.proj]
2021-12-02T02:40:47.3941457Z   Internal.TypeSystem.TypeSystemException+FileNotFoundException: Failed to load assembly 'Microsoft.CodeAnalysis.Workspaces'
2021-12-02T02:40:47.4279226Z      at Internal.TypeSystem.ThrowHelper.ThrowFileNotFoundException(ExceptionStringID id, String fileName)
2021-12-02T02:40:47.4842620Z      at Internal.TypeSystem.ResolutionFailure.Throw()
2021-12-02T02:40:47.5516622Z      at Internal.TypeSystem.Ecma.EcmaModule.GetType(EntityHandle handle)
2021-12-02T02:40:47.6595806Z      at Internal.TypeSystem.Ecma.EcmaType.InitializeBaseType()
2021-12-02T02:40:47.7145432Z      at Internal.TypeSystem.TypeDesc.get_IsDelegate()
2021-12-02T02:40:47.8321293Z      at ILCompiler.DependencyAnalysis.TypeMetadataNode.GetStaticDependencies(NodeFactory factory)
2021-12-02T02:40:47.9451587Z      at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.GetStaticDependenciesImpl(DependencyNodeCore`1 node)
2021-12-02T02:40:48.0827007Z      at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.GetStaticDependencies(DependencyNodeCore`1 node)
2021-12-02T02:40:48.1301405Z      at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.ProcessMarkStack()
2021-12-02T02:40:48.2247466Z      at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.ComputeMarkedNodes()
2021-12-02T02:40:48.3161684Z      at ILCompiler.RyuJitCompilation.CompileInternal(String outputFile, ObjectDumper dumper)
2021-12-02T02:40:48.3941818Z      at ILCompiler.Compilation.ILCompiler.ICompilation.Compile(String outputFile, ObjectDumper dumper)
2021-12-02T02:40:48.4579501Z      at ILCompiler.Program.Run(String[] args)
2021-12-02T02:40:48.5047892Z      at ILCompiler.Program.Main(String[] args)
2021-12-02T02:40:48.5821786Z D:\a\_work\1\s\artifacts\tests\coreclr\windows.x64.Release\Tests\Core_Root\nativeaot\build\Microsoft.NETCore.Native.targets(293,5): error MSB3073: The command ""D:\a\_work\1\s\artifacts\tests\coreclr\windows.x64.Release\Tests/Core_Root/nativeaot\tools\ilc" @"D:\a\_work\1\s\artifacts\obj\/NativeAOTFX\Microsoft.Interop.DllImportGenerator.ilc.rsp"" exited with code 1.
```

jkotas:
> Oh good, DllImportGenerator binplaces itself somewhere where we consider it part of the framework.

We are not hitting this problem for other source generators. What makes DllImportGenerator special?

Looks like that this was introduced by https://github.com/dotnet/runtime/pull/62156. cc @jkoritzinsky

jkoritzinsky:
DllImportGenerator is the only generator that we reference within the shared framework other than the CoreLib-only generator (which gets bin placed differently).

jkotas:
I assume that there is filter somewhere that prevents the DllImportGenerator from being included in the shipping shared framework set. Is that right? So the problem is that this filter is not kicking for NativeAOT (or NativeAOT tests at least).

jkotas:
It sounds like that it is just a problem for tests. The DllImportGenerator is not in the .nupkg.

jkoritzinsky:
Yeah we don't include it in the shared framework today. That's a separate step that we still need to do when we get ready to ship it.

jkotas:
> It sounds like that it is just a problem for tests

Looks similar to https://github.com/dotnet/runtime/issues/62372

MichalStrehovsky:
> Looks similar to dotnet/runtime#62372

Thanks, yes, looks similar. I've made a quick local workaround.

I also looked into fixing the compiler to just handle this input because it's just garbage input like any other, but this is the known issue with nested types - a nested type that is fine by itself is nested under something with a damaged base type. Because the type is nested, we need to put the owning type into the dependency graph, but having a type with a broken base hierarchy in the dependency graph requires adding a lot of try/catch in various places. We generally try to avoid broken things in the graph. The ultimate fix would be to rewrite metadata generation in the compiler to be S.R.Metadata-based, not type system based. Then these things are less annoying to deal with.

PR #1763, opened 2021-12-02
[NativeAOT-LLVM] Support for (U)Shorts and struct pointers

yowl:
This PR:

- Add support for short/ushort conversons
- Uses the same name for types as the IL backend (adds assembly prefix)
- Fix pointer arithmetic when not i8* (`MethodTable*` arithmetic was broken)
- Support for struct pointer types in signatures
- Mark shadow stack stores as not heap (removes a lot of unnecessary `RhpAssignRef` )


yowl:
cc @SingleAccretion

src/coreclr/jit/llvm.cpp
```diff
@@ -721,11 +728,18 @@ void Llvm::emitDoNothingCall()
...
+    if (op1Type->isPointerTy() && op2->getType()->isIntegerTy())
     {
-        mapGenTreeToValue(node, _builder.CreateGEP(op1, op2));
+        // the second operator to gep is effectively the number of pointer contained types to add.
+        // For i8*, this is the same as the llvm add operator, but for i32* it would be 4 * add
+        // E.g.
+        // %27 = getelementptr %"[S.P.CoreLib]Internal.Runtime.MethodTable", %"[S.P.CoreLib]Internal.Runtime.MethodTable"* %1, i32 %3
+        // Does not add %3 to the *MethodTable, but adds %3 * sizeof(MethodTable)
+        // Force all pointers to i8* to get the IR GT_ADD semantics
+        mapGenTreeToValue(node, _builder.CreateGEP(castIfNecessary(op1, Type::getInt8PtrTy(_llvmContext)), op2));
```
SingleAccretion: Feels like a losing battle, reconciling the fact pointers in IR are just integers with the fact they are not in LLVM...

Let's call this good enough for now (obviously, we could have `integer + pointer` here, but this is a bit of a best effort code anyway).

Nit: I would shorten the comment to something like `GEPs scale indices, bitcasting to i8* makes them equivalent to the raw offsets we have in IR`.

src/coreclr/jit/llvm.cpp
```diff
@@ -1752,6 +1766,10 @@ void Llvm::populateLlvmArgNums()
         {
             varDsc->lvLlvmArgNum  = nextLlvmArgNum++;
             varDsc->lvCorInfoType = corInfoType;
+            if (classHnd != NO_CLASS_HANDLE && !varTypeIsStruct(varDsc))
+            {
+                varDsc->lvClassHnd = classHnd;
+            }
```
SingleAccretion: We could use `lvClassHnd` to store the exact class always (not just for non structs). Seems that would be a simpler model?

src/coreclr/jit/llvm.cpp
```diff
@@ -1929,6 +1947,7 @@ GenTree* Llvm::createStoreNode(var_types nodeType, GenTree* addr, GenTree* data,
     else
     {
         storeNode = new (_compiler, GT_STOREIND) GenTreeStoreInd(nodeType, addr, data);
+        storeNode->gtFlags |= GTF_IND_TGT_NOT_HEAP;
```
SingleAccretion: The helper is a generic `createStoreNode`, so it seems not right to bake into it the assumptions about the target address' location.

Make a new helper, say `createShadowStackStoreNode` that just adds the `GTF_IND_TGT_NOT_HEAP` (which will be valid for the `STORE_OBJ` too)?

src/coreclr/jit/llvm.cpp
yowl: yes, that's better thanks.

src/coreclr/jit/llvm.cpp
yowl: Thanks, done

SingleAccretion (approved):

jkotas (approved):
Thank you!

PR #1768, opened 2021-12-05
[NativeAOT-LLVM]: Initial support for IsVirtualVtable calls

yowl:
This PR adds the start of support for `CT_INDIRECT` by supporting the most simple cases, when the target is known at compile time, and when the target is a simple indirect call from a variable.

To support this I've added a flag to the call node to indicate if there is a hidden parameter so that the signature can be generated correctly (similar to how `this` is handled)

Added a fix for a store op in between phi nodes which is illegal in LLVM.  I thought I'd already done that but apparently not.

Included the  delegates test as it passes


src/coreclr/jit/indirectcalltransformer.cpp
```diff
@@ -491,7 +494,14 @@ class IndirectCallTransformer
         }
 
     private:
+        // Wasm stores function pointers as indicies in a function table
```
SingleAccretion:
```suggestion
        // Wasm stores function pointers as indices in a function table.
```

src/coreclr/jit/llvm.cpp
```diff
@@ -864,31 +867,82 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
...
+Function* Llvm::getOrCreateLlvmFunction(const char* symbolName, GenTreeCall* call)
+{
+    Function* llvmFunc = _module->getFunction(symbolName);
+
+    if (llvmFunc == nullptr)
+    {
+        // assume ExternalLinkage, if the function is defined in the clrjit module, then it is replaced and an
+        // extern added to the Ilc module
+        llvmFunc =
+            Function::Create(getFunctionTypeForCall(call), Function::ExternalLinkage, 0U, symbolName, _module);
```
SingleAccretion: It would be nice to rename `getFunctionTypeForCall` to `createFunctionTypeForCall`, for symmetry.

src/coreclr/jit/llvm.cpp
```diff
@@ -2091,10 +2170,26 @@ void Llvm::lowerToShadowStack()
...
-
-                failUnsupportedCalls(callNode, calleeSigInfo);
+                // For CT_INDIRECT calls, the callNode->callSig is present,
+                // other times it's not, so get the siginfo from the GenTree call node if possible, else from eeGetMethodSig
+                CORINFO_SIG_INFO* calleeSigInfo = callNode->callSig;
+                if (calleeSigInfo == nullptr)
+                {
+                    CORINFO_SIG_INFO eeCalleeSigInfo;
+                    _compiler->eeGetMethodSig(callNode->gtCallMethHnd, &eeCalleeSigInfo);
+                    calleeSigInfo = &eeCalleeSigInfo;
```
SingleAccretion: It is the case that for indirect calls the method handle is in a union with `gtCallMethHnd`:
```cpp
union {
    CORINFO_METHOD_HANDLE gtCallMethHnd; // CT_USER_FUNC or CT_HELPER
    GenTree*              gtCallAddr;    // CT_INDIRECT
};
```
So it is not clear how does this work to me.

In what cases is a `nullptr` signature seen? We should probably fix those. I note that the signature field is under `#ifdef DEBUG`, so it is not terribly surprising we can see it not preserved.

src/coreclr/jit/llvm.cpp
```diff
@@ -2006,26 +2062,49 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
...
+    {
+        if (callThisArg != nullptr)
+        {
+            hiddenArgNum++;
+        }
+        if (callNode->HasRetBufArg())
+        {
+            hiddenArgNum++;
+        }
+    }
```
SingleAccretion: I actually wonder if we need `gtHasHiddenArgument` and this logic. It seems we will need to handle every arg-not-in-signature here, which seems unfortunate (we will essentially have to duplicate part of logic from `fgMorphArgs`).

We can exploit the fact that in the sorted args order, all non-user (and thus non-signature) args always come first. Subtracting the number of args in the signature from the number of args in the arg info should give us the threshold after which we can start using the signature for argument types.

src/coreclr/jit/llvm.cpp
```diff
@@ -864,31 +867,82 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
...
+    if (call->gtCallType == CT_USER_FUNC)
+    {
+        const char* symbolName = (*_getMangledSymbolName)(_thisPtr, call->gtEntryPoint.handle);
 
-    std::vector<llvm::Value*> argVec = std::vector<llvm::Value*>();
+        (*_addCodeReloc)(_thisPtr, call->gtEntryPoint.handle);
+        Function* llvmFunc = getOrCreateLlvmFunction(symbolName, call);
+        llvmFuncCallee = llvmFunc;
+    }
+    else if (call->gtCallType == CT_INDIRECT)
```
```suggestion
    else if (call->gtCallType == CT_INDIRECT)
```
I would drop the `else if`, leaving only the `else`, and assert under it that the call is indirect.

This captures the idea that this method only builds 2 call types, and will only ever do that (well, it could eventually morph into just `buildCall`, but that'll be a different function).

src/coreclr/jit/llvm.cpp
```diff
@@ -864,31 +867,82 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
...
+
+                const char* methodName = _getMangledSymbolNameFromHelperTarget(_thisPtr, methodHandle);
+                if (methodName == nullptr) // TODO-LLVM: abstract or virtual call
+                {
+                    failFunctionCompilation();
+                }
+
+                (*_addCodeReloc)(_thisPtr, methodHandle);
+
+                llvmFuncCallee = { functionType, getOrCreateLlvmFunction(methodName, call) };
```
SingleAccretion: We are effectively doing "devirtualization for indirect calls" here, manually.

Why?

Is the issue that we need to call `(*_addCodeReloc)(_thisPtr, methodHandle);`? It does not seem right to have it be "pulled" from under the call, it should instead be registered at the logical `ldftn` point, when building the the icon node.

yowl: Thanks, it's good that  you've questioned this, as it gives something concrete to talk about.  I didn't understand how to do the virtual calls so restricted myself to those calls that could be determined at runtime.  The `addCodeReloc` is there to make sure we get the target method compiled.  Happy to expand the PR to cover virtual calls but I didn't see how to make the slot lookup work, and I suspected I was missing something.  How do the other Ilc backends do this I wonder?

yowl: @Jkotas @MichalStrehovsky  Maybe I can pull one of you in here please ?

SingleAccretion: @yowl let me test my understanding here. Is the following true?

> `addCodeReloc` is only required for helpers. User methods are fine as they have been added as dependencies by the ILC's front-end.

yowl: Ummm.... If I delete the call to `_addCodeReloc` , then it works.  Oh dear, sorry about that.  I still could use some help on the slot lookup side though.  Let me get some IR for that.

yowl: Lets take https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Debug.cs#L137-L138
The IR for this is
```cs

------------ BB02 [000..00F) (return), preds={BB01} succs={}
               [000017] ------------                 IL_OFFSET void   IL offset: 0x0
N001 (  1,  4) [000006] H-----------         t6 =    CNS_INT(h) int    0x420058 ftn
                                                  /--*  t6     int
N003 (  5,  7) [000008] DA----------              *  STORE_LCL_VAR int    V02 tmp1         d:1
N001 ( 14,  2) [000001] H-CXG-------         t1 =    CALL help byref  HELPER.CORINFO_HELP_READYTORUN_STATIC_BASE
N002 (  1,  1) [000002] ------------         t2 =    CNS_INT   int    4 Fseq[s_provider]
                                                  /--*  t1     byref
                                                  +--*  t2     int
N003 ( 15,  3) [000003] ---XG--N----         t3 = *  ADD       byref
                                                  /--*  t3     byref
N004 ( 17,  5) [000004] V--XGO-N----         t4 = *  IND       ref
               [000020] ------------        t20 =    CNS_INT   int    4
               [000021] ------------        t21 =    LCL_VAR   int    V04 tmp3         u:1
                                                  /--*  t21    int
                                                  +--*  t20    int
               [000022] ------------        t22 = *  ADD       int
                                                  /--*  t22    int
                                                  +--*  t4     ref
N006 ( 21,  8) [000014] -A-XGO----L-              *  STOREIND  ref
               [000023] ------------        t23 =    CNS_INT   int    4
               [000024] ------------        t24 =    LCL_VAR   int    V04 tmp3         u:1
                                                  /--*  t24    int
                                                  +--*  t23    int
               [000025] ------------        t25 = *  ADD       int
                                                  /--*  t25    int
N008 (  3,  2) [000015] ------------        t15 = *  IND       ref
               [000026] ------------        t26 =    CNS_INT   int    0
               [000027] ------------        t27 =    LCL_VAR   int    V04 tmp3         u:1
                                                  /--*  t27    int
                                                  +--*  t26    int
               [000028] ------------        t28 = *  ADD       int
                                                  /--*  t28    int
N009 (  3,  2) [000005] ------------         t5 = *  IND       ref
N010 (  3,  2) [000009] ------------         t9 =    LCL_VAR   int    V02 tmp1         u:1 (last use)
               [000029] ------------        t29 =    LCL_VAR   int    V04 tmp3         u:1
               [000030] ------------        t30 =    CNS_INT   int    8
                                                  /--*  t29    int
                                                  +--*  t30    int
               [000031] ------------        t31 = *  ADD       int
                                                  /--*  t31    int
               [000032] ------------        t32 = *  PUTARG_TYPE int
               [000033] ------------        t33 =    LCL_VAR   int    V04 tmp3         u:1
               [000034] ------------        t34 =    CNS_INT   int    8
                                                  /--*  t33    int
                                                  +--*  t34    int
               [000035] ------------        t35 = *  ADD       int
                                                  /--*  t35    int
                                                  +--*  t15    ref
               [000036] ------------              *  STOREIND  ref
               [000037] ------------        t37 =    LCL_VAR   int    V04 tmp3         u:1 (last use)
               [000038] ------------        t38 =    CNS_INT   int    12
                                                  /--*  t37    int
                                                  +--*  t38    int
               [000039] ------------        t39 = *  ADD       int
                                                  /--*  t39    int
                                                  +--*  t5     ref
               [000040] ------------              *  STOREIND  ref
                                                  /--*  t32    int    arg-1
                                                  +--*  t9     int    calli tgt
N011 ( 47, 19) [000010] --CXGO------              *  CALL ind nullcheck void
               [000018] ------------                 IL_OFFSET void   IL offset: 0xd
               [000019] ------------                 IL_OFFSET void   IL offset: 0xe
N001 (  0,  0) [000012] ------------                 RETURN    void
```
`t9` is the target, coming from `t9 =    LCL_VAR   int    V02 tmp1         u:1 (last use)` and is the method handle for https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/libraries/System.Private.CoreLib/src/System/Diagnostics/DebugProvider.cs#L66  

To resolve that, somewhere, and with the IL->LLVM compiler, this would be in the IlToLLVMImporter, this needs to be a look up into the vtable for the runtime type, i.e. `s_provider`/`this`.  Easier to see before LLVM lowering maybe:
```
N001 (  0,  0) [000000] ------------                 NOP       void

------------ BB02 [000..00F) (return), preds={BB01} succs={}
               [000017] ------------                 IL_OFFSET void   IL offset: 0x0
N001 (  1,  4) [000006] H-----------         t6 =    CNS_INT(h) int    0x420058 ftn
                                                  /--*  t6     int
N003 (  5,  7) [000008] DA----------              *  STORE_LCL_VAR int    V02 tmp1
N001 ( 14,  2) [000001] H-CXG-------         t1 =    CALL help byref  HELPER.CORINFO_HELP_READYTORUN_STATIC_BASE
N002 (  1,  1) [000002] ------------         t2 =    CNS_INT   int    4 Fseq[s_provider]
                                                  /--*  t1     byref
                                                  +--*  t2     int
N003 ( 15,  3) [000003] ---XG--N----         t3 = *  ADD       byref
                                                  /--*  t3     byref
N004 ( 17,  5) [000004] V--XGO-N----         t4 = *  IND       ref
                                                  /--*  t4     ref
N006 ( 21,  8) [000014] DA-XGO----L-              *  STORE_LCL_VAR ref    V03 tmp2
N008 (  3,  2) [000015] ------------        t15 =    LCL_VAR   ref    V03 tmp2
N009 (  3,  2) [000005] ------------         t5 =    LCL_VAR   ref    V00 arg0
N010 (  3,  2) [000009] ------------         t9 =    LCL_VAR   int    V02 tmp1
                                                  /--*  t15    ref    this in rcx
                                                  +--*  t5     ref    arg1 in rdx
                                                  +--*  t9     int    calli tgt
N011 ( 47, 19) [000010] --CXGO------              *  CALL ind nullcheck void
               [000018] ------------                 IL_OFFSET void   IL offset: 0xd
N001 (  1,  1) [000011] ------------                 NO_OP     void
               [000019] ------------                 IL_OFFSET void   IL offset: 0xe
N001 (  0,  0) [000012] ------------                 RETURN    void
```
I could attempt that as part of this PR if you like?  Basically find the symbol for the slot, then look that up in the EE type - these LLVM steps
```llvm
 %LoadAddressOfSymbolNode2 = load i32*, i32** @__getslot__S_P_CoreLib_System_Diagnostics_DebugProvider__WriteLine___SYMBOL
%WriteLine_slot = load i32, i32* %LoadAddressOfSymbolNode2
.
.
  %CastPtrthis = bitcast i8* %LoadField_s_provider to void (i8*)***,
  %ldEEType = load void (i8*)**, void (i8*)*** %CastPtrthis, align 4, !dbg !446704
  %__getslot__ = getelementptr void (i8*)*, void (i8*)** %ldEEType, i32 %WriteLine_slot
  %ld__getslot__ = load void (i8*)*, void (i8*)** %__getslot__
 ```
My question for Jan or Michael would be is there something in implementation for other backends that helps us?

Comment (SingleAccretion) [reply src/coreclr/jit/llvm.cpp]: > I could attempt that as part of this PR if you like?

I think it is best we scope this PR to what it is right now, i. e. to indirect calls. I suspect the case being shown here will "just work" once that has been done.

I looked around a bit, and it appears the method in question is using an implementation of "virtualness" through the `CORINFO_CALL_CODE_POINTER` path, where we use an indirect invoke for a helper which dispatches to the actual target method:
```
// CORINFO_CALL_CODE_POINTER (shared generic code only) :
//   Indicates that the JIT should do an indirect call to the entrypoint given by address, which may be specified
//   as a runtime lookup by CORINFO_CALL_INFO::codePointerLookup.
//   [targetMethodHandle] will not hold a valid value.
//   This flag may be combined with nullInstanceCheck=TRUE for uses of callvirt on methods whose target method can
//   be resolved at compile-time but whose instantiation can be resolved only through runtime lookup.
```
Apparently this path is taken because, at least for my reproduction code, the layout of the VTable is not fixed.

Virtuals implemented via VTables go though [`fgExpandVirtualVtableCallTarget`](https://github.com/dotnet/runtime/blob/5dc1b4c31312b6778b62ee28ba4a2b0c54f34898/src/coreclr/jit/morph.cpp#L9273) which builds the required indirections, so we won't have to do much if anything except build the LLVM for `gtControlExpr` (hopefully).

Comment (yowl) [reply src/coreclr/jit/llvm.cpp]: Yes I think I see what you are looking at, `CORINFO_CALL_CODE_POINTER` is the `callInfo->kind`, but I'm investigating that as the comment suggests that's only for generics ("shared generic code only")

Comment (yowl) [reply src/coreclr/jit/llvm.cpp]: I wonder if instead of the `LazyVTableSliceProvider` which always returns a `LazilyBuiltVTableSliceNode` and hence never has fixed slots:  https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/VTableSliceNode.cs#L178-L184

I want something more like the `ScannedVTableProvider` which has https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs#L271-L294

Comment (yowl) [reply src/coreclr/jit/llvm.cpp]: I think this might work: a new slice provider and node:
```
    public class LLVMVTableSliceProvider : VTableSliceProvider
    {
        internal override VTableSliceNode GetSlice(TypeDesc type)
        {
            if (type.GetTypeDefinition() is Internal.TypeSystem.Ecma.EcmaType)
            {
                return new LLVMVTableSliceNode(type);
            }
            else
                return new LazilyBuiltVTableSliceNode(type);
        }
    }

    class LLVMVTableSliceNode : VTableSliceNode
    {
        public LLVMVTableSliceNode(TypeDesc type) : base(type)
        {
        }

        public override IReadOnlyList<MethodDesc> Slots => new MethodDesc []{};

        public override bool HasFixedSlots => true;
    }
```
to send it the path of `fgExpandVirtualVtableCallTarget`.  Instead of 
https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/coreclr/jit/morph.cpp#L9539
call something that returns the symbol from a method similar to https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs#L2107-L2113
Then before adding the GT_ADD, add a  GT_CNS_INT (GTF_ICON_CONST_PTR?) and GT_IND 

Is that crazy ?

Comment (MichalStrehovsky) [reply src/coreclr/jit/llvm.cpp]: The compiler doesn't build the vtable the way the CoreCLR runtime does (`foreach (method in type) if method.IsVirtual AddToVTable(method)`). Instead, the AOT compiler tracks which virtual methods are actually used and only once the method is actually called virtually, it's added to the vtable. Thanks to this, we compile a lot less methods. But because of this, we don't know slot numbers until we looked at all the callvirts in the program.

You'll get the ScannedVTableProvider if you add `--scan` to the ILC command line (or add `-O` to enable optimizations - optimizations imply running the scanner). It will run a pass over the whole program that will import all the IL and build a whole program view. Once the scanner is done, RyuJIT will be instructed to do vtable lookups and all the offsets get provided through JitInterfaces. But without the scanner running, we can't tell the slots because we haven't seen all the virtual calls yet. So we ask RyuJIT to do the call through a helper instead. Helper is in assembly and we figure out what assembly instructions to put there once everything is compiled.

You can just add `--scan` to unblock yourself now. But the helper is similar to the one that is used for grabbing static bases of types - the implementation strategy could be similar (I assume we would want to emit some LLVM IR for the helper from the ILCompiler.LLVM side). The emission of those helpers can only happen once the main compilation is done; otherwise they'll run into the same problem with slot not being known yet.

Comment (MichalStrehovsky) [reply src/coreclr/jit/llvm.cpp]: Look for `ReadyToRunHelperId.VirtualCall` in the ILCompiler source tree (e.g. under `src\coreclr\tools\aot\ILCompiler.Compiler\Compiler\DependencyAnalysis\Target_X64\X64ReadyToRunHelperNode.cs`) to see how the helper looks like.

Comment (yowl) [reply src/coreclr/jit/llvm.cpp]: @MichalStrehovsky Thanks, yes I see that running the scanner would be better than my idea, less work and one less indirection.  I think that is the way to unblock this PR in the short term.

I have a problem with the scanner on and
```
GenericDictionaryLookup lookup = _compilation.ComputeGenericLookup(_method, helperId, helperArg);
``` 
returning a lookup with an invalid `HelperId` for the IL->LLVM backend so I'll need to see what's up with that.

Comment (yowl) [reply src/coreclr/jit/llvm.cpp]: > I assume we would want to emit some LLVM IR for the helper from the ILCompiler.LLVM side

I'm not sure we want to emit LLVM from the Ilc side, I'd like to remove LLVMSharp as a dependency

Comment (yowl) [reply src/coreclr/jit/llvm.cpp]: Hmm mix n match between fixed slots for RyuJit and always helper lookups for the IL backend looks awkward.  I guess I'll need to adapt the IL backend for fixed slots (for GetThreadStaticBase for example).  This code assumes there is always a helper:
```cs
        ISymbolNode GetGenericLookupHelperAndAddReference(ReadyToRunHelperId helperId, object helperArg, out LLVMValueRef helper, IEnumerable<LLVMTypeRef> additionalArgs = null)
        {
            ISymbolNode node;
            GenericDictionaryLookup lookup = _compilation.ComputeGenericLookup(_method, helperId, helperArg);

            var retType = helperId == ReadyToRunHelperId.DelegateCtor
                ? LLVMTypeRef.Void
                : LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0);
            var helperArgs = new List<LLVMTypeRef>
            {
                LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0),
                LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0),
            };
            if (additionalArgs != null) helperArgs.AddRange(additionalArgs);
            if (_method.RequiresInstMethodDescArg())
            {
                node = _compilation.NodeFactory.ReadyToRunHelperFromDictionaryLookup(lookup.HelperId, lookup.HelperObject, _method);
                helper = GetOrCreateLLVMFunction(node.GetMangledName(_compilation.NameMangler),
                    LLVMTypeRef.CreateFunction(retType, helperArgs.ToArray(), false));
            }
            else
            {
                Debug.Assert(_method.RequiresInstMethodTableArg() || _method.AcquiresInstMethodTableFromThis());
                node = _compilation.NodeFactory.ReadyToRunHelperFromTypeLookup(lookup.HelperId, lookup.HelperObject, _method.OwningType);
                helper = GetOrCreateLLVMFunction(node.GetMangledName(_compilation.NameMangler),
                    LLVMTypeRef.CreateFunction(retType, helperArgs.ToArray(), false));
            }
```   
Bit surprised as Release configuration used to work in CoreRT, it would have had the scanner on I would have thought.

MichalStrehovsky: > returning a lookup with an invalid HelperId for the IL->LLVM backend so I'll need to see what's up with that

Generic lookups have the same problem as vtable lookups - we don't know the slots until after we looked at the whole program - that's why they need to go through helpers. They also have the same solution - scanner. Once the scanner runs, helper calls are no longer needed - the lookup sequence can be inlined - it's similar to how vtable lookup is done - just a couple indirections. The result from `ComputeGenericLookup` tells what offsets from the generic context to indirect. But you can always just ifdef things out and and unblock yourself (e.g. make it always take the helper route).

> Bit surprised as Release configuration used to work in CoreRT, it would have had the scanner on I would have thought.

We used to have the scanner disabled on WASM and CppCodegen: https://github.com/dotnet/corert/blob/c6af4cfc8b625851b91823d9be746c4f7abdc667/src/ILCompiler/src/Program.cs#L652-L658 because it was easier to disable than adding support for the inlined lookups.

yowl:
@MichalStrehovsky Thanks, looking at https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs#L120-L138 and https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/coreclr/jit/importer.cpp#L2230 the steps from the result of `ComputeGenericLookup` are:

- For each indirection, indirect (starting from the context) 
- add the offset.
- After repeating `pRuntimeLookup->indirections` number of times, indirect one more time

Does that sound right?

I'm going to `#if` it out so need to answer this really.  I'll get back to it at some point.

MichalStrehovsky:
> After repeating pRuntimeLookup->indirections number of times, indirect one more time

Not sure that one is needed. It should match the indirections that are done for the ReadyToRunGenericLookupHelper (the LLVMSharp-based WASM backend already does all of this because it needs to generate the corresponding helper; it's just in a different location).

src/coreclr/jit/llvm.cpp
yowl: I can get back to this now.  With the scanner enabled I get different IR
```
------------ BB02 [000..00F) (return), preds={BB01} succs={}
               [000018] ------------                 IL_OFFSET void   IL offset: 0x0
N001 ( 14,  2) [000001] H-CXG-------         t1 =    CALL help byref  HELPER.CORINFO_HELP_READYTORUN_STATIC_BASE
N002 (  1,  1) [000002] ------------         t2 =    CNS_INT   int    4 Fseq[s_provider]
                                                  /--*  t1     byref
                                                  +--*  t2     int
N003 ( 15,  3) [000003] ---XG--N----         t3 = *  ADD       byref
                                                  /--*  t3     byref
N004 ( 17,  5) [000004] V--XGO-N----         t4 = *  IND       ref
                                                  /--*  t4     ref
N006 ( 21,  8) [000010] DA-XGO----L-              *  STORE_LCL_VAR ref    V02 tmp1
N008 (  3,  2) [000011] ------------        t11 =    LCL_VAR   ref    V02 tmp1
N009 (  3,  2) [000005] ------------         t5 =    LCL_VAR   ref    V00 arg0
N010 (  3,  2) [000013] ------------        t13 =    LCL_VAR   ref    V02 tmp1
                                                  /--*  t13    ref
N011 (  6,  4) [000014] #--X--------        t14 = *  IND       int
N012 (  1,  1) [000015] ------------        t15 =    CNS_INT   int    40
                                                  /--*  t14    int
                                                  +--*  t15    int
N013 (  7,  5) [000016] ---X---N----        t16 = *  ADD       int
                                                  /--*  t16    int
N014 (  9,  7) [000017] n--X--------        t17 = *  IND       int
                                                  /--*  t11    ref    this in rcx
                                                  +--*  t5     ref    arg1 in rdx
                                                  +--*  t17    int    control expr
N015 ( 56, 26) [000006] --CXGO------              *  CALLV vt-ind void   System.Diagnostics.DebugProvider.WriteLine
```
I assume this is what we were expecting - instead of the `CALL ind nullcheck void` we now have `CALLV vt-ind ` ?

SingleAccretion: Yes, that looks right to me.

yowl: When we lower the call args, we remove the `gtCallThisArg` as its pushed to the shadowstack.  I'm going to need that aren't I, to get to the method table?

SingleAccretion: >  I'm going to need that aren't I, to get to the method table?

Why? The idea here is that we don't need to do that, the control expression (== function address) has already been computed for us by morph.

yowl: Thanks, just me being slow.

yowl: Ok, cool, this is super simple now it seems, thanks.
image
```
        if (call->IsVirtualVtable())
        {
            FunctionType* functionType = createFunctionTypeForCall(call);

            Value* funcPtr = castIfNecessary(getGenTreeValue(call->gtControlExpr), functionType->getPointerTo());

            llvmFuncCallee = {functionType, funcPtr};
        }
```

src/coreclr/jit/llvm.cpp
yowl: now we are not doing `CT_INDIRECT` but `CALLV` I'm just reverting this.

src/coreclr/jit/llvm.cpp
yowl: You are right, I've removed it.

yowl:
Latest commit addresses remaining feedback.  Thanks for reviewing.

yowl:
Disabling the Delegates test which I had enabled here.  It fails in Release config.

jkotas:
@SingleAccretion Could you please do final review of this change?

SingleAccretion:
Sorry for the delay, will do so tomorrow.

src/coreclr/jit/llvm.cpp
```diff
@@ -518,14 +521,17 @@ FunctionType* Llvm::getFunctionType()
...
+FunctionType* Llvm::createFunctionTypeForCall(GenTreeCall* call)
 {
     llvm::Type* retLlvmType = getLlvmTypeForCorInfoType(call->gtCorInfoType, call->gtRetClsHnd);
 
     std::vector<llvm::Type*> argVec = std::vector<llvm::Type*>();
 
     for (GenTreeCall::Use& use : call->Args())
     {
+        if (use.GetNode() == call->gtControlExpr)
+            continue;
```
SingleAccretion: When is this case hit?

The control expression should never be in the args (in fact, it would violate the "single use" invariant of an SDSU).

src/coreclr/jit/llvm.cpp
```diff
@@ -1026,7 +1058,7 @@ void Llvm::buildCall(GenTree* node)
     {
         buildHelperFuncCall(call);
     }
-    else if (call->gtCallType == CT_USER_FUNC && !call->IsVirtualStub() /* TODO: Virtual stub not implemented */)
+    else if ((call->gtCallType == CT_USER_FUNC || call->gtCallType == CT_INDIRECT) && !call->IsVirtualStub() /* TODO: Virtual stub not implemented */)
```
SingleAccretion: `call->gtCallType == CT_INDIRECT` no longer needed (as it is no longer handled)?

src/coreclr/jit/llvm.cpp
```diff
@@ -2014,26 +2043,39 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
...
+        {
+            failFunctionCompilation();
+        }
         unsigned int   argNum         = curArgTabEntry->argNum;
         OperandArgNum  opAndArg       = {argNum, curArgTabEntry->GetNode()};
         sortedData[argNum]            = opAndArg;
     }
 
-    CORINFO_ARG_LIST_HANDLE sigArgs = calleeSigInfo.args;
+    // calculate hiddenArg number if present
```
```suggestion
    // Relies on the fact all arguments not in the signature come before those that are.
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1988,7 +2017,7 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
     GenTreeCall::Use* callThisArg = callNode->gtCallThisArg;
 
     callNode->ResetArgInfo();
-    callNode->gtCallThisArg = nullptr;
+    //callNode->gtCallThisArg = nullptr;
```
SingleAccretion: Why comment it out?

The way I read the code, it will cause a call node to reference the `this` SDSU twice (e. g. for `VisitOperands`).

(If some dumping code relies on it - we should adjust it)

src/coreclr/jit/llvm.cpp
```diff
@@ -2014,26 +2043,39 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
     for (unsigned i = 0; i < argCount; i++)
     {
         fgArgTabEntry* curArgTabEntry = argTable[i];
+        if (curArgTabEntry->nonStandardArgKind == NonStandardArgKind::VirtualStubCell)
```
SingleAccretion: Cleaner to catch it in `failUnsupportedCalls`?

src/coreclr/jit/llvm.cpp
```diff
@@ -2014,26 +2043,39 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode, CORINFO_SIG_INFO& calle
...
 
     for (OperandArgNum opAndArg : sortedArgs)
     {
         CORINFO_CLASS_HANDLE clsHnd      = NO_CLASS_HANDLE;
         CorInfoType          corInfoType = CORINFO_TYPE_UNDEF;
 
         // "this" not in sigInfo arg list
-        bool isThis = callThisArg != nullptr && opAndArg.argNum == 0 && calleeSigInfo.hasThis();
-        if (!isThis)
+        bool isThis = callThisArg != nullptr && opAndArg.argNum == 0 && calleeSigInfo->hasThis();
```
SingleAccretion, line 2064: I think it would be a little more robust to check against `callThisArg`:

`isThis = callThisArg != nullptr && callThisArg->GetNode() == opAndArg.operand`.

yowl: That is not quite the same.  E.g. for the IR
```
N001 (  0,  0) [000000] ------------                 NOP       void

------------ BB02 [000..00B) -> BB04 (cond), preds={BB01} succs={BB03,BB04}
               [000021] ------------                 IL_OFFSET void   IL offset: 0x0
N001 (  1,  1) [000001] ------------                 NO_OP     void
               [000022] ------------                 IL_OFFSET void   IL offset: 0x1
N002 (  3,  2) [000002] ------------         t2 =    LCL_VAR   byref  V00 this
                                                  /--*  t2     byref  this in rcx
N003 ( 17,  6) [000003] --CXG-------         t3 = *  CALL      int    System.Threading.CancellationToken.get_IsCancellationRequested
                                                  /--*  t3     int
N004 ( 18,  8) [000004] ---XG-------         t4 = *  CAST      int <- bool <- int
                                                  /--*  t4     int
N006 ( 22, 11) [000006] DA-XG-------              *  STORE_LCL_VAR int    V03 tmp1
               [000023] ------------                 IL_OFFSET void   IL offset: 0x7
N001 (  3,  2) [000007] ------------         t7 =    LCL_VAR   int    V03 tmp1
                                                  /--*  t7     int
N002 (  4,  4) [000019] ------------        t19 = *  CAST      int <- bool <- int
                                                  /--*  t19    int
N004 (  8,  7) [000009] DA----------              *  STORE_LCL_VAR int    V01 loc0
               [000024] ------------                 IL_OFFSET void   IL offset: 0x8
N001 (  3,  2) [000010] ------------        t10 =    LCL_VAR   int    V01 loc0
N002 (  1,  1) [000011] ------------        t11 =    CNS_INT   int    0
                                                  /--*  t10    int
                                                  +--*  t11    int
N003 (  5,  4) [000012] J------N----        t12 = *  EQ        int
                                                  /--*  t12    int
N004 (  7,  6) [000013] ------------              *  JTRUE     void

------------ BB03 [00B..012), preds={BB02} succs={BB04}
               [000025] ------------                 IL_OFFSET void   IL offset: 0xb
N002 (  3,  2) [000015] ------------        t15 =    LCL_VAR   byref  V00 this
                                                  /--*  t15    byref  this in rcx
N003 ( 17,  6) [000016] --CXG-------              *  CALL      void   System.Threading.CancellationToken.ThrowOperationCanceledException
               [000026] ------------                 IL_OFFSET void   IL offset: 0x11
N001 (  1,  1) [000017] ------------                 NO_OP     void

------------ BB04 [012..013) (return), preds={BB02,BB03} succs={}
               [000027] ------------                 IL_OFFSET void   IL offset: 0x12
N001 (  0,  0) [000014] ------------                 RETURN    void
```
For the call at tree id 16, `callThisArg->GetNode()` is a `GT_ARGPLACE` whereas `opAndArg.operand` has become `GT_IND` at treed id 15.

SingleAccretion: Ahh, the ever-present late-arg-not-late arg distinction. You could obtain the real node from the arg info but that point I agree it is not worth the trouble.

src/coreclr/jit/llvm.cpp
yowl: Never apparently.  Have deleted it.

src/coreclr/jit/llvm.cpp
yowl: Thanks, deleted.

src/coreclr/jit/llvm.cpp
yowl: mm, Have reinstated it, thanks

src/coreclr/jit/llvm.cpp
yowl: yes, that's better

src/coreclr/jit/llvm.cpp
yowl: ok,thanks

yowl:
The OSX build failed
```
      nativeaot/SmokeTests/Threading/Threading/Threading.sh [FAIL]
        Unhandled Exception: AssertionFailureException: Exception of type 'AssertionFailureException' was thrown.
           at Assert.True(Boolean) + 0x59
           at WaitSubsystemTests.VerifyWaitDuration(WaitHandle, Int32, Int32) + 0xfb
           at WaitSubsystemTests.WaitDurationTest() + 0x51
           at Runner.Main() + 0x81
           at Threading!<BaseAddress>+0x487b9f
           at Threading!<BaseAddress>+0x487c17
        /Users/runner/work/1/s/artifacts/tests/coreclr/OSX.x64.Debug/nativeaot/SmokeTests/Threading/Threading/Threading.sh: line 408: 13719 Abort trap: 6           $LAUNCHER $ExePath "${CLRTestExecutionArguments[@]}"
        
        Return code:      1
```
And 
```
ld : warning : linker symbol '$ld$hide$os10.4$__Unwind_Backtrace' hides a non-existent symbol '__Unwind_Backtrace' [/Users/runner/work/1/s/src/tests/run.proj]
ld : warning : linker symbol '$ld$hide$os10.5$__Unwind_Backtrace' hides a non-existent symbol '__Unwind_Backtrace' [/Users/runner/work/1/s/src/tests/run.proj]
ld : warning : linker symbol '$ld$hide$os10.4$__Unwind_DeleteException' hides a non-existent symbol '__Unwind_DeleteException' [/Users/runner/work/1/s/src/tests/run.proj]
ld : warning : linker symbol '$ld$hide$os10.5$__Unwind_DeleteException' hides a non-existent symbol '__Unwind_DeleteException' [/Users/runner/work/1/s/src/tests/run.proj]

```
Doesn't seem relevant to this PR.  See what else fails...

yowl:
Will retry.

SingleAccretion (approved):
Nice!

jkotas:
@yowl @SingleAccretion Thank you!

PR #1769, opened 2021-12-06
Move new System.Native shim methods out of pal_environment

jkotas:
pal_environment should only contain methods related to env variables to make the iOS specific variant work well.


MichalStrehovsky (approved):
Thank you!

PR #1776, opened 2021-12-09
Fix InvokeMember corner case

MichalStrehovsky:
The change in DefaultBinder is in a shared file and it didn't feel right to move it to dotnet/runtime with a known runtimelab bug.


jkotas (approved):

MichalStrehovsky:
Closing in favor of https://github.com/dotnet/runtime/pull/62891.

PR #1781, opened 2021-12-10
Use OutputRid instead of __DistroRid

MichalStrehovsky:
Apparently __DistroRid in the musl-cross container is linux-arm64.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
+++ b/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
@@ -21,7 +21,7 @@
   <PropertyGroup>
     <SelfContained>true</SelfContained>
     <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj
```diff
@@ -21,7 +21,7 @@
   <PropertyGroup>
     <SelfContained>true</SelfContained>
     <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
-    <RuntimeIdentifier>$(__DistroRid)</RuntimeIdentifier>
+    <RuntimeIdentifier>$(OutputRid)</RuntimeIdentifier>
 
     <!-- DistroRid is injected through environment variables in build scripts. Provide reasonable default
```
AntonLapounov, line 26: Nit: This comment may need updating.

AntonLapounov: Ah, this was reverted.

PR #1796, opened 2021-12-28
Add ILLink.Shared project to NativeAOT

tlakollo:
Add ILLink.Shared project to NativeAOT
Change diagnostics to use ILLink.Shared utilities
Update ReferenceSource


src/coreclr/tools/aot/ILLink.Shared/SharedStrings.resx
```diff
@@ -0,0 +1,213 @@
...
+  </data>
+  <data name="RequiresUnreferencedCodeOnStaticConstructorTitle" xml:space="preserve">
+    <value>The use of 'RequiresUnreferencedCodeAttribute' on static constructors is disallowed since is a method not callable by the user, is only called by the runtime. Placing the attribute directly on the static constructor will have no effect, instead use 'RequiresUnreferencedCodeAttribute' on the type which will handle warning and silencing from the static constructor.</value>
+  </data>
+  <data name="MakeGenericMethodCannotBeStaticallyAnalyzedMessage" xml:space="preserve">
+    <value>Call to '{0}' can not be statically analyzed. It's not possible to guarantee the availability of requirements of the generic method.</value>
+  </data>
+  <data name="PropertyAccessorParameterInLinqExpressionsCannotBeStaticallyDeterminedMessage" xml:space="preserve">
+    <value>Value passed to the '{0}' parameter of method '{1}' cannot be statically determined as a property accessor.</value>
+  </data>
```
tlakollo, line 209: I guess we need a special review on these since they don't exist currently in dotnet/linker, I just choose a name based on what I saw on Microsoft docs but we can choose a more appropriate string if necessary and add it to dotnet/linker too.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -2221,45 +2222,24 @@ public override bool HandleCall(MethodIL callingMethodBody, MethodDesc calledMet
                         if (shouldEnableReflectionWarnings &&
                             calledMethod.HasCustomAttribute("System.Diagnostics.CodeAnalysis", "RequiresUnreferencedCodeAttribute"))
                         {
-                            string attributeMessage = DiagnosticUtilities.GetRequiresUnreferencedCodeAttributeMessage(calledMethod);
+                            string arg1 = MessageFormat.FormatRequiresAttributeMessageArg(DiagnosticUtilities.GetRequiresUnreferencedCodeAttributeMessage(calledMethod));
+                            string message = new DiagnosticString(DiagnosticId.RequiresUnreferencedCode).GetMessage(calledMethod.GetDisplayName(), arg1, "");
```
MichalStrehovsky: Could you add some sort of TODO that we also need to pass the URI? I left the commented out code because I didn't want to deal with it that kind of serves as a TODO.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -2221,45 +2222,24 @@ public override bool HandleCall(MethodIL callingMethodBody, MethodDesc calledMet
...
                         static void LogDynamicCodeWarning(Logger logger, MethodIL callingMethodBody, int offset, MethodDesc calledMethod)
                         {
-                            string attributeMessage = DiagnosticUtilities.GetRequiresDynamicCodeAttributeMessage(calledMethod);
-
-                            if (attributeMessage != null && attributeMessage.Length > 0 && !attributeMessage.EndsWith('.'))
-                                attributeMessage += '.';
-
-                            string message = $"{String.Format(Resources.Strings.IL3050, calledMethod.GetDisplayName())} {attributeMessage}";
+                            string arg1 = MessageFormat.FormatRequiresAttributeMessageArg(DiagnosticUtilities.GetRequiresDynamicCodeAttributeMessage(calledMethod));
+                            string message = new DiagnosticString(DiagnosticId.RequiresDynamicCode).GetMessage(calledMethod.GetDisplayName(), arg1, "");
```
MichalStrehovsky: Same here - some kind of TODO that the `""` is a URI we're supposed to read.

src/coreclr/tools/aot/ILLink.Shared/DiagnosticId.cs
```diff
@@ -0,0 +1,62 @@
+﻿namespace ILLink.Shared
```
MichalStrehovsky: These will need copyright headers. We should probably do those changes in the linker repo and then propagate here. Ideally, files in ILLink.Shared will be just straight copies out of the linker repo with no additions.

src/coreclr/tools/aot/ILLink.Shared/ILLink.Shared.projitems.bak
```diff
@@ -0,0 +1,20 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
MichalStrehovsky: Unnecessary file?

MichalStrehovsky (commented):
Nit: We might want to also reconsider the name of ILLink.Shared. Something along "form factors analysis" might be a better fit for what this is project contains.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Dataflow/ReflectionMethodBodyScanner.cs
tlakollo: Added a method to check for URL

MichalStrehovsky (approved):
Thanks!

Could you please make a pull request against the runtime repo? We don't plan to do integrations from runtimelab to runtime anymore.

You should be able to just run this when on the main branch of the runtime repo:

```
curl -L https://patch-diff.githubusercontent.com/raw/dotnet/runtimelab/pull/1796.diff | git apply
```

MichalStrehovsky:
Merged into runtime/main.

PR #1800, opened 2022-01-06
[NativeAOT-LLVM] Add runtime package for browser-wasm

### LINKED ISSUE #1792: [NativeAOT-LLVM] Can we produce package artifacts
Would it be ok to produce nuget package artifacts for this branch?  Even if they're just in Azure and not published to nuget, it would be helpful?

yowl:
Also add a few instructions which I hope are going to work once the package is available.

Closes #1792

### DIFF
--- a/src/installer/pkg/projects/Microsoft.DotNet.ILCompiler.LLVM/ILCompilerRIDs.props
+++ b/src/installer/pkg/projects/Microsoft.DotNet.ILCompiler.LLVM/ILCompilerRIDs.props
@@ -1,0 +1,12 @@
+<Project>
+  <!-- The list of RIDs supported by ILCompiler -->
+  <ItemGroup>
+    <OfficialBuildRID Include="linux-arm64" Platform="arm64" />
+    <OfficialBuildRID Include="linux-musl-x64" Platform="x64" />
+    <OfficialBuildRID Include="linux-x64" Platform="x64" />
+    <OfficialBuildRID Include="osx-x64" Platform="x64" />
+    <OfficialBuildRID Include="win-arm64" Platform="arm64" />
+    <OfficialBuildRID Include="win-x64" Platform="x64" />
+    <OfficialBuildRID Include="browser-wasm" Platform="wasm" />
+  </ItemGroup>
+</Project>
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -47,13 +47,29 @@ If the compilation succeeds, the native executable will be placed under the `bin
 
 Native AOT toolchain allows targeting ARM64 on an x64 host and vice versa for both Windows and Linux. Cross-OS compilation, such as targeting Linux on a Windows host, is not supported. To target win-arm64 on a Windows x64 host, in addition to the `Microsoft.DotNet.ILCompiler` package reference, also add the `runtime.win-x64.Microsoft.DotNet.ILCompiler` package reference to get the x64-hosted compiler:
 ```xml
-<PackageReference Include="Microsoft.DotNet.ILCompiler; runtime.win-x64.Microsoft.DotNet.ILCompiler" Version="7.0.0-alpha.1.21423.2" />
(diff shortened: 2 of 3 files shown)


yowl:
cc @SingleAccretion

src/installer/pkg/projects/Microsoft.DotNet.ILCompiler.LLVM/ILCompilerRIDs.props
```diff
@@ -0,0 +1,12 @@
+<Project>
+  <!-- The list of RIDs supported by ILCompiler -->
+  <ItemGroup>
+    <OfficialBuildRID Include="linux-arm64" Platform="arm64" />
+    <OfficialBuildRID Include="linux-musl-x64" Platform="x64" />
+    <OfficialBuildRID Include="linux-x64" Platform="x64" />
+    <OfficialBuildRID Include="osx-x64" Platform="x64" />
+    <OfficialBuildRID Include="win-arm64" Platform="arm64" />
+    <OfficialBuildRID Include="win-x64" Platform="x64" />
+    <OfficialBuildRID Include="browser-wasm" Platform="wasm" />
```
SingleAccretion, line 10: I guess this one is required to produce the actual WASM runtime (one won't be running ILC on WASM)?

Would be worth a comment to that effect (`This one is a little special because...`).

yowl:
I did try
```
build nativeaot.packages -c Debug -lc Debug -a wasm -os Browser
```
locally and it built:
```
Microsoft.DotNet.ILCompiler.LLVM -> E:\GitHub\runtimelab\artifacts\packages\Debug\Shipping\runtime.browser-wasm.Microsoft.DotNet.ILCompiler.LLVM.7.0.0-dev.nupkg
```

src/installer/pkg/projects/Microsoft.DotNet.ILCompiler.LLVM/ILCompilerRIDs.props
yowl: Yes, that's right, its the cross architecture publish flow I suppose you call it.  Is it special, it's the same as compiling for linux on Windows isn't it?

SingleAccretion: On second thought, I agree, it is not that special after all. We're just not at the point of self-hosting ILC yet, I suppose :).

SingleAccretion (approved):
I am not really familiar with this publishing infra, but can say at least that last time I fixed the packages themselves, they worked locally.

src/installer/pkg/projects/Microsoft.DotNet.ILCompiler.LLVM/ILCompilerRIDs.props
yowl: Yes, that' would be a project.  Although some parts of this, e.g. emscripten already run under Wasm.  A project for someone with more time on their hands than I have.

jkotas (approved):
Ok, let's see whether the package gets actually published by the official build.

yowl:
Thanks!

yowl:
I think the publishing to https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-experimental has finished, but unfortunately nothing for `runtime.browser-wasm....`

image

yowl:
Is it the `Publish Using Darc` job that should do this?  Are the logs from the official build that might have some clues online, or just to MS employees?

jkotas:
The official builds (including logs) do not have anonymous access. I will be happy to share the logs with you as needed.

I think the problem is that the official build is only building this: https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/eng/pipelines/runtimelab.yml#L100-L105 . I think we are missing wasm in this list.

yowl:
I see it's just the Release build.  I've not tested that in ages.  I suppose I'll need to resurrect that first.

Scott
________________________________

PR #1801, opened 2022-01-07
[NativeAOT-LLVM]: Build Release configuration

yowl:
This PR enables the Release configuration on Windows.  To do so it

- Fix handling for params such as `--codegenopt:NgenDump=`
- Remove -O for Release build as it fails - will have to come back to this
- Remove branch merging due to struct return Type see #1813 
- Add LLVM verifyModule to Debug build to aid debugging
- Change retrieval of method name for DWARF to ICorInfoImpl call that works in Debug and Release
- Add temporary inlining policy that returns false - I've not forgotten about getting the scanner working, but I've hit a problem with generics so this is just to get the Release package out


yowl:
@SingleAccretion

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -1,3 +1,7 @@
+param (
+    [string]$buildConfig = "Release"
```
SingleAccretion: Do we want this defaulted?

Seems it'd be easier to debug bad parameter failures if its required (but it's not a big deal).

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -252,7 +252,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <IlcArg Condition="$(TargetArchitecture) != ''" Include="--targetarch:$(TargetArchitecture)" />
       <IlcArg Condition="'$(TargetArchitecture)' == 'wasm'" Include="--targetos:wasm" />
       <IlcArg Condition="$(IlcMultiModule) == 'true'" Include="--multifile" />
-      <IlcArg Condition="$(Optimize) == 'true'" Include="-O" />
+      <!-- TODO-LLVM: when optimize is on, the clrjit is asserting so turn off to get the CI working -->
+      <IlcArg Condition="$(Optimize) == 'true' and '$(TargetArchitecture)' != 'wasm'" Include="-O" />
```
SingleAccretion: Hmm, maybe this would be a little more visible to turn `Optimize` off at the top of the file (where `<NativeCodeGen>llvm</NativeCodeGen>` is defined), it seems very easy to miss and spend time wondering "why are optimizations off".

eng/pipelines/runtimelab/install-llvm.cmd
```diff
@@ -3,14 +3,17 @@ cd /D "%1"
 
 set RepoRoot=%2\
 
+set buildConfig=%3
+echo Building LLVM with config %buildConfig%
```
SingleAccretion:
```suggestion
echo Building LLVM with config: %buildConfig%
```

SingleAccretion (approved):
The return merging code failure is suspect, it should not really depend on platform quirks to work (feels like a latent upstream bug). It is also not optional (the "add internal blocks" phase depends on it being run), so it will need to be fixed at some point.

But that is separate, this change LGTM. Very nice to see more CI configurations enabled!

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: Thanks, yes that's better

eng/pipelines/runtimelab/install-llvm.ps1
yowl: Removed.

yowl:
I doubled the timeout (I hope), it seems to fail more often than succeed.  It didn't make any difference, still 900000 milliseconds, so I reverted that.

jkotas (approved):
Thanks

PR #1814, opened 2022-01-09
[NativeAOT-LLVM] Wasm official job changes

yowl:
@jkotas 

I don't know if you want to try this, I admit this is a shot in the dark.  I've removed the Browser_wasm lines in case we are looking at some mix of the mono and coreclr flavors, and added some Browser_wasm_win lines to try to stop it picking the ubuntu image.

### DIFF
--- a/eng/pipelines/common/xplat-setup.yml
+++ b/eng/pipelines/common/xplat-setup.yml
@@ -115,7 +115,7 @@ jobs:
           demands: ImageOverride -equals Build.Ubuntu.1804.Amd64.Open
 
         # Official Build Linux Pool
-        ${{ if and(or(in(parameters.osGroup, 'Linux', 'FreeBSD', 'Browser', 'Android'), eq(parameters.hostedOs, 'Linux')), ne(variables['System.TeamProject'], 'public')) }}:
+        ${{ if and(or(in(parameters.osGroup, 'Linux', 'FreeBSD', 'Android'), eq(parameters.hostedOs, 'Linux')), ne(variables['System.TeamProject'], 'public')) }}:
           name: NetCore1ESPool-Internal
           demands: ImageOverride -equals Build.Ubuntu.1804.Amd64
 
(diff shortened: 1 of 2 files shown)


jkotas:
I do not think that changing runtime-official is going to help. This file is not used for runtimelab as all.

The runtimelab official builds are configured to use eng/pipelines/runtimelab.yml . So the problem will be in one of the files transitively included by runtimelab.yml.

I have looked around in the logs for more clues. I think the problem happens very early when the wasm job is starting and the system picks up machine pool and OS image to use:
```
2022-01-08T17:55:28.9860198Z ##[section]Starting: Initialize job
2022-01-08T17:55:28.9863139Z Agent name: 'NetCore1ESPool-Internal 16'
2022-01-08T17:55:28.9863795Z Agent machine name: '531d1268c000000'
2022-01-08T17:55:28.9864064Z Current agent version: '2.195.2'
2022-01-08T17:55:28.9919655Z ##[group]1ES Hosted Pool
2022-01-08T17:55:28.9920162Z SKU: Standard_D4_v3
2022-01-08T17:55:28.9920376Z Image: Build.Ubuntu.1804.Amd64 <- this is wrong. It needs to be windows
```

I think that the problem is here: https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/eng/pipelines/common/xplat-setup.yml#L112-L134 . Notice that the conditions for official and non-official builds differ by browser. I think it may be the reason why the system may be picking wrong machine to use to official builds. Could you please update this PR to fix these conditions instead?

yowl:
Thanks for looking at this (and on a Sunday).  Have reverted the runtime-official change and simply removed the forcing of `osGroup` == `Browser` to Ubuntu.  Looking at the "parent" yml, `platform-matrix.yml`, this should just work out as `Browser_wasm_win` goes to `hostedOs` of `Windows` and `Browser_wasm` will still go the same route courtesy of it having `hostedOs: Linux` (although `runtimelab.yml` does not have a platform line for `Browser_wasm` anyway.

jkotas (approved):

PR #1818, opened 2022-01-12
[NativeAOT-LLVM] Include clr.wasmjit subset when building on windows for x64

yowl:
This PR as per the title changes the subset of `DefaultNativeAotSubsets` to include `clr.wasmjit` when building on windows for x64.  This will get the clrjit for wasm built when doing a `build nativeaot` and gets it in the runtime package.  Hopefully this will fix the official job packaging.

cc @SingleAccretion


jkotas (approved):
Let's see whether we are going to get a good package with these fixes.

yowl:
I dont think its published successfully, was there an error?

PR #1822, opened 2022-01-15
[NativeAOT-LLVM] Remove erroneous extra indirection loading field address from static base.

yowl:
This PR removes an indirection when loading the field address from the static base.  Nothing blew up before, but I noticed it playing with zerosharp and w4 (https://wasm4.org/).  The game crashed accessing invalid memory (it only has 64KB so things are more easily noticed)

cc @SingleAccretion

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -4618,7 +4618,7 @@ namespace Internal.IL
                             {
                                 node = _compilation.NodeFactory.TypeGCStaticsSymbol(owningType);
                                 LLVMValueRef basePtrPtr = LoadAddressOfSymbolNode(node);
-                                staticBase = _builder.BuildLoad(_builder.BuildPointerCast(basePtrPtr, LLVMTypeRef.CreatePointer(LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0), 0), "castBasePtr"), "base");
+                                staticBase = _builder.BuildPointerCast(basePtrPtr, LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0), "castBase");
                             }
                         }
                         else


yowl:
@MichalStrehovsky I'm not understanding this clearly.  When you build with `--preinitstatics` you get a symbol such as `@__GCStaticBase_HelloWasm_Program__PreInitData` which points to a frozen segment which would include the data for something like this array, ie. there would be a slice with the eetype, the length, then the 2 bytes.  

```cs
static readonly byte[] DirectionSprites = new byte[]
{
        0b11111111,
        0b11111111,
};
```
However when I compare the generated LLVM code to get the field address for this
```cs
        fixed (byte* spriteAddress = &(DirectionSprites[0]))
        {
```
The LLVM is the same with or without `preinitstatics` - basically indirect the static base twice, add 4 (which I suppose is the field offset for `DirectionSprites`,  then add 8 (skip the eetype and array length). Without `preinitstatics` there is the class constructor call, but that seems to be the only difference.

`@__GCStaticBase_HelloWasm_Program__PreInitData` is added to the gc static base:
```
@__GCStaticBase_HelloWasm_Program = global [2 x i32*] 
	[i32* bitcast (i8* getelementptr (i8, i8* bitcast ([6 x i32*]* @__GCStaticEEType_011111___REALBASE to i8*), i32 15) to i32*), 
	 i32* bitcast ([5 x i32*]* @__GCStaticBase_HelloWasm_Program__PreInitData to i32*)]
```
So shouldn't someting be using that?

yowl:
So what I've done with the change, which is clearly not right, is to make that "add 4" I mention above, add 4 to `@__GCStaticBase_HelloWasm_Program` and get into the `@__GCStaticBase_HelloWasm_Program__PreInitData`.  A happy coincidence that the field offset is the same as the offest into `@__GCStaticBase_HelloWasm_Program` to get the preinit symbol.

yowl:
Looking at the memory at runtime, there seems to be two routes to the field address.  The one taken by the actual code which works fine.  And another which I think is through the preinit symbol.   I cant get my head round how the first works but it does. I'm beginning to think that the problem is just in the linking, will try linking with more memory next.

yowl:
Abandoning, I don't why the zerosharp code fails.

MichalStrehovsky:
The bases of GC static fields are allocated on the GC heap (as opposed to bases of non-GC static fields that are simply stored in the data segment). There's this code on the startup code path that checks if we have any preinitialized data and if we do, we copy it to the newly allocated object:

https://github.com/dotnet/runtimelab/blob/19bbf57b23bbc20de850eddc26a767da2d63e9d3/src/coreclr/nativeaot/Common/src/Internal/Runtime/CompilerHelpers/StartupCodeHelpers.cs#L228-L257

ZeroSharp doesn't really have any startup code, so this is missing. I guess for ZeroSharp the only way to get to preinitialized data is through the `ReadOnlySpan<byte> s = new byte[] { 1,2,3,4 }` trick (that won't actually allocate an array in IL).

yowl:
@MichalStrehovsky Thankyou very much, that was driving me nuts.

Issue #1783, opened 2021-12-11
[NativeAOT-LLVM] Question: ILScanner and Generics test


yowl:
I'm enabling the scanner for the IL backend and when compiling Generics.cs I'm hitting the assert https://github.com/dotnet/runtimelab/blob/dd8c43a825c41ccfc21a8ceefa79e232adbe02a1/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs#L316-L325
with stacktrace
```
>	ILCompiler.Compiler.dll!ILCompiler.ILScanResults.ScannedDictionaryLayoutProvider.GetPrecomputedLayout(Internal.TypeSystem.TypeSystemEntity methodOrType) Line 325	C#
 	ILCompiler.Compiler.dll!ILCompiler.ILScanResults.ScannedDictionaryLayoutProvider.GetLayout(Internal.TypeSystem.TypeSystemEntity methodOrType) Line 350	C#
 	[External Code]	
 	ILCompiler.Compiler.dll!ILCompiler.DependencyAnalysis.NodeFactory.NodeCache<Internal.TypeSystem.TypeSystemEntity, ILCompiler.DependencyAnalysis.DictionaryLayoutNode>.GetOrAdd(Internal.TypeSystem.TypeSystemEntity key) Line 156	C#
 	ILCompiler.Compiler.dll!ILCompiler.DependencyAnalysis.NodeFactory.GenericDictionaryLayout(Internal.TypeSystem.TypeSystemEntity methodOrType) Line 785	C#
 	ILCompiler.Compiler.dll!ILCompiler.DependencyAnalysis.MethodGenericDictionaryNode.GetDictionaryLayout(ILCompiler.DependencyAnalysis.NodeFactory factory) Line 248	C#
 	ILCompiler.Compiler.dll!ILCompiler.DependencyAnalysis.MethodGenericDictionaryNode.ComputeNonRelocationBasedDependencies(ILCompiler.DependencyAnalysis.NodeFactory factory) Line 206	C#
 	ILCompiler.Compiler.dll!ILCompiler.DependencyAnalysis.ObjectNode.GetStaticDependencies(ILCompiler.DependencyAnalysis.NodeFactory factory) Line 59	C#
[...]

PR #1823, opened 2022-01-16
[NativeAOT-LLVM] Make boxing on reference types a noop

yowl:
This PR addresses the first point from https://github.com/dotnet/runtimelab/issues/1783#issuecomment-1011488581
To avoid adding depenencies for boxing reference types this becomes a NO-OP to bring the compilation in line with the scanner.


yowl:
cc @SingleAccretion

jkotas (approved):

PR #1824, opened 2022-01-17
[NativeAOT-LLVM] For TARGET_WASM do not expand to GT_INDEX_ADDR

yowl:
This PR adds support in the LLVM clrjit for the `GT_INDEX_ADDR` instruction.  Do I need some todo's in here for 64 bit Wasm on either the index, the element offset (unlikely I suppose, its just the type + length I think), or the element size?

@cc SingleAccretion.

Thanks,

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1122,15 +1122,6 @@ void Llvm::buildInd(GenTree* node, Value* ptr)
                                      getLlvmTypeForVarType(node->TypeGet())->getPointerTo())));
 }
 
-void Llvm::buildIndexAddr(GenTreeIndexAddr* indexAddr)
-{
-    Value* firstElementAddr =
-        _builder.CreateGEP(getGenTreeValue(indexAddr->gtGetOp1()), _builder.getInt32(indexAddr->gtElemOffset));
-    Value* offset = _builder.CreateMul(_builder.getInt32(indexAddr->gtElemSize), getGenTreeValue(indexAddr->Index()));
-
-    mapGenTreeToValue(indexAddr, _builder.CreateGEP(firstElementAddr, offset));
-}
-
 Value* Llvm::buildJTrue(GenTree* node, Value* opValue)
 {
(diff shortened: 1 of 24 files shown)


yowl:
cc @SingleAccretion  (had @ in the wrong place)

src/coreclr/jit/llvm.cpp
```diff
@@ -1122,6 +1122,15 @@ void Llvm::buildInd(GenTree* node, Value* ptr)
...
 
+void Llvm::buildIndexAddr(GenTreeIndexAddr* indexAddr)
+{
+    Value* firstElementAddr =
+        _builder.CreateGEP(getGenTreeValue(indexAddr->gtGetOp1()), _builder.getInt32(indexAddr->gtElemOffset));
+    Value* offset = _builder.CreateMul(_builder.getInt32(indexAddr->gtElemSize), getGenTreeValue(indexAddr->Index()));
+
+    mapGenTreeToValue(indexAddr, _builder.CreateGEP(firstElementAddr, offset));
+}
+
```
SingleAccretion: Hmm, this won't work as-is, `INDEX_ADDR` is effectively bounds-checked (and we're missing the bounds check).

I think it would be better to just always fully expand `INDEX` in morph for LLVM, `INDEX_ADDR`'s only purpose is to increase throughput under MinOpts, and I think that sort of tuning can wait (and is probably not worthwhile anyway).

yowl: I see you've got a PR in runtime around this area.  Would you prefer this was left until that was accepted and merged here?

SingleAccretion: I don't think there is a need to block this change on my changes in the runtime.

The fix here should be very simple: `#ifdef`-out the logic that creates `INDEX_ADDR`s for `TARGET_WASM` in `fgMorphArrayIndex`.

yowl: thanks, have #if'ed it out.

SingleAccretion (approved):

PR #1825, opened 2022-01-18
[NativeAOT-LLVM] Add support for GT_STORE_OBJ and GT_OBJ

### LINKED ISSUE #1757: [NativeAOT-LLVM]: GT_STORE_OBJ implementation
We create this op but we don't actually have an implementation for it.  

The current IL code is https://github.com/dotnet/runtimelab/blob/3d515fbe1daa73e3bb765b626f3028589f26d816/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs#L1119-L1160

Is the same approach correct for the RyuJIT backend?

cc @SingleAccretion

Thanks,

yowl:
This PR add support for GT_STORE_OBJ and GT_OBJ, including support for the zero initializer when storing a 0 constant.

It implements a call back to Ilc to get the structure layout in one call, and then implements a recursive store similar to what was done in Ilc.

Closes #1757 

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1442,13 +1442,12 @@ void Llvm::buildStoreInd(GenTreeStoreInd* storeIndOp)
 // Copies endOffset - startOffset bytes, endOffset is exclusive.
 unsigned Llvm::buildMemCpy(Value* baseAddress, unsigned startOffset, unsigned endOffset, Value* srcAddress)
 {
-    Value* paddingDestAddress = _builder.CreateGEP(baseAddress, _builder.getInt32(startOffset));
+    Value* destAddress = _builder.CreateGEP(baseAddress, _builder.getInt32(startOffset));
+    unsigned size = endOffset - startOffset;
 
-    const unsigned paddingSize = endOffset - startOffset;
-    // copy the bytes in the padding
-    _builder.CreateMemCpy(paddingDestAddress, llvm::Align(), srcAddress, llvm::Align(), paddingSize);
+    _builder.CreateMemCpy(destAddress, llvm::Align(), srcAddress, llvm::Align(), size);
 
-    return paddingSize;
+    return size;
 }
 
 void Llvm::storeObjAtAddress(Value* baseAddress, Value* data, StructDesc* structDesc)
@@ -1488,7 +1487,7 @@ void Llvm::storeObjAtAddress(Value* baseAddress, Value* data, StructDesc* struct
             // recurse into struct
             storeObjAtAddress(address, fieldData, getStructDesc(fieldDesc->getClassHandle()));
 
-            bytesStored += fieldData->getType()->getPrimitiveSizeInBits() / 8;
+            bytesStored += fieldData->getType()->getPrimitiveSizeInBits() / BITS_PER_BYTE;
         }
         else
         {
@@ -1508,7 +1507,7 @@ void Llvm::storeObjAtAddress(Value* baseAddress, Value* data, StructDesc* struct
         }
     }
 
-    unsigned llvmStructSize = data->getType()->getPrimitiveSizeInBits() / 8;
+    unsigned llvmStructSize = data->getType()->getPrimitiveSizeInBits() / BITS_PER_BYTE;
     if (structDesc->isExplicitLayout() && llvmStructSize > bytesStored)
     {
         Value* srcAddress = _builder.CreateGEP(baseAddress, _builder.getInt32(bytesStored));
@@ -1539,7 +1538,7 @@ void Llvm::buildStoreObj(GenTreeIndir* indirOp)
 
     CORINFO_CLASS_HANDLE structClsHnd  = structLayout->GetClassHandle();
     StructDesc*          structDesc    = getStructDesc(structClsHnd);
-    bool                 targetNotHeap = (indirOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0;
+    bool                 targetNotHeap = ((indirOp->gtFlags & GTF_IND_TGT_NOT_HEAP) != 0) || genTreeObj->Addr()->OperIsLocalAddr();
 
     Value* dataValue = getGenTreeValue(genTreeObj->Data());
     if (targetNotHeap)
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -41,6 +41,7 @@ static const uint32_t(*_structIsWrappedPrimitive)(void*, CORINFO_CLASS_STRUCT_*,
 static const uint32_t(*_padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned);
 static const CorInfoTypeWithMod(*_getArgTypeIncludingParameterized)(void*, CORINFO_SIG_INFO*, CORINFO_ARG_LIST_HANDLE, CORINFO_CLASS_HANDLE*);
 static const CorInfoTypeWithMod(*_getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*);
+static const uint32_t (*_getObjectLayout)(void*, CORINFO_CLASS_HANDLE, FieldStoreLayout**);
 
 static char*                              _outputFileName;
 static Function*                          _doNothingFunction;
@@ -64,7 +65,9 @@ extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
                                                 const uint32_t(*structIsWrappedPrimitive)(void*, CORINFO_CLASS_STRUCT_*, CorInfoType),
                                                 const uint32_t(*padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned),
                                                 const CorInfoTypeWithMod(*getArgTypeIncludingParameterized)(void*, CORINFO_SIG_INFO*, CORINFO_ARG_LIST_HANDLE, CORINFO_CLASS_HANDLE*),
-                                                const CorInfoTypeWithMod(*getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*))
+                                                const CorInfoTypeWithMod(*getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*),
+                                                const uint32_t (*getObjectLayout)(void*, CORINFO_CLASS_HANDLE, FieldStoreLayout**)
+)
 {
     _thisPtr = thisPtr;
     _getMangledMethodName         = getMangledMethodNamePtr;
@@ -80,6 +83,7 @@ extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
     _padOffset = padOffset;
     _getArgTypeIncludingParameterized = getArgTypeIncludingParameterized;
     _getParameterType = getParameterType;
+    _getObjectLayout = getObjectLayout;
 
     if (_module == nullptr) // registerLlvmCallbacks is called for each method to compile, but must only created the module once.  Better perhaps to split this into 2 calls.
     {
@@ -131,7 +135,7 @@ bool structIsWrappedPrimitive(CORINFO_CLASS_HANDLE classHnd, CorInfoType primiti
     return (*_structIsWrappedPrimitive)(_thisPtr, classHnd, primitiveType);
 }
 
-void addPaddingFields(unsigned paddingSize, std::vector<Type*> llvmFields)
+void addPaddingFields(unsigned paddingSize, std::vector<Type*>& llvmFields)
 {
     unsigned numInts = paddingSize / 4;
     unsigned numBytes = paddingSize - numInts * 4;
@@ -949,7 +953,9 @@ bool Llvm::helperRequiresShadowStack(CORINFO_METHOD_HANDLE corinfoMethodHnd)
            corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2INT_OVF) ||
            corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2LNG_OVF) ||
            corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2UINT_OVF) ||
-           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2ULNG_OVF);
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_DBL2ULNG_OVF) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_LMUL_OVF) ||
+           corinfoMethodHnd == _compiler->eeFindHelper(CORINFO_HELP_ULMUL_OVF);
 }
 
 void Llvm::buildHelperFuncCall(GenTreeCall* call)
@@ -1149,6 +1155,13 @@ void Llvm::buildInd(GenTree* node, Value* ptr)
     mapGenTreeToValue(node, _builder.CreateLoad(
                                  castIfNecessary(ptr,
                                      getLlvmTypeForVarType(node->TypeGet())->getPointerTo())));
+}
+
+void Llvm::buildObj(GenTreeObj* node, Value* ptr)
+{
+    // cast the pointer to create the correct load instructions
+    mapGenTreeToValue(node, _builder.CreateLoad(
+                                castIfNecessary(ptr, getLlvmTypeForStruct(node->GetLayout()->GetClassHandle())->getPointerTo())));
 }
 
 Value* Llvm::buildJTrue(GenTree* node, Value* opValue)
@@ -1393,6 +1406,74 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
     }
 }
 
+void Llvm::storeObjAtAddress(Value* baseAddress, Value* data, unsigned startIx, unsigned gcFieldCount, FieldStoreLayout* gcFieldLayout)
+{
+    unsigned i = startIx;
+    while (!gcFieldLayout[i].IsEndStruct)
+    {
+        Value* address = _builder.CreateGEP(baseAddress, _builder.getInt32(gcFieldLayout[i].FieldOffset));
+
+        // get the data if this instruction has data
+        Value* fieldData = nullptr;
+        if (!gcFieldLayout[i].IsEndStruct)
+        {
+            if (data->getType()->isStructTy())
+            {
+                unsigned index = _module->getDataLayout().getStructLayout((llvm::StructType*)(data->getType()))->getElementContainingOffset(gcFieldLayout[i].FieldOffset);
+                fieldData = _builder.CreateExtractValue(data, index);
+            }
+            else
+            {
+                // single field IL structs are not LLVM structs
+                fieldData = data;
+            }
+        }
+
+        if (gcFieldLayout[i].IsStartStruct)
+        {
+            storeObjAtAddress(address, fieldData, i + 1, gcFieldCount, gcFieldLayout);
+        }
+        else
+        {
+            _builder.CreateCall(getOrCreateRhpAssignRef(),
+                                ArrayRef<Value*>{address, castIfNecessary(fieldData, Type::getInt8PtrTy(_llvmContext))});
+        }
+        i++;
+    }
+}
+
+void Llvm::importStoreObj(GenTreeStoreInd* storeIndOp)
+{
+    unsigned gcFieldCount;
+    Value* baseAddress = getGenTreeValue(storeIndOp->Addr());
+
+    if (!storeIndOp->OperIsBlk())
+    {
+        failFunctionCompilation(); // cant get struct handle. TODO-LLVM: try as an assert
+    }
+
+    ClassLayout* structLayout = storeIndOp->AsBlk()->GetLayout();
+
+    // zero initialization  check
+    GenTree* dataOp = storeIndOp->Data();
+    if (dataOp->OperIs(GT_CNS_INT) && dataOp->AsIntCon()->IconValue() == 0)
+    {
+        _builder.CreateMemSet(baseAddress, _builder.getInt8(0), _builder.getInt32(structLayout->GetSize()), {});
+        return;
+    }
+
+    CORINFO_CLASS_HANDLE structClsHnd = structLayout->GetClassHandle();
+    FieldStoreLayout* gcFieldLayout;
+    gcFieldCount = _getObjectLayout(_thisPtr, structClsHnd, &gcFieldLayout);
+
+    storeObjAtAddress(baseAddress, getGenTreeValue(storeIndOp->Data()), 0, gcFieldCount, gcFieldLayout);
+
+    // just copy all the fields again for simplicity,
+    // if all the fields were set using RhpAssignRef then a possible optimisation would be to skip this line
+    Value* toStore = getGenTreeValue(storeIndOp->Data());
+    _builder.CreateStore(toStore, castIfNecessary(baseAddress, toStore->getType()->getPointerTo()));
+}
+
 Value* Llvm::localVar(GenTreeLclVar* lclVar)
 {
     Value*       llvmRef;
@@ -1578,6 +1659,9 @@ void Llvm::visitNode(GenTree* node)
         case GT_NO_OP:
             emitDoNothingCall();
             break;
+        case GT_OBJ:
+            buildObj(node->AsObj(), getGenTreeValue(node->AsOp()->gtOp1));
+            break;
         case GT_PHI:
             buildEmptyPhi(node->AsPhi());
             break;
@@ -1592,6 +1676,9 @@ void Llvm::visitNode(GenTree* node)
             break;
         case GT_STOREIND:
             importStoreInd((GenTreeStoreInd*)node);
+            break;
+        case GT_STORE_OBJ:
+            importStoreObj((GenTreeStoreInd*)node);
             break;
         case GT_AND:
         case GT_OR:
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -176,10 +176,10 @@ StructDesc* Llvm::getStructDesc(CORINFO_CLASS_HANDLE structHandle)
             sparseFields[i] = nullptr;
 
         // determine the largest field for unions, and get fields in order of offset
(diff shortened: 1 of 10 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -1593,6 +1677,9 @@ void Llvm::visitNode(GenTree* node)
         case GT_STOREIND:
             importStoreInd((GenTreeStoreInd*)node);
             break;
+        case GT_STORE_OBJ:
+            importStoreObj((GenTreeStoreInd*)node);
+            break;
```
```suggestion
        case GT_STOREIND:
            importStoreInd(node->AsStoreInd());
            break;
        case GT_STORE_OBJ:
            importStoreObj(node->AsStoreInd());
            break;
```
(Would also be nice to rename these `buildStoreXYZ` for consistency, or at least the new function)

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter_Statics.cs
```diff
@@ -379,6 +379,40 @@ public static int PadOffset(TypeDesc type, int atOffset)
             return padding;
         }
 
+        public static void GetObjectLayoutInstructions(TypeDesc type, List<FieldStoreLayout> layout, int startOffset)
+        {
+            foreach (FieldDesc f in type.GetFields())
+            {
+                if (f.IsStatic) continue;
```
SingleAccretion:
```suggestion
                if (f.IsStatic)
                {
                    continue;
                }

```
(Or without the braces if that's the prevailing style)

src/coreclr/jit/llvm.h
```diff
@@ -64,6 +64,14 @@ struct SsaPairHash
     }
 };
 
+struct FieldStoreLayout
+{
+    unsigned AbsoluteOffset;
+    unsigned FieldOffset;
+    unsigned IsStartStruct;
+    unsigned IsEndStruct;
+};
```
SingleAccretion: Hmmm, `FieldStoreLayout`.

This feels more like `FieldSlotDesc` or something to me.

This needs some documentation, on the whole and for the individual members: what sort of fields are recorded, some examples on what the array of these (which is what the "type layout" as a whole will be) looks like for some interesting cases, e. g.
```cs
struct StructWithObj { object Obj; }
struct NestedStructWithObj { StructWithObj StructWithObj; }
struct StructNoGcPtr { int A, B, C, D; }
struct StructComposite { int A; object Obj; }
struct StructMixed { int A; object Obj, double C; StructWithObj StructWithObj; }
```
Also, I wonder if it would have been fruitful to try and pass the full representation of the VM layout to the Jit (build a small copy of `FieldDesc` in the Jit). Seems like it could've been shared with the "build LLVM struct type" code.

src/coreclr/jit/llvm.h
```diff
@@ -133,6 +142,7 @@ class Llvm
     void buildHelperFuncCall(GenTreeCall* call);
     llvm::FunctionType* buildHelperLlvmFunctionType(GenTreeCall* call, bool withShadowStack);
     void buildInd(GenTree* node, Value* ptr);
+    void buildObj(GenTreeObj* node, Value* ptr);
```
SingleAccretion: Nit: `void buildObj(GenTreeObj* node, Value* ptr)` -> `void buildObj(GenTreeObj* obj)` and retrieve the pointer inside it.

(Realize `buildInd` is doing this, it should be fixed too -- possibly as a separate cleanup)

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -287,6 +288,31 @@ public static CorInfoTypeWithMod getParameterType(IntPtr thisHandle, CORINFO_CLA
...
+            TypeDesc type = _this.HandleToObject(inputType);
+
+            List<FieldStoreLayout> layout = new();
+
+            _this._compilation.GetObjectLayoutInstructions(type, layout);
+
+            uint fieldCount = (uint)layout.Count;
+
+            // TODO-LLVM: release?  Is there a smarter way of doing this
+            var layoutPtr = Marshal.AllocHGlobal((int)(sizeof(FieldStoreLayout) * fieldCount));
```
SingleAccretion: Does the compiler project reference .NET ref recent enough for it to have `NativeMemory.Alloc`? Would be nice to use (if possible) it as it's much faster than `AllocHGlobal`.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -287,6 +288,31 @@ public static CorInfoTypeWithMod getParameterType(IntPtr thisHandle, CORINFO_CLA
...
+
+            TypeDesc type = _this.HandleToObject(inputType);
+
+            List<FieldStoreLayout> layout = new();
+
+            _this._compilation.GetObjectLayoutInstructions(type, layout);
+
+            uint fieldCount = (uint)layout.Count;
+
+            // TODO-LLVM: release?  Is there a smarter way of doing this
```
SingleAccretion: This does seem like it would leak memory globally for the process, doesn't it.

Presumably there is a "cleanup once the compilation is over" mechanism in this Jit-EE interface implementation, and presumably we should be using it here to release the memory.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -287,6 +288,31 @@ public static CorInfoTypeWithMod getParameterType(IntPtr thisHandle, CORINFO_CLA
...
+        [UnmanagedCallersOnly]
+        public static uint getObjectLayout(IntPtr thisHandle, CORINFO_CLASS_STRUCT_* inputType, void** layoutPtrPtr)
+        {
+            var _this = GetThis(thisHandle);
+
+            TypeDesc type = _this.HandleToObject(inputType);
+
+            List<FieldStoreLayout> layout = new();
+
+            _this._compilation.GetObjectLayoutInstructions(type, layout);
```
SingleAccretion: Nit: cleaner to return the list from `GetObjectLayoutInstructions` instead of mutating it?

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -287,6 +288,31 @@ public static CorInfoTypeWithMod getParameterType(IntPtr thisHandle, CORINFO_CLA
...
+
+            uint fieldCount = (uint)layout.Count;
+
+            // TODO-LLVM: release?  Is there a smarter way of doing this
+            var layoutPtr = Marshal.AllocHGlobal((int)(sizeof(FieldStoreLayout) * fieldCount));
+            *layoutPtrPtr = (void*)layoutPtr;
+
+            for (int i = 0; i < fieldCount; i++)
+            {
+                Marshal.StructureToPtr(layout[i], layoutPtr + sizeof(FieldStoreLayout) * i, false);
```
SingleAccretion: Isn't this just:
```cs
((FieldStoreLayout*)layoutPtr)[i] = layout[i];
```
?

(Also would be nice to get rid of the casts by typing `layoutPtr` as `FieldStoreLayout` and `layoutPtrPtr` as `FieldStoreLayout**`)

src/coreclr/jit/llvm.cpp
```diff
@@ -1151,6 +1157,13 @@ void Llvm::buildInd(GenTree* node, Value* ptr)
                                      getLlvmTypeForVarType(node->TypeGet())->getPointerTo())));
 }
 
+void Llvm::buildObj(GenTreeObj* node, Value* ptr)
+{
+    // cast the pointer to create the correct load instructions
+    mapGenTreeToValue(node, _builder.CreateLoad(
+                                castIfNecessary(ptr, getLlvmTypeForStruct(node->GetLayout()->GetClassHandle())->getPointerTo())));
```
SingleAccretion: I think we'd want to make an overload of `getLlvmTypeForStruct` that takes a `ClassLayout*` directly.

That way we'll be able to deal with "block layouts" (that do not have a struct handle) transparently in the future.

src/coreclr/jit/llvm.cpp
```diff
@@ -1393,6 +1406,74 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+    {
+        Value* address = _builder.CreateGEP(baseAddress, _builder.getInt32(gcFieldLayout[i].FieldOffset));
+
+        // get the data if this instruction has data
+        Value* fieldData = nullptr;
+        if (!gcFieldLayout[i].IsEndStruct)
+        {
+            if (data->getType()->isStructTy())
+            {
+                unsigned index = _module->getDataLayout().getStructLayout((llvm::StructType*)(data->getType()))->getElementContainingOffset(gcFieldLayout[i].FieldOffset);
```
SingleAccretion: This could use an intermediate variable for the struct layout.

(Also, `index` -> `llvmFieldIndex`)

src/coreclr/jit/llvm.cpp
```diff
@@ -1393,6 +1406,74 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+    CORINFO_CLASS_HANDLE structClsHnd = structLayout->GetClassHandle();
+    FieldStoreLayout* gcFieldLayout;
+    gcFieldCount = _getObjectLayout(_thisPtr, structClsHnd, &gcFieldLayout);
+
+    storeObjAtAddress(baseAddress, getGenTreeValue(storeIndOp->Data()), 0, gcFieldCount, gcFieldLayout);
+
+    // just copy all the fields again for simplicity,
+    // if all the fields were set using RhpAssignRef then a possible optimisation would be to skip this line
+    Value* toStore = getGenTreeValue(storeIndOp->Data());
+    _builder.CreateStore(toStore, castIfNecessary(baseAddress, toStore->getType()->getPointerTo()));
```
SingleAccretion: Hmm, it does not seem right to me to store to the same location twice.

Presumably it breaks the GC to store a ref to the heap without a barrier, and the second store would be doing that.

Also, why do we need to do it all? Wouldn't the equivalent of the below (that looks a lot like the current IL code):
```cs
void StoreObjFields(addr, offset, obj)
{
    if (obj.IsGCPtr()) // TYP_REF/TYP_BYREF
    {
        StoreWithWriteBarrier(addr + offset, obj);
        return;
    }
    if (!obj.HasGCPtr())
    {
        StoreLlvm(addr + offset, obj);
        return;
    }

    foreach (field in obj)
    {
        StoreObjFields(addr, offset, field);
        offset += field.Size; // With padding
    }
}
```
Store the whole struct (modulo padding -- we'll need to handle it with `memcpy` for explicitly-sized structs)?

src/coreclr/jit/llvm.cpp
```diff
@@ -1393,6 +1406,74 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+
+    ClassLayout* structLayout = storeIndOp->AsBlk()->GetLayout();
+
+    // zero initialization  check
+    GenTree* dataOp = storeIndOp->Data();
+    if (dataOp->OperIs(GT_CNS_INT) && dataOp->AsIntCon()->IconValue() == 0)
+    {
+        _builder.CreateMemSet(baseAddress, _builder.getInt8(0), _builder.getInt32(structLayout->GetSize()), {});
+        return;
+    }
```
SingleAccretion: This needs a `TODO-LLVM` for `handle non-zero InitBlks.`.

(These are generated for the `initblk` IL instruction that has the same semantics as `memset`. In IR they will be block stores with `INIT_VAL` `data`)

src/coreclr/jit/llvm.cpp
```diff
@@ -1393,6 +1406,74 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+    if (!storeIndOp->OperIsBlk())
+    {
+        failFunctionCompilation(); // cant get struct handle. TODO-LLVM: try as an assert
+    }
+
+    ClassLayout* structLayout = storeIndOp->AsBlk()->GetLayout();
+
+    // zero initialization  check
+    GenTree* dataOp = storeIndOp->Data();
+    if (dataOp->OperIs(GT_CNS_INT) && dataOp->AsIntCon()->IconValue() == 0)
```
```suggestion
    if (dataOp->IsIntegralConst(0))
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1393,6 +1406,74 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+    {
+        _builder.CreateMemSet(baseAddress, _builder.getInt8(0), _builder.getInt32(structLayout->GetSize()), {});
+        return;
+    }
+
+    CORINFO_CLASS_HANDLE structClsHnd = structLayout->GetClassHandle();
+    FieldStoreLayout* gcFieldLayout;
+    gcFieldCount = _getObjectLayout(_thisPtr, structClsHnd, &gcFieldLayout);
```
SingleAccretion: Re-creating this for each node seems expensive. We need a side table that'll associate it with `ClassLayout*` pointers.

SingleAccretion (commented):
Thinking we should probably try and see how replicating the full field information on the Jit side is going to look like, seems we need it.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
yowl: No, this project is .net 5.0 unfortunately.  We are some way behind `main`, if we caught up we could.  I'll put a todo for now.

yowl: Or we could just change it, and worry about merging later.

SingleAccretion: Up to you, it's a little thing, may not be worth the hassle of updating to 6.

src/coreclr/jit/llvm.h
yowl: I suppose this goes in hand with not double copying the memory below.  I think the problem I'm having is how to create a recursive struct in the C# side, that I can marshal to the C++ side.  I'll need that wont I?

SingleAccretion: Yes, but it does not have to be recursive on the interface side. E. g.:
```c
struct TypeDescriptor
{
     size_t fieldCount;
     CORINFO_FIELD_HANDLE* fields;
}

TypeDescriptor getTypeDescriptor(CORINFO_CLASS_HANDLE clsHnd);
```
Then we should be able to build an appropriate recursive abstraction on the Jit side.

yowl: Thanks.  What I'll do then is create another PR for just this and incorporate the new struct into the build LLVM struct type code.  Then I'll come back to this one.

yowl:
@SingleAccretion The change from #1854 is now incorporated here, removing the "instruction" struct.  I've not needed to make it recursive in the end.

It's passing the tests, but I have 2 outstanding questions

1. How can I identify whether a field in the struct needs an `RhpAssignRef`?  Right now I just have
```
if (fieldClassHandle != NO_CLASS_HANDLE)
{
    isGcPtr = corInfoType == CORINFO_TYPE_CLASS;  //TODO-LLVM: how to tell if the field needs an RhpAssignRef - does _getTypeDescriptor have to do that?
}
```
2. Should I be filling in spaces between fields with 0s ?

SingleAccretion:
> How can I identify whether a field in the struct needs an `RhpAssignRef`?

`isGcPtr = corInfoType == CORINFO_TYPE_CLASS` looks right, but the existing backend code also does this for `BYREF`s (see `genCodeForCpObj` in `codegenxarch.cpp`), but I do not sure why, given byrefs shouldn't be living on the heap. You can (should) also omit the write barrier if the destination address is local (or `TGT_NOT_HEAP`).

> Should I be filling in spaces between fields with 0s ?

In RyuJIt we assume padding is not significant in case it is "natural", that is, the struct is not `LayoutKind.Explicit` and doesn't have the `Size` metadata.

So if the struct "has custom layout" like that, it'll need to be copied wholesale, if not - it can be copied field by field, ignoring padding. In either case, you don't need to zero anything.

yowl:
I'll write some tests for `ExplicitLayout`

yowl:
I added some code to add an explicitlayout flag to TypeDescriptor so that a decision can be made if copying of gaps is required.  I added a test for that, but it doesn't compile with the clrjit module yet.

If the target is not heap, then it will just do a memcpy.

src/coreclr/jit/llvm.cpp
```diff
@@ -1424,6 +1444,119 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+unsigned Llvm::buildMemCpy(Value* baseAddress, unsigned startOffset, unsigned endOffset, Value* srcAddress)
+{
+    Value* paddingDestAddress = _builder.CreateGEP(baseAddress, _builder.getInt32(startOffset));
+
+    const unsigned paddingSize = endOffset - startOffset;
+    // copy the bytes in the padding
+    _builder.CreateMemCpy(paddingDestAddress, llvm::Align(), srcAddress, llvm::Align(), paddingSize);
+
+    return paddingSize;
+}
```
```suggestion
// Copies endOffset - startOffset bytes, endOffset is exclusive.
unsigned Llvm::buildMemCpy(Value* baseAddress, unsigned startOffset, unsigned endOffset, Value* srcAddress)
{
    Value* destAddress = _builder.CreateGEP(baseAddress, _builder.getInt32(startOffset));
    unsigned size = endOffset - startOffset;

    _builder.CreateMemCpy(destAddress, llvm::Align(), srcAddress, llvm::Align(), size);

    return size;
}
```
Doesn't seem like there is great a reason to make this padding-copying-specific.

src/coreclr/jit/llvm.cpp
```diff
@@ -1424,6 +1444,119 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+        }
+
+        if (fieldData->getType()->isStructTy())
+        {
+            assert(fieldDesc->getClassHandle() != NO_CLASS_HANDLE);
+
+            // recurse into struct
+            storeObjAtAddress(address, fieldData, getStructDesc(fieldDesc->getClassHandle()));
+
+            bytesStored += fieldData->getType()->getPrimitiveSizeInBits() / 8;
```
```suggestion
            bytesStored += fieldData->getType()->getPrimitiveSizeInBits() / BITS_PER_BYTE;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1424,6 +1444,119 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+                _builder.CreateCall(getOrCreateRhpAssignRef(),
+                                    ArrayRef<Value*>{address,
+                                                     castIfNecessary(fieldData, Type::getInt8PtrTy(_llvmContext))});
+                bytesStored += TARGET_POINTER_SIZE;
+            }
+            else
+            {
+                _builder.CreateStore(fieldData, castIfNecessary(address, fieldData->getType()->getPointerTo()));
+
+                bytesStored += fieldData->getType()->getPrimitiveSizeInBits() / 8;
```
```suggestion
                bytesStored += fieldData->getType()->getPrimitiveSizeInBits() / BITS_PER_BYTE;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1424,6 +1444,119 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+            }
+        }
+    }
+
+    unsigned llvmStructSize = data->getType()->getPrimitiveSizeInBits() / 8;
```
```suggestion
    unsigned llvmStructSize = data->getType()->getPrimitiveSizeInBits() / BITS_PER_BYTE;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1424,6 +1444,119 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+    GenTree*    dataOp     = genTreeObj->Data();
+    if (dataOp->IsIntegralConst(0))
+    {
+        _builder.CreateMemSet(baseAddress, _builder.getInt8(0), _builder.getInt32(structLayout->GetSize()), {});
+        return;
+    }
+
+    CORINFO_CLASS_HANDLE structClsHnd  = structLayout->GetClassHandle();
+    StructDesc*          structDesc    = getStructDesc(structClsHnd);
+    bool                 targetNotHeap = (indirOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0;
```
```suggestion
    bool                 targetNotHeap = ((indirOp->gtFlags & GTF_IND_TGT_NOT_HEAP) != 0) || genTreeObj->Addr()->OperIsLocalAddr();
```
The condition seems to have been inverted.

src/coreclr/jit/llvm.cpp
```diff
@@ -1424,6 +1444,119 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+    unsigned llvmStructSize = data->getType()->getPrimitiveSizeInBits() / 8;
+    if (structDesc->isExplicitLayout() && llvmStructSize > bytesStored)
+    {
+        Value* srcAddress = _builder.CreateGEP(baseAddress, _builder.getInt32(bytesStored));
+
+        buildMemCpy(baseAddress, bytesStored, llvmStructSize, srcAddress);
+    }
+}
+
+void Llvm::buildStoreObj(GenTreeIndir* indirOp)
```
SingleAccretion: The method is named `buildStoreObj` but takes `GenTreeIndir*`. Why not make it take `GenTreeObj*`?

(Also I would rename `indirOp` to something with `store` in the name)

I think we need to establish a naming convention for trees-vs-LLVM-values, they are mixed right now and it can be confusing. Appending `node` for trees seems a fine solution, e. g. here we could have:
```c
GenTree* dstAddrNode = storeObjNode->Addr();
Value* baseAddress = getGenTreeValue(dstAddrNode);
```
Or we could leave the trees with "simple" names and append `Value` to `llvm::Value*`s instead, e. g.:
```
GenTree* dstAddr = storeObj->Addr();
Value* baseAddressValue = getGenTreeValue(dstAddr);
```
Or have it be as it is (simple names for both).

What do you think?

src/coreclr/jit/llvm.cpp
```diff
@@ -1424,6 +1444,119 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+{
+    Value* baseAddress = getGenTreeValue(indirOp->Addr());
```
SingleAccretion: This should be moved down to where it is first used.

src/coreclr/jit/llvm.cpp
```diff
@@ -1424,6 +1444,119 @@ void Llvm::importStoreInd(GenTreeStoreInd* storeIndOp)
...
+    {
+        _builder.CreateMemSet(baseAddress, _builder.getInt8(0), _builder.getInt32(structLayout->GetSize()), {});
+        return;
+    }
+
+    CORINFO_CLASS_HANDLE structClsHnd  = structLayout->GetClassHandle();
+    StructDesc*          structDesc    = getStructDesc(structClsHnd);
+    bool                 targetNotHeap = (indirOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0;
+
+    Value* toStore = getGenTreeValue(genTreeObj->Data());
```
SingleAccretion: `toStore` -> `dataValue`.

src/coreclr/jit/llvm_types.h
```diff
@@ -47,18 +70,24 @@ struct FieldDesc
     {
         return m_corType;
     }
+
+    bool isGcPointer()
+    {
+        return m_isGcPointer;
+    }
```
```suggestion
    bool isGcPointer()
    {
        return m_corType == CORINFO_TYPE_CLASS || m_corType == CORINFO_TYPE_BYREF;
    }
```
Seems like there's no need to cache it.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -295,6 +295,7 @@ public struct TypeDescriptor
         {
             public uint FieldCount;
             public CORINFO_FIELD_STRUCT_** Fields; // array of CORINFO_FIELD_STRUCT_*
+            public uint IsExplicitLayout;
```
SingleAccretion: We should be mindful not to waste too much space here, since it is cached (this is 32 bits for a boolean). It's fine as-is right now, but if we add more booleans like this, we should make this `uint` a flags value.

src/coreclr/jit/llvm_types.h
```diff
@@ -12,8 +12,30 @@
...
+
+    CORINFO_FIELD_HANDLE getField(unsigned index)
+    {
+        return m_fields[index];
+    }
+
+    unsigned getIsExplicitLayout()
+    {
+        return m_isExplicitLayout;
+    }
```
SingleAccretion: The condition here is: `LayoutKind.Explicit || HasSizeMetadata`, i. e. the following struct should also be treated as one with significant padding:
```cs
[StructLayout(LayoutKind.Sequential, Size = 10)]
struct S
{
    public byte Field;
}
```
This would also imply renaming this (and related interface code) to something like `hasSignificantPadding`.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter_Statics.cs
```diff
@@ -379,6 +379,44 @@ public static int PadOffset(TypeDesc type, int atOffset)
             return padding;
         }
 
+        public static void GetObjectLayoutInstructions(TypeDesc type, List<FieldStoreLayout> layout, int startOffset)
```
SingleAccretion: This and related code unused?

src/coreclr/jit/llvm.cpp
yowl: oops

src/coreclr/jit/llvm_types.h
yowl: Thanks

yowl:
I think I've done all the review items now.

src/coreclr/jit/llvm.cpp
SingleAccretion: @yowl ^

SingleAccretion (approved):

jkotas (approved):
Thank you!

PR #1836, opened 2022-01-30
Update location of Emscripten

kant2002:
Docs mention `dotnet/emsdk` but CI use `emscripten-core/emsdk`. Also very specific commit needed. Document that with snippet of code.


yowl:
I wonder if its best to leave these docs alone as they relate to the official coreclr build.  We can add your instructions about the specific version of emsdk to  docs/using-nativeaot/compiling.md ?

kant2002:
This part is specifically for NativeAOT-LLVM, I did not see any issue changes them since nothing remotely like that in NativeAOT and CoreCLR.

kant2002:
And `docs/using-nativeaot/compiling.md` specifically for developer who want compile application using NativeAOT-LLVM, not develop using it.

yowl:
Ok, fine with me.

jkotas (approved):
Thank you

PR #1843, opened 2022-02-01
[NativeAOT-LLVM] map runtimeimports to their native name for ldftn

yowl:
This PR partially addresses a problem with ldftn for RuntimeImports, i.e. from

https://github.com/dotnet/runtimelab/blob/3dd23bfdd00013b38bbc9b8de27cc9a2dd7cce12/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/RuntimeExports.cs#L455

Right now these are reported as undefined as they are not translated to the native names:
```
warning: undefined symbol: S_P_CoreLib_System_Runtime_InternalCalls__RhpNewArray (referenced by top-level compiled C/C++ code)
```
Emscripten will create imports for these in the wasm module, and stub them in JS.  However a WASI host like wasmtime cannot satisfy these imports and just refuses to start.  This PR at least lets the module run in a WASI host, although if these function pointers were ever called, they might be treated as managed calls, get a shadow stack arg, and fail.  I propose to not solve that in the IL->LLVM compilation, but leave it until it's hit in the RyuJIT compilation.

There is still one undefined symbol preventing wasmtime starting, `dotnet_browser_entropy`, but that can be tackled in it's own PR.

cc @SingleAccretion

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -3069,47 +3069,7 @@ namespace Internal.IL
 
         private ExpressionEntry ImportRawPInvoke(MethodDesc method, StackEntry[] arguments, LLVMBuilderRef builder, TypeDesc forcedReturnType = null)
         {
-            string realMethodName = method.Name;
-
-            if (method.IsPInvoke)
-            {
-                string entrypointName = method.GetPInvokeMethodMetadata().Name;
-                if (!String.IsNullOrEmpty(entrypointName))
-                {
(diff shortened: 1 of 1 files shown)


yowl:
Oh we are hitting the hash problem https://github.com/emscripten-core/emscripten/issues/16164  We'll have to wait for that to be resolved.

yowl:
The emscripten bug is fixed so this can be reviewed now.  Thanks.

jkotas:
> propose to not solve that in the IL->LLVM compilation, but leave it until it's hit in the RyuJIT compilation.

Ok, sounds reasonable.

jkotas:
> propose to not solve that in the IL->LLVM compilation, but leave it until it's hit in the RyuJIT compilation.

Ok, sounds reasonable.

jkotas (approved):

Issue #1390, opened 2021-08-06
Question: NativeAOT-LLVM: For WebAssembly imports, DllImport Value misuse?


yowl:
When importing functions from other modules in WebAssembly using the proposed WebAssembly canonical ABI, https://github.com/bytecodealliance/witx-bindgen, , the final Wasm/Wat looks like 
```
  (import "hellowasi" "reverse" (func $reverse (type 4)))
```
This imports the `reverse` function from the `hellowasi` module.  In order to generate this from LLVM, the LLVM has to look like:
```
declare void @reverse(i8*, i32, i8*) #2
...
attributes #2 = { "wasm-import-module"="hellowasi" "wasm-import-name"="reverse" }
```
For information, the C equivalent for emscripte/wasm-ld would be
```
__attribute__((import_module("hellowasi"), import_name("reverse"))) void reverse();
```

In c# this starts with currently
```
        [DllImport("*")]
        private static extern void reverse(byte* utf8Bytes, int len, IntPtr returnArea);
```

Not all `DllImports` are going to be Wasm imports, some like e.g. printf are supplied by the Emscripten runtime.  A rule could be created:

`DllImportAttribute.Value == "*"`  - treat as now.
`DllImportAttribute.Value != "*"`- assume its a Wasm import from another module and add the LLVM attributes - the Value property goes to the `wasm-import-module` attribute.

So this example would be
```
 [DllImport("hellowasi")]
```

Would that be acceptable or a misuse of DllImport and "The name of the DLL file that contains the entry point" https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute.value?view=net-5.0

jkotas:
>  A rule could be created:

I think we may want this rule to be configurable, similar to how the direct PInvoke calls are configurable today: https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/docs/using-nativeaot/interop.md#direct-pinvoke-calls

Otherwise, it looks reasonable to me.

jkotas:
@lewing Is there any thinking about how the support for WebAssembly canonical ABI is going to look like in Mono?

yowl:
Looking at the PInvoke config, how about
```
<WasmImport Import="wasi_snapshot_preview1!random_get" />
```

Then when 
```
[DllImport("*")]
private static extern int random_get(byte* utf8Bytes, int len);
```
Ilc would match `random_get` to the `WasmImport` and generate the correct LLVM import?

[...]

PR #1845, opened 2022-02-01
[NativeAOT-LLVM] Add support for controlling the import name used for PInvokes.

yowl:
This PR allows the module name used for WebAssembly imports to be controlled.  It follows #1390.

Using these changes it's possible to add
```xml
<ItemGroup>
	<WasmImport Include="wasi_snapshot_preview1!random_get" />
</ItemGroup>
```
to the project file and have the final WebAssembly WAT look like:

```
 (import "wasi_snapshot_preview1" "random_get" (func $random_get (type 3)))
```
This will allow WASI functions to be used in user code and will facilitate module linking, e.g. for `wit-bindgen`.

I realise that mono may put forward an alternative proposal that will take precedence at which point this will have to be redone.

### DIFF
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -88,6 +88,25 @@ public static int Answer()
 > dotnet publish /p:NativeLib=Static /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false /p:EmccExtraArgs="-s EXPORTED_FUNCTIONS=_Answer -s EXPORTED_RUNTIME_METHODS=cwrap" --self-contained
 ```
 
+#### WebAssembly module imports
+Functions in other WebAssembly modules can be imported and invoked using `DllImport` e.g.
+```cs
+[DllImport("*")]
+static extern int random_get(byte* buf, uint size);
+```
+Be default emscripten will create a WebAssembly import for this function, importing from the `env` module.  This can be controlled with `WasmImport` items in the project file.  For example
+```xml
+<ItemGroup>
+    <WasmImport Include="wasi_snapshot_preview1!random_get" />
+</ItemGroup>
+```
+Will cause the above `random_get` to create this WebAssembly:
+```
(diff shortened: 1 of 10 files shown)


MichalStrehovsky:
Does WASI support dlopen/dlsym? Just wondering whether we should just reuse the directpinvoke mechanism instead of building an alternative on the side. If dlopen/dlsym doesn't exist, we could make directpinvoke mean something else on WASM.

yowl:
Yes, Wasm has the module linking proposal which covers dynamic loading of other modules (I think the term is going be adapter modules)  Its in flux as you can see from work the ByteCode Alliance are currently doing https://github.com/bytecodealliance/wasm-tools/issues/450.  Ive not read the proposal in detail but the structure looks a little more complex that just importing some symbols from another module, there is a concept of nesting and what they call link time virtualization (https://github.com/WebAssembly/module-linking/blob/main/design/proposals/module-linking/Explainer.md#link-time-virtualization) for example:
```
(adapter module
  (module $A
    (func (export "answer") (result i32) (i32.const 42))
  )
  (module $B
    (func (import "the" "answer") (result i32))
  )
  (instance $a (instantiate $A))
  (instance $b (instantiate $B (import "the" (instance $a))))
)
```
I'm happy to reuse the directpinvoke configuration here though, and can revisit as necessary when the module linking proposal firms up.  I didn't use it initialliy as it seemed to be about the code generation and thought I might be abusing the intention.

Wasi SDK also implements `dlopen` in it's libc, but I think that predates the changes for the module linking proposal, so not sure how much effort should be spent on that as it's changing.

yowl:
So in short, I can reuse the directpinvoke configuration, no problem.. Want me to change it?

MichalStrehovsky:
> So in short, I can reuse the directpinvoke configuration, no problem.. Want me to change it?

If WASM is going to have a working dlopen, we'll need a mechanism for deciding whether to generate a WASM import or use dlopen/dlsym, and that's what directpinvoke is for. Let's not overload it then.

I think eventually it would be nice if we could use the name of the DllImport, but that's something Mono will do their own way and then we'll need to adapt here I guess.

It's sad that WASM has this extra attribute and it doesn't work the way it works elsewhere (the decision on how to resolve symbols is determined at linking time, not at compile time).

MichalStrehovsky (approved):

PR #1846, opened 2022-02-02
[NativeAOT-LLVM] Change importance of missing EMSDK Message task to High so it appears on the console

### LINKED ISSUE #1839: [NativeAOT-LLVM] Surface message for missing EMSDK
When publishing for Wasm, and Emscripten is not installed, the message is produced (and can be found with difficulty in the binlog), but is not surfaced to stdout.  It should be.

yowl:
This PR just does what it says in the title.  Fixes #1839 

It's not super visible, but better than nothing:

image


jkotas:
> It's not super visible, but better than nothing:

You can make it warning if you wanted it to be more visible.

jkotas (approved):
LGTM

Issue #1848, opened 2022-02-05
[NativeAOT-LLVM] For WASI are we going to need another build of the C parts of the runtime?


yowl:
In Emscripten, for opening files, `O_CREAT` is `0x40` .  In wasi-sdk it is `0x1000000`.  This makes opening files with the current emscripten build of the runtime (`libPortableRuntime.a` I think is the affected library, but not checked), difficult.  It would be nice to support wasi features like opening files.  Is there an alternative like switch `ConvertOpenFlags` based on something (https://github.com/dotnet/runtimelab/blob/714aef1181f175ab0f45930eb89f90dfd53d2064/src/libraries/Native/Unix/System.Native/pal_io.c#L233) ?

I suppose really we should build the whole library with wasi-sdk, but that's not simple either as it mean another dependency and I think their Windows build also needs MinGW.

jkotas:
> Is there an alternative like switch ConvertOpenFlags based on something

I think you can implement a short-term workaround in these methods.

> I suppose really we should build the whole library with wasi-sdk,

It is likely unavoidable in the long run. I doubt that this is the only difference.

yowl:
Ok, thanks.  I've got something working for now anyway.  You can opine on how bad the hack is when the PR is in :-)

PR #1850, opened 2022-02-08
[NativeAOT-LLVM]  Enable support for WASI SDK

yowl:
This PR adds support for linking with the WASI SDK to enable opening of files on WASI hosts, e.g. wasmtime.

I'm changing the build property `EmccExtraArgs` to `WasmExtraLinkerArgs` as it only applies at the linking step and now applies to both Emscripten and Wasi SDK.

Adding build property `WasmWasi` to indicate that the build is for WASI, this turns off some unsupported Emscripten features, exceptions mainly.

Adding build property `WasiSdkPath` (which implies `WasmWasi`).  This causes the linking to be done by the Wasi SDK, and hence links the WASI SDK libc which has good support for WASI.

Added a hack to get around the fact that we compile the runtime with emscripten's headers, but use libc from the WASI SDK where some constants are different.  I'm using the `FileOptions`  to set a bit to indicate "when opening files, use the WASI SDK constants".  I did think about adding new enum values to other parameters to `new FileStream` but that breaks the ABI checks.  To use this hack, the user code would be:

```
            using (var fs = new FileStream("/tmp/direction", FileMode.OpenOrCreate, FileAccess.Write, FileShare.None, 4096, (FileOptions)0x20000000))
```

WIP as not sure this is a good idea.  wasi-libc originates from cloudlibc which has the different constants: Emscripten and musl are the same.  The alternative fix here would be to not use the libc file operations at all, but to straight to the wasi calls.  That would mean that running in Emscripten would not work as WASI in Emscripten is a WIP.

Updated the documentation.

Follows #1848


Cedtur (approved):

ChristianWeyer:
Do you know what the current state of this effort is @yowl ? Thanks!

yowl:
What I did here was a hack really to workaround different c header constants in the different libc(s).

There's maybe at least 2 ways to go forward with this

1.  Create a package that continues to use libc, but built against the libc in the WASI SDK (as opposed to emscripten)
2.  Do not use libc where WASI itself supplies the ABI, go straight to the WASI call (This is the recommendation I got from the Wasi SDK people - https://groups.google.com/g/emscripten-discuss/c/a97HaictQLo last comment).  Also requires another package.

So the state of the effort is that the IT problem is understood (I think!) , but the human time problem is harder to solve!

yowl:
Closing this one in favour of #2268

PR #1852, opened 2022-02-08
Complete JsonCodeGen experiment

kant2002:
Json source generator is part of dotnet/runtime as of .NET 6


kant2002:
I would speculate that all work would happens now in runtime, since what's planned is for .NET 7 is just regular work without need for "experimentation".

PR #1854, opened 2022-02-17
[NativeAOT-LLVM] Build and cache struct types in llvm.cpp

yowl:
This PR caches struct `CORINFO_CLASS_STRUCT_` fields and replicates some fields of `FieldDesc` in llvm.cpp facilitating the eventual reuse of the struct definitions for  GT_STORE_OBJ and GT_OBJ (to come in #1825 )

cc @SingleAccretion

Possibly coincidentally I've noticed that on my desktop with 16GB RAM I can no longer run the tests in parallel (x4) as it runs out of memory.  I did print the structs as they were cached and that does appear to be fine: it doesn't `new`multiple storage for the same struct, so maybe I was just close to the limit previously.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -37,15 +37,17 @@ static const uint32_t (*_isRuntimeImport)(void*, CORINFO_METHOD_STRUCT_*);
 static const char* (*_getDocumentFileName)(void*);
 static const uint32_t (*_firstSequencePointLineNumber)(void*);
 static const uint32_t (*_getOffsetLineNumber)(void*, unsigned int ilOffset);
-static const uint32_t(*_structIsWrappedPrimitive)(void*, CORINFO_CLASS_STRUCT_*, CorInfoType);
-static const uint32_t(*_padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned);
-static const CorInfoTypeWithMod(*_getArgTypeIncludingParameterized)(void*, CORINFO_SIG_INFO*, CORINFO_ARG_LIST_HANDLE, CORINFO_CLASS_HANDLE*);
-static const CorInfoTypeWithMod(*_getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*);
+static const uint32_t (*_structIsWrappedPrimitive)(void*, CORINFO_CLASS_STRUCT_*, CorInfoType);
+static const uint32_t (*_padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned);
+static const CorInfoTypeWithMod (*_getArgTypeIncludingParameterized)(void*, CORINFO_SIG_INFO*, CORINFO_ARG_LIST_HANDLE, CORINFO_CLASS_HANDLE*);
+static const CorInfoTypeWithMod (*_getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*);
+static const TypeDescriptor (*_getTypeDescriptor)(void*, CORINFO_CLASS_HANDLE);
 
 static char*                              _outputFileName;
 static Function*                          _doNothingFunction;
 
 static std::unordered_map<CORINFO_CLASS_HANDLE, Type*>* _llvmStructs = new std::unordered_map<CORINFO_CLASS_HANDLE, Type*>();
+static std::unordered_map<CORINFO_CLASS_HANDLE, StructDesc*>* _structDescMap = new std::unordered_map<CORINFO_CLASS_HANDLE, StructDesc*>();
 
 
 extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
@@ -64,7 +66,8 @@ extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
                                                 const uint32_t(*structIsWrappedPrimitive)(void*, CORINFO_CLASS_STRUCT_*, CorInfoType),
                                                 const uint32_t(*padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned),
                                                 const CorInfoTypeWithMod(*getArgTypeIncludingParameterized)(void*, CORINFO_SIG_INFO*, CORINFO_ARG_LIST_HANDLE, CORINFO_CLASS_HANDLE*),
-                                                const CorInfoTypeWithMod(*getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*))
+                                                const CorInfoTypeWithMod(*getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*),
+                                                const TypeDescriptor(*getTypeDescriptor)(void*, CORINFO_CLASS_HANDLE))
 {
     _thisPtr = thisPtr;
     _getMangledMethodName         = getMangledMethodNamePtr;
@@ -79,7 +82,8 @@ extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
     _structIsWrappedPrimitive     = structIsWrappedPrimitive;
     _padOffset = padOffset;
     _getArgTypeIncludingParameterized = getArgTypeIncludingParameterized;
-    _getParameterType = getParameterType;
+    _getParameterType             = getParameterType;
+    _getTypeDescriptor            = getTypeDescriptor;
 
     if (_module == nullptr) // registerLlvmCallbacks is called for each method to compile, but must only created the module once.  Better perhaps to split this into 2 calls.
     {
@@ -125,6 +129,16 @@ Value* Llvm::getGenTreeValue(GenTree* op)
     return _sdsuMap->at(op);
 }
 
+TypeDescriptor Llvm::getTypeDescriptor(CORINFO_CLASS_HANDLE structHandle)
+{
+    if (_typeDescriptorsMap->find(structHandle) == _typeDescriptorsMap->end())
+    {
+        _typeDescriptorsMap->insert({structHandle, _getTypeDescriptor(_thisPtr, structHandle)});
+    }
+
+    return _typeDescriptorsMap->at(structHandle);
+}
+
 // maintains compatiblity with the IL->LLVM generation.  TODO-LLVM, when IL generation is no more, see if we can remove this unwrapping
 bool structIsWrappedPrimitive(CORINFO_CLASS_HANDLE classHnd, CorInfoType primitiveType)
 {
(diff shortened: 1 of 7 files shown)


yowl:
Interesting, they fail here with OOM also.  I guess I've messed up somewhere, although they  do pass locally with `sequential` added to the command.

src/coreclr/jit/llvm.cpp
```diff
@@ -125,6 +129,16 @@ Value* Llvm::getGenTreeValue(GenTree* op)
...
 
+TypeDescriptor Llvm::getTypeDescriptor(CORINFO_CLASS_HANDLE structHandle)
+{
+    if (_typeDescriptorsMap->find(structHandle) == _typeDescriptorsMap->end())
+    {
+        _typeDescriptorsMap->insert({structHandle, _getTypeDescriptor(_thisPtr, structHandle)});
+    }
+
+    return _typeDescriptorsMap->at(structHandle);
+}
```
SingleAccretion: Appears unused (as well as the associated map).

src/coreclr/jit/llvm.h
```diff
@@ -64,6 +64,27 @@ struct SsaPairHash
...
+    int                  fieldOffset;
+    CORINFO_CLASS_HANDLE classHandle;
+    CorInfoType          corType;
+};
+
+struct StructDesc
+{
+    size_t     fieldCount;
+    FieldDesc* fieldDesc;
+};
```
SingleAccretion: It would be nice to make `StructDesc` and `FieldDesc` into "proper" types, with constructors, private members, accessor methods, etc.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -287,6 +291,58 @@ public static CorInfoTypeWithMod getParameterType(IntPtr thisHandle, CORINFO_CLA
...
+            {
+                if (!field.IsStatic)
+                {
+                    fieldCount++;
+                }
+            }
+
+            //TODO-LLVM: change to NativeMemory.Alloc when upgrade to .net6
+            IntPtr allocedMemory = Marshal.AllocHGlobal((int)(sizeof(CORINFO_FIELD_STRUCT_*) * fieldCount));
+            _allocedMemory.Add(allocedMemory);
```
SingleAccretion: I think it is not necessary to give type descriptors process-level lifetime. They should be just fine with compilation-level lifetime.

Nit: `allocedMemory` -> `fieldsArray`.

SingleAccretion (commented):
Not sure why are we OOMing...

src/coreclr/jit/llvm.h
yowl: ok, in the same files (llvm.cpp/llvm.h) or add new ones?

SingleAccretion: Up to you, I would slightly prefer new files I suppose.

src/coreclr/jit/llvm.cpp
yowl: Thanks, yes you are right..

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
yowl: I free them up when shutting down the jit

https://github.com/dotnet/runtimelab/blob/b495aecf7c51dc85cba6e1ee43a3c4f76d5632fb/src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs#L78-L79

If I make them non static and instance fields of `CorInfoImpl` then I'll have to implement the dispose pattern?

SingleAccretion: I would have expected it to be freed somewhere around `CompileMethodCleanup`.

However, now that I look at it again, the descriptors are dictionary-cached, and it is fine for that to have "static" lifetime. So let's leave them as-is.

yowl:
Thanks for the review.  I don't know what is going on with the OOM error.  I see this in the event viewer:
```
Faulting application name: ilc.exe, version: 42.42.42.42424, time stamp: 0x612ca7ed
Faulting module name: KERNELBASE.dll, version: 10.0.19041.1503, time stamp: 0xb2acaea9
Exception code: 0xe0434352
Fault offset: 0x0000000000034f69
Faulting process ID: 0x725c
Faulting application start time: 0x01d8259a02119621
Faulting application path: E:\GitHub\runtimelab\artifacts\tests\coreclr\Browser.wasm.Debug\Tests\Core_Root\nativeaot\tools\ilc.exe
Faulting module path: C:\Windows\System32\KERNELBASE.dll
Report ID: 63842411-7df7-4e12-9cd9-52d2bf945f4a
Faulting package full name: 
Faulting package-relative application ID: 
```
I don't know if its a dotnet error message, or what.  I've set the tests to sequential for now.

src/coreclr/jit/llvm_types.h
```diff
@@ -0,0 +1,83 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+/*
+ * Types used in LLVM compilation that are abstractions of the VM types
+*/
```
```suggestion
//
// Types used in LLVM compilation that are abstractions of the VM types.
//
```

src/coreclr/jit/llvm_types.h
```diff
@@ -0,0 +1,83 @@
...
+
+#ifndef _LLVM_TYPES_H_
+#define _LLVM_TYPES_H_
+
+#include "corinfo.h"
```
SingleAccretion: It is unusual to directly include `corinfo.h` in Jit code. Would including `jitpch.h` not work?

src/coreclr/jit/llvm_types.h
```diff
@@ -0,0 +1,83 @@
...
+    size_t                fieldCount;
+    CORINFO_FIELD_HANDLE* fields;
+};
+
+struct FieldDesc
+{
+private:
+    int                  _fieldOffset;
+    CORINFO_CLASS_HANDLE _classHandle;
+    CorInfoType          _corType;
```
SingleAccretion: Nit: the Jit coding convention is to have fields prefixed with `m_`: `m_fieldOffset`, etc.

src/coreclr/jit/llvm_types.h
```diff
@@ -0,0 +1,83 @@
...
+        _corType     = corType;
+        _classHandle = classHandle;
+    }
+};
+
+struct StructDesc
+{
+private:
+    size_t     _fieldCount;
+    FieldDesc* _fieldDesc;
```
SingleAccretion: Nit: `m_` prefix.

`_fieldDesc` naming seems off: it is an array, so should be `m_fieldDescs` (or `m_fields`)?

src/coreclr/jit/llvm_types.h
```diff
@@ -0,0 +1,83 @@
...
+
+public:
+    // deleting the StructDesc will also delete the fieldDesc array
```
```suggestion
    // This constructor takes the ownership of the passed in array of field descriptors.
```

src/coreclr/jit/llvm_types.h
```diff
@@ -0,0 +1,83 @@
...
+    {
+        return _classHandle;
+    }
+
+    CorInfoType getCorType()
+    {
+        return _corType;
+    }
+
+    void setFieldData(int fieldOffset, CorInfoType corType, CORINFO_CLASS_HANDLE classHandle)
```
SingleAccretion: It would be cleaner to set these in the constructor and make the type fully immutable.

SingleAccretion (commented):
The OOM issue concerns me.

I will try and see what debugging it locally brings.

src/coreclr/jit/llvm.h
```diff
@@ -198,7 +200,9 @@ class Llvm
     CorInfoType toCorInfoType(var_types varType);
     CORINFO_CLASS_HANDLE tryGetStructClassHandle(LclVarDsc* varDsc);
     void visitNode(GenTree* node);
-    Value* Llvm::zextIntIfNecessary(Value* intValue);
+    Value* zextIntIfNecessary(Value* intValue);
+    TypeDescriptor getTypeDescriptor(CORINFO_CLASS_HANDLE structHandle);
```
```suggestion
```

src/coreclr/jit/llvm_types.h
yowl: ok, I copied GenTree https://github.com/dotnet/runtimelab/blob/ea61843f3125bba9e12c7844f70b0406883e6179/src/coreclr/jit/gentree.h#L857 but will change it.

src/coreclr/jit/llvm_types.h
yowl: Yes, I did think about that but as the structs are created `new StructDesc(fieldCount, new FieldDesc[fieldCount]);`  I didn't know how what would be the right pattern to avoid reassigning the `FieldDesc` in each array index.

SingleAccretion: I think it's fine to re-assign them, it is what `setFieldData` already effectively does.
```c
FieldDesc* fields = new FieldDesc[fieldCount]; // Will require a parameterless constructor

for (...) { fields[i] = StructDesc(offset, type, handle); }

StructDesc* structDesc = new StructDesc(fieldCount, fields);
```

src/coreclr/jit/llvm_types.h
SingleAccretion: As you may have noticed, the actual Jit code does not match up with the conventions :). But it's nice for at least new code to be uniform.

SingleAccretion:
As far as the OOM is concerned, I can definitely reproduce the significantly increased memory consumption locally. Yet to pinpoint the exact cause.

src/coreclr/jit/llvm_types.h
yowl: Yes, of course.

yowl:
That's good.  I wondered if I had a field count of -1 somewhere, but haven't checked that  yet.

yowl:
I wonder if its a forward reference for a struct missing because I notice the Release build is failing locally with 
```
Assertion failed: N->isResolved() && "Expected forward references to be resolved", file F:\llvm11\llvm-11.0.0.src\lib\Bitcode\Writer\BitcodeWriter.cpp, line 2059
```
And I added the comment :

```
                // Forward-declare the struct in case there's a reference to it in the fields.
                // This must be a named struct or LLVM hits a stack overflow
```

SingleAccretion:
The issue is definitely in `getLlvmTypeForStruct`. I am seeing lots of `llvm::Type`s in the memory profile, and reverting changes to that method fixes the problem.

src/coreclr/jit/llvm.cpp
```diff
@@ -159,6 +163,70 @@ unsigned Llvm::getElementSize(CORINFO_CLASS_HANDLE classHandle, CorInfoType corI
...
+
+            FieldDesc* fieldDesc = structDesc->getFieldDesc(fieldIx);
+
+            CORINFO_FIELD_HANDLE fieldHandle = structTypeDescriptor.fields[fieldIx];
+            CORINFO_CLASS_HANDLE fieldClassHandle = NO_CLASS_HANDLE;
+
+            fieldDesc->setFieldData(_info.compCompHnd->getFieldOffset(fieldHandle),
+                                    _info.compCompHnd->getFieldType(fieldHandle, &fieldClassHandle), fieldClassHandle);
+            fieldIx++;
+        }
```
SingleAccretion: It took some time waiting around in the debugger, but I did eventually find the cause behind the OOM issue...

Reproduction:
```cs
private static void Problem(DateTime v1, DateTimeOffset v2, DateTime* p1, DateTimeOffset* p2)
{
    *p1 = v1;
    *p2 = v2;
}
```
This code is using `structTypeDescriptor.fields[fieldIx]` to get the field handle, but at this point, the fields may have been reversed (such as the case for `DateTimeOffset` auto-layout struct), it should be using `sparseFields[fldOffset]` instead.

Also, seems like the `getFieldOffset` call is not necessary (we already have `fldOffset`)?

Also, I believe it is the case in C++ that the `getFieldType`'s result has to be put into its own local for this to have well-defined behavior, otherwise the compiler can reorder the arguments such that `fieldClassHandle` is evaluated before `getFieldType` has been called.

Edit: oh, and the failure mode was that we'd (try to) add a `uint.MaxValue` padding fields when building the LLVM type.

yowl: Thanks for looking at this.  And confirm it allows the tests to run in parallel again.  See what the CI makes of the release build now.  The LLVM was missing all the structs in the RyuJIT module so not sure it is related.

yowl:
I'm not sure that
```
llvm::Type* Llvm::getLlvmTypeForStruct(CORINFO_CLASS_HANDLE structHandle)
{
    if (_llvmStructs->find(structHandle) == _llvmStructs->end())
```
is doing what I expect in release configuration.  If I print the struct name inside that if, in the `default` switch block:
```

            default:
                // Forward-declare the struct in case there's a reference to it in the fields.
                // This must be a named struct or LLVM hits a stack overflow
                const char* name = _getTypeName(_thisPtr, structHandle);
                printf("name %s\n", name);
```
And run the compiler on one of the test `rsp` files, then I see things like
```
name [S.P.CoreLib]System.DateTime
name [S.P.CoreLib]System.ValueTuple`2<int32,int32>
name [S.P.CoreLib]System.ValueTuple`2<int32,int32>
name [S.P.CoreLib]System.ValueTuple`2<int64,System.DateTime>
name [S.P.CoreLib]System.ValueTuple`2<int64,System.DateTime>
name [S.P.CoreLib]System.DateTime
```
I would not expect to see the names repeated but a casual glance at the output, it looks like every type appears twice.  I'll have to see if there is more than one thread running, or a problem with the hash map.

Edit. `getLlvmTypeForStruct` is throwing an exception before getting to add the struct to the map, hence it is processing each one multiple times.

yowl:
@SingleAccretion The release build is looking green now.  I had a `long long` mismatched with an `int`.  

Thanks for reviewing.

SingleAccretion (approved):
Looks great!

Issue #1860, opened 2022-03-08
[NativeAOT-LLVM] [Question] CT_INDIRECT (revisited)


yowl:
We still dont have support for `CT_INDIRECT` calls and one of the reasons why is that I don't know how to get the `CORINFO_SIG_INFO`.  The `gtCallMethHnd` is the call address node so `eeGetMethodSig` is out.  There are other methods like `eeGetSig` and `eeGetCallSiteSig` but they need a token.  We currently need the `CORINFO_SIG_INFO` in order to `lowerCallToShadowStack`  .  Are we going to have to try to rewrite `lowerCallToShadowStack` to not require a `CORINFO_SIG_INFO` or is there a way to get one from the `CT_INDIRECT` node?

Thanks

jkotas:
`CORINFO_SIG_INFO` is available when `CT_INDIRECT` call is imported in `impImportIndirectCall`. Update `impImportIndirectCall` to store `CORINFO_SIG_INFO` so that it is available in lowering where you need it?

yowl:
That might work, but `CORINFO_SIG_INFO ` is quite big, so I can't just add it as a struct to `GenTree` as it exceeds `TREE_NODE_SZ_LARGE` and I think there is a concern in making `GenTree` much larger.  If I just assign the pointer then it goes out of scope.

jkotas:
Yes, it would probably need to be a separate struct and GenTree would have to have a pointer to it.

yowl:
Righto, let me try that.  Thanks

yowl:
Or maybe I should just `|| defined(TARGET_WASM)`  at

https://github.com/dotnet/runtimelab/blob/258c8cee8cec40849127995fee955c117ced3d42/src/coreclr/jit/gentree.h#L4214-L4217

As that seems to be doing just this:

https://github.com/dotnet/runtimelab/blob/258c8cee8cec40849127995fee955c117ced3d42/src/coreclr/jit/importer.cpp#L9230-L9235

PR #1863, opened 2022-03-09
[NativeAOT-LLVM] Add support CT_INDIRECT calls

### LINKED ISSUE #1862: [NativeAOT-LLVM] [Question] consuming return values from CT_INDIRECT
Looking at this IR:
```
                                                  /--*  t84    int    arg-1
                                                  +--*  t32    int    calli tgt
N006 ( 26, 11) [000029] --CXG-------        t29 = *  CALL ind  int
                                                  /--*  t29    int
N008 ( 30, 14) [000027] DA-XG-------              *  STORE_LCL_VAR int    V05 tmp3         d:4```
```
`t29` is stored in the sdsu map as an `i8`, and then fails to be consumed to an `int` for the `STORE_LCL_VAR`.

@SingleAccretion, previously you wrote

> To maintain the invariant under 1 (only exact types stored in locations), we are to pass the type of the local to consumeValue (and ignore the node type).

Is this the erroneous logic in `consumeValue` at

https://github.com/dotnet/runtimelab/blob/258c8cee8cec40849127995fee955c117ced3d42/src/coreclr/jit/llvm.cpp#L923

Should that be using `targetLlvmType` which comes from the type of the local?

yowl:
This PR adds support for CT_INDIRECT calls.  It promotes `GenTree`'s `callSig` to a non DEBUG only field for `TARGET_WASM`   Doesn't add a huge amount of methods to the clrjit coverage, about a 1% increase.

cc @SingleAccretion

Closes #1862  #1860

### DIFF
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -8799,7 +8799,11 @@ GenTreeCall* Compiler::gtCloneExprCallHelper(GenTreeCall* tree,
     // we only really need one physical copy of it. Therefore a shallow pointer copy will suffice.
     // (Note that this still holds even if the tree we are cloning was created by an inlinee compiler,
     // because the inlinee still uses the inliner's memory allocator anyway.)
+#if TARGET_WASM
+    copy->callSig = tree->callSig;
+#else
     INDEBUG(copy->callSig = tree->callSig;)
+#endif // TARGET_WASM
 
     // The tail call info does not change after it is allocated, so for the same reasons as above
     // a shallow copy suffices.
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -8799,11 +8799,9 @@ GenTreeCall* Compiler::gtCloneExprCallHelper(GenTreeCall* tree,
     // we only really need one physical copy of it. Therefore a shallow pointer copy will suffice.
     // (Note that this still holds even if the tree we are cloning was created by an inlinee compiler,
     // because the inlinee still uses the inliner's memory allocator anyway.)
-#if TARGET_WASM
+#if defined(DEBUG) || defined(TARGET_WASM)
     copy->callSig = tree->callSig;
-#else
-    INDEBUG(copy->callSig = tree->callSig;)
-#endif // TARGET_WASM
+#endif // DEBUG || TARGET_WASM
 
     // The tail call info does not change after it is allocated, so for the same reasons as above
     // a shallow copy suffices.
--- a/src/coreclr/jit/importer.cpp
+++ b/src/coreclr/jit/importer.cpp
@@ -8645,8 +8645,9 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
                 assert((sig->callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_VARARG &&
                        (sig->callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_NATIVEVARARG);
 
+#if TARGET_WASM
                 call->AsCall()->callSig = new (this, CMK_Generic) CORINFO_SIG_INFO(*sig);
-
+#endif
                 goto DONE;
             }
 
--- a/src/coreclr/jit/importer.cpp
+++ b/src/coreclr/jit/importer.cpp
@@ -8645,10 +8645,8 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
                 assert((sig->callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_VARARG &&
                        (sig->callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_NATIVEVARARG);
 
-#if TARGET_WASM // TODO-LLVM
-                call->AsCall()->callSig  = new (this, CMK_Generic) CORINFO_SIG_INFO;
-                *call->AsCall()->callSig = *sig;
-#endif 
+                call->AsCall()->callSig = new (this, CMK_Generic) CORINFO_SIG_INFO(*sig);
+
                 goto DONE;
             }
 
@@ -8744,8 +8742,7 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
 
 #if TARGET_WASM
     assert(call->AsCall()->callSig == nullptr);
-    call->AsCall()->callSig  = new (this, CMK_Generic) CORINFO_SIG_INFO;
-    *call->AsCall()->callSig = *sig;
+    call->AsCall()->callSig = new (this, CMK_Generic) CORINFO_SIG_INFO(*sig);
 #endif
 
     /*-------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -919,8 +919,10 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
                nodeValue->getType()->getPrimitiveSizeInBits() <= 32 && targetLlvmType->getPrimitiveSizeInBits() <= 32);
         if (nodeValue->getType()->getPrimitiveSizeInBits() < targetLlvmType->getPrimitiveSizeInBits())
         {
+            var_types trueNodeType = node->OperIs(GT_CALL) ? static_cast<var_types>(node->AsCall()->gtReturnType) : node->TypeGet();
```
SingleAccretion: I think we can convert the whole chain of `if ... else`s in the method to this now:
```cs
var_types trueNodeType = TYP_UNDEF;
switch (node->OperGet())
    case GT_CALL: { };
    case GT_LCL_VAR: { };
    case GT_[RELOPS]: { };
    default: trueNodeType = node->TypeGet();

assert(varTypeIsSmall(trueNodeType));

finalValue = varTypeIsSigned(trueNodeType) ? _builder.CreateSExt(nodeValue, targetLlvmType)
                                           : _builder.CreateZExt(nodeValue, targetLlvmType);
```
(Also, right now the case under `// Upcast.` is using `varTypeIsSigned(node)`, it should use `varTypeIsSigned(trueNodeType)`)

src/coreclr/jit/llvm.cpp
```diff
@@ -2399,12 +2400,13 @@ void Llvm::lowerToShadowStack()
...
                 }
 
-                CORINFO_SIG_INFO calleeSigInfo;
-                _compiler->eeGetMethodSig(callNode->gtCallMethHnd, &calleeSigInfo);
-
-                failUnsupportedCalls(callNode, &calleeSigInfo);
+                if (callNode->gtCallType == CT_INDIRECT)
+                {
+                    unsigned i = 1;
+                }
```
```suggestion
```

src/coreclr/jit/llvm.cpp
yowl: I made a change here .  For Relops I did that in the default block, not seeing a way to make it a `case`.

I also added a couple of more sets for the `callSig`.  The clone was the one breaking the release tests I think.  The `CORINFO_VIRTUALCALL_LDVIRTFTN` case also looked like it needs it.

src/coreclr/jit/llvm.cpp

src/coreclr/jit/gentree.cpp
```diff
@@ -8799,7 +8799,11 @@ GenTreeCall* Compiler::gtCloneExprCallHelper(GenTreeCall* tree,
     // we only really need one physical copy of it. Therefore a shallow pointer copy will suffice.
     // (Note that this still holds even if the tree we are cloning was created by an inlinee compiler,
     // because the inlinee still uses the inliner's memory allocator anyway.)
+#if TARGET_WASM
+    copy->callSig = tree->callSig;
+#else
     INDEBUG(copy->callSig = tree->callSig;)
+#endif // TARGET_WASM
```
SingleAccretion:
```suggestion
#if defined(DEBUG) || defined(TARGET_WASM)
    copy->callSig = tree->callSig;
#endif // DEBUG || TARGET_WASM
```

src/coreclr/jit/llvm.cpp
```diff
@@ -919,28 +919,40 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
...
+                    {
+                        // This is the special case for relops. Ordinary codegen "just knows" they need zero-extension.
+                        assert(nodeValue->getType() == Type::getInt1Ty(_llvmContext));
+                        trueNodeType = TYP_BOOL;
+                    }
+                    else
+                    {
+                        trueNodeType = node->TypeGet();
+                    }
+                }
```
```suggestion
                case GT_CALL:
                    trueNodeType = static_cast<var_types>(node->AsCall()->gtReturnType);
                    break;

                case GT_LCL_VAR:
                    trueNodeType = _compiler->lvaGetDesc(node->AsLclVarCommon())->TypeGet();
                    break;

                case GT_EQ:
                case GT_NE:
                case GT_LT:
                case GT_LE:
                case GT_GE:
                case GT_GT:
                    // This is the special case for relops. Ordinary codegen "just knows" they need zero-extension.
                    assert(nodeValue->getType() == Type::getInt1Ty(_llvmContext));
                    trueNodeType = TYP_UBYTE;
                    break;

                default:
                    trueNodeType = node->TypeGet();
                    break;
```
(The relops can be just listed explicitly, we'll not be adding new ones any time soon)

src/coreclr/jit/importer.cpp
```diff
@@ -8644,6 +8644,11 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
                 // Sine we are jumping over some code, check that its OK to skip that code
                 assert((sig->callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_VARARG &&
                        (sig->callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_NATIVEVARARG);
+
+#if TARGET_WASM // TODO-LLVM
+                call->AsCall()->callSig  = new (this, CMK_Generic) CORINFO_SIG_INFO;
+                *call->AsCall()->callSig = *sig;
```
SingleAccretion:
```suggestion
                call->AsCall()->callSig  = new (this, CMK_Generic) CORINFO_SIG_INFO(*sig);
```
(It *should* work, though I did not test it)

src/coreclr/jit/importer.cpp
```diff
@@ -8644,6 +8644,11 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
                 // Sine we are jumping over some code, check that its OK to skip that code
                 assert((sig->callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_VARARG &&
                        (sig->callConv & CORINFO_CALLCONV_MASK) != CORINFO_CALLCONV_NATIVEVARARG);
+
+#if TARGET_WASM // TODO-LLVM
```
SingleAccretion: Why the `TODO`?

src/coreclr/jit/importer.cpp
```diff
@@ -8737,6 +8742,12 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
     assert(((mflags & CORINFO_FLG_STATIC) != 0) == ((sig->callConv & CORINFO_CALLCONV_HASTHIS) == 0));
     assert(call != nullptr);
 
+#if TARGET_WASM
+    assert(call->AsCall()->callSig == nullptr);
+    call->AsCall()->callSig  = new (this, CMK_Generic) CORINFO_SIG_INFO;
+    *call->AsCall()->callSig = *sig;
```
SingleAccretion:
```suggestion
    call->AsCall()->callSig  = new (this, CMK_Generic) CORINFO_SIG_INFO(*sig);
```

SingleAccretion (approved):
Just some code simplification suggestions, otherwise LGTM.

src/coreclr/jit/importer.cpp
yowl: Dont remember, I'll remove it.

src/coreclr/jit/importer.cpp
yowl: Thanks, looks fine here.

src/coreclr/jit/llvm.cpp

yowl:
Oops, removed the #ifs by mistake

jkotas (approved):
Thank you!

PR #1864, opened 2022-03-10
[NativeAOT-LLVM] FlushProcessWriteBuffers to a NOP

### LINKED ISSUE #1861: [NativeAOT-LLVM] Could use a stub for __syscall_mprotect
There's no concept of this, at least currently, for wasm, so NativeAOT-LLVM could use a stub to prevent the warnings:

```
warning: unsupported syscall: __syscall_mprotect
```

yowl:
This PR makes `FlushProcessWriteBuffers` a NOP for `TARGET_WASM` and removes a load of warnings about missing syscall for mprotect.

Closes #1861 

Thanks @jkotas 
cc @SingleAccretion

### DIFF
--- a/src/coreclr/gc/unix/gcenv.unix.cpp
+++ b/src/coreclr/gc/unix/gcenv.unix.cpp
@@ -515,10 +515,10 @@ bool GCToOSInterface::CanGetCurrentProcessorNumber()
     return HAVE_SCHED_GETCPU;
 }
 
+#if !TARGET_WASM
 // Flush write buffers of processors that are executing threads of the current process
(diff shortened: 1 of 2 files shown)


SingleAccretion:
For some other functions in the file, they've been `#if`-ed out completely and an implementation provided in [gcenv.wasm.cpp](https://github.com/dotnet/runtimelab/blob/27876060d64632edcfdbc88d461c83a4bebbd293/src/coreclr/gc/wasm/gcenv.wasm.cpp).

Should/could `FlushProcessWriteBuffers` follow the same pattern?

yowl:
Yes, well spotted.

jkotas (approved):

PR #1866, opened 2022-03-11
[NativeAOT-LLVM] Support for storing locals of type struct

yowl:
This PR adds support for getting the LLVM type for locals of type `TYP_STRUCT` and moves local structs where an implicit cast appears to the LLVM frame.

(I believe `LCL_FLD` will be similar, but not done that here)

BTW, this enables the `TestStructStoreWithSignificantPadding` test from HelloWasm in the RyuJit module.

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1766,14 +1766,13 @@ bool Llvm::isLlvmFrameLocal(LclVarDsc* varDsc)
 
 void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
 {
-    Type* destLlvmType = getLlvmTypeForLclVar(lclVar);
-
-    Value*   localValue;
+    Type*  destLlvmType = getLlvmTypeForLclVar(lclVar);
+    Value* localValue   = nullptr;
 
     // zero initialization check
-    if (lclVar->TypeGet() == TYP_STRUCT && lclVar->gtGetOp1()->IsIntegralConst(0))
+    if (lclVar->TypeIs(TYP_STRUCT) && lclVar->gtGetOp1()->IsIntegralConst(0))
     {
-        localValue = llvm::Constant::getNullValue(getLlvmTypeForStruct(_compiler->lvaGetDesc(lclVar)->GetLayout()));
+        localValue = llvm::Constant::getNullValue(destLlvmType);
     }
     else
     {
@@ -2431,7 +2430,7 @@ void Llvm::convertLclStructToLoad(GenTreeLclVarCommon* lclNode, ClassLayout* cls
     GenTree* lclAddr = _compiler->gtNewLclVarAddrNode(lclNode->GetLclNum());
 
     lclNode->ChangeOper(GT_OBJ);
-    lclNode->gtOp1 = lclAddr;
+    lclNode->AsObj()->SetAddr(lclAddr);
     lclNode->AsObj()->SetLayout(clsLayout);
 
     _currentRange->InsertBefore(lclNode, lclAddr);
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -925,17 +925,6 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
             return _builder.CreateBitCast(nodeValue, targetLlvmType);
         }
 
-        // implicit struct cast from alloca
-        if (nodeValue->getType()->isStructTy())
-        {
-            GenTreeLclVarCommon* lclVar = node->AsLclVarCommon();
-            if (isLlvmFrameLocal(_compiler->lvaGetDesc(lclVar)))
-            {
-                return _builder.CreateLoad(
-                    _builder.CreateBitCast(m_allocas[lclVar->GetLclNum()], targetLlvmType->getPointerTo()));
-            }
-        }
-
         // int and smaller int conversions
         assert(targetLlvmType->isIntegerTy() && nodeValue->getType()->isIntegerTy() &&
                nodeValue->getType()->getPrimitiveSizeInBits() <= 32 && targetLlvmType->getPrimitiveSizeInBits() <= 32);
@@ -2420,6 +2409,17 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
     }
 }
 
+void Llvm::convertLclStructToLoad(GenTreeLclVarCommon* lclNode, ClassLayout* clsLayout)
+{
+    GenTree* lclAddr = _compiler->gtNewLclVarAddrNode(lclNode->GetLclNum(), TYP_STRUCT);
+
+    lclNode->ChangeOper(GT_OBJ);
+    lclNode->gtOp1 = lclAddr;
+    lclNode->AsObj()->SetLayout(clsLayout);
+
+    _currentRange->InsertBefore(lclNode, lclAddr);
+}
+
 void Llvm::lowerToShadowStack()
 {
     for (BasicBlock* _currentBlock : _compiler->Blocks())
@@ -2475,7 +2475,7 @@ void Llvm::lowerToShadowStack()
                 _compiler->lvaGetDesc(node->AsLclVarCommon())->lvHasLocalAddr = 1;
             }
             // check for implicit struct cast
-            else if (node->OperIsLocalStore())
+            else if (node->OperIs(GT_STORE_LCL_VAR))
             {
                 GenTreeLclVarCommon* addrLcl = node->AsLclVarCommon();
                 GenTree* dataOp  = addrLcl->gtGetOp1();
@@ -2492,6 +2492,8 @@ void Llvm::lowerToShadowStack()
                             // mark both as requiring lvHasLocalAddr as we will be doing a bitcast on the pointers
                             addrVarDsc->lvHasLocalAddr = 1;
                             dataVarDsc->lvHasLocalAddr = 1;
+
+                            convertLclStructToLoad(dataLcl, addrVarDsc->GetLayout());
                         }
                     }
                 }
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1227,7 +1227,24 @@ void Llvm::buildCnsInt(GenTree* node)
 {
     if (node->gtType == TYP_INT)
     {
-        mapGenTreeToValue(node, _builder.getInt32(node->AsIntCon()->IconValue()));
+        if (node->IsIconHandle())
+        {
+            if (node->IsIconHandle(GTF_ICON_TOKEN_HDL))
+            {
+                const char* symbolName = (*_getMangledSymbolName)(_thisPtr, (void*)(node->AsIntCon()->IconValue()));
+                (*_addCodeReloc)(_thisPtr, (void*)node->AsIntCon()->IconValue());
+                mapGenTreeToValue(node, _builder.CreateLoad(getOrCreateExternalSymbol(symbolName)));
+            }
+            else
+            {
+                //TODO-LLVML: other ICON handle types
+                failFunctionCompilation();
+            }
+        }
+        else
+        {
+            mapGenTreeToValue(node, _builder.getInt32(node->AsIntCon()->IconValue()));
+        }
         return;
     }
     if (node->gtType == TYP_REF)
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -2452,20 +2452,17 @@ void Llvm::lowerStoreLcl(GenTreeLclVarCommon* storeLclNode)
             else if (dataOp->OperIs(GT_LCL_VAR)) // can get icon 0 here
             {
                 GenTreeLclVarCommon* dataLcl = dataOp->AsLclVarCommon();
-                if (dataLcl->TypeGet() == TYP_STRUCT)
-                {
-                    LclVarDsc* dataVarDsc = _compiler->lvaGetDesc(dataLcl->GetLclNum());
+                LclVarDsc* dataVarDsc = _compiler->lvaGetDesc(dataLcl->GetLclNum());
 
-                    dataVarDsc->lvHasLocalAddr = 1;
+                dataVarDsc->lvHasLocalAddr = 1;
 
-                    GenTree* dataAddrNode = _compiler->gtNewLclVarAddrNode(dataLcl->GetLclNum());
+                GenTree* dataAddrNode = _compiler->gtNewLclVarAddrNode(dataLcl->GetLclNum());
 
-                    dataLcl->ChangeOper(GT_OBJ);
-                    dataLcl->AsObj()->SetAddr(dataAddrNode);
-                    dataLcl->AsObj()->SetLayout(addrVarDsc->GetLayout());
+                dataLcl->ChangeOper(GT_OBJ);
+                dataLcl->AsObj()->SetAddr(dataAddrNode);
+                dataLcl->AsObj()->SetLayout(addrVarDsc->GetLayout());
 
-                    _currentRange->InsertBefore(dataLcl, dataAddrNode);
-                }
+                CurrentRange().InsertBefore(dataLcl, dataAddrNode);
             }
         }
     }
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -2497,7 +2497,7 @@ void Llvm::lowerToShadowStack()
                         }
                     }
                 }
-                else if (dataOp->OperIs(GT_IND))
+                else if (dataOp->OperIs(GT_IND) && addrLcl->TypeGet() == TYP_STRUCT && dataOp->TypeGet() == TYP_STRUCT)
                 {
                     dataOp->ChangeOper(GT_OBJ);
                     LclVarDsc* addrVarDsc = _compiler->lvaGetDesc(addrLcl->GetLclNum());
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -2497,6 +2497,12 @@ void Llvm::lowerToShadowStack()
                         }
                     }
                 }
+                else if (dataOp->OperIs(GT_IND))
+                {
+                    dataOp->ChangeOper(GT_OBJ);
+                    LclVarDsc* addrVarDsc = _compiler->lvaGetDesc(addrLcl->GetLclNum());
+                    dataOp->AsObj()->SetLayout(addrVarDsc->GetLayout());
+                }
             }   
         }
     }
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -2411,7 +2411,7 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
 
 void Llvm::convertLclStructToLoad(GenTreeLclVarCommon* lclNode, ClassLayout* clsLayout)
 {
-    GenTree* lclAddr = _compiler->gtNewLclVarAddrNode(lclNode->GetLclNum(), TYP_STRUCT);
+    GenTree* lclAddr = _compiler->gtNewLclVarAddrNode(lclNode->GetLclNum());
 
     lclNode->ChangeOper(GT_OBJ);
     lclNode->gtOp1 = lclAddr;
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -2427,49 +2427,44 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
     }
 }
 
-void Llvm::convertLclStructToLoad(GenTreeLclVarCommon* lclNode, ClassLayout* clsLayout)
-{
-    GenTree* lclAddr = _compiler->gtNewLclVarAddrNode(lclNode->GetLclNum());
-
-    lclNode->ChangeOper(GT_OBJ);
-    lclNode->AsObj()->SetAddr(lclAddr);
-    lclNode->AsObj()->SetLayout(clsLayout);
-
-    _currentRange->InsertBefore(lclNode, lclAddr);
-}
-
 void Llvm::lowerStoreLcl(GenTreeLclVarCommon* storeLclNode)
 {
-    GenTree* dataOp = storeLclNode->gtGetOp1();
-    CORINFO_CLASS_HANDLE dataHandle = _compiler->gtGetStructHandleIfPresent(dataOp);
-    if (dataOp->OperIs(GT_IND))
-    {
-        // Special case: "gtGetStructHandleIfPresent" sometimes guesses the handle from
-        // field sequences, but we will always need to transform TYP_STRUCT INDs into OBJs.
-        dataHandle = NO_CLASS_HANDLE;
-    }
-
     LclVarDsc* addrVarDsc = _compiler->lvaGetDesc(storeLclNode->GetLclNum());
 
-    if (storeLclNode->TypeIs(TYP_STRUCT) && addrVarDsc->GetStructHnd() != dataHandle)
+    if (storeLclNode->TypeIs(TYP_STRUCT))
     {
-        if (dataOp->OperIsIndir())
+        GenTree* dataOp = storeLclNode->gtGetOp1();
+        CORINFO_CLASS_HANDLE dataHandle = _compiler->gtGetStructHandleIfPresent(dataOp);
+        if (dataOp->OperIs(GT_IND))
         {
-            dataOp->SetOper(GT_OBJ);
-            dataOp->AsObj()->SetLayout(addrVarDsc->GetLayout());
+            // Special case: "gtGetStructHandleIfPresent" sometimes guesses the handle from
+            // field sequences, but we will always need to transform TYP_STRUCT INDs into OBJs.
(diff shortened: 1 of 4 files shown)


yowl:
Wasm tests are green, but OSX failed with a restore error, so will retry

yowl:
Maybe I reopened this too soon... and the previous win x64 job was still running

yowl:
Hmm, I don't know why this job just seems to start as Cancelled .  If I try to look into the pipeline I get

image

@jkotas any ideas?

jkotas:
> @jkotas any ideas?

It looks fine to me know. Intermittent infrastructure issue probably.

jkotas:
@SingleAccretion Could you please review this change?

src/coreclr/jit/llvm.cpp
```diff
@@ -2435,11 +2468,34 @@ void Llvm::lowerToShadowStack()
...
                 CurrentRange().InsertBefore(node, retAddressLocal, storeNode);
             }
+
             if (node->OperIsLocalAddr())
             {
                 // Indicates that this local is to live on the LLVM frame, and will not participate in SSA.
                 _compiler->lvaGetDesc(node->AsLclVarCommon())->lvHasLocalAddr = 1;
             }
+            // check for implicit struct cast
+            else if (node->OperIsLocalStore())
```
SingleAccretion: This should be `OperIs(GT_STORE_LCL_VAR)` for now. `STORE_LCL_FLD` is not supported for `TYP_STRUCT` currently.

That aside, I think this should be structured a little differently to avoid handling mismatched struct types in 
codegen -- convert the `LCL_VAR` source to `OBJ(LCL_VAR_ADDR)`, for `IND` sources, convert them to `OBJ<dest struct>` (same with mismatched `OBJ/BLK`).

yowl: Thanks. 
I've removed the codegen conversion for `LCL_VAR` source.  I don't seem to have any failures after that.  For the `IND` sources problem, is there a test that you can think of that would exercise it?

SingleAccretion: > is there a test that you can think of that would exercise it

Something like this should hit it:
```cs
private static StructWithIndex Problem(StructWithStructWithIndex* p, StructWithIndex b)
{
    b = p->StructWithIndex;
    JitUse(&b);

    return b;
}

struct StructWithIndex
{
    public int Index;
    public int Value;
}

struct StructWithStructWithIndex
{
    public StructWithIndex StructWithIndex;
    public int AnotherIndex;
}

public static void JitUse<T>(T* arg) where T : unmanaged { }
```
Basically, the Jit likes to dissolve `TYP_STRUCT` nodes into layout-less `IND TYP_STRUCT` nodes that only the parent nodes will know the true type of. This is most prevalent for the RHSs of struct stores (aka `OperIsCopyBlkOp`).

yowl: Thanks.  that looks good to me.  Before:
```
N001 (  3,  2) [000002] ------------         t2 =    LCL_VAR   int    V00 arg0         u:1 (last use) Zero Fseq[StructWithIndex]
                                                  /--*  t2     int
N002 (  6,  4) [000003] D--XG-------         t3 = *  IND       struct
                                                  /--*  t3     struct
N004 ( 10,  7) [000006] DA-XG-------              *  STORE_LCL_VAR struct<Program+StructWithIndex, 8>(AX) V01 arg1
```
After with latest change:
```
N001 (  3,  2) [000002] ------------         t2 =    LCL_VAR   int    V00 arg0         u:1 (last use) Zero Fseq[StructWithIndex]
                                                  /--*  t2     int
N002 (  6,  4) [000003] D--XG-------         t3 = *  OBJ       struct<Program+StructWithIndex, 8>
                                                  /--*  t3     struct
N004 ( 10,  7) [000006] DA-XG-------              *  STORE_LCL_VAR struct<Program+StructWithIndex, 8>(AX) V01 arg1
```

We can't handle return types of struct, so it fails the actual codegen at the moment, but the test is in at least.

yowl:
Looks to be corrupting the vslot tables.  Either due to a latent problem or this change.  So far tried -fsanitize=undefined and -fsanitize=address, but not luck yet finding the cause.

src/coreclr/jit/llvm.cpp
```diff
@@ -2387,6 +2409,17 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
     }
 }
 
+void Llvm::convertLclStructToLoad(GenTreeLclVarCommon* lclNode, ClassLayout* clsLayout)
+{
+    GenTree* lclAddr = _compiler->gtNewLclVarAddrNode(lclNode->GetLclNum(), TYP_STRUCT);
```
SingleAccretion: Doubtful this is the corruption cause, but this should be `TYP_I_IMPL`, not `TYP_STRUCT`.

(Or actually omitted -- I believe `gtNewLclVarAddrNode` has `TYP_I_IMPL` as the default parameter already).
```suggestion
    GenTree* lclAddr = _compiler->gtNewLclVarAddrNode(lclNode->GetLclNum());
```

yowl: Yes, you are right `GenTreeLclVar* gtNewLclVarAddrNode(unsigned lclNum, var_types type = TYP_I_IMPL);`  Thanks

yowl:
`String__FastAllocateString` looks like the problem

Its not handling
```
CNS_INT(h) int    0x420078 token
```
I think the fix is something like
```c++
        if (node->IsIconHandle())
        {
            const char* symbolName = (*_getMangledSymbolName)(_thisPtr, (void*)(node->AsIntCon()->IconValue()));
            (*_addCodeReloc)(_thisPtr, (void*)node->AsIntCon()->IconValue());
            mapGenTreeToValue(node, _builder.CreateLoad(getOrCreateExternalSymbol(symbolName)));
        }
```
but maybe not all Icon handles, as i'm getting
```
wasm-ld: error: symbol type mismatch: S_P_TypeLoader_Internal_Runtime_TypeLoader_ModuleInfo__ResolveTypeSlotDispatch
>>> defined as WASM_SYMBOL_TYPE_FUNCTION in E:\GitHub\runtimelab\artifacts\tests\coreclr\Browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.o
>>> defined as WASM_SYMBOL_TYPE_DATA in E:\GitHub\runtimelab\artifacts\tests\coreclr\Browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasmclrjit.o
```

yowl:
Added fix for memory corruption.  RyuJIT coverages drops by around 7% with this but at least it runs.  Funny how a seemingly catastrophic error can go undetected for so long.

src/coreclr/jit/llvm.cpp
```diff
@@ -2435,11 +2485,42 @@ void Llvm::lowerToShadowStack()
...
                 CurrentRange().InsertBefore(node, retAddressLocal, storeNode);
             }
+
             if (node->OperIsLocalAddr())
             {
                 // Indicates that this local is to live on the LLVM frame, and will not participate in SSA.
                 _compiler->lvaGetDesc(node->AsLclVarCommon())->lvHasLocalAddr = 1;
             }
+            // check for implicit struct cast
+            else if (node->OperIs(GT_STORE_LCL_VAR))
```
SingleAccretion, line 2525: I think we should start to create more structure to the lowering, wherein we have one `lowerNode` method with a `switch` that calls the individual lowering methods. Let's start by outlining this into `lowerStoreLcl`.

src/coreclr/jit/llvm.cpp
```diff
@@ -2435,11 +2485,42 @@ void Llvm::lowerToShadowStack()
...
+                    GenTreeLclVarCommon* dataLcl = dataOp->AsLclVarCommon();
+                    if (addrLcl->TypeGet() == TYP_STRUCT && dataLcl->TypeGet() == TYP_STRUCT)
+                    {
+                        LclVarDsc* addrVarDsc = _compiler->lvaGetDesc(addrLcl->GetLclNum());
+                        LclVarDsc* dataVarDsc = _compiler->lvaGetDesc(dataLcl->GetLclNum());
+
+                        if (tryGetStructClassHandle(addrVarDsc) != tryGetStructClassHandle(dataVarDsc))
+                        {
+                            // mark both as requiring lvHasLocalAddr as we will be doing a bitcast on the pointers
+                            addrVarDsc->lvHasLocalAddr = 1;
```
SingleAccretion: Why mark the destination local as address-taken? It does not seem necessary.

src/coreclr/jit/llvm.cpp
```diff
@@ -2435,11 +2485,42 @@ void Llvm::lowerToShadowStack()
...
+                            dataVarDsc->lvHasLocalAddr = 1;
+
+                            convertLclStructToLoad(dataLcl, addrVarDsc->GetLayout());
+                        }
+                    }
+                }
+                else if (dataOp->OperIs(GT_IND) && addrLcl->TypeGet() == TYP_STRUCT && dataOp->TypeGet() == TYP_STRUCT)
```
SingleAccretion: We can have `OBJ` or `BLK` mismatches here too. You could structure it like this:
```c
CORINFO_CLASS_HANDLE dataHandle = _compiler->gtGetStructHandleIfPresent(dataOp);
if (dataOp->OperIs(GT_IND))
{
    // Special case: "gtGetStructHandleIfPresent" sometimes guesses the handle from
    // field sequences, but we will always need to transform TYP_STRUCT INDs into OBJs.
    dataHandle = NO_CLASS_HANDLE;
}

if (addrVarDsc->GetStructHnd() != dataHandle)
{
    if (OperIsIndir()) { /* bash to OBJ */ }
    else { assert(dataOp->OperIs(GT_LCL_VAR) /* LCL_VAR transform */ }
}
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2435,11 +2485,42 @@ void Llvm::lowerToShadowStack()
...
+                {
+                    dataOp->ChangeOper(GT_OBJ);
```
SingleAccretion: This should be `SetOper`.

The difference between `SetOper` and `ChangeOper` is that the former doesn't clear "node-specific" flags. So, generally, when you want to bash to a completely different oper, `ChangeOper` should be used, and if the opers are of the same class (defined loosely as accepting the same flags) - `SetOper`, s. t. the flags aren't lost.

src/coreclr/jit/llvm.cpp
```diff
@@ -2387,6 +2426,17 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
     }
 }
 
+void Llvm::convertLclStructToLoad(GenTreeLclVarCommon* lclNode, ClassLayout* clsLayout)
+{
+    GenTree* lclAddr = _compiler->gtNewLclVarAddrNode(lclNode->GetLclNum());
+
+    lclNode->ChangeOper(GT_OBJ);
+    lclNode->gtOp1 = lclAddr;
```
```suggestion
    lclNode->AsObj()->SetAddr(lclAddr);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1738,8 +1766,19 @@ bool Llvm::isLlvmFrameLocal(LclVarDsc* varDsc)
...
-    Type* destLlvmType = getLlvmTypeForVarType(lclVar->TypeGet());
-    Value* localValue = consumeValue(lclVar->gtGetOp1(), destLlvmType);
+    Type* destLlvmType = getLlvmTypeForLclVar(lclVar);
+
+    Value*   localValue;
+
+    // zero initialization check
+    if (lclVar->TypeGet() == TYP_STRUCT && lclVar->gtGetOp1()->IsIntegralConst(0))
+    {
+        localValue = llvm::Constant::getNullValue(getLlvmTypeForStruct(_compiler->lvaGetDesc(lclVar)->GetLayout()));
```
```suggestion
        localValue = llvm::Constant::getNullValue(destLlvmType);
```
(Presumably the same thing)

src/coreclr/jit/llvm.cpp
```diff
@@ -1738,8 +1766,19 @@ bool Llvm::isLlvmFrameLocal(LclVarDsc* varDsc)
 
 void Llvm::storeLocalVar(GenTreeLclVar* lclVar)
 {
-    Type* destLlvmType = getLlvmTypeForVarType(lclVar->TypeGet());
-    Value* localValue = consumeValue(lclVar->gtGetOp1(), destLlvmType);
+    Type* destLlvmType = getLlvmTypeForLclVar(lclVar);
+
+    Value*   localValue;
```
```suggestion
    Type*  destLlvmType = getLlvmTypeForLclVar(lclVar);
    Value* localValue   = nullptr;
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1738,8 +1766,19 @@ bool Llvm::isLlvmFrameLocal(LclVarDsc* varDsc)
...
+
+    // zero initialization check
+    if (lclVar->TypeGet() == TYP_STRUCT && lclVar->gtGetOp1()->IsIntegralConst(0))
```
```suggestion
    if (lclVar->TypeIs(TYP_STRUCT) && lclVar->gtGetOp1()->IsIntegralConst(0))
```

src/coreclr/jit/llvm.cpp
```diff
@@ -656,6 +656,17 @@ Type* Llvm::getLlvmTypeForVarType(var_types type)
     }
 }
 
+Type* Llvm::getLlvmTypeForLclVar(GenTreeLclVarCommon* lclVar)
```
SingleAccretion: Hmm, this should either only accept `GenTreeLclVar`, or `assert(!lclVar->OperIsLocalField())` under `varType == TYP_STRUCT` - it won't work for local fields.

src/coreclr/jit/llvm.cpp
```diff
@@ -656,6 +656,17 @@ Type* Llvm::getLlvmTypeForVarType(var_types type)
...
+{
+    var_types varType = lclVar->TypeGet();
```
```suggestion
    var_types nodeType = lclVar->TypeGet();
```
Nit: the `varType` strongly implies we're dealing with `varDsc->TypeGet()` but actually that's not the case (they will be different for `LCL_FLD`s).

src/coreclr/jit/llvm.cpp
```diff
@@ -1216,7 +1227,24 @@ void Llvm::buildCnsInt(GenTree* node)
...
     if (node->gtType == TYP_INT)
     {
-        mapGenTreeToValue(node, _builder.getInt32(node->AsIntCon()->IconValue()));
+        if (node->IsIconHandle())
+        {
+            if (node->IsIconHandle(GTF_ICON_TOKEN_HDL))
+            {
+                const char* symbolName = (*_getMangledSymbolName)(_thisPtr, (void*)(node->AsIntCon()->IconValue()));
+                (*_addCodeReloc)(_thisPtr, (void*)node->AsIntCon()->IconValue());
+                mapGenTreeToValue(node, _builder.CreateLoad(getOrCreateExternalSymbol(symbolName)));
```
SingleAccretion, line 1238: Hmmm, I would like to understand -- what was the problem here?

yowl: We were producing this 
```
define void @String__FastAllocateString(i8* %0, i8* %1, i32 %2) !dbg !864 {
Prolog:
  %3 = alloca i32, align 4
  %4 = alloca i32, align 4
  br label %5

5:                                                ; preds = %Prolog
  br label %6

6:                                                ; preds = %5
  store i32 4345336, i32* %3, align 4, !dbg !865
  %7 = load i32, i32* %3, align 4, !dbg !865
  store i32 %7, i32* %4, align 4, !dbg !865
  %8 = load i32, i32* %4, align 4, !dbg !866
  %9 = getelementptr i8, i8* %0, i32 12, !dbg !866
  %10 = getelementptr i8, i8* %0, i32 12, !dbg !866
  %11 = ptrtoint i8* %10 to i32, !dbg !866
  %12 = inttoptr i32 %11 to i8*, !dbg !866
  call void @S_P_CoreLib_System_Runtime_RuntimeImports__RhNewString_0(i8* %9, i8* %12, i32 %8, i32 %2), !dbg !866
```

That number, `4345336` was the `@__EEType_String___SYMBOL` .  The IL backend was producing this for the same method:

```
define void @String__FastAllocateString(i8* %0, i8* %1, i32 %2) !dbg !476660 {
Prolog:
  %length_arg0_ = alloca i32, align 4
  store i32 %2, i32* %length_arg0_, align 4
  %Temp0_ = getelementptr i8, i8* %0, i32 8
  %Local0_ = getelementptr i8, i8* %0, i32 0
  %Local1_ = getelementptr i8, i8* %0, i32 4
  br label %Block0

Block0:                                           ; preds = %Prolog
  call void @llvm.donothing(), !dbg !476661
  %LoadAddressOfSymbolNode = load i32*, i32** @__EEType_String___SYMBOL, align 4, !dbg !476662
  %3 = getelementptr i8, i8* %0, i32 8, !dbg !476663
  %CastInteeTypePtr = ptrtoint i32* %LoadAddressOfSymbolNode to i32, !dbg !476663
  %Loadarg0_ = load i32, i32* %length_arg0_, align 4, !dbg !476663
  call void @S_P_CoreLib_System_Runtime_RuntimeImports__RhNewString_0(i8* %3, i8* %Temp0_, i32 %CastInteeTypePtr, i32 %Loadarg0_), !dbg !476663
```

SingleAccretion: I see. This icon handle is meant to be a `MethodTable*`. Unfortunate that it's got this generic `GTF_ICON_TOKEN_HDL` kind that is a byproduct of how importation for the intrinsic works.

I would think you can enable this for `GTF_ICON_CLASS_HDL`, `GTF_ICON_METHOD_HDL` and `GTF_ICON_FIELD_HDL` too.

I would also leave a `TODO-LLVM: consider lowering these to "IND(CLS_VAR_ADDR)"`.

yowl: Thanks, that has got back the coverage

src/coreclr/jit/llvm.cpp

src/coreclr/jit/llvm.cpp
yowl: Added, with a couple of extra checks that I seemed to need.

src/coreclr/jit/llvm.cpp
yowl: Right, started as suggested.

src/coreclr/jit/llvm.cpp
yowl: Ok.  For LCL_FLD I'll need to get the LLVM Type but I guess we can pick that up in the next PR.

src/coreclr/jit/llvm.cpp
```diff
@@ -2387,6 +2427,55 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
...
+    GenTree* dataOp = storeLclNode->gtGetOp1();
+    CORINFO_CLASS_HANDLE dataHandle = _compiler->gtGetStructHandleIfPresent(dataOp);
+    if (dataOp->OperIs(GT_IND))
+    {
+        // Special case: "gtGetStructHandleIfPresent" sometimes guesses the handle from
+        // field sequences, but we will always need to transform TYP_STRUCT INDs into OBJs.
+        dataHandle = NO_CLASS_HANDLE;
+    }
+
+    LclVarDsc* addrVarDsc = _compiler->lvaGetDesc(storeLclNode->GetLclNum());
```
SingleAccretion: This should be moved under the `storeLclNode->TypeIs(TYP_STRUCT)` clause, since it's not relevant for non-struct stores:

```c
if (storeLclNode->OperIsCopyBlkOp())
{
    dataHandle = { ... };
    if (addrVarDsc->GetStructHnd() != dataHandle) { ... }
}
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2435,11 +2524,17 @@ void Llvm::lowerToShadowStack()
 
                 CurrentRange().InsertBefore(node, retAddressLocal, storeNode);
             }
+
             if (node->OperIsLocalAddr())
             {
                 // Indicates that this local is to live on the LLVM frame, and will not participate in SSA.
                 _compiler->lvaGetDesc(node->AsLclVarCommon())->lvHasLocalAddr = 1;
             }
+            // check for implicit struct cast
```
```suggestion
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2387,6 +2427,55 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
...
+            GenTreeLclVarCommon* dataLcl = dataOp->AsLclVarCommon();
+            if (storeLclNode->TypeGet() == TYP_STRUCT && dataLcl->TypeGet() == TYP_STRUCT)
+            {
+                LclVarDsc* dataVarDsc = _compiler->lvaGetDesc(dataLcl->GetLclNum());
+
+                if (tryGetStructClassHandle(addrVarDsc) != tryGetStructClassHandle(dataVarDsc))
+                {
+                    dataVarDsc->lvHasLocalAddr = 1;
+
+                    convertLclStructToLoad(dataLcl, addrVarDsc->GetLayout());
```
SingleAccretion: I think `convertLclStructToLoad` can be inlined into this method (can be outlined again if needed in the future).

src/coreclr/jit/llvm.cpp
```diff
@@ -2387,6 +2427,55 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
...
+    {
+        if (dataOp->OperIsIndir())
+        {
+            dataOp->SetOper(GT_OBJ);
+            dataOp->AsObj()->SetLayout(addrVarDsc->GetLayout());
+        }
+        else if (dataOp->OperIs(GT_LCL_VAR)) // can get icon 0 here
+        {
+            GenTreeLclVarCommon* dataLcl = dataOp->AsLclVarCommon();
+            if (storeLclNode->TypeGet() == TYP_STRUCT && dataLcl->TypeGet() == TYP_STRUCT)
```
SingleAccretion: Redundant with the `storeLclNode->TypeIs(TYP_STRUCT)` above.

src/coreclr/jit/llvm.cpp
```diff
@@ -2387,6 +2427,55 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
...
+            {
+                LclVarDsc* dataVarDsc = _compiler->lvaGetDesc(dataLcl->GetLclNum());
+
+                if (tryGetStructClassHandle(addrVarDsc) != tryGetStructClassHandle(dataVarDsc))
```
SingleAccretion: Redundant with the handle check above.

src/coreclr/jit/llvm.cpp
yowl: Are we removing the `if (dataOp->OperIs(GT_IND))` check ?

yowl: I tried changing `storeLclNode->TypeIs(TYP_STRUCT)` to `storeLclNode->OperIsCopyBlkOp()` but I'm getting some assert.  Let me see what that is.

SingleAccretion: Ah, I am sorry for that, `OperIsCopyBlkOp` will indeed not work here, it doesn't recognize `STORE_LCL_VAR`. `TypeIs(TYP_STRUCT)` should be left as it was.

(There is an interesting inconsistency: `ASG(LCL struct, ...)` is `OperIsCopyBlkOp` but the equivalent `STORE_LCL_VAR` is)

yowl:
I think we might be there.  Thanks

src/coreclr/jit/llvm.cpp
```diff
@@ -2387,6 +2427,50 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
...
+        {
+            if (dataOp->OperIsIndir())
+            {
+                dataOp->SetOper(GT_OBJ);
+                dataOp->AsObj()->SetLayout(addrVarDsc->GetLayout());
+            }
+            else if (dataOp->OperIs(GT_LCL_VAR)) // can get icon 0 here
+            {
+                GenTreeLclVarCommon* dataLcl = dataOp->AsLclVarCommon();
+                if (dataLcl->TypeGet() == TYP_STRUCT)
```
SingleAccretion: I think the check should be removed.

We can have IR like `STORE_LCL_VAR<struct>(LCL_VAR long)`, and for that case we'd want this transformation too.

```suggestion
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2387,6 +2427,50 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
...
+
+                    dataVarDsc->lvHasLocalAddr = 1;
+
+                    GenTree* dataAddrNode = _compiler->gtNewLclVarAddrNode(dataLcl->GetLclNum());
+
+                    dataLcl->ChangeOper(GT_OBJ);
+                    dataLcl->AsObj()->SetAddr(dataAddrNode);
+                    dataLcl->AsObj()->SetLayout(addrVarDsc->GetLayout());
+
+                    _currentRange->InsertBefore(dataLcl, dataAddrNode);
```
```suggestion
                    CurrentRange().InsertBefore(dataLcl, dataAddrNode);
```

src/coreclr/jit/llvm.cpp
yowl: Yes, removed.

SingleAccretion (approved):

yowl:
Died after 200 mins, and now is having some HTTP problems.

yowl:
Increased timeout

PR #1871, opened 2022-03-14
[NativeAOT-LLVM] Add support for GT_NULLCHECK

yowl:
This PR adds support for `GT_NULLCHECK` using the the compiler helper  `ThrowNullReferenceException`.  

I thought there was an outside chance this would fail in a more interesting way than #1866, but it doesn't seem to .

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -27,6 +27,9 @@ using llvm::Module;
 static Module*          _module    = nullptr;
 static llvm::DIBuilder* _diBuilder = nullptr;
 static LLVMContext _llvmContext;
+
+static Function* _nullCheckFunction = nullptr;
+
 static void* _thisPtr; // TODO: workaround for not changing the JIT/EE interface.  As this is static, it will probably fail if multithreaded compilation is attempted
 static const char* (*_getMangledMethodName)(void*, CORINFO_METHOD_STRUCT_*);
 static const char* (*_getMangledSymbolName)(void*, void*);
@@ -42,6 +45,7 @@ static const uint32_t (*_padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned);
 static const CorInfoTypeWithMod (*_getArgTypeIncludingParameterized)(void*, CORINFO_SIG_INFO*, CORINFO_ARG_LIST_HANDLE, CORINFO_CLASS_HANDLE*);
 static const CorInfoTypeWithMod (*_getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*);
 static const TypeDescriptor (*_getTypeDescriptor)(void*, CORINFO_CLASS_HANDLE);
+static CORINFO_METHOD_HANDLE (*_getCompilerHelpersMethodHandle)(void*, const char*, unsigned, const char*, unsigned);
 
 static char*                              _outputFileName;
 static Function*                          _doNothingFunction;
@@ -67,7 +71,8 @@ extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
                                                 const uint32_t(*padOffset)(void*, CORINFO_CLASS_STRUCT_*, unsigned),
                                                 const CorInfoTypeWithMod(*getArgTypeIncludingParameterized)(void*, CORINFO_SIG_INFO*, CORINFO_ARG_LIST_HANDLE, CORINFO_CLASS_HANDLE*),
                                                 const CorInfoTypeWithMod(*getParameterType)(void*, CORINFO_CLASS_HANDLE, CORINFO_CLASS_HANDLE*),
-                                                const TypeDescriptor(*getTypeDescriptor)(void*, CORINFO_CLASS_HANDLE))
+                                                const TypeDescriptor(*getTypeDescriptor)(void*, CORINFO_CLASS_HANDLE),
+                                                CORINFO_METHOD_HANDLE (*getCompilerHelpersMethodHandle)(void*, const char*, unsigned, const char*, unsigned))
 {
     _thisPtr = thisPtr;
     _getMangledMethodName         = getMangledMethodNamePtr;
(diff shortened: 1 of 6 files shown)


yowl:
@jkotas @singleaccretion  I know I've not done much here for a while, but this PR could be reviewed if there is still interest.

src/coreclr/jit/llvm.cpp
```diff
@@ -1639,6 +1645,65 @@ unsigned Llvm::buildMemCpy(Value* baseAddress, unsigned startOffset, unsigned en
     return size;
 }
 
+void Llvm::buildThrowException(llvm::IRBuilder<>& builder, const char* helperClass, const char * helperMethodName, Value* shadowStack)
+{
+    CORINFO_METHOD_HANDLE methodHandle = _getCompilerHelpersMethodHandle(_thisPtr, helperClass, strlen(helperClass),
```
jkotas: The JIT/EE interface typically sends the strings over as zero terminated strings (without explicit length) and then constructs the managed string using `Marshal.PtrToStringUTF8`. Fewer arguments to pass through.

src/coreclr/jit/llvm.cpp
```diff
@@ -1639,6 +1645,65 @@ unsigned Llvm::buildMemCpy(Value* baseAddress, unsigned startOffset, unsigned en
...
+                                                                         helperMethodName, strlen(helperMethodName));
```
jkotas:
```suggestion
    CORINFO_METHOD_HANDLE methodHandle = _getCompilerHelpersMethodHandle(_thisPtr, helperClass, helperMethodName);
```

src/coreclr/jit/llvm.h
```diff
@@ -223,6 +224,10 @@ class Llvm
     StructDesc* getStructDesc(CORINFO_CLASS_HANDLE structHandle);
     unsigned buildMemCpy(Value* baseAddress, unsigned startOffset, unsigned endOffset, Value* srcAddress);
     void buildLocalField(GenTreeLclFld* lclFld);
+    void buildNullCheck(GenTreeUnOp* nullCheckNode);
+    void buildThrowException(llvm::IRBuilder<>& builder, const char*        helperClass, const char*        helperMethodName, Value* shadowStack);
```
```suggestion
    void buildThrowException(llvm::IRBuilder<>& builder, const char* helperClass, const char* helperMethodName, Value* shadowStack);
```

src/coreclr/jit/llvm.cpp
```diff
@@ -27,6 +27,9 @@ using llvm::Module;
 static Module*          _module    = nullptr;
 static llvm::DIBuilder* _diBuilder = nullptr;
 static LLVMContext _llvmContext;
+
+static Function* _nullCheckFunction = nullptr;
```
SingleAccretion, line 33: Unprotected (with locks) static state like this is concerning. It is fine as-is currently, but we should make it more apparent what's missing.

Ideally, there'd be a `StaticCompiler`-like class (which should also encapsulate other `static` things above) that would be responsible for managing state shared across compilations. This would also make building SPMI-like tooling for the LLVM Jit easier.

src/coreclr/jit/llvm.cpp
yowl: Thanks, using `Marshal.PtrToStringAnsi((IntPtr)className)` to go to a dotnet string

src/coreclr/jit/llvm.cpp
yowl: Thanks, I copied
```
private int doAssert(byte* szFile, int iLine, byte* szExpr)
        {
            Log.WriteLine(Marshal.PtrToStringAnsi((IntPtr)szFile) + ":" + iLine);
            Log.WriteLine(Marshal.PtrToStringAnsi((IntPtr)szExpr));

            return 1;
        }
```

src/coreclr/jit/llvm.cpp
yowl: Thanks, added a comment.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -351,6 +362,14 @@ public static TypeDescriptor getTypeDescriptor(IntPtr thisHandle, CORINFO_CLASS_
             return typeDescriptor;
         }
 
+        [UnmanagedCallersOnly]
+        public static CORINFO_METHOD_STRUCT_* getCompilerHelpersMethodHandle(IntPtr thisHandle, byte* className, byte* methodName)
+        {
+            var _this = GetThis(thisHandle);
+
+            return _this.ObjectToHandle(_this._compilation.GetCompilerHelpersMethodDesc(Marshal.PtrToStringAnsi((IntPtr)className), Marshal.PtrToStringAnsi((IntPtr)methodName)));
```
jkotas:
```suggestion
            return _this.ObjectToHandle(_this._compilation.GetCompilerHelpersMethodDesc(Marshal.PtrToStringUTF8((IntPtr)className), Marshal.PtrToStringUTF8((IntPtr)methodName)));
```
PtrToStringAnsi is legacy

SingleAccretion (approved):

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
yowl: THanks, you did say.  Sorry.

yowl:
Seems like the OSX build was ok after the merge commit a month ago but now is consistently failing with 
```
Cloning into 'llvm-project'...
  fatal: specify directories rather than patterns (no leading slash)
  Already on 'release/12.x'
  Your branch is up to date with 'origin/release/12.x'.
  /Users/runner/work/1/s/src/coreclr/tools/aot/ObjWriter/build.sh: line 48: cd: llvm: No such file or directory

```
Does that ring any bells ?

jkotas:
I guess that the new macOS git client is complaining about this line https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/src/coreclr/tools/aot/ObjWriter/build.sh#L46 ?

src/coreclr/tools/aot/ObjWriter/build.sh
```diff
@@ -43,7 +43,7 @@ if [ ! -d llvm-project ]; then
     cd llvm-project || exit 1
     # Purposefuly ignoring exit codes of sparse-checkout so that this works with git lower than 2.26
     git sparse-checkout init
-    git sparse-checkout set /llvm/\* !/llvm/test/\*
+    git sparse-checkout set llvm/\* !llvm/test/\*
```
jkotas:
```suggestion
    git sparse-checkout set --no-cone /llvm/\* !/llvm/test/\*
```

PR #1873, opened 2022-03-16
[NativeAOT-LLVM] Call methods with byref parameters, e.g. this

yowl:
This PR enables `TYP_BYREF` args, basically by just removing the blocks on them and disabling `FEATURE_FIXED_OUT_ARGS`

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1589,7 +1589,7 @@ void Llvm::buildStoreInd(GenTreeStoreInd* storeIndOp)
 {
     Value* address = getGenTreeValue(storeIndOp->Addr());
     Value* toStore = getGenTreeValue(storeIndOp->Data());
-    if (toStore->getType()->isPointerTy() && (storeIndOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0)
+    if (toStore->getType()->isPointerTy() && (storeIndOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0 && address->getType()->isPointerTy())
     {
         // RhpAssignRef will never reverse PInvoke, so do not need to store the shadow stack here
         _builder.CreateCall(getOrCreateRhpAssignRef(), ArrayRef<Value*>{address, castIfNecessary(toStore, Type::getInt8PtrTy(_llvmContext))});
@@ -2317,7 +2317,7 @@ void Llvm::failUnsupportedCalls(GenTreeCall* callNode)
 
             fgArgTabEntry* curArgTabEntry = _compiler->gtArgEntryByNode(callNode, operand);
             regNumber      argReg         = curArgTabEntry->GetRegNum();
-            if (argReg == REG_STK || curArgTabEntry->argType == TYP_BYREF) // TODO-LLVM: out and ref args
+            if (argReg == REG_STK) // TODO-LLVM: out args
             {
                 failFunctionCompilation();
             }
@@ -2641,12 +2641,6 @@ void Llvm::createAllocasForLocalsWithAddrOp()
 
     for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
     {
-        // TODO-LLVM: Consider turning off FEATURE_FIXED_OUT_ARGS.
-        if(_compiler->lvaOutgoingArgSpaceVar == lclNum)
-        {
-            continue;
-        }
-
         LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
 
         if (canStoreLocalOnLlvmStack(varDsc) && isLlvmFrameLocal(varDsc))
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -617,7 +617,27 @@ Function* getOrCreateRhpAssignRef()
     Function* llvmFunc = _module->getFunction("RhpAssignRef");
     if (llvmFunc == nullptr)
     {
-        llvmFunc = Function::Create(FunctionType::get(Type::getVoidTy(_llvmContext), ArrayRef<Type*>{Type::getInt8PtrTy(_llvmContext), Type::getInt8PtrTy(_llvmContext)}, false), Function::ExternalLinkage, 0U, "RhpAssignRef", _module); // TODO: ExternalLinkage forced as linked from old module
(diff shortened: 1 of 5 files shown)


yowl:
closing to rerun

src/coreclr/jit/llvm.cpp
```diff
@@ -1589,7 +1589,7 @@ void Llvm::buildStoreInd(GenTreeStoreInd* storeIndOp)
 {
     Value* address = getGenTreeValue(storeIndOp->Addr());
     Value* toStore = getGenTreeValue(storeIndOp->Data());
-    if (toStore->getType()->isPointerTy() && (storeIndOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0)
+    if (toStore->getType()->isPointerTy() && (storeIndOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0 && address->getType()->isPointerTy())
```
SingleAccretion: I don't think this is correct. The address can be an integer here, that points to the heap, to a pinned object.

(Side note: we should `consumeValue` it as a pointer to the type we're going store instead of manually `castIfNecessary`ing)

The logic required to know whether we need the barrier lives here: https://github.com/dotnet/runtime/blob/10ac54d21b60263a94596cba441bdd68f370d480/src/coreclr/jit/gcinfo.cpp#L224-L272

(`WBF_BarrierUnchecked` is `RhpAssignRef`, `WBF_BarrierChecked` and `WBF_Unknown` are `RhpCheckedAssignRef`)

It would be nice to reuse it here, if it is not possible, the simplified (inefficient) version is that we need a write barrier for `storeIndOp->TypeIs(TYP_REF)`, and it should be checked.

yowl: I can change this to `consumeValue`,  but I'll still need to cast to `i8*` as that will be the types for the args in `RhpAssignRef` and `RhpCheckedAssignRef`   LLVM is going to opaque pointers which might remove this (i've not read up on that).

src/coreclr/jit/llvm.cpp
```diff
@@ -1587,16 +1607,28 @@ void Llvm::buildReturn(GenTree* node)
...
-    {
-        // RhpAssignRef will never reverse PInvoke, so do not need to store the shadow stack here
-        _builder.CreateCall(getOrCreateRhpAssignRef(), ArrayRef<Value*>{address, castIfNecessary(toStore, Type::getInt8PtrTy(_llvmContext))});
-    }
-    else
+    GenTree* data = storeIndOp->Data();
+    Value* toStore = getGenTreeValue(data);
+    Value* address = consumeValue(storeIndOp->Addr(), toStore->getType()->getPointerTo());
+
+    GCInfo gcInfo(_compiler);
```
SingleAccretion: `GCInfo` is a somewhat large class, it is meant to be constructed once and then re-used.

src/coreclr/jit/llvm.cpp
```diff
@@ -1587,16 +1607,28 @@ void Llvm::buildReturn(GenTree* node)
...
+                                                 castIfNecessary(toStore, Type::getInt8PtrTy(_llvmContext))});
+            break;
+        case GCInfo::WriteBarrierForm::WBF_BarrierChecked:
+        case GCInfo::WriteBarrierForm::WBF_BarrierUnknown:
+            _builder.CreateCall(getOrCreateRhpCheckedAssignRef(),
+                                ArrayRef<Value*>{castIfNecessary(address, Type::getInt8PtrTy(_llvmContext)),
+                                                 castIfNecessary(toStore, Type::getInt8PtrTy(_llvmContext))});
+            break;
+        default:
+            castingStore(toStore, address, storeIndOp->gtType);
```
SingleAccretion: It would be better to have an explicit `WBF_NoBarrier` case for this and then `unreached` for the `default`.

Also, we should `consumeValue(data, getLlvmTypeForVarType(storeIndOp->TypeGet()))` for `toStore` and delete `castingStore` (you can also delete `storeOnShadowStack` that is its only other user, and which itself is unused).

src/coreclr/jit/llvm.cpp
```diff
@@ -1587,16 +1607,28 @@ void Llvm::buildReturn(GenTree* node)
...
-    Value* toStore = getGenTreeValue(storeIndOp->Data());
-    if (toStore->getType()->isPointerTy() && (storeIndOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0)
-    {
-        // RhpAssignRef will never reverse PInvoke, so do not need to store the shadow stack here
-        _builder.CreateCall(getOrCreateRhpAssignRef(), ArrayRef<Value*>{address, castIfNecessary(toStore, Type::getInt8PtrTy(_llvmContext))});
-    }
-    else
+    GenTree* data = storeIndOp->Data();
+    Value* toStore = getGenTreeValue(data);
+    Value* address = consumeValue(storeIndOp->Addr(), toStore->getType()->getPointerTo());
```
```suggestion
  Value* address = consumeValue(storeIndOp->Addr(), getLlvmTypeForVarType(storeIndOp->TypeGet())->getPointerTo());
```
(Coupled with the comment about `data`)

src/coreclr/jit/llvm.cpp
yowl: Thanks, I created a static with a member lazy getter

SingleAccretion: Hmm, it cannot be static, since it would outlive the compiler then, it should be a member of the `Llvm` class.

yowl: Ok, I'll change that

src/coreclr/jit/llvm.cpp
```diff
@@ -98,6 +97,15 @@ extern "C" DLLEXPORT void registerLlvmCallbacks(void*       thisPtr,
     }
 }
 
+GCInfo* Llvm::getGCInfo()
+{
+    if (_gcInfo == nullptr)
+    {
+        _gcInfo = new GCInfo(_compiler);
```
SingleAccretion: This should use the compiler's allocator to avoid leaking: `new (_compiler->getAllocator(CMK_GC)) GCInfo(_compiler);`

SingleAccretion (approved):
LGTM modulo one comment.

yowl:
@SingleAccretion Thanks very much for the review

@jkotas I think this can be merged when you get the chance, thanks!

jkotas (approved):
Thanks

PR #1874, opened 2022-03-16
[NativeAOT-LLVM] To support helper functions that return structs....

yowl:
... change `getLlvmTypeForVarType` to accept an optional `CORINFO_CLASS_HANDLE`


yowl:
not needed

PR #1878, opened 2022-03-17
[NativeAOT-LLVM] Support for GT_LCL_FLD

yowl:
This PR adds support for GT_LCL_FLD ~~and disables FEATURE_FIXED_OUT_ARGS~~

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -2625,6 +2625,11 @@ void Llvm::createAllocasForLocalsWithAddrOp()
 
     for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
     {
+        if(_compiler->lvaOutgoingArgSpaceVar == lclNum)
+        {
+            continue;
+        }
+
         LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
 
         if (canStoreLocalOnLlvmStack(varDsc) && isLlvmFrameLocal(varDsc))
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -436,9 +436,8 @@ CorInfoType Llvm::toCorInfoType(var_types varType)
             return CorInfoType::CORINFO_TYPE_BYTE;
         case TYP_UBYTE:
             return CorInfoType::CORINFO_TYPE_UBYTE;
-        case TYP_LCLBLK: // TODO: outgoing args space - need to get an example compiling, e.g. https://github.com/dotnet/runtimelab/blob/40f9ff64ae80596bcddcec16a7e1a8f57a0b2cff/src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs#L3492 to see what's
-            // going on.  CORINFO_TYPE_VALUECLASS is a better mapping but if that is mapped as of now, then canStoreTypeOnLlvmStack will fail compilation for most methods.
-            failFunctionCompilation();
+        case TYP_LCLBLK:
+            return CorInfoType::CORINFO_TYPE_VALUECLASS;
         case TYP_DOUBLE:
             return CorInfoType::CORINFO_TYPE_DOUBLE;
         case TYP_FLOAT:
@@ -1685,6 +1684,43 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
     return llvmRef;
 }
 
+void Llvm::buildLocalField(GenTreeLclFld* lclFld)
+{
+    unsigned int lclNum = lclFld->GetLclNum();
+
+    Value*     structAddrValue;
+    LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
+
+    if (canStoreLocalOnLlvmStack(varDsc))
+    {
+        assert(m_allocas[lclNum] != nullptr);
+
+        structAddrValue = m_allocas[lclNum];
+    }
+    else
+    {
+        structAddrValue = getLocalVarAddress(lclFld);
+    }
+
+    // TODO-LLVM: if this is an only value type field, or at offset 0, we can optimize
+    Value* structAddrInt8Ptr = castIfNecessary(structAddrValue, Type::getInt8PtrTy(_llvmContext));
+    Value* fieldAddressValue = _builder.CreateGEP(structAddrInt8Ptr, _builder.getInt16(lclFld->GetLclOffs()));
+
+    Type* fieldType;
+    if (lclFld->TypeGet() == TYP_STRUCT)
+    {
+        fieldType = getLlvmTypeForStruct(_compiler->lvaGetDesc(lclFld)->GetLayout());
+    }
+    else
+    {
+        fieldType = getLlvmTypeForVarType(lclFld->TypeGet());
+    }
+
+    Value* fieldAddressTypedValue = _builder.CreateBitCast(fieldAddressValue, fieldType->getPointerTo());
+
+    mapGenTreeToValue(lclFld, _builder.CreateLoad(fieldAddressTypedValue));
+}
+
 void Llvm::buildLocalVarAddr(GenTreeLclVarCommon* lclAddr)
 {
     unsigned int lclNum = lclAddr->GetLclNum();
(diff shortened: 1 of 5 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -1685,6 +1684,43 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
...
+    }
+
+    // TODO-LLVM: if this is an only value type field, or at offset 0, we can optimize
+    Value* structAddrInt8Ptr = castIfNecessary(structAddrValue, Type::getInt8PtrTy(_llvmContext));
+    Value* fieldAddressValue = _builder.CreateGEP(structAddrInt8Ptr, _builder.getInt16(lclFld->GetLclOffs()));
+
+    Type* fieldType;
+    if (lclFld->TypeGet() == TYP_STRUCT)
+    {
+        fieldType = getLlvmTypeForStruct(_compiler->lvaGetDesc(lclFld)->GetLayout());
```
yowl: I think you mentioned that this was not goting to work for field, but I wasn't sure what to replace it with so have left it here for review.

src/coreclr/jit/target.h
```diff
@@ -276,7 +276,9 @@ typedef unsigned char   regNumberSmall;
 #define FEATURE_PARTIAL_SIMD_CALLEE_SAVE 1 // Whether SIMD registers are partially saved at calls
 #endif // !UNIX_AMD64_ABI
 #endif
+#ifndef TARGET_WASM
```
SingleAccretion: It is unfortunate that this will conflict on upstream merge, `target.h` was split into individual files there.

src/coreclr/jit/reglist.h
```diff
@@ -8,7 +8,7 @@
 #include "tinyarray.h"
 
 // The "regList" type is a small set of registerse
-#ifdef TARGET_X86
+#if defined(TARGET_X86) || defined(TARGET_WASM)
```
SingleAccretion: This will also conflict on merge (the reglist stuff was deleted as dead code)...

Maybe we should consider turning off `FEATURE_FIXED_OUT_ARGS` after merging upstream.

src/coreclr/jit/llvm.cpp
```diff
@@ -2527,6 +2566,7 @@ void Llvm::lowerToShadowStack()
                 lowerStoreLcl(node->AsLclVarCommon());
             }
```
```suggestion
            }
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1685,6 +1684,43 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
...
+
+    if (canStoreLocalOnLlvmStack(varDsc))
+    {
+        assert(m_allocas[lclNum] != nullptr);
+
+        structAddrValue = m_allocas[lclNum];
+    }
+    else
+    {
+        structAddrValue = getLocalVarAddress(lclFld);
```
SingleAccretion: This should be handled in `ConvertShadowStackLocalNode`.

Once that's moved, you can delete `getLocalVarAddress` and `getLocalOffsetAtIndex`, they're unused.

src/coreclr/jit/llvm.cpp
SingleAccretion: Yeah, the problem is that `LCL_FLD` does not support `TYP_STRUCT` (no way to get the layout out of the node). It's a very large IR issue I am working on right now.

So right now `buildLocalField` should just `assert(!lclFld->TypeIs(TYP_STRUCT);`.

src/coreclr/jit/llvm.cpp
```diff
@@ -1685,6 +1684,43 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
...
+    if (lclFld->TypeGet() == TYP_STRUCT)
+    {
+        fieldType = getLlvmTypeForStruct(_compiler->lvaGetDesc(lclFld)->GetLayout());
+    }
+    else
+    {
+        fieldType = getLlvmTypeForVarType(lclFld->TypeGet());
+    }
+
+    Value* fieldAddressTypedValue = _builder.CreateBitCast(fieldAddressValue, fieldType->getPointerTo());
```
SingleAccretion: Should this be `castIfNecessary` to avoid nop bitcasts?

src/coreclr/jit/reglist.h
yowl: ok, I reverted that, I added a bit of handling for `lvaOutgoingArgSpaceVar` which may have been the reason I did it here.

yowl:
Infra problems

yowl:
Annoying as it actually did pass:
image

src/coreclr/jit/llvm.cpp
```diff
@@ -1685,6 +1684,29 @@ Value* Llvm::localVar(GenTreeLclVar* lclVar)
...
+    Value*     structAddrValue;
+    LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
+
+    assert(canStoreLocalOnLlvmStack(varDsc));
+    assert(m_allocas[lclNum] != nullptr);
+
+    structAddrValue = m_allocas[lclNum];
+
+    // TODO-LLVM: if this is an only value type field, or at offset 0, we can optimize
+    Value* structAddrInt8Ptr = castIfNecessary(structAddrValue, Type::getInt8PtrTy(_llvmContext));
```
```suggestion
    unsigned   lclNum = lclFld->GetLclNum();
    LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
    assert(isLlvmFrameLocal(varDsc));

    // TODO-LLVM: if this is an only value type field, or at offset 0, we can optimize.
    Value* structAddrValue = m_allocas[lclNum];
    Value* structAddrInt8Ptr = castIfNecessary(structAddrValue, Type::getInt8PtrTy(_llvmContext));
```

src/coreclr/jit/llvm.cpp
```diff
@@ -2123,6 +2105,15 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
...
                 break;
+            case GT_LCL_FLD:
+                // TODO-LLVM: TYP_STRUCT
+                if (lclVar->TypeIs(TYP_STRUCT))
+                {
+                    failFunctionCompilation();
+                }
+
+                indirOper = GT_IND;
+                break;
```
```suggestion
                assert(!lclVar->TypeIs(TYP_STRUCT));
                indirOper = GT_IND;
                break;
```
It's not that we don't have an implementation for it in the LLVM backend, it's that it cannot exist in the compiler at all (with some exceptions that should not be relevant here).

src/coreclr/jit/llvm.cpp
```diff
@@ -2475,7 +2466,7 @@ void Llvm::lowerToShadowStack()
         _currentRange = &LIR::AsRange(_currentBlock);
         for (GenTree* node : CurrentRange())
         {
-            if (node->OperIs(GT_STORE_LCL_VAR, GT_LCL_VAR, GT_LCL_VAR_ADDR, GT_LCL_FLD_ADDR))
+            if (node->OperIs(GT_STORE_LCL_VAR, GT_LCL_VAR, GT_LCL_VAR_ADDR, GT_LCL_FLD_ADDR, GT_LCL_FLD))
```
```suggestion
            if (node->OperIs(GT_STORE_LCL_VAR, GT_LCL_VAR, GT_LCL_FLD, GT_LCL_VAR_ADDR, GT_LCL_FLD_ADDR))
```
(Symmetry...)

I guess the only thing left to implement here is `STORE_LCL_FLD` and then we can convert this to `OperIsLocal() || OperIsLocalAddr()`.

src/coreclr/jit/llvm.cpp
```diff
@@ -2585,10 +2576,14 @@ void Llvm::createAllocasForLocalsWithAddrOp()
 
     for (unsigned lclNum = 0; lclNum < _compiler->lvaCount; lclNum++)
     {
+        if(_compiler->lvaOutgoingArgSpaceVar == lclNum)
```
```suggestion
    {
        // TODO-LLVM: Consider turning off FEATURE_FIXED_OUT_ARGS.
        if(_compiler->lvaOutgoingArgSpaceVar == lclNum)
```
(I think the TODO is still relevant -- just the matter is not pressing and merge conflicts are best avoided)

src/coreclr/jit/llvm.cpp
yowl: ok, thanks

SingleAccretion (approved):

jkotas:
Thank you!

PR #1888, opened 2022-04-18
NativeAOT-LLVM: Dont overwrite shadow stack in filters and finally blocks

yowl:
In CoreRT I made some changes to get the Wasm stable long enough to start an Uno Platform HelloWorld Xaml page.  I never created a PR for those as I never did the work to establish which changes were actually necessary.  I got a repro from a user and was able to replicate some memory corruption issues and then bring in some of the changes from the CoreRT branch I had.  This PR includes those changes that I have found make the system stable .  When calculating the shadow stack from filters and finally blocks, there was an assumption of compilation order matching execution order so when that was not the case it was possible for a finally block to overwrite local on the shadow stack.  This PR fixes that by calculating the total shadow stack space required for the whole method including finally/filter blocks and backfilling the shadow stack use once the method has been fully compiled.

### DIFF
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -35587,6 +35587,7 @@ uint8_t* gc_heap::find_first_object (uint8_t* start, uint8_t* first_object)
             prev_brick = (brick_entry + prev_brick);
 
         }
+		      /*
 	if(prev_brick < min_brick){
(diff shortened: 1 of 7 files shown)


yowl:
This is not going to get committed so will close.  The RyuJIT backend is getting close to 100% and will remove the need for fixing the IL backend

PR #1899, opened 2022-06-15
Empty Assembly Functionality

MosheWolberg:
This is the first prototype of writing and reading of a simple blank assembly with a single module with a new version of Assembly Builder and ModuleBuilder. The other Builders will be implemented as the project expands scope. There was theoretical support for multiple modules and that's why I created a dictionary for storing modules. But I discovered that MetadataBuilder wouldn't let me add multiple modules so that was rendered moot and anyways @AaronRobinsonMSFT told me not practical or relevant.
See the comments for design points and some questions.
I assume I'll move onto basic types (Interfaces) next.
There is also the second niche constructor for AssemblyBuilder but I'm not sure if that's as important to work on now.


src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
...
+using System.Reflection.PortableExecutable;
+using System.Runtime.CompilerServices;
 
 namespace System.Reflection.Emit.Experimental
 {
+
     public class AssemblyBuilder: System.Reflection.Assembly
     {
-        internal AssemblyBuilder() { }
+        private static readonly Guid _s_guid = new Guid("87D4DBE1-1143-4FAD-AAB3-1001F92068E6");//Some random ID, Need to look into how these should be generated.
```
jkotas:
```suggestion
        private static readonly Guid s_guid = new Guid("87D4DBE1-1143-4FAD-AAB3-1001F92068E6");//Some random ID, Need to look into how these should be generated.
```
The prefix for statics is just `s_`

jkotas: `Guid.NewGuid()` is a method to call to generate random Guid

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
...
+        private static readonly BlobContentId _s_contentId = new BlobContentId(_s_guid, 0x04030201);
+        private BlobBuilder _emptyBlob = new BlobBuilder();
+        public override string? FullName { get; }
+        private static MetadataBuilder _metadata = new MetadataBuilder();
+        private static IDictionary<string,ModuleBuilder> moduleStorage = new Dictionary<string, ModuleBuilder>();
```
jkotas:
```suggestion
        private static MetadataBuilder _metadata = new MetadataBuilder();
        private static IDictionary<string,ModuleBuilder> moduleStorage = new Dictionary<string, ModuleBuilder>();
```
These two should not be static. Also, `moduleStorage` should have `_` prefix.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
...
+                throw new ArgumentNullException();
+            }
+            if (FullName == null)
+            {
+                throw new ArgumentNullException();
+            }
+            //Add assembly metadata
+            _metadata.AddAssembly(//Metadata is added for the new assembly - Current design - metdata generated only when Save method is called.
+               _metadata.GetOrAddString(value: FullName),//FullName, CultureName?
+               version: new Version(1, 0, 0, 0),
```
jkotas: version, culture and publicKey should come from the AssemblyName that was used to create the assembly.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
...
+               hashAlgorithm: AssemblyHashAlgorithm.None);
+            //Add each module's medata
+            foreach (KeyValuePair<string, ModuleBuilder> entry in moduleStorage)
+            {
+                entry.Value.AppendMetadata(_metadata);
+            }
+            using var peStream = new FileStream(assemblyFileName, FileMode.OpenOrCreate, FileAccess.ReadWrite);
+            var ilBuilder = new BlobBuilder();
+            WritePEImage(peStream, _metadata, ilBuilder);
+            peStream.Dispose();//close or dispose?
```
jkotas: The `using ...` above will call `Dispose` for you at the end of the scope. You do not need to call it manually.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
...
-            throw new NotImplementedException(); 
+        public static System.Reflection.Emit.Experimental.AssemblyBuilder DefineDynamicAssembly(System.Reflection.AssemblyName name, System.Reflection.Emit.AssemblyBuilderAccess access)
+        {
+            if (name == null || name.Name == null)
+            {
+                throw new ArgumentNullException();
+            }
+            //AssemblyBuilderAccess affects runtime managment only and is not relevant for saving to disk.
+            AssemblyBuilder currentAssembly = new AssemblyBuilder(name.Name);
+            //We need to create module becaue even a blank assembly has one module.
```
jkotas:
```suggestion
            //We need to create module because even a blank assembly has one module.
```
Typo

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,136 @@
...
+    public class ModuleBuilder : System.Reflection.Module
+    {
+        private readonly Guid _s_guid;
+        public override System.Reflection.Assembly Assembly { get; }
+        public override string ScopeName { get; }
+
+        internal ModuleBuilder(string name, Assembly assembly)
+        {
+         //Example random GUID - I still need to understand when, how and why to generate them.
+        _s_guid= Guid.NewGuid();
```
jkotas: It should not be necessary to create the random Guid upfront and store it. You can create just when you needed, like:

`metadataBuilder.GetOrAddGuid(Guid.NewGuid())`

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
...
-        { 
-            throw new NotImplementedException(); 
+        public static System.Reflection.Emit.Experimental.AssemblyBuilder DefineDynamicAssembly(System.Reflection.AssemblyName name, System.Reflection.Emit.AssemblyBuilderAccess access)
+        {
+            if (name == null || name.Name == null)
+            {
+                throw new ArgumentNullException();
+            }
+            //AssemblyBuilderAccess affects runtime managment only and is not relevant for saving to disk.
+            AssemblyBuilder currentAssembly = new AssemblyBuilder(name.Name);
```
AaronRobinsonMSFT: I would thread the `AssemblyName` into the `AssemblyBuilder`. As @jkotas has mentioned you need those details during the save routine.

src/Experiment/src/MyClass.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using System;
```
AaronRobinsonMSFT: I would remove this file unless you have a purpose for it.

src/Experiment/tests/MyClassTests.cs
```diff
@@ -1,14 +1,68 @@
 using System;
```
AaronRobinsonMSFT: License header is needed here. Also, let's rename the file to what it is testing.

src/Experiment/tests/MyClassTests.cs
```diff
@@ -1,14 +1,68 @@
...
+                {
+                    throw new IOException();
+                }
+
+            }
+            finally
+            {
+                if (File.Exists(fileName))
+                {
+                    File.Delete(fileName);
```
AaronRobinsonMSFT: I think `File.Delete` handles the case where `fileName` doesn't exist.

src/Experiment/tests/MyClassTests.cs
```diff
@@ -1,14 +1,68 @@
...
+                using var fs = new FileStream(fileName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+                using var peReader = new PEReader(fs);
+                MetadataReader mr = peReader.GetMetadataReader();
+                var asssemblyData = mr.GetAssemblyDefinition();
+                var moduleData = mr.GetModuleDefinition();
+                Debug.WriteLine("Assembly name is: " + mr.GetString(asssemblyData.Name));
+                Assert.Equal(_name.Name, mr.GetString(asssemblyData.Name));
+                Debug.WriteLine("Module name is: " + mr.GetString(moduleData.Name));
+                Assert.Equal(_name.Name, mr.GetString(moduleData.Name));
+                fs.Dispose();
```
AaronRobinsonMSFT: The `using` clause handles this.

src/Experiment/tests/MyClassTests.cs
```diff
@@ -1,14 +1,68 @@
...
         {
-            Assert.True(MyClass.ReturnTrue);
+            const String _fileName = "BlankAssembly.dll";
+            AssemblyBuilder assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(_name, System.Reflection.Emit.AssemblyBuilderAccess.Run);
+            assemblyBuilder.Save(_fileName);
+            Reader(_fileName);//It seems that sometimes ILSpy will point out a DLL is malformed but MetadataReader doesn't throw an error (MetadataLoadContext rather?) .
+        }
+       
+
+        private void Reader(String fileName)
```
AaronRobinsonMSFT:
```suggestion
        private void Reader(string fileName)
```

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
 using System.Text;
 using System.Threading.Tasks;
 using System.Reflection;
+using System.Reflection.Metadata.Ecma335;
+using System.Reflection.Metadata;
```
joperezr: NIT: we usually try to keep our `using` statements in alphabetical order and remove the unused ones. There is a VS context menu option that can do that for you.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
...
+using System.Runtime.CompilerServices;
 
 namespace System.Reflection.Emit.Experimental
 {
+
     public class AssemblyBuilder: System.Reflection.Assembly
     {
-        internal AssemblyBuilder() { }
+        private static readonly Guid _s_guid = new Guid("87D4DBE1-1143-4FAD-AAB3-1001F92068E6");//Some random ID, Need to look into how these should be generated.
+        private static readonly BlobContentId _s_contentId = new BlobContentId(_s_guid, 0x04030201);
```
joperezr: Where does `0x04030201` come from or why was it picked as a Stamp ID? If it the value is a constant, we should probably declare it as a constant with a comment explaining the value, and then passing that constant here.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -7,21 +7,69 @@
...
+        private BlobBuilder _emptyBlob = new BlobBuilder();
+        public override string? FullName { get; }
+        private static MetadataBuilder _metadata = new MetadataBuilder();
+        private static IDictionary<string,ModuleBuilder> moduleStorage = new Dictionary<string, ModuleBuilder>();
+        public AssemblyBuilder() { }
```
joperezr: NIT: Please leave an empty line to separate fields from methods, and also make sure there is an empty line between each Method.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,69 @@
...
+using System.Reflection.Metadata.Ecma335;
+using System.Reflection.PortableExecutable;
 
 namespace System.Reflection.Emit.Experimental
 {
+
     public class AssemblyBuilder: System.Reflection.Assembly
     {
-        internal AssemblyBuilder() { }
+        private static readonly Guid s_guid = Guid.NewGuid();//Some random ID, Need to look into how these should be generated.
```
joperezr: NIT: looks like we can get rid of the comment now.

src/Experiment/src/AssemblyBuilder.cs
joperezr: @MosheWolberg Looks like this feedback hasn't been addressed yet.

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Ecma335;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace System.Reflection.Emit.Experimental
+{
+    public class ModuleBuilder : System.Reflection.Module
+    {
+        private readonly Guid _s_guid;
```
joperezr: As per Jan comment, this is not needed

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,136 @@
...
+        public override System.Reflection.Assembly Assembly { get; }
+        public override string ScopeName { get; }
+
+        internal ModuleBuilder(string name, Assembly assembly)
+        {
+         //Example random GUID - I still need to understand when, how and why to generate them.
+        _s_guid= Guid.NewGuid();
+        ScopeName = name;
```
joperezr: NIT: indentation isn't right on this method.

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,136 @@
...
+        public void CreateGlobalFunctions()
+        {
+            throw new NotImplementedException();
+        }
+
+        // For all these methods, return type will be changed to "System.Reflection.Emit.Experiment.x" once non-empty modules/assemblies are supported.
+        public System.Reflection.Emit.EnumBuilder DefineEnum(string name, System.Reflection.TypeAttributes visibility, System.Type underlyingType) 
+        { 
+            throw new NotImplementedException(); 
+        }
```
joperezr:
```suggestion
        public System.Reflection.Emit.EnumBuilder DefineEnum(string name, System.Reflection.TypeAttributes visibility, System.Type underlyingType) 
            => throw new NotImplementedException();
```
NIT: all of these can just be simplifyed to not have braces for now.

src/Experiment/tests/BlankAssemblyTest.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using System.Runtime.CompilerServices;
+using Xunit;
+
+namespace Experiment.Tests
+{
+    public class BlankAssemblyTest
+    {
+        private readonly AssemblyName _name = new AssemblyName("MyBlankAssembly");
```
joperezr: NIT: instead of declaring a field, you can just pass this as a parameter to your helper function. That way that helper can be made static.

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,136 @@
...
+        public void SetCustomAttribute(System.Reflection.ConstructorInfo con, byte[] binaryAttribute)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void SetCustomAttribute(System.Reflection.Emit.CustomAttributeBuilder customBuilder)
+        {
+            throw new NotImplementedException();
+        }
+
```
joperezr: Nit: remove all these empty lines here

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
```
joperezr: Please add the licensing headers into this file as well as the test file.

src/Experiment/tests/BlankAssemblyTest.cs
```diff
@@ -0,0 +1,68 @@
...
+        public void GenerateBlankAssemblyReadToDisk()
+        {
+            const String _fileName = "BlankAssembly.dll";
+            AssemblyBuilder assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(_name, System.Reflection.Emit.AssemblyBuilderAccess.Run);
+            assemblyBuilder.Save(_fileName);
+            Reader(_fileName);//It seems that sometimes ILSpy will point out a DLL is malformed but MetadataReader doesn't throw an error (MetadataLoadContext rather?) .
+        }
+       
+
+        private void Reader(String fileName)
```
joperezr: Should this be something like:
`private static bool TryLoadAssembly(string filePath)`

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,68 @@
...
         }
 
-        public static System.Reflection.Emit.AssemblyBuilder DefineDynamicAssembly(System.Reflection.AssemblyName name, System.Reflection.Emit.AssemblyBuilderAccess access) 
-        { 
-            throw new NotImplementedException(); 
+        public static System.Reflection.Emit.Experimental.AssemblyBuilder DefineDynamicAssembly(System.Reflection.AssemblyName name, System.Reflection.Emit.AssemblyBuilderAccess access)
+        {
+            if (name == null || name.Name == null)
+            {
+                throw new ArgumentNullException();
```
danmoseley, line 70:
```suggestion
                throw new ArgumentNullException(nameof(name));
```
So the message includes the parameter's name

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -30,14 +71,36 @@ public static System.Reflection.Emit.AssemblyBuilder DefineDynamicAssembly(Syste
...
+        public System.Reflection.Emit.Experimental.ModuleBuilder? GetDynamicModule(string name) 
+        {
+            return _moduleStorage[name];
+        }
+
+        private static void WritePEImage(Stream peStream, MetadataBuilder metadataBuilder, BlobBuilder ilBuilder) // MethodDefinitionHandle entryPointHandle when we have main method.
+        {
+            // Create executable with the managed metadata from the specified MetadataBuilder.
+            var peHeaderBuilder = new PEHeaderBuilder(
+                imageCharacteristics: Characteristics.Dll //Start off with a simple DLL
```
danmoseley, line 135: Nit please one space after //

src/Experiment/tests/BlankAssemblyTest.cs
```diff
@@ -0,0 +1,71 @@
...
+                }
+                else
+                {
+                    throw new IOException();
+                }
+
+            }
+            finally
+            {
+                if (File.Exists(filePath))
```
danmoseley: No need to protect with Exists. If the file to be deleted does not exist, no exception is thrown.

danmoseley: Oh, since you want the Debug check, you would write it as `Debug.Assert(File.Exists...`

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,68 @@
...
+using System.Reflection.PortableExecutable;
 
 namespace System.Reflection.Emit.Experimental
 {
+
     public class AssemblyBuilder: System.Reflection.Assembly
     {
-        internal AssemblyBuilder() { }
+        private static readonly Guid s_guid = Guid.NewGuid();
+        private BlobBuilder _emptyBlob = new BlobBuilder();//For later
```
joperezr: If this is not used, let's just remove it and add it once we are ready to add it.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,68 @@
...
+                throw new ArgumentNullException();
+            }
+            if (_assemblyName==null||_assemblyName.Name==null)
+            {
+                throw new ArgumentNullException();
+            }
+            //Add assembly metadata
+            _metadata.AddAssembly(//Metadata is added for the new assembly - Current design - metdata generated only when Save method is called.
+               _metadata.GetOrAddString(value: _assemblyName.Name),
+               version: _assemblyName.Version ?? new Version(1, 0, 0, 0),
```
joperezr: .NET Framework defaults to `0.0.0.0` when a version isn't specified, so we should match that behavior here.

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+            ScopeName = name;
+            Assembly = assembly;
+        }
+
+        internal void AppendMetadata(MetadataBuilder metadataBuilder)
+        {
+            metadataBuilder.AddModule(
+                generation: 0,
+                metadataBuilder.GetOrAddString(ScopeName),
+                metadataBuilder.GetOrAddGuid(Guid.NewGuid()),//See above comment on GUID
```
joperezr: NIT: remove comment.

src/Experiment/src/AssemblyBuilder.cs
joperezr: Also, Culture defaults to `neutral` and PublicKeyToken to `null`

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,67 @@
...
+using System.Reflection.Metadata.Ecma335;
+using System.Reflection.PortableExecutable;
 
 namespace System.Reflection.Emit.Experimental
 {
+
     public class AssemblyBuilder: System.Reflection.Assembly
     {
-        internal AssemblyBuilder() { }
+        private static readonly Guid s_guid = Guid.NewGuid();
```
AaronRobinsonMSFT:
```suggestion
        private readonly Guid _guid = Guid.NewGuid();
```

Should be an instance field not a static field.

joperezr:
Just as an FYI for other reviewers, @MosheWolberg and I discussed and this PR will also include the support for interfaces. Some of the changes are already in, and @MosheWolberg will ping this PR once it is ready for another look.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,67 @@
...
+        }
 
         public void Save(string assemblyFileName)
         {
-            //New method
+            if (assemblyFileName == null)
+            {
+                throw new ArgumentNullException(nameof(assemblyFileName));
+            }
+            if (_assemblyName==null||_assemblyName.Name==null)
```
joperezr:
```suggestion
            if (_assemblyName == null || _assemblyName.Name == null)
```
NIT: adding spaces in conditions and statements help with readability

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System.Collections.Generic;
```
joperezr: All new files should have the license headers

src/Experiment/tests/BlankAssemblyTest.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void GenerateBlankAssemblyReadToDisk()
+        {
+            AssemblyName _name = new AssemblyName("MyBlankAssembly");
+            _name.Version = new Version("4.3");
+            const String _fileName = "BlankAssembly.dll";
+            AssemblyBuilder assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(_name, System.Reflection.Emit.AssemblyBuilderAccess.Run);
+            //We need to create module because even a blank assembly has one module.
+            assemblyBuilder.DefineDynamicModule(_name.Name);
+            assemblyBuilder.Save(_fileName);
+            TryLoadAssembly(_fileName,_name);//It seems that sometimes ILSpy will point out a DLL is malformed but MetadataReader doesn't throw an error (MetadataLoadContext rather?) .
```
joperezr, line 27: It would be good if TryLoadAssembly returns back the `Assembly` object, so that you can then assert things in the test after loading it. For example: it would be good in this test that we validate that the assembly name and version are what we expect

src/Experiment/tests/BlankAssemblyTest.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using Xunit;
+
+namespace Experiment.Tests
+{
+    public class BlankAssemblyTest
+    {
+        [Fact]
+        public void GenerateBlankAssemblyReadToDisk()
```
joperezr, line 18: Can we also add a test that only passes in the name and checks that version is 0.0.0.0, and culture and public key are none and null respectively.

src/Experiment/tests/InterfaceAssembly.cs
```diff
@@ -0,0 +1,72 @@
...
+            _name.Version = new Version("4.3");
+            const String _fileName = "MyInterfaceAssembly.dll";
+            AssemblyBuilder assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(_name, System.Reflection.Emit.AssemblyBuilderAccess.Run);
+            //We need to create module because even a blank assembly has one module.
+            ModuleBuilder moduleBuilder = assemblyBuilder.DefineDynamicModule(_name.Name);
+            moduleBuilder.DefineType(interfaceName, TypeAttributes.Interface);
+            assemblyBuilder.Save(_fileName);
+            TryLoadAssembly(_fileName,_name);//It seems that sometimes ILSpy will point out a DLL is malformed but MetadataReader doesn't throw an error (MetadataLoadContext rather?) .
+        }  
+    private static void TryLoadAssembly(string filePath, AssemblyName _name)
```
joperezr: This method could be moved to a test utilities class so it can be used on both tests

src/Experiment/tests/InterfaceAssembly.cs
```diff
@@ -0,0 +1,72 @@
...
+                MetadataReader mr = peReader.GetMetadataReader();
+                var asssemblyData = mr.GetAssemblyDefinition();
+                var moduleData = mr.GetModuleDefinition();
+                Debug.WriteLine("Assembly name is: " + mr.GetString(asssemblyData.Name));
+                Assert.Equal(_name.Name, mr.GetString(asssemblyData.Name));
+
+                Debug.WriteLine("Module name is: " + mr.GetString(moduleData.Name));
+                Assert.Equal(_name.Name, mr.GetString(moduleData.Name));
+
+                foreach (TypeDefinitionHandle tdefh in mr.TypeDefinitions)
```
joperezr: This part of the validation should go on the test, that way the helper method can be reused.

joperezr (commented):
Left a few more comments, please let us know when you are ready for another look.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,67 @@
...
+            _assemblyName = name;
+        }
 
         public void Save(string assemblyFileName)
         {
-            //New method
+            if (assemblyFileName == null)
+            {
+                throw new ArgumentNullException(nameof(assemblyFileName));
+            }
```
joperezr, line 36: NIT: new line after code blocks.

src/Experiment/src/MethodBuilder.cs
```diff
@@ -0,0 +1,191 @@
...
+            Module = declaringType.Module;
+        }
+
+        internal MethodDefinitionHandle AppendMetadata(MetadataBuilder _metadata)
+        {
+            // Encoding return type and parameters.
+            var methodSignature = new BlobBuilder();
+            ParametersEncoder _parEncoder;
+            ReturnTypeEncoder _retEncoder;
+            new BlobEncoder(methodSignature).
```
joperezr, line 37:
```suggestion
            _ = new BlobEncoder(methodSignature).
```
NIT: use discard.

src/Experiment/tests/InterfaceAssemblyTest.cs
```diff
@@ -0,0 +1,217 @@
...
+namespace Experiment.Tests
+{
+    public class InterfaceAssemblyTest
+    {
+        const bool _keepFiles = true; // keep files after testing for inspection
+        TempFileCollection _tfc;
+        string _assemblyFile;
+
+        [Fact]
+        public void Setup()
```
AaronRobinsonMSFT, line 25: This isn't the correct way to share assets between tests. See https://xunit.net/docs/shared-context.

src/Experiment/src/MethodBuilder.cs
```diff
@@ -0,0 +1,191 @@
...
+                    break;
+                case "System.Object":
+                    signature.Object();
+                    break;
+                case "System.String":
+                    signature.String();
+                    break;
+                default: throw new NotImplementedException("This parameter type is not yet supported");
+            }
+        }
```
joperezr, line 100:
```suggestion
        => type.FullName switch
        {
            "System.Boolean" => signature.Boolean(),
            "System.Byte"       => singature.Byte(),
            //...
            _                           => throw new NotImplementedException("This parameter type is not yet supported");
        };
```
NIT: use the new Switch syntax if you want to.

src/Experiment/src/MethodBuilder.cs
```diff
@@ -0,0 +1,191 @@
...
+        // These methods seems like they should be implemented next.
+        public System.Reflection.Emit.ParameterBuilder DefineParameter(int position, System.Reflection.ParameterAttributes attributes, string? strParamName)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void SetImplementationFlags(System.Reflection.MethodImplAttributes attributes)
+        {
+            throw new NotImplementedException();
+        }
```
joperezr, line 111: NIT: new line after method

src/Experiment/tests/InterfaceAssemblyTest.cs
```diff
@@ -0,0 +1,217 @@
...
+                }
+            }
+        }
+
+        [Fact]
+        public void ThreeInterfacesWithThreeMethods()
+        {
+            int _typeCount = 3;
+            int _methodCount = 3;
+            Setup();
```
AaronRobinsonMSFT, line 118: This is going to make testing very unstable.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,67 @@
...
 
 namespace System.Reflection.Emit.Experimental
 {
+
     public class AssemblyBuilder: System.Reflection.Assembly
     {
-        internal AssemblyBuilder() { }
+        private AssemblyName? _assemblyName;
+        private MetadataBuilder _metadata = new MetadataBuilder();
+        private IDictionary<string,ModuleBuilder> _moduleStorage = new Dictionary<string, ModuleBuilder>();
```
joperezr: As discussed offline, if we are only going to support one module, then let's get rid of the dictionary.

src/Experiment/tests/InterfaceAssemblyTest.cs
```diff
@@ -0,0 +1,217 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Reflection.Emit.Experimental;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using System.Runtime.InteropServices;
+using Xunit;
+
+namespace Experiment.Tests
```
AaronRobinsonMSFT, line 16: The testing methodology here is difficult to maintain and will likely be fragile as more complexity is introduced. Instead, I would consider the following.

Create a shape in C# for the type you want to build. For example, define the following interface in a new file that contains all the types you plan to add to a new assembly. Once the type is defined here you can use Reflection on the C# definition and use the resulting `MetadataLoadContext` assets to confirm the shape is identical. This will also allow you to use the `nameof` operator in the tests and make sure the strings and type details are checked by the compiler.

```csharp
namespace Tests
{
    public interface Interface1
    {
        string Func(int a);
    }
}
```

Usage example in test:
```csharp
const string _interfaceName = $"{nameof(Tests)}.{nameof(Tests.Interface1)}";
```

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Ecma335;
+
+namespace System.Reflection.Emit.Experimental
+{
+    public class ModuleBuilder : System.Reflection.Module
+    {
+        internal LinkedList<TypeBuilder> _typeStorage = new LinkedList<TypeBuilder>(); // LinkedList because each type needs to need know what methods are in the next type.
```
joperezr: Can you add bigger comment in here that talks about the special case of types with no methods, and explain why we need to track this list?

AaronRobinsonMSFT (commented):
It might be appropriate to set a defined point when this PR will be merged in. It is getting more difficult to review.

src/Experiment/tests/BlankAssemblyTest.cs
```diff
@@ -0,0 +1,64 @@
...
+            _name.Version = new Version("4.3");
+            const String _fileName = "BlankAssembly.dll";
+            AssemblyBuilder assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(_name, System.Reflection.Emit.AssemblyBuilderAccess.Run);
+            //We need to create module because even a blank assembly has one module.
+            assemblyBuilder.DefineDynamicModule(_name.Name);
+            assemblyBuilder.Save(_fileName);
+            TryLoadAssembly(_fileName,_name);//It seems that sometimes ILSpy will point out a DLL is malformed but MetadataReader doesn't throw an error (MetadataLoadContext rather?) .
+        }
+       
+    private static void TryLoadAssembly(string filePath, AssemblyName _name)
```
joperezr, line 30:
```suggestion
    private static bool TryLoadAssembly(string filePath, AssemblyName _name, [NotNullWhen(true)] out Assembly? assembly)
```

src/Experiment/src/AssemblyBuilder.cs
AaronRobinsonMSFT: > if we are only going to support one module

To be clear about this, .NET Core doesn't support multi-module assemblies so this is "by design". See https://github.com/dotnet/coreclr/commit/1aa7d6b8796f7e28a63162117c5bb16a207a472b.

src/Experiment/tests/InterfaceAssemblyTest.cs
AaronRobinsonMSFT: Concrete example:

```csharp
namespace Test
{
    public interface IInterface
    {
        string Func(int a);
    }
}

static MetadataLoadContext CreateTypeInAssembly(Type type)
{
    var asmBuilder = AssemblyBuilder.DefineDynamicAssembly(default, default);
    var modBuilder = asmBuilder.DefineDynamicModule(default);

    var typeBuilder = modBuilder.DefineType($"{type.Namespace}.{type.Name}");
    foreach (MethodInfo mi in type.GetMethods())
    {
        typeBuilder.DefineMethod(mi.Name, mi.Attributes);
    }

    // fields, events, properties, etc.

    return default;
}

static Assembly ReadTypeFromMetadata(MetadataLoadContext cxt);

public static void Main(string[] args)
{
    // Create an assembly and return the MetadataLoadContext of the Assembly on disk.
    MetadataLoadContext cxt = CreateTypeInAssembly(typeof(Test.IInterface));

    // From the MetadataLoadContext, read in the assembly and module.
    Assembly asmFromMetadata = ReadTypeFromMetadata(cxt);
    Module moduleFromMetadata = asmFromMetadata.Modules.First();

    // Validate metadata matches the type defined in the test
    Type typeFromMetadata = moduleFromMetadata.GetType($"{nameof(Test)}.{nameof(Test.IInterface)}");
}
```

MosheWolberg: Will do a separate PR with this new testing methodology.

MosheWolberg:
Can we merge now and I'll push the testing changes in a new PR?

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,78 @@
...
+using System.Reflection.PortableExecutable;
 
 namespace System.Reflection.Emit.Experimental
 {
+
     public class AssemblyBuilder: System.Reflection.Assembly
     {
-        internal AssemblyBuilder() { }
+        private bool _previouslySaved=false;
+        private AssemblyName? _assemblyName;
```
jkotas:
```suggestion
        private AssemblyName _assemblyName;
```

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,78 @@
...
+            if (assemblyFileName == null)
+            {
+                throw new ArgumentNullException(nameof(assemblyFileName));
+            }
+
+            if (_assemblyName == null || _assemblyName.Name == null)
+            {
+                throw new ArgumentNullException(nameof(_assemblyName));
+            }
+
```
jkotas:
```suggestion
```
This was validated in `DefineDynamicAssembly` already. No need to do it again here.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,78 @@
...
+            if(_module == null)
+            {
+                throw new InvalidOperationException("Assembly needs at least one module defined");
+            }
+
+            //Add assembly metadata
+            _metadata.AddAssembly(//Metadata is added for the new assembly - Current design - metdata generated only when Save method is called.
```
jkotas: This is fine design. It means that you do not need actually need to do `new MetadataBuilder()` upfront in the constructor, but you can do it here instead. This can be addressed later.

src/Experiment/src/AssemblyBuilder.cs
jkotas: I see many of these nit with extra spaces, etc.

Are you using Visual Studio as the editor? You can go to `Edit / Advanced / Format Document` that will fix many of these nits for you.

src/Experiment/src/AssemblyBuilder.cs
jkotas:
```suggestion
            _metadata.AddAssembly(//Metadata is added for the new assembly - Current design - metadata generated only when Save method is called.
```

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,78 @@
...
+               _metadata.GetOrAddString(value: _assemblyName.Name),
+               version: _assemblyName.Version ?? new Version(0, 0, 0, 0),
+               culture: (_assemblyName.CultureName==null) ?  default : _metadata.GetOrAddString(value: _assemblyName.CultureName),
+               publicKey: (_assemblyName.GetPublicKey() is byte[] publicKey) ? _metadata.GetOrAddBlob(value: publicKey) : default,
+               flags: (AssemblyFlags) _assemblyName.Flags,
+               hashAlgorithm: AssemblyHashAlgorithm.None);//It seems AssemblyName.HashAlgorithm is obslete so default value used.
```
jkotas:
```suggestion
               hashAlgorithm: AssemblyHashAlgorithm.None);//AssemblyName.HashAlgorithm is obsolete so default value used.
```

src/Experiment/src/MethodBuilder.cs
```diff
@@ -0,0 +1,195 @@
...
+        public override string Name { get; }
+        public override System.Reflection.MethodAttributes Attributes { get; }
+        public override System.Reflection.CallingConventions CallingConvention { get; }
+        public override System.Type? DeclaringType { get; }
+        public override System.Reflection.Module Module { get; }
+        internal MethodDefinitionHandle _handle;
+        private Type? _returnType;
+        private Type[]? _parameters;
+
+        internal MethodBuilder(string name, System.Reflection.MethodAttributes attributes, CallingConventions callingConventions, Type? returnType, Type[]? mthdParams,Type declaringType)
```
jkotas:
```suggestion
        internal MethodBuilder(string name, System.Reflection.MethodAttributes attributes, CallingConventions callingConventions, Type? returnType, Type[]? parameters, Type declaringType)
```

src/Experiment/src/MethodBuilder.cs
```diff
@@ -0,0 +1,195 @@
...
+        {
+            Name = name;
+            Attributes = attributes;
+            CallingConvention = callingConventions;
+            _returnType = returnType;
+            _parameters = mthdParams;
```
jkotas:
```suggestion
            _parameters = parameters;
```

src/Experiment/src/MethodBuilder.cs
```diff
@@ -0,0 +1,195 @@
...
+            DeclaringType = declaringType;
+            Module = declaringType.Module;
+        }
+
+        internal MethodDefinitionHandle AppendMetadata(MetadataBuilder _metadata)
```
jkotas, line 31: `_metadata` argument name should be replaced with `metadata` everywhere. We do not use `_` for argument names.

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,137 @@
+﻿using System.Collections.Generic;
```
jkotas:
```suggestion
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.

﻿using System.Collections.Generic;
```

src/Experiment/src/ModuleBuilder.cs
```diff
@@ -0,0 +1,137 @@
...
+{
+    public class ModuleBuilder : System.Reflection.Module
+    {
+        /* According to ECMA-335, (II.22.37), a TypeDef refrences the row of its first method in the MethodDef table. It owns all methods that lie between 
+         * that row and the row refrenced by the next TypeDef.
+         * For a TypeA that has no methods followed by a TypeB that has methods, MetaDataBuilder indicates this by having TypeA refrence the row of the first method of TypeB. Since TypeB
+         * has the same row refrence, TypeA will have no methods. (Why it needs to do this is unclear to me, since according to ECMA, ibid. 18, MethodList can be null).
+         * See https://docs.microsoft.com/en-us/dotnet/api/system.reflection.metadata.ecma335.metadatabuilder.addtypedefinition?view=net-6.0#system-reflection-metadata-ecma335-metadatabuilder-addtypedefinition(system-reflection-typeattributes-system-reflection-metadata-stringhandle-system-reflection-metadata-stringhandle-system-reflection-metadata-entityhandle-system-reflection-metadata-fielddefinitionhandle-system-reflection-metadata-methoddefinitionhandle)
+         * Using a LinkedList allows Types to inspect later Types when needed.
```
jkotas, line 17: Per our offline conversation, this linked list can be replaced with a just a regular `List<TypeBuilder>` and keeping the track of next method token instead

jkotas (approved):

jkotas:
> Can we merge now and I'll push the testing changes in a new PR?

We can merge this to make progress, with the caveat that the feedback will be addressed later.

For next time, it is best to keep the PR focused on one thing, drive that one thing to 100% and get that merged.

If you keep adding more features to a PR, it is best way to ensure that reviewing and merging it will drag forever. It is also not efficient for reviewers since they have to look at the same code multiple times.

src/Experiment/src/AssemblyBuilder.cs
```diff
@@ -1,27 +1,73 @@
...
+            {
+                throw new InvalidOperationException("Cannot save an assembly multiple times");
+            }
+
+            if (assemblyFileName == null)
+            {
+                throw new ArgumentNullException(nameof(assemblyFileName));
+            }
+
+            if(_module == null)
```
jkotas, line 43:
```suggestion
            if (_assemblyName == null || _assemblyName.Name == null)
            {
                throw new ArgumentNullException(nameof(_assemblyName));
            }
            
            if(_module == null)
```
Ok, this was actually needed to make the compiler happy. Adding it back

PR #1900, opened 2022-06-18
Initial hot cold splitting support for crossgen2/VM

cshung:
This PR marks the start of the work to get crossgen2 to support hot/cold splitting.

Here are the changes:

1. Added a section named `Scratch`, to be described in (13) below for more detail.
2. Create the `_methodColdCode` from `_coldCode` and `_coldCodeRelocs` during `PublishCode`.
3. Null `_methodColdCode` and empty `_coldCodeRelocs` during `CompileMethodCleanup`.
4. Create `_methodColdCode` during `allocMem`.
5. Change `reserveUnwindInfo` to ignore unwind reservation for `coldCode`.
6. Change `allocUnwindInfo` to ignore unwindInfo allocation for 0 bytes.
7. Change `findKnownBlocks` to return `_coldCodeReloc` when `BlockType.ColdCode` is asked for.
8. Change `recordRelocation` to set `relocTarget` to `_methodColdCodeNode` when `BlockType.ColdCode` is asked for.
9. Implement `MethodColdCodeNode`
- Storing the cold code.
- Make sure sorting will place cold code blocks after hot code blocks.
- Make sure cold code blocks are ordered the same as the hot code blocks.
10. Change `MethodGCInfoNode` to 
- Emit the chained unwind info for the cold code block, after the hot code GCInfo, so the hot and cold code unwind info are currently mixed.
- Changed  `CalculateFuncletOffsets` so that the offset for the cold code chained unwind info is included if it exists.
11. Changed `MethodWithGCInfo` to:
- Have a reference to the `MethodColdCodeNode`.
- Changed `GetFixupBlob` to include relocation for the cold code.
12. Changed `RuntimeFunctionTableNode` to: 
- Emit the RuntimeFunction entries for the cold code block after the hot code blocks.
- Prepare the cold to hot code mapping for `ScratchNode`.
13. Implement `ScratchNode` - the session is simply an array of cold code runtime function index to hot code runtime function index, so all even entries (e.g. 0, 2, 4) are cold code runtime function index and the corresponding odd indexes (e.g. 1, 3, 5) are the corresponding runtime function index for the corresponding hot code. The array is sorted by the cold code indexes, by (9), the hot code indexes should be sorted as well.
14. Implemented the `--hot-cold-splitting` command-line option, and turn on the `CORJIT_FLAG_PROCSPLIT` flag as needed.
15. Implemented the various code manager contracts:
- `JitCodeToMethodInfo` already gave us a runtime function index, we do not know if it is hot or cold, so we use the scratch table to find it. If we got an even index, that would be cold code and therefore we map it back to the hot one.
- Change `GetFuncletStartOffsets` to return the cold code block as well - technically the cold block is not a funclet, but it appears that the debugger might want it.
- Override `IsFunclet` so that it avoids returning `true` for the cold code block, by searching the scratch table.
- Change `JitTokenToMethodRegionInfo` so that it will find and return the cold code block - again, by search the scratch table since we already have the hot part runtimeFunction.
- Change `DebugDebugger` so that it uses `EECodeInfo` to compute relative offset instead of a simple subtraction, that will allow us to account for the cold code.

This work is far from complete - here is a list of issues filed for the project - these are known issues for now:
- https://github.com/dotnet/runtimelab/issues/1901
- https://github.com/dotnet/runtimelab/issues/1902
- https://github.com/dotnet/runtimelab/issues/1903
- https://github.com/dotnet/runtimelab/issues/1904
- https://github.com/dotnet/runtimelab/issues/1905
- https://github.com/dotnet/runtimelab/issues/1906
- https://github.com/dotnet/runtimelab/issues/1907
- https://github.com/dotnet/runtimelab/issues/1908
- https://github.com/dotnet/runtimelab/issues/1909
- https://github.com/dotnet/runtimelab/issues/1910
- https://github.com/dotnet/runtimelab/issues/1911
- https://github.com/dotnet/runtimelab/issues/1912
- https://github.com/dotnet/runtimelab/issues/1913

@amanasifkhalid, @BruceForstall 

@dotnet/crossgen-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/DependencyAnalysis/ReadyToRun/MethodColdCodeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/DependencyAnalysis/ReadyToRun/MethodColdCodeNode.cs
@@ -1,0 +1,65 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Diagnostics;
+using Internal.Text;
+using Internal.TypeSystem;
+
+namespace ILCompiler.DependencyAnalysis.ReadyToRun
+{
+    public class MethodColdCodeNode : ObjectNode, ISymbolDefinitionNode
+    {
+        private ObjectData _methodColdCode;
+        private MethodDesc _owningMethod;
+
+        public MethodColdCodeNode(MethodDesc owningMethod)
+        {
+            _owningMethod = owningMethod;
+        }
+
+        public int Offset => 0;
+
+        public override ObjectNodeSection Section
+        {
+            get
+            {
+                // TODO, Unix
+                return ObjectNodeSection.ManagedCodeWindowsContentSection;
+            }
+        }
+
+        public override bool IsShareable => false;
+
+        // This ClassCode must be larger than that of MethodCodeNode to ensure it got sorted at the end of the code
+        public override int ClassCode => 788492408;
+
+        public override bool StaticDependenciesAreComputed => _methodColdCode != null;
+
+        public void AppendMangledName(NameMangler nameMangler, Utf8StringBuilder sb)
+        {
+            sb.Append("__coldcode_" + nameMangler.GetMangledMethodName(_owningMethod));
+        }
+
+        public override int CompareToImpl(ISortableNode other, CompilerComparer comparer)
+        {
+            MethodColdCodeNode otherNode = (MethodColdCodeNode)other;
+            return comparer.Compare(_owningMethod, otherNode._owningMethod);
+        }
+
+        public override ObjectData GetData(NodeFactory factory, bool relocsOnly = false) => _methodColdCode;
+
+        protected override string GetName(NodeFactory context) => throw new NotImplementedException();
+
+        public void SetCode(ObjectData data)
+        {
+            Debug.Assert(_methodColdCode == null);
+            _methodColdCode = data;
+        }
+
+        public int GetColdCodeSize()
+        {
+            return _methodColdCode.Data.Length;
+        }
+    }
+}
--- a/src/coreclr/vm/codeman.cpp
+++ b/src/coreclr/vm/codeman.cpp
@@ -5830,7 +5830,7 @@ GCInfoToken ReadyToRunJitManager::GetGCInfoToken(const METHODTOKEN& MethodToken)
     SIZE_T nUnwindDataSize;
     PTR_VOID pUnwindData = GetUnwindDataBlob(baseAddress, pRuntimeFunction, &nUnwindDataSize);
 
-    // GCInfo immediatelly follows unwind data
+    // GCInfo immediately follows unwind data
     PTR_BYTE gcInfo = dac_cast<PTR_BYTE>(pUnwindData) + nUnwindDataSize;
     UINT32 gcInfoVersion = JitTokenToGCInfoVersion(MethodToken);
 
@@ -6073,8 +6073,6 @@ BOOL ReadyToRunJitManager::JitCodeToMethodInfo(RangeSection * pRangeSection,
         SUPPORTS_DAC;
     } CONTRACTL_END;
 
-    // READYTORUN: FUTURE: Hot-cold spliting
-
     // If the address is in a thunk, return NULL.
     if (GetStubCodeBlockKind(pRangeSection, currentPC) != STUB_CODE_BLOCK_UNKNOWN)
     {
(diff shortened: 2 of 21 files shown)


src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/DependencyAnalysis/ReadyToRun/MethodColdCodeNode.cs
```diff
@@ -0,0 +1,65 @@
...
+            get
+            {
+                // TODO, Unix
+                return ObjectNodeSection.ManagedCodeWindowsContentSection;
+            }
+        }
+
+        public override bool IsShareable => false;
+
+        // This ClassCode must be larger than that of MethodCodeNode to ensure it got sorted at the end of the code
```
MichalStrehovsky, line 34: ClassCode should be chosen by a fair dice roll. There are no guarantees on ordering and this will be fragile/subtle.

I would suggest creating a new ObjectNodeSection for it and have the PE emitter emit it however it sees fit (at the end of the physical .text section?).

It's similar how this is done for native code: https://devblogs.microsoft.com/oldnewthing/20181107-00/?p=100155 where we would probably use `.text` for hot code and `.text$cold` for cold code.

The PE emitter can apply additional sorting/policies within the section.

cshung: I filed #1912 to track this issue, meanwhile, would you mind if I merge this in first to unblock the team? 
I am trying to get this feature branch ready so that our interns can start work on top of it.

BruceForstall (approved):
This is a good start. Let's get it merged so we can enable collaboration.

BruceForstall:
Added https://github.com/dotnet/runtimelab/issues/1913

src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/DependencyAnalysis/ReadyToRun/MethodColdCodeNode.cs
MichalStrehovsky: Didn't realize this is runtimelab (I got pinged because of the codeowners file). This is good enough for runtimelab for sure

BruceForstall:
Have you followed all the instructions on https://github.com/dotnet/runtimelab/blob/docs/CreateAnExperiment.md to make this a "proper" runtimelab feature branch?

cshung:
I merged the PR for now, feel free to work on top of it.
In the meantime, I will follow up with @hoyosjs to make sure the branch is set up correctly.
In case it isn't, there might be some churns to the actual commits, but it shouldn't impact the code at all.

src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/DependencyAnalysis/ReadyToRun/MethodColdCodeNode.cs
trylek: I concur with Michal that a new ObjectNode section should be used for cold code (not sure whether it needs to be Windows / Unix-specific, I don't know the rationale for the pre-existing choice regarding hot code) and ReadyToRunObjectWriter should decode it to a new PE code section to be put after the (hot) code, see here:

https://github.com/dotnet/runtime/blob/8558cdc3e26a0e28c386168be519f818678aadef/src/coreclr/tools/aot/ILCompiler.ReadyToRun/ObjectWriter/R2RPEBuilder.cs#L253

Alternatively (if there's pushback towards generating two code sections as indicated by the code comment at line 256) we'll probably need to tweak the code around

https://github.com/dotnet/runtime/blob/8558cdc3e26a0e28c386168be519f818678aadef/src/coreclr/tools/aot/ILCompiler.ReadyToRun/CodeGen/ReadyToRunObjectWriter.cs#L301

to perhaps stash cold code sections instead of sending them to the section writer right away and flush them after all hot code has been processed.

src/coreclr/vm/codeman.cpp
```diff
@@ -6194,11 +6205,64 @@ DWORD ReadyToRunJitManager::GetFuncletStartOffsets(const METHODTOKEN& MethodToke
...
+
+    ReadyToRunInfo * pInfo = JitTokenToReadyToRunInfo(pCodeInfo->GetMethodToken());
+
+    COUNT_T nRuntimeFunctions = pInfo->m_nRuntimeFunctions;
+    PTR_RUNTIME_FUNCTION pRuntimeFunctions = pInfo->m_pRuntimeFunctions;
+
+    ULONG methodIndex = (ULONG)(pCodeInfo->GetFunctionEntry() - pRuntimeFunctions);
+
+    // If it is either the main hot-code or the cold code, then it is not a funclet
+    for (DWORD i = 0; i < pInfo->m_nScratch; i++)
```
trylek, line 6249: This looks pretty inefficient and may show up in perf runs; ultimately we may need to encode this as a VertexMap or something similar in the R2R native PE file to allow for faster lookup.

trylek (commented):
I have added some feedback that should be ultimately addressed by the final implementor of hot-cold code splitting in Crossgen2 (Eugenio); in general the code changes seem reasonable to me, thanks Andrew for providing our intern with such a great head start!

src/coreclr/vm/codeman.cpp
cshung: Yes, it is known and it is tracked. To get to a better data structure, we have #1903. To avoid repeating searching, we have #1904.

hoyosjs:
> I merged the PR for now, feel free to work on top of it. In the meantime, I will follow up with @hoyosjs to make sure the branch is set up correctly. In case it isn't, there might be some churns to the actual commits, but it shouldn't impact the code at all.

Looked at all things I can think of - looks fine to me

MichalStrehovsky:
> > I merged the PR for now, feel free to work on top of it. In the meantime, I will follow up with @hoyosjs to make sure the branch is set up correctly. In case it isn't, there might be some churns to the actual commits, but it shouldn't impact the code at all.
> 
> Looked at all things I can think of - looks fine to me

The instructions also suggest deleting `.github/CODEOWNERS`. It's the reason why I was pinged for review (I don't mind that part). But it will also cause a lot of people to be pinged for review if you ever do a merge pull request from runtime main branch (because those touch a lot of files). Deleting it will avoid bothering those people. Doesn't apply if you don't intend to merge from main.

Issue #1901, opened 2022-06-18
[Hot Cold Splitting] Generate the cold code in the proper session for Unix.


cshung:
Right now, the cold code for Unix is generated in `ObjectNodeSection.ManagedCodeWindowsContentSection`, we need to put it in a proper place for Unix.

PR #1916, opened 2022-06-22
Generated the cold code in the proper session

EugenioPena:
Fixes https://github.com/dotnet/runtimelab/issues/1901

Validated the correct OS target to generate the code in the proper session depending on whether it's Unix or Windows.


cshung:
@EugenioPena, did you get a chance to run some tests on Linux to make sure it works?

PR #1926, opened 2022-07-02
Support EH funclet splitting in Crossgen2

amanasifkhalid:
Opening this to share initial Crossgen2 work with @cshung , @EugenioPena .

(EH splitting is enabled in the JIT, so expect pipelines to fail for now).

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -3247,10 +3247,10 @@ void Compiler::compInitOptions(JitFlags* jitFlags)
 #endif
     }
 
-#ifdef DEBUG
-    // TODO: Implement support for EH splitting in Crossgen2.
-    opts.compProcedureSplittingEH = false;
-#endif
+// #ifdef DEBUG
+//     // TODO: Implement support for EH splitting in Crossgen2.
+//     opts.compProcedureSplittingEH = false;
+// #endif
 
 #ifdef DEBUG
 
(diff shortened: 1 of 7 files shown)


cshung:
The NativeAOT compilation of crossgen2 on Linux Release is failing because during the compilation of `[ILCompiler.DependencyAnalysisFramework]ILCompiler.RequireTotalOrderAssert.__GetFieldHelper(int32,EETypePtr&)`, we have got a call to `allocUnwindInfo` with `pUnwindInfo == null`.

In my design, I was thinking using `pUnwindInfo == null` to signal this cold main function has no unwind info and therefore we should generate a chained unwind info instead. The fact that this function's got a call to `allocUnwindInfo` with `pUnwindInfo == null` in NativeAOT (i.e. without hot-cold splitting at all) surprised me, we need to do something about that. For now, my patch should fix the compilation so that it at least compiles.

amanasifkhalid:
Interesting, thanks for pushing a fix for that. There is logic [in the JIT](https://github.com/dotnet/runtimelab/blob/775723ee1a191c52c8de27a9bfd7a83c23abe53a/src/coreclr/jit/compiler.cpp#L3202) to prevent hot/cold splitting on NativeAOT for now (since we aren't testing it in our pipelines), so it looks like my changes to unwind info generation for splitting EH funclets may have introduced a regression in NativeAOT that we didn't catch. I'll look into the JIT side of things.

[Here](https://github.com/dotnet/runtimelab/blob/feature/hot-cold-splitting/src/coreclr/jit/unwindamd64.cpp#L823) is the JIT logic for determining `pUnwindInfo` -- looks like we might be running into an issue where `func->cfiCodes->size() = 0`.

amanasifkhalid:
Scratch my last comment -- after discussing, it looks like it is possible for hot functions (such as a trivial, empty function) to not have any unwind info. This warrants changing our design, as we cannot rely on our assumption that `pUnwindInfo` will only be null for cold main function fragments. We now check if `pUnwindInfo == null` and `pColdCode != null` to determine if chained unwind info is needed.

Issue #1918, opened 2022-06-30
[Hot cold splitting] Support split-eh


cshung:
As the PR https://github.com/dotnet/runtime/pull/71236 is merged, we should update crossgen2 to support cold funclets.

Here are a few things to do:

On the JIT side:

- [x] Consider making eh-split configurable. As you can see, there is a lot of work on the crossgen2/vm side to catch up. It would be bad if we can't update the lab branch before we have done all of these.

- [x] Make sure the JIT reports the unwind info accurately. The expectation is we get exactly one pair of `reserveUnwindInfo` and `allocUnwindInfo` for each runtime function. There are only two cases, case 1 is that it is the main function or a funclet, then in the JIT, we should have generated the right unwind info data, we need to right data and the right flags. Another case is that it is a cold block that is not a funclet, and we would like to get a callback with the right range but with a `unwindSize == 0` so that we know we have to generate a chained unwind info.

On the crossgen2 side:

- [x] Expect more than 1 calls to `reserveUnwindInfo` [here](https://github.com/dotnet/runtimelab/blob/7ccc22fabe7d516d9d8ea7c124fa79dbca654e8a/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs#L3465) and `allocUnwindInfo` [here](https://github.com/dotnet/runtimelab/blob/7ccc22fabe7d516d9d8ea7c124fa79dbca654e8a/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs#L3473) for the cold code. They are both in `CorInfoImpl.cs` and it is our entry point to the JIT questions. Before eh-splitting, the only cold funclet appears to be always zero sized, so I just ignored it [here](https://github.com/dotnet/runtimelab/blob/7ccc22fabe7d516d9d8ea7c124fa79dbca654e8a/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs#L3502). After eh-splitting, there could be non-trivial unwind-info allocated with the cold funclet, and we need to remember them in a structure parallel to `_frameInfos`.

[...]

PR #1929, opened 2022-07-07
Support cold EH funclets in VM

amanasifkhalid:
This PR closes out the remaining tasks in #1918 by adding support for hot/cold splitting of EH funclets in the VM. This support requires the following changes:
* In `ReadyToRunJitManager::JitCodeToMethodInfo`, when searching the Scratch table for the corresponding hot MethodIndex of a given cold MethodIndex, we now handle MethodIndices of cold EH funclets that aren't in the table (i.e. when part of the main body is split).
* `ReadyToRunJitManager::IsFunclet` now returns true for cold EH funclets that are in the Scratch table.

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -3247,11 +3247,6 @@ void Compiler::compInitOptions(JitFlags* jitFlags)
 #endif
     }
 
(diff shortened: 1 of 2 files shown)


amanasifkhalid:
Tagging @BruceForstall

PR #1933, opened 2022-07-11
Custom attributes

MosheWolberg:
This is the core functionality for Custom Attributes.

At present, only Custom Attributes on types are supported but that extending to Assemblies, Members etc. shouldn't be difficult since the foundation is in place.
I reorganized several key parts of the projects:

1. Most storage and all construction of metadata was centralized into the ModuleBuilder class.
This made more sense than the previous local construction the metadata of each type, member etc within itself - a lot of metadata across types may overlap (for example, assembly refs) so constructing data in the central module made sense and also simplifies the code going forward.

2.  Actual metadata writing moved to static Metadata Helper class.

3. Signature construction moved to Signature Helper class. 

And then I added necessary code for Custom Attributes, (porting needed encoder classes as well like Custom Attribute Builder).

Fields are next but I want to tie up a few loose ends before then (I'll share a list of them) - mainly, the SignatureEncoder class is currently just a basic prototype I built much earlier but to support more complex things like arrays in signatures I should port the real SignatureEnconder.


MosheWolberg:
@joperezr

MosheWolberg:
Hey, not urgent, not to ping or pressure anyone, I believe it's a very busy time with preview 7, - but just tagging you guys because it doesn't look like you were automatically tagged.
@ericstj @jkotas @AaronRobinsonMSFT

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+using System.Buffers.Binary;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+
+namespace System.Reflection.Emit.Experimental
+{
+    public class CustomAttributeBuilder
+    {
+        internal ConstructorInfo m_con;
```
joperezr: NIT: Why do we have `m_` prefixes for these? I'd assume these should all just be prefixed by `_`. Also, for this particular one, can we name it `_constructorInfo`?

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+        private object?[] m_constructorArgs;
+        internal byte[] m_blob;
```
joperezr: NIT: it would be good if we can also add some comments that explain what is this used for, and why it's internal instead of private.

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+
+        public ConstructorInfo Con { get => m_con;  }
```
joperezr: `Con` => `Constructor`. Also, where is it used?

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+
+        // public constructor to form the custom attribute with constructor and constructor
+        // parameters.
```
joperezr, line 23: Instead of adding  normal comments `//`, you could just add tripple-slash comments instead `///` which VS will automatically template for you so that you get a nice intellisense experience when consuming these.

src/Experiment/src/CustomAttributeBuilder.cs
joperezr: Same goes for all of the places where you take a parameter named `con`, we should just instead use `constructorInfo`

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+            ArgumentNullException.ThrowIfNull(namedProperties);
+            ArgumentNullException.ThrowIfNull(propertyValues);
+            ArgumentNullException.ThrowIfNull(namedFields);
+            ArgumentNullException.ThrowIfNull(fieldValues);
+
+#pragma warning disable CA2208 // Instantiate argument exceptions correctly, combination of arguments used
+            if (namedProperties.Length != propertyValues.Length)
+                throw new ArgumentException("namedProperties, propertyValues differ");
+            if (namedFields.Length != fieldValues.Length)
+                throw new ArgumentException("namedFields, fieldValues differ");
```
joperezr:
```suggestion
            if (namedProperties.Length != propertyValues.Length)
            {
                throw new ArgumentException($"{nameof(namedProperties)} and {nameof(propertyValues)} should have the same length.");
            }

            if (namedFields.Length != fieldValues.Length)
            {
                throw new ArgumentException($"{nameof(namedFields)} and {nameof(fieldValues)} should have the same length.");
            }
```

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+#pragma warning restore CA2208
+
+            if ((con.Attributes & MethodAttributes.Static) == MethodAttributes.Static ||
+                (con.Attributes & MethodAttributes.MemberAccessMask) == MethodAttributes.Private)
+                throw new ArgumentException("Bad Constructor");
```
joperezr: Can we provide better exception messages like: `The passed-in constructor is either static or private`

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+                    case TypeCode.SByte:
+                    case TypeCode.Byte:
+                    case TypeCode.Int16:
+                    case TypeCode.UInt16:
+                    case TypeCode.Int32:
+                    case TypeCode.UInt32:
+                    case TypeCode.Int64:
+                    case TypeCode.UInt64:
+                        return true;
+                    default:
```
joperezr, line 223: Is this just the case for nint and nuint?

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+            paramTypes = Array.ConvertAll(parameters, parameter => parameter.ParameterType);
+
+            // Since we're guaranteed a non-var calling convention, the number of arguments must equal the number of parameters.
+            if (paramTypes.Length != constructorArgs.Length)
+                throw new ArgumentException("Bad Parameter Count For Constructor");
+
+            // Verify that the constructor has a valid signature (custom attributes only support a subset of our type system).
+            for (i = 0; i < paramTypes.Length; i++)
+                if (!ValidateType(paramTypes[i]))
+                    throw new ArgumentException("Bad Type In Custom Attribute");
```
joperezr: Should we mention here which one was the one at fault?

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+                    {
+                        throw new ArgumentNullException($"{nameof(constructorArgs)}[{i}]");
+                    }
+                    continue;
+                }
+                VerifyTypeAndPassedObjectType(paramTypes[i], constructorArg.GetType(), $"{nameof(constructorArgs)}[{i}]");
+            }
+
+            // Allocate a memory stream to represent the CA blob in the metadata and a binary writer to help format it.
+            MemoryStream stream = new MemoryStream();
```
joperezr, line 108: shouldn't these two be wrapped in a using statement to get them disposed?

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+
+                // Validate property type.
+                if (!ValidateType(propType))
+                    throw new ArgumentException("Bad Type In Custom Attribute");
+
+                // Property has to be writable.
+                if (!property.CanWrite)
+                    throw new ArgumentException("Not A Writable Property");
+
+                //// Property has to be from the same class or base class as ConstructorInfo. Note: This edge case is currently difficult to port and needs to be added into prototype.
```
joperezr: Should we remove this for now and instead have an issue tracking this?

src/Experiment/src/CustomAttributeBuilder.cs
```diff
@@ -0,0 +1,581 @@
...
+                Type fldType = namedField.FieldType;
+                object? fieldValue = fieldValues[i];
+                if (fieldValue == null && fldType.IsValueType)
+                    throw new ArgumentNullException("fieldValues[" + i + "]");
+
+                // Validate field type.
+                if (!ValidateType(fldType))
+                    throw new ArgumentException("Bad Type In Custom Attribute");
+
+                // Field has to be from the same class or base class as ConstructorInfo. Note: This edge case is currently difficult to port and needs to be added into prototype.
```
joperezr: Same here, perhaps remove this and just track it in an issue.

src/Experiment/src/SignatureHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Ecma335;
+
+namespace System.Reflection.Emit.Experimental
+{
+    //This is prototype code, to generate simple signatures.
+    //For more complex signatures, port System.Reflection.Emit's SignatureHelper.
+    internal static class SignatureHelper
+    {
+        internal static BlobBuilder MethodSignatureEnconder(ParameterInfo[]? parameters, ParameterInfo? returnType, bool isInstance)
```
AaronRobinsonMSFT, line 13:
```suggestion
        internal static BlobBuilder MethodSignatureEncoder(ParameterInfo[]? parameters, ParameterInfo? returnType, bool isInstance)
```

Several places have the same typo.
```suggestion
        internal static BlobBuilder MethodSignatureEnconder(ParameterInfo[]? parameters, ParameterInfo? returnType, bool isInstance)
```

src/Experiment/src/SignatureHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+        internal static BlobBuilder MethodSignatureEnconder(Type[]? parameters, Type? returnType, bool isInstance)
+        {
+            // Encoding return type and parameters.
+            var methodSignature = new BlobBuilder();
+
+            ParametersEncoder _parEncoder;
+            ReturnTypeEncoder _retEncoder;
+
+            new BlobEncoder(methodSignature).
```
AaronRobinsonMSFT, line 38: I'd place the `.` on the next line.

src/Experiment/src/TypeNameBuilder.cs
```diff
@@ -0,0 +1,346 @@
...
+
+        private void OpenGenericArguments()
+        {
+            _instNesting++;
+            _firstInstArg = true;
+
+            if (_useAngleBracketsForGenerics)
+                Append('<');
+            else
+                Append('[');
```
AaronRobinsonMSFT, line 36: Seems common enough to make a single method.

AaronRobinsonMSFT: Could be combined to do both open and close.

src/Experiment/tests/ReflectionTesting.cs
```diff
@@ -0,0 +1,465 @@
...
+    {
+        string Func(int a, string b);
+    }
+
+    public class ReflectionTesting
+    {
+        const bool _keepFiles = true; // keep files after testing for inspection
+        TempFileCollection _tfc;
+
+        [Fact]
```
AaronRobinsonMSFT: I mentioned this previously. This is not the correct way to setup a test class asset. See https://xunit.net/docs/shared-context

src/Experiment/src/CustomAttributeBuilder.cs
MosheWolberg: This is straight from Reflection.Emit.

src/Experiment/src/CustomAttributeBuilder.cs
MosheWolberg: This is straight from Reflection.Emit.

MosheWolberg:
I've addressed these issues in my code. Since mostly small, added them into new work I've been doing, instead of writing them twice, to this branch and new branch (A lot of the changes in files overlap between branches). Will push them as part them as part next feature PR. 
Can we merge?

joperezr:
I think if possible it would be better to add the fixes here, that way there is no un-relevant code changes in your next PR. I'm happy to work with you offline to get these changes here.

MosheWolberg:
Merge?

joperezr (approved):

Issue #1912, opened 2022-06-20
[Hot Cold Splitting] Avoid relying on class code value 


cshung:
Right now, the code relies on class code value to ensure cold code blocks get generated after hot code, we should not rely on that. See [this comment](https://github.com/dotnet/runtimelab/pull/1900#discussion_r901216913) for more details.

PR #1934, opened 2022-07-13
Emit cold nodes last without relying on ClassCode

amanasifkhalid:
This PR addresses #1912 by removing `MethodColdCodeNode`'s dependency on its ClassCode to be emitted after all hot nodes in the PE execution section. When emitting `ObjectNodes`, `MethodColdCodeNodes` are stashed in a list, and all other `ObjectNodes` are emitted first. Finally, all `MethodColdCodeNodes` in the list are emitted to the end of the execution section. This implementation removes the need for another execution section.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.ReadyToRun/CodeGen/ReadyToRunObjectWriter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.ReadyToRun/CodeGen/ReadyToRunObjectWriter.cs
@@ -239,11 +239,21 @@ namespace ILCompiler.DependencyAnalysis
                 ISymbolDefinitionNode lastImportThunk = null;
                 ObjectNode lastWrittenObjectNode = null;
 
+                // Save cold method nodes here, and emit them to execution section last.
+                List<ObjectNode> methodColdCodeNodes = new List<ObjectNode>();
+
                 int nodeIndex = -1;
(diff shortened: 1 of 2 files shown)


src/coreclr/tools/aot/ILCompiler.ReadyToRun/CodeGen/ReadyToRunObjectWriter.cs
```diff
@@ -307,6 +317,37 @@ public void EmitPortableExecutable()
...
+
+                    if (node == null)
+                    {
+                        continue;
+                    }
+
+                    ObjectData nodeContents = node.GetData(_nodeFactory);
+                    string name = null;
+
+                    if (_mapFileBuilder != null)
```
trylek: Could we move this to a helper method to avoid duplicating the generic hackery?

trylek (approved):
LGTM, thank you!

src/coreclr/tools/aot/ILCompiler.ReadyToRun/CodeGen/ReadyToRunObjectWriter.cs
amanasifkhalid: Sure thing, just pushed a revision. Thank you for the review!

PR #1936, opened 2022-07-14
Generate Scratch only if there is cold code

### LINKED ISSUE #1906: [Hot Cold Splitting] Conditional hot-cold mapping generation.
Right now, the hot-cold mapping is generated unconditionally. We should not generate it if there are no cold blocks.

amanasifkhalid:
Fixes #1906 by generating the Scratch table only if a call to `CorInfoImpl.CompileMethod()` results in cold code being generated. This work does not require any changes in the VM, as iterations over the Scratch table already check the table's size before accessing it, thus avoiding any null pointer dereferences.

Below is an excerpt from the R2R dump for `CoreLab`, showcasing the Scratch table exists:
```
Type:  ManifestAssemblyMvids (118)
RelativeVirtualAddress: 0x0000180C
Size: 0 bytes

_______________________________________________

Type:  Scratch (119)
RelativeVirtualAddress: 0x00001853
Size: 8 bytes
1,0


======================== R2R Methods ========================
```

After editing `CoreLab\Program.cs` so that the sole function is empty (and thus does not have any cold code), the R2R dump excerpt looks like this:

```
Type:  ManifestAssemblyMvids (118)
RelativeVirtualAddress: 0x000017C4
Size: 0 bytes


======================== R2R Methods ========================
```

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/DependencyAnalysis/ReadyToRun/RuntimeFunctionsTableNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/DependencyAnalysis/ReadyToRun/RuntimeFunctionsTableNode.cs
@@ -59,8 +59,6 @@ namespace ILCompiler.DependencyAnalysis.ReadyToRun
 
         public override ObjectData GetData(NodeFactory factory, bool relocsOnly = false)
         {
-            // This node does not trigger generation of other nodes.
-            // TODO: Make this generate the generation of the Scratch node
             if (relocsOnly)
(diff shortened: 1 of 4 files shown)


amanasifkhalid:
@cshung PTAL, thank you!

amanasifkhalid:
I realized `NodeFactory.Scratch` is a public field, so I can remove the redundant `_generatedScratch` boolean and replace it with a null check.

amanasifkhalid:
Removed an obsolete comment from `RuntimeFunctionsTableNode.cs`. CI was previously clean, so we shouldn't need to wait for another run.

cshung (approved):

PR #1937, opened 2022-07-19
Disable splitting for switch tables when inlining

amanasifkhalid:
Due to an implementation limitation, the JIT disables hot/cold splitting for functions that may contain switch tables. However, this limitation could be bypassed if we inline a function containing switch tables: `opts.compProcedureSplitting` would be disabled for the inlining compiler, but not for the parent compiler, which could result in the switch table being incorrectly split. This PR prevents this from happening by disabling `opts.compProcedureSplitting` for the parent compiler.

### DIFF
--- a/src/coreclr/jit/fgbasic.cpp
+++ b/src/coreclr/jit/fgbasic.cpp
@@ -2656,7 +2656,7 @@ unsigned Compiler::fgMakeBasicBlocks(const BYTE* codeAddr, IL_OFFSET codeSize, F
                 jmpKind     = BBJ_SWITCH;
                 fgHasSwitch = true;
 
-                if (opts.compProcedureSplitting)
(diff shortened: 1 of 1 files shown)


amanasifkhalid:
I added the if statement back in to conditionally output the JIT_DUMP message.

cshung (approved):
LGTM

AndyAyersMS:
This should wait until we know for sure that the inline will succeed. There's a common place where we propagate this sort of info from inlinee to root:

https://github.com/dotnet/runtime/blob/16f13eb0707c63e8e5ba8f34373308a43914388a/src/coreclr/jit/fginline.cpp#L1361-L1433

amanasifkhalid:
@AndyAyersMS thanks for catching this. I've opened a fix here: #1944.

PR #1939, opened 2022-07-19
Fix duplicate Scratch generation

amanasifkhalid:
This PR fixes a bug where the Scratch table could be generated multiple times when the dependency graph computes its dependencies in parallel. `NodeFactory.GenerateScratch()` has been updated to prevent duplicate generation.


cshung (approved):
LGTM

PR #1942, opened 2022-07-20
Fix GetUnwindInfoHelper so that we will not get a chained unwind info

amanasifkhalid:
When unwinding from cold main method code, we risk breaking invariants the security stack walk cache assumes -- the cache does not expect chained unwind info, and thus can return incorrect results later if not handled. We handle this by simply invalidating the cache; assuming cold code is rarely executed, this should not happen frequently.

### DIFF
--- a/src/coreclr/vm/jitinterface.cpp
+++ b/src/coreclr/vm/jitinterface.cpp
@@ -78,8 +78,6 @@
 #define JIT_TO_EE_TRANSITION_LEAF()
 #define EE_TO_JIT_TRANSITION_LEAF()
 
-
-
 #ifdef DACCESS_COMPILE
 
 // The real definitions are in jithelpers.cpp. However, those files are not included in the DAC build.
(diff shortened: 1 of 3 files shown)


src/coreclr/vm/jitinterface.cpp
```diff
@@ -14572,10 +14573,13 @@ void EECodeInfo::GetOffsetsFromUnwindInfo(ULONG* pRSPOffset, ULONG* pRBPOffset)
...
     UNWIND_INFO * pInfo = GetUnwindInfoHelper(unwindInfo);
+
+    // Chained unwind info use violates assumptions of the security stackwalk cache.
+    // To avoid returning incorrect cache entries, invalidate the cache.
+    // Since chained unwind info is used only for cold main body code, this should not happen frequently.
     if (pInfo->Flags & UNW_FLAG_CHAININFO)
     {
-        _ASSERTE(!"GetRbpOffset() - chained unwind info used, violating assumptions of the security stackwalk cache");
-        DebugBreak();
+        StackwalkCache::Invalidate(pLoaderAllocator);
```
jkotas: This method is called before the entry is inserted into the cache, so there is nothing to invalidate at this point. 

Maybe a better approach is to ensure that the entry won't get inserted into the stackwalk cache when we hit this case. Ie suppress `m_crawl.stackWalkCache.Insert(&cacheEntry);` in stackwalk.cpp for this case.

amanasifkhalid: Thanks for catching that -- I'll push a revision with this approach.

amanasifkhalid:
I've updated the implementation to not insert chained unwind info into `stackWalkCache`. Avoiding invalidating the entire cache should benefit performance, too.

src/coreclr/vm/stackwalk.cpp
```diff
@@ -2455,6 +2455,13 @@ StackWalkAction StackFrameIterator::NextRaw(void)
             if(isCollectableMethod)
                 fInsertCacheEntry = FALSE;
 
+            // Security stackwalk cache assumes it will not contain chained unwind info, so do not insert it.
+            UNWIND_INFO * pUnwindInfo = m_crawl.GetCodeInfo()->GetUnwindInfo();
```
jkotas: This won't work on non-x64. `UNWIND_INFO` and `UNW_FLAG_CHAININFO` are x64 specific structures. It would be nice to keep all special casing for this in EECodeInfo and avoid platform specific code here.

cshung:
@jkotas Can you please take a look?

jkotas (approved):

Issue #1903, opened 2022-06-18
[Hot Cold Splitting] Redesign the hot cold mapping data structure to allow fast lookup


cshung:
Right now, the data structure is simply a list of (cold runtime function index, hot runtime function index), this is not efficient for lookup. We need to redesign it so that it can be lookup fast.

amanasifkhalid:
# Design Proposal

## Current Implementation
@cshung's initial prototype for hot/cold splitting in Crossgen2 introduces a new data structure mapping indices of cold runtime functions to indices of their hot counterparts in the runtime function table. This mapping (currently referred to as `Scratch` in the codebase) is implemented as a list of integer indices: Mapped cold/hot integer indices are inserted into `Scratch` in pairs. This creates some important invariants:

* Function indices at even indices in `Scratch` point to cold functions, while function indices at odd indices point to hot functions.
    * In the prototype codebase, you will see checks like `if (i % 2 == 0)` to check if `Scratch[i]` is the index of a cold function, etc.
* The index of a cold function is immediately succeeded by the index of its hot counterpart in `Scratch`.
    * This facilitates obtaining the hot/cold counterpart of a cold/hot function, given its index in `Scratch`

`Scratch` is written to file as a linear collection of unsigned integer indices, each one 4 bytes long.

## Problems
* When writing to the R2R format, space is critical. If `N` functions are split into hot/cold sections, `Scratch` will contain `2N` indices. If each index is 4 bytes long, we will write `8N` bytes to file, plus potentially any extra data needed by `Scratch`. Using a more succinct data structure, we can approach the theoretical lower limit for the amount of memory required to represent this hot/cold function mapping.
* Furthermore, searching `Scratch` at runtime requires a linear search. Ideally, we want a dictionary-like data structure that achieves constant-time lookup; if this takes too much space, a data structure capable of binary search would still be an improvement.

## Proposal
We propose replacing `Scratch` with a succinct [rank/select dictionary](https://en.wikipedia.org/wiki/Succinct_data_structure#Succinct_dictionaries). Using a succinct dictionary would enable constant-time lookup while significantly reducing the amount of data written to file. Consider the following implementation:

At compile-time,
[...]

PR #1943, opened 2022-07-21
Convert Scratch lookup to binary search

amanasifkhalid:
Preliminary work for #1903: Before making any design changes, we should try using binary search over the Scratch table, and profile the performance gain. I likely won't have time to fully pursue this task, but we can at least get rid of the linear search for now.

### DIFF
--- a/src/coreclr/vm/codeman.cpp
+++ b/src/coreclr/vm/codeman.cpp
@@ -5756,6 +5756,86 @@ int NativeUnwindInfoLookupTable::LookupUnwindInfoForMethod(DWORD RelativePc,
     return -1;
 }
 
+int HotColdMappingLookupTable::LookupMappingForMethod(ReadyToRunInfo* pInfo, ULONG MethodIndex)
+{
+    CONTRACTL {
+        NOTHROW;
+        GC_NOTRIGGER;
+        SUPPORTS_DAC;
+    } CONTRACTL_END;
+
+    if (pInfo->m_nScratch == 0)
+    {
+        return -1;
+    }
+
+    // Casting the lookup table's size to an int is safe:
+    // We index the RUNTIME_FUNCTION table with ints, and the lookup table
+    // contains a subset of the indices in the RUNTIME_FUNCTION table.
+    // Thus, |lookup table| <= |RUNTIME_FUNCTION table|.
+    _ASSERTE(pInfo->m_nScratch <= pInfo->m_nRuntimeFunctions);
+    const int nLookupTable = (int)(pInfo->m_nScratch);
+
+    // The lookup table contains pairs of hot/cold indices, and thus should have an even size.
+    _ASSERTE((nLookupTable % 2) == 0);
+    int high = ((nLookupTable - 1) / 2);
+    int low  = 0;
+
+    const int indexCorrection = (int)(MethodIndex < pInfo->m_pScratch[0]);
+
+    // Binary search the lookup table.
+    // Use linear search once we get down to a small number of elements
+    // to avoid binary search overhead.
+    while (high - low > 10)
+    {
+        const int middle = low + (high - low) / 2;
+        const int index = (middle * 2) + indexCorrection;
+
+        if (MethodIndex < pInfo->m_pScratch[index])
+        {
+            high = middle - 1;
+        }
+        else
+        {
+            low = middle;
+        }
+    }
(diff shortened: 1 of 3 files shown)


src/coreclr/vm/codeman.cpp
```diff
@@ -5756,6 +5756,84 @@ int NativeUnwindInfoLookupTable::LookupUnwindInfoForMethod(DWORD RelativePc,
...
+
+    if (nLookupTable == 0)
+    {
+        return -1;
+    }
+
+    // Casting the lookup table's size to an int is safe:
+    // We index the RUNTIME_FUNCTION table with ints, and the lookup table
+    // contains a subset of the indices in the RUNTIME_FUNCTION table.
+    // Thus, |lookup table| <= |RUNTIME_FUNCTION table|.
```
cshung, line 5775: Why not just assert that fact instead?

src/coreclr/vm/codeman.cpp
```diff
@@ -5756,6 +5756,84 @@ int NativeUnwindInfoLookupTable::LookupUnwindInfoForMethod(DWORD RelativePc,
...
+
+    int high = ((nLookupTable - 1) / 2);
```
cshung, line 5781: `nLookupTable` is always even, we can assert that fact. Should we really subtract 1 here?

src/coreclr/vm/codeman.h
```diff
@@ -1499,6 +1499,14 @@ class NativeUnwindInfoLookupTable
                                          int EndIndex);
 };
 
+class HotColdMappingLookupTable
+{
+public:
+    static int LookupMappingForMethod(PTR_ULONG pLookupTable,
```
cshung: Can we have a comment saying what this function is supposed to return?

cshung (approved):

src/coreclr/vm/codeman.cpp
amanasifkhalid: By copying the binary search implementation in `NativeUnwindInfoLookupTable::LookupUnwindInfoForMethod`, I made `low` and `high` inclusive bounds, so I think subtracting 1 is required here. Without this, it is possible for us to index out of bounds in the `for` loop below (since we iterate while `i <= high`).

src/coreclr/vm/codeman.cpp
amanasifkhalid: I can add an assert for this -- we'll just pass `pInfo` instead of `m_nScratch` and `m_pScratch` separately so we can access the size of the runtime function table.

src/coreclr/vm/codeman.h
amanasifkhalid: Sure thing.

amanasifkhalid:
@cshung I pushed changes addressing your feedback. For asserting the lookup table's size is <= the runtime function table's size, I changed the method's parameter list to just pass `pInfo`, but this required making `HotColdMappingLookupTable` a friend of `ReadyToRunInfo` -- not sure if this is ideal or not...

src/coreclr/vm/codeman.cpp
cshung: Got it, would be more clear to write `nLookupTable/2 - 1` so that I know you want an inclusive upper bound, but that's okay.

src/coreclr/vm/codeman.cpp
```diff
@@ -5756,6 +5756,84 @@ int NativeUnwindInfoLookupTable::LookupUnwindInfoForMethod(DWORD RelativePc,
...
+    if (pInfo->m_nScratch == 0)
+    {
+        return -1;
+    }
+
+    // Casting the lookup table's size to an int is safe:
+    // We index the RUNTIME_FUNCTION table with ints, and the lookup table
+    // contains a subset of the indices in the RUNTIME_FUNCTION table.
+    // Thus, |lookup table| <= |RUNTIME_FUNCTION table|.
+    _ASSERTE(pInfo->m_nScratch <= pInfo->m_nRuntimeFunctions);
```
cshung, line 5776: While we are at it, we might as well `_ASSERTE(pInfo->m_nScratch % 2 == 0)`

amanasifkhalid: Gotcha, just pushed a revision.

PR #1944, opened 2022-07-26
Move inline switch table check to fginline.cpp

amanasifkhalid:
Follow-up to #1937 : According to @AndyAyersMS, we should disable hot/cold splitting in the root compiler if its inlinee compiler finds a switch table only if the inline succeeds. Thus, the logic to disable `opts.compProcedureSplitting` in the root compiler has been moved to a common place in `fginline.cpp` for propagating compiler arguments. This change does not reproduce the related regression exposed by #1937 when stress-splitting `System.Private.CoreLib`.


amanasifkhalid:
@cshung @AndyAyersMS PTAL, thank you!

AndyAyersMS (approved):
Looks good. Thanks for moving this bit of logic.

Issue #1950, opened 2022-08-03
[Hot Cold Splitting] Test failures


cshung:
# Repro steps:

1. Checkout my `private/hot-cold-devenv` branch [here](https://github.com/cshung/runtimelab/tree/private/hot-cold-devenv)
2. Run these commands in a Command Prompt (not Powershell)

```
bc.cmd
blo.cmd
bctg.cmd
bs.cmd
cr.cmd
cgt_on.cmd
rs.cmd
```

> Note, these scripts create environment variables, make sure you close the prompt when you are done with it.

After building once, you can run `rs.cmd` again to reproduce the issue again without rebuilding, you can also run it with `ds.cmd` to debug.

Feel free to change `bs.cmd`, `ds.cmd` and `rs.cmd` so that you can run another test.

PR #1951, opened 2022-08-04
Fixed indexing error

EugenioPena:
Fixed an indexing error that would cause the index to be -1, which would cause some test failures when reproducing the steps stated on issue #1950. Did this by validating that the number of functions in the scratch map wasn't 0 before calling `HotColdMappingLookupTable::LookupMappingForMethod`.


cshung (approved):
It looks like ​`LookupMappingForMethod`​ expects a non-empty scratch table.
Can you add ​`assert(pInfo->n_Scratch != 0)​` in the implementation of `LookupMappingForMethod​`

PR #1953, opened 2022-08-05
Fix Standalone template's official build and publishing

joperezr:
Fixing Standalone template so that it is able to successfully have an official build and publish packages.


jkotas (approved):
I assume that this is just arcade update without any real changes. Is that correct?

Also, the PR title should be fixed up.

joperezr:
Oops, yeah I don't know where that title came from.

> I assume that this is just arcade update without any real changes. Is that correct?

Correct. The only manual changes where the nuget.config removal of nuget.org and the runtimelab.yml pool definitions, all of the rest is the result of running `darc update-dependencies -c ".NET Eng - Latest"` which automatically migrates the repo to the latest version of Arcade.

Issue #1859, opened 2022-03-06
[NativeAOT-LLVM] Optimisation on compiling problem


yowl:
Looking at what is wrong with Release config builds and why Optimisation was removed in #1801, this is the first problem.  
We have 
```c++
bool Llvm::isLlvmFrameLocal(LclVarDsc* varDsc)
{
    assert(canStoreLocalOnLlvmStack(varDsc) && (_compiler->fgSsaPassesCompleted >= 1));
    return !varDsc->lvInSsa;
}
```
But when compiling with optimisation, you can get methods with LIR:
```
------------ BB01 [000..008) (return), preds={} succs={}
N001 (  0,  0) [000002] ------------                 RETURN    void   $c0
```
In this case the shadow stack arg is not tracked as there are no references, `lvInSsa` is then false, and `isLlvmFrameLocal` returns true.  Should this method include `varDsc->lvTracked` or `lvRefCnt() != 0` or `lclNum != _shadowStackLclNum` ?

cc @SingleAccretion

SingleAccretion:
Depends on what the actual problem is.

If it's that we're creating unused `alloca`s (which I would guess is what is happening), the place that is creating them should be checking `lvRefCnt`, yes (it is what ordinary codegen does).

yowl:
In `createAllocasForLocalsWithAddrOp` it then goes on to hit the assert :
```c++
            if (varDsc->lvIsParam)
            {
                LlvmArgInfo argInfo = getLlvmArgInfoForArgIx(lclNum);
                assert(argInfo.IsLlvmArg())
```

SingleAccretion:
It does not seem right that `getLlvmArgInfoForArgIx` effectively fails if the shadow stack local is not referenced.

yowl:
The next assert after this one is
https://github.com/dotnet/runtimelab/blob/1960cfa58f3aa2ab46355947c19b7cfd33dbe60b/src/coreclr/jit/importer.cpp#L17454
Is this related to tail calls?  Tail calls are at phase 3 for WebAssembly, so would be good not to do anything that makes enabling them more difficult.

SingleAccretion:
That looks like some importer code being all too aware of ABI details to me, yikes...

~~LLVM should be taking the "no return buffer path" there. I think you can add `TARGET_WASM` to the `#elif defined(TARGET_X86)` (in general our "virtual" (before lower) ABI is very similar to x86).~~

^ Above not true, let me look again.

[...]

PR #1966, opened 2022-08-25
NativeAOT-LLVM: Allow `RyuJitCompilation` to turn off tail calls

yowl:
As the first part of progressing #1859 this PR allows tail calls to be turned off based on the method signature.

For `object.ToString`:

IL
```
// object.ToString()
.method instance string ToString() cil managed
{
  // Code size: 12
  .maxstack 8

  IL_0000:  ldarg.0
  IL_0001:  call        instance class System.Type System.Object::GetType()
  IL_0006:  callvirt    instance string System.Object::ToString()
  IL_000B:  ret
}
```
The `callvirt/ret` are identified as a tail call
```
    [ 1]   6 (0x006) callvirt 06000268
 (Implicit Tail call: prefixFlags |= PREFIX_TAILCALL_IMPLICIT)
In Compiler::impImportCall: opcode is callvirt, kind=4, callRetType is ref, structSize is 0
```
It would be nice to allow tail calls in Wasm, LLVM has tail convention, but Wasm can't do it when the type needs to go in a return slot on the shadow stack, so to allow tail calls for simple types that can go on the LLVM stack, but turn it off for return types that need a shadow stack slot, this PR adds a virtual call to `RyuJitCompilation` that LLVM can override and turn off the tail call.  This wont prevent the assert at https://github.com/dotnet/runtimelab/blob/706c9c1ad1f4d3d150d08cd02d4ac507d02302f9/src/coreclr/jit/llvm.cpp#L2753 for simple types, but should clear the way for adding LLVM tail convention for simple types.


yowl:
cc @SingleAccretion

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -159,9 +159,9 @@ private unsafe void CompileSingleMethod(CorInfoImpl corInfo, LLVMMethodCodeNode
                     ILImporter.GenerateRuntimeExportThunk(this, method, externFunc);
 
                     ryuJitMethodCount++;
+                    }
+                    else ILImporter.CompileMethod(this, methodCodeNodeNeedingCode);
```
jkotas: Revert this indentation change?

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -284,5 +284,10 @@ public override MethodDesc GetCompilerHelpersMethodDesc(string className, string
             MetadataType helperType = TypeSystemContext.SystemModule.GetKnownType("Internal.Runtime.CompilerHelpers", className);
             return helperType.GetKnownMethod(helperMethodName, null);
         }
+
+        public override bool CanTailCall(MethodDesc caller)
+        {
+            return !NeedsReturnStackSlot(caller.Signature);
```
jkotas: The codegen arch specific tailcall limitations are handled in the JIT importer and morph. Can this check be moved there?

yowl: Thanks,
I've moved some stuff to morph, that looked straightforward enough, I made a few methods in Llvm static and had the usual `min`/`std::min` problem in morph.  What is it that I should be changing in importer?  I had a look at `impTailCallRetTypeCompatible` and probably `allowWidening` should always be false for LLVM.  I'm not sure what else will fail with the compatible check at the moment.  As there are no unsigned types, probably the size check is going to work out the same as an actual LLVM type check.

SingleAccretion: I think morph is a reasonable place for this check. It does similar ABI-dependent checks for other targets as well.

src/coreclr/jit/morph.cpp
```diff
@@ -17,6 +17,13 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
 #include "allocacheck.h" // for alloca
 
+#undef min
+#undef max
+
+#if TARGET_WASM
+#include "llvm.h"
+#endif // TARGET_WASM
```
SingleAccretion: Would redefining `min`/`max` after the `#include` work here? Seems better to avoid these small divergences with upstream.

src/coreclr/jit/morph.cpp
```diff
@@ -3634,7 +3655,7 @@ void Compiler::fgInitArgInfo(GenTreeCall* call)
 #ifdef WINDOWS_AMD64_ABI
                             // Whenever we pass an integer register argument
                             // we skip the corresponding floating point register argument
-                            intArgRegNum = min(intArgRegNum + size, MAX_REG_ARG);
+                            intArgRegNum = min(intArgRegNum + size, (unsigned int)MAX_REG_ARG);
```
SingleAccretion: Are the casts still needed?

(If they are, it suggests the `min/max` templates aren't quite what the original code used)

src/coreclr/jit/compiler.cpp
```diff
@@ -26,12 +26,26 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
 extern ICorJitHost* g_jitHost;
 
+#if TARGET_WASM
```
```suggestion
#ifdef TARGET_WASM
```

src/coreclr/jit/morph.cpp
yowl: yes, that's annoying.  Without them 
```
 E:\GitHub\runtimelab\src\coreclr\jit\morph.cpp(3658): error C2672: 'min': no matching overloaded function found
  E:\GitHub\runtimelab\src\coreclr\jit\morph.cpp(27): note: could be 'const T &min(const T &,const T &)'
  E:\GitHub\runtimelab\src\coreclr\jit\morph.cpp(3658): note: 'const T &min(const T &,const T &)': template parameter 'T' is ambiguous
  E:\GitHub\runtimelab\src\coreclr\jit\morph.cpp(27): note: see declaration of 'min'
  E:\GitHub\runtimelab\src\coreclr\jit\morph.cpp(3658): note: could be 'int'
  E:\GitHub\runtimelab\src\coreclr\jit\morph.cpp(3658): note: or       'unsigned int'
  E:\GitHub\runtimelab\src\coreclr\jit\morph.cpp(3658): note: 'const T &min(const T &,const T &)': could not deduce template argument for 'const T &' from 'int'
  E:\GitHub\runtimelab\src\coreclr\jit\morph.cpp(27): note: see declaration of 'min'
```
I copied the templates from 
https://github.com/dotnet/runtimelab/blob/706c9c1ad1f4d3d150d08cd02d4ac507d02302f9/src/coreclr/inc/clr_std/utility#L53-L68

I'll have another look around, maybe there's some others.

yowl: Ah, but `-DUSE_STL` is set so they might be from 
```
#include <utility>
```

yowl: According to VS they are from `C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\shared\minwindef.h` so I've redefined them from those and removed the casts.

SingleAccretion (approved):

jkotas (approved):
LGTM. Thank you!

yowl:
Thanks!

PR #1967, opened 2022-08-26
Fix some mistakes in `docs/using-nativeaot/interop.md`

ero-qt:
Simply some typo fixes, as well as trying to keep the language in line with existing documentation.


jkotas:
The native AOT project was moved to dotnet/runtime repo. It is no longer maintained here.

Could you please submit this PR against https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/docs/interop.md instead?

PR #1968, opened 2022-08-27
NativeAOT-LLVM: do not mark the shadow stack as a local on the llvm stack if not referenced

yowl:
Continuing #1859 this PR addresses the first issue hit in that PR: https://github.com/dotnet/runtimelab/issues/1859#issue-1160716397

For the IL: 
```
// [S.P.CoreLib]System.Reflection.Module+<>c..ctor()
.method instance void .ctor() cil managed
{
  // Code size: 7
  .maxstack 8

  IL_0000:  ldarg.0
  IL_0001:  call        instance void System.Object::.ctor()
  IL_0006:  ret
}
```

We get the IR
```
------------ BB01 [000..007) (return), preds={} succs={}
N001 (  0,  0) [000002] ------------                 RETURN    void   $c0
```

The call is inlined if I read the log properly, and `$c0` is `{MemOpaque:NotInLoop}` <- I don't know what that means, is it the reason for inlining?

Anyway, that fails currently for the reason in the original disucssion: it tries to create an LLVM `alloca` for the shadow stack which is not referenced and `getLlvmArgInfoForArgIx` hits an assert.

With this change the final LLVM is
```
define void @S_P_CoreLib_System_Reflection_Module___c___ctor(i8* %0) {
Prolog:
  br label %1

1:                                                ; preds = %Prolog
  ret void
}
```

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -791,6 +791,13 @@ LlvmArgInfo Llvm::getLlvmArgInfoForArgIx(unsigned int lclNum)
                                                                                        the shadow stack */
     };
 
+    if (lclNum == _shadowStackLclNum)
+    {
+        llvmArgInfo.m_argIx             = 0;
+        llvmArgInfo.m_shadowStackOffset = 0;
+        return llvmArgInfo;
+    }
+
     if (_compiler->info.compRetBuffArg != BAD_VAR_NUM)
     {
         if (lclNum == 0)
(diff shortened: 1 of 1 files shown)


yowl:
As an aside, how does this work if a method gets it's implementation inlined to basically nothing like this, does the compiler then go about checking callers of this method to inline those calls and so on?

SingleAccretion (approved):
> I don't know what that means, is it the reason for inlining?

No, it's a value numbering annotation. You can read about our value numbering [here](https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/jit/Jit%20Architecture%20Plan%202009.md#332--value-numbering) and [here](https://github.com/dotnet/runtime/blob/main/src/coreclr/jit/valuenum.h).

> As an aside, how does this work if a method gets it's implementation inlined to basically nothing like this, does the compiler then go about checking callers of this method to inline those calls and so on?

No, we don't have inlining heuristics like that (yet), so it is treated just like any other method.

PR #1971, opened 2022-08-29
NativeAOT-LLVM: Add option to disable the RyuJIT backend

yowl:
This PR adds a codegenopt:

```
--codegenopt:DisableRyuJit=1
```
This will turn off the RyuJIT backend and force all compilation through the old IL->LLVM backend.  This can be helpful if a user encounters a problem with the evolving RyuJIT backend that is not present in the IL backend.

cc @SingleAccretion

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
@@ -23,6 +23,7 @@ namespace ILCompiler
     {
         private readonly ConditionalWeakTable<Thread, CorInfoImpl> _corinfos = new ConditionalWeakTable<Thread, CorInfoImpl>();
         private string _outputFile;
+        bool _disableRyuJit;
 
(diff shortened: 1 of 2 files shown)


src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -23,6 +23,7 @@ public sealed class LLVMCodegenCompilation : RyuJitCompilation
     {
         private readonly ConditionalWeakTable<Thread, CorInfoImpl> _corinfos = new ConditionalWeakTable<Thread, CorInfoImpl>();
         private string _outputFile;
+        bool _disableRyuJit;
```
SingleAccretion:
```suggestion
        private readonly bool _disableRyuJit;
```
(Should also be reordered with `_outputFile` as it is `readonly`)

SingleAccretion (approved):
It's unconventional to have `--codegenopt`s that are meant to influence the compiler driver, as opposed to the code generator, but I don't think it matters much for this temporary switch.

yowl:
Thanks

Issue #1970, opened 2022-08-29
NativeAOT-LLVM: packages not building or publishing


yowl:
Hi,

The last couple of merges have not produced new artifacts at

https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-experimental

![image](https://user-images.githubusercontent.com/2720041/187204634-d99786a4-2f47-436e-9eba-9e5d185af11a.png)

Can you see what the problem is?

jkotas:
It fails with:
```
##[error]The Windows 2016 environment is deprecated, consider switching to windows-2019 or windows-2022 (windows-latest). For more details, see https://github.com/actions/virtual-environments/issues/5403
##[error]The remote provider was unable to process the request.
Pool: Azure Pipelines
Image: vs2017-win2016
```
It may need https://github.com/dotnet/runtime/commit/e233b7b5b00d9b80f155cd484e246dfc24fdf801 to pick up current machine pools.

yowl:
Thanks, has published successfully:
![image](https://user-images.githubusercontent.com/2720041/187435449-17b155c6-9436-4da8-929c-7e40dc142934.png)

PR #1972, opened 2022-08-29
NativeAOT-LLVM: upgrade DevOps agent as  win2016 is deprecated

yowl:
This PR attempts to address #1970 taking the commit https://github.com/dotnet/runtime/commit/e233b7b5b00d9b80f155cd484e246dfc24fdf801

Thanks @jkotas


yowl:
Did the agent just die?
```
We stopped hearing from agent NetCore1ESPool-Public 38
```
Will close and reopen in case its not related to this change.

yowl:
transient I guess.

PR #1975, opened 2022-08-31
NativeAOT-LLVM: enable the fallback TypeSystemThrowingILEmitter method

yowl:
This PR enables the method that is emitted in case of a `TypeSystemException` in compilation.  It registers the callbacks just once, earlier, in case the method that fails is the first one through `corInfo` or if the RyuJIT backend is disabled.  The RyuJIT backend even if disabled is used for the `TypeSystemThrowingILEmitter`.  Also sets up the debug info to prevent a null ref.

cc @SingleAccretion


yowl:
fatal error LNK1318: Unexpected PDB error
Will retry.

yowl:
Wip, as this is not enough for all cases.

yowl:
No longer needed

PR #1977, opened 2022-09-01
NativeAOT-LLVM: Add support for GT_CNS_LNG

yowl:
This PR add support for `GT_CNS_LNG`.  I've left in the the possibilty of `node->gtType == TYP_REF` in case this could happen with Wasm64?

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1384,27 +1384,7 @@ void Llvm::buildCnsInt(GenTree* node)
 
 void Llvm::buildCnsLng(GenTree* node)
 {
-    if (node->gtType == TYP_LONG)
-    {
-        mapGenTreeToValue(node, _builder.getInt64(node->AsLngCon()->LngValue()));
-        return;
-    }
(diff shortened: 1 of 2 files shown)


jkotas:
The CI is down in the whole dotnet org for configuration changes.

src/coreclr/jit/llvm.cpp
```diff
@@ -1382,6 +1382,31 @@ void Llvm::buildCnsInt(GenTree* node)
...
+        if (longCon != 0)
+        {
+            failFunctionCompilation();
+        }
+
+        mapGenTreeToValue(node, _builder.CreateIntToPtr(_builder.getInt64(longCon),
+                                                        Type::getInt8PtrTy(_llvmContext))); // TODO: wasm64
+        return;
+    }
+    failFunctionCompilation();
```
```suggestion
    mapGenTreeToValue(node, _builder.getInt64(node->AsLngCon()->LngValue()));
```
`CNS_LNG` is only used for `TYP_LONG` on 32 bit targets.

yowl:
Thanks both.

SingleAccretion (approved):

Issue #1987, opened 2022-09-10
NativeAOT-LLVM: RyuJIT backend is using RhpAssignRef for addresses outside the heap


yowl:
Not looked closely at this, but at least for `S_P_TypeLoader_Internal_NativeFormat_NativeHashtable_Enumerator___ctor` I noticed that the RyuJIT->LLVM was using `RhpAssignRef ` to store a value at an address outside the heap

yowl:
This method
https://github.com/dotnet/runtimelab/blob/8ac90a4815a96f5470e753960dcc616d007795c6/src/coreclr/tools/Common/Internal/NativeFormat/NativeFormatReader.cs#L494-L499

produces IR

```cs
------------ BB01 [???..???), preds={} succs={BB02}

------------ BB02 [000..017) (return), preds={BB01} succs={}
               [000022] ------------                 IL_OFFSET void   IL offset: 0x0
               [000023] ------------                 IL_OFFSET void   IL offset: 0x1
               [000030] ------------        t30 =    LCL_VAR   int    V04 tmp0         u:1
               [000029] ------------        t29 =    CNS_INT   int    0
                                                  /--*  t30    int
                                                  +--*  t29    int
               [000031] ------------        t31 = *  ADD       int
                                                  /--*  t31    int
N001 (  3,  2) [000002] ------------         t2 = *  IND       byref
N002 (  1,  1) [000017] ------------        t17 =    CNS_INT   int    8 field offset Fseq[_parser]
                                                  /--*  t2     byref
                                                  +--*  t17    int
N003 (  5,  4) [000018] ------------        t18 = *  ADD       byref
               [000033] ------------        t33 =    LCL_VAR   int    V04 tmp0         u:1
               [000032] ------------        t32 =    CNS_INT   int    4
                                                  /--*  t33    int
                                                  +--*  t32    int
               [000034] ------------        t34 = *  ADD       int
                                                  /--*  t34    int
N005 (  3,  2) [000003] ------------         t3 = *  OBJ       struct<Internal.NativeFormat.NativeParser, 8>
                                                  /--*  t18    byref
                                                  +--*  t3     struct
N004 ( 11,  8) [000006] -A-XG-------              *  STORE_OBJ struct<Internal.NativeFormat.NativeParser, 8> (copy)
```

The `NativeParser` struct is `STORE_OBJ` at the result of an `ADD` and in llvm.cpp at

[...]

Issue #1984, opened 2022-09-08
NativeAOT-LLVM: heap corruption


yowl:
Hi,

I wonder if anyone can look at this, and maybe give me some more ideas as I've been looking at it for quite a few hours without success, thanks.

If I turn on the `HeapVerifyLevel` setting in `gcconfig.h` to 7 (i.e. all 3 flags), then I can make the verify_heap function fail in node's v8 with this code inserted into `HelloWasm.cs`:

Fails at the third GC.Collect.  Previous to this code, there has been a around a 100 other GC Collects, manual, and automatic that pass the verify checks.
```
GC.Collect();
PrintLine("collect 4");

GC.Collect();
var arrayType = GetArrayType();
PrintLine("collect 4a");
GC.Collect();
```
where `GetArrayType` is 
```
private static Type GetArrayType()
{
    return typeof(object[]);
}
```

Fails with (I have a printf("garbage_collect") in `garbage_collect`)
```
garbage_collect
collect 4
garbage_collect
collect 4a
garbage_collect
E:\GitHub\runtimelab\artifacts\tests\coreclr\Browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:386
      throw ex;
      ^

RuntimeError: unreachable
    at GCToOSInterface::DebugBreak() (wasm-function[21832]:0xf33c8b)
    at WKS::gc_heap::verify_heap(int) (wasm-function[21191]:0xeada88)
    at WKS::gc_heap::garbage_collect(int) (wasm-function[21108]:0xe7ad2a)
    at WKS::GCHeap::GarbageCollectGeneration(unsigned int, gc_reason) (wasm-function[21105]:0xe79ec3)
    at WKS::GCHeap::GarbageCollectTry(int, int, int) (wasm-function[21487]:0xf117ea)
    at WKS::GCHeap::GarbageCollect(int, bool, int) (wasm-function[21486]:0xf11494)
    at RhpCollect (wasm-function[20100]:0xe0ee7b)
    at S_P_CoreLib_System_Runtime_InternalCalls__RhpCollect (wasm-function[10626]:0x689ecf)
    at S_P_CoreLib_System_Runtime_InternalCalls__RhCollect (wasm-function[7478]:0x4dfc6f)
    at RhCollect (wasm-function[7477]:0x4dfc06)
    at S_P_CoreLib_System_GC__Collect_0 (wasm-function[8161]:0x54580a)
[...]

PR #1988, opened 2022-09-10
NativeAOT-LLVM: Change all cases of RhpAssignRef to RhpCheckedAssignRef 

yowl:
This PR address one cause of memory corruption where the IL->LLVM backend is using `RhpAssignRef` for addresses that may be outside of the heap, e.g. on the shadow stack.  This at least partially addresses #1984 .    All three instances of `RhpAssignRef` are changed as it was observed that they are all culprits.

I didn't create a test as this is difficult to make consistently fail.

The RyuJIT->LLVM backend is also doing this, which is why I've not added the assert mentioned in the above issue.  The RyuJIT problem can be looked at separately, #1987


yowl:
cc @SingleAccretion

SingleAccretion (approved):

PR #1989, opened 2022-09-11
NativeAOT-LLVM: Use RhpCheckedAssignRef from RyuJIT backend when location of address is unknown

### LINKED ISSUE #1987: NativeAOT-LLVM: RyuJIT backend is using RhpAssignRef for addresses outside the heap

yowl:
This PR fixes a problem where the RyuJIT backend is using RhpAssignRef for addresses on the stack, notably when the local var has had an `ADD` operation before the `STORE_OBJ`.   `RhpCheckedAssignRef ` is used instead as it is not easily known if the address is on the heap.  This is not the most efficient solution maybe, so input welcome as always.

Added an assert in `RhpAssignRef` to catch this problem when it occurs as opposed to later during a GC collection.

cc @SingleAccretion

Fixes #1987 

WIP as I'll wait for #1988 to be merged

### DIFF
--- a/src/coreclr/nativeaot/Runtime/GCMemoryHelpers.inl
+++ b/src/coreclr/nativeaot/Runtime/GCMemoryHelpers.inl
@@ -156,6 +156,10 @@ static const uint32_t INVALIDGCVALUE = 0xcccccccd;
 
 FORCEINLINE void InlineWriteBarrier(void * dst, void * ref)
 {
+#ifdef TARGET_WASM
+    ASSERT(((uint8_t*)dst >= g_lowest_address) && ((uint8_t*)dst < g_highest_address))
+#endif // TARGET_WASM
+
     if (((uint8_t*)ref >= g_ephemeral_low) && ((uint8_t*)ref < g_ephemeral_high))
     {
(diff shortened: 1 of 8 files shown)


src/coreclr/nativeaot/Runtime/GCMemoryHelpers.inl
```diff
@@ -156,6 +156,10 @@ static const uint32_t INVALIDGCVALUE = 0xcccccccd;
 
 FORCEINLINE void InlineWriteBarrier(void * dst, void * ref)
 {
+#ifdef TARGET_WASM
```
jkotas: This does not need to be under wasm ifdef.

yowl: thanks

SingleAccretion (approved):

Issue #1990, opened 2022-09-12
Question: NativeAOT-LLVM: leave targets


yowl:
The IL backend has a bug that means the finally block is called twice if an exception occurs and is caught as in the following code:
```cs
private static void TryFinallyWithCatch()
{
    try
    {
        throw new Exception();
    }
    catch {}
    finally
    {
        finallyCalled++;
    }
}
```
The IL for this looks like:
```cs
.method void TryFinallyWithCatch() cil managed
{
  // Code size: 31
  .maxstack 2

  .try IL_0001 to IL_0008 catch object handler IL_0008 to IL_000D
  .try IL_0001 to IL_000F finally handler IL_000F to IL_001E
  IL_0000:  nop
  IL_0001:  nop
  IL_0002:  newobj      instance void [System.Private.CoreLib]System.Exception::.ctor()
  IL_0007:  throw
  IL_0008:  pop
  IL_0009:  nop
  IL_000A:  nop
  IL_000B:  leave.s     IL_000D
  IL_000D:  leave.s     IL_001E
  IL_000F:  nop
  IL_0010:  ldsfld      int32 Program::finallyCalled
  IL_0015:  ldc.i4.1
  IL_0016:  add
  IL_0017:  stsfld      int32 Program::finallyCalled
  IL_001C:  nop
  IL_001D:  endfinally
  IL_001E:  ret
}
```
The 2 leave targets 
```
  IL_000B:  leave.s     IL_000D
  IL_000D:  leave.s     IL_001E
```
Is one for the try and one for the catch?.  If so why are they both in the catch handler region, after all it is `IL_0008 to IL_000D`  ?

Thanks

yowl:
The EH second pass handler for the exception should call the finally handler, shouldn't it?

jkotas:
This is how the IL matches the C#. 
```
try
{
    try
    {
        // newobj      instance void [System.Private.CoreLib]System.Exception::.ctor()
        // throw
        throw new Exception();
    }
    catch
    {
        // leave.s     IL_000D
    }
    
    // IL_000D:
    // leave.s     IL_001E
}
finally
{
    // ldsfld      int32 Program::finallyCalled
    // ldc.i4.1
    // add
    // stsfld      int32 Program::finallyCalled
    finallyCalled++;
    
    // endfinally
}

// IL_001E
// ret
```

> The EH second pass handler for the exception should call the finally handler, 

Yes, that's correct.

yowl:
[...]

PR #1991, opened 2022-09-14
NativeAOT-LLVM: Fix order of catch and second pass + calling finally twice

yowl:
This PR fixes a couple of issues with the LLVM IL backend for exception handling.  The second pass handler is now called before the catch (oops), and the finally block when adjacent to a catch, is only now called once.

I expect there are other problems, but at least this fixes 2 big ones and allows 

https://github.com/dotnet/runtimelab/blob/f9cf5b0e52c4645c0a76ad3770d0c63c712c3a05/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/CompilerServices/ClassConstructorRunner.cs#L109-L115

to throw without dying in the `Release`

Long term wasm-exceptions may help depending on how close the semantics are to .net I guess.

cc @SingleAccretion

### DIFF
--- a/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
+++ b/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
@@ -92,7 +92,6 @@ namespace System.Runtime
                 }
                 else
                 {
-                    tryRegionIdx = MaxTryRegionIdx;
                     bool shouldInvokeHandler = InternalCalls.RhpCallFilterFunclet(exception, ehClause._filterAddress, shadowStack);
                     if (shouldInvokeHandler)
                     {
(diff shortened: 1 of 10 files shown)


yowl:
closes #1990

jkotas:
I think you will need to port everything in https://github.com/dotnet/runtime/pull/75294/ to fix the macOS build failures.

yowl:
> I think you will need to port everything in [dotnet/runtime#75294](https://github.com/dotnet/runtime/pull/75294) to fix the macOS build failures.

Thanks I was having a bit of a problem

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
```diff
@@ -92,7 +92,7 @@ public bool ContainsCodeOffset(uint idxTryLandingStart)
                 }
                 else
                 {
-                    tryRegionIdx = 0;
+                    tryRegionIdx = MaxTryRegionIdx;
```
jkotas: Is this needed? We have set `tryRegionIdx = MaxTryRegionIdx;` before the loop.

yowl: No, deleted.  Thanks

jkotas (approved):

PR #1993, opened 2022-09-15
Add green thread api skeleton

davidwrighton:
Provide a green thread api that just uses threads  (`Task.RunAsGreenThread`)
- This is probably not completely correct, but if users just call .Wait on the returned tasks, it'll probably be ok.

Also added `Thread.IsGreenThread` to check to see if a thread is green.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs
@@ -5382,6 +5382,7 @@ namespace System.Threading.Tasks
                 TaskCreationOptions.DenyChildAttach, InternalTaskOptions.None);
         }
 
+        [UnsupportedOSPlatform("browser")]
         public static Task RunAsGreenThread(Action action, CancellationToken cancellationToken)
         {
             if (cancellationToken.IsCancellationRequested)
@@ -5409,6 +5410,7 @@ namespace System.Threading.Tasks
             return tcs.Task;
         }
 
+        [UnsupportedOSPlatform("browser")]
         public static Task RunAsGreenThread(Action action)
         {
             var tcs = new TaskCompletionSource();
@@ -5428,6 +5430,7 @@ namespace System.Threading.Tasks
             return tcs.Task;
         }
 
+        [UnsupportedOSPlatform("browser")]
         public static Task<TResult> RunAsGreenThread<TResult>(Func<TResult> function)
         {
             var tcs = new TaskCompletionSource<TResult>();
@@ -5446,6 +5449,7 @@ namespace System.Threading.Tasks
             return tcs.Task;
         }
 
+        [UnsupportedOSPlatform("browser")]
         public static Task<TResult> RunAsGreenThread<TResult>(Func<TResult> function, CancellationToken cancellationToken)
         {
             if (cancellationToken.IsCancellationRequested)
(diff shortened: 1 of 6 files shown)


davidwrighton:
@stephentoub @davidfowl Could you guys give an opinion to the api surface, good enough for now?

src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs
```diff
@@ -5382,6 +5382,52 @@ public static Task Run(Action action, CancellationToken cancellationToken)
...
+
+        public static Task RunAsGreenThread(Action action)
+        {
+            Task greenThreadTask = new Task(action);
+            new Thread(()=>
+            {
+                Thread.t_IsGreenThread = true;
+                greenThreadTask.RunSynchronously();
+            }).Start();
+            return greenThreadTask;
```
stephentoub: Even though it's a placeholder, for now I'd suggest instead doing:
```suggestion
            var tcs = new TaskCompletionSource();
            new Thread(()=>
            {
                Thread.t_IsGreenThread = true;
                try
                {
                    action();
                    tcs.SetResult();
                }
                catch (Exception e)
                {
                    tcs.SetException(e);
                }
            }) { IsBackground = true }.Start();
            return tcs.Task;
```
and similarly in the other overloads.  Using Task's ctor + Start isn't a great pattern and is something we'd not have made available if we could do it over, e.g. the caller could mess things up by calling Start on the returned task.

src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs
```diff
@@ -5382,6 +5382,52 @@ public static Task Run(Action action, CancellationToken cancellationToken)
...
+        public static Task RunAsGreenThread(Action action, CancellationToken cancellationToken)
+        {
+            Task greenThreadTask = new Task(action, cancellationToken);
+            new Thread(()=>
+            {
+                Thread.t_IsGreenThread = true;
+                // TODO? Do something like register the cancellation token as an AsyncLocal or something.
+                greenThreadTask.RunSynchronously();
+            }).Start();
+            return greenThreadTask;
```
stephentoub: As with below, this can be something like:
```suggestion
            if (cancellationToken.IsCancellationRequested)
                return Task.FromCanceled(cancellationToken);
                
            var tcs = new TaskCompletionSource();
            new Thread(()=>
            {
                if (cancellationToken.IsCancellationRequested)
                {
                    tcs.SetCanceled(cancellationToken);
                }
                
                Thread.t_IsGreenThread = true;
                try
                {
                    action();
                    tcs.SetResult();
                }
                catch (Exception e)
                {
                    tcs.SetException(e);
                }
            }) { IsBackground = true }.Start();
            return tcs.Task;
```

stephentoub:
> Could you guys give an opinion to the api surface, good enough for now?

Looks fine for now.

I'm expecting the only thing the library implementations will need is Thread.IsGreenThread, as every relevant sync method will end up getting a block added to it like:
```C#
public int Read(byte[] buffer, int offset, int count)
{
    if (Thread.IsGreenThread)
    {
        Task t = ReadAsync(buffer, offset, count);
        return t.GetAwaiter().GetResult();
    }

    ... // normal sync thing
}
```
with the wait logic inside of `GetResult`/`Wait` doing the magic interaction with the runtime, which I'm expecting will be along the lines of:
```C#
if (Thread.IsGreenThread)
{
    Thread currentThread = Thread.CurrentThread;
    ContinueWith(_ => MagicGreenThreadResume(currentThread));
    MagicGreenThreadSuspend();
}
```

davidwrighton:
@stephentoub I agree, with your theory on the code that will need to appear in the libraries.

PR #1994, opened 2022-09-26
NativeAOT-LLVM: cast unsigned ints to floats using uitofp

yowl:
This PR fixes a bug when casting ints to floating points.  It assumed signed.  If the source is an unsigned int, then make the cast unsigned (https://llvm.org/docs/LangRef.html#uitofp-to-instruction) .

Adds a test using a byte > 0x7f

cc @SingleAccretion

### DIFF
--- a/src/coreclr/tools/Common/TypeSystem/IL/ILImporter.cs
+++ b/src/coreclr/tools/Common/TypeSystem/IL/ILImporter.cs
@@ -581,7 +581,7 @@ namespace Internal.IL
                         ImportConvert(WellKnownType.Double, false, false);
                         break;
                     case ILOpcode.conv_u4:
-                        ImportConvert(WellKnownType.UInt32, false, false);
+                        ImportConvert(WellKnownType.UInt32, false, true);
                         break;
                     case ILOpcode.conv_u8:
                         ImportConvert(WellKnownType.UInt64, false, true);
--- a/src/coreclr/tools/Common/TypeSystem/IL/ILImporter.cs
+++ b/src/coreclr/tools/Common/TypeSystem/IL/ILImporter.cs
@@ -825,7 +825,7 @@ namespace Internal.IL
                         ImportStoreIndirect(WellKnownType.IntPtr);
                         break;
                     case ILOpcode.conv_u:
-                        ImportConvert(WellKnownType.UIntPtr, false, false);
+                        ImportConvert(WellKnownType.UIntPtr, false, true);
                         break;
                     case ILOpcode.prefix1:
                         opCode = (ILOpcode)(0x100 + ReadILByte());
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -1217,7 +1217,9 @@ namespace Internal.IL
             else if (toStoreKind == LLVMTypeKind.LLVMIntegerTypeKind && (valueTypeKind == LLVMTypeKind.LLVMDoubleTypeKind || valueTypeKind == LLVMTypeKind.LLVMFloatTypeKind))
             {
                 //TODO: keep track of the TypeDesc so we can call BuildUIToFP when the integer is unsigned
-                typedToStore = builder.BuildSIToFP(source, valueType, "CastSIToFloat" + (name ?? ""));
+                typedToStore = unsigned
+                    ? builder.BuildUIToFP(source, valueType, "CastUIToFloat" + (name ?? ""))
+                    : builder.BuildSIToFP(source, valueType, "CastSIToFloat" + (name ?? ""));
             }
             else if ((toStoreKind == LLVMTypeKind.LLVMDoubleTypeKind || toStoreKind == LLVMTypeKind.LLVMFloatTypeKind) &&
                 valueTypeKind == LLVMTypeKind.LLVMIntegerTypeKind)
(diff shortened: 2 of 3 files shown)


jkotas:
@SingleAccretion Could you please take a look at this change? I guess it might have slipped through your inbox.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
```diff
@@ -4035,7 +4037,15 @@ private void ImportConvert(WellKnownType wellKnownType, bool checkOverflow, bool
             }
             else
             {
-                LLVMValueRef converted = CastIfNecessary(loadedValue, GetLLVMTypeForTypeDesc(destType), value.Name(), wellKnownType == WellKnownType.UInt64 /* unsigned is always false, so check for the type explicitly */);
+                TypeDesc sourceType = value.Type;
```
SingleAccretion: Who calls `ImportConvert`?

Edit: I see, it is called from the shared logic.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
```diff
@@ -4035,7 +4037,15 @@ private void ImportConvert(WellKnownType wellKnownType, bool checkOverflow, bool
...
+
+                LLVMValueRef converted = CastIfNecessary(loadedValue, GetLLVMTypeForTypeDesc(destType), value.Name(),
+                    wellKnownType == WellKnownType.UInt64 /* unsigned is always false, so check for the type explicitly */
+                    || sourceType.IsWellKnownType(WellKnownType.Byte)
+                    || sourceType.IsWellKnownType(WellKnownType.UInt16)
+                    || sourceType.IsWellKnownType(WellKnownType.UInt32)
+                    || sourceType.IsWellKnownType(WellKnownType.UInt64)
+                    );
```
SingleAccretion: How does the ILImporter stack work -- does it keep the precise types?

If it does, it does not look correct to look at them. E. g.
```
ldobj int64
conv.r.un
```
Should still be treated as an unsigned cast.

> `/* unsigned is always false, so check for the type explicitly */`

Is that true? I see it should be `true` for `conv.r.un`.

yowl: Hmm, no that's not true, let me try to clean this up.

yowl: There were 2 reported problems, using a byte as an index and casting a byte to a float (for an alpha channel).  But looking at the code I don't see why I can't just pass the right value from the op code.

src/coreclr/tools/Common/TypeSystem/IL/ILImporter.cs
```diff
@@ -581,7 +581,7 @@ private void ImportBasicBlock(BasicBlock basicBlock)
                         ImportConvert(WellKnownType.Double, false, false);
                         break;
                     case ILOpcode.conv_u4:
-                        ImportConvert(WellKnownType.UInt32, false, false);
+                        ImportConvert(WellKnownType.UInt32, false, true);
```
jkotas, line 584: Could you please submit this fix to dotnet/runtime as well?

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
SingleAccretion:
```suggestion
                LLVMValueRef converted = CastIfNecessary(loadedValue, GetLLVMTypeForTypeDesc(destType), value.Name(), unsigned);
```
Presumably that should now work?

I also see there is:
```
            else if ((toStoreKind == LLVMTypeKind.LLVMDoubleTypeKind || toStoreKind == LLVMTypeKind.LLVMFloatTypeKind) &&
                valueTypeKind == LLVMTypeKind.LLVMIntegerTypeKind)
            {
                //TODO: keep track of the TypeDesc so we can call BuildFPToUI when the integer is unsigned
                typedToStore = builder.BuildFPToSI(source, valueType, "CastFloatSI" + (name ?? ""));
            }
```
In `CastIfNecessary` that, could, presumably also be fixed.

yowl:
I've put code to widen the bytes/shorts to ints before the `conv`.  Added another test.  Hopefully this fits the IL semantics better.

Example LLVM is now
```ll
  store i8 -1, i8* %alpha_local0_, align 1, !dbg !389570
  %Loadloc0_ = load i8, i8* %alpha_local0_, align 1, !dbg !389571
  %CastZIntloc0_ = zext i8 %Loadloc0_ to i32, !dbg !389571
  %CastSIToFloatloc0_ = sitofp i32 %CastZIntloc0_ to float, !dbg !389571
  %fpextop2 = fpext float %CastSIToFloatloc0_ to double, !dbg !38957
```

yowl:
ugg, I've messed up the branches somehow... moving to WIP

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
```diff
@@ -4026,27 +4027,34 @@ private void ImportConvert(WellKnownType wellKnownType, bool checkOverflow, bool
...
             //TODO checkOverflow - r_un & r_4, i & i_un
             StackEntry value = _stack.Pop();
+            TypeDesc stackType = value.Type;
+
             TypeDesc destType = GetWellKnownType(wellKnownType);
 
             // Load the value and then convert it instead of using ValueAsType to avoid loading the incorrect size
             LLVMValueRef loadedValue = value.ValueAsType(value.Type, _builder);
+            LLVMValueRef widenedStackValue = CastIfNecessary(loadedValue, GetLLVMTypeForTypeDesc(WidenBytesAndShorts(value.Type)), value.Name(),
+                stackType.IsWellKnownType(WellKnownType.Byte) ||
```
SingleAccretion, line 4038: Missing `Bool`?

SingleAccretion (approved):
LGTM modulo one comment.

jkotas:
@yowl @SingleAccretion Thank you!

PR #2001, opened 2022-09-28
NativeAOT-LLVM: extern globals should not be constant 

yowl:
This PR fixes an optimisation problem where clang will remove the externs if they are constant and there are only `store`s in the input and optimisation is on

cc @SingleAccretion.


SingleAccretion (approved):
> if they are constant and there are only stores in the input and optimisation is on

Does this mean the code was storing into constant globals?

yowl:
yes

yowl:
The WAT difference between unoptimised and optimised was like this:
image

The `store` was dropped

PR #2002, opened 2022-09-29
Initial green thread prototype

davidwrighton:
Initial stab at enabling green threads.
- Implement via the strategy of stack segments, with a greenfield implementation of concept to avoid IP concerns although the design is not that unusual
  - This is likely not a viable long term strategy, but it should work for the purposes of a prototype
- Tie it into Tasks. An infinite wait in a Green Thread will suspend the GC thread. Waits with timeouts are not integrated.

Issues...
- While the suspension resumption logic is capable of working with multiple segments, our GC stackwalker is not yet, and the JIT does not generate the appropriate prolog, so the size of a stack segment is very large.
- GC tracking in suspended threads is not enabled. If a GC occurs, there will be heap corruption
- ThreadStatics are not integrated
- PInvokes to non-green-thread safe apis do not trigger thread pinning, and so can result in incorrect behavior
- Unique managed thread objects/ids are not implemented
- JIT does not generate the new prolog
- Integration into `Task.Wait` is probably wrong in various ways.
- FCalls do not swap to the OS stack
- PInvokes do not swap to the OS stack


src/coreclr/System.Private.CoreLib/src/System/Threading/GreenThreadTasks.CoreCLR.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+                finally
+                {
+                    GreenThreadStatics.t_TaskToWaitFor = null;
+                    Thread.t_IsGreenThread = false;
+                }
+            }
+        }
+
+        static partial void RunOnActualGreenThread(Action action, ref bool ranAsActualGreenThread)
```
AntonLapounov, line 89: Is there a reason to use `ref bool` instead of `out bool`?

src/coreclr/vm/amd64/AsmHelpers.asm
```diff
@@ -698,5 +698,240 @@ NESTED_END OnCallCountThresholdReachedStub, _TEXT
...
+        sub eax, 1
+        call _more_stack
+        add rsp, 28h
+        ret
+        jmp rcx
+NESTED_END TransitionToOSThreadHelper, _TEXT
+
+
+; extern "C" void YieldOutOfGreenThreadHelper(StackRange *pOSStackRange, uint8_t* osStackCurrent, uint8_t** greenThreadStackCurrent);
+; This function is supposed to hop back to the OS thread, with the original OS thread stack in place, and
```
AntonLapounov: The comment seems unfinished.

src/coreclr/vm/amd64/AsmHelpers.asm
```diff
@@ -698,5 +698,240 @@ NESTED_END OnCallCountThresholdReachedStub, _TEXT
...
+    mov [r8], rsp ; Capture current rsp into the greenTheadStackCurrent variable
+    lea rsp, [rbp-0e0h]
+    jmp yield_point
+ALTERNATE_ENTRY resume_point
+    EPILOG_WITH_TRANSITION_BLOCK_RETURN
+NESTED_END YieldOutOfGreenThreadHelper, _TEXT
+
+extern GetResumptionStackPointerAndSaveOSStackPointer:proc
+
+; This is a paired function with the _more_stack function to allow for
```
AntonLapounov: And this one.

src/coreclr/vm/threads.h
```diff
@@ -1245,11 +1245,7 @@ class Thread
     }
 #endif // DACCESS_COMPILE
 
-    DWORD IsExecutingOnAltStack()
-    {
-        LIMITED_METHOD_CONTRACT;
-        return (m_State & TS_ExecutingOnAltStack);
-    }
+    DWORD IsExecutingOnAltStack();
```
AntonLapounov: I am wondering why the implementation was moved to .cpp.

src/coreclr/System.Private.CoreLib/src/System/Threading/GreenThreadTasks.CoreCLR.cs
davidwrighton: Yes, partial methods are not permitted to have out parameters. Effectively, if the method implementation does not exist, then the method isn't called and any out parameters would have no meaning.

src/coreclr/vm/threads.h
davidwrighton: Ah, I should move it back. Initially I had this checking the thread local variables to report this, but I realized that we use during GC stack walking, and checking the current threads green thread state is a bad idea.

src/coreclr/vm/amd64/AsmHelpers.asm
davidwrighton: It does doesn't it. :) Thank you for reading enough to find this. I'll see if I can remember the thought I had before I went and got distracted trying to put something else together.

src/coreclr/vm/amd64/AsmHelpers.asm
```diff
@@ -698,5 +698,246 @@ NESTED_END OnCallCountThresholdReachedStub, _TEXT
...
+        mov             r8,  [rsp + 80h]
+        mov             r9,  [rsp + 88h]
+        movdqa          xmm0, [rsp + 20h]
+        movdqa          xmm1, [rsp + 30h]
+        movdqa          xmm2, [rsp + 40h]
+        movdqa          xmm3, [rsp + 50h]
+
+        ; Swap to new stack
+        mov             rbx, rsp ; Stash old stack pointer into rbx (saved register)
+        mov             rsp, rax
```
kouvel, line 768: Probably ok for now, but wondering if it would be feasible to use rsp for the virtual stack pointer at all, as I figure the OS could interrupt the thread and end up using the virtual stack while it's close to the limit.

davidwrighton: Agreed. We've had an interesting idea around this. At the moment I leave a fair bit of space as a "red-zone" but I'm not convinced it is a good idea long-term. Send me a message on Teams to discuss whenever you'd like to.

PR #2003, opened 2022-09-29
NativeAOT-LLVM: stelem always use unsigned cast

yowl:
This PR fixes a bug when using a byte > 0x7f for an index as it was using a signed cast to `i32`

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -5100,7 +5100,7 @@ namespace Internal.IL
             StackEntry index = _stack.Pop();
             StackEntry arrayReference = _stack.Pop();
             var nullSafeElementType = elementType ?? GetWellKnownType(WellKnownType.Object);
-            LLVMValueRef elementAddress = GetElementAddress(index.ValueAsInt32(_builder, true), arrayReference.ValueAsType(LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0), _builder), nullSafeElementType);
+            LLVMValueRef elementAddress = GetElementAddress(index.ValueAsInt32(_builder, false), arrayReference.ValueAsType(LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0), _builder), nullSafeElementType);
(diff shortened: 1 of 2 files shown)


yowl:
cc @SingleAccretion

therightstatement:
This would be awesome! Thank you for the fix

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
```diff
@@ -5100,7 +5100,7 @@ private void ImportStoreElement(TypeDesc elementType)
             StackEntry index = _stack.Pop();
             StackEntry arrayReference = _stack.Pop();
             var nullSafeElementType = elementType ?? GetWellKnownType(WellKnownType.Object);
-            LLVMValueRef elementAddress = GetElementAddress(index.ValueAsInt32(_builder, true), arrayReference.ValueAsType(LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0), _builder), nullSafeElementType);
+            LLVMValueRef elementAddress = GetElementAddress(index.ValueAsInt32(_builder, false), arrayReference.ValueAsType(LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0), _builder), nullSafeElementType);
```
SingleAccretion, line 5103: Zero-extending signed small types does not look correct, but then the values for which that would matter (negative ones) would also induce an IORE, which this code (seemingly) doesn't throw, and presumably we're not investing in fixing that, so it is fine to leave things be as they are.

SingleAccretion (approved):

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
yowl: Yes, we are missing IORE.

yowl: ...which would have saved me hours of debugging, but heh ho.

yowl: @jkotas Sorry to prod, but I can't build working packages locally for some reason at the moment.  If you can merge this that would be great, thanks

PR #2004, opened 2022-09-29
NativeAOT-LLVM: Add void to .c functions that have no parameters - something about K&R functions??

yowl:
This PR fixes this warning which seems to be new in later versions of clang.  Interestingly runtime doesn't have this fix, so don't know how that is working, maybe its not on the later versions of clang yet.  If the header declares the args as `void` the `.c` needs to do the same to avoid the warning.  At least that's my observation...

```
E:/GitHub/runtimelab/src/libraries/Native/Unix/System.Globalization.Native/pal_icushim.c(509,42): error G0E77CA38: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes] [E:\GitHub\runtimelab\src\libraries\Native\build-native.proj]
```

https://reviews.llvm.org/D122895  

Might be relevant, there's a mention of void in the comments.

### DIFF
--- a/src/libraries/Native/Unix/CMakeLists.txt
+++ b/src/libraries/Native/Unix/CMakeLists.txt
@@ -129,6 +129,7 @@ if(CMAKE_C_COMPILER_ID STREQUAL Clang)
     add_compile_options(-Wno-typedef-redefinition)
     add_compile_options(-Wno-c11-extensions)
     add_compile_options(-Wno-thread-safety-analysis)
+    add_compile_options(-Wno-strict-prototypes)
 endif()
 
 add_subdirectory(System.Native)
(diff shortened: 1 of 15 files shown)


yowl:
cc @SingleAccretion

jkotas:
dotnet/runtime "fix" for this problem was to disable the warning: https://github.com/dotnet/runtime/pull/70693/files#diff-7f8d8bfbe22b648f50efce6df1dddb9fb53e13efa083a0f54c7a3f2ac5e5bb15R153 . Would you like to do the same in this branch to avoid growing unnecessary delta?

Also, I am happy to take this fix and delete the warning disable in dotnet/runtime if you would like to submit it there. Thanks!

jkotas:
For the feature/NativeAOT-LLVM branch, we should just disable the warning, and wait for the full fix to come with merge from main.

@yowl Could you please update this PR to just disable the warnings?

yowl:
Sure, this looks ok locally, hopefully it will pass.

PR #2005, opened 2022-09-29
NativeAOT-LLVM: Add support for GT_STORE_BLK

yowl:
This PR adds support for `GT_STORE_BLK`.  For this IR:
```cs
                                                  /--*  t51    byref
                                                  +--*  t48    struct
N004 ( 19,  6) [000053] -ACXG-------              *  STORE_BLK struct<System.DateTimeOffset, 16> (copy)
```
We generate
```ll
  %37 = bitcast i8* %35 to %"[S.P.CoreLib]System.DateTimeOffset"*, !dbg !8
  store %"[S.P.CoreLib]System.DateTimeOffset" %36, %"[S.P.CoreLib]System.DateTimeOffset"* %37, align 1, !dbg !8
```

WIP as behind some other things

cc @SingleAccretion

I think this will help with the FIELD_LIST work in https://github.com/dotnet/runtimelab/issues/1859

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1825,16 +1825,16 @@ void Llvm::buildStoreBlk(GenTreeBlk* blockOp)
 
     CORINFO_CLASS_HANDLE structClsHnd  = structLayout->GetClassHandle();
     StructDesc*          structDesc    = getStructDesc(structClsHnd);
-    bool targetNotHeap = blockOp->OperIsBlk() || ((blockOp->gtFlags & GTF_IND_TGT_NOT_HEAP) != 0) || blockOp->Addr()->OperIsLocalAddr();
 
     Value* dataValue = getGenTreeValue(blockOp->Data());
-    if (targetNotHeap)
+    if (structLayout->HasGCPtr() && ((blockOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0) &&
+        !blockOp->Addr()->OperIsLocalAddr())
     {
-        _builder.CreateStore(dataValue, castIfNecessary(baseAddressValue, dataValue->getType()->getPointerTo())); 
+            storeObjAtAddress(baseAddressValue, dataValue, structDesc);
     }
     else
     {
-        storeObjAtAddress(baseAddressValue, dataValue, structDesc);
+        _builder.CreateStore(dataValue, castIfNecessary(baseAddressValue, dataValue->getType()->getPointerTo()));
     }
 }
 
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1620,7 +1620,13 @@ void Llvm::buildReturn(GenTree* node)
         case TYP_UINT:
         case TYP_LONG:
         case TYP_ULONG:
-            _builder.CreateRet(castIfNecessary(getGenTreeValue(node->gtGetOp1()), getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass)));
+            if (node->gtGetOp1()->TypeIs(TYP_FLOAT))
+            {
+                // TODO-LLVM: remove this case by lowering see
+                // https://github.com/dotnet/runtimelab/pull/2007#issuecomment-1264715441
+                failFunctionCompilation();
+            }
+            _builder.CreateRet(consumeValue(node->gtGetOp1(), getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass)));
             return;
         case TYP_VOID:
             _builder.CreateRetVoid();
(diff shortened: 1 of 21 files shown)


yowl:
Removed the sig(void) changes and putting back for review, thanks.

src/coreclr/jit/llvm.cpp
```diff
@@ -2043,6 +2049,9 @@ void Llvm::visitNode(GenTree* node)
         case GT_STOREIND:
             buildStoreInd(node->AsStoreInd());
             break;
+        case GT_STORE_BLK:
+            buildStoreBlk(node->AsBlk());
```
SingleAccretion: `STORE_BLK` is almost identical to `STORE_OBJ` (the only difference is that `STORE_BLK` can represent `Unsafe.BlockCopy`), so I think it would make sense to just use `buildStoreObj`.

yowl: and rename to `buildStoreBlk` I suppose as `OBJ` is a subset of `BLK`

yowl: I'm assuming `BLK` operations are never on the heap
```
bool targetNotHeap = blockOp->OperIsBlk() || ((blockOp->gtFlags & GTF_IND_TGT_NOT_HEAP) != 0) || blockOp->Addr()->OperIsLocalAddr();
```

SingleAccretion: Well, it's not that they never store to heap, it's that the structs never have GC pointers. The conditions could be something like this:
```
if (structLayout->HasGCPtr() && ((blockOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0) && !blockOp->Addr()->OperIsLocalAddr())
{
    // Use the complex store with write barriers.
}
else
{
    // Use the simple LLVM store.
}
```

src/coreclr/jit/llvm.cpp
```diff
@@ -1804,19 +1804,19 @@ void Llvm::storeObjAtAddress(Value* baseAddress, Value* data, StructDesc* struct
...
 }
 
-void Llvm::buildStoreObj(GenTreeObj* storeOp)
+void Llvm::buildStoreBlk(GenTreeBlk* blockOp)
 {
-    if (!storeOp->OperIsBlk())
+    if (!blockOp->OperIsBlk())
     {
         failFunctionCompilation(); // cant get struct handle. TODO-LLVM: try as an assert
     }
```
```suggestion
```
Dead code; both `GT_STORE_BLK` and `GT_STORE_BLK` are `OperIsBlk`.

src/coreclr/jit/llvm.cpp
```diff
@@ -1825,9 +1825,9 @@ void Llvm::buildStoreObj(GenTreeObj* storeOp)
 
     CORINFO_CLASS_HANDLE structClsHnd  = structLayout->GetClassHandle();
     StructDesc*          structDesc    = getStructDesc(structClsHnd);
-    bool targetNotHeap = ((storeOp->gtFlags & GTF_IND_TGT_NOT_HEAP) != 0) || storeOp->Addr()->OperIsLocalAddr();
+    bool targetNotHeap = blockOp->OperIsBlk() || ((blockOp->gtFlags & GTF_IND_TGT_NOT_HEAP) != 0) || blockOp->Addr()->OperIsLocalAddr();
```
```suggestion
    bool targetNotHeap = ((blockOp->gtFlags & GTF_IND_TGT_NOT_HEAP) != 0) || blockOp->Addr()->OperIsLocalAddr();
```
(See also the write barrier discussion)

src/coreclr/jit/llvm.cpp
yowl: ok, thanks, thought you might comment on that.

src/coreclr/jit/llvm.cpp
yowl: We dont need `targetNotHeap`  if we have the other check do we?

SingleAccretion: Yep.

SingleAccretion (approved):

src/coreclr/jit/llvm.cpp
```diff
@@ -1825,16 +1821,16 @@ void Llvm::buildStoreObj(GenTreeObj* storeOp)
...
-    bool targetNotHeap = ((storeOp->gtFlags & GTF_IND_TGT_NOT_HEAP) != 0) || storeOp->Addr()->OperIsLocalAddr();
 
-    Value* dataValue = getGenTreeValue(storeOp->Data());
-    if (targetNotHeap)
+    Value* dataValue = getGenTreeValue(blockOp->Data());
+    if (structLayout->HasGCPtr() && ((blockOp->gtFlags & GTF_IND_TGT_NOT_HEAP) == 0) &&
+        !blockOp->Addr()->OperIsLocalAddr())
     {
-        _builder.CreateStore(dataValue, castIfNecessary(baseAddressValue, dataValue->getType()->getPointerTo())); 
+            storeObjAtAddress(baseAddressValue, dataValue, structDesc);
```
```suggestion
        storeObjAtAddress(baseAddressValue, dataValue, structDesc);
```

src/coreclr/jit/llvm.h
```diff
@@ -203,7 +203,7 @@ class Llvm
     unsigned int getTotalLocalOffset();
     bool helperRequiresShadowStack(CORINFO_METHOD_HANDLE corinfoMethodHnd);
     void buildStoreInd(GenTreeStoreInd* storeIndOp);
-    void buildStoreObj(GenTreeObj* indirOp);
+    void buildStoreBlk(GenTreeBlk* indirOp);
```
```suggestion
    void buildStoreBlk(GenTreeBlk* blockOp);
```

yowl:
Thanks!

PR #2006, opened 2022-10-01
NativeAOT-LLVM: Allow cast from i1 to i8

yowl:
This PR allows a widening from `i1` to `i8` using a zero extend. With optimisations we can generate the IR
```
N005 ( 16, 15) [000028] ------------        t28 = *  EQ        int    $180
                                                  /--*  t28    int
N006 ( 17, 16) [000016] ------------              *  RETURN    int    $1c0
```
But the LLVM backend currently represents bools as `i8`s.  This PR allows the return to widen to match the signature.  The resulting LLVM with this change:
```ll
  %14 = icmp eq i64 %and, %and1
  %15 = zext i1 %14 to i8
  ret i8 %15
```

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1620,6 +1620,12 @@ void Llvm::buildReturn(GenTree* node)
         case TYP_UINT:
         case TYP_LONG:
         case TYP_ULONG:
+            if (node->gtGetOp1()->TypeIs(TYP_FLOAT))
+            {
+                // TODO-LLVM: remove this case by lowering see
+                // https://github.com/dotnet/runtimelab/pull/2007#issuecomment-1264715441
+                failFunctionCompilation();
+            }
             _builder.CreateRet(consumeValue(node->gtGetOp1(), getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass)));
             return;
         case TYP_VOID:
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1625,7 +1625,7 @@ void Llvm::buildReturn(GenTree* node)
         case TYP_UINT:
         case TYP_LONG:
         case TYP_ULONG:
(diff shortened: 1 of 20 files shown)


yowl:
cc @SingleAccretion

SingleAccretion:
I think it would be better to switch `buildReturn` to use `consumeValue`. It should already handle this case.

SingleAccretion:
There is also dead code in the form of `zextIntIfNecessary` which mentions this problem and can be deleted.

yowl:
Thanks, changed to `consumeValue` and removed the dead method.

src/coreclr/jit/llvm.cpp
```diff
@@ -748,6 +748,11 @@ llvm::Instruction* Llvm::getCast(llvm::Value* source, Type* targetType)
                 {
                     return new llvm::TruncInst(source, targetType, "TruncInt");
                 }
+                // allow just the specific case of bools being represented in LLVM as i8
+                else if (sourceType == Type::getInt1Ty(_llvmContext) && targetType == Type::getInt8Ty(_llvmContext))
+                {
+                    return new llvm::ZExtInst(source, targetType, "BitToByte");
+                }
```
```suggestion
```

yowl: oh sorry, too much haste..  Thanks for the commit suggestion

SingleAccretion (approved):
LGTM modulo the `void` changes that have inadvertently snuck in.

jkotas:
Build break
```
  RyuJIT compilation results, total methods 16917 RyuJit Methods 10070 59.5259%
EXEC : error : Invalid cast (Producer: 'LLVM11.0.0' Reader: 'LLVM 15.0.0git') [D:\a\_work\1\s\artifacts\tests\coreclr\Browser.wasm.Release\nativeaot\SmokeTests\DeadCodeElimination\DeadCodeElimination\nativebuild.proj]
```

yowl:
> Build break
> 
> ```
> EXEC : error : Invalid cast (Producer: 'LLVM11.0.0' Reader: 'LLVM 15.0.0git') [D:\a\_work\1\s\artifacts\tests\coreclr\Browser.wasm.Release\nativeaot\SmokeTests\DeadCodeElimination\DeadCodeElimination\nativebuild.proj]
> ```

Thanks, this is https://github.com/dotnet/runtimelab/pull/2007#issuecomment-1264715441.  I'll bring failing these functions to this PR.

Issue #2017, opened 2022-10-03
[Green Threads] FCalls do not swap to the OS stack


davidwrighton:
Known issue from https://github.com/dotnet/runtimelab/pull/2002

PR #2019, opened 2022-10-04
Make FCall transition go through lambda that can be translated into a functionpointer + pointer arg parameter

davidwrighton:
Make FCall helper frames actual enclose a lambda, and if not on the OS thread, transition to it and call the lambda.

Fixes dotnet/runtimelab#2017

### DIFF
--- a/src/coreclr/vm/fcall.h
+++ b/src/coreclr/vm/fcall.h
@@ -560,11 +560,7 @@ LPVOID __FCThrowArgument(LPVOID me, enum RuntimeExceptionKind reKind, LPCWSTR ar
 
 // END: after gcpoll
 //__fcallGcCanTrigger.Leave(__FUNCTION__, __FILE__, __LINE__);
-struct PairOfPointers
-{
-    void *innerLambda;
-    void *innerLambdaFunctionPointer;
-};
+
 // We have to put DEBUG_OK_TO_RETURN_BEGIN around the FORLAZYMACHSTATE
 // to allow the HELPER_FRAME to be installed inside an SO_INTOLERANT region
 // which does not allow a return.  The return is used by FORLAZYMACHSTATE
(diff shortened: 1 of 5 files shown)


src/coreclr/vm/fcall.h
```diff
@@ -559,7 +560,11 @@ LPVOID __FCThrowArgument(LPVOID me, enum RuntimeExceptionKind reKind, LPCWSTR ar
 
 // END: after gcpoll
 //__fcallGcCanTrigger.Leave(__FUNCTION__, __FILE__, __LINE__);
-
+struct PairOfPointers
```
markples: I don't think you're using this (yet?)

src/coreclr/vm/fcall.h
```diff
@@ -584,12 +589,14 @@ LPVOID __FCThrowArgument(LPVOID me, enum RuntimeExceptionKind reKind, LPCWSTR ar
             /* gcpoll; */                                                       \
             INSTALL_MANAGED_EXCEPTION_DISPATCHER;                               \
             __helperframe.Push();                                               \
+            auto helperFrameLambda = [&](){                                     \
```
markples: Although it's a formatting nit, it would probably be good to indent the lambda bodies just to help with readability.  Perhaps that's just special for me since this is my first time read this file..

src/coreclr/vm/greenthreads.cpp
```diff
@@ -220,6 +225,30 @@ uintptr_t TransitionToOSThread(TakesOneParam functionToExecute, uintptr_t param)
     return result;
 }
 
+void TransitionToOSThread(TakesOneParamNoReturn functionToExecute, uintptr_t param)
```
markples, line 256: I suspect you're going to want/need to combine the code for these fairly soon as it will be difficult to keep them in sync, review, etc.

src/coreclr/vm/fcall.h
```diff
@@ -616,11 +623,21 @@ LPVOID __FCThrowArgument(LPVOID me, enum RuntimeExceptionKind reKind, LPCWSTR ar
 
 #define HELPER_METHOD_FRAME_END_EX(gcpoll,allowGC)                          \
             UNINSTALL_UNWIND_AND_CONTINUE_HANDLER;                          \
+            };                                                              \
+            auto nonCapturingLambda = [](uintptr_t pointerToLambda) { \
+                (*(decltype(helperFrameLambda)*)pointerToLambda)(); \
+            }; \
+            CallOnOSThread(nonCapturingLambda, (uintptr_t)&helperFrameLambda); \
```
markples, line 625: I assume I haven't gotten far enough past this PR... while obviously you need a function pointer to pass off, why do you need the two layers (or "functionpointer + pointer arg" as in the PR title")?  Is this setting up for a future where the wrapper function can be replaced by something else?

markples (approved):

src/coreclr/vm/fcall.h
davidwrighton: Oh, yeah, I should delete this.

src/coreclr/vm/greenthreads.cpp
davidwrighton: I'm hoping that we don't need to write any more of these manually, and the rest of them get autogenerated by something (hopefully in the JIT or by whatever we need for the JIT to make this work.)

src/coreclr/vm/fcall.h
davidwrighton: Oh, that's a peculiarity of C++ lambdas. You cannot get a normal function pointer to a lambda which captures any variables. However, I need that for the TransitionToOSThread function. So we have a 2 level structure where the functionpointer passed to the CallOSThread function is a captureless lambda, but we can then call through to the inner lambda with all the args in place. (Its legal to have a pointer to any lambda that can be converted to/from void*). We need the captureless lambda, as that is needed to know the actual type of the "useful" lambda.

PR #2020, opened 2022-10-07
Instrument RandomAccess.Read/Write for green threads

stephentoub:
This _should_ make most file reading/writing throughout the system (e.g. File, FileStream, StreamWriter, etc.) go through the async paths for the sync operations if on a green thread.  It's completely untested, but, you know, it builds.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/IO/RandomAccess.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/IO/RandomAccess.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Buffers;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


davidwrighton:
I think this should work, but I wonder if we need another api, something like `IsGreenThreadAndIsSuspendable` There are conditions where this would end up generating somewhat pessimized logic which injects sync over async behavior.

stephentoub:
> IsGreenThreadAndIsSuspendable

This would be for a case such as where the thread has a native frame on the stack and so there's no point in doing this work because the thread's going to block anyway?

That seems reasonable.  At the same time, though, we'd hopefully expect that to be the minority case... if it's not, then this whole effort is unlikely to be fruitful, since it prevents the essence of what the whole feature achieves.  I'm not sure spending a little more effort in the minority case will significantly move the needle when we have to exert that effort in what's hopefully the vast majority case.

davidwrighton (approved):

Issue #2018, opened 2022-10-03
[Green Threads] PInvokes do not swap to the OS stack


davidwrighton:
Known issue from https://github.com/dotnet/runtimelab/pull/2002

PR #2022, opened 2022-10-07
Conservative GC reporting for suspended green threads

cshung:
The GC changes come from https://github.com/dotnet/runtime/pull/76737 which is required for conservative reporting. Once we have that PR merged, we should take an FI instead to avoid unnecessary conflicts.

The green thread changes slightly extended the lifetime of the `SuspendedGreenThread` object so that I can use it for GC reporting. The change simply loop through all of them and reported all the stack segments conservatively.

In debug mode, the process terminates with an exit code `1073807370` == `0x4001000a`, this is because we hit an assert [here](https://github.com/dotnet/runtimelab/blob/5074833015a2f1bce6539c4bcd8b641699726d27/src/coreclr/vm/jithelpers.cpp#L5951). I assume this will be handled as part of #2018

### DIFF
--- a/src/coreclr/vm/gcenv.ee.cpp
+++ b/src/coreclr/vm/gcenv.ee.cpp
@@ -274,6 +274,36 @@ void GCToEEInterface::GcScanRoots(promote_func* fn, int condemned, int max_gen, 
         STRESS_LOG2(LF_GC | LF_GCROOTS, LL_INFO100, "Ending scan of Thread %p ID = 0x%x }\n", pThread, pThread->GetThreadId());
     }
 
+#ifdef FEATURE_GREENTHREADS
+    {
+        if (sc->promotion && green_head.next)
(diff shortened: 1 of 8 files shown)


src/coreclr/vm/gcenv.ee.cpp
```diff
@@ -274,6 +274,34 @@ void GCToEEInterface::GcScanRoots(promote_func* fn, int condemned, int max_gen,
         STRESS_LOG2(LF_GC | LF_GCROOTS, LL_INFO100, "Ending scan of Thread %p ID = 0x%x }\n", pThread, pThread->GetThreadId());
     }
 
+    {
```
davidwrighton, line 278: Could you put this under FEATURE_GREENTHREADS

davidwrighton (approved):

PR #2023, opened 2022-10-08
Green thread objects

### LINKED ISSUE #2014: [Green Threads] Unique managed thread objects/ids are not implemented

davidwrighton:
This adds managed `Thread` objects for green threads as well as unique managed thread ids. There is a test for that. 

In addition this adds logic to moves handling of Monitor locks to a per-green thread structure.

In addition it fixes a number of bugs

- In the reverse p/invoke logic where the current native `Thread*` pointer is cached, and when we hop threads isn't properly restored. (See change in jithelpers.cpp)
- This also fixes bugs where the stack limits were not properly updated on resume
- This fixes bugs where in the presence of optimization, when the thread was resumed on a different thread the Thread* was not refreshed from the underlying TLS data structure. This issue was found in several functions in C++ (which are now marked with optimizations disabled, and the GreenThread_Yield P/Invoke which has the same issue.)

Fixes #2014

### DIFF
--- a/src/coreclr/System.Private.CoreLib/src/System/Threading/GreenThreadTasks.CoreCLR.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/Threading/GreenThreadTasks.CoreCLR.cs
@@ -41,6 +41,7 @@ namespace System.Threading.Tasks
             try
             {
                 var action = (Action?)obj;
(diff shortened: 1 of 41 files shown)


cshung (approved):
I read through the changes in Thread/SyncBlock, I only skimmed through the refactoring.
Looks good to me.

PR #2034, opened 2022-10-12
NativeAOT-LLVM: remove an SSA assert for TARGET_WASM

yowl:
This PR removes an assert we hit in optimized code.  Would you like the full dump for an example method or is it obvious (it's not to me..)?

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/ssabuilder.cpp
+++ b/src/coreclr/jit/ssabuilder.cpp
@@ -1184,8 +1184,10 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
     {
         if ((memoryKind == GcHeap) && m_pCompiler->byrefStatesMatchGcHeapStates)
         {
+#if !defined(TARGET_WASM)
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/ssabuilder.cpp
```diff
@@ -1184,8 +1184,10 @@ void SsaBuilder::BlockRenameVariables(BasicBlock* block)
     {
         if ((memoryKind == GcHeap) && m_pCompiler->byrefStatesMatchGcHeapStates)
         {
+#if !defined(TARGET_WASM)
             // ByrefExposed and GcHeap share any phi this block may have,
             assert(block->bbMemorySsaPhiFunc[memoryKind] == block->bbMemorySsaPhiFunc[ByrefExposed]);
```
SingleAccretion, line 1191: It means memory liveness on LIR is broken. This is a known upstream issue. Would be nice to add a comment to that effect.

SingleAccretion (approved):

jkotas (approved):

PR #2041, opened 2022-10-16
NativeAOT-LLVM: remove load expressions from expression stack when not immediately used

yowl:
This PR removes the use of the LoadExpression stack entry for cases where the use or pop of the expression may occur after the value in the address being loaded is modified.  This bug looks to have been present for some time, but has been highlighted by recent versions of Roslyn which appear to eliminate more locals in exchange for leaving expressions on the stack.


yowl:
I added a test, but it only repros with .net 7 roslyn in release config it seems.  So its a bit redundant with the repo on .net 6 sdk, but doesn't harm.

yowl:
cc @SingleAccretion

SingleAccretion (approved):

PR #2042, opened 2022-10-16
NativeAOT-LLVM: Process blocks in SSA dominance to fix CSE mapping issue

yowl:
This PR allows more CSEs to be processed in the clrjit backend.  It changes the order of compiling basic blocks to SSA dominance preorder so that the map of lclnum/ssanum finds the SSA definition before the use.

WIP as requires optimisation in the clrjit backend.


SingleAccretion:
@yowl seems we can proceed with this now?

yowl:
> @yowl seems we can proceed with this now?

yes, this should go green.

SingleAccretion (approved):
LGTM.

yowl:
@jkotas with all the package problems, this may have got lost in the noise.  Is it ok?  Thanks very much.

jkotas (approved):

yowl:
Thank you!

PR #2048, opened 2022-10-18
Add support for ThreadStatic variables to the green threads implementation

### LINKED ISSUE #2012: [Green Threads] ThreadStatics are not integrated

davidwrighton:
- Move thread statics into the ThreadBase type
- Update the green thread startup/suspension/resumption code to deal with the fact that managed thread locals are now truly local to the green thread.

Fixes #2012


davidwrighton:
@kouvel With this change I think we now have the right bits and pieces in place to implement flowing execution context correctly. Would you be the right person to figure out what flowing them correctly actually means? And to write some tests to validate whatever the correct behavior is?

kouvel:
> Would you be the right person to figure out what flowing them correctly actually means? And to write some tests to validate whatever the correct behavior is?

Sure I can take a look

PR #2053, opened 2022-10-21
Transition to os thread on pinvoke

### LINKED ISSUE #2018: [Green Threads] PInvokes do not swap to the OS stack

davidwrighton:
Following the example of @jakobbotsch's implementation of CFG, I've implemented a small helper call to transition to the OS thread. (Effectively, it replaces the function pointer jumped to for the P/Invoke to jump through a helper function, after stashing away the extra details needed for the OS thread transition in TLS variables.

This change includes
- The JIT changes to do that
- New helper method enum for the magic helper function
- Implementation of arbitrary helper method
- Refactoring how inGreenThread flag is set so that we don't need an interstitial frame to make it work

Fixes #2018


src/coreclr/jit/lower.cpp
```diff
@@ -2198,6 +2210,131 @@ GenTree* Lowering::LowerTailCallViaJitHelper(GenTreeCall* call, GenTree* callTar
...
+            //
+            // ta... = <early args> (without GT_PUTARG_* nodes)
+            // tb = callTarget
+            // GT_CALL CORINFO_HELP_VALIDATE_INDIRECT_CALL, tb
+            // tc = GT_PHYSREG REG_VALIDATE_INDIRECT_CALL_ADDR (preserved by helper)
+            // td = <moved GT_PUTARG_* nodes>
+            // GT_CALL tb, ta..., td..
+            //
+
+            GenTree* regNode = PhysReg(REG_INTRET, TYP_I_IMPL);
```
jakobbotsch: I realize this is an experiment, just pointing out that eventually there should be no need to use `PhysReg` here, which may pessimize codegen. Instead you can just replace with the use of the `validate` call created below.

src/coreclr/jit/gentree.h
```diff
@@ -4319,6 +4320,7 @@ enum class WellKnownArg
     R2RIndirectionCell,
     ValidateIndirectCallTarget,
     DispatchIndirectCallTarget,
+    StackSizeArg,
```
jakobbotsch: Should be unnecessary to add here if it has no custom register assignment.

src/coreclr/vm/greenthreads.cpp
```diff
@@ -54,6 +54,9 @@ extern "C" uintptr_t AllocateMoreStackHelper(int argumentStackSize, void* stackP
     uint8_t *newArgsLocation;
     if (argumentStackSize < 0)
     {
+        assert(t_greenThread.inGreenThread);
+        t_greenThread.inGreenThread = false;
```
markples, line 58: This pattern seems good for catching issues.  Perhaps use it every single time?  (even down below at line 231 which is 20 lines after the __debugbreak check, just in case, and again there when setting it back because the call could do something strange)

src/coreclr/vm/greenthreads.cpp
```diff
@@ -401,20 +404,69 @@ SuspendedGreenThread* GreenThread_ResumeThread(SuspendedGreenThread* pSuspendedT
...
-    if (t_greenThread.inGreenThread) // This should avoid doing this for the initial jump into the green thread and jumps from green to OS
+    if (t_greenThread.inGreenThread)
     {
         if (t_greenThread.pStackListCurrent->prev == NULL)
-            __debugbreak();
-        t_greenThread.pStackListCurrent = t_greenThread.pStackListCurrent->prev;
+        {
+            // We should only hit this path when a green thread is finishing
+            t_greenThread.inGreenThread = false;
+        }
```
markples, line 426: Do you want to keep the current pStackListCurrent value rather than null it out here?  (current = t = current->prev;  if (t == nullptr) inGT=false)?

davidwrighton: Its a bit ugly, but yes, I think I do. We use it as part of cleanup. :(

src/coreclr/jit/lower.cpp
```diff
@@ -2198,6 +2210,124 @@ GenTree* Lowering::LowerTailCallViaJitHelper(GenTreeCall* call, GenTree* callTar
...
+    {
+// Unlike CFG we need to handle integral constants
+#if 0
+        if (callTarget->IsIntegralConst())
+        {
+            // This is a direct call, no CFG check is necessary.
+            return;
+        }
+#endif
+    }
```
markples, line 2243: It's ok with me if you remove the parallels to CFG here (assert not null, delete the else)

src/coreclr/jit/lower.cpp
```diff
@@ -2198,6 +2210,124 @@ GenTree* Lowering::LowerTailCallViaJitHelper(GenTreeCall* call, GenTree* callTar
     return result;
 }
 
+//------------------------------------------------------------------------
+// LowerCFGCall: Potentially lower a call to use control-flow guard. This
```
markples, line 2214: copy pasta.  There's more in the internal comment, but it's probably not worth trying to rewrite it yet.

src/coreclr/vm/greenthreads.cpp
```diff
@@ -219,7 +228,10 @@ SuspendedGreenThread* GreenThread_StartThread(TakesOneParam functionToExecute, u
     GetThread()->SetActiveThreadBase(pGreenThread);
     pGreenThread->m_currentThreadObj = GetThread();
 
+    assert(t_greenThread.inGreenThread == false);
+    t_greenThread.greenThreadOnStack = true;
     GreenThread_StartThreadHelper((uintptr_t)FirstFrameInGreenThread, &detailsAboutWhatToCall);
+    t_greenThread.greenThreadOnStack = false;
```
markples, line 234: You tried but removed `assert(t_greenThread.inGreenThread == true);` here.  Is it always `false`?  Or is this a multiple exit conditions thing?

markples (approved):

src/coreclr/vm/greenthreads.cpp
davidwrighton: Its a multiple exit conditions thing. if inGreenThread, we're exiting due to a suspension, else we just finished the green thread without ever suspending.

src/coreclr/jit/lower.cpp
davidwrighton: Eh, it'll work for now.

Issue #2016, opened 2022-10-03
[Green Threads] - Integration into Task.Wait is probably wrong in various ways.


davidwrighton:
Known issue from https://github.com/dotnet/runtimelab/pull/2002

PR #2054, opened 2022-10-22
Flow the execution context into a green thread

kouvel:
Relevant to https://github.com/dotnet/runtimelab/issues/2016


kouvel:
There may be a few more things left to do regarding value change notifications during implicit transitions, I'll look into that separately

davidwrighton (approved):

PR #2056, opened 2022-10-26
New helper for green thread stack segments

davidwrighton:
Add new helper and update the various bits so that it can be referred to from the JIT. This will require managed functions to have a prolog like...

The expected prologs will look something like:
```
   lea         rax, [rsp-850h]
   cmp         rax, qword ptr gs:[10h]
   ja          prolog
   mov         eax, 50h
   call        JIT_GreenThreadMoreStack
prolog:
   push        rbp
   ...
```


src/coreclr/vm/amd64/AsmHelpers.asm
```diff
@@ -974,6 +983,18 @@ NESTED_ENTRY TransitionToOSThreadHelper2, _TEXT
     jmp rax
 NESTED_END TransitionToOSThreadHelper2, _TEXT
 
+; This function is called with a misaligned stack, and the amount of needed stack space in the RAX register  
+NESTED_ENTRY JIT_GreenThreadMoreStack, _TEXT
+    pop r11
```
markples:
```suggestion
    pop r11  ; explicitly passing the return address as a continuation in r11
```
(or something like this)

src/coreclr/vm/amd64/AsmHelpers.asm
```diff
@@ -721,6 +728,7 @@ NESTED_ENTRY _more_stack, _TEXT
         ;
         ; Save integer parameter registers.
         ;
+        save_reg_postrsp    r11, 68h
```
markples, line 731: This needs to be copied to ResumeSuspendedThreadHelper2 to keep them identical

markples (approved):

src/coreclr/vm/amd64/AsmHelpers.asm
davidwrighton: Eh, true, although it doesn't actually matter for correctness.

PR #2058, opened 2022-10-27
Update README.md

k3ldar:
remove duplicate wording


jkotas:
The NativeAOT experiment in this repo was completed and the branch is archived. We are not accepting any new PR into the NativeAOT branch.

The current version of NativeAOT lives in dotnet/runtime and other (non-experimental) repos.

Could you please submit this change against https://github.com/dotnet/samples/tree/main/core/nativeaot/NativeLibrary  instead? It is the new home of this sample.

PR #2059, opened 2022-10-27
allow TYP_STRUCT for GT_RETURN - special case CNS_INT 0 initialization

yowl:
This PR is pretty much just as per the title.  Added a case for `TYP_STRUCT` in `buildReturn` with special handling for 0 initialization.

Slight difference to the other location where we 0 initialize structs is that the size here is taken from the LLVM type not the `ClassLayout->GetSize` as we dont have a `GenTreeBlk`

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -1629,12 +1629,12 @@ void Llvm::buildReturn(GenTree* node)
             }
             if (node->gtGetOp1()->IsIntegralConst(0))
             {
-                // special case returning 0 initialized structs
+                // Special-case returning zero-initialized structs.
                 Type* structLlvmType = getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass);
                 Value* structAddrValue = _builder.CreateAlloca(structLlvmType, 0U);
-                _builder.CreateMemSet(structAddrValue, _builder.getInt8(0), _builder.getInt32(structLlvmType->getScalarSizeInBits() >> 3), {});
+                Value* structSizeValue = _builder.getInt32(structLlvmType->getScalarSizeInBits() / BITS_PER_BYTE);
+                _builder.CreateMemSet(structAddrValue, _builder.getInt8(0), structSizeValue, {});
                 _builder.CreateRet(_builder.CreateLoad(structAddrValue));
-
                 return;
             }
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -1620,12 +1620,23 @@ void Llvm::buildReturn(GenTree* node)
...
                 failFunctionCompilation();
             }
+            if (node->gtGetOp1()->IsIntegralConst(0))
+            {
+                // special case returning 0 initialized structs
+                Type* structLlvmType = getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass);
+                Value* structAddrValue = _builder.CreateAlloca(structLlvmType, 0U);
+                _builder.CreateMemSet(structAddrValue, _builder.getInt8(0), _builder.getInt32(structLlvmType->getScalarSizeInBits() >> 3), {});
+                _builder.CreateRet(_builder.CreateLoad(structAddrValue));
+
```
```suggestion
                // Special-case returning zero-initialized structs.
                Type* structLlvmType = getLlvmTypeForCorInfoType(_sigInfo.retType, _sigInfo.retTypeClass);
                Value* structAddrValue = _builder.CreateAlloca(structLlvmType, 0U);
                Value* structSizeValue = _builder.getInt32(structLlvmType->getScalarSizeInBits() / BITS_PER_BYTE);
                _builder.CreateMemSet(structAddrValue, _builder.getInt8(0), structSizeValue, {});
                _builder.CreateRet(_builder.CreateLoad(structAddrValue));
```

SingleAccretion (approved):
LGTM modulo nit.

src/coreclr/jit/llvm.cpp
yowl: Isn't that changing a division by 8 (going from bits to bytes) to a division by 4 (`TARGET_POINTER_SIZE` for wasm32).  If the `>> 3` is not great, we can just do `/ 8` to go from bits to bytes?

SingleAccretion: Ah, sorry, indeed it should be `/ BITS_PER_BYTE` (I edited the suggestion).

yowl:
oops.  Probably should only do the `memset` for `IsIntegralConst(0)` for `TYP_STRUCT`...

yowl:
Green now, @SingleAccretion when you have a moment, please take a look at the last commit to make sure we are happy with this.  Thanks!

SingleAccretion (approved):

jkotas:
Thank you!

PR #2060, opened 2022-10-28
Swap to OS thread for special cases

### LINKED ISSUE #2051: [Green Threads] Various helper calls do not transition to the OS thread
In particular, those that use PROLOG_WITH_TRANSITION_BLOCK and then just call a native function are in this category such as the Prestub, VSD resolution, and ComPlusCalls.

davidwrighton:
- where we don't use an FCall or standard P/Invoke transition to go to C++ code
- Do so by generalizing the C++ lambda technique used for the FCall transitions and make it available as a macro for these other transitions to use

Fixes #2051

### DIFF
--- a/src/coreclr/vm/comcallablewrapper.cpp
+++ b/src/coreclr/vm/comcallablewrapper.cpp
@@ -395,7 +395,10 @@ bool IsOleAutDispImplRequiredForClass(MethodTable *pClass)
 //--------------------------------------------------------------------------
 extern "C" PCODE ComPreStubWorker(ComPrestubMethodFrame *pPFrame, UINT64 *pErrorReturn)
 {
-    CONTRACT (PCODE)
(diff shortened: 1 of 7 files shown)


src/coreclr/vm/comcallablewrapper.cpp
```diff
@@ -602,8 +604,9 @@ extern "C" PCODE ComPreStubWorker(ComPrestubMethodFrame *pPFrame, UINT64 *pError
 Exit:
 
     END_ENTRYPOINT_VOIDRET;
-
-    RETURN retAddr;
+    RETURN;
+    END_ENSURE_ON_OS_THREAD();
+    return retAddr;
```
markples, line 609: Sorry for the late review - why does only this one have `RETURN` + `return`?

PR #2064, opened 2022-11-01
Add prolog stack check for green threads

markples:
Adds a notion of a "pre-prolog" to functions that does a bounds check on the stack and conditionally calls a helper to grow (copy) the stack region.  For now, this avoids allowing more than one prolog IG, but the handling is messy.  The pre-prolog must be added to place that require prolog-like processing.

Add support for encoding GS accesses - include the offset and avoid RIP-relative addressing.  (not sure if FS worked before)

Fix encoding of forward branches by N instructions.  (all previous jumps were backwards by N instructions for forward/backwards to a label)

Includes @davidwrighton's fix to hook up the new stack helper

TODO:
- (probably later - code sequence allows execution) Any external requirements (like unwind)
- funclets?


markples:
Replaced by #2065

PR #2067, opened 2022-11-07
Fix a few things regarding EC, await, and thread pool local queuing

kouvel:
- Modified to send EC value change notifications in some transitions between a green thread and the OS thread, as some of those notifications can take effect on the OS thread. Currently this includes exiting a green thread, yield, and resume. It currently doesn't cover transitions to the OS thread for fcalls/pinvokes.
- Updated to have await and task continuations run as a green thread when the previous task ran as a green thread. Added a green thread task scheduler. Currently it creates new green threads for continuations, which can probably be improved.
- Fixed local queuing to the thread pool while running as a green thread to actually queue locally. Replaced the thread pool thread-static field with a field on the thread object, along with adding a way to get the OS thread object.
- Disabled a few tests that appear to be crashing at various places frequently or occasionally, which seem unrelated to these changes

Closes https://github.com/dotnet/runtimelab/issues/2016, though there are probably things to improve.


kouvel:
cc @davidwrighton @stephentoub

davidwrighton:
Looks good. I've merged my stack fix, could you merge that in, and see if you can avoid disabling so many tests?

kouvel:
It looks like `executioncontext\FlowTest` is still crashing occasionally in Checked builds and more frequently in Debug builds on my machine, even without this change. I have enabled all of the other tests, which seem to be passing now.

kouvel:
Think I've found the issue, I'll put up a fix and reenable FlowTest after this is merged

kouvel:
Some of the new tests are failing after https://github.com/dotnet/runtimelab/pull/2079, I think I'll go ahead and merge this change since the failures are unrelated.

PR #2070, opened 2022-11-08
Fix stack corruption bug on green threads

davidwrighton:
- Caused by windows updating the stack limit in the TEB while expanding the OS stack.
- This isn't an ideal fix, but it seems to work for now


davidwrighton:
@kouvel, @cshung I think this is the issue which was breaking the flow test in recent builds.

kouvel (approved):
Nice!

Issue #2072, opened 2022-11-09
[NativeAOT-LLVM] Split RyuJit backend code into different files


SingleAccretion:
As the NativeAOT-LLVM's RyuJit backend grows, it will eventually become untenable for all of the code to live in one `llvm.cpp` file. I propose to split the code in the somewhat near future, to avoid having to do that down the road, when there will be more of it.

The proposed changes (and file names) are modeled against similar files for non-WASM backends:
1) General-purpose utilities specific to LLVM-based compilation (e. g. "does this local need to exist on the shadow stack"): stay in `llvm.cpp`, in the `Llvm` class (as it exists today).
2) Things related to the type system (type descriptors, field descriptors, `XYZtoLLVMtype` methods): move to `llvmtypes.cpp`, (but stay part of the `Llvm` class, possibly as `static` methods)
3) Lowering: `lowerllvm.cpp`, in a new `LowerLlvm` class.
4) Codegen: `codegenllvm.cpp`, in a new `CodegenLlvm` class.

I think it is reasonable for the header file to remain singular (i. e. `llvm.h`) for now (it will be straightforward to split it in the future if needed).

Also, it could be better to name lower/codegen files `llvmlower.cpp`/`llvmcodegen.cpp`, so that everything shares a common prefix (I don't have a strong opinion on this).

@yowl thoughts? I will do this if you agree it is a good idea (which it may not necessarily be, considering the risk of conflicts etc).

yowl:
I like this.  Prefix of `llvm` for the files would be my preference.

SingleAccretion:
I will consider this done for now. Initially I wanted to split lowering and codegen into different classes, but will defer that for the time being. Doing it will not be as disruptive as moving code across files.

PR #2077, opened 2022-11-10
[NativeAOT-LLVM] Split RyuJit backend into multiple files

SingleAccretion:
First part of #2072:

1) Group functions in `llvm.h`.
2) Move them to different files (with declaration order same as implementation order).
3) Enclose all of the custom Jit-EE interface functions in wrappers for easier future refactorings (inevitable if we want to support parallel compilation).
4) Minor dead code cleanup / comment fixes.

Since this is already pretty much irreviewable, the class split was left for a different change.

cc @yowl


src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -0,0 +1,1714 @@
...
+    if (isLlvmFrameLocal(varDsc))
+    {
+        llvmRef = _builder.CreateLoad(m_allocas[lclNum]);
+    }
+    else if (_localsMap->find({lclNum, ssaNum}) == _localsMap->end())
+    {
+        if (varDsc->lvLlvmArgNum != BAD_LLVM_ARG_NUM)
+        {
+            // TODO-LLVM: we dont handle this yet
+            if (_info.compRetBuffArg != BAD_VAR_NUM)
```
SingleAccretion, line 454: By the way, all of the handling of return buffers can just be deleted, since we have a purely "by-value" (to the frontend) ABI now.

yowl:
Thanks, this looks much better than before.

PR #2078, opened 2022-11-11
Comment out the new frame in FirstFrameInGreenThreadCpp for now

kouvel:
- Yields appear to be failing (returning false) with that frame, seems to have started happening after be2df4a1e76cc1e44d9b1b702812232145982edd
- Commented out the push/pop of the new frame for now


kouvel:
CC @cshung @davidwrighton

davidwrighton (approved):

Issue #1992, opened 2022-09-14
[NativeAOT-LLVM, WASM] Disable Reflection, IL scanner, RuntimeError: memory access out of bounds


Kanawanagasaki:
Hi, I was experimenting with compiling dotnet console application to WebAssembly. When disable reflection by adding `<IlcDisableReflection>true</IlcDisableReflection>` in .csproj i got `ILCompiler.ScannerFailedException`:
```
ILCompiler.ScannerFailedException: VTable of type 'System.Runtime.InteropServices.DllImportSearchPath' not computed by the IL scanner. You can work around by running the compilation with scanner disabled.
     at ILCompiler.ILScanResults.ScannedVTableProvider.GetSlice(TypeDesc type)
     at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
     at ILCompiler.DependencyAnalysis.ConstructedEETypeNode.ComputeNonRelocationBasedDependencies(NodeFactory factory)
     at ILCompiler.DependencyAnalysis.ObjectNode.GetStaticDependencies(NodeFactory factory)
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.GetStaticDependenciesImpl(DependencyNodeCore`1 node)
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.GetStaticDependencies(DependencyNodeCore`1 node)
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.ProcessMarkStack()
     at ILCompiler.DependencyAnalysisFramework.DependencyAnalyzer`2.ComputeMarkedNodes()
     at ILCompiler.LLVMCodegenCompilation.CompileInternal(String outputFile, ObjectDumper dumper)
     at ILCompiler.Compilation.ILCompiler.ICompilation.Compile(String outputFile, ObjectDumper dumper)
     at ILCompiler.Program.Run(String[] args)
     at ILCompiler.Program.Main(String[] args)
```
The exception message prompts me to disable the scanner, but I don't know how.
While trying to add various properties to the .csproj, I found that with  `<InvariantGlobalization>true</InvariantGlobalization>` property console application would compile successfully. But this time after i run it with node or in browser it will throw `RuntimeError: memory access out of bounds` Exception:
```
RuntimeError: memory access out of bounds
[...]

PR #2081, opened 2022-11-13
NativeAOT-LLVM: add a dependency for the constructed eetype for newobj

yowl:
This PR adds a dependency for the fully constructed EE type for a call to newobj.  Fixes the reflection free mode and #1992

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
@@ -1909,16 +1909,9 @@ namespace Internal.IL
                         {
                             typeToAlloc = callee.OwningType;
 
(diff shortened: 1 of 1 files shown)


SingleAccretion (approved):

yowl:
Thanks, I've removed the other symbol and just left the constructed eetype.  This is actually how the code was before I put in the ILScanner back in February.  Guess I just got that wrong at the time.

MichalStrehovsky (approved):
Looks good! Yes, we should definitely be getting the version of EEType that has a working vtable for newobj and not the one used for casting/other non-creation operations.

(Sorry I didn't respond on the issue - I don't look at work email on weekends these days.)

yowl:
> (Sorry I didn't respond on the issue - I don't look at work email on weekends these days.)

No need to apologize, no work at the weekend is best.

jkotas (approved):

PR #2084, opened 2022-11-14
[NativeAOT-LLVM] Enable `lvaTableDump` for WASM

SingleAccretion:
And clean up the `target.h` situation as well (i. e. delete everything but the bare necessities).

We will now dump the shadow stack layout after building it:
```scala
Locals after shadow stack layout:
; Tentative local variable assignments
;
;* V00 arg0         [V00    ] (  0,  0   )     int  ->  zero-ref    class-hnd single-def
;  V01 arg1         [V01,T04] (  1,  0.50)     ref  ->  [SS+00H]   class-hnd single-def
;  V02 arg2         [V02,T03] (  1,  1   )     ref  ->  [SS+04H]   class-hnd single-def
;  V03 arg3         [V03,T02] (  2,  1.50)     ref  ->  [SS+08H]   class-hnd single-def
;  V04 tmp0         [V04,T01] (  3,  2   )     int  ->  [SS+00H]
;  V05 tmp1         [V05,T00] (  3,  6   )     ref  ->  [SS+0CH]   class-hnd exact "Single-def Box Helper"
;  V06 rat0         [V06    ] (  1,  1   )     int  ->  [SS+00H]   "shadowstack"
;  V07 rat1         [V07    ] (  1,  1   )     int  ->  [SS+00H]   "returnslot"
```
(One can notice a couple locals there don't actually go on the shadow stack, but it is good enough for now)


SingleAccretion:
cc @yowl

yowl:
Nice!

Issue #2073, opened 2022-11-09
[Bug] [NativeAOT-LLVM]: Unable to compile native WebAssembly library


just-ero:
## Description
Attempted compilation of a native WASM library aborts with `System.AccessViolationException`:
```
Determining projects to restore...
  All projects are up-to-date for restore.
  NativeWasm -> ~\bin\Debug\net7.0\browser-wasm\NativeWasm.dll
  Generating compatible native code.

/_/src/libraries/System.Private.CoreLib/src/System/Resources/ManifestBasedResourceGroveler.cs(238): Trim analysis warning IL2026: System.Resources.ManifestBasedResourceGroveler.CreateResourceSet(Stream,Assembly): Using method 'System.Resources.ManifestBasedResourceGroveler.InternalGetResourceSetFromSerializedData(Stream,String,String,ResourceManagerMediator)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The CustomResourceTypesSupport feature switch has been enabled for this app which is being trimmed. Custom readers as well as custom objects on the resources file are not observable by the trimmer and so required assemblies, types and members may be removed.

  Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
  Repeat 2 times:
  --------------------------------
     at LLVMSharp.Interop.LLVM.BuildLoad(LLVMSharp.Interop.LLVMOpaqueBuilder*, LLVMSharp.Interop.LLVMOpaqueValue*, SByte*)
  --------------------------------
     at LLVMSharp.Interop.LLVMBuilderRef.BuildLoad(LLVMSharp.Interop.LLVMValueRef, System.ReadOnlySpan`1<Char>)
     at Internal.IL.ILImporter.LoadAddressOfSymbolNode(ILCompiler.DependencyAnalysis.ISymbolNode, LLVMSharp.Interop.LLVMBuilderRef)
     at Internal.IL.ILImporter.GetFieldAddress(Internal.TypeSystem.FieldDesc, Internal.TypeSystem.FieldDesc, Boolean)
     at Internal.IL.ILImporter.ImportLoadField(Int32, Boolean)
     at Internal.IL.ILImporter.ImportBasicBlock(BasicBlock)
     at Internal.IL.ILImporter.ImportBasicBlocks()
     at Internal.IL.ILImporter.Import()
[...]

PR #2087, opened 2022-11-14
NatvieAOT-LLVM: Update the image for prepare-signed-artifacts.yml

yowl:
This PR takes the change from upstream/runtime for the image used for publishing the nuget packages.  Should unblock #2073.

cc @SingleAccretion
cc @jkotas


jkotas:
Fix this one too? https://github.com/dotnet/runtimelab/blob/06fe8e962060e318ea16986c09a331d3d6896ed7/eng/pipelines/official/stages/publish.yml#L23

yowl:
Thanks, didn't see it for the different case...

PR #2091, opened 2022-11-16
[NativeAOT-LLVM] Initialize locals in codegen

SingleAccretion:
When we have `opts.compInitMem` set to `true` (i. e. the default for C# without `SkipLocalsInit`), we need to zero-initialize non-temps. We also need to set the initial SSA values for use-before-def locals, and allocas, as otherwise we run into UB issues with LLVM.

This change fixes the above, modulo the requirement to zero-init locals with GC pointers, which will have to be done in lowering.

In doing this, it also cleans up the LLVM type system management a little and generalizes `buildAdd/buildCmp`, which gets us up to 70% coverage for the HellowWasm test.

Example C#/LLVM:
```cs
private static double Problem()
{
    double a = *&a;
    double b = *&b;
    JitUse(&b);

    return a;
}
```
```llvm
define double @RyuJitReproduction_RyuJitReproduction_Program__Problem(i8* %0) {
Prolog:
  %1 = freeze double undef
  %2 = alloca double, align 8
  store double 0.000000e+00, double* %2, align 8
  br label %BB01

BB01:                                             ; preds = %Prolog
  %3 = load double, double* %2, align 8
  store double %3, double* %2, align 8
  %4 = getelementptr i8, i8* %0, i32 0
  call void @"RyuJitReproduction_RyuJitReproduction_Program__JitUse_0<Double>"(i8* %4, double* %2)
  ret double 0.000000e+00
}
```

### DIFF
--- a/src/coreclr/jit/llvmtypes.cpp
+++ b/src/coreclr/jit/llvmtypes.cpp
@@ -233,11 +233,14 @@ Type* Llvm::getLlvmTypeForLclVar(LclVarDsc* varDsc)
     {
         return getLlvmTypeForStruct(varDsc->GetLayout());
     }
-    if (varDsc->TypeGet() == TYP_BLK)
-    {
-        assert(varDsc->lvExactSize != 0);
-        return llvm::ArrayType::get(Type::getInt8Ty(_llvmContext), varDsc->lvExactSize);
-    }
+    // TODO-LLVM: enable. Currently broken because RyuJit inserts RPI helpers for RPI methods,
+    // with a TYP_BLK frame variable, and then we also create an RPI wrapper stub, resulting
+    // in a double transition.
+    // if (varDsc->TypeGet() == TYP_BLK)
+    // {
+    //     assert(varDsc->lvExactSize != 0);
+    //     return llvm::ArrayType::get(Type::getInt8Ty(_llvmContext), varDsc->lvExactSize);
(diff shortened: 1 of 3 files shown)


src/coreclr/jit/llvmtypes.cpp
```diff
@@ -212,81 +212,59 @@ Type* Llvm::getLlvmTypeForVarType(var_types type)
...
+Type* Llvm::getLlvmTypeForLclVar(LclVarDsc* varDsc)
 {
-    var_types nodeType = lclVar->TypeGet();
-
-    if (nodeType == TYP_STRUCT)
+    if (varDsc->TypeGet() == TYP_STRUCT)
+    {
+        return getLlvmTypeForStruct(varDsc->GetLayout());
+    }
+    // TODO-LLVM: enable. Currently broken because RyuJit inserts RPI helpers for RPI methods,
```
SingleAccretion, line 236: Presumably this is not a difficult fix, but it'd require some VM changes to not emit thunks but still have the right mangling for the export.

SingleAccretion:
@yowl

yowl:
LGTM, although it's a bit over my pay grade!  I like that you got rid of the ugly `getLlvmArgInfoForArgIx`.

jkotas (approved):
Thank you!

PR #2093, opened 2022-11-17
[NativeAOT-LLMV] Implement `GT_LCLHEAP` codegen

SingleAccretion:
I. e. `stackalloc` in C# or `localloc` in IL. Reference implementation can be found in e. g. `codegenxarch.cpp, genLclHeap`. The tricky bit is ensuring we return a null pointer for zero-sized allocations.

Example C#/codegen:
```cs
private static byte Problem(int size)
{
    var p = stackalloc byte[size];
    return *p;
}
```
```llvm
=============== Generating BB01 [000..006) (return), preds={} succs={} flags=0x00000000.20000020: i LIR
Generating:                [000004] ------------                 IL_OFFSET void   IL offset: 0x0
Generating: N001 (  3,  2) [000000] ------------         t0 =    LCL_VAR   int    V00 arg0         u:1 (last use) $80
                                                              /--*  t0     int
Generating: N002 (  4,  3) [000001] ---X---N----         t1 = *  LCLHEAP   int    $81
  %4 = alloca i8, i32 %1, align 8
  call void @llvm.memset.p0i8.i32(i8* align 8 %4, i8 0, i32 %1, i1 false)
  %5 = icmp ne i32 %1, 0
  %6 = select i1 %5, i8* %4, i8* null
                                                              /--*  t1     int
Generating: N003 (  8,  6) [000002] *--XG-------         t2 = *  IND       ubyte  <l:$1c1, c:$1c0>
  %7 = load i8, i8* %6, align 1
                                                              /--*  t2     ubyte
Generating: N004 (  9,  7) [000003] ---XG-------              *  RETURN    int    $82
  ret i8 %7
```

cc @yowl

### DIFF
--- a/src/coreclr/jit/llvm.h
+++ b/src/coreclr/jit/llvm.h
@@ -228,6 +228,7 @@ private:
     void buildAdd(GenTreeOp* node);
     void buildDiv(GenTree* node);
     void buildCast(GenTreeCast* cast);
+    void buildLclHeap(GenTreeUnOp* lclHeap);
     void buildCmp(GenTreeOp* node);
     void buildCnsDouble(GenTreeDblCon* node);
     void buildCnsInt(GenTree* node);
--- a/src/coreclr/jit/llvmcodegen.cpp
+++ b/src/coreclr/jit/llvmcodegen.cpp
@@ -778,7 +778,7 @@ void Llvm::buildCmp(GenTreeOp* node)
                                    : (isUnordered ? Predicate::FCMP_UGE : Predicate::FCMP_OGE);
             break;
         case GT_GT:
-            predicate = isIntOrPtr ? (isUnordered ? Predicate::ICMP_UGT : Predicate::ICMP_SGT)
+            predicate = isIntOrPtr ? (isUnsigned ? Predicate::ICMP_UGT : Predicate::ICMP_SGT)
                                    : (isUnordered ? Predicate::FCMP_UGT : Predicate::FCMP_OGT);
             break;
         default:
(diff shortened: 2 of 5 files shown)


yowl:
Thanks, looks good.  

I see you have a problem with `GetNonCryptographicallySecureRandomBytes` using the undefined `dotnet_browser_entropy`.  I'm not sure why that is becoming a problem with this change.  Would be nice from a WASI point of view if we could use a WASI call for that eventually so as not to introduce browser dependencies, but I'm not sure where we are with enabling WASI in emscripten, nor where emscripten itself is.

`Interop.GetRandomBytes` is "handled" with a stub:
```
if ((method.OwningType as EcmaType)?.Name == "Interop" && method.Name == "GetRandomBytes")
{
    // this would normally fill the buffer parameter, but we'll just leave the buffer as is and that will be our "random" data for now
    return new ILStubMethodIL(method, new byte[] { (byte)ILOpcode.ret }, Array.Empty<LocalVariableDefinition>(), null);
}
```
on the IL side.

SingleAccretion:
Indeed, something is off with the handling here. Theory number one is that RyuJit inlined `Interop.GetRandomBytes` and that broke things.

SingleAccretion:
> Theory number one is that RyuJit inlined `Interop.GetRandomBytes` and that broke things.

Looks to me that was indeed the case.

jkotas:
@yowl Does this look good to you?

yowl:
@jkotas Yes, thanks, absolutely.

jkotas:
@yowl I have sent you invitation for dotnet org so that I can just ask you for codereview via github UI next time.

yowl:
@jkotas Thanks, it looks like it got cancelled

"[GitHub] Your invitation to dotnet has been canceled"

and the link:
image

jkotas:
Hmm, I will try to find out what happened there.

jkotas:
You should have a new invitation in your inbox now.

yowl:
Thanks, worked.

PR #2094, opened 2022-11-18
Add deep stack test, reduce standard stack depth, and fix the GC stackwalking problems

davidwrighton:
- Reduce the minimum stack size for a green thread stack segment
  - This caused broken stackwalking to become evident, so I then fixed THAT
  - 
- Add new concept for stack pointer comparison (courtesy of @cshung)
- Always unwind through _more_stack and JIT_GreenThreadMoreStack
 - To simplify the implementation, I've disabled QuickUnwind (I'm not sure if this is necessary, but it appears like it might be)
 - This allows the stackwalker to ignore the existence of these frames
- DACize a bunch of the green thread data structures so that the updates to the stackwalker work

### DIFF
--- a/src/coreclr/vm/amd64/AsmHelpers.asm
+++ b/src/coreclr/vm/amd64/AsmHelpers.asm
@@ -871,7 +871,7 @@ NESTED_END TransitionToOSThreadHelper, _TEXT
 ; onwards in execution with the green thread suspended. When the thread is resumed, it will jump to resume_point.
 NESTED_ENTRY YieldOutOfGreenThreadHelper, _TEXT
     PROLOG_WITH_TRANSITION_BLOCK
-    sub rdx, 0e8h         ; This should be the RSP before we did the transition to the OS thread
+    add rdx, 0e0h         ; This should be the RSP before we did the transition to the OS thread
     mov rbp, rdx          ; Set RBP to what it was before we transitioned to the green thread
 
     mov rdx, [rcx]        ; Get the latest OS thread stack limit from t_greenThread.osStackRange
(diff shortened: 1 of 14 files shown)


src/coreclr/vm/amd64/AsmHelpers.asm
```diff
@@ -1019,6 +1021,8 @@ JumpToMoreStack:
     add rsp, 28h
     ret
     jmp r11
+ALTERNATE_ENTRY JIT_GreenThreadMoreStack_end ; When yielding, we will jmp back to this location with the registers back in their original config
```
markples: This comment is a copy/paste mistake.  (or I'm very confused about how this works)

src/coreclr/vm/greenthreads.cpp
```diff
@@ -16,18 +16,114 @@ FCIMPL0(void, JIT_GreenThreadMoreStack)
...
+    {
+        b2 = blockNumber;
+    }
+    if (b1 == b2)
+    {
+        return sp1 < sp2;
+    }
+    else
+    {
+        // The smaller the block number, the earlier it was created, the larger it is "logically"
```
markples, line 108: I think you could
- remove the pre-walk through `prev`
- start blockNumber at -1 and decrement it
- iterate through `prev` instead of `next`
- flip the direction on this `b2 < b1`

I bring it up because that's what I spent my time on reviewing this, but also it wasn't that hard either.

markples (approved):

src/coreclr/vm/amd64/AsmHelpers.asm
davidwrighton: Very much a copy past mistake.

PR #2098, opened 2022-11-19
[NativeAOT-LLVM] Expand support for indirections

SingleAccretion:
1) Implement null checking for all types of indirs in codegen.
2) Implement complex `initblk`s, as well as simple `BLK`s.
3) Implement retyping for struct stores, including shadow stack ones.
4) Also remove zero-sized block stores in lowering.

### DIFF
--- a/src/coreclr/jit/llvmlower.cpp
+++ b/src/coreclr/jit/llvmlower.cpp
@@ -378,6 +378,7 @@ void Llvm::ConvertShadowStackLocalNode(GenTreeLclVarCommon* node)
         {
             node->ChangeOper(indirOper);
             node->AsIndir()->SetAddr(lclAddress);
+            node->gtFlags |= GTF_IND_NONFAULTING;
         }
         if (GenTree::OperIsStore(indirOper))
         {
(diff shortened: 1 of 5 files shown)


SingleAccretion:
@yowl

src/coreclr/jit/llvmlower.cpp
```diff
@@ -635,7 +662,9 @@ GenTreeCall::Use* Llvm::lowerCallReturn(GenTreeCall*      callNode,
         {
             indirNode = _compiler->gtNewIndir(callReturnType, returnAddrLclAfterCall);
         }
-        indirNode->gtFlags |= GTF_IND_TGT_NOT_HEAP; // No RhpAssignRef required
```
yowl, line 638: can you explain why we dont need this anymore, thanks?

yowl:
Looks great.  You are really pulling this thing into shape.   :-)

src/coreclr/jit/llvmlower.cpp
SingleAccretion: The flag is only relevant for stores i. e. `STOREIND`. This is a use, so it has not effect.

yowl: oh yes, thanks

PR #2100, opened 2022-11-21
[NativeAOT-LLVM] Support some throwing operators via throw helper blocks

### LINKED ISSUE #2061: Question: NativeAOT-LLVM: GT_ARR_LENGTH semantics
Is there any difference between `GT_IND` and `GT_ARR_LENGTH` for LLVM?  If I remember rightly, the length is in the first 4 bytes so a `GT_IND` for an `int` should get the length.  Or does GT_ARR_LENGTH imply something additional, e.g. a null ref check?

SingleAccretion:
With this change, we begin to support various throwing (casts, range checks, division, remainder) operators in codegen, under the "throw helper blocks" scheme, similar to what other backends do for optimized code. Inlined throws for the unoptimized case are not yet implemented, as are checked addition / subtraction / multiplication and null checks.

Closes #2061 (I decided to just do the simple thing and enable simple lowering; it will also provide MD array operator support once more of upstream has been merged).

Example IR for `checked((byte)floatValue)`:
```llvm
define i8 @RyuJitReproduction_RyuJitReproduction_Program__Problem(i8* %0, float %1) {
Prolog:
  br label %BB01.1

BB01.1:                                           ; preds = %Prolog
  %2 = fcmp ule float %1, -1.000000e+00
  %3 = fcmp uge float %1, 2.560000e+02
  %4 = or i1 %2, %3
  br i1 %4, label %BB02, label %BB01.2

BB01.2:                                           ; preds = %BB01.1
  %5 = fptoui float %1 to i8
  ret i8 %5

BB02:                                             ; preds = %BB01.1
  call void @S_P_CoreLib_Internal_Runtime_CompilerHelpers_ThrowHelpers__ThrowOverflowException(i8* %0)
  unreachable
}
```
For `a % b`:
```llvm
define i32 @RyuJitReproduction_RyuJitReproduction_Program__Problem(i8* %0, i32 %1, i32 %2) {
Prolog:
  br label %BB01.1

BB01.1:                                           ; preds = %Prolog
  %3 = icmp eq i32 %2, 0
  br i1 %3, label %BB02, label %BB01.2

BB01.2:                                           ; preds = %BB01.1
  %4 = icmp eq i32 %2, -1
  %5 = icmp eq i32 %1, -2147483648
  %6 = and i1 %4, %5
  br i1 %6, label %BB03, label %BB01.3

BB01.3:                                           ; preds = %BB01.2
  %7 = srem i32 %1, %2
  ret i32 %7

BB02:                                             ; preds = %BB01.1
  store i8* %0, i8** @t_pShadowStackTop, align 4
  call void @S_P_CoreLib_Internal_Runtime_CompilerHelpers_ThrowHelpers__ThrowDivideByZeroException()
  unreachable

BB03:                                             ; preds = %BB01.2
  call void @S_P_CoreLib_Internal_Runtime_CompilerHelpers_ThrowHelpers__ThrowOverflowException(i8* %0)
  unreachable
}
```

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -5190,6 +5190,22 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
     Rationalizer rat(this); // PHASE_RATIONALIZE
     rat.Run();
 
+    // Here we do "simple lowering".  When the RyuJIT backend works for all
+    // platforms, this will be part of the more general lowering phase.  For now, though, we do a separate
+    // pass of "final lowering."  We must do this before (final) liveness analysis, because this creates
+    // range check throw blocks, in which the liveness must be correct.
+    //
+    DoPhase(this, PHASE_SIMPLE_LOWERING, &Compiler::fgSimpleLowering);
+
+#ifdef DEBUG
+    fgDebugCheckBBlist();
+    fgDebugCheckLinks();
+#endif
+
+    // Enable this to gather statistical data such as
+    // call and register argument info, flowgraph and loop info, etc.
+    compJitStats();
+
 #if defined(TARGET_WASM)
     if (opts.OptimizationEnabled())
     {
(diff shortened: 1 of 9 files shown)


SingleAccretion:
@yowl

yowl:
Looks good, thanks!

yowl:
BTW, I was joking about `GT_SUB` :-).  I'm going to be merging for quite a while, so feel free if you want to do more math ops.

SingleAccretion:
Hehe :). I am researching the EH situation right now, so next on my immediate list are things related to enabling it (invokes, funclets and the corresponding LLVM).

yowl:
Ok, that's cool.  There is a bug (well at least 2) I believe in the current emscripten exception implementation.  If there is a nested handler call, the shadow stack offset that is used in the inner handler for locals, is the same as the outer handler and can overwrite shadowstack locals.  I believe I fixed that in https://github.com/yowl/runtimelab/blob/uo-loop-test/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs by adding another parameter (usedSS) to the second pass handler, or something like that.  Never got round to merging that bit as it was mixed up in other quick fixes trying to make the GC stable.

jkotas (approved):
Thank you!

PR #2103, opened 2022-11-28
[NativeAOT-LLVM] Unify call lowering and codegen; support inline throw helpers

SingleAccretion:
A number of changes, but the essence is simple:

1) Define signatures for helpers. Requires for multiple compilations to agree with each other on the finer details, such as whether we have a pointer, byref, integer, etc. Includes an extension point ("flags") in case we need more info in the future.
2) Unify codegen and lowering for user calls and helper calls using this.
3) Implement general-purpose helper emission facility in codegen and use it to implement inline throw helpers (used for debug codegen); implement NRE throwing as well as write barriers on top of this. Some light scanner fixups to accompany this.


SingleAccretion:
@yowl we should wait for #2101 before merging this, but it is otherwise ready for review.

src/coreclr/jit/llvm.cpp
```diff
@@ -174,7 +210,7 @@ CorInfoType Llvm::getCorInfoTypeForArg(CORINFO_SIG_INFO* sigInfo, CORINFO_ARG_LI
 
 // When looking at a sigInfo from eeGetMethodSig we have CorInfoType(s) but when looking at lclVars we have LclVarDsc or var_type(s),
 // This method exists to allow both to map to LLVM types.
-CorInfoType Llvm::toCorInfoType(var_types varType)
+/* static */ CorInfoType Llvm::toCorInfoType(var_types varType)
```
yowl, line 213: Is `static` a desired modifier that we can't take right now?

yowl (approved):
LGTM, just the minor question on the commented `/* static */`

src/coreclr/jit/llvm.cpp
SingleAccretion: No, it's just for documentation purposes. `static` on a member function definition is not valid in C++.

<details>
<summary>Styles of "static" documentation</summary>

There are generally three ways definitions of `static` member functions are annotated:
1\) Not at all.
2\)
```
// static
void Class::Func()
```
3\)
```
/* static */ void Class::Func()
```
The Jit uses both `2` and `3`, I just prefer `3` because it looks better in my opinion.
</details>

jkotas (approved):
Thank you!

PR #2105, opened 2022-11-29
Add some perf tests and don't spin-wait on a task when on a green thread

kouvel:
- The spin-waiting on tasks is tuned for thread context switches. Yield/Resume are much faster, updated to skip spin-waiting when on a green thread.
- Added a makeshift way to reuse green threads. For now, reusable green threads don't go away. Set `DOTNET_GreenThreads_DontReuse=1` to not reuse green threads.
- Added some small perf tests. They get built as part of the runtime tests and don't run in the CI, pass "run" as the argument to manually run them.


kouvel:
@cshung, @davidwrighton, or @mangod9, would one of you be able to review this PR?

davidwrighton (approved):
Nice.

PR #2111, opened 2022-12-12
Merge to last NativeAot commit

yowl:
This merges up to the last NativeAot commit in runtimelab, with hash `04b092003db5ba207d4fa3f3becb7f01828bf16c `

Not too much of interest here, except perhaps I reinstated the `ClrNativeAot` check in `externals.csproj` to stop it checking for the IL CoreClr files - as they are not built.

cc @SingleAccretion 

I added the subset `clr.nativeaotlibs` because the new project `ILCompiler.TypeSystem.Tests` fails without it
```
CSC : error CS2001: Source file 'E:\GitHub\runtimelab2\artifacts\obj\\coreclr\windows.x64.Debug\nativeaot\Runtime\Full\AsmOffsets.cs' could not be found. [E:\GitHub\runtimelab2\src\coreclr\nativeaot\Test.CoreLib\src\Test.CoreLib.csproj]
```

From here, I will try a commit from runtime, maybe just a day or two after this commit


yowl:
The OSX build is failing with a Fiji time zone problem it seems
```
 [FAIL] System.Tests.TimeZoneInfoTests.FijiTimeZoneTest
  Assert.Equal() Failure
  Expected: 13:00:00
  Actual:   12:00:00
     at System.Tests.TimeZoneInfoTests.FijiTimeZoneTest() + 0x32f
     at System.Runtime!<BaseAddress>+0x10f09b1
     at System.InvokeUtils.CallDynamicInvokeMethod(Object, IntPtr, IntPtr, IntPtr, Object, Object[], BinderBundle, Boolean, Boolean) + 0x163

```

A known issue perhaps, https://github.com/dotnet/runtime/issues/69801.  I could try merging in https://github.com/dotnet/runtime/pull/70309   @jkotas  do you want me to try that?

jkotas:
> I could try merging in https://github.com/dotnet/runtime/pull/70309 @jkotas do you want me to try that?

Yes, it would be nice to merge it so that we do not have to look at the failure log every time.

eng/pipelines/common/global-build-job.yml
```diff
@@ -160,9 +159,6 @@ jobs:
         displayName: Install/build LLVM
 
     - ${{ if eq(parameters.nameSuffix, 'Browser_wasm_Windows') }}:
-=======
-    - ${{ if eq(parameters.platform, 'Browser_wasm_win') }}:
->>>>>>> origin/feature/NativeAOT
```
SingleAccretion: Hmm, is this right? There is no `Browser_wasm_Windows` platform that I can see, only `Browser_wasm_win`.

Should it be the upstream's version, perhaps with an exclusion for not-Mono?

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -14,20 +14,15 @@ parameters:
 steps:
 - ${{ if and(ne(parameters.archType, 'arm'), ne(parameters.archType, 'arm64')) }}:
 
-<<<<<<< HEAD
   # For NativeAOT-LLVM, we have just built the Wasm-targeting native artifacts (the runtime and libraries).
   # Now we need to build the cross-targeting compilers, RyuJit and ILC.
 
   - ${{ if eq(parameters.archType, 'wasm') }}:
-    - script: $(Build.SourcesDirectory)/build$(scriptExt) clr.alljits+clr.wasmjit+clr.tools -c $(buildConfigUpper)
+    - script: $(Build.SourcesDirectory)/build$(scriptExt) clr.alljits+clr.wasmjit+clr.nativeaotlibs+clr.tools -c $(buildConfigUpper)
```
SingleAccretion, line 21: Should this be mentioned in the documentation too?

It is a bit of a shame one needs to build the NAOT libs now for a test project to compile, though I suppose it doesn't take that much time.

yowl: Indeed it is a shame.  Doc updated.

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -55,7 +55,7 @@ For the compilers:
...
 - Build LLVM either from the command line (`cmake --build . --target LLVMCore LLVMBitWriter`) or from VS 2022. Currently the Jit depends only on the output of LLVMCore and LLVMBitWriter projects.  For the Release configuration, `cmake --build . --config Release  --target LLVMCore LLVMBitWriter`
 - Set the enviroment variable `LLVM_CMAKE_CONFIG` to locate the LLVM config: `set LLVM_CMAKE_CONFIG=path-to-the-build-directory/lib/cmake/llvm`. This location should contain the file `LLVMConfig.cmake`.
-- Build the Jits and the ILC: `build clr.jit+clr.wasmjit+clr.tools -c [Debug|Release]`. Note that `clr.jit` only needs to be built once.  Add the `clr.nativeaotlibs+libs` subsets if you want the packages for publishing, e.g. `build clr.jit+clr.wasmjit+clr.tools+clr.nativeaotlibs+libs -c Debug`
+- Build the Jits and the ILC: `build clr.jit+clr.wasmjit+clr.tools+clr.nativeaotlibs -c [Debug|Release]`. Note that `clr.jit` only needs to be built once.  Add the `libs` subsets if you want the packages for publishing, e.g. `build clr.jit+clr.wasmjit+clr.tools+clr.nativeaotlibs+libs -c Debug`
```
```suggestion
- Build the Jits and the ILC: `build clr.jit+clr.wasmjit+clr.nativeaotlibs+clr.tools -c [Debug|Release]`. Note that `clr.jit` only needs to be built once.  Add the `libs` subsets if you want the packages for publishing, e.g. `build clr.jit+clr.wasmjit+clr.nativeaotlibs+clr.tools+libs -c Debug`.
```
Presumably the order is important.

eng/pipelines/common/global-build-job.yml
yowl: I'll take upstream, we needed this certificate update some time back.  It may be ok without it now.  I think at the time, mono took what we did and used the same.

docs/workflow/building/coreclr/nativeaot.md
jkotas: The order of subsets on the build command line is not important. The subsets are mapped to what to build in https://github.com/dotnet/runtime/blob/1529058b7e6b7a13253278f45e28f73c04149896/eng/Subsets.props#L336-L370 . This mapping will create the right order.

SingleAccretion: I see. I suppose there is only aesthetic benefit to having dependencies come before dependents then.

yowl:
Failing with the ILLink problem, as there's a lot of changes I will give it another go.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -213,17 +213,9 @@ The .NET Foundation licenses this file to you under the MIT license.
     <MSBuild Projects="@(ProjectToBuild)" BuildInParallel="true" />
   </Target>
 
-<<<<<<< HEAD
-  <!-- NativeAOT-LLVM note: 'BeforeTargets="_RunILLink"' should be deleted once https://github.com/dotnet/runtime/pull/63914 is merged from upstream -->
-  <Target Name="IlcCompile"
-      Inputs="@(IlcCompileInput);@(RdXmlFile)"
-      Outputs="$(NativeIntermediateOutputPath)%(ManagedBinary.Filename)$(IlcOutputFileExt)"
-      BeforeTargets="_RunILLink"
```
SingleAccretion, line 212: We may need to bring this `BeforeTargets="_RunILLink"` back and reword the `TODO` to say `once upstream uses .NET 7 SDK`.

PR #2113, opened 2022-12-13
NativeAOT-LLVM : Fix assert when GT_FIELD_LIST has > 1 member and is emitted as an LLVM primitive

### LINKED ISSUE #2112: NativeAOT-LLVM: compiler crashes for some struct optimisations
When compiling `S_P_CoreLib_System_Collections_Generic_List_1<S_P_CoreLib_System_ValueTuple_2<Char__Char>>__Contains` with `-O` `Ilc` hits the assert at 

https://github.com/dotnet/runtimelab/blob/4022968bf4d938b83c8b0f19b627eae06acf3f9f/src/coreclr/jit/llvmcodegen.cpp#L1280

[dump.txt](https://github.com/dotnet/runtimelab/files/10212644/dump.txt)

yowl:
This PR extends the condition when we build out GT_FIELD_LIST to include cases like 
```
struct { char, char }
````
This is emitted as an LLVM `i32` but needs to be alloc'ed, and built out as field stores.

Fixes #2112 which contains a log of an example that failed.

cc @SingleAccretion

### DIFF
--- a/src/coreclr/jit/llvmcodegen.cpp
+++ b/src/coreclr/jit/llvmcodegen.cpp
@@ -1206,7 +1206,7 @@ Value* Llvm::buildFieldList(GenTreeFieldList* fieldList, Type* llvmType)
 {
     assert(fieldList->TypeIs(TYP_STRUCT));
 
-    if (llvmType->isStructTy() || fieldList->Uses().begin() != fieldList->Uses().end())
+    if (llvmType->isStructTy() || fieldList->Uses().begin()->GetNext() != nullptr)
     {
         Value* alloca = _builder.CreateAlloca(llvmType);
         Value* allocaAsBytePtr = _builder.CreatePointerCast(alloca, Type::getInt8PtrTy(_llvmContext));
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/coreclr/jit/llvmcodegen.cpp [hunk 1]
```diff
@@ -1206,7 +1206,7 @@ Value* Llvm::buildFieldList(GenTreeFieldList* fieldList, Type* llvmType)
 {
     assert(fieldList->TypeIs(TYP_STRUCT));
 
-    if (llvmType->isStructTy())
+    if (llvmType->isStructTy() || fieldList->Uses().begin() != fieldList->Uses().end())
```
SingleAccretion: > `fieldList->Uses().begin() != fieldList->Uses().end()`

I think that's going to always be `true`, since the `end` iterator is initialized with `nullptr` while `begin` - with an always non-null `m_head`. So probably the better option is to just use the assert condition here (and delete the assert itself).

SingleAccretion (commented):
I assume it is expected we normalize `struct { char, char }` to `int32`?

src/coreclr/jit/llvmcodegen.cpp
yowl: ok, thanks!

SingleAccretion (approved):

yowl:
> I assume it is expected we normalize `struct { char, char }` to `int32`?

Yes, at least that's what the IL backend does.  struct sizes of 1,2,4,8 go to LLVM primitives

jkotas (approved):
Thank you!

PR #2124, opened 2022-12-16
NativeAOT-LLVM: revert change that caused the Ilc exit status to be lost

yowl:
This PR removes the `/b` from the `exit` command so the wrapper detects the exit status correctly.  In commit 87777b624b0b8ea295b0fe9cd8505ce262ed75c0 we will remove this file completely.


yowl:
Hopefully this will fail the CI now.  So I guess I leave this draft until a PR is merged that fixes the real problem.

Edit. There is no problem with the current code, so it doesn't fail.  However I think this is a worth taking in case Ilc does fail.

yowl:
cc @dotnet/nativeaot-llvm

SingleAccretion (approved):

PR #2127, opened 2022-12-17
Build wasm packages for official builds

jkotas:


jkotas:
This is the next problem that the package publishing is failing on. The wasm package is not getting built.

yowl:
Thanks, this is right for the wasm package.

PR #2128, opened 2022-12-17
Disable Build and run libraries tests for official builds

jkotas:
These tests are just a sanity check that non-wasm config is not broken. We do not need to run this sanity check for official builds.


SingleAccretion (approved):
Thank you for working on all these fixes!

PR #2129, opened 2022-12-17
[NativeAOT-LLVM] Fix local libraries build

SingleAccretion:
S.N.Quic ref project is special because it sets "IsNETCoreAppRef" to "false", which ends up meaning implicit framework references aren't disabled.

This is a problem because we build with the 6.0 SDK currently, and that doesn't support running source generators from the 7.0 reference pack that is pulled in.

I do not know why CI isn't hitting this. Perhaps a side effect of the locally installed 7.0 SDK?


yowl:
Thanks for the merge comments!

jkotas (approved):

PR #2130, opened 2022-12-17
[NativeAOT-LLVM] Add `LLVM_CMAKE_CONFIG_DEBUG` and `LLVM_CMAKE_CONFIG_RELEASE`

SingleAccretion:
Without this capability, one needs to manually switch `LLVM_CMAKE_CONFIG` when building for different configurations.

@dotnet/nativeaot-llvm


yowl (approved):
This is good, thanks.

PR #2132, opened 2022-12-17
[NativeAOT-LLVM] Fix NAOT runtime libraries VS experience for WASM

SingleAccretion:
It was not possible to simply open VS and start editing code for the WASM version of SPCL because some key MSBuild properties weren't set properly. This change fixes that: one can now simply define a new config in VS named "wasm" and things "just work".


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
LGTM!

PR #2133, opened 2022-12-18
[NativeAOT-LLVM] Build target and host packages for WASM on the same platform

SingleAccretion:
This matches how the build for other artifacts is set up.

We must take care to disable building the corresponding "host" platform, otherwise the publishing process will be racy (since both builds will end up producing "host" packages of the same name).

Hopefully, this is the last fix needed for package publishing to work.

### DIFF
--- a/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+++ b/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
@@ -15,12 +15,17 @@ steps:
 - ${{ if and(ne(parameters.archType, 'arm'), ne(parameters.archType, 'arm64')) }}:
 
   # For NativeAOT-LLVM, we have just built the Wasm-targeting native artifacts (the runtime and libraries).
-  # Now we need to build the cross-targeting compilers, RyuJit and ILC.
+  # Now we need to build the cross-targeting compilers, RyuJit and ILC. Likewise with packages.
 
   - ${{ if eq(parameters.archType, 'wasm') }}:
     - script: $(Build.SourcesDirectory)/build$(scriptExt) clr.alljits+clr.wasmjit+clr.nativeaotlibs+clr.tools -c $(buildConfigUpper)
       displayName: Build the ILC and RyuJit cross-compilers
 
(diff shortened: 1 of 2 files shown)


SingleAccretion:
Packages produced in the test run looked reasonable.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -34,8 +39,8 @@ steps:
   - ${{ elseif eq(parameters.osGroup, 'windows') }}:
     - script: $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) ${{ parameters.archType }} ${{ parameters.testFilter }} /p:NativeAotMultimodule=true
       displayName: Build tests
-  - ${{ elseif ne(parameters.osGroup, 'windows') }}:
-    - script: $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) ${{ parameters.archType }} 'tree nativeaot'
+  - ${{ else }}:
+    - script: $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) ${{ parameters.archType }} ${{ parameters.testFilter }}
```
SingleAccretion: Miscellaneous YAML cleanup.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl:
This looks good to me.

jkotas (approved):
LGTM. Thank you!

SingleAccretion:
Part of the change unintentionally made the CI run all of the Pri0 runtime tests. I have reverted that.

yowl:
Was the loss of the WIndows x64 Release build an intentional effect of this PR?

SingleAccretion:
Yes, it was. What used to be built "on" the windows_x64 platform is now built "on" browser_wasm_win. Otherwise we would be in a situation where the win-x64 package would be produced twice (and only one of them would have the right content).

PR #2135, opened 2022-12-21
Fix publishing of RID-agnostic packages

jkotas:


eng/pipelines/official/stages/publish.yml
```diff
@@ -1,5 +1,5 @@
 parameters:
-  PublishRidAgnosticPackagesFromPlatform: windows_x64
+  PublishRidAgnosticPackagesFromPlatform: browser_wasm
```
jkotas: `windows_x64` leg is not building the packages after #2133. We need to tell the publish process to pick-up the RID-neutral packages from `browser_wasm` leg.

eng/pipelines/runtimelab.yml
```diff
@@ -127,7 +127,7 @@ stages:
         buildArgs: -s clr.alljits+clr.tools+clr.nativeaotlibs+libs+nativeaot.packages -c Release
         extraStepsTemplate: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
         extraStepsParameters:
-          uploadIntermediateArtifacts: ${{ variables.isOfficialBuild }}
+          isOfficialBuild: ${{ variables.isOfficialBuild }}
```
jkotas, line 130: `isOfficialBuild` flag was not propagated correctly.

jkotas:
cc @yowl

yowl:
Thanks, guess this is another consequence of NativeAOT-LLVM being so far behind.

jkotas:
I think this one is caused more by NativeAOT-LLVM official build flow diverging from dotnet/runtime. I suspect that we will end up redoing #2133 at some point to make the official build flow more similar to dotnet/runtime again.

SingleAccretion:
Does this fix the `win-x64` package publish as well? If it does not and there is more friction to overcome to fix that, then we should revert #2133 and go in the other direction (i. e. build LLVM on the the host platforms, at least for official builds). The reason I thought the current solution better is that it is consistent with the WASM build itself hence easier to understand (and doesn't need to build LLVM twice), but consistency with the runtime in official builds trumps that, even at some complexity overhead, because of the high debugging cost.

jkotas:
Let's give this change a try. If it still does not work, we can re-evaluate the approach.

PR #2137, opened 2022-12-22
 [NativeAOT-LLVM] Pass OfficialBuildId when building packages

SingleAccretion:
Should fix the missing build numbers for packages.

@dotnet/nativeaot-llvm

### DIFF
--- a/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+++ b/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
@@ -22,7 +22,7 @@ steps:
       displayName: Build the ILC and RyuJit cross-compilers
 
     - ${{ if eq(parameters.isOfficialBuild, true) }}:
-      - script: $(Build.SourcesDirectory)/build$(scriptExt) libs+nativeaot.packages -c $(buildConfigUpper)
+      - script: $(Build.SourcesDirectory)/build$(scriptExt) libs+nativeaot.packages -c $(buildConfigUpper) /p:OfficialBuildId=$(Build.BuildNumber)
         displayName: Build host packages
 
   # Run libraries tests
--- a/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+++ b/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
@@ -18,11 +18,11 @@ steps:
   # Now we need to build the cross-targeting compilers, RyuJit and ILC. Likewise with packages.
 
   - ${{ if eq(parameters.archType, 'wasm') }}:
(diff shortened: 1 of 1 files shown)


eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -22,7 +22,7 @@ steps:
       displayName: Build the ILC and RyuJit cross-compilers
 
     - ${{ if eq(parameters.isOfficialBuild, true) }}:
-      - script: $(Build.SourcesDirectory)/build$(scriptExt) libs+nativeaot.packages -c $(buildConfigUpper)
+      - script: $(Build.SourcesDirectory)/build$(scriptExt) libs+nativeaot.packages -c $(buildConfigUpper) /p:OfficialBuildId=$(Build.BuildNumber)
```
jkotas: The build number is included in all binaries as version resource. Should it be passed in for "Build the ILC and RyuJit cross-compilers" step as well?

jkotas (approved):
Thanks

yowl:
I dont know when this changed, but looking at https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-experimental/NuGet/Microsoft.DotNet.ILCompiler.LLVM/versions/7.0.0-preview.2.22623.1
you can see:

image

The name as changed from alpha->dev->preview.  I don't know if that is a problem or its fine.  What does look odd, is that if I try a publish, it pulls `7.0.0-alpha.1.22617.2` still.  I.e. create a new console project and `dotnet publish /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false`  Then for packages I have:

image

image

image

While at https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-experimental we have

image

image

image

SingleAccretion:
The `preview` bits look as expected, since we are at about the point of P2 in runtime commits. The `dev` branding on the win-x64 package is not expected, presumably some parameters to `build.cmd` are still missing.

I am not sure about the local publish. Will investigate more tomorrow.

jkotas:
We are missing `-ci` argument for build.cmd
 https://github.com/dotnet/arcade/blob/e7ede87875f41a9b3df898ae08da5ebc96e24f56/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.BeforeCommonTargets.targets#L113-L114

jkotas:
Fixed by #2140

yowl:
Thanks the `alpha` versions in my publish was my fault.

Issue #2138, opened 2022-12-23
NativeAOT-LLVM: LLVM version upgrade to v15 and opaque pointers


yowl:
In the docs we recommend a version of emscripten, that uses v14 of LLVM.  We currently write LLVM using v11 and we've gotten away with this for a while, but its not recommended and it can produce

```
EXEC : error : Invalid cast (Producer: 'LLVM11.0.0' Reader: 'LLVM 14.0.0git')
```
At compile time.  We should move forward to LLVM v15 and at the same time move to opaque pointers.

https://llvm.org/docs/OpaquePointers.html

(I'm taking a look at this now).

yowl:
This is implemented and merged

PR #2139, opened 2022-12-23
NativeAOT-LLVM: Upgrade to LLVM v15 and opaque pointers

yowl:
This PR upgrades emsdk, and LLVM.  LLVM is moving to opaque pointers where the pointers have no types.  E.g. `i8*` becomes `ptr` .  LLVM `load`s need the type specified.  All the `bitcast` noops are gone from the clrjit module.   I didn't do the the IL module as it's going.

Note, there are some changes to the setup instructions for LLVM as it now requires the cmake tarball downloaded and installed.

Closes: #2138


yowl:
cc @dotnet/nativeaot-llvm

yowl:
I added a `nuget.config` for the `Ilc` solution to enable LLVMSharp v15 as its not in the normal feeds.

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -50,9 +50,16 @@ For the runtime libraries:
...
 
 For the compilers:
-- Download the LLVM 11.0.0 source from https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/llvm-11.0.0.src.tar.xz
-- Extract it and create a subdirectory in the `llvm-11.0.0.src` folder (`path-to-the-build-directory`).
-- Configure the LLVM source to use the same runtime as the Jit: `cmake -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE=Debug -D LLVM_USE_CRT_DEBUG=MTd path-to-the-build-directory` or if building for the Release configuration `cmake -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE=Release -D LLVM_USE_CRT_RELEASE=MT path-to-the-build-directory`
+- Download the LLVM 15.0.6 source from https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.6/lldb-15.0.6.src.tar.xz
+- Extract it and create a subdirectory in the `llvm-15.0.6.src` folder (`path-to-the-build-directory`).  Do not create at the root of a drive.
+- Download the CMake 15.0.6 source from https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.6/cmake-15.0.6.src.tar.xz
```
SingleAccretion, line 55: Hmm, custom CMake is quite an unusual requirement. Why is it needed?

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -552,10 +552,10 @@ Value* Llvm::consumeValue(GenTree* node, Type* targetLlvmType)
             return _builder.CreatePtrToInt(nodeValue, Type::getInt32Ty(_llvmContext));
         }
 
-        // i32* e.g symbols, to i8*
+        // all pointers are opaque
         if (nodeValue->getType()->isPointerTy() && targetLlvmType->isPointerTy())
```
SingleAccretion: Can this just be deleted then? Since all pointer types are (at the moment at least, since we don't use address spaces) are, presumably, equal?

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -847,25 +848,22 @@ void Llvm::buildLocalField(GenTreeLclFld* lclFld)
...
 
 void Llvm::buildLocalVarAddr(GenTreeLclVarCommon* lclAddr)
 {
     unsigned int lclNum = lclAddr->GetLclNum();
+    Value* localAddr = getLocalAddr(lclNum);
     if (lclAddr->OperIs(GT_LCL_FLD_ADDR))
     {
-        Value* bytePtr = castIfNecessary(getLocalAddr(lclNum), Type::getInt8PtrTy(_llvmContext));
-        mapGenTreeToValue(lclAddr, gepOrAddr(bytePtr, lclAddr->GetLclOffs()));
+        mapGenTreeToValue(lclAddr, gepOrAddr(localAddr, lclAddr->GetLclOffs()));
```
SingleAccretion: I think we can delete the branch (`if (lclAddr->OperIs`...) now, since it will work for both `LCL_VAR_ADDR` and `LCL_FLD_ADDR`.

docs/workflow/building/coreclr/nativeaot.md
yowl: I hit this problem: https://github.com/llvm/llvm-project/issues/57573#issuecomment-1237955018

SingleAccretion: Sad. I suppose there is nothing we can do then.

src/coreclr/jit/llvmcodegen.cpp
yowl: yes, it can, thanks

src/coreclr/jit/llvmcodegen.cpp
yowl: Thanks, looks like it.

docs/workflow/building/coreclr/nativeaot.md
yowl: but I've not fixed then yml scripts...  better do that.

src/coreclr/tools/aot/ILCompiler.LLVM/ILCompiler.LLVM.csproj
```diff
@@ -67,6 +63,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="libLLVM" Version="11.0.0" />
+    <PackageReference Include="LLVMSharp" Version="15.0.0-beta1" />
```
jkotas, line 66: What's the difference between libLLVM and LLVMSharp?

src/coreclr/tools/aot/nuget.config
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!-- For LLVMSharp -->
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
jkotas: This will fail in official builds. I will need to ask for the required packaged mirrored in the dotnet-public feed, and then this nuget.config can be deleted.

src/coreclr/tools/aot/ILCompiler.LLVM/ILCompiler.LLVM.csproj
yowl: libLLVM contains `libLLVM.dll` (via the `runtime.json` and the libllvm.runtime.win-x64 package) which is the native LLVM library for writing LLVM bitcode.  LLVMSharp is the object oriented interface over libLLVM (via the interop package) that Ilc uses in the IL backend.  When we delete the IL backend, we can remove LLVMSharp.

src/coreclr/tools/aot/nuget.config
yowl: ok, I will convert to draft.  When/if the package is mirrored I will remove the file and try another CI run.  Thanks

jkotas: The following packages have been added to the public feed:

LLVMSharp 15.0.0-beta1
libLLVM 15.0.0 
libLLVM.runtime.win-x64 15.0.0
libLLVM.runtime.ubuntu.20.04-x64 15.0.0
libLLVM.runtime.osx-x64 15.0.0
libLLVM.runtime.ubuntu.18.04-x64 15.0.0
libLLVM.runtime.win-x86 15.0.0
libLLVM.runtime.freebsd.11-x86 12.0.0
libLLVM.runtime.freebsd.11-x64 12.0.0
libLLVM.runtime.linux-arm64 15.0.0 
libLLVM.runtime.win-arm64 15.0.0
libLLVM.runtime.freebsd.13-x64 15.0.0
libLLVM.runtime.freebsd.12-x64 15.0.0
libLLVM.runtime.linux-x64 15.0.0 
libLLVM.runtime.osx-arm64 15.0.0 
libLLVM.runtime.ubuntu.22.04-x64 15.0.0

You should be able to delete this nuget.config now.

yowl: Please could you request LLVMSharp.Interop 15.0.0-beta1 as well?

jkotas: LLVMSharp.Interop should be there now

yowl:
Still got zlib problems, will try cherry picking https://github.com/dotnet/runtime/pull/73082

yowl:
I'm going to leave this as I can't find a way to build zlib with the latest emscripten, hopefully a merge from runtime will fix it in the future.

yowl:
Changing the correct cmakelist.txt helped.

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -50,9 +50,16 @@ For the runtime libraries:
...
+- Extract it to a folder called cmake at the same level as the llvm-15.0.0.src.
+- A sample layout could be:
+  f:\llvm-project
+  f:\llvm-project\llvm-15.0.6.src
+  f:\llvm-project\llvm-15.0.6.src\build # The subdirectory you created
+  f:\llvm-project\cmake
```
```suggestion
- Extract it to a folder called cmake at the same level as the llvm-15.0.0.src.
  - Sample layout:
    - f:\llvm-project
    - f:\llvm-project\llvm-15.0.6.src
    - f:\llvm-project\llvm-15.0.6.src\build # The subdirectory you created
    - f:\llvm-project\cmake
```
Looks like this rendered:

- Extract it to a folder called cmake at the same level as the llvm-15.0.0.src.
  - Sample layout:
    - f:\llvm-project
    - f:\llvm-project\llvm-15.0.6.src
    - f:\llvm-project\llvm-15.0.6.src\build # The subdirectory you created
    - f:\llvm-project\cmake

Otherwise the rendered version doesn't look great.

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -50,9 +50,16 @@ For the runtime libraries:
...
+- Configure the LLVM source to use the same runtime as the Jit (from the subdirectory you created, e.g. f:\llvm-project\llvm-15.0.6\src\build : `cmake -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE=Debug -D LLVM_USE_CRT_DEBUG=MTd -DLLVM_INCLUDE_BENCHMARKS=OFF path-to-the-build-directory` or if building for the Release configuration `cmake -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE=Release -D LLVM_USE_CRT_RELEASE=MT -DLLVM_INCLUDE_BENCHMARKS=OFF path-to-the-build-directory`
```
```suggestion
- Configure the LLVM source to use the same runtime as the Jit (from the subdirectory you created, e.g. f:\llvm-project\llvm-15.0.6\src\build: `cmake -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE=Debug -D LLVM_USE_CRT_DEBUG=MTd -DLLVM_INCLUDE_BENCHMARKS=OFF path-to-the-build-directory` or if building for the Release configuration `cmake -G "Visual Studio 17 2022" -DCMAKE_BUILD_TYPE=Release -D LLVM_USE_CRT_RELEASE=MT -DLLVM_INCLUDE_BENCHMARKS=OFF path-to-the-build-directory`
```

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1254,7 +1239,8 @@ void Llvm::buildCnsInt(GenTree* node)
             {
                 const char* symbolName = GetMangledSymbolName((void*)(node->AsIntCon()->IconValue()));
                 AddCodeReloc((void*)node->AsIntCon()->IconValue());
-                mapGenTreeToValue(node, _builder.CreateLoad(getOrCreateExternalSymbol(symbolName)));
+                mapGenTreeToValue(node, _builder.CreateLoad(Type::getInt8PtrTy(_llvmContext),
```
SingleAccretion: > `getInt8PtrTy`

Shouldn't we be always using `PointerType::getUnqual` from now on?

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1722,13 +1706,13 @@ void Llvm::storeObjAtAddress(Value* baseAddress, Value* data, StructDesc* struct
             {
                 // We can't be sure the address is on the heap, it could be the result of pointer arithmetic on a local var.
                 emitHelperCall(CORINFO_HELP_CHECKED_ASSIGN_REF,
-                               {address, castIfNecessary(fieldData, Type::getInt8PtrTy(_llvmContext))});
+                               {address, castIfNecessary(fieldData, llvm::PointerType::getUnqual(_llvmContext))});
```
SingleAccretion, line 1709: Hmm, when is `fieldData` not a pointer here?

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1738,7 +1722,8 @@ void Llvm::storeObjAtAddress(Value* baseAddress, Value* data, StructDesc* struct
     unsigned llvmStructSize = data->getType()->getPrimitiveSizeInBits() / BITS_PER_BYTE;
     if (structDesc->hasSignificantPadding() && llvmStructSize > bytesStored)
     {
-        Value* srcAddress = _builder.CreateGEP(baseAddress, _builder.getInt32(bytesStored));
+        Value* srcAddress =
+            _builder.CreateGEP(Type::getInt8Ty(_llvmContext), baseAddress, _builder.getInt32(bytesStored));
```
SingleAccretion: Can this use `gepOrAddr`?

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -156,7 +163,10 @@ private unsafe void CompileSingleMethod(CorInfoImpl corInfo, LLVMMethodCodeNode
                 
                 if (methodIL.GetExceptionRegions().Length == 0 && !_disableRyuJit)
                 {
+                    // S_P_CoreLib_System_Globalization_UmAlQuraCalendar___cctor
                     var mangledName = NodeFactory.NameMangler.GetMangledMethodName(method).ToString();
+                    // if (mangledName == "S_P_CoreLib_System_Globalization_UmAlQuraCalendar___cctor")
+                    // {
                     var sig = method.Signature;
```
SingleAccretion:
```suggestion
                    var mangledName = NodeFactory.NameMangler.GetMangledMethodName(method).ToString();
                    var sig = method.Signature;
```

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -165,13 +175,14 @@ private unsafe void CompileSingleMethod(CorInfoImpl corInfo, LLVMMethodCodeNode
...
-                    LLVMValueRef externFunc = ILImporter.GetOrCreateLLVMFunction(Module, mangledName,
-                        GetLLVMSignatureForMethod(sig, method.RequiresInstArg()));
+                    LLVMValueRef externFunc = ILImporter.GetOrCreateLLVMFunction(Module, mangledName, GetLLVMSignatureForMethod(sig, method.RequiresInstArg()));
                     externFunc.Linkage = LLVMLinkage.LLVMExternalLinkage;
 
                     ILImporter.GenerateRuntimeExportThunk(this, method, externFunc);
 
                     ryuJitMethodCount++;
+                    // }
+                    // else ILImporter.CompileMethod(this, methodCodeNodeNeedingCode);
```
SingleAccretion:
```suggestion
```

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -60,6 +65,8 @@ public sealed class LLVMCodegenCompilation : RyuJitCompilation
             DIBuilder = Module.CreateDIBuilder();
             DebugMetadataMap = new Dictionary<string, DebugMetadata>();
             ILImporter.Context = Module.Context;
+            LLVMContextSetOpaquePointers(Module.Context, false);
```
SingleAccretion, line 68: I am curious how this works linking wise. Does the Jit module still produce declarations/definitions of functions using precise pointer types, or...?

SingleAccretion (commented):
Overall this is a great change! Opaque pointers will make things faster and simpler for us.

docs/workflow/building/coreclr/nativeaot.md
yowl: Thanks!

src/coreclr/jit/llvmcodegen.cpp
yowl: Yes thanks, fixed.

src/coreclr/jit/llvmcodegen.cpp
yowl: E.g. Its the `RhpCheckedAssignRef` for `S_P_TypeLoader_Internal_TypeSystem_Instantiation___cctor` :

```
Generating:                [000043] ------------        t43 =    LCL_VAR   int    V01 tmp1         u:1 (last use)
                                                              /--*  t43    int
Generating: N005 (  3,  2) [000007] n-----------         t7 = *  OBJ       struct<Internal.TypeSystem.Instantiation, 4>
  %16 = load i32, ptr %0, align 4, !dbg !5
                                                              /--*  t10    byref
                                                              +--*  t7     struct
Generating: N004 ( 22,  8) [000012] -ACXG-------              *  STORE_OBJ struct<Internal.TypeSystem.Instantiation, 4> (copy)
```

`Instantiation` is 
```
    public struct Instantiation : IEquatable<Instantiation>
    {
        private TypeDesc[] _genericParameters;
```
So is this because of the LLVM struct representation of an `i32` I guess

SingleAccretion: I see. Bit unexpected (one would think the struct would be normalized to `ptr`), but also pre-existing so nothing to do.

src/coreclr/jit/llvmcodegen.cpp
yowl: yes, changed

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
yowl: right, removed that debug stuff

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
yowl: removed

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
yowl: In the Jit module 
```
declare i8 @S_P_CoreLib_System_Runtime_CompilerServices_RuntimeHelpers__Equals(ptr)
```
In the IL module
```
define i8 @S_P_CoreLib_System_Runtime_CompilerServices_RuntimeHelpers__Equals(i8* %0)
```
This is used in `Object__Equals` from the Jit module.  You can run wasm2wat on the final wasm and you see:

```
  (func $Object__Equals (type 1) (param i32) (result i32)
    (local i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32 i32)
    local.get 0
    i32.load
    local.set 1
    i32.const 4
    local.set 2
    local.get 0
    local.get 2
    i32.add
    local.set 3
    local.get 3
    i32.load
    local.set 4
    i32.const 8
    local.set 5
    local.get 0
    local.get 5
    i32.add
    local.set 6
    i32.const 8
    local.set 7
    local.get 0
    local.get 7
    i32.add
    local.set 8
    local.get 8
    local.get 1
    i32.store
    i32.const 12
    local.set 9
    local.get 0
    local.get 9
    i32.add
    local.set 10
    local.get 10
    local.get 4
    i32.store
    local.get 6
    call $S_P_CoreLib_System_Runtime_CompilerServices_RuntimeHelpers__Equals
```
As Wasm just uses i32 for the pointers, I guess the linker doesn't really care.

SingleAccretion: I see. Presumably means we can drop all of `GetParameterType`/`GetArgTypeIncludingParameterized` (in a follow up change) as well.

yowl: The wat for `$S_P_CoreLib_System_Runtime_CompilerServices_RuntimeHelpers__Equals`
```
func $S_P_CoreLib_System_Runtime_CompilerServices_RuntimeHelpers__Equals (type 1) (param i32) (result i32)
```
Where `type 1 ` is
```
  (type (;1;) (func (param i32) (result i32)))
```

SingleAccretion (approved):
:shipit:

yowl:
Our "normal" 
```
    WaitSubsystemTests.DoubleSetOnEventWithTimedOutWaiterShouldNotStayInWaitersList
    WaitSubsystemTests.ManualResetEventTest
    WaitSubsystemTests.AutoResetEventTest
    WaitSubsystemTests.SemaphoreTest
    WaitSubsystemTests.MutexTest
    WaitSubsystemTests.WaitDurationTest
```
Failures, retrying

yowl:
hurrah!

yowl:
Thanks for the packages and the reviews!

PR #2140, opened 2022-12-24
Fix pre-release label for official builds

jkotas:


jkotas:
cc @dotnet/nativeaot-llvm

PR #2136, opened 2022-12-22
NativeAOT-LLVM: Merge nativeaotruntime subset change

yowl:
This PR inches forward merging in the changed subset `NativeAotBuild`/`NativeAotRuntime`

PR #2143, opened 2022-12-27
NativeAOT-LLVM: merge the change to running tests

yowl:
This PR merges up to the change to how the NativeAOT tests are run.  You will see I struggled a bit with how to get this running from a windows.x64 host, and ended up setting `IlcToolsPath` and `IlcSdkPath` which was the least invasive way I thought of.

WIP for yml changes


yowl:
I removed the multi module tests as it looked like they were removed upstream.

yowl:
Cc @dotnet/nativeaot-llvm

jkotas:
It looks like the merge commit was squashed. The list of original commits is missing (the previous merge PR https://github.com/dotnet/runtimelab/pull/2136 was done correctly).

Could you please redo the merge without squashing so that we have proper history?

yowl:
Reapplied merge without fast-forward

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -68,8 +68,8 @@ Working on the Jit itself, one possible workflow is taking advantage of the gene
 It is also possible to publish an ordinary console project for Wasm using packages produced by the build: `build nativeaot.packages && build nativeaot.packages -a wasm -os Browser`, assuming all the binaries mentioned above have been built (note that the order is important - the build always produces an architecture-independent package that has a dependency on an architecture-dependent one, and we want that architecture-dependent package to be built for Wasm). Add the `path-to-repo/artifacts/packages/[Debug|Release]/Shipping` directory to your project's `NuGet.Config`, and the following two references to the project file itself:
 ```xml
 <ItemGroup>
-  <PackageReference Include="Microsoft.DotNet.ILCompiler.LLVM" Version="7.0.0-dev" />
-  <PackageReference Include="runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM" Version="7.0.0-dev" />
+  <PackageReference Include="Microsoft.DotNet.ILCompiler.LLVM" Version="7.0.0-*" />
+  <PackageReference Include="runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM" Version="7.0.0-*" />
```
SingleAccretion: Did the suffix change for locally built packages?

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -57,7 +57,7 @@ For the compilers:
...
 - Set the enviroment variable `LLVM_CMAKE_CONFIG` to locate the LLVM config: `set LLVM_CMAKE_CONFIG=path-to-the-build-directory/lib/cmake/llvm`. This location should contain the file `LLVMConfig.cmake`. `LLVM_CMAKE_CONFIG_DEBUG` and `LLVM_CMAKE_CONFIG_RELEASE` can be used instead of `LLVM_CMAKE_CONFIG` to allow the build to select the config file based on the configuration. If set, these variables take precedence over `LLVM_CMAKE_CONFIG`.
-- Build the Jits and the ILC: `build clr.jit+clr.wasmjit+clr.nativeaotruntime+clr.nativeaotlibs+clr.tools -c [Debug|Release]`. Note that `clr.jit` only needs to be built once.  Add the `libs` subsets if you want the packages for publishing, e.g. `build clr.jit+clr.wasmjit+clr.nativeaotruntime+clr.nativeaotlibs+clr.tools+libs -c Debug`.
+- Build the Jits and the ILC: `build clr.alljits+clr.wasmjit+clr.nativeaotruntime+clr.nativeaotlibs+clr.tools -c [Debug|Release]`. Note that `clr.jit` only needs to be built once.  Add the `libs` subsets if you want the packages for publishing, e.g. `build clr.alljits+clr.wasmjit+clr.nativeaotruntime+clr.nativeaotlibs+clr.tools+libs -c Debug`.
```
SingleAccretion: Why is `clr.alljits` needed now?

eng/Subsets.props
```diff
@@ -310,6 +310,11 @@
     <ProjectToBuild Include="$(CoreClrProjectRoot)nativeaot\BuildIntegration\BuildIntegration.proj" Category="clr" />
   </ItemGroup>
 
+  <!-- Running the tests for browser wasm requires BuildIntegration -->
+  <ItemGroup Condition="$(_subset.Contains('+clr.nativeaotlibs+')) and ('$(TargetOS)' == 'Browser' and '$(TargetArchitecture)' == 'wasm')">
+    <ProjectToBuild Include="$(CoreClrProjectRoot)nativeaot\BuildIntegration\BuildIntegration.proj" Category="clr" />
+  </ItemGroup>
+
```
SingleAccretion: The build integration is built for `-a wasm` when we request `clr.nativeaotruntime` and for the host as part of `clr.tools`. Why is it needed in `clr.nativeaotlibs` as well?

src/coreclr/jit/llvm.cpp
```diff
@@ -588,7 +588,11 @@ const HelperFuncInfo& Llvm::getHelperFuncInfo(CorInfoHelpFunc helperFunc)
         { FUNC(CORINFO_HELP_PATCHPOINT) },
         { FUNC(CORINFO_HELP_CLASSPROFILE32) },
         { FUNC(CORINFO_HELP_CLASSPROFILE64) },
-        { FUNC(CORINFO_HELP_PARTIAL_COMPILATION_PATCHPOINT) }
+        { FUNC(CORINFO_HELP_PARTIAL_COMPILATION_PATCHPOINT) },
+
+        // TODO-LLVM: validating indirect calls, does LLVM need or want this?
+        { FUNC(CORINFO_HELP_VALIDATE_INDIRECT_CALL) },
+        { FUNC(CORINFO_HELP_DISPATCH_INDIRECT_CALL) }
```
```suggestion
        { FUNC(CORINFO_HELP_PARTIAL_COMPILATION_PATCHPOINT) },
        { FUNC(CORINFO_HELP_VALIDATE_INDIRECT_CALL) },
        { FUNC(CORINFO_HELP_DISPATCH_INDIRECT_CALL) }
```
These `VALIDATE` helpers are part of the CFG feature which is [Windows-only](https://learn.microsoft.com/en-us/windows/win32/secbp/control-flow-guard).

It is possible (probable) that once/if we start thinking about doing non-WASM LLVM targeting we'll need it, but I believe at that point it'd be as simple as adding a flag for LLVM to generate the required code.

src/tests/Common/Directory.Build.targets
```diff
@@ -171,7 +171,6 @@
       <RunTimeDependencyCopyLocal Include="$(CoreDisToolsLibrary)" Condition="$(CopyCoreDisToolsToCoreRoot)" />
     </ItemGroup>
 
-<<<<<<< HEAD
     <!-- When cross-compiling tests (such as is the case for Wasm), we depends on the host's ILC being built. -->
     <!-- This is a workaround for ILC's project not supporting cross-targeting. Remove once that's implemented. -->
```
SingleAccretion:
```suggestion
    <!-- TODO-LLVM: investigate refactoring in light of ILC project supporting cross-targeting. -->
```
Evidently we now do have cross-compilation for ILC, but it is not totally clear to me we can use it "as-is" (if at all).

src/tests/build.proj
```diff
@@ -483,6 +483,10 @@
       <GroupBuildCmd Condition="'$(TestBuildMode)' == 'nativeaot'">$(GroupBuildCmd) "/p:DefaultBuildAllTarget=BuildNativeAot"</GroupBuildCmd>
       <GroupBuildCmd Condition="'$(IlcMultiModule)' == 'true'">$(GroupBuildCmd) "/p:IlcMultiModule=true"</GroupBuildCmd>
       <GroupBuildCmd Condition="'$(BuildNativeAotFrameworkObjects)' == 'true'">$(GroupBuildCmd) "/p:BuildNativeAotFrameworkObjects=true"</GroupBuildCmd>
+
+      <!-- Set Ilc and the SDK patch for cross compilation -->
```
SingleAccretion: Am I right in understanding this is added here because `src/tests/Directory.Build.targets` doesn't set the properties for WASM?

docs/workflow/building/coreclr/nativeaot.md
yowl: Thanks, whatever problem I thought there was does not exist.  Removed change

docs/workflow/building/coreclr/nativeaot.md
yowl: I was confused with the docs and the public packages, reverted.

eng/Subsets.props
yowl: I guess I added that because we dont ask for `clr.nativeaotruntime` at line 52 (hyperlink not working):
```
 Run `build clr.nativeaotlibs+libs -c [Debug|Release] -a wasm -os Browser`
```

yowl: Let me try without it.

yowl: reverted

yowl: Should we add `clr.nativeaotruntime` to line 52 of the nativeaot.md?

```
- Run `build clr.nativeaotlibs+libs -c [Debug|Release] -a wasm -os Browser`. This will create the architecture-dependent libraries needed for linking and runtime execution, as well as the managed binaries to be used as input to ILC. Add the `clr.nativeaotbuild` subset if you want the packages for publishing, e.g. `build clr.nativeaotlibs+clr.nativeaotbuild+libs -c Debug -a wasm -os Browser`
```

Without it you can't run the tests

SingleAccretion: I think #2146 already added it.

src/tests/build.proj
yowl: I think you can remove the conditions from `src/tests/Directory.Build.targets` and get past the missing `IlcSdkPath`.

You then hit 
```
  E:\GitHub\runtimelab2\.dotnet\sdk\6.0.100\MSBuild.dll -distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,E:\GitHub\runtimelab2\.dotnet\sdk\6.0.100\dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,E:\GitHub\runtimelab2\.dotnet\sdk\6.0.100\dotnet.dll -maxcpucount /maxcpucount -verbosity:m /bl:E:\GitHub\runtimelab2\artifacts\/log/Debug/InnerManagedTestBuild.3.binlog /nodeReuse:false /p:TargetArchitecture=wasm /p:Configuration=Debug /p:LibrariesConfiguration=debug /p:TargetOS=Browser /p:RuntimeOS= /p:RuntimeFlavor=coreclr /p:RuntimeVariant= /p:CLRTestBuildAllTargets= /p:__TestGroupToBuild=3 /p:__SkipRestorePackages=1 /p:DefaultBuildAllTarget=BuildNativeAot /t:Build E:\GitHub\runtimelab2\src\tests\build.proj
    XUnitWrapperGenerator -> E:\GitHub\runtimelab2\artifacts\tests\coreclr\Browser.wasm.Debug\Common\XUnitWrapperGenerator\XUnitWrapperGenerator\XUnitWrapperGenerator.dll
    Dataflow -> E:\GitHub\runtimelab2\artifacts\tests\coreclr\Browser.wasm.Debug\nativeaot\SmokeTests\Dataflow\Dataflow\Dataflow.dll
    Generating native code
    '"E:\GitHub\runtimelab2\artifacts\bin\coreclr\Browser.wasm.Debug\ilc\\ilc"' is not recognized as an internal or external command,
    operable program or batch file.
E:\GitHub\runtimelab2\artifacts\bin\coreclr\Browser.wasm.Debug\build\Microsoft.NETCore.Native.targets(292,5): error MSB3073: The command ""E:\GitHub\runtimelab2\artifacts\bin\coreclr\Browser.wasm.Debug\ilc\\ilc" @"E:\GitHub\runtimelab2\artifacts\tests\coreclr/obj/Browser.wasm.Debug/Managed/nativeaot\SmokeTests\Dataflow\Dataflow\native\Dataflow.ilc.rsp"" exited with code 9009. [E:\GitHub\runtimelab2\src\tests\nativeaot\SmokeTests\Dataflow\Dataflow.csproj] [E:\GitHub\runtimelab2\src\tests\build.proj]
```
Because:
image

SingleAccretion: The way I imagine this should work is that when we are cross-compiling the tests (which should always be the case for WASM), the ILC path should be set to the appropriate "host" value.

We used to achieve this with the code in `src/tests/Common/Directory.Build.targets` that constructed the core root. I now see [#64738](https://github.com/dotnet/runtime/pull/64738) actually just deleted that code, so it doesn't work anymore.

One particular piece of experience that needs to work is compiling individual tests without going through `src/tests/build.proj`.

SingleAccretion: I see we already have `CoreCLRCrossILCompilerDir` defined `liveBuilds.targets`, so it should be straightforward to adapt code in `src/tests/Directory.Build.targets` to use it I would hope.

yowl: ok let me look into that, thanks.

yowl: Doesn't look like it gets set properly, will trace that
image

SingleAccretion: I think the value is right, just the location somewhat unusual. What I do not recall is part of which build is this cross-component, host or target.

SingleAccretion: Hmm, so I looked again at this cross-arch stuff and it appeared simpler to me to keep doing what we've been doing with the core root construction code, i. e. pick the host compiler out of host's build.

The cross-arch setup looks inconvenient for our scenario where cross-compilation is the mainline workflow.

yowl: I think `CORE_ROOT` is setup in run.py, is that what you had in mind?

SingleAccretion: I basically meant that we should migrate what we did in `src/tests/Common/Directory.Build.targets` with `<!-- When cross-compiling tests (such as is the case for Wasm), we depends on the host's ILC being built. -->` to here (and delete from there as it is now evidently dead code).

yowl: Thanks, I see better now.  `src/tests/Common/Directory.Build.targets` would be ok perhaps, but the build fails before it gets a chance to run.  The target we want from there is `CopyDependencyToCoreRoot` however that ultimately is `dependsOn` for `GenerateLayout`.  That is a `TestBuildSteps` for `TestBuild`, the entry target, but its after `ManagedBuild` which is the failing step.  I tried to set up another dependency to `CopyDependencyToCoreRoot`:
```
  <Target Name="ManagedBuild"
      DependsOnTargets="CopyDependencyToCoreRoot;BuildManagedTestGroups"
```
 but it runs into problems because the msbuild properties are not inherited to `src/tests/Common/Directory.Build.targets`, the `RepoRoot` is not set, and `<Import Project="$(RepoRoot)eng/liveBuilds.targets" />` fails.  It might be possible to make this approach work, but I left it there.

Have moved the copy into a new `Target` in `src\tests\Directory.Build.targets`

SingleAccretion: I don't think there is a need to move the core root construction code here, just the path setting.

I. e. before what happened was:
```
// 1) Core root is constructed with the paths / copy target
// 2) Compiler from core root is used when running the tests
```
Now what should happen is:
```
// 1) Compiler from the path we provide is used during the build
```
I. e. there is no more copying of ILC to the core root directory.

yowl:
```
  <Target Name="NativeAotWasmCopyDependencyToCoreRoot" BeforeTargets="IlcCompile" Condition="'$(TestBuildMode)' == 'nativeaot' and '$(TargetArchitecture)' == 'wasm'">

    <!-- Set IlcToolsPath for cross compilation -->
    <PropertyGroup>
      <IlcToolsPath>$([MSBuild]::NormalizeDirectory('$(RepoRoot)', 'artifacts', 'bin', 'coreclr', '$(HostOS).$(BuildArchitecture).$(CoreCLRConfiguration)\ilc'))</IlcToolsPath>
    </PropertyGroup>
    <Error Condition="!Exists('$(IlcToolsPath)')"
           Text="'$(IlcToolsPath)' does not exist. Subsets clr.jit, clr.wasmjit and clr.tools for $(HostOS) $(BuildArchitecture) in $(CoreCLRConfiguration) configuration must be built before building this project." />
  </Target>
```
Added this target which is passing in CI now.

yowl: hold on, let me rename it

SingleAccretion: The name's not quite right but we can fix it separately.

Edit: or not :).

SingleAccretion (approved):

src/tests/build.proj
yowl: ok, async human communication, beep.  I changed it.  @jkotas I guess if it it passes this time, then it can go, thanks

yowl: Looks like I need to also fix a conflict

yowl:
Passed. thanks for the help.

PR #2144, opened 2022-12-28
[NativeAOT-LLVM] Various fixes and code cleanup

SingleAccretion:
Another set of changes extracted from the EH work. Please refer to the individual commits for the contents.

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -5071,6 +5071,16 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
     compJitStats();
 
 #if defined(TARGET_WASM)
+#ifdef DEBUG
+    static ConfigMethodRange EnableLlvmRange;
+    EnableLlvmRange.EnsureInit(JitConfig.EnableLlvmRange());
+
+    if (!EnableLlvmRange.Contains(info.compMethodHash()))
+    {
+        fatal(CORJIT_SKIPPED);
+    }
+#endif // DEBUG
+
     if (opts.OptimizationEnabled())
     {
         // When optimizing, we'll sort the locals on the shadow stack by ref count.
--- a/src/coreclr/jit/llvmlower.cpp
+++ b/src/coreclr/jit/llvmlower.cpp
@@ -106,8 +106,6 @@ void Llvm::lowerLocals()
             if (varDsc->lvIsParam && !isLlvmParam)
             {
                 shadowStackParamCount++;
-                varDsc->lvIsParam = false;
-
                 shadowStackLocals.push_back(varDsc);
                 continue;
             }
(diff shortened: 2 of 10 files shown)


src/coreclr/jit/compiler.cpp
```diff
@@ -5071,6 +5071,16 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
...
 
 #if defined(TARGET_WASM)
+#ifdef DEBUG
+    static ConfigMethodRange EnableLlvmRange;
+    EnableLlvmRange.EnsureInit(JitConfig.EnableLlvmRange());
+
+    if (!EnableLlvmRange.Contains(info.compMethodHash()))
+    {
+        fatal(CORJIT_SKIPPED);
+    }
```
SingleAccretion, line 5081: This allows one to binary search for methods causing silent bad codegen. The syntax is `--codegenopt:EnableLlvmRange=00000000-FFFFFFFF` (method hashes are unsigned 4-byte integers).

src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
```diff
@@ -433,6 +433,11 @@ private void ImportCall(ILOpcode opcode, int token)
                                 _dependencies.Add(_factory.ReadyToRunHelper(ReadyToRunHelperId.DelegateCtor, info), reason);
                             }
 
+                            // TODO-LLVM: delete when IL->LLVM module is gone.
+                            if (_compilation.TargetArchIsWasm())
+                            {
+                                _dependencies.Add(info.Constructor, "LLVM delegate ctor");
+                            }
```
SingleAccretion, line 440: There were some scenarios where scanning missed the appropriate intrinsic initializer; I am pretty sure this is overly conservative but debugging deeper did not seem worth it.

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvmlower.cpp
```diff
@@ -214,20 +212,59 @@ void Llvm::assignShadowStackOffsets(std::vector<LclVarDsc*>& shadowStackLocals,
...
+            varDsc->SetStackOffset(offset);
+            offset = padNextOffset(corInfoType, classHandle, offset);
+        }
+
+        // We will use this as the indication that the local has a home on the shadow stack.
+        varDsc->SetRegNum(REG_STK);
+    };
+
+    // First, we process the parameters, since their offsets are fixed by the ABI. Then, we process the rest.
+    // Doing this ensures we don't count LLVM parameters live on the shadow stack as shadow parameters.
```
yowl, line 240: What is a "shadow parameter" in this context?

yowl:
Looks good, thanks, I just had the one comment about the comment "LLVM parameters live on the shadow stack as shadow parameters" .  I.e. an LLVM parameter that is live on the shadow stack, is a parameter to a method that is called from the compiling method?

yowl (approved):

src/coreclr/jit/llvmlower.cpp
SingleAccretion: One that is live on the shadow stack on entry, i. e. not an LLVM parameter. So these will be things with GC references in them.

We can also have `LLVM parameters live on the shadow stack`, e. g. `int`s, for other reasons, currently those include being live into (or out of) handlers (I suspect there will be more in the future), and those we have to explicitly spill as the code does.

PR #2145, opened 2023-01-01
[NativeAOT-LLVM] Exception handling: dispatch

SingleAccretion:
This change implements exception handling dispatch for the RyuJit-based LLVM backend.

All funclets will have two LLVM parameters: the shadow stack and the original shadow stack, which is shared among all of the functions belonging to the same logical parent. The shadow stack is used for calls, while the original shadow stack - to access state from the original shadow frame. Once we move to the correct pass order, the frames while dispatch is active will look something like this (note I am using the traditional "stacks grow down" notation here, while physically the shadow stack grows toward higher addresses in our implementation):
```
; First pass

  ; Shadow frames                            ; Native frames

  [Filtering F0] (with C0 catch)             [Filtering F0]
  [Finally   S0]                                            [Dispatcher]  ^ ; Progression of the native exception
  [Filtering F1]                             [Filtering F1] [F0 frames]   | ; stops once we find a filter which
  [Finally   S1]                                            [Dispatcher]  | ; accepts its managed counterpart.
  [Filtering F2]                             [Filtering F2] [F1 frames]   |
  [Finally   S2]                                            [Dispatcher]  |
  [Throw]                                    [Throw]        [F2 frames]   |
  [Dispatcher(s)]
  [F2 frames] [F1 frames] ... ; Native exception carries the dispatcher's shadow stack

; Second pass                                

  ; Shadow frames                            ; Native frames

  [Filtering F0] <-------------------------| [Filtering F0] <---------------------------------------------|
  [Finally   S0]                           |                [Dispatcher]  ; The handler was found         |
  [Filtering F1]                           |                [S2 frames] [S1 frames] ... [C0 frames]-------|
  [Finally   S1]                           |
  [Filtering F2]                           |
  [Finally   S2]                           |
  [Throw]                                  |
  [Dispatcher]                             |
  [S2 frames] [S1 frames] ... [C0 frames]--| ; Normal "ret" from the dispatcher
```
So in both passes we will have funclets located above the dispatcher. While it is conceivable to avoid this for the second pass, in the first pass it is not avoidable. The simple solution using the original shadow frame pointer solves both cases nicely. It also solves the issue of reserving space for the exception object: it is passed on the ordinary shadow stack and thus cannot clobber any state from the original shadow frame.

While the above is the high level design of the overall dispatch, the bulk of implementation is concerned with solving a narrower problem: per-frame dispatch. Here a simple approach of encoding the EH tree directly into code with one native handler (landing pad) per one managed clause was chosen, for three reasons:
1) It is simple.
2) It transparently handles throws from funclets.
3) WASM exceptions use similar nesting structure, so it should make the code easier to adapt when moving to them.

Here is the comment from code which describes the specifics of how we lay out the dispatch blocks for one protected region:
```
// INNER_TRY_REGION:
//   invoke ThrowingCall()
//          unwind to DISPATCH_PAD_INNER;
//
// DISPATCH_PAD_INNER:
//   dispatchData.CppExceptionTuple = landingPadInst
//   dispatchData.DispatcherData = null
//   goto DISPATCH_INNER;
//
// DISPATCH_INNER:
//   dispatchDest = DispatchFunction(FuncletShadowStack(), &dispatchData, &HandlerFunclet, ...)
//                  unwind to DISPATCH_PAD_OUTER
//   if (dispatchDest == 0)
//      goto DISPATCH_OUTER; // For nested regions; top-level ones will use the "switch".
//   goto UNIFIED_DISPATCH;
//
// UNIFIED_DISPATCH:
//   switch (dispatchDest) {
//       case 0: goto RESUME;
//       case 1: goto BB01;
//       case 2: goto BB02;
//       ...
//       default: goto FAIL_FAST;
//   }
//
// RESUME:
//   resume(dispatchData.CppExceptionTuple); // Rethrow the exception and unwind to caller.
//
// FAIL_FAST:
//   FailFast();
//
// What is the possibe set of dispatch destinations (aka why have "UNIFIED_DISPATCH")?
//
// We consider the tree of active protected regions above this one, that are also contained in the same funclet.
// For each region with a (possibly filtered) catch handler, we consider successors of all "catchret" blocks.
// The union of these will form the set of all possible dispatch destinations for the current protected region.
// However, we do not actually emit the "switch" code for each individual region, as it would mean quadratic
// code size growth (number of dispatch destinations X number of protected regions) for deeply nested EH trees.
// Instead, we create one "universal" dispatch block for each funclet, and jump to it from each dispatch. Note
// that thanks to the step blocks inserted by "impImportLeave", we do not need to consider cases where a jump
// from a funclet to its caller would be required.
```
Notable difference from the IL backend: since we're generating one dispatcher for one clause (note we count mutually protecting clauses as one clause), we employ dispatchers specialized for the clause type. This is expected to be a somewhat significant saving, both in code size and dispatch time, since the vast majority (95%+) of clauses are either faults/finallys or single catches. Here are the statistics I gathered using SPMI:
```
benchmarks.run

Total clauses       : 2514
Finally             : 281    11.2%
Fault               : 1413   56.2%
Single catch        : 724    28.8%
Single filter       : 49      1.9%
Mutually protecting : 47      1.9%
Among them, catches : 108    86.4%
Among them, filters : 17     13.6%

libraries.pmi

Total clauses       : 19568
Finally             : 1937    9.9%
Fault               : 11399  58.3%
Single catch        : 5373   27.5%
Single filter       : 349     1.8%
Mutually protecting : 510     2.6%
Among them, catches : 1240   95.0%
Among them, filters : 65      5.0%
```
The prevalence of faults here, which are not expressible in C#, is due to finally cloning, which duplicates the code inside the finally so we save on the call overhead for the common path and gain additional optimization opportunities. We may choose to disable finally cloning as a size-increasing optimization in the future for WASM, but for now it is, I suppose, a good demonstration of taking advantage of existing RyuJit optimizations.

It was also chosen that we emit a fail fast in case a catch handler returns an unexpected dispatch location. This is technically unreachable code; the thinking was to leave it there while the EH implementation matures and delete once we are confident in its correctness.


SingleAccretion:
@dotnet/nativeaot-llvm 

Please feel free to ask any and all questions about the implementation / design. This is a fairly large and fundamental change.

SingleAccretion:
There is a silent bad codegen problem here, as evidenced by CI. Not the first one (a lot of the fixes present here that are nominally unrelated to EH also manifested themselves as silent bad codegen), but the hardest one to debug so far because it does not reproduce locally. Probably will take a while to sort it out...

yowl:
The tests that are removed, is that because the IL backend is no longer being used for them, and so the cases are not applicable ?

SingleAccretion:
All of the commits after and including `Testing...` are temporary, for debugging the elusive bad codegen issue, and will be reverted.

yowl:
Is it always in `TestFilterNested` ?

SingleAccretion:
Yes, here's the call stack:
```
JS:     at invoke_ii (D:\a\_work\1\s\artifacts\tests\coreclr\Browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:7279:36)
IL:     at S_P_CoreLib_Internal_Reflection_Extensions_NonPortable_CustomAttributeSearcher_1__GetMatchingCustomAttributesIterator_d__2<System___Canon>__MoveNext (<anonymous>:wasm-function[5360]:0x4e7c34)
IL:     at S_P_CoreLib_System_Attribute__OneOrNull (<anonymous>:wasm-function[4943]:0x462cff)
RyuJit: at S_P_CoreLib_System_Attribute__GetCustomAttribute (<anonymous>:wasm-function[11794]:0xb102e4)
RyuJit: at S_P_CoreLib_System_Reflection_CustomAttributeExtensions__GetCustomAttribute (<anonymous>:wasm-function[11829]:0xb151d9)
IL:     at S_P_CoreLib_System_Reflection_CustomAttributeExtensions__GetCustomAttribute_3<System___Canon> (<anonymous>:wasm-function[1650]:0x12be05)
IL:     at S_P_CoreLib_System_Resources_ManifestBasedResourceGroveler__GetNeutralResourcesLanguage (<anonymous>:wasm-function[6018]:0x5e93b1)
RyuJit: at S_P_CoreLib_System_Resources_ResourceManager__CommonAssemblyInit (<anonymous>:wasm-function[10795]:0xaa010c)
RyuJit: at S_P_CoreLib_System_Resources_ResourceManager___ctor_3 (<anonymous>:wasm-function[12358]:0xb3caed)
RyuJit: at S_P_CoreLib_System_SR__get_ResourceManager (<anonymous>:wasm-function[17090]:0xd46ee4)
JS:     at invoke_vii (D:\a\_work\1\s\artifacts\tests\coreclr\Browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:7257:29)
IL:     at S_P_CoreLib_System_SR__InternalGetResourceString (<anonymous>:wasm-function[6380]:0x66b883)
JS:     at invoke_vii (D:\a\_work\1\s\artifacts\tests\coreclr\Browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:7257:29)
RyuJit: at S_P_CoreLib_System_SR__GetResourceString (<anonymous>:wasm-function[8913]:0x99225d)
RyuJit: at S_P_CoreLib_System_SR__get_Arg_ArgumentException (<anonymous>:wasm-function[10960]:0xab0c0c)
RyuJit: at S_P_CoreLib_System_ArgumentException___ctor (<anonymous>:wasm-function[8975]:0x995d7a)
RyuJit: at HelloWasm_Program__TestFilterNested (<anonymous>:wasm-function[12648]:0xb5d517)
RyuJit: at HelloWasm_Program__TestTryCatch (<anonymous>:wasm-function[12855]:0xb75599)
RyuJit: at HelloWasm_Program__Main (<anonymous>:wasm-function[3668]:0x301ef0)
RyuJit: at HelloWasm__Module___MainMethodWrapper (<anonymous>:wasm-function[13246]:0xb9344e)
```
As one of the commits notes, prime suspect is shadow stack corruption.

yowl:
Hmm, I wonder if its the same stack corruption with the IL second pass handler.  I got that better, but wasn't sure it was totally ok in https://github.com/yowl/runtimelab/tree/uo-loop-test .  Nested handlers can corrupt the shadow stack on calls made in the handlers.

Once the shadow stack corrupts its really hard, I feel the pain here.  If it doesn't repro locally, somethings you could try to make it repro are running without memory growth and playing with the fixed memory size.

SingleAccretion:
> Hmm, I wonder if its the same stack corruption with the IL second pass handler. I got that better, but wasn't sure it was totally ok in https://github.com/yowl/runtimelab/tree/uo-loop-test . Nested handlers can corrupt the shadow stack on calls made in the handlers.

Yes, that is the first-order suspicion... It would be helpful is understanding how does this corruption happen, i. e. where do the shadow stack pointers start to alias.

> Once the shadow stack corrupts its really hard, I feel the pain here. If it doesn't repro locally, somethings you could try to make it repro are running without memory growth and playing with the fixed memory size.

Thank you for the advice!

yowl:
From memory and a glance at the changes in that branch, if you have a stack (top down)

```
InvokeSecondPassWasm
FinallyFunclet1
.
InvokeSecondPassWasm
FinallyFunclet2
MethodA
```
Then (from memory), the call to MethodA passes the wrong shadow stack - it doesn't cater for locals used in `FinallyFunclet1`, I think that's right and is why in the `uo-loop-test` branch I added another parameter to `InvokeSecondPassWasm`.

Plus the other change in that branch, is the removal a shadow stack block for every reverse pinvoke, although I dont remember if that was just a performance benefit, or there was a logical problem with the way it was.

You could take the changes to the IL side and just try it I suppose.  If it doesn't help then it can be discounted.  Let me know if you want me to do that to this branch.

SingleAccretion:
>  Let me know if you want me to do that to this branch.

That would be greatly appreciated \:).

yowl:
> That would be greatly appreciated :).

in https://github.com/dotnet/runtimelab/pull/2149.  I realised while merging that there is third change here that may be more obvious.  In a couple of places I added an RhpCheckedAssignRef, where previously there was just an LLVM store.  If this goes anywhere we could see which is helping.  Throwing from finally blocks is not that common so that idea is maybe not a strong theory.

SingleAccretion:
Good news on the SBCG reduction front: the latest minimal example reproduces for me locally. Will be digging in.

SingleAccretion:
Alright, I think I have a pretty good idea of what the bug is about now.

In short, we start with this IL:
```il
IL_00d6: nop
IL_00d7: ldarg.0
IL_00d8: ldfld class System.Func`2<class System.Type, bool> class Internal.Reflection.Extensions.NonPortable.CustomAttributeSearcher`1/'<GetMatchingCustomAttributesIterator>d__2'<!E>::'<passesFilter>5__2'
IL_00dd: ldarg.0
IL_00de: ldfld class System.Reflection.CustomAttributeData class Internal.Reflection.Extensions.NonPortable.CustomAttributeSearcher`1/'<GetMatchingCustomAttributesIterator>d__2'<!E>::'<cad>5__5'
IL_00e3: callvirt instance class System.Type System.Reflection.CustomAttributeData::get_AttributeType()
IL_00e8: callvirt instance !1 class System.Func`2<class System.Type, bool>::Invoke(!0)
IL_00ed: stloc.2
IL_00ee: ldloc.2
IL_00ef: brfalse.s IL_0127
```
Which translates to a lot of LLVM, but the important part is that we have a GC value live in an LLVM register across a safe point:
```llvm
%passesFilter = load %someAddress

call get_AttributeType() ; Tiggers GC, relocates %passesFilter

%passesFilterInvokeFunc = [%passesFilter] ; Some lookup, uses outdated %passesFilter
call (%passesFilterInvokeFunc)(...)       ; Crashes
```
@yowl would this ring any bells (I am yet to look at the IL backend code for this)?

RyuJit backend has the same bug currently, but it should be mitigated by the fact RyuJit importation aggressively spills call arguments (not that this is great protection; it must eventually be fixed properly of course).

yowl:
I thought everything on the SS was pinned.
"
@yowl in conservative stack scanning, every pointer is assumed to be interior and pinned, interior pointer scanning is expensive and pinned means the GC cannot relocate them. For precise stack scanning you would need to differentiate if it's an object reference or managed (interior) pointer and whether they're pinned
"
Do you agree with this statement, it came up in gitter in 2020!

SingleAccretion:
I would certainly expect that to be true, otherwise everything would be quite broken.

The problem here is that the value in question never gets to live on the shadow stack, so my diagram is actually not quite accurate. The accurate version is this:
```scala
%passesFilter = load %someAddress

call get_AttributeType() ; Tiggers GC, relocates %passesFilter

%passesFilterInvokeFunc = [%passesFilter] ; Some lookup, uses outdated %passesFilter
call (%passesFilterInvokeFunc)(...)       ; Crashes
```

yowl:
Yes, I can believe that depending on the branch order in the IL and execution order.  However I'd have thought the RyuJit backend fixed this as it evaluates the SS placements up front.  I sort of did something similar in that other branch "backfilling" the shadow stack placements.

SingleAccretion:
It is still a problem for RyuJit because we don't spill some SDSUs to the shadow stack (loads most notably), so if we have something like this:
```
t1  = IND<ref>(some address)
      CALL GC.Collect()
      USE(t1) // Will use the outdated version of 't1'
```
But here the bug is actually in the IL backend-produced method, so the question is how we fix it there.

(For RyuJit I will be working on a new SS pass which will handle this as well as other things)

yowl:
Is this `S_P_CoreLib_Internal_Reflection_Extensions_NonPortable_CustomAttributeSearcher_1__GetMatchingCustomAttributesIterator_d__2<System___Canon>__MoveNext`  ?

SingleAccretion:
Yes.

yowl:
```cs
  %59 = getelementptr i8, i8* %0, i32 40, !dbg !277196
  %60 = getelementptr i8, i8* %59, i32 0, !dbg !277196
  %CastPtrTypedStore53 = bitcast i8* %60 to i8**, !dbg !277196
  store i8* %"loadField_<>4__this", i8** %CastPtrTypedStore53, align 4, !dbg !277196
  %61 = getelementptr i8, i8* %59, i32 4, !dbg !277196
  %CastPtrTypedStore54 = bitcast i8* %61 to i8**, !dbg !277196
  store i8* %loadField_element, i8** %CastPtrTypedStore54, align 4, !dbg !277196
  invoke void %ld__getslot__(i8* %59, i8* %Temp1_)
          to label %Try3EC55 unwind label %LandingPad0, !dbg !277196

Try3EC55:                                         ; preds = %Try3EC52
  %"CastPtrVariable15_[S.P.CoreLib]System.__Canon" = bitcast i8* %Local15_ to i8**, !dbg !277196
  %CastPtrTemp1_ = bitcast i8* %Temp1_ to i8**, !dbg !277196
  %LdTemp1_ = load i8*, i8** %CastPtrTemp1_, align 4, !dbg !277196
```
SS is `%59 ` but and is `getelementptr i8, i8* %0, i32 40` , however the prolog has:

```cs
    %Temp1_ = getelementptr i8, i8* %0, i32 40
```
so the `invoke void %ld__getslot__(i8* %59, i8* %Temp1_)` is a potential problem I think, depending on what it does with the SS, as there is a `load` of `%Temp1_` after the call.

yowl:
> But here the bug is actually in the IL backend-produced method, so the question is how we fix it there.

I would be tempted to just take just the backfill stuff from that branch, I can prepare a PR so you can look at that ?

SingleAccretion:
> I would be tempted to just take just the backfill stuff from that branch

It is not clear to me it would solve this particular problem. As I understand, the problem it is solving is the lazy buildup of the shadow stack being not quite correct, but if we have GC refs live in LLVM registers only, it is a problem of its own. Presumably, we need to spill them to the shadow stack in some way. Which implies more chances to hit issues, so... both should be done I suppose?

Edit: also, here's the LLVM that triggers the bug:
```llvm
Prolog:
  %Argument0_ = getelementptr i8, i8* %0, i32 0

...

TryD6:                                            ; preds = %TryCB
  call void @llvm.donothing(), !dbg !337096
  %CastPtr462 = bitcast i8* %Argument0_ to i8**, !dbg !337096
  %load_arg0_463 = load i8*, i8** %CastPtr462, align 4, !dbg !337096
  %332 = getelementptr i8, i8* %0, i32 100, !dbg !337097
  invoke void @corert.throwifnull(i8* %332, i8* %load_arg0_463)
          to label %TryDD unwind label %LandingPad0, !dbg !337097

TryDD:                                            ; preds = %TryD6
  %333 = getelementptr i8, i8* %load_arg0_463, i32 32, !dbg !337097
  %CastPtr464 = bitcast i8* %333 to i8**, !dbg !337097
  %"loadField_<passesFilter>5__2465" = load i8*, i8** %CastPtr464, align 4, !dbg !337097
  
  %PassedFilterObj = ptrtoint i8* %"loadField_<passesFilter>5__2465" to i32
  %SS_0 = getelementptr i8, i8* %0, i32 100, !dbg !337097
  call void @S_P_CoreLib_System_Runtime_DispatchResolve__LogHex(i8* %SS_0, i32 %PassedFilterObj), !dbg !337097
  call void @S_P_CoreLib_System_Runtime_DispatchResolve__LogChar(i8* %SS_0, i16 10), !dbg !337097
  
  %CastPtr466 = bitcast i8* %Argument0_ to i8**, !dbg !337097
  %load_arg0_467 = load i8*, i8** %CastPtr466, align 4, !dbg !337097
  %334 = getelementptr i8, i8* %0, i32 100, !dbg !337097
  invoke void @corert.throwifnull(i8* %334, i8* %load_arg0_467)
          to label %TryE3 unwind label %LandingPad0, !dbg !337097

TryE3:                                            ; preds = %TryDD
  %335 = getelementptr i8, i8* %load_arg0_467, i32 44, !dbg !337097
  %CastPtr468 = bitcast i8* %335 to i8**, !dbg !337097
  %"loadField_<cad>5__5469" = load i8*, i8** %CastPtr468, align 4, !dbg !337097
  %336 = getelementptr i8, i8* %0, i32 100, !dbg !337097
  invoke void @corert.throwifnull(i8* %336, i8* %"loadField_<cad>5__5469")
          to label %TryE8 unwind label %LandingPad0, !dbg !337097

TryE8:                                            ; preds = %TryE3
  %LoadAddressOfSymbolNode470 = load i32*, i32** @__getslot__S_P_CoreLib_System_Reflection_CustomAttributeData__get_AttributeType___SYMBOL, align 4, !dbg !337097
  %get_AttributeType_slot471 = load i32, i32* %LoadAddressOfSymbolNode470, align 4, !dbg !337097
  %337 = getelementptr i8, i8* %0, i32 100, !dbg !337097
  invoke void @corert.throwifnull(i8* %337, i8* %"loadField_<cad>5__5469")
          to label %TryE8472 unwind label %LandingPad0, !dbg !337097

TryE8472:                                         ; preds = %TryE8
  %CastPtrthis473 = bitcast i8* %"loadField_<cad>5__5469" to void (i8*, i8*)***, !dbg !337097
  %ldEEType474 = load void (i8*, i8*)**, void (i8*, i8*)*** %CastPtrthis473, align 4, !dbg !337097
  %__getslot__475 = getelementptr void (i8*, i8*)*, void (i8*, i8*)** %ldEEType474, i32 %get_AttributeType_slot471, !dbg !337097
  %ld__getslot__476 = load void (i8*, i8*)*, void (i8*, i8*)** %__getslot__475, align 4, !dbg !337097
  %338 = getelementptr i8, i8* %0, i32 100, !dbg !337097
  %339 = getelementptr i8, i8* %338, i32 0, !dbg !337097
  %CastPtrTypedStore477 = bitcast i8* %339 to i8**, !dbg !337097
  store i8* %"loadField_<cad>5__5469", i8** %CastPtrTypedStore477, align 4, !dbg !337097
  invoke void %ld__getslot__476(i8* %338, i8* %Temp16_) ; GC triggered by this call
          to label %TryE8478 unwind label %LandingPad0, !dbg !337097

TryE8478:                                         ; preds = %TryE8472
  %340 = getelementptr i8, i8* %0, i32 104, !dbg !337097
  invoke void @corert.throwifnull(i8* %340, i8* %"loadField_<passesFilter>5__2465")
          to label %TryED unwind label %LandingPad0, !dbg !337097

TryED:                                            ; preds = %TryE8478
  %LoadAddressOfSymbolNode479 = load i32*, i32** @"__getslot__S_P_CoreLib_System_Func_2<S_P_CoreLib_System_Type__Bool>__Invoke___SYMBOL", align 4, !dbg !337097
  %Invoke_slot480 = load i32, i32* %LoadAddressOfSymbolNode479, align 4, !dbg !337097
  %341 = getelementptr i8, i8* %0, i32 104, !dbg !337097
  invoke void @corert.throwifnull(i8* %341, i8* %"loadField_<passesFilter>5__2465")
          to label %TryED481 unwind label %LandingPad0, !dbg !337097

TryED481:                                         ; preds = %TryED
  %CastPtrthis482 = bitcast i8* %"loadField_<passesFilter>5__2465" to i8 (i8*)***, !dbg !337097 ; Stale value use
  %ldEEType483 = load i8 (i8*)**, i8 (i8*)*** %CastPtrthis482, align 4, !dbg !337097
  %__getslot__484 = getelementptr i8 (i8*)*, i8 (i8*)** %ldEEType483, i32 %Invoke_slot480, !dbg !337097
  %ld__getslot__485 = load i8 (i8*)*, i8 (i8*)** %__getslot__484, align 4, !dbg !337097
  %342 = getelementptr i8, i8* %0, i32 104, !dbg !337097
  %343 = getelementptr i8, i8* %342, i32 0, !dbg !337097
  %CastPtrTypedStore486 = bitcast i8* %343 to i8**, !dbg !337097
  store i8* %"loadField_<passesFilter>5__2465", i8** %CastPtrTypedStore486, align 4, !dbg !337097
  %CastPtrTemp16_ = bitcast i8* %Temp16_ to i8**, !dbg !337097
  %LdTemp16_ = load i8*, i8** %CastPtrTemp16_, align 4, !dbg !337097
  %344 = getelementptr i8, i8* %342, i32 4, !dbg !337097
  %CastPtrTypedStore487 = bitcast i8* %344 to i8**, !dbg !337097
  store i8* %LdTemp16_, i8** %CastPtrTypedStore487, align 4, !dbg !337097

  %345 = invoke i8 %ld__getslot__485(i8* %342) ; We crash here
          to label %TryED488 unwind label %LandingPad0, !dbg !337097
```

yowl:
My theory is that "GC refs live in LLVM registers only"  does not happen in the IL backend, but we can have GC refs that are not included in the SS range (i.e. when GC.Collect is called, SS_bottom and SS_top do not cover the address in the load.  All the loads into LLVM registers for GC refs come from memory that should be in the SS range, that's my assertion.  For example the first load in the code is `%"loadField_<passesFilter>5__2465" = load i8*, i8** %CastPtr464, align 4` which is `SS[32]`.  It's on the SS in theory, but only if SS_top > (or >=?) SS+32

Maybe you have already spotted a load from a location that is suspect in that code?

SingleAccretion:
> the first load in the code is `%"loadField_<passesFilter>5__2465" = load i8*, i8** %CastPtr464, align 4` which is SS[32]

Hmm, I don't think that is the case. `%CastPtr464` aka `%load_arg0_463` _itself_ is loaded from the shadow stack, but `passesFilter` is then loaded from it.

(I have also included more of the relevant LLVM code which shows this above)

yowl:
Doh, you are right.  The object referenced through the field could be moved.
```cs
        private void ImportLoadField(int token, bool isStatic)
        {
            FieldDesc field = (FieldDesc)_methodIL.GetObject(token);
            FieldDesc canonFieldDesc = (FieldDesc)_canonMethodIL.GetObject(token);
            LLVMValueRef fieldAddress = GetFieldAddress(field, canonFieldDesc, isStatic);
            TypeDesc fieldTypeDesc = canonFieldDesc.FieldType;
            LLVMValueRef fieldValue = LoadValue(_builder, fieldAddress, fieldTypeDesc,
                GetLLVMTypeForTypeDesc(fieldTypeDesc), SignExtendTypeDesc(fieldTypeDesc), $"loadField_{field.Name}");

            _stack.Push(new ExpressionEntry(GetStackValueKind(fieldTypeDesc), $"Field_{field.Name}", fieldValue, fieldTypeDesc));
        }
```
Pushing  a `SpilledExpressionEntry` for GC fields might be enough here, maybe

SingleAccretion:
I will work on a minimal test for this issue and then look into the `SpilledExpressionEntry` solution; I was a bit afraid this would blow up the shadow stack size, but perhaps as long as it works well enough to not fail our tests it is fine...

yowl:
Just to throw this out there, there is a Wasm GC which Google are working on in Chrome: https://bugs.chromium.org/p/v8/issues/detail?id=7748

If it worked for us we could drop the whole SS stuff.  Would it be a better GC than dotnet's, only benchmarks would show I suppose.

SingleAccretion:
> Just to throw this out there, there is a Wasm GC which Google are working on in Chrome: https://bugs.chromium.org/p/v8/issues/detail?id=7748

> If it worked for us we could drop the whole SS stuff. Would it be a better GC than dotnet's, only benchmarks would show I suppose.

I suspect we will have to use shadow stacks for a very long time / forever. In the first approximation, WASM GC seems to be missing quite a few features we need: https://github.com/WebAssembly/gc/issues/77. In the second, we still need something for EH to work.

SingleAccretion:
Alright, while the GC hole is still to be resolved, I believe this change by itself is (finally) ready for review.

Edit: but we'll have to wait until the hole is fixed before it can be merged.

yowl:
Another big step forward!  Do you know if any of the exception based tests are now compiled via the RyuJit backend, I assume some are?

SingleAccretion:
>  Do you know if any of the exception based tests are now compiled via the RyuJit backend, I assume some are?

A good number of `HelloWasm` EH tests are compiled with RyuJit now:
```
TestThrowIfNull: WASM EH: exception thrown
Catch not called when no exception test: Ok.
Catch called when exception thrown test: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002676]
caught
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
Catch called when exception thrown from call: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002677]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
Catch called for exception type and order: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: mutually protecting catches, table at [0x00005F64]
WASM EH/RyuJit [SF: 0x03804394] [1]: Clause: catch, class [EETypePointer:0x00156EA4(System.ArgumentException)]
WASM EH/RyuJit [SF: 0x03804394] [1]: Clause: catch, class [EETypePointer:0x0015839C(System.NullReferenceException)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002700]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
Try/Finally calls finally when exception thrown test: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x038043A8] [1]: Handling: catch, class [EETypePointer:0x00156750(System.Object)]
WASM EH/RyuJit [SF: 0x038043A8] [2]: Calling catch funclet at [0x00002642]
warning: Source map information is not available, emscripten_log with EM_LOG_C_STACK will be ignored. Build with "--pre-js $EMSCRIPTEN/src/emscripten-source-map.min.js" linker flag to add source map loading to code.
WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x038043A8] [2]: Handling: fault
WASM EH/RyuJit [SF: 0x038043A8] [2]: Calling fault funclet at [0x00002643]
WASM EH/RyuJit [SF: 0x038043A8] [2]: Funclet returned: success
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002678]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
Try/Finally calls finally once when exception thrown and caught test: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804398] [1]: Handling: catch, class [EETypePointer:0x00156750(System.Object)]
WASM EH/RyuJit [SF: 0x03804398] [2]: Calling catch funclet at [0x00002640]
WASM EH/RyuJit [SF: 0x03804398] [2]: Funclet returned: 00000001
Ok.
Inner try finally called before outer catch: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [2]: Handling: fault
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling fault funclet at [0x0000267A]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: success
WASM EH/RyuJit [SF: 0x03804394] [2]: Handling (active): fault
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling fault funclet at [0x0000267B]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: success
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling (active): catch, class [EETypePointer:0x00156750(System.Object)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x0000267C]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
Test invoke when last instruction in if block: Ok.
Throw exception in catch: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156750(System.Object)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x0000267E]
WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x0000267F]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
TestExceptionInGvmCall: WASM EH: exception thrown
WASM EH/IL [SF: 0x0380439C] [1]: Pass start; at: 0000000C, exception: [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/IL [SF: 0x0380439C] [1]: EH#00000000: try [00000001..0000001A) handled by [0x00001AB9], class [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/IL [SF: 0x0380439C] [1]: Pass end; handler found at [0x00001AB9]
Ok.
Catch handler can access generic context: WASM EH: exception thrown
WASM EH/IL [SF: 0x0380439C] [1]: Pass start; at: 00000001, exception: [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/IL [SF: 0x0380439C] [1]: EH#00000000: try [00000001..00000008) handled by [0x00001ACA], class [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/IL [SF: 0x0380439C] [1]: Pass end; handler found at [0x00001ACA]
Ok.
Filter funclet can access generic context: WASM EH: exception thrown
WASM EH/IL [SF: 0x0380439C] [1]: Pass start; at: 00000001, exception: [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/IL [SF: 0x0380439C] [1]: EH#00000000: try [00000001..00000008) handled by [0x00001ACB], filter at [0x00001ACC]
WASM EH/IL [SF: 0x0380439C] [1]: Calling filter funclet at [0x00001ACC]
WASM EH/IL [SF: 0x0380439C] [1]: Funclet returned: true
WASM EH/IL [SF: 0x0380439C] [1]: Pass end; handler found at [0x00001ACB]
Ok.
TestFilter: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: filtered catch
WASM EH/RyuJit [SF: 0x03804394] [1]: Calling filter funclet at [0x00002681]
WASM EH/RyuJit [SF: 0x03804394] [1]: Funclet returned: true
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002680]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
TestFilterNested: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: filtered catch
WASM EH/RyuJit [SF: 0x03804394] [1]: Calling filter funclet at [0x00002684]
WASM EH/RyuJit [SF: 0x03804394] [1]: Funclet returned: false
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling (active): catch, class [EETypePointer:0x00156EA4(System.ArgumentException)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002685]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000002
WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: filtered catch
WASM EH/RyuJit [SF: 0x03804394] [1]: Calling filter funclet at [0x00002684]
WASM EH/RyuJit [SF: 0x03804394] [1]: Funclet returned: false
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling (active): catch, class [EETypePointer:0x00156EA4(System.ArgumentException)]
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling (active): filtered catch
WASM EH/RyuJit [SF: 0x03804394] [1]: Calling filter funclet at [0x00002687]
WASM EH/RyuJit [SF: 0x03804394] [1]: Funclet returned: true
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002686]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: filtered catch
WASM EH/RyuJit [SF: 0x03804394] [1]: Calling filter funclet at [0x00002684]
WASM EH/RyuJit [SF: 0x03804394] [1]: Funclet returned: true
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002683]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000003
In middle catchRunning outer filterIn outer catchRunning inner filterIn inner catch
Ok.
Test catch and throw different exception: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156750(System.Object)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002688]
WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x00002689]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
Test rethrow: WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156750(System.Object)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x0000268A]
WASM EH: exception thrown
WASM EH/RyuJit [SF: 0x03804394] [1]: Handling: catch, class [EETypePointer:0x00156A9C(System.Exception)]
WASM EH/RyuJit [SF: 0x03804394] [2]: Calling catch funclet at [0x0000268B]
WASM EH/RyuJit [SF: 0x03804394] [2]: Funclet returned: 00000001
Ok.
```
The `Exceptions.cs` test doesn't use RyuJit yet (probably because it creates a delegate).

yowl:
Looks impressive, do you expect https://stackoverflow.com/questions/57544142/in-c-sharp-try-finally-how-to-catch-the-original-exception/57544409#57544409 to pass with this (assuming there was no other reason why it fell back to the IL backend) ?

SingleAccretion:
No, not currently. We are still doing per-frame dispatch here because attempting to not do this will run into problem with the IL backend catching in-progress exceptions. But it is certainly the case this design should be capable of doing the whole first pass before the second pass (PR's description has a schematic for how it will work).

For full inter-frame dispatch, the currently missing piece is shadow stack from the throw frame, which is expected to be carried by the native exception. The rest of the system should then "just work".

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -132,47 +134,50 @@ bool Llvm::initializeFunctions()
...
-                kindName = "Fault";
-                break;
-            case EH_HANDLER_FINALLY:
-                kindName = "Finally";
-                break;
-            default:
-                unreached();
-        }
+            // All funclets have two arguments: original and actual shadow stacks.
+            Type* ptrLlvmType = Type::getInt8PtrTy(_llvmContext);
```
yowl: Can we use the unqualified pointer type here?

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -398,7 +377,8 @@ void Llvm::generateEHDispatch()
     // Recover the C++ personality function.
     Type* ptrLlvmType = Type::getInt8PtrTy(_llvmContext);
```
yowl: opaque pointer?

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -410,54 +390,501 @@ void Llvm::generateEHDispatch()
...
+        Function* llvmFunc = getLlvmFunctionForIndex(funcIdx);
         if (!llvmFunc->hasPersonalityFn())
         {
             llvmFunc->setPersonalityFn(gxxPersonalityLlvmFunc);
         }
 
-        // Move it after the last block in the corresponding protected region.
-        llvmBlock->moveAfter(getLastLlvmBlockForBlock(ehDsc->ebdTryLast));
+        // The code we will generate effectively inlines the usual runtime dispatch logic. The main reason this
+        // scheme was chosen is the fact (re)throwing funclets are handled by it semlessly and efficiently. The
```
yowl:
```suggestion
        // scheme was chosen is the fact (re)throwing funclets are handled by it seamlessly and efficiently. The
```

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
```diff
@@ -2,19 +2,178 @@
...
+        }
+
+        private struct CppExceptionTuple
+        {
+            public void* ExceptionData;
+            public int Selector;
+        }
+
+        // These per-clause handlers are invoked by RyuJit-generated LLVM code. TODO-LLVM: once we move to opaque
+        // pointers, we can change the signatures of these functions to use precise pointer types instead of "void*"s.
```
yowl: Is this possible now?  Not that it needs to be in this PR.

yowl:
Some minor comments.  I think that the fact the tests are passing, is as good a proof (and clearly better) than me eyeballing the code that this is awesome!

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
SingleAccretion: Yes; addressed.

src/coreclr/jit/llvmcodegen.cpp
SingleAccretion: Addressed in d5e9d29132ad3caaad6148c75678688f6de800c8.

src/coreclr/jit/llvmcodegen.cpp
SingleAccretion: Addressed in https://github.com/dotnet/runtimelab/commit/d5e9d29132ad3caaad6148c75678688f6de800c8.

src/coreclr/jit/llvmcodegen.cpp
SingleAccretion: Yes; addressed in https://github.com/dotnet/runtimelab/commit/d5e9d29132ad3caaad6148c75678688f6de800c8.

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
```diff
@@ -2,19 +2,178 @@
...
```
SingleAccretion:
```suggestion
        // These per-clause handlers are invoked by RyuJit-generated LLVM code.
```

jkotas:
@yowl Is this good to merge?

yowl:
Yes absolutely

jkotas:
@SingleAccretion Thank you! This is great work

PR #2149, opened 2023-01-05
NativeAOT-LLVM: test PR for Exception handling dispatch upstream

yowl:
This is test PR for the CI for some changes that may fix some memory corruption


SingleAccretion:
Thank you for helping with this!

I would revert everything I did after and including `Testing...`, the issue is known to reproduce with the HelloWasm test fully intact.

yowl:
This failed with 
```
Failed.
object alive when has no references-
```
Which I remember now happens with these changes.  An object held only be a weak ref is not getting GC'ed.  It hasn't crashed though.  I can either investigate that or I can remove changes until it crashes?

SingleAccretion:
No crash does support the hypothesis of these changes resolving the underlying cause. Let's leave this change here for now; I hope the more expedient way of reducing the test case in #2145 will get to the exact corruption point in a relatively timely manner.

yowl:
Ok, good luck.  If there's a GC object, referenced from heap 2, but not marked, then it can be a bar of soap.  I don't know that is the problem of course.  GC heap verification can also help.

PR #2154, opened 2023-01-06
[NativeAOT-LLVM] Delete pointee type related Jit/EE APIs

SingleAccretion:
No longer needed with opaque pointers.


jkotas (approved):
LGTM

PR #2158, opened 2023-01-07
NativeAOT-LLVM: Merge to April 22

yowl:
to commit 1a31bf638c6c220d20ef65f43a07a9ac562d92d9


yowl:
Cc @dotnet/nativeaot-llvm

yowl:
I need to revert the Ilc.sln change.  Aside from normal merge resolve changes there is support for Methodtable.Of<> and something changed regards constrained calls that is not resolved resulting in the removal of those tests.

eng/Subsets.props
```diff
@@ -242,8 +242,8 @@
       Category="clr" />
   </ItemGroup>
 
-  <!-- Build the CoreCLR cross-arch tools when we're doing a cross-architecture build and either we're building any CoreCLR native tools for platforms CoreCLR fully supports or when someone explicitly requests them -->
-  <ItemGroup Condition="(('$(ClrRuntimeBuildSubsets)' != '' and '$(PrimaryRuntimeFlavor)' == 'CoreCLR' and '$(TargetsMobile)' != 'true') or $(_subset.Contains('+clr.crossarchtools+'))) and '$(BuildArchitecture)' != '$(TargetArchitecture)'">
+  <!-- Build the CoreCLR cross-arch tools when we're doing a cross-architecture build and either we're building any CoreCLR native tools for platforms CoreCLR fully supports or when someone explicitly requests them, but never for Browser/Wasm -->
+  <ItemGroup Condition="(('$(ClrRuntimeBuildSubsets)' != '' and '$(PrimaryRuntimeFlavor)' == 'CoreCLR' and '$(TargetsMobile)' != 'true') or $(_subset.Contains('+clr.crossarchtools+'))) and '$(BuildArchitecture)' != '$(TargetArchitecture)' and !('$(TargetOS)' == 'Browser' and '$(TargetArchitecture)' == 'wasm')">
```
SingleAccretion: What in our build requests `clr.crossarchtools`?

eng/native/configureplatform.cmake
```diff
@@ -392,10 +392,10 @@ if(CLR_CMAKE_TARGET_OS STREQUAL SunOS)
     set(CLR_CMAKE_TARGET_SUNOS 1)
 endif(CLR_CMAKE_TARGET_OS STREQUAL SunOS)
 
-if(CLR_CMAKE_TARGET_OS STREQUAL Emscripten)
+if(CLR_CMAKE_HOST_OS STREQUAL Emscripten)
```
SingleAccretion: Everything else here is checking "if (we were told the target is XYZ) SetTargetXYZ()", why does emscripten need to be different? It feels like something upstream (e. g. build scripts) isn't behaving correctly.

src/coreclr/build-runtime.cmd
```diff
@@ -181,46 +167,25 @@ set __ExtraCmakeArgs="-DRUNTIME_FLAVOR=CoreCLR"
...
-=======
 if %__TargetArchX64%==1   set __TargetArch=x64
 if %__TargetArchX86%==1   set __TargetArch=x86
 if %__TargetArchArm%==1   set __TargetArch=arm
 if %__TargetArchArm64%==1 set __TargetArch=arm64
+if %__TargetArchWasm%==1 (
+echo setting wasm
+    set __TargetOS=browser
+    set __TargetArch=wasm
+)
```
SingleAccretion:
```suggestion
if %__TargetArchWasm%==1 (
    set __TargetOS=browser
    set __TargetArch=wasm
)
```

src/coreclr/jit/gentree.h
```diff
@@ -8329,12 +8213,8 @@ inline regNumber GenTree::GetRegByIndex(int regIndex) const
         return AsPutArgSplit()->GetRegNumByIdx(regIndex);
     }
 #endif
-<<<<<<< HEAD
-#if !defined(TARGET_64BIT) && !defined(TARGET_WASM32) && !defined(TARGET_WASM64)
-=======
 
-#if !defined(TARGET_64BIT)
->>>>>>> 1a31bf638c6c220d20ef65f43a07a9ac562d92d9
+#if !defined(TARGET_64BIT) && !defined(TARGET_WASM32) && !defined(TARGET_WASM64)
```
SingleAccretion: Why is `!defined(TARGET_WASM32) && !defined(TARGET_WASM64)` needed? This is already under `#if FEATURE_MULTIREG_RET` which is `0` for WASM.

src/coreclr/jit/gentree.h
```diff
@@ -8396,14 +8276,9 @@ inline var_types GenTree::GetRegTypeByIndex(int regIndex) const
...
     }
-<<<<<<< HEAD
-#endif
-#if !defined(TARGET_64BIT) && !defined(TARGET_WASM32) && !defined(TARGET_WASM64)
-=======
 #endif // FEATURE_ARG_SPLIT
 
-#if !defined(TARGET_64BIT)
->>>>>>> 1a31bf638c6c220d20ef65f43a07a9ac562d92d9
+#if !defined(TARGET_64BIT) && !defined(TARGET_WASM32) && !defined(TARGET_WASM64)
```
SingleAccretion: Ditto: `!defined(TARGET_WASM32) && !defined(TARGET_WASM64)`?

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -880,7 +880,7 @@ void Llvm::buildStoreLocalVar(GenTreeLclVar* lclVar)
 
     if (lclVar->HasSsaName())
     {
-        _localsMap.Set({lclNum, lclVar->GetSsaNum()}, localValue);
+        _localsMap.Set({lclNum, lclVar->GetSsaNum()}, localValue, JitHashTable<Compiler::SSAName, Compiler::SSAName, Value*>::Overwrite);
```
SingleAccretion: This is not right. It indicates we are visiting the same SSA definition twice, which is not legal. What is the cause?

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -892,21 +891,17 @@ void SsaBuilder::RenameDef(GenTree* defNode, BasicBlock* block)
...
             else
 #endif
-                asgPhiOp = asgNode->gtGetOp2();
+                asgPhiOp = defNode;
 
-            if (!asgPhiOp->OperIs(GT_PHI))
-=======
-            if (!defNode->IsPhiDefn())
->>>>>>> 1a31bf638c6c220d20ef65f43a07a9ac562d92d9
+            if (!asgPhiOp->IsPhiDefn())
```
SingleAccretion: This is not quite right for `#ifdef TARGET_WASM`, since `asgPhiOp` will be `Data()`, i. e. not the actual definition node.

I think we can simplify the whole thing down to just `!defNode->IsPhiDefn()`:
```diff
-            GenTree* asgPhiOp;
-#if defined(TARGET_WASM)
-            if (block->IsLIR())
-            {
-                asgPhiOp = defNode->gtGetOp1();
-            }
-            else
-#endif
-                asgPhiOp = defNode;
-
-            if (!asgPhiOp->IsPhiDefn())
+            // NativeAOT-LLVM: revert the deletion of STORE_LCL_VAR handling in "IsPhiDefn" when merging https://github.com/dotnet/runtime/pull/76139 (Oct 20)
+            if (!defNode->IsPhiDefn())
             {
                 AddDefToHandlerPhis(block, lclNum, ssaNum);
             }
```

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
```diff
@@ -5228,6 +5242,18 @@ private void ImportFallthrough(BasicBlock next)
         private const string ThreadStatics = "ThreadStatics";
         private const string ClassConstructorRunner = "ClassConstructorRunner";
 
+        private ExpressionEntry LoadInstanceField(LLVMValueRef thisValueRef, FieldDesc fieldDesc, TypeDesc fieldTypeDesc, StackValueKind stackValueKind, LLVMBuilderRef builder = default(LLVMBuilderRef))
+        {
+            if (builder.Handle == IntPtr.Zero)
+                builder = _builder;
+
+            LLVMValueRef fieldAddress = _builder.BuildGEP(thisValueRef, new LLVMValueRef[] { LLVMValueRef.CreateConstInt(LLVMTypeRef.Int32, (ulong)fieldDesc.Offset.AsInt) });
```
SingleAccretion, line 5250: Is `thisValueRef` guaranteed to be `i8*` here s.t. GEP adds the right offset?

src/tests/nativeaot/SmokeTests/UnitTests/Generics.cs
```diff
@@ -20,7 +20,9 @@ internal static int Run()
         TestDelegateVirtualMethod.Run();
         TestDelegateInterfaceMethod.Run();
         TestThreadStaticFieldAccess.Run();
+#if !CODEGEN_WASM // TODO-LLVM: PrecomputedDictionaryLayoutNode.GetSlotForEntry System.InvalidOperationException: [UnitTests]Generics+TestConstrainedMethodCalls.DoFrob<Foo`1<__Canon>,__Canon>(Foo`1<__Canon>&,object): TypeHandle: T_.Foo`1<System.__Canon>
         TestConstrainedMethodCalls.Run();
```
SingleAccretion: Presumably the cause of `.constrained` handling perturbations is https://github.com/dotnet/runtime/pull/66936, but it is not trivial to tell what needs to be changed for things to work again at a glance...

Hopefully we will just be able to re-enable the tests with the IL backend gone.

eng/native/configureplatform.cmake
yowl: It was just because sometimes the taget is `Browser` and sometimes `browser`.  I did have some lower case logic in here at one point, i could put that back instead of this?

SingleAccretion: Hmm, so would the following would be equivalent?
```
if(CLR_CMAKE_TARGET_OS STREQUAL browser OR CLR_CMAKE_TARGET_OS STREQUAL Browser)
    set(CLR_CMAKE_TARGET_UNIX 1)
    set(CLR_CMAKE_TARGET_BROWSER 1)
endif(CLR_CMAKE_TARGET_OS STREQUAL browser OR CLR_CMAKE_TARGET_OS STREQUAL Browser)
```
If so it would be better to do this (or normalize somewhere above).

Evidently the convention here is to require exact casing of target OSes when passing them to CMake, so "the best" solution would be to make sure it is always `Browser` but I do not know how costly that would be.

yowl: Yes, that should be equivalent.  I will at least investigate where `browser` is coming from.

SingleAccretion: One place that's immediately apparent is `build-runtime.cmd` here:

https://github.com/dotnet/runtimelab/blob/10a88e02a5b239693af0f63a060659818d03ec35/src/coreclr/build-runtime.cmd#L183-L186

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs
yowl: Yes, I think we are ok, It is only set from one path:
```
thisRef = thisPointer.ValueAsType(LLVMTypeRef.CreatePointer(LLVMTypeRef.Int8, 0), _builder);
```

https://github.com/yowl/runtimelab/blob/merge-apr-22/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/ILToLLVMImporter.cs#L2055

src/coreclr/jit/ssabuilder.cpp
yowl: Thanks, looks fine.

src/coreclr/jit/gentree.h
yowl: removed

src/coreclr/jit/gentree.h
yowl: removed

src/coreclr/jit/llvmcodegen.cpp
yowl: Dump attached `Object__GetHashCode`
[dump.txt](https://github.com/dotnet/runtimelab/files/10376220/dump.txt)

What happens is that the prolog generation for 
```
Setting V02's initial value to uninit
  %2 = freeze i32 undef
```
Sets lcl/ssa 2:1
And then 
```
Generating: N004 ( 24,  8) [000005] DA-XG-------              *  STORE_LCL_VAR int    V02 tmp0         d:1
```
fails setting the same key.  Is `lvaInSsa` returning true now in the prolog and it wasn't before?  I could check that.

SingleAccretion: No need... The bug is here:
```
unsigned ssaNum = varDsc->lvPerSsaData.AllocSsaNum(m_allocator, block,
                                                   defNode->OperIs(GT_ASG) ? defNode->AsOp() : nullptr);
```
It needs to record the SSA def node for `STORE_LCL_VAR` as well. Flipping the condition from `defNode->OperIs(GT_ASG) ? defNode->AsOp() : nullptr)` to `!defNode->OperIs(GT_CALL) ? defNode->AsOp() : nullptr)` should achieve this.

yowl: I see, should have been no need for initialisation. Thanks

eng/native/configureplatform.cmake
yowl: Looks like that was the only one that needs changing

eng/Subsets.props
yowl: Its `'$(ClrRuntimeBuildSubsets)' != '' and '$(PrimaryRuntimeFlavor)' == 'CoreCLR' and '$(TargetsMobile)' != 'true'` that evaluates true. (and `$(BuildArchitecture)' != '$(TargetArchitecture)`

yowl:
A crash appeared in one of the runs around MD array types and then disappeared on its own.  Strange, but I guess we leave it.

SingleAccretion:
> A crash appeared in one of the runs around MD array types and then disappeared on its own. Strange, but I guess we leave it.

With the same stack as I am investigating, presumably (hopefully) the same underlying cause as well.

eng/Subsets.props
SingleAccretion: I see. A nit, but it would be clearer to move this to be logically under the "CoreCLR doesn't fully support this platform" umbrella then:
```
  <!-- Build the CoreCLR cross-arch tools when we're doing a cross-architecture build and either we're building any CoreCLR native tools for platforms CoreCLR fully supports or when someone explicitly requests them -->
  <ItemGroup Condition="(('$(ClrRuntimeBuildSubsets)' != '' and '$(PrimaryRuntimeFlavor)' == 'CoreCLR' and '$(TargetsMobile)' != 'true' and !('$(TargetOS)' == 'Browser' and '$(TargetArchitecture)' == 'wasm')) or $(_subset.Contains('+clr.crossarchtools+'))) and '$(BuildArchitecture)' != '$(TargetArchitecture)'">
```
If someone "explicitly requests" the cross arch tools they will get build failures (most likely) but that's by design.

SingleAccretion (approved):
LGTM (modulo one nit)!

yowl:
nit done, thanks

PR #2159, opened 2023-01-07
NativeAOT-LLVM: Remove nmake from wasm build

yowl:
For some historical reason nmake was used for the wasm side of the build, this PR removes that and uses the default of Ninja on Windows, removing one small annoyance when merging.

cc @dotnet/nativeaot-llvm


SingleAccretion (approved):
Nice!

Issue #2121, opened 2022-12-16
[NativeAOT-LLVM] Publishing Hello World for WebAssembly target on Windows arm64 fails


ChristianWeyer:
When trying to publish the _Hello World_ project (https://github.com/dotnet/runtimelab/tree/feature/NativeAOT-LLVM/samples/HelloWorld) for WebAssembly target on a Windows ARM machine, with this command:

```
dotnet publish -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false --self-contained
```

I get this error:
```
MSBuild version 17.4.0+18d5aef85 for .NET
  Determining projects to restore...
  Restored C:\Users\christianweyer\source\repos\ConsoleApp1\ConsoleApp1\ConsoleApp1.csproj (in 41.75 sec).
  ConsoleApp1 -> C:\Users\christianweyer\source\repos\ConsoleApp1\ConsoleApp1\bin\Debug\net7.0\browser-wasm\ConsoleApp1.dll
  Optimizing assemblies for size may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
  Generating compatible native code. To optimize for size or speed, visit https://aka.ms/OptimizeCoreRT
**EXEC : error : Unable to load DLL 'clrjit_browser_wasm32_arm64' or one of its dependencies: The specified module could not be found. (0x8007007E) [C:\Users\christianweyer\source\repos\ConsoleApp1\C
onsoleApp1\ConsoleApp1.csproj]**
  System.DllNotFoundException: Unable to load DLL 'clrjit_browser_wasm32_arm64' or one of its dependencies: The specified module could not be found. (0x8007007E)
     at System.Runtime.InteropServices.NativeLibrary.LoadByName(String libraryName, QCallAssembly callingAssembly, Boolean hasDllImportSearchPathFlag, UInt32 dllImportSearchPathFlag, Boolean throwO
  nError)
     at System.Runtime.InteropServices.NativeLibrary.LoadLibraryByName(String libraryName, Assembly assembly, Nullable`1 searchPath, Boolean throwOnError)
     at System.Runtime.InteropServices.NativeLibrary.Load(String libraryName, Assembly assembly, Nullable`1 searchPath)
[...]

PR #2161, opened 2023-01-10
NativeAOT-LLVM: Fix publishing of additional wasm files

yowl:
This PR addresses the comment at https://github.com/dotnet/runtimelab/issues/2121#issuecomment-1362271101 where the js and wasm files do not get copied to the publish folder.

It changes condition of the additional copies to use TargetArchitecure=wasm.  (Platform is AnyCPU now)

cc @dotnet/nativeaot-llvm


SingleAccretion (approved):

yowl:
@MichalStrehovsky  I don't know what I clicked to tag you here...  You can ignore it, it wasn't intentional

SingleAccretion:
I suspect it is automatic tagging based on the code owners file. We should probably delete that / mute it / something else.

MichalStrehovsky:
> I suspect it is automatic tagging based on the code owners file. We should probably delete that / mute it / something else.

Yup, the set-up a new experiment doc suggests deleting it: https://github.com/dotnet/runtimelab/blob/docs/CreateAnExperiment.md

I would probably rename it instead to avoid edit-delete merge conflicts.

PR #2164, opened 2023-01-11
Patch a GC hole with SDSUs in the IL backend

SingleAccretion:
Spill GC values for a few select operators to the shadow stack to make them visible to the GC scan.

Includes a test.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl:
That looks good to me!

PR #2167, opened 2023-01-12
Fix package publishing

jkotas:
The condition to detect official builds changed in dotnet/runtime, and it is not detecting official runtimelab build correctly anymore. Revert to the original condition.

### DIFF
--- a/eng/pipelines/common/variables.yml
+++ b/eng/pipelines/common/variables.yml
@@ -10,7 +10,7 @@ variables:
   value: 1000
 
 - name: isOfficialBuild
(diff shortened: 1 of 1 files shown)


yowl:
I think I'll get a conflict with d9433fb4bf59490e928195e5efedaad6d76a69d9 at some point, which one should I take at that point?

jkotas:
`value: ${{ and(eq(variables['System.TeamProject'], 'internal'), eq(variables['Build.DefinitionName'], 'dotnet.runtimelab-official')) }}`  should work.

Issue #2165, opened 2023-01-11
[NativeAOT-LLVM] Runtime tests generate a significant number of warnings


SingleAccretion:
Let's get at least `HelloWasm` clean, eleven warnings is too high a number to ignore.

Some of these are upstream problems we've picked up, but most should be actionable directly.

Current list of warnings:

1\) Four `MSB4011` - https://github.com/dotnet/runtime/issues/70658
  - Introduced by the manual inclusion of `test_dependencies.*.props` in https://github.com/dotnet/runtime/pull/63978. There are now two NuGet-related `.props` that are almost identical. The batched build does its own restore and so isn't affected.
  - Upstream fix: https://github.com/dotnet/runtime/pull/80573.

2\) Two `IL2070`.
  - One DAM will solve the problem.

3\) `clang++ : warning : argument unused during compilation: '-I C:\Users\Accretion\source\dotnet\emsdk\upstream\emscripten\cache\sysroot\include\SDL'`.
  - Emscripten adds SDL headers by default, and this triggers the warnings (since, presumably, nothing references the code in them). We can pass `sUSE_SDL=0` ourselves to disable the default behavior, which could then go away once we update to emscripten which has https://github.com/emscripten-core/emscripten/pull/18443.

4\) `EXEC : warning : undefined symbol: S_P_CoreLib_Internal_Runtime_CompilerHelpers_ThrowHelpers__ThrowBodyRemoved` - appeared after the recent merge.
  - We are using the wrong helper for tentative instance method stubs.

5\) `EXEC : warning : undefined symbol: dotnet_browser_entropy` - good old `dotnet_browser_entropy`.
  - Presumably there because System.Native references it.

6\) `emcc : warning : warnings in JS library compilation [-Wjs-compiler]`.
  - Fallout from the undefined symbols warnings.

SingleAccretion:
FYI @yowl

PR #2168, opened 2023-01-13
 [NativeAOT-LLVM] Fix most warnings for the `HelloWasm` runtime test

SingleAccretion:
Only two, related to the missing `dotnet_browser_entropy` symbol, remain.

Contributes to #2165.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeInstanceMethodNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeInstanceMethodNode.cs
@@ -35,11 +35,11 @@ namespace ILCompiler.DependencyAnalysis
 
         public override bool HasConditionalStaticDependencies => true;
 
-        protected override ISymbolNode GetTarget(NodeFactory factory)
+        public override IMethodNode GetTarget(NodeFactory factory)
         {
             // If the class library doesn't provide this helper, the optimization is disabled.
             MethodDesc helper = factory.InstanceMethodRemovedHelper;
-            return helper == null ? RealBody: factory.MethodEntrypoint(helper);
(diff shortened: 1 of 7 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeInstanceMethodNode.cs
```diff
@@ -35,11 +35,11 @@ public TentativeInstanceMethodNode(IMethodBodyNode methodNode)
 
         public override bool HasConditionalStaticDependencies => true;
 
-        protected override ISymbolNode GetTarget(NodeFactory factory)
+        public override IMethodNode GetTarget(NodeFactory factory)
```
yowl, line 38: This visibility change seems harmless to me, but maybe @jkotas you'd like to cast an eye on it?

yowl (approved):
LGTM, modulo if @jkotas has any comment on the protected/public comment

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TentativeInstanceMethodNode.cs
jkotas: It looks good to me. @MichalStrehovsky Any concerns?

MichalStrehovsky: Nope, looks good. All AssemblyStubNode descendants need some special treatment because on WASM we're not able to generate byte buffers with extra code in it (which is what we do elsewhere).

PR #2170, opened 2023-01-15
 [NativeAOT-LLVM] Implement a few more opcodes, restore IL tests and fix bugs

SingleAccretion:
Please refer to the commit list for the precise set of changes.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (commented):
Looks good, thanks!

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1835,6 +1907,20 @@ void Llvm::buildBoundsCheck(GenTreeBoundsChk* boundsCheckNode)
...
 
+void Llvm::buildCkFinite(GenTreeUnOp* ckNode)
+{
+    assert(varTypeIsFloating(ckNode));
+    Type* fpLlvmType = getLlvmTypeForVarType(ckNode->TypeGet());
+    Value* opValue = consumeValue(ckNode->gtGetOp1(), fpLlvmType);
+
+    // Taken from IR Clang generates for "isfinite".
+    Value* absOpValue = _builder.CreateIntrinsic(llvm::Intrinsic::fabs, fpLlvmType, opValue);
+    Value* isNotFiniteValue = _builder.CreateFCmpUEQ(absOpValue, llvm::ConstantFP::get(fpLlvmType, INFINITY));
```
jkotas, line 1918: Is this doing the right thing for NaN?

jkotas: I see that there is a test case for this, so it presumably works.

SingleAccretion: > Is this doing the right thing for NaN?

Yes, unordered compares return `true` if any one of the operands is `NaN`.

yowl: I think we have tests for it, but looking at them I think this should be `||` 
https://github.com/SingleAccretion/runtimelab/blob/c59764fb6cef967f06edcf3deab73574f2b498f8/src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs#L2714
I hope we are getting away with that as they are all false, so `||` or `&&` is coming to the same thing.

PR #2171, opened 2023-01-16
 [NativeAOT-LLVM] Simplify LLVM object writer

SingleAccretion:
The writer seems to have been written as an amalgam of its native version and (presumably) the C++ codegen version, inheriting some dead code and unnecessary complexity. This change removes some of that.

It also adds a few comments and reorders members of the class in a more logical fashion.

I would recommend reviewing the change commit-by-commit (possibly skipping the last one altogether).


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
This certainly was an old and neglected side of the code.  Thanks for sorting it out.

PR #2172, opened 2023-01-17
NativeAOT-LLVM: Merge to May 22

yowl:
Just a small step forward as there are quite a few conflicts.


SingleAccretion:
Hi @yowl, it appears commits between 8th of April (where we are now in terms of upstream, i. e. the last commit merged in #2158) and 23rd of April (https://github.com/dotnet/runtimelab/pull/2172/commits/75ee11c68ef1481712a7d8b0ef67fe76b0709eae, first commit here) are missing.

https://github.com/dotnet/runtime/pull/67238 is among those commits. If it would not be too bothersome, I would appreciate it if we merged that one as a separate PR. It is impactful to our backend (implies removing the `PUTARG_TYPE` mechanism we've been using so far).

yowl:
I wonder how that happens.  Sure, I can merge https://github.com/dotnet/runtime/pull/67238 first on its own.

PR #2173, opened 2023-01-17
[NativeAOT-LLVM] Remove unnecessary symbol indirections

SingleAccretion:
Up until this change, we used a scheme where each `ISymbolNode` was referenced via a load from a `__SYMBOL` shadow symbol containing its address. With this change, we use LLVM's global alias feature instead and reference things directly, removing the unnecessary indirection.

Happily, that means removing more code from the object writer.


SingleAccretion:
OSX x64 failure is an upstream problem and can be ignored.

@dotnet/nativeaot-llvm

yowl (approved):
Nice reduction, thanks

PR #2175, opened 2023-01-18
NativeAOT-LLVM: Merge up to but not including PR67238

yowl:
This PR merges up to but does not include https://github.com/dotnet/runtime/pull/67238

Mostly TARGET_LOONGARCH64 changes plus removal of `Lowering::TransformUnusedIndirection` for Wasm


yowl:
2 OSX failures, the Release one looks the same as mentioned in https://github.com/dotnet/runtimelab/pull/2173#issuecomment-1385832016.  The Debug one looks like an undefined reference to "_PAL_fprintf" which I'll have a look at.  Conflict resolutions can be reviewed, along with the exclusion of `TransformUnusedIndirection ` for `TARGET_WASM`

yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/jit/instr.cpp
```diff
@@ -1589,7 +1589,6 @@ instruction CodeGen::ins_Copy(var_types dstType)
     {
         return INS_mov;
     }
-<<<<<<< HEAD
 #elif defined(TARGET_WASM)
```
SingleAccretion: Is `instr.cpp` compiled for WASM at all? If not (which I think is the case), we should just delete all `TARGET_WASM`s from it.

src/coreclr/jit/lower.cpp
```diff
@@ -5922,11 +5922,7 @@ GenTree* Lowering::LowerConstIntDivOrMod(GenTree* node)
             return nullptr;
         }
 
-<<<<<<< HEAD
-#if defined(TARGET_XARCH) || defined(TARGET_ARM64) || defined(TARGET_WASM) // TODO Wasm
-=======
-#if defined(TARGET_XARCH) || defined(TARGET_ARM64) || defined(TARGET_LOONGARCH64)
->>>>>>> 6543a048d7242ddf204f2e1ba0723d27c02bdfc7
+#if defined(TARGET_XARCH) || defined(TARGET_ARM64) || defined(TARGET_WASM) || defined(TARGET_LOONGARCH64) // TODO Wasm
```
SingleAccretion: Same question as with `instr.cpp`.

src/coreclr/jit/unwind.cpp
```diff
@@ -447,13 +447,10 @@ UNATIVE_OFFSET Compiler::unwindGetCurrentOffset(FuncInfoDsc* func)
 
 // See unwindX86.cpp
 
-<<<<<<< HEAD
 #elif defined(TARGET_WASM) // TODO
```
SingleAccretion: Let's change it like so:
```
#elif defined(TARGET_WASM)

// No unwind info on WASM.

```

src/coreclr/tools/aot/ilc.sln
```diff
@@ -18,9 +18,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ILCompiler.LLVM", "ILCompil
 EndProject
```
SingleAccretion, line 18: Solution changes - to be reverted?

SingleAccretion (commented):
I would be tempted to disable OSX x64 Release in CI until we merge [upstream fixes for it](https://github.com/dotnet/runtime/pull/75421).

src/coreclr/tools/aot/ilc.sln
yowl: yes, I don't know why VS does this all the time for me.  I'll clear my .vs folder.

yowl: I see, VS wants to create the configuration entries for the 2 vcxproj jit entries which are missing.  Does that not happen for you @SingleAccretion ?

SingleAccretion: It certainly does add things, but I always `git restore` + `Ignore` them (so not sure what they actually are).

Would be no objections from me for checking it in and seeing whether VS holds its peace from there.

src/coreclr/jit/lower.cpp
```diff
@@ -5922,7 +5922,7 @@ GenTree* Lowering::LowerConstIntDivOrMod(GenTree* node)
             return nullptr;
         }
 
-#if defined(TARGET_XARCH) || defined(TARGET_ARM64) || defined(TARGET_WASM) || defined(TARGET_LOONGARCH64) // TODO Wasm
+#if defined(TARGET_XARCH) || defined(TARGET_ARM64) || defined(TARGET_LOONGARCH64) // TODO Wasm
```
SingleAccretion:
```suggestion
#if defined(TARGET_XARCH) || defined(TARGET_ARM64) || defined(TARGET_LOONGARCH64)
```

yowl:
I removed the OSX Debug build for the fprintf issue, but I will try putting it back as it could be related to https://github.com/dotnet/runtime/pull/64631

src/coreclr/jit/instr.cpp
yowl: No, its not used, have removed as suggested

src/coreclr/jit/lower.cpp
yowl: No, its not used, have removed as suggested

src/coreclr/jit/unwind.cpp
yowl: Done, thanks

yowl:
This is ready for final review, thanks.

SingleAccretion (approved):
Looks good!

PR #2177, opened 2023-01-20
NativeAOT-LLVM: merge PR 67238 - JIT: Refactor call argument representation

yowl:
This PR merges runtime PR https://github.com/dotnet/runtime/pull/67238 where the call args are refactored into `callNode->gtArgs`  .  Most of the work is in `lowerCallToShadowStack`.

The "special" LLVM args, shadow stack, and return slot are now inserted in reverse order at the end using `PushFront`.


yowl:
@SingleAccretion `GT_PUTARG_TYPE` is not removed in this PR, but it is the intention that it is removed later.  Are we going to remove `GT_PUTARG_TYPE` when lowering calls, i.e. remove calls to `gtNewPutArgType`, with this PR or later when `GT_PUTARG_TYPE` is removed upstream?

SingleAccretion:
To remove `GT_PUTARG_TYPE` we need https://github.com/dotnet/runtime/pull/68748 first, otherwise there is no place to store the signature types, so I suppose the latter.

yowl:
As it looks harder than just `callNode->gtCallArgs = nullptr` to start with new args, I'm thinking there are a couple of options for `lowerCallToShadowStack` 

* capture the existing set of args somehow, remove the existing args using `callNode->gtArgs.Remove()` and add new ones
* create a new `GenTreeCall` node and build using the old one as the source (instead of `argInfo->ArgTable`)

Did you have any thoughts about what would be best?

SingleAccretion:
Recreating the argument list should no longer be necessary. Only the underlying nodes need to be changed.

yowl:
I have a question about morphing the args.  I'm hitting the assert at 

https://github.com/yowl/runtimelab/blob/d1a050268b7a41d4e529baf28203576825a8200b/src/coreclr/jit/morph.cpp#L8910

Because `this` had not been morphed into a temp.  Tracing the execution, this seems to be because, although `this` is identified as needing a temp and get `m_needTmp` set `true`, the execution goes into 

https://github.com/yowl/runtimelab/blob/pr67238/src/coreclr/jit/morph.cpp#L1589-L1590

And skips the creation of the temp.  The comment above this says " On x86 and other archs that use push instructions to pass arguments:" .

We set `REG_STK` for `this` at 

https://github.com/yowl/runtimelab/blob/pr67238/src/coreclr/jit/morph.cpp#L3020

I don't know if we should be looking to see how x86 gets past the assert, or for `TARGET_WASM` we should be ignoring the `arg.AbiInfo.GetRegNum() == REG_STK` check and creating the temp?

(dump attached)
[dump.txt](https://github.com/dotnet/runtimelab/files/10472492/dump.txt)

SingleAccretion:
Upstream assumes all non-local `this` args are forced into temps because they are always late, which is in turn assumed because all targets pass `this` in a register.

We used to get away with this because of the special handling of `this` that forced it into a register, but now that special handling is gone.

Now, let's consider the case where `this` would be passed on stack via `push`. We would still not be able to evaluate it into a temp, since that would change `push` order:
```
push this
push <other args>
->
push <other args>
push this
```
Essentially, to put `push`ed args into temps, `EvalArgsToTemps` would need to do what `fgMakeOutgoingStructArgCopy` does i. e. put the expression into a `COMMA` instead of using the late arg list. It does not do that.

Let's now consider the shadow stack ABI. Here we don't actually do `push`, we just store the arguments into the outgoing area right before the call, so it is not a problem if we "reorder" the arguments like above.

So I think we can just put ` && !defined(TARGET_WASM)` to the `#if !FEATURE_FIXED_OUT_ARGS` in `EvalArgsToTemps`.

yowl:
Thanks, that clears that problem.  I'm going back to something earlier:

> Only the underlying nodes need to be changed.

I'm having a spot of bother with this because, for arguments on the shadow stack, I'm attempting to remove them from `gtArgs` with `Remove`.  Firstly that doesn't like being called when `m_abiInformationDetermined == true`, but if I ignore that for now, it only removes the early arg, and when a temp has been created, the `CallArg` has both an early and late node by virtue of the code in `EvalArgsToTemps` :
```
  if (setupArg != nullptr)
  {
      noway_assert(arg.GetEarlyNode() == argx);
      arg.SetEarlyNode(setupArg);
  }

  arg.SetLateNode(defArg);
  *lateTail = &arg;
  lateTail  = &arg.LateNextRef();
```
Am I going to need to adjust `Remove` for us to remove from the late arg list as well, or am I on the wrong path?

SingleAccretion:
It seems like the right direction.

We'd need a different helper though, as we cannot really use `Remove` as-is because it would mean the algorithm is going to be quadratic. The helper would be something like `RemoveAfter(CallArg* arg)` (basically a wrapper over `SetNext`).

Removing late args on the fly seems a bit hard since they're unordered (the list could be: `arg3 -> arg1 -> arg2`, etc), but we can just null out the entire list by transferring late nodes to early ones, doing `SetLateNext(nullptr)` and nulling out `m_lateHead`.

yowl:
I was hoping I could get rid of the siginfo determination of call signature and rely on the args, but that's not going to be the case it seems (shorts->ints is one problem).  We will keep the siginfo code, that's as you expect?

SingleAccretion:
We need the signature for helpers since they aren't precisely tracked. For user/indirect calls, I'd expect we could use the fields added in https://github.com/dotnet/runtime/pull/68748 to track them, with a little change to use `CorInfoType` instead of `var_types` because we need to distinguish pointers from non-pointers.

yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/jit/compiler.h
```diff
@@ -8567,10 +7742,9 @@ class Compiler
     */
 
 public:
-    CodeGenInterface* codeGen;
-
 #ifndef TARGET_WASM
     // Record the instr offset mapping to the generated code
+    CodeGenInterface* codeGen;
 
     jitstd::list<IPmappingDsc> genIPmappings;
```
```suggestion
    CodeGenInterface* codeGen;

    // Record the instr offset mapping to the generated code
    jitstd::list<IPmappingDsc> genIPmappings;
```

src/coreclr/jit/gentree.cpp
```diff
@@ -1840,6 +1842,39 @@ void CallArgs::Remove(CallArg* arg)
     assert(!"Did not find arg to remove in CallArgs::Remove");
 }
 
+#if TARGET_WASM
+void CallArgs::RemoveAfter(CallArg* arg)
```
SingleAccretion, line 1857: This needs a [standard function header](https://github.com/dotnet/runtimelab/blob/24524863c43f97a42a55466b5de4cee3fd6d22fb/src/coreclr/jit/conventions.txt#L57-L81).

Also, `RemovedWellKnownArg` needs to be called for any removed args.

src/coreclr/jit/gentree.cpp
```diff
@@ -1840,6 +1842,39 @@ void CallArgs::Remove(CallArg* arg)
...
+    while (lateArg != nullptr)
+    {
+        lateArg->SetEarlyNode(lateArg->GetLateNode());
+        lateArg->SetLateNode(nullptr);
+        lateArg = lateArg->GetLateNext();
+    }
+    
+    m_lateHead = nullptr;
+}
+
```
SingleAccretion:
```suggestion
```

src/coreclr/jit/gentree.cpp
```diff
@@ -1840,6 +1842,39 @@ void CallArgs::Remove(CallArg* arg)
...
+        m_head = m_head->GetNext();
+        return;
+    }
+
+    assert(arg->GetNext() != nullptr);
+    
+    arg->SetNext(arg->GetNext()->GetNext());
+}
+
+void CallArgs::MoveLateToEarly()
```
SingleAccretion, line 1883: Ditto: comment header.

src/coreclr/jit/llvmlower.cpp
```diff
@@ -597,13 +597,10 @@ void Llvm::lowerRethrow(GenTreeCall* callNode)
...
     // explicitly. Add it here, before the general call lowering.
-    assert(callNode->gtCallArgs == nullptr);
-    callNode->ResetArgInfo();
+    assert(callNode->gtArgs.IsEmpty());
 
     GenTree* excObjAddr = insertShadowStackAddr(callNode, getCatchArgOffset(), _shadowStackLclNum);
-    callNode->gtCallArgs = _compiler->gtNewCallArgs(excObjAddr);
-
-    _compiler->fgInitArgInfo(callNode);
+    callNode->gtArgs.InsertAfterThisOrFirst(_compiler, excObjAddr);
```
SingleAccretion: Why not `PushFront`?

src/coreclr/jit/llvmlower.cpp
```diff
@@ -785,48 +782,15 @@ void Llvm::lowerReturn(GenTreeUnOp* retNode)
 //     in a simple increasing order, matching the signature. We also rewrite returns
 //     that must be on the shadow stack, see "lowerCallReturn".
 //
+// LVVM Arg layout:
+//    - Shadow stack (if required)
+//    - Return slot (if required)
+//    - Generic context (if required)
+//    - args passed as LLVM parameters (not on the shadow stack)
```
```suggestion
//    - Generic context (if required)
//    - Args passed as LLVM parameters (not on the shadow stack)
//
```

src/coreclr/jit/llvmlower.cpp
```diff
@@ -775,7 +772,7 @@ void Llvm::lowerReturn(GenTreeUnOp* retNode)
 // lowerCallToShadowStack: Lower the call, rewriting its arguments.
 //
 // This method has two primary objectives:
-//  1) Transfer the information about the arguments from arg info to explicit
+//  1) Transfer the information about the arguments from gtArgs to explicit
 //     PutArgType nodes, to make it easy for codegen to consume it. Also, get
 //     rid of the late/non-late argument distinction, by sorting the inserted nodes
 //     in the original evaluation order, matching that of them in the signature.
```
```suggestion
//     PutArgType nodes, to make it easy for codegen to consume it. Also, all
//     of the late argument nodes are moved (back) to the early list.
```

src/coreclr/jit/llvmlower.cpp
```diff
@@ -997,36 +974,31 @@ void Llvm::failUnsupportedCalls(GenTreeCall* callNode)
         failFunctionCompilation();
     }
 
-    if (callNode->gtCallArgs != nullptr)
+    for (CallArg& callArg : callNode->gtArgs.Args())
     {
-        for (GenTree* operand : callNode->Operands())
+        GenTree* operand = callArg.GetNode();
```
SingleAccretion: I think the whole:
```
        GenTree* operand = callArg.GetNode();
        if (operand->IsArgPlaceHolderNode() || !operand->IsValue())
        {
            // Either of these situations may happen with calls.
            continue;
        }
        if (operand == callNode->gtControlExpr || operand == callNode->gtCallAddr)
        {
            // vtable target or indirect target
            continue;
        }
```
Can just be deleted as it's a no-op. We only care about blocking VSD.

src/coreclr/jit/morph.cpp
```diff
@@ -3141,6 +3095,12 @@ void CallArgs::AddFinalArgsAndDetermineABIInfo(Compiler* comp, GenTreeCall* call
     m_abiInformationDetermined = true;
 }
 
+
+// normally in target<arch>.h but llvm/wasm does not have that file and seems overkill for this one constant
```
SingleAccretion: Hmm, but we do have `targetwasm.h`?

src/coreclr/jit/llvmlower.cpp
```diff
@@ -929,40 +895,51 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
...
-            sigArgs = _info.compCompHnd->getArgNext(sigArgs);
+            lastLlvmStackArg = &callArg;
         }
 
         argIx++;
     }
+
+    callNode->gtArgs.MoveLateToEarly();
+
+    // Insert special args after finished iterating >gtArgs.Args
```
SingleAccretion: To be honest, I liked the setup where we were inserting these in-order at the top more. But it's not a big deal.

src/coreclr/jit/llvmlower.cpp
```diff
@@ -929,40 +895,51 @@ void Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
...
-        if (isSigArg && (sigInfo != nullptr))
-        {
-            sigArgs = _info.compCompHnd->getArgNext(sigArgs);
+            lastLlvmStackArg = &callArg;
         }
 
         argIx++;
     }
+
+    callNode->gtArgs.MoveLateToEarly();
```
SingleAccretion: Can we do this before at the top, before rewriting the argument list? It'd be a bit clearer I think.

src/coreclr/jit/morph.cpp
yowl: /shuffles feet/

src/coreclr/jit/llvmlower.cpp
yowl: I just assumed it was going to be a problem with iterator.  I could do the iteration manually, starting from the last of shadow stack or return slot.  Do you want to see what that looks like?

SingleAccretion: >  Do you want to see what that looks like?

It would be appreciated.

I don't think manual `GetNext`-style iteration is a big downside, since we already effectively depend on the argument list being linked with `RemoveAfter`.

src/coreclr/jit/llvmlower.cpp
yowl: Yes, done

src/coreclr/jit/morph.cpp
yowl: moved to `targewasm.h`

src/coreclr/jit/llvmlower.cpp
yowl: Indeed.

src/coreclr/jit/gentree.cpp
yowl: Thanks, added both

src/coreclr/jit/gentree.cpp
yowl: Do we need something in these comments to say it applies to LLVM lowering?

SingleAccretion: No need I think. The `#ifdef` is self-explanatory and the functions are general-purpose.

src/coreclr/jit/gentree.cpp
```diff
@@ -1840,6 +1842,57 @@ void CallArgs::Remove(CallArg* arg)
...
+//          if the first argument should be removed.
+//
+void CallArgs::RemoveAfter(CallArg* arg)
+{
+    // if arg == nullptr, we are removing the first arg
+    if(arg == nullptr)
+    {
+        assert(m_head != nullptr);
+
+        m_head = m_head->GetNext();
```
SingleAccretion, line 1864: `RemovedWellKnownArg` is needed here too.

src/coreclr/jit/llvmlower.cpp
yowl: Last commit has this.  There's some long compound ways to get the head of the list `callNode->gtArgs.Args().begin().GetArg()`.  I could add a `GetHead` method to make that look nicer?

SingleAccretion (approved):
LGTM, thank you!

src/coreclr/jit/llvmlower.cpp
SingleAccretion: Not worth it at present I think, the use in return lowering could be replaced with `GetArgByIndex(0)`, and then only only one is left (modulo upstream)

yowl: >  I think, the use in return lowering could be replaced with GetArgByIndex(0)

I added a commit for this.  If it goes green I will ping Jan.  Thanks for the review!

yowl:
@jkotas This looks good to go, thanks!

jkotas:
Great work! Thank you!

PR #2178, opened 2023-01-21
[NativeAOT-LLVM] Debug info refactoring

SingleAccretion:
Two changes:
1) Initialize the DI structures eagerly instead of lazily.
2) Insert a zero-offset ILOffset node so that we don't have problems with missing DI locations for calls.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -8,11 +8,10 @@
 #include "llvm/Bitcode/BitcodeWriter.h"
 #pragma warning (error: 4459)
 
-LLVMContext      _llvmContext;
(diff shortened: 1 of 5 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, did you try Chrome/Edge or VS (remote debugging)?  I don't see anything that would have made it stop working, just wondered?

SingleAccretion:
> did you try Chrome/Edge or VS (remote debugging)?

I have not; will do.

SingleAccretion:
> I have not; will do.

Done. Debugging in Chrome works, breakpoints are hit and expected source lines shown / stepped through.

Issue #2095, opened 2022-11-18
[NativeAOT-LLVM] Remove the workaround for `CalendarData.EnumCalendarInfo`


SingleAccretion:
The functions need real implementations. Opening an issue for tracking purposes.

Code references:

https://github.com/dotnet/runtimelab/blob/8284a6d8d46866f25f1ed605a9c4b6ac579a844c/src/coreclr/tools/Common/TypeSystem/IL/ILImporter.cs#L953-L969

https://github.com/dotnet/runtimelab/blob/8284a6d8d46866f25f1ed605a9c4b6ac579a844c/src/coreclr/jit/llvmcodegen.cpp#L23-L32

SingleAccretion:
I am working on #2165 and the last warning to be dealt with is the missing symbol for `dotnet_browser_entropy`, which is referenced by System.Native now. I see two possible solutions:

1\) Properly plumb the JS through our build and use it.
2\) Add another workaround (either in compiler or System.Native) to silence the warning.

I like doing `1` now more because:
a\) It means we get properly working random data.
b\) The concern that we'll tie ourselves to JS-providing WASM hosts seems to no longer be as relevant, as [upstream is adding support for WASI builds directly into System.Native](https://github.com/dotnet/runtime/pull/79046), so, presumably, we'll be able to pick that up as well.

@yowl what do you think?

yowl:
Yes, makes sense, i.e. point 1

PR #2179, opened 2023-01-22
[NativeAOT-LLVM] Integrate the JS implementation for `dotnet_browser_entropy`

### LINKED ISSUE #2165: [NativeAOT-LLVM] Runtime tests generate a significant number of warnings

SingleAccretion:
Overall a pretty simple change:

1) Consider `pal_random.lib.js` an artifact of the `System.Native` build and pass it as `--js-library` to `emcc`.
2) Add a polyfill, since our Node doesn't support `crypto` directly (taken more or less verbatim from https://github.com/dotnet/runtime/pull/78696).

As a result, we now have a fully functional implementation and zero warnings in the runtime test build.

Fixes #2165.
Contributes to #2095.


SingleAccretion:
@dotnet/nativeaot-llvm

eng/liveBuilds.targets
```diff
@@ -176,6 +176,10 @@
         $(LibrariesNativeArtifactsPath)*.pdb"
         IsNative="true"
         Exclude="@(ExcludeNativeLibrariesRuntimeFiles)" />
+      <LibrariesRuntimeFiles Condition="'$(TargetOS)' == 'Browser' and '$(RuntimeFlavor)' == 'CoreCLR'"
+                             Include="
+        $(LibrariesNativeArtifactsPath)pal_random.lib.js;"
```
yowl, line 181: nit; this wrapping looks odd, is that intentional?

yowl (approved):
Nice, just the nit on the wrapping of the line.

eng/liveBuilds.targets
SingleAccretion: It is meant to be consistent with the rest of the file:
```
      <LibrariesRuntimeFiles Condition="'$(TargetOS)' == 'Browser' and '$(RuntimeFlavor)' == 'CoreCLR'"
                             Include="
        $(LibrariesNativeArtifactsPath)pal_random.lib.js;"
        IsNative="true" />
      <LibrariesRuntimeFiles Condition="'$(TargetOS)' == 'Browser' and '$(RuntimeFlavor)' == 'Mono'"
                             Include="
        $(LibrariesNativeArtifactsPath)dotnet.js;
        $(LibrariesNativeArtifactsPath)dotnet.d.ts;
        $(LibrariesNativeArtifactsPath)package.json;
        $(LibrariesNativeArtifactsPath)dotnet.wasm;
        $(LibrariesNativeArtifactsPath)dotnet.js.symbols;
        $(LibrariesNativeArtifactsPath)dotnet.timezones.blat;
        $(LibrariesNativeArtifactsPath)*.dat;"
        IsNative="true" />
```

yowl: Good to go then, thanks

jkotas:
Thank you!

Issue #2151, opened 2023-01-05
NativeAOT-LLVM: LLVM structs over single pointers, should be ptr


yowl:
Following from https://github.com/dotnet/runtimelab/pull/2139#discussion_r1062703990

If we use reduce a structs over a pointer to an intrinsic, we should use `ptr`  not `i32` so we can remove a cast when storing to the field.

PR #1791, opened 2021-12-21
NativeAOT-LLVM: Support nativelib for Wasm 

yowl:
This PR removes the restriction of having a main enabling the `--nativelib` option.
Fixes a bug with a nativelib where the shadow stack would grow for every pinvoke call from the nativelib.
Fixes a bug processing options to enable `--codegenopt:Target=wasm32-unknown-unknown`

cc @SingleAccretion.

PR #2180, opened 2023-01-27
[NativeAOT-LLVM] Implement delegate constructors and managed helpers

SingleAccretion:
Two core changes:

1) Implement lowering support for helpers that take arguments on the shadow stack; implement the distinction of "managed" (shadow) vs "unmanaged" calling conventions. In the former, the caller uses the shadow stack to pass GC values, in the latter they are passed directly as pointer values. FCalls as well as runtime imports use the "unmanaged" calling convention, and must therefore ensure they report any live references properly, or forbid the GC from occurring.
   - This also lays the groundwork for supporting true native calls, i. e. PInvokes.
2) Implement `CORINFO_HELP_READYTORUN_DELEGATE_CTOR`:
   - Object writer code emission support
   - ILImporter rewrite to use it in the same manner RyuJit does (modulo `.constrained; ldftn` handling).

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -27,9 +27,6 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 extern ICorJitHost* g_jitHost;
 
 #ifdef TARGET_WASM
-#undef min
-#undef max
-
 #include "llvm.h"
 
 #define max(a, b) (((a) > (b)) ? (a) : (b))
@@ -1930,7 +1927,9 @@ void Compiler::compInit(ArenaAllocator*       pAlloc,
     {
 #ifndef TARGET_WASM
         codeGen = getCodeGenerator(this);
-#endif // !TARGET_WASM
+#else
+        m_llvm = new (getAllocator(CMK_Codegen)) Llvm(this);
+#endif
         optInit();
         hashBv::Init(this);
 
@@ -4384,13 +4383,6 @@ void Compiler::EndPhase(Phases phase)
     mostRecentlyActivePhase = phase;
 }
 
-#if defined(TARGET_WASM)
-inline void DoLlvmPhase(Llvm* llvm)
-{
-    llvm->Compile();
-}
-#endif
-
 //------------------------------------------------------------------------
 // compCompile: run phases needed for compilation
 //
@@ -5130,9 +5122,9 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
         lvaMarkLocalVars();
     }
 
-    Llvm* llvm = new Llvm(this);
-    auto  lowerPhase = [llvm]() {
-        llvm->Lower();
+    assert(m_llvm != nullptr);
+    auto lowerPhase = [this]() {
+        m_llvm->Lower();
     };
     DoPhase(this, PHASE_LOWER_LLVM, lowerPhase);
 
@@ -5141,11 +5133,10 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
     fgResetForSsa();
     DoPhase(this, PHASE_BUILD_SSA, &Compiler::fgSsaBuild);
 
-    auto buildLlvmPhase = [llvm]() {
-        llvm->Compile();
+    auto buildLlvmPhase = [this]() {
+        m_llvm->Compile();
     };
     DoPhase(this, PHASE_BUILD_LLVM, buildLlvmPhase);
-    delete llvm;
 #else
 
 #ifdef TARGET_ARM
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -67,6 +67,12 @@ CorInfoType HelperFuncInfo::GetSigReturnType() const
     return static_cast<CorInfoType>(SigReturnType);
 }
 
+CORINFO_CLASS_HANDLE HelperFuncInfo::GetSigReturnClass(Compiler* compiler) const
+{
+    assert(GetSigReturnType() != CORINFO_TYPE_VALUECLASS);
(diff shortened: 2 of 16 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -245,12 +316,13 @@ bool Llvm::callHasShadowStackArg(GenTreeCall* call)
...
 //    Reference to the info structure for "helperFunc".
 //
-const HelperFuncInfo& Llvm::getHelperFuncInfo(CorInfoHelpFunc helperFunc)
+/* static */ const HelperFuncInfo& Llvm::getHelperFuncInfo(CorInfoHelpFunc helperFunc)
 {
     // Note on Runtime[Type|Method|Field]Handle: it should faithfully be represented as CORINFO_TYPE_VALUECLASS.
     // However, that is currently both not necessary due to the unwrapping performed for LLVM types and not what
     // the Jit expects. When deleting the unwrapping, fix the runtime signatures to take the underlying pointer instead.
-    const int CORINFO_TYPE_RT_HANDLE = CORINFO_TYPE_PTR;
+    const int CORINFO_TYPE_RT_HANDLE = CORINFO_TYPE_NATIVEINT;
```
SingleAccretion, line 324: Due to https://github.com/dotnet/runtimelab/issues/2151, runtime handles show up in signatures as `int32`s.

src/coreclr/jit/compiler.cpp
```diff
@@ -27,9 +27,6 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 extern ICorJitHost* g_jitHost;
 
 #ifdef TARGET_WASM
-#undef min
```
yowl, line 30: This is interesting, were they causing a problem?  They were part of the problem with the linux build if I remember correctly.

SingleAccretion: Not really, but `llvm.h` already does this for LLVM headers to compile and there is no need to spread the workaround elsewhere.

(It would be ideal to redefine them after the `#include`s in `llvm.h` itself, so we don't need to duplicate that, but that's a task for another day)

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvm.cpp
```diff
@@ -245,12 +316,13 @@ bool Llvm::callHasShadowStackArg(GenTreeCall* call)
 // Return Value:
 //    Reference to the info structure for "helperFunc".
 //
-const HelperFuncInfo& Llvm::getHelperFuncInfo(CorInfoHelpFunc helperFunc)
+/* static */ const HelperFuncInfo& Llvm::getHelperFuncInfo(CorInfoHelpFunc helperFunc)
```
yowl, line 319: nit: is this comment intentional?

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1933,6 +1937,13 @@ void Llvm::buildCall(GenTreeCall* call)
     }
 
     Value* callValue = emitCallOrInvoke(llvmFuncCallee, argVec);
+
+    // Restore the saved shadow stack.
+    if (shadowStackToRestoreValue != nullptr)
+    {
+        _builder.CreateStore(shadowStackToRestoreValue, shadowStackTopAddrValue);
+    }
```
yowl, line 1945: I'm sure there's a good reason why we need to do this, can you remind me what it is please?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/InterfaceDispatchCellNode.cs
```diff
@@ -93,23 +93,35 @@ public override void EncodeData(ref ObjectDataBuilder objData, NodeFactory facto
             }
 
             IEETypeNode interfaceType = factory.NecessaryTypeSymbol(_targetMethod.OwningType);
-            if (interfaceType.RepresentsIndirectionCell)
+            if (factory.Target.SupportsRelativePointers)
             {
-                objData.EmitReloc(interfaceType, RelocType.IMAGE_REL_BASED_RELPTR32,
-                    (int)InterfaceDispatchCellCachePointerFlags.CachePointerIsIndirectedInterfaceRelativePointer);
+                if (interfaceType.RepresentsIndirectionCell)
```
yowl, line 98: @MichalStrehovsky  Can you have a quick review of this please?  Thanks.

yowl (approved):
Looks great, just a request for Michal with his superior knowledge to review one bit please.  

Lots of code deleted - always a good sign :-)

src/coreclr/jit/llvmcodegen.cpp
SingleAccretion: This was carried over from the IL importer version of the code, where it was added in https://github.com/dotnet/runtimelab/pull/1791.

If my inference is correct, the reason shadow stack top needs to be restored is so that we don't end up doing this:
```
[Native caller] ; i = 0; ManagedCallee();
[ManagedCallee] ; SS = new (...)]
[Stack grows by X]
[NativeCall] ; SSTop = SS + X

  [Native caller] ; i = 1; ManagedCallee();
  [ManagedCallee] ; SS = SS + X]
  [Stack grows by X]
  [NativeCall] ; SSTop = SS + 2X

    [Native caller] ; i = 2; ManagedCallee();
    [ManagedCallee] ; SS = SS + 2X]
    [Stack grows by X]
    [NativeCall] ; SSTop = SS + 3X

; And so on until SSTop grows out of bounds
```
Presumably, we could restore it in the RPI frame instead though, which would be better since RPI is much less common than PI (once we don't have it everywhere with runtime imports/exports, that is).

src/coreclr/jit/llvm.cpp
SingleAccretion: Yep, `getHelperFuncInfo` went from instance to static.

MichalStrehovsky (approved):
InterfaceDispatchCellNode.cs change looks good!

PR #2181, opened 2023-01-28
NativeAot-LLVM: Merge to apr 20

yowl:
This PR moves a little forward as there seems to be quite a few conflicts coming up.


yowl:
Had a problem with `liveBuild.targets` and `ResolveRuntimeFilesFromLocalBuild` because `RuntimeFiles` was empty when building the tests.  Previously it looks like `RuntimeFiles` only had a `pdb` file in it:

image

So didn't look like it was necessary for building the tests to pass anyway.  Maybe we should be looking into why it is a dependency?

src/coreclr/jit/gentree.h
```diff
@@ -4022,12 +4022,8 @@ const char* getWellKnownArgName(WellKnownArg arg);
 struct CallArgABIInformation
 {
     CallArgABIInformation()
-<<<<<<< HEAD
         : ArgNum((unsigned)-1)
         , NumRegs(0)
```
SingleAccretion: This does not look correct. Upstream deleted `ArgNum` and so should we. If there are dependencies on it in `TARGET_WASM` code they should be deleted.

src/coreclr/jit/gentree.h
```diff
@@ -4059,13 +4055,10 @@ struct CallArgABIInformation
         }
     }
 
-<<<<<<< HEAD
     // The original argument number, also specifies the required argument
     // evaluation order from the IL
     unsigned ArgNum;
```
SingleAccretion: Ditto: `ArgNum` deleted.

src/coreclr/jit/lower.cpp
```diff
@@ -2414,10 +2414,7 @@ void Lowering::LowerCFGCall(GenTreeCall* call)
             call->gtArgs.PushLateBack(targetArg);
 
             // Set up ABI information for this arg.
-<<<<<<< HEAD
             targetArg->AbiInfo.ArgNum  = call->gtArgs.CountArgs() - 1;
```
SingleAccretion: Ditto: `ArgNum`.

src/coreclr/jit/morph.cpp
```diff
@@ -650,12 +650,8 @@ const char* getWellKnownArgName(WellKnownArg arg)
 //
 void CallArg::Dump(Compiler* comp)
 {
-<<<<<<< HEAD
     printf("CallArg[arg %u", AbiInfo.ArgNum);
     printf(" [%06u].%s", comp->dspTreeID(GetNode()), GenTree::OpName(GetEarlyNode()->OperGet()));
-=======
-    printf("CallArg[[%06u].%s", comp->dspTreeID(GetNode()), GenTree::OpName(GetEarlyNode()->OperGet()));
->>>>>>> b39c723a9b6c07f85304bf41a97bcc18225b1206
```
SingleAccretion: Ditto: `ArgNum`.

src/coreclr/jit/morph.cpp
```diff
@@ -2845,10 +2829,7 @@ void CallArgs::AddFinalArgsAndDetermineABIInfo(Compiler* comp, GenTreeCall* call
 #endif // TARGET_ARM
 
         arg.AbiInfo          = {};
-<<<<<<< HEAD
         arg.AbiInfo.ArgNum   = argIndex;
```
SingleAccretion: Ditto: `ArgNum`.

SingleAccretion (commented):
The `RuntimeFiles` logic look CoreCLR-specific, so I am wondering if a `!NativeAot` condition is more appropriate. Also wondering whether upstream NAOT testing doesn't have the same problem by design of by accident (will look into it).

src/coreclr/jit/morph.cpp
```diff
@@ -650,7 +650,6 @@ const char* getWellKnownArgName(WellKnownArg arg)
 //
 void CallArg::Dump(Compiler* comp)
 {
-    printf("CallArg[arg %u", AbiInfo.ArgNum);
     printf(" [%06u].%s", comp->dspTreeID(GetNode()), GenTree::OpName(GetEarlyNode()->OperGet()));
```
SingleAccretion: Should be `printf("CallArg[[%06u].%s", comp->dspTreeID(GetNode()), GenTree::OpName(GetEarlyNode()->OperGet()));` to match upstream.

yowl:
> The `RuntimeFiles` logic look CoreCLR-specific, so I am wondering if a `!NativeAot` condition is more appropriate. 

I added a condition to the target to exclude it for nativeaot test builds.

src/coreclr/jit/gentree.h
yowl: Thanks, removed `ArgNum` now.

SingleAccretion (approved):
LGTM, modulo one comment! Sorry for the delays, working through understanding how the build fits together is always a bit of a slow process.

src/coreclr/jit/morph.cpp
yowl: Thanks, now producing:
```
Args for call [000006] CALL after AddFinalArgsAndDetermineABIInfo:
CallArg[[000005].RUNTIMELOOKUP
```

yowl:
Thanks, @jkotas this is good to go now.  The `Dump` is now aligned with upstream.

jkotas:
Thank you!

PR #2182, opened 2023-01-28
Expose MSbuild property for setting shell file

kant2002:
Current shell file shipped by Emscripten is ugly, and you definitely want provide something alternative, even for demo purposes.

### DIFF
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -110,7 +110,7 @@ This can be used to import WASI functions that are in other modules, either as t
 # WASM applications configuration
 
 Currently NativeAOT-LLVM supports following additional properties
-- `WasmHtmlTemplate` which specified path to the HTML template within which WASM application would be embeded.
+- `WasmHtmlTemplate`: specifies path to the HTML template within which the WASM application will be embedded. Example of minimal template from Emscipten repo: https://github.com/emscripten-core/emscripten/blob/main/src/shell_minimal.html
 
 ### Cross-compiling on Linux
 Similarly, to target linux-arm64 on a Linux x64 host, in addition to the `Microsoft.DotNet.ILCompiler` package reference, also add the `runtime.linux-x64.Microsoft.DotNet.ILCompiler` package reference to get the x64-hosted compiler:
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -107,6 +107,11 @@ Will cause the above `random_get` to create this WebAssembly:
 
 This can be used to import WASI functions that are in other modules, either as the above, in WASI, `wasi_snapshot_preview1`, or in other WebAssembly modules that may be linked with [WebAssembly module linking](https://github.com/WebAssembly/module-linking)
 
+# WASM applications configuration
+
+Currently NativeAOT-LLVM supports following additional properties
+- `WasmHtmlTemplate` which specified path to the HTML template within which WASM application would be embeded.
+
 ### Cross-compiling on Linux
 Similarly, to target linux-arm64 on a Linux x64 host, in addition to the `Microsoft.DotNet.ILCompiler` package reference, also add the `runtime.linux-x64.Microsoft.DotNet.ILCompiler` package reference to get the x64-hosted compiler:
 ```xml
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -110,7 +110,7 @@ This can be used to import WASI functions that are in other modules, either as t
 # WASM applications configuration
 
 Currently NativeAOT-LLVM supports following additional properties
-- `WasmHtmlTemplate`: specifies path to the HTML template within which the WASM application will be embedded. Example of minimal template from Emscipten repo: https://github.com/emscripten-core/emscripten/blob/main/src/shell_minimal.html
+- `WasmHtmlTemplate`: specifies path to the HTML template within which the WASM application will be embedded. An example of a minimal template can be found in the Emscripten repo: https://github.com/emscripten-core/emscripten/blob/main/src/shell_minimal.html
 
 ### Cross-compiling on Linux
 Similarly, to target linux-arm64 on a Linux x64 host, in addition to the `Microsoft.DotNet.ILCompiler` package reference, also add the `runtime.linux-x64.Microsoft.DotNet.ILCompiler` package reference to get the x64-hosted compiler:
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -349,7 +349,6 @@ The .NET Foundation licenses this file to you under the MIT license.
       <EmccArgs>&quot;$(LlvmClrjitObject)&quot; -c -o &quot;$(NativeClrjitObject)&quot;  -s DISABLE_EXCEPTION_CATCHING=0</EmccArgs>
       <EmccArgs Condition="'$(NativeDebugSymbols)' != 'true'">$(EmccArgs) -O2</EmccArgs>
       <EmccArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(EmccArgs) -g3</EmccArgs>
-      <EmccArgs Condition="'$(WasmHtmlTemplate)' != ''">$(EmccArgs) --shell-file $(WasmHtmlTemplate)</EmccArgs>    
       <!-- TODO-LLVM: remove when we update to Emscripten with https://github.com/emscripten-core/emscripten/pull/18443 -->
       <EmccArgs>$(EmccArgs) -s USE_SDL=0</EmccArgs>
     </PropertyGroup>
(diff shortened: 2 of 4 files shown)


yowl:
I'm not sure about this because it is already possible with something like:

https://github.com/yowl/CSharpWebGpu/blob/6f91408b1e944b4ace3d58dbacefe6d7f588e219/Particles/Particles.csproj#L10

and the `--shell-file=$(MSBuildThisFileDirectory)shell_minimal.html` argument.

Do we want to be adding msbuild properties for what is effectively just another linker argument?  Where would we stop?  Perhaps we should be renaming `EmccExtraArgs` to  `CustomLinkerArg` to better follow NativeAOT for `link` and `ld`.  I believe mono has `EmccExtraCompileArgs` and `EmccExtraLinkArgs` (or something similar) but as NativeAOT has `CustomLinkerArg` we should probably follow that.

kant2002:
I want to have something which can be documented. I know that developer can use `EmccExtraArgs`, but isn't it easier to have something well known as parameter?

SingleAccretion:
I think in the current state of things, people will need to use `EmccExtraArgs` if they want to do something interesting regardless.

We could certainly add a documentation section somewhere for "useful Emscripten options".

kant2002:
Honestly the less I know about Emscripten the better for me. For me this kind of changes is to smooth learning curve and make life of outsiders easier. I think having separate properties allow to make these command lines more composable in a sense that you can build different combinations of command lines using MSbuild facilities easier.

I think most people would like to have something more interesting as shell because current default for EMCC is not very pleasant and for web devs it look *strange*. 

> Where do we stop?

I think we should provide sensible defaults where you don't need specify anything and options should be opt-in. In case of Hml template, I think we can ship separate better looking HTML template and allow override it. 

for example, I think we can have two templates or some configurations, Not everybody need canvas, maybe for some samples console would be enough (?) Just thinking aloud. With linker switch it's on developer to provide correct template. If do not provided, NativeAOT app would look worse then it can be.

SingleAccretion:
I guess the question here boils down to whether we want to make `EmccExtraArgs` a niche property and provide MSBuild wrappers for mainstream scenarios or not.

One angle to look at this: what if this was `dotnet/runtime`? Would it be acceptable to have the mainstream experience be using `EmccExtraArgs`? Probably not. But then we need to look for the end-to-end as well, i. e. what set of Emscripten options to expose.

yowl:
Right, if this was mainstream would you be doing "dotnet new wasm node/console", "dotnet new wasm canvas" "dotnew new wasm minimal".  I'm fine with adding the mainstream wrappers as msbuild properties as it does make thing that bit easier as long as they are documented somewhere.

kant2002:
My idea for now, because I just start to play with WASM following. I do not want to know anything about Emscripten which is not that good documented IMO and I do want to have reasonably nice defaults.

Currently we have UI mode where @yowl do his magic.
and we also have console mode where I plan to play a bit to create more or less useful samples.

I really want 
1. replace ugly Emscripten shell with something more nice.
image
2. Make sure that full screen working by default. it's not without some tweaking.
3. Preferably if something is not available, then hide that from shell. For example if I plan to work with console I prefer have default shell like that.
image

I really dislike idea that developer should know all EMCC parameters. For example, finding what to do to make full screen mode works was non-trivial. I assume that if you use some tech it will start magically working, but otherwise we have broken sample by default which give bad impression.
That's repo where I want to show some ideas. https://github.com/kant2002/NativeAOTWasm ,

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -337,6 +337,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <EmccArgs>&quot;$(LlvmObject)&quot; -c -o &quot;$(NativeObject)&quot; -s DISABLE_EXCEPTION_CATCHING=0</EmccArgs>
       <EmccArgs Condition="'$(NativeDebugSymbols)' != 'true'">$(EmccArgs) -O2</EmccArgs>
       <EmccArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(EmccArgs) -g3</EmccArgs>
+      <EmccArgs Condition="'$(EmccShellFile)' != ''">$(EmccArgs) --shell-file $(EmccShellFile)</EmccArgs>
```
SingleAccretion: This is the compile-time args section, `--shell-file` is a link-time flag, so it would have to be added [below](https://github.com/dotnet/runtimelab/blob/50151a1fd5cbf06e024338895ce9e18b8278058e/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L418-L422).

Additionally, if we don't want too much Emscripten specifics to leak out, `EmccShellFile` -> `HtmlShellFile`?

src/coreclr/nativeaot/docs/wasm.md
```diff
@@ -0,0 +1,4 @@
+# WASM applications configuration with Native AOT
+
+Currently NativeAOT-LLVM supports following additional properties
+- `WasmHtmlTemplate` which specified path to the HTML template within which WASM application would be embeded.
```
SingleAccretion: Instead of a new file, I think it would be better to use [the existing section](https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/docs/using-nativeaot/compiling.md#webassembly).

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -337,6 +337,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <EmccArgs>&quot;$(LlvmObject)&quot; -c -o &quot;$(NativeObject)&quot; -s DISABLE_EXCEPTION_CATCHING=0</EmccArgs>
       <EmccArgs Condition="'$(NativeDebugSymbols)' != 'true'">$(EmccArgs) -O2</EmccArgs>
       <EmccArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(EmccArgs) -g3</EmccArgs>
+      <EmccArgs Condition="'$(WasmHtmlTemplate)' != ''">$(EmccArgs) --shell-file $(WasmHtmlTemplate)</EmccArgs>
```
SingleAccretion: `--shell-file` is a link-time option, which need to be specified here:
https://github.com/dotnet/runtimelab/blob/3256ce868921db2e1bf09d0fb5aa2a86f0bc2a5e/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L416-L422

src/coreclr/nativeaot/docs/wasm.md
SingleAccretion:
```suggestion
- `WasmHtmlTemplate`: specifies path to the HTML template within which the WASM application will be embedded.
```
We also need to document its format (by linking to Emscripten documentation, probably)

SingleAccretion (commented):
I think we have a broad enough agreement now that this is something we'd like to have in some capacity, so let's get it in. It would be awesome to have something prettier for the default shell, I think \:).

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -349,6 +349,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <EmccArgs>&quot;$(LlvmClrjitObject)&quot; -c -o &quot;$(NativeClrjitObject)&quot;  -s DISABLE_EXCEPTION_CATCHING=0</EmccArgs>
       <EmccArgs Condition="'$(NativeDebugSymbols)' != 'true'">$(EmccArgs) -O2</EmccArgs>
       <EmccArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(EmccArgs) -g3</EmccArgs>
+      <EmccArgs Condition="'$(WasmHtmlTemplate)' != ''">$(EmccArgs) --shell-file $(WasmHtmlTemplate)</EmccArgs>
```
SingleAccretion: ...this is still the compile-time section.

Also, we need to `&quot;` the path.

docs/using-nativeaot/compiling.md
```diff
@@ -107,6 +107,11 @@ Will cause the above `random_get` to create this WebAssembly:
 
 This can be used to import WASI functions that are in other modules, either as the above, in WASI, `wasi_snapshot_preview1`, or in other WebAssembly modules that may be linked with [WebAssembly module linking](https://github.com/WebAssembly/module-linking)
 
+# WASM applications configuration
+
+Currently NativeAOT-LLVM supports following additional properties
+- `WasmHtmlTemplate`: specifies path to the HTML template within which the WASM application will be embedded. Example of minimal template from Emscipten repo: https://github.com/emscripten-core/emscripten/blob/main/src/shell_minimal.html
```
yowl:
```suggestion
- `WasmHtmlTemplate`: specifies path to the HTML template within which the WASM application will be embedded. An example of a minimal template can be found in the Emscripten repo: https://github.com/emscripten-core/emscripten/blob/main/src/shell_minimal.html
```

yowl (commented):
I think this is a good start, just the minor grammar change.  Thanks!

jkotas (approved):
Thanks

Issue #2000, opened 2022-09-28
NativeAOT-LLVM: optimisation for managed helpers


yowl:
The RyuJit backend should use the managed helpers where possible to avoid an extra call.  Follows from https://github.com/dotnet/runtimelab/issues/1999

SingleAccretion:
Adding this optimization will result in a size regression because of (LLVM-driven) inlining and because setting up shadow arguments is more expensive than LLVM ones:
```
Summary of Code Size diffs:
(Lower is better)

Total bytes of base: 3303065
Total bytes of diff: 3308866
Total bytes of delta: 5801 (0.18% % of base)
Average relative delta: 0.53%
    diff is a regression
    average relative diff is a regression

Top method regressions (percentages):
        1527 (152.40% of base) : 1939.dasm - S_P_CoreLib_System_Reflection_Runtime_General_NativeFormatMetadataReaderExtensions__TryParseConstantEnumArray
         119 (108.18% of base) : 2070.dasm - S_P_CoreLib_Internal_TypeSystem_LockFreeReaderHashtable_2<S_P_TypeLoader_Internal_TypeSystem_TypeSystemContext_MethodForInstantiatedTypeKey__System___Canon>__TryWriteSentinelToLocation
         113 (62.43% of base) : 2071.dasm - S_P_CoreLib_Internal_TypeSystem_LockFreeReaderHashtable_2<S_P_TypeLoader_Internal_TypeSystem_TypeSystemContext_MethodForInstantiatedTypeKey__System___Canon>__VolatileReadNonSentinelFromHashtable
          48 (61.54% of base) : 2004.dasm - HelloWasm_Program__TestObjectRefInUncoveredShadowStackSlot
         275 (52.68% of base) : 1833.dasm - S_P_CoreLib_System_Reflection_Runtime_General_NativeFormatMetadataReaderExtensions__ToArray_2
         133 (51.15% of base) : 2072.dasm - S_P_CoreLib_Internal_TypeSystem_LockFreeReaderHashtable_2<S_P_TypeLoader_Internal_TypeSystem_TypeSystemContext_MethodForInstantiatedTypeKey__System___Canon>__WaitForSentinelInHashtableToDisappear
         231 (37.50% of base) : 1461.dasm - S_P_TypeLoader_Internal_Runtime_TypeLoader_TypeLoaderEnvironment__GetTypeSequence
[...]

PR #2184, opened 2023-01-30
[NativeAOT-LLVM] Implement cached interface dispatch for WASM

SingleAccretion:
This change implements the cached interface dispatch feature for the RyuJit backend of our WASM target.

First, a little overview of regular CID I will steal from myself:
```
The compiler generates a "VSD cell": struct { m_pTarget, m_Cache } and gives its address to codegen (this is the
WellKnownArg::VirtualStubCell). Codegen will dereference it and call m_pTarget, passing the cell's address in a
special register.

This then goes into one of the assembly stubs which handle dispatch. The stubs are specialized for the cache sizes,
e. g. there is a stub for no cache (aka the initial stub, one compiler initialized the cell with), one for 1 entry,
2 entries, 4 entries ... 64 entries.

The stubs do this:

MethodTable* pObjectType = pObject->get_EEType();
InterfaceDispatchCache* pCache = reinterpret_cast<InterfaceDispatchCache*>(pCell->m_pCache);
for (size_t i = 0; i < CacheEntryCount; i++)
{
    InterfaceDispatchCacheEntry* pEntry = &pCache->m_rgEntries[i];
    if (pEntry->m_pInstanceType == pObjectType)
    {
        jmp pEntry->m_pTargetCode;
    }
}

jmp UniversalTranstion to RhpCidResolve(pObject, pCell);


UniversalTransition is a special assembly helper that saves argument registers, etc, and then calls the target it is
given (RhpCidResolve), which gives it the ultimate resolved managed target, to which it jmps (restoring the state it
has just saved).

RhpCidResolve then calls into the managed runtime and does two things:
1) Finds the target with some very complicated logic you'd expect.
2) Updates the cell's cache to include this new target.

Finally, the cache update mechanism has logic in it for recycling caches which were thrown away in the update process,
by registering them in a list which is emptied at GC time (when no managed code can run hence all the entries are
definitely no longer in use), with the cache entries transferred to a free list, from which the cache allocator can
then draw new ones. 
```
For WASM, we cannot do the same, since there is no way to implement signature-agnostic calls (`jmp`s). Instead, we shift the responsibility to call the target to codegen and let the stubs only do its resolution.

There are a number of auxiliary changes here as well:
1) Null checking fix.
2) PI helpers changed names and we didn't update the portable versions.
3) `OS_PAGE_SIZE` was upped to 4K s.t. the native allocator used for stub caches remains efficient.
4) Calls that needed null checks are now null-checked.
5) Runtime exports will now also have the managed entrypoint exported, for now this is only used by the stub code but in the future will help with fixing #2000.
6) Stack traces are now captured for RyuJit-thrown exceptions.

Finally, this brings us to 97%/88% RyuJit coverage for HelloWasm, with the only remaining significant not implemented features being PI, RPI and atomics / volatile memory operations.


src/coreclr/nativeaot/Runtime/CachedInterfaceDispatch.h
```diff
@@ -25,6 +25,13 @@ struct InterfaceDispatchCacheEntry
 // cache miss processing needs to determine this value in a synchronized manner, so it can't be contained in
 // the owning interface dispatch indirection cell) and a list entry used to link the caches in one of a couple
 // of lists related to cache reclamation.
+
+#if defined(HOST_ARM) || defined(HOST_X86)
+// On ARM and x86 the slow path in the stubs needs to reload the cell pointer from the cache due to the lack
+// of available (volatile non-argument) registers.
+#define INTERFACE_DISPATCH_CACHE_HAS_CELL_BACKPOINTER
```
SingleAccretion, line 32: This bit is intended for upstreaming, so that we don't need to add WASM-specific `#ifdef`s.

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvmlower.cpp
```diff
@@ -1175,6 +1288,17 @@ void Llvm::normalizeStructUse(GenTree* node, ClassLayout* layout)
     }
 }
 
+unsigned Llvm::representAsLclVar(LIR::Use& use)
+{
+    GenTree* node = use.Def();
+    if (!node->OperIs(GT_LCL_VAR))
```
yowl: nit: would this be clearer with a positive if condition?

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/CachedInterfaceDispatch.cs
```diff
@@ -20,6 +20,7 @@ private static unsafe IntPtr RhpCidResolve(IntPtr callerTransitionBlockParam, In
             return dispatchResolveTarget;
         }
 
+        [RuntimeExport("RhpCidResolveWasm")]
```
yowl, line 23: Should this be in a #if ?

yowl (commented):
This is another great advancement.  Admittedly I'm not reviewing much here.  There's some common code changes which maybe you want to tag someone else for?  Otherwise, thanks for the comments and the explanation. Awesome!

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/CachedInterfaceDispatch.cs
SingleAccretion: I did originally write it that way, but it looked ugly:
```
#if TARGET_WASM
        [RuntimeExport("RhpCidResolveWasm")]
#endif
        private static IntPtr RhpCidResolve_Worker(...)
```
Since the export name makes it pretty clear this is Wasm-only functionality, I decided to just delete the `#if`.

SingleAccretion:
@jkotas - could you please take a quick look at the `INTERFACE_DISPATCH_CACHE_HAS_CELL_BACKPOINTER` refactoring? I am planning to get it into runtime proper.

src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
```diff
@@ -3080,8 +3080,18 @@ private void getEEInfo(ref CORINFO_EE_INFO pEEInfoOut)
 
             pEEInfoOut.osPageSize = new UIntPtr(0x1000);
 
-            pEEInfoOut.maxUncheckedOffsetForNullObject = (_compilation.NodeFactory.Target.IsWindows) ?
-                new UIntPtr(32 * 1024 - 1) : new UIntPtr((uint)pEEInfoOut.osPageSize / 2 - 1);
+#if !READYTORUN
+            if (_compilation.TargetArchIsWasm())
+            {
+                // The low 1024 bytes are reserved by Emscripten; no valid address will point there.
+                pEEInfoOut.maxUncheckedOffsetForNullObject = new UIntPtr(1024 - 1);
```
yowl, line 3087: I wonder if this should be configurable, because you can change it with `--global-base` I believe: https://github.com/aduros/wasm4/issues/217#issuecomment-998293727

SingleAccretion: It's a good question. I basically took this value from https://github.com/emscripten-core/emscripten/issues/5775#issuecomment-344049528, which suggests that it would be unsupported to compile with Emscripten and use this region.

I would tend towards saying the same for us. RyuJit is built to assume this value is "reasonably large", if you set it to something small (like zero), you will encounter correctness issues and poor code quality.

yowl: For wasm4 I was not using emscripten for the final link, (this was zero-sharp as well as there is only 64KB to play with).  The global-base required for wasm4 was/is 6560.  So, 1024 would not cause me a problem as any write >= 6560 would be allowed, just the range 1024-6560 would not throw.  We can leave it at 1024 on the basis that it is low enough to allow other runtimes, and the right value for 99% of cases.

jkotas:
> could you please take a quick look at the INTERFACE_DISPATCH_CACHE_HAS_CELL_BACKPOINTER refactoring?

Looks good to me.

PR #2185, opened 2023-01-30
Add Stream boilerplate source-generator and attribute

jozkee:
Forked from https://github.com/dotnet/runtimelab/tree/standalone-template.

* Adds the `System.IO.StreamSourceGeneration` project containing the generator and the `[GenerateStreamBoilerplate]` attribute.

* Adds test verifying generation of a variety of streams that implement one or more members that can be used to hint to the generator what it should do:
  - `StreamThatImplementsNothing`
  - `StreamThatImplementsReadAsyncByte`
  - `StreamThatImplementsReadAsyncMemory`
  - `StreamThatImplementsReadByte`
  - `StreamThatImplementsReadSpan`
  - `StreamThatImplementsSeek`
  - `StreamThatImplementsWriteAsyncByte`
  - `StreamThatImplementsWriteAsyncMemory`
  - `StreamThatImplementsWriteByte`
  - `StreamThatImplementsWriteSpan`

~~This PR does not add, but is intended as a foundation for adding analyzer rules that hint the user about potential improvements on their Stream implementation. e.g: "Prefer overriding Span/Memory overloads instead of calling byte[] ones."~~

* Adds several diagnostics (without fixers).


src/System.IO.StreamSourceGeneration/src/BoilerplateCandidateInfo.cs
```diff
@@ -0,0 +1,60 @@
...
+            new(StreamMembersConstants.EndRead, StreamOperationKind.None, Constants.EndRead, Constants.EndReadUnsupported),
+            new(StreamMembersConstants.EndWrite, StreamOperationKind.None, Constants.EndWrite, Constants.EndWriteUnsupported),
+            new(StreamMembersConstants.Seek, StreamOperationKind.Seek, null, Constants.SeekUnsupported),
+            new(StreamMembersConstants.SetLength, StreamOperationKind.None, null, Constants.SetLengthUnsupported),
+            new(StreamMembersConstants.Length, StreamOperationKind.None, null, Constants.LengthUnsupported),
+            new(StreamMembersConstants.Position, StreamOperationKind.None, null, Constants.PositionUnsupported),
+        };
+
+        s_lookupBoilerplateGenerationCandidates = CandidatesList.ToDictionary(x => x.Name);
+    }
```
bartonjs: These look like they can be done as field initializers and not need a static initializer body.

src/System.IO.StreamSourceGeneration/src/BoilerplateCandidateInfo.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Constants = System.IO.StreamSourceGeneration.StreamBoilerplateConstants;
+
+namespace System.IO.StreamSourceGeneration;
```
bartonjs: Have we used file-scoped namespaces elsewhere? I don't see any in `dotnet/runtime/src/libraries/...`

src/System.IO.StreamSourceGeneration/src/GenerateStreamBoilerplateAttribute.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace System.IO.StreamSourceGeneration;
+
+/// <summary>
+/// Instructs the System.IO.StreamSourceGeneration source generator to generate boilerplate implementations for multiple legacy members, leaving the implementer only with the task of providing implementations for the core operations.
```
bartonjs: Even with a widescreen monitor, maximized browser window, and GitHub using a wider view, this almost requires a horizontal scrollbar.

225 characters is way too long.  120 is a much more reasonable length.

The type marker for System.IO.StreamSourceGeneration should be either in `<c>` if it can't be referenced; or as `<see cref="System.IO.StreamSourceGeneration" />` if it can.


```suggestion
/// Instructs the <c>System.IO.StreamSourceGeneration</c> source generator to generate boilerplate implementations for
/// multiple legacy members, leaving the implementer only with the task of providing implementations for the core operations.
```

src/System.IO.StreamSourceGeneration/src/GenerateStreamBoilerplateAttribute.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace System.IO.StreamSourceGeneration;
```
bartonjs: The `api-suggestion` proposes this in `System.IO`.  Is it intentionally different here?

src/System.IO.StreamSourceGeneration/src/Helpers.cs
```diff
@@ -0,0 +1,170 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.DotnetRuntime.Extensions;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+//using System.Threading;
```
bartonjs: Don't commit commented out code, please (unless it's being used as part of a comment, such as an example)

src/System.IO.StreamSourceGeneration/src/Helpers.cs
```diff
@@ -0,0 +1,170 @@
...
+                IPropertySymbol property => property.OverriddenProperty,
+                _ => null
+            };
+    }
+
+    internal static string GetMemberToCallForTemplate(string candidateName, string memberToCall)
+    {
+        return candidateName switch
+        {
+            StreamMembersConstants.ReadByte => memberToCall switch
```
bartonjs: Can `candidateName` be an enum, instead?  As it is, when "ReadSpan" is passed in it'll have to do a full string comparison against ReadByte, fail, then move to the next item (since they're the same length it can't skip it as "obviously wrong").

Since I don't think the candidate list is user extensible, using an enum and faster numeric matching seems better (at least at this part... maybe strings make sense later).

src/System.IO.StreamSourceGeneration/src/Helpers.cs
```diff
@@ -0,0 +1,170 @@
...
+                StreamMembersConstants.WriteByte => StreamBoilerplateConstants.WriteAsyncMemoryCallsToWriteByte,
+                StreamMembersConstants.WriteSpan => StreamBoilerplateConstants.WriteAsyncMemoryCallsToWriteSpan,
+                StreamMembersConstants.WriteAsyncByte => StreamBoilerplateConstants.WriteAsyncMemoryCallsToWriteAsyncByte,
+                _ => throw new InvalidOperationException()
+            },
+            _ => throw new InvalidOperationException()
+        };
+    }
+
+    internal static List<StreamTypeInfo>? GetClassesWithGenerationOptions(Compilation compilation, ImmutableArray<ClassDeclarationSyntax> classes, Threading.CancellationToken cancellationToken)
```
bartonjs: Line length

src/System.IO.StreamSourceGeneration/src/Helpers.cs
```diff
@@ -0,0 +1,170 @@
...
+                    retVal ??= new List<StreamTypeInfo>();
+                    retVal.Add(streamTypeInfo);
+                }
+            }
+        }
+
+        return retVal;
+    }
+
+    // TODO: merge this method with DerivesFromJsonSerializerContext.
```
bartonjs: I don't see a method named DerivesFromJsonSerializerContext anywhere; how is this supposed to merge with it?

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Read.cs
```diff
@@ -0,0 +1,97 @@
...
+            EnsureCanRead();
+            {0}
+        }}
+";
+    internal const string ReadByteCallsToReadSpan = @"
+            return Read(buffer.AsSpan(offset, count));";
+    internal const string ReadByteCallsToReadAsyncByte = @"
+            return ReadAsync(buffer, offset, count).GetAwaiter().GetResult();";
+    internal const string ReadByteCallsToReadAsyncMemory = @"
+            return ReadAsync(buffer.AsMemory(offset, count), CancellationToken.None).GetAwaiter().GetResult();";
```
bartonjs: Since `ReadByte` is an actual named method that the generator will need to generate / work with, you shouldn't also call `Read(byte[] buffer, int offset, int count)` "ReadByte".  "ReadBytes" or "ReadByteArray" seem more appropriate if "Read" by itself feels lacking.

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Read.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    internal const string ReadSpanTemplate = @"
+        public override int Read(Span<byte> buffer)
+        {{
+            EnsureCanRead();
+            {0}
+        }}
+";
+    internal const string ReadSpanCallsToReadByte = @"
+            return base.Read(buffer);";
```
bartonjs: If we're just going to define the implementation as calling base, why have a declared body at all?

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Unsupported.cs
```diff
@@ -0,0 +1,126 @@
...
+        public override int Read(Span<byte> buffer)
+        {
+            throw new NotSupportedException(""Stream does not support reading."");
+        }
+";
+
+    internal const string ReadAsyncByteArrayUnsupported = @"
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            throw new NotSupportedException(""Stream does not support reading."");
```
bartonjs, line 65: I had to stop and think a bit on the exception model here.

Since this body is presumably only generated when there are no `Read*` methods, and `CanRead` is generated as `false` in that case, it can be said that the caller has committed a usage error by calling `ReadAsync` without checking `CanRead`; and that therefore it is correct to throw this exception in lieu of creating a Task (versus creating a Task to represent the exception).

Of course, if that's the actual justification then it seems like the better answer is an InvalidOperationException... but maybe we have precedent on Stream implementations that NotSupportedException is what we use in these cases.

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Unsupported.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+";
+
+    internal const string WriteAsyncMemoryUnsupported = @"
+        public override ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default)
+        {
+            throw new NotSupportedException(""Stream does not support writing."");
+        }
+";
+
```
bartonjs, line 117: Presumably there should be something here for the `ReadByte` method (in particular, it's better for it to defer to Read(Span) over Read(Array), if Read(Span) is not itself just deferring to Read(Array))

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Write.cs
```diff
@@ -0,0 +1,88 @@
...
+            var vt = new ValueTask(Task.Run(() =>
+            {
+                Write(sharedBuffer, 0, buffer.Length);
+                ArrayPool<byte>.Shared.Return(sharedBuffer);
+            }, cancellationToken));
+
+            return vt;";
+
+    internal const string WriteAsyncMemoryCallsToWriteSpan = @"
+            return new ValueTask(Task.Run(() => Write(buffer.Span), cancellationToken));";
```
bartonjs: @stephentoub Is it better for these to use `Task.Run` with a captured local, or `Task.Factory.StartNew(func, state)` with a manual check on the cancellation token?  (And should we have an overload for Task.Run that takes `(Func<object, Task> func, object state, CancellationToken=default)`?)

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Write.cs
```diff
@@ -0,0 +1,88 @@
+﻿namespace System.IO.StreamSourceGeneration;
```
bartonjs: For all of these files, I assume they should have the standard .NET comment/header.

src/System.IO.StreamSourceGeneration/src/StreamCapabilityInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    internal string GetPreferredMemberName(bool isAsync)
+    {
+        string? retVal = isAsync ?
+            AsyncPreferredMemberName ?? SyncPreferredMemberName :
+            SyncPreferredMemberName ?? AsyncPreferredMemberName;
+
+        Debug.Assert(retVal != null, "Both properties can't be null.");
```
bartonjs: Sure they can.  Maybe you want to move/copy the assert up to the ctor.

src/System.IO.StreamSourceGeneration/src/StreamSourceGen.cs
```diff
@@ -0,0 +1,177 @@
...
+{
+    internal const string StreamBoilerplateAttributeFullName = "System.IO.StreamSourceGeneration.GenerateStreamBoilerplateAttribute";
+    internal const string StreamFullName = "System.IO.Stream";
+
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        context.RegisterPostInitializationOutput(ctx =>
+        {
+            //Diagnostics.Debugger.Launch();
+            ctx.AddSource("TaskToApm.g.cs", System.IO.StreamSourceGeneration.Properties.Resources.TaskToApm);
```
bartonjs: Will this work if more than one Stream type in the same project uses the generator?

Alternatives: Nested types that are generated into each .g.cs as-needed; or just deciding that the odds of someone calling into the IAsyncResult methods is so low that we simplify the machinery (return the Task from BeginRead, and make EndRead just be a nullcheck then `return ((Task<int>)asyncResult).GetAwaiter().GetResult());`)

src/System.IO.StreamSourceGeneration/src/StreamSourceGen.cs
```diff
@@ -0,0 +1,177 @@
...
+public partial class StreamSourceGen : IIncrementalGenerator
+{
+    internal const string StreamBoilerplateAttributeFullName = "System.IO.StreamSourceGeneration.GenerateStreamBoilerplateAttribute";
+    internal const string StreamFullName = "System.IO.Stream";
+
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        context.RegisterPostInitializationOutput(ctx =>
+        {
+            //Diagnostics.Debugger.Launch();
```
bartonjs: Don't commit commented out code, please.

src/System.IO.StreamSourceGeneration/src/StreamSourceGen.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            return;
+        }
+
+#if LAUNCH_DEBUGGER
+        if (!Diagnostics.Debugger.IsAttached)
+        {
+            Diagnostics.Debugger.Launch();
+        }
+#endif
```
bartonjs: If possible, I think it'd be better to leave these lines out of the committed code.

src/System.IO.StreamSourceGeneration/src/StreamSourceGen.cs
```diff
@@ -0,0 +1,177 @@
...
+
+{StreamBoilerplateConstants.UsingDirectives}
+
+namespace {streamTypeInfo.TypeSymbol.ContainingNamespace}
+{{");
+        sb.Append($@"
+    partial class {streamTypeInfo.TypeSymbol.Name}
+    {{");
+
+        HashSet<string> overriddenMembers = streamTypeInfo.OverriddenMembers;
```
bartonjs: Since there aren't a lot of methods, and strings are expensive, I wonder if this wouldn't be better off as just having a struct for the overridden members, with it having a field for each one that we care about.

```C#
internal struct OverriddenMembers
{
    internal SomeWordHere ReadArray;
    internal SomeWordHere ReadSpan;
    ...
}
```

Or, a bit more loosey-goosey, an array indexed by named positions.

src/System.IO.StreamSourceGeneration/src/StreamSourceGen.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            if (overriddenMembers.Contains(candidateName))
+            {
+                continue;
+            }
+
+            StreamOperationKind candidateOperationKind = candidateInfo.OperationKind;
+
+            if (candidateOperationKind >= StreamOperationKind.Read &&
+                candidateOperationKind <= StreamOperationKind.WriteAsync)
```
bartonjs: The range on this feels pretty arbitrary, and I had to look it up.

I don't necessarily have a replacement strategy off the top of my head; but I just thought I'd point it out.

src/System.IO.StreamSourceGeneration/src/Properties/Resources.resx
```diff
@@ -0,0 +1,259 @@
...
+
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace System.Threading.Tasks
+{
+    /// &lt;summary&gt;
+    /// Provides support for efficiently using Tasks to implement the APM (Begin/End) pattern.
+    /// &lt;/summary&gt;
+    internal static class TaskToApm
```
bartonjs: Why is this a resource, when everything else is a const string?

src/System.IO.StreamSourceGeneration/src/System.IO.StreamSourceGeneration.csproj
```diff
@@ -0,0 +1,35 @@
...
+
+	<PropertyGroup>
+		<TargetFramework>netstandard2.0</TargetFramework>
+		<LangVersion>latest</LangVersion>
+		<Nullable>enable</Nullable>
+	</PropertyGroup>
+
+	<PropertyGroup>
+		<DefineConstants Condition="'$(LaunchDebugger)' == 'true'">$(DefineConstants);LAUNCH_DEBUGGER</DefineConstants>
+	</PropertyGroup>
```
bartonjs: Probably doesn't need to be committed.

src/System.IO.StreamSourceGeneration/src/BoilerplateCandidateInfo.cs
jozkee: I see 122 matches of `namespace [A-Za-z0-9.]*;` using VSCode Regex Search. 

Just to name one: https://github.com/dotnet/runtime/blob/53bb518ea482a5960577b151c11cc1297e5dea6d/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/GeneratedRegexAttribute.cs#L8

stephentoub: > Just to name one

That's... surprising.  I must have done that accidentally, since a) we're not changing existing usage and b) our general approach is to match the style used by the rest of the project.

src/System.IO.StreamSourceGeneration/src/GenerateStreamBoilerplateAttribute.cs
jozkee: My mistake, it was not intentional. now that you mention it, I think the correct namespace should be `System.IO`.

For reference:

On System.Text..Json, fully qualified name of source generator is `System.Text.Json.SourceGeneration.JsonSourceGenerator`, of attribute is `System.Text.Json.Serialization.JsonSerializableAttribute`.

On Ms.Extensions.Logging.Abstractions (this one only ships as a package), generator is `Microsoft.Extensions.Logging.Generators.LoggerMessageGenerator`, attribute is `Microsoft.Extensions.Logging.LoggerMessageAttribute`.

On System.Text.RegularExpressions, generator is `System.Text.RegularExpressions.Generator.RegexGenerator`, attribute is `System.Text.RegularExpressions.GeneratedRegexAttribute`.

jozkee: I don't know if this project being in runtimelab, and not in runtime might have weight in the decision.

src/System.IO.StreamSourceGeneration/src/Helpers.cs
jozkee: I was referring to [this method](https://github.com/dotnet/runtime/blob/30cc26fae6439707097ebd07145e8600e99a416c/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Parser.cs#L400) which can be abstracted as a helper and be reused.

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Read.cs
jozkee: If this is the best implementation of `Read(Span<byte>)` calling `Read(byte[], int, int)`, then yes, we can add logic to skip code emission for this case.

stephentoub: Yes, if the base implementation is the best implementation, we just shouldn't emit an override.

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Write.cs
stephentoub: I've not yet read through the PR yet.  What is the case where this would be emitted?  If we'd be using this when no other WriteAsync methods were overridden, the right answer is most likely to not override this at all and just let the base implementation handle it.

src/System.IO.StreamSourceGeneration/src/StreamCapabilityInfo.cs
jozkee: > Sure they can. 

How? ctor sets one or the other with a non-nullable `string`.

src/System.IO.StreamSourceGeneration/src/StreamSourceGen.cs
jozkee: > Will this work if more than one Stream type in the same project uses the generator?

Yes, `RegisterPostInitializationOutput` is meant to emit things only once (per compilation (is that the right term?)).

> Alternatives:

Before we go further to think about alternatives, I want to mention that the `TaskToApm` helpers are meant to be public API for 8.0, see https://github.com/dotnet/runtime/issues/61729.

src/System.IO.StreamSourceGeneration/src/StreamSourceGen.cs
jozkee: Once you build the generator with `Diagnostics.Debugger.Launch()`, the debugging prompt becomes quite annoying and I've noticed that it doesn't go away even if you build again without it. you need to re-launch Visaul Studio. That's why I kept this in the source; also, I borrowed this from https://github.com/dotnet/runtime/blob/07f10e789f06e8ab94dbbe8c870858ab2eb69cf9/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Roslyn4.0.cs#L41-L46.

src/System.IO.StreamSourceGeneration/src/Properties/Resources.resx
jozkee: Tbh, I was looking for the best way to embed a full file that is meant to be emitted. If having a const with the whole definition is better, then I can stick to that.

src/System.IO.StreamSourceGeneration/src/System.IO.StreamSourceGeneration.csproj
jozkee: Same reply as https://github.com/dotnet/runtimelab/pull/2185#discussion_r1107754088.

src/System.IO.StreamSourceGeneration/src/StreamCapabilityInfo.cs
bartonjs: > ctor sets one or the other with a non-nullable string.

Nullability annotations are a hint, not a guarantee.  If this assert is just to make the nullability flow happy, then OK; but if it's actually trying to detect problems then the ctor should also have an assert (or throw) that the input wasn't actually null.

src/System.IO.StreamSourceGeneration/src/StreamSourceGen.cs
jozkee: I switched to a `HashSet<StreamMember>`, where StreamMember is an enum, let me know if that still concerns you.

src/System.IO.StreamSourceGeneration/src/Properties/Resources.resx
jozkee: I moved the TaskToApm code to a string const and removed the resx file.

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Unsupported.cs
jozkee: > we have precedent on Stream implementations that NotSupportedException is what we use in these cases.

https://github.com/dotnet/runtime/blob/c108266e9b10deb7ca3fb92ac90a6016b5f04f03/src/libraries/System.Private.CoreLib/src/System/IO/FileStream.cs#L302

https://github.com/dotnet/runtime/blob/c108266e9b10deb7ca3fb92ac90a6016b5f04f03/src/libraries/System.Private.CoreLib/src/System/IO/BufferedStream.cs#L1013

src/System.IO.StreamSourceGeneration/src/StreamBoilerplateConstants.Unsupported.cs
jozkee: Ah right, I forgot about ReadByte and WriteByte.

jozkee:
Ping @stephentoub @bartonjs, this is ready for re-review.

jozkee:
Ping @stephentoub.

I'm planning on adding tests to ensure we emit the proper diagnostics, functional tests checking the behavior of the generated code already exists. In the meantime, this PR could benefit from a code review.

stephentoub (approved):

PR #2195, opened 2023-02-08
NativeAOT-LLVM: Remove GT_PUTARG_TYPE

yowl:
This PR merges PR 68748 from runtime, removing the `GT_PUTARG_TYPE` node.  It adds a `CorInfoType` field to `NewCallArg` to track pointer types.

### DIFF
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -1765,8 +1765,6 @@ public:
     // Tunnel through any GT_RET_EXPRs
     GenTree* gtRetExprVal(BasicBlockFlags* pbbFlags = nullptr);
 
-    inline GenTree* gtSkipPutArgType();
-
     // Return the child of this node if it is a GT_RELOAD or GT_COPY; otherwise simply return the node itself
     inline GenTree* gtSkipReloadOrCopy();
 
@@ -4098,6 +4096,8 @@ public:
     // argument type, but when a struct is passed as a scalar type, this is
     // that type. Note that if a struct is passed by reference, this will still
     // be the struct type.
+    // TODO-ARGS: Reconsider whether we need this, it does not make much sense
+    // to have this instead of using just the type of the arg node.
     var_types ArgType : 5;
     // True when the argument fills a register slot skipped due to alignment
     // requirements of previous arguments.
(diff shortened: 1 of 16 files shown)


yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -2767,8 +2765,8 @@ FunctionType* Llvm::createFunctionTypeForCall(GenTreeCall* call)
 
     for (CallArg& callArg: call->gtArgs.Args())
     {
-        GenTreePutArgType* putArg = callArg.GetNode()->AsPutArgType();
-        argVec.push_back(getLlvmTypeForCorInfoType(putArg->GetCorInfoType(), putArg->GetClsHnd()));
+        GenTree* putArg = callArg.GetNode();
```
```suggestion
```
Now unused.

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1983,23 +1982,22 @@ void Llvm::buildCall(GenTreeCall* call)
...
-        Type*    argLlvmType = getLlvmTypeForCorInfoType(lastArg->GetCorInfoType(), lastArg->GetClsHnd());
+        Type*    argLlvmType = getLlvmTypeForCorInfoType(callArg.GetSignatureCorInfoType(), callArg.GetSignatureClassHandle());
         Value*   argValue;
 
-        if (argNode->OperIs(GT_FIELD_LIST))
+        if (lastArg->OperIs(GT_FIELD_LIST))
         {
-            assert(lastArg->GetCorInfoType() == CORINFO_TYPE_VALUECLASS);
-            argValue = buildFieldList(argNode->AsFieldList(), argLlvmType);
+            assert(callArg.GetSignatureType() == var_types::TYP_STRUCT);
```
```suggestion
            assert(callArg.GetSignatureType() == TYP_STRUCT);
```

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1983,23 +1982,22 @@ void Llvm::buildCall(GenTreeCall* call)
...
-        lastArg = callArg.GetNode()->AsPutArgType();
+        lastArg = callArg.GetNode();
 
-        GenTree* argNode     = lastArg->gtGetOp1();
-        Type*    argLlvmType = getLlvmTypeForCorInfoType(lastArg->GetCorInfoType(), lastArg->GetClsHnd());
+        Type*    argLlvmType = getLlvmTypeForCorInfoType(callArg.GetSignatureCorInfoType(), callArg.GetSignatureClassHandle());
         Value*   argValue;
 
-        if (argNode->OperIs(GT_FIELD_LIST))
+        if (lastArg->OperIs(GT_FIELD_LIST))
```
SingleAccretion: Let's keep `argNode`, `lastArg` isn't as descriptive (and not really correct, since it is actually the current arg).

src/coreclr/jit/llvmlower.cpp
```diff
@@ -929,12 +916,15 @@ void Llvm::lowerVirtualStubCallAfterArgs(
     stubAddr->gtFlags |= GTF_IND_NONFAULTING;
     CurrentRange().InsertBefore(callNode, stubAddr);
 
+    GenTree* thisForStub  = _compiler->gtNewLclvNode(thisArgLclNum, TYP_REF);
```
SingleAccretion: Nit: I would prefer to keep the original order for how `thisForStub` is created/inserted. It is meant to match the final operand layout (SS -> `this` -> stub call target as `stubAddr`).

src/coreclr/jit/llvmlower.cpp
```diff
@@ -1128,14 +1111,10 @@ unsigned Llvm::lowerCallToShadowStack(GenTreeCall* callNode)
             if (!argNode->OperIs(GT_FIELD_LIST) && argNode->TypeIs(TYP_STRUCT))
             {
                 normalizeStructUse(argNode, _compiler->typGetObjLayout(clsHnd));
+                callArg->SetSignatureClassHandle(clsHnd);
```
SingleAccretion, line 1121:
```suggestion
                // TODO-LLVM: delete (together with 'SetSignatureClassHandle') when merging https://github.com/dotnet/runtime/pull/69969 (May 31).
                callArg->SetSignatureClassHandle(clsHnd);
```
I realize it is currently necessary to reset the class handle because of issues with canonical structs, but it should be done for all `TYP_STRUCT` args (this condition excludes `FIELD_LIST`s).

src/coreclr/jit/gentree.h
```diff
@@ -4253,12 +4258,19 @@ struct NewCallArg
         return arg;
     }
 
+#ifdef TARGET_WASM
+    static NewCallArg Primitive(GenTree* node, var_types type = TYP_UNDEF, CorInfoType corInfoType = CORINFO_TYPE_UNDEF)
```
SingleAccretion: Instead of this `#ifdef`y way, I think it would better to define a new function:
```
static NewCallArg Primitive(GenTree* node, CorInfoType corInfoType)
{
    assert(corInfoType != CORINFO_TYPE_UNDEF);
    NewCallArg arg = Primitive(node, JITtype2varType(corInfoType));
#ifdef TARGET_WASM
    arg.SignatureCorInfoType = corInfoType;
#endif
    return arg;
}
```
We should use it in `impPopCallArgs` to store the signature type directly. It should allow us to delete the `#ifdef TARGET_WASM` code dealing with signatures for user calls from importer. There is still one case in rationalization (`Rationalizer::RewriteNodeAsCall`) where user calls are created that will need to be handled specially for WASM.

src/coreclr/jit/gentree.h
```diff
@@ -4332,7 +4354,11 @@ class CallArg
     // clang-format off
     GenTree*& EarlyNodeRef() { return m_earlyNode; }
     GenTree* GetEarlyNode() { return m_earlyNode; }
+#ifdef TARGET_WASM
+    void SetEarlyNode(GenTree* node, CorInfoType signatureCorInfoType = CORINFO_TYPE_UNDEF) { m_earlyNode = node; m_signatureCorInfoType = signatureCorInfoType; }
```
SingleAccretion: Adding a separate setter for `m_signatureCorInfoType` seems better.

src/coreclr/jit/gentree.h
yowl: I think I'm going to have to allow the var_type to be passed in as well as from `JITtype2varType` or else I hit a problem in `impCheckImplicitArgumentCoercion` with `CORINFO_TYPE_PTR` and `TYP_REF` for `NewCallArg::Primitive(thisForStub, CORINFO_TYPE_PTR)`.  `thisForStub` is of type `TYP_REF`

SingleAccretion: The original code used `CORINFO_TYPE_CLASS` for `thisForStub`, was it changed because that caused other issues?

yowl: No, I can put that back.  I've lost ~18% of coverage somewhere, so need to see what happened there.

SingleAccretion: It may be due to the removal of signature code. We probably have to keep until the issue with canonical structs is resolved.

yowl: Yes, that is when it went.  I've put it back

src/coreclr/jit/llvmlower.cpp
yowl: Right, thanks, I've put that back now.

src/coreclr/jit/gentree.h
```diff
@@ -4287,6 +4300,10 @@ class CallArg
     unsigned m_tmpNum;
     // The type of the argument in the signature.
     var_types m_signatureType : 5;
+#ifdef TARGET_WASM
+    // The CorInfoType of argument in the signature.
+    CorInfoType m_signatureCorInfoType : 8;
```
SingleAccretion, line 4305: The field should also be copied here:

https://github.com/dotnet/runtimelab/blob/55680b0e768fd6ad4a23c12b8859266b79faedc9/src/coreclr/jit/gentree.cpp#L9133-L9146

src/coreclr/jit/llvmlower.cpp
```diff
@@ -930,11 +918,13 @@ void Llvm::lowerVirtualStubCallAfterArgs(
...
     GenTreeCall* stubCall = _compiler->gtNewIndCallNode(stubAddr, TYP_I_IMPL);
-    stubCall->gtArgs.PushFront(_compiler, shadowStackForStubPutArg, thisForStubPutArg, cellForStubPutArg);
+    stubCall->gtArgs.PushFront(_compiler, NewCallArg::Primitive(shadowStackForStub, CORINFO_TYPE_PTR),
+                               NewCallArg::Primitive(thisForStub, CORINFO_TYPE_CLASS),
+                               NewCallArg::Primitive(cellArgNode, CORINFO_TYPE_PTR));
     stubCall->gtCorInfoType = CORINFO_TYPE_PTR;
     stubCall->gtFlags |= GTF_CALL_UNMANAGED;
     stubCall->gtCallMoreFlags |= GTF_CALL_M_SUPPRESS_GC_TRANSITION;
-    CurrentRange().InsertBefore(callNode, stubCall);
+    CurrentRange().InsertBefore(callNode, thisForStub, stubCall);
```
```suggestion
    CurrentRange().InsertBefore(callNode, stubCall);
```

src/coreclr/jit/llvmlower.cpp
```diff
@@ -892,20 +892,8 @@ void Llvm::lowerVirtualStubCallAfterArgs(
...
-    GenTree* thisForStub = _compiler->gtNewLclvNode(thisArgLclNum, TYP_REF);
-    GenTreePutArgType* thisForStubPutArg = _compiler->gtNewPutArgType(thisForStub, CORINFO_TYPE_CLASS, NO_CLASS_HANDLE);
-    INDEBUG(thisForStubPutArg->SetArgNum(0));
-    CurrentRange().InsertBefore(callNode, thisForStub, thisForStubPutArg);
-
-    GenTreePutArgType* cellForStubPutArg = _compiler->gtNewPutArgType(cellArgNode, CORINFO_TYPE_PTR, NO_CLASS_HANDLE);
-    INDEBUG(cellForStubPutArg->SetArgNum(1));
-    CurrentRange().InsertBefore(callNode, cellForStubPutArg);
+    GenTree* shadowStackForStub       = insertShadowStackAddr(callNode, shadowStackOffsetForStub, _shadowStackLclNum);
+    GenTree* thisForStub              = _compiler->gtNewLclvNode(thisArgLclNum, TYP_REF);
```
```suggestion
    GenTree* shadowStackForStub = insertShadowStackAddr(callNode, shadowStackOffsetForStub, _shadowStackLclNum);
    GenTree* thisForStub = _compiler->gtNewLclvNode(thisArgLclNum, TYP_REF);
    CurrentRange().InsertBefore(callNode, thisForStub);
```
(Indentation changes - since we decided not aligning things)

SingleAccretion (approved):
Looks good to me, thank you!

jkotas:
Thank you!

PR #2201, opened 2023-02-21
[NativeAOT-LLVM] Implement PInvoke

SingleAccretion:
A large change, but pretty much everything here is required for a quality implementation. Even so, the ABI handling is still missing.

The largest challenge with implementing foreign calls for WASM is, of course, the problem of mismatched signatures. There are two cases to consider:
1) Static linking.
    - Here, the linker will know what the ultimate definition is, so we should be able to replicate the desired behavior of runtime errors for wrong signatures (instead of compile-time ones). Indeed, this is more or less straightforward if we "hide" the target symbol from LLVM, such that all foreign calls are indirect and guaranteed to stay that way in the WASM object file. The static linker (`wasm-ld`) allows functions which only show up as table index relocations to have mismatched signatures. In this change, the "hiding" at LLVM level is achieved by putting all declarations into a separate module. This is not quite resilient against LLVM-level LTO, but we don't currently support that, and if/when when we cross that bridge, a more sophisticated strategy (involving inline assembly) could (will) be used. I implemented this strategy in an earlier version of this change, which attempted to support mismatched signatures for dynamically imported functions, but that ultimately did not work out and so got pulled out as unnecessary.
2) Dynamic linking aka WASM imports:
    - Here, we will not have the definition until runtime. For this, the thinking was that we could take advantage of the overloading that WASM imports support (you can import the same name with different signatures), however, this proved to not be feasible, for two reasons:
      - Emscripten's JS-generating code doesn't expect or handle this.
      - It is not possible to tell whether a given symbol was statically linked because we have to compile with `--import-undefined` which turns all undefined functions into WASM imports automatically. So, while we could generate:
```
if (symbol_statically_linked)
    call symbol
else
    call dynamic_symbol_<sig> (imported as "symbol")
```
There is no way, as far as I able to tell (and I did spend an inordinate amount of time on this), to implement `symbol_statically_linked` without doing disproportionately expensive things like post-processing the WASM object files or calling out to JS functions.

Thus, the compromise solution, implemented in this change, is to only handle the static case and warn the user for the dynamic one. This required some Jit/EE interface perturbations as it implies the generation of external symbols has to be delayed until after all of the references have been processed. An earlier version of this change required more extensive interface changes, I've kept the parts meant to make MT compilation easier in the future.

With this, we are at 95.8% coverage for the Release version of `HelloWasm` and very close to 100% for the Debug version.


SingleAccretion:
Took some iterations but I am reasonably happy with the result now.

@dotnet/nativeaot-llvm

src/coreclr/jit/llvm.cpp
```diff
@@ -208,6 +131,28 @@ bool Llvm::needsReturnStackSlot(const GenTreeCall* callee)
...
+var_types Llvm::GetArgTypeForStructWasm(CORINFO_CLASS_HANDLE structHnd, structPassingKind* pPassKind, unsigned size)
+{
+    assert(size != 0);
+    //
+    // WASM C ABI is documented here: https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md.
+    // In essense, structs are passed by reference except if they are trivial wrappers of a primitive (scalar).
+    // Additionally, structs which cannot be passed on the LLVM stack are passed on the shadow one in the managed
+    // calling convention.
+    //
+    // However, we currently do not conform to this ABI and so cannot pass non-trivial structs to PI methods.
```
yowl, line 143: This presumably doesn't stop us passing complex struct by `ref` or with `&` ?

yowl (approved):
Lots of good things here.  Just a comment about ref structs which I assume are fine as its just a pointer.

src/coreclr/jit/llvm.cpp
SingleAccretion: Yep.

I hope to fix it soon enough anyway.

jkotas (commented):
Thank you!

Issue #2196, opened 2023-02-09
Recent WASM package have some issues.


kant2002:
I update version of packages in my test project to
```
<PackageReference Include="Microsoft.DotNet.ILCompiler.LLVM" Version="7.0.0-preview.5.23106.1" />
<PackageReference Include="runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM" Version="7.0.0-preview.5.23106.1" />
```
and now have issues. 
```
c:\cache\nuget\packages\microsoft.dotnet.ilcompiler.llvm\7.0.0-preview.5.23106.1\build\Microsoft.DotNet.ILCompiler.targets(55,5): error : The PackageDefinitions 
ItemGroup is required for target ImportRuntimeIlcPackageTarget [C:\github\NativeAOTWasm\NativeAOTWasm.csproj]
```

I manage to fix this using following chages
```
<PropertyGroup>
    <PublishAot>true</PublishAot>
</PropertyGroup>

<ItemGroup>
      <KnownILCompilerPack Update="Microsoft.DotNet.ILCompiler"
                         TargetFramework="net7.0"
                         ILCompilerPackNamePattern="runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM"
                         ILCompilerPackVersion="7.0.0-preview.5.23106.1"
                         ILCompilerRuntimeIdentifiers="browser-wasm;linux-musl-x64;linux-x64;win-x64;linux-arm;linux-arm64;linux-musl-arm;linux-musl-arm64;osx-arm64;osx-x64;win-arm;win-arm64;win-x86"
                         />
</ItemGroup>
```
and adding to file `microsoft.dotnet.ilcompiler.llvm\7.0.0-preview.5.23106.1\build\Microsoft.NETCore.Native.Browser.props`
```
<PropertyGroup>
   <!-- Should we have Microsoft.DotNet.ILCompiler.SingleEntry.targets ? -->
      <ILCompilerTargetsPath>$(MSBuildThisFileDirectory)Microsoft.DotNet.ILCompiler.targets</ILCompilerTargetsPath>
</PropertyGroup>
```

Even after that I receive following error
```
EXEC : error : One or more errors occurred. (Code generation failed for method '[S.P.CoreLib]Internal.Runtime.CompilerHelpers.ThrowHelpers.ThrowInstanceBodyRemoved()'
[...]

PR #2203, opened 2023-02-23
Create Microsoft.DotNet.ILCompiler.LLVM.props

kant2002:
Missing bit fo https://github.com/dotnet/runtimelab/issues/2196

You would still required to have 
```
<KnownILCompilerPack Update="Microsoft.DotNet.ILCompiler" 
       TargetFramework="net7.0"
       ILCompilerPackNamePattern="runtime.**RID**.Microsoft.DotNet.ILCompiler.LLVM"
       ILCompilerPackVersion="7.0.0-preview.5.23113.1"
       ILCompilerRuntimeIdentifiers="browser-wasm;linux-musl-x64;linux-x64;win-x64;linux-arm;linux-arm64;linux-musl-arm;linux-musl-arm64;osx-arm64;osx-x64;win-arm;win-arm64;win-x86"
       />
```

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.LLVM.props
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.LLVM.props
@@ -1,0 +1,16 @@
+<!--
+***********************************************************************************************
+Microsoft.DotNet.ILCompiler.props
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
+<Project>
+  <PropertyGroup>
+    <!-- N.B. The ILCompilerTargetsPath is used as a sentinel to indicate a version of this file has already been imported. It will also be the path
+         used to import the targets later in the SDK. -->
+    <ILCompilerTargetsPath>$(MSBuildThisFileDirectory)Microsoft.DotNet.ILCompiler.SingleEntry.targets</ILCompilerTargetsPath>
+  </PropertyGroup>
+</Project>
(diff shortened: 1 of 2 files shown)


SingleAccretion (commented):
I need to understand why the `EmitLegacyAssetsFileItems` workaround did not work.

Do you have a binlog of the failure?

kant2002:
You can add `<PublishAot>true</PublishAot>` and you will see an issue.
Process of resolving host/runtime ILC packages depends on the `KnownILCompilerPack`. If you set `PublishAot=true`, that’s required to have set `IlcHostPackagePath` and `RuntimePackagePath` and point them to proper locations.

if you do not specify `<PublishAot>true</PublishAot>` then `EmitLegacyAssetsFileItems` indeed helps.

SingleAccretion:
> `PublishAot` breaks things

I see. In the current state of things, this is "intentional", `PublishAot` is used as a discriminator for "doing SDK-based publish" versus "doing package-based publish".

Once we merge enough of upstream this will, of course, change.

> Process of resolving host/runtime ILC packages depends on the KnownILCompilerPack. If you set PublishAot=true, that’s required to have set IlcHostPackagePath and RuntimePackagePath and point them to proper locations.

This is concerning. Am I correct this means this change doesn't fix the publishing as-is, and you need to override `KnownILCompilerPack` as well (how would you do that)?

kant2002:
I put snippet in the original comment to PR. That’s all what’s needed.

SingleAccretion:
Okay, I will need some time to internalize how the SDK-driven publishing works. The goal would of course be that the user doesn't need to override anything manually.

kant2002:
I hope that my suggestion https://github.com/dotnet/runtime/issues/73678#issuecomment-1442220142 would be approved. That’s all what we need from what I see. And we are have default experience “as documented”.

kant2002:
@SingleAccretion I think we can safely merge this, since it works in 2 places 😄

SingleAccretion (approved):
LGTM, thank you!

Sorry for the wait; the logic flow here was not straightforward to untangle.

PR #2208, opened 2023-02-25
[NativeAOT-LLVM] Implement RPI

SingleAccretion:
Interesting parts:

1) Introduce the "LLVM-specific helper" mechanism. Use it to hide the shadow stack top variable.
2) Delete the bespoke RPI generation for managed main/startup methods by supporting non-ECMA UCO methods in ILC's LLVM generation.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/wasm/PInvoke.cpp
+++ b/src/coreclr/nativeaot/Runtime/wasm/PInvoke.cpp
@@ -1,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#include <stdlib.h>
+
+#include "CommonTypes.h"
+#include "CommonMacros.h"
+
+extern "C" void* t_pShadowStackBottom = nullptr;
+extern "C" void* t_pShadowStackTop;
+
+COOP_PINVOKE_HELPER(void*, RhpGetOrInitShadowStackTop, ())
+{
+    void* pShadowStack = t_pShadowStackTop;
+    if (pShadowStack == nullptr)
+    {
+        pShadowStack = malloc(1000000); // ~1MB.
+        t_pShadowStackBottom = pShadowStack;
+    }
+
+    return pShadowStack;
+}
+
+COOP_PINVOKE_HELPER(void*, RhpGetShadowStackTop, ())
+{
+    return t_pShadowStackTop;
+}
+
+COOP_PINVOKE_HELPER(void, RhpSetShadowStackTop, (void* pShadowStack))
+{
+    t_pShadowStackTop = pShadowStack;
+}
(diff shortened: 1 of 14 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/nativeaot/Runtime/wasm/PInvoke.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include <stdlib.h>
+
+#include "CommonTypes.h"
+#include "CommonMacros.h"
+
+extern "C" void* t_pShadowStackBottom = nullptr;
+extern "C" void* t_pShadowStackTop;
+
+COOP_PINVOKE_HELPER(void*, RhpGetOrInitShadowStackTop, ())
```
yowl, line 12: This inits the `t_pShadowStackBottom` but the name reflects the logical function rather than the implementation?

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -570,95 +552,6 @@ private void EmitReversePInvokesAndShadowStackBottom()
...
-
-            LLVMValueRef rhpReversePInvokeReturn = Module.GetNamedFunction("RhpReversePInvokeReturn");
-            LLVMTypeRef reversePInvokeFunctionType = LLVMTypeRef.CreateFunction(LLVMTypeRef.Void, new LLVMTypeRef[] { LLVMTypeRef.CreatePointer(reversePInvokeFrameType, 0) }, false);
-            builder.BuildCall(rhpReversePInvokeReturn, new LLVMValueRef[] { reversePinvokeFrame }, "");
-
-            builder.BuildRet(mainReturn);
-            mainFunc.Linkage = LLVMLinkage.LLVMExternalLinkage;
-        }
-
-        private void EmitNativeStartup(LLVMContextRef context)
```
yowl, line 621: I dont think we have a test for this do we?

yowl (commented):
Looks like some nice code deletion.  A couple of comments really.  Theres  a SO article that gets a few views about the native lib, so I just wondered if that was tested?

src/coreclr/nativeaot/Runtime/wasm/PInvoke.cpp
SingleAccretion: Yes... I renamed this a few times, happy to try more options if they are clearer, e. g. `RhpGetShadowStackTopWithInit`.

The semantics of these helpers are tied very strongly to what the code generator does, so the fact we don't set `t_pShadowStackTop` is because we know it would be a dead store.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
SingleAccretion: Indeed we do not (but should).

I have tested it manually and it works.

Notably, https://stackoverflow.com/a/70705808 needs to be updated to account for the fact `CoreRT_StaticInitialization` was renamed into `NativeAOT_StaticInitialization`.

yowl (approved):

src/coreclr/nativeaot/Runtime/wasm/PInvoke.cpp
yowl: I think its fine, nothing better comes to mind really.

jkotas:
Thank you!

PR #2199, opened 2023-02-16
NativeAOT-LLVM: Merge to August 2022

yowl:
Contrary to the branch name this PR merges up to August, commit 442c137891821a567e9a05411f821dbf2aec5aa5

LoongArch addition, some test changes.

partial implementation of `GT_STORE_LCL_FLD` - there are some TODO's left .  The merge is large already so can leave the TODO's for later or do them here?

Excluded some of the new tests that failed.

Removed the Linux release job as it was failing - maybe more merging will make it work again...

PR #2212, opened 2023-02-25
[NativeAOT-LLVM] Implement more things

SingleAccretion:
Please refer to the commit list for details.

List of things still not implemented in RyuJit:
1) MD arrays - will be fixed in #2199.
2) `TYP_STRUCT` `LCL_FLD` - will be fixed in #2199.
3) Correct ABI for structs - will be fixed after the IL backend is phased out.
4) Some calls without signatures - also depends on #2199.
5) Volatile operations and memory ordering - not critical for IL backend deletion.

The Debug logs have some nice numbers:
```
RyuJIT compilation results, total methods 16168 RyuJit Methods 16168 100.0000%
```

### DIFF
--- a/src/coreclr/jit/llvmcodegen.cpp
+++ b/src/coreclr/jit/llvmcodegen.cpp
@@ -2698,13 +2698,7 @@ llvm::CallBase* Llvm::emitCallOrInvoke(llvm::FunctionCallee callee, ArrayRef<Val
 
 FunctionType* Llvm::getFunctionType()
 {
-    // TODO-LLVM: delete this when these signatures implemented
-    if (_sigInfo.hasExplicitThis())
-        failFunctionCompilation();
-
     std::vector<llvm::Type*> argVec(_llvmArgCount);
-    llvm::Type*              retLlvmType;
-
     for (unsigned i = 0; i < _compiler->lvaCount; i++)
     {
         LclVarDsc* varDsc = _compiler->lvaGetDesc(i);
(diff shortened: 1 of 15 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -2313,6 +2313,12 @@ void Llvm::buildIntrinsic(GenTreeIntrinsic* intrinsicNode)
     mapGenTreeToValue(intrinsicNode, intrinsicValue);
 }
 
+void Llvm::buildMemoryBarrier(GenTree* node)
+{
+    assert(node->OperIs(GT_MEMORYBARRIER));
+    _builder.CreateFence(llvm::AtomicOrdering::AcquireRelease);
```
yowl, line 2319: Just for my education, what does this actually do?  I read that it orders memory instructions?

yowl (approved):
looks good, just a question for my own benefit, thanks!

src/coreclr/jit/llvmcodegen.cpp
SingleAccretion: Yes. It is required for ordering loads/stores in the presence of multi-threaded access.

The whole picture is very complicated, but the simple version is that "acquire" operations disallow moving loads/stores before them (they "acquire a lock" / "enter the critical section"), while "release" - past them (they "release the lock" / "exit the critical section"), thus "acquire + release" disallows moving memory accesses across it in either direction.

PR #2217, opened 2023-02-27
[NativeAOT-LLVM] Let the sun set over the IL backend

SingleAccretion:
This changes deletes the last two fallbacks to the IL backend from RyuJit. Unmanaged struct ABI is for now NYI, and we stop using signatures when lowering user calls.

I have also fixed a number of small bugs, re-enabled a couple of tests and started deleting IL backend-specific code.


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvm.cpp
```diff
@@ -744,19 +744,6 @@ CORINFO_GENERIC_HANDLE Llvm::getSymbolHandleForClassToken(mdToken token)
     return CORINFO_GENERIC_HANDLE(typeSymbolHandle);
 }
 
-[[noreturn]] void Llvm::failFunctionCompilation()
```
yowl, line 747: deleted, awesome!

yowl (approved):
Brilliant.  I don't know how many times faster this got done with your input, but it was a lot.  Many thanks!

PR #2218, opened 2023-03-01
[NativeAOT-LLVM] Delete the IL backend

SingleAccretion:
This change deletes the last dependency on the IL backend code (`cctor` trigger code and thread static base access), removes it and migrates all of the remaining LLVM generation in ILC to the object writer, which gains support for opaque pointers.

Notably, the LLVM type system code in ILC remains for now, as we need to handle defining arbitrary functions for tentative (instance) methods and unboxing thunks.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
LGTM, thanks

Issue #2190, opened 2023-02-03
[NativeAOT-LLVM] PI-related issues


SingleAccretion:
We need to implement PI for WASM, there are two issues:

1) LLVM doesn't support lowering its aggregate types using C ABI rules, thus we need to lower them ourselves. The scheme would likely involve taking advantage of the frontend-provided handling for argument morphing.
2) We need to support wrong PI declarations, i. e. we need to be able to compile the following:
```cs
private static double Problem(int arg)
{
    [DllImport("*", EntryPoint = "__pinvoke")]
    static extern double __pinvoke1(double a, long b);
    [DllImport("*", EntryPoint = "__pinvoke")]
    static extern long __pinvoke2(long a, double b);

    switch (arg)
    {
        case 1:
            return __pinvoke1(1, 1);
        case 2:
            return __pinvoke2(2, 2);
        default:
            return 1;
    }
}
```
And have it work correctly at compile time and runtime when:
1) The ultimate definition of `__pinvoke` matches `__pinvoke1` and `arg` equals `1`.
2) The ultimate definition of `__pinvoke` matches `__pinvoke2` and `arg` equals `2`.
3) The ultimate definition of `__pinvoke` doesn't match either `__pinvoke1` or `__pinvoke2` but `arg` never equals `1` or `2`.

This implies we cannot emit direct calls to any PI methods as they could be using the wrong definition and that would fail at compile time (as invalid LLVM IR).

We cannot emit "naive" indirect calls like the IL backend does today either, because LLVM's WASM lowering contains a [pass](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/WebAssembly/WebAssemblyFixFunctionBitcasts.cpp) which legalizes such calls by creating trapping stubs for all references save those matching the declaration. We cannot let LLVM create the stub for any single reference as we do not know which will be the correct one at runtime.

We therefore need to wrap the PI symbol reference into something that will force LLVM to emit a `call_indirect` in WASM, e. g. a volatile load from a global initialized with the function's address.

jkotas:
> We need to support wrong PI declarations

Do these wrong PI declarations exist in something we care about? I am wondering whether it would be better to not support them, for now at last.

SingleAccretion:
> Do these wrong PI declarations exist in something we care about?

Nothing that I would know, though that is a weak assertion as I haven't really looked.

> I am wondering whether it would be better to not support them, for now at last.

[...]

PR #2220, opened 2023-03-02
[NativeAOT-LLVM] Pass structs by reference in accordance with the WASM C ABI

### LINKED ISSUE #2151: NativeAOT-LLVM: LLVM structs over single pointers, should be ptr

SingleAccretion:
And switch from generating named struct types to unnamed; we no longer need the former with opaque pointers.

Closes #2151.
Contributes to #2190.

We still need to implement the return counterpart to this code, that will come in a separate change.


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas:
@yowl Could you please review this and other @SingleAccretion PRs when you get a chance?

yowl (approved):
LGTM, thanks!

PR #2222, opened 2023-03-03
[NativeAOT-LLVM] Spill GC temps live across safe points

SingleAccretion:
We now run a pre-pass before lowering, detecting cases when we need to spill and doing so. The algorithm is the simple "produce, check interference, consume" one, augmented a little with a cache for the spill temps, and complicated by the presence of contained `FIELD_LIST`s.

It is rather unfortunate we need another full IR walk for this, but in the long run, this code will get folded into the general shadow stack allocator algorithm (which will also be significantly more expensive), so it's not that important to optimize it now.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks.

PR #2224, opened 2023-03-04
[NativeAOT-LLVM] Implement support for lazy virtual tables

SingleAccretion:
We can now compile things with scanner disabled.

Two motivating reasons:
1) Reducing differences with other targets.
2) Allows users to work around scanner-specific bugs.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

PR #2226, opened 2023-03-05
[NativeAOT-LLVM] Fix warnings and use LLVM helpers for EH dispatchers

SingleAccretion:
Recent changes have lead to a number of warnings when compiling tests, this change solves them by `#ifdef`ing things out.

It also moves the EH dispatchers to use the recently introduced LLVM-specific helper mechanism and enables a passing test.


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvm.h
```diff
@@ -78,11 +85,11 @@ enum HelperFuncInfoFlags
 
 struct HelperFuncInfo
 {
-    static const int MAX_SIG_ARG_COUNT = 3;
+    static const int MAX_SIG_ARG_COUNT = 4;
 
     INDEBUG(unsigned char Func);
     unsigned char SigReturnType;
-    unsigned char SigArgTypes[MAX_SIG_ARG_COUNT];
+    unsigned char SigArgTypes[MAX_SIG_ARG_COUNT + 1];
```
yowl, line 92: Why is the `+1` needed here?

SingleAccretion: The array is zero-terminated - we need one type at the end to always be UNDEF. It's a bug fix - we just didn't have any helpers with 3 arguments before.

yowl (approved):

PR #2229, opened 2023-03-18
[NativeAOT-LLVM] Implement basic debug info support for types and variables

SingleAccretion:
This implements basic debug info emission for variables and complex types. We can now inspect parameters, locals and instance fields. This is sufficient for low-level runtime debugging. See also the newly added (unfortunately, manual) test for supported scenarios.

Not implemented are static fields, debugging in funclets, dynamic type discovery, pretty printing of arrays/strings, calling methods from the debugger, etc. Additionally, compiler-generated helpers such as "get static base" are stepped into and show raw WASM, so step-in experience is subpar.

That said, this is still approximately a lot better than not being able to inspect anything at all \:).

The implementation is somewhat complex, due to the desire to reuse code intended for the non-LLVM object writer (`UserDefinedTypeDescriptor`). At the end of the day, though, it is just converting EE data structures into Jit/EE data structures into the DWARF-like metadata LLVM uses.

Notably, this disables SSA for debug code as the toolchain cannot handle `dbg.value`-based variables, everything has be spilled to be accessible in the debugger.

### DIFF
--- a/src/coreclr/jit/llvmdebuginfo.cpp
+++ b/src/coreclr/jit/llvmdebuginfo.cpp
@@ -14,8 +14,10 @@ using llvm::DINode;
 using llvm::DINodeArray;
 using llvm::DIFile;
 using llvm::DIType;
+using llvm::DIDerivedType;
 using llvm::DISubroutineType;
 using llvm::DILocation;
+using llvm::DIExpression;
 
 enum CorInfoLlvmDebugTypeKind
 {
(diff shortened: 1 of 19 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
This is great - way better than anything we've had before.

A random screenshot:

image

src/coreclr/jit/llvmdebuginfo.cpp
```diff
@@ -0,0 +1,602 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// ================================================================================================================
+// |                            DWARD debug info generation for the LLVM backend                                  |
```
jkotas:
```suggestion
// |                            DWARF debug info generation for the LLVM backend                                  |
```

PR #2230, opened 2023-03-19
[NativeAOT-LLVM] Fix a bug in the delegate ctor helper

SingleAccretion:
We were not properly accounting for cases where the generic dictionary pointer needed to be loaded from the method table.


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas (approved):
LGTM. Thank you!

PR #2231, opened 2023-03-19
[NativeAOT-LLVM] Delete the `IsRhpUnmanagedIndirection` workaround

SingleAccretion:
Instead, handle the CC difference at the point of use by routing the indirect call through a runtime import.

This fixes the dynamic generics test modulo some things which rely on threading and finalization (which is why I am not enabling it here).


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/RuntimeImports.cs
```diff
@@ -972,6 +972,13 @@ public GCFrameRegistration(void* allocation, uint elemCount, bool areByRefs = tr
         internal static unsafe partial void RhCpuIdEx(int* cpuInfo, int functionId, int subFunctionId);
 #endif
 
+#if TARGET_WASM
+        [Intrinsic]
```
SingleAccretion, line 976:
```suggestion
```

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -83,7 +83,7 @@ public static int getSignatureForMethodSymbol(IntPtr thisHandle, IntPtr symbolHa
             if (node is IMethodNode { Offset: 0, Method: MethodDesc method })
             {
                 _this.Get_CORINFO_SIG_INFO(method, pSig, scope: null);
-                if (method.IsUnmanagedCallersOnly)
+                if (method.IsUnmanagedCallersOnly || node is RuntimeImportMethodNode)
```
jkotas: `CORINFO_CALLCONV_UNMANAGED` implies GC transition. Are we compensating for it somewhere?

SingleAccretion: > Are we compensating for it somewhere?

There is no need as the API is only used to retrieve the signature from a function pointer representative to construct the LLVM equivalent. Making it work for other cases would require adding a way to describe the "FCall calling convention", i. e. native ABI + no marshaling, no transition and no poll.

(Additionally, whether the transition will be needed is influenced by SPGCT and that is callsite specific.)

yowl (approved):

Issue #2214, opened 2023-02-25
[NativeAOT-LLVM] Managed ABI


SingleAccretion:
As we near the completion of the RyuJit transition, we should reflect on some decisions inherited from the IL backend.

One of those decisions is the managed (or "shadow") ABI.

Recall that in this ABI, arguments are passed in the following manner:
1) Types that do not contain GC references (`unmanaged` in C# parlance) are passed as LLVM arguments.
2) GC primitives as well as structs that have at least one GC field are passed on the shadow stack by value à la x86 or Unix x64.
3) Types that do not contain GC references are returned the same way as in the native calling convention. For WASM, this means that any non-trivial aggregates are returned by reference, using a hidden return buffer parameter.
4) Types that do contain GC references, which, importantly, includes GC primitives, are also returned by reference, with an explicit pointer parameter. Notable: this is currently not really implemented correctly as the return buffer logically resides below the caller's shadow frame, thus any intervening calls and/or safe points risk trashing it.
5) The shadow stack is aligned to the target's pointer size, but the alignment of arguments can be greater than that. The result is that some arguments can still end up misaligned.
6) The shadow stack grows up and there are two pointers prominent on the frame: the passed-in one, used to access local state, and the one passed to callees (currently recomputed at each call).

I propose a number of changes to this scheme. These represent "obvious" improvements that are easily justifiable based on experience with other ABIs. Further adjustments would require more targeted code size and speed benchmarks. So:
1) Return GC primitives and trivial GC aggregates by value. Such SDSU GC values are unlikely to cross safe points, so it does not seem pointful to force them to the stack.
    - #2236
2) Figure out what to do with the stack alignment - either align `long`/`double` to 4 bytes only (on 32 bit), or increase shadow stack alignment to 8. Increasing shadow stack alignment is not great because in general we want as little "wasted" shadow space as possible, to make GC scans faster.
    - #2315 - dynamic double alignment.
3) Move to the traditional "stacks grow down" model. This would allow us to use only one frame pointer throughout the whole frame in most cases, both for accessing the locals, and calling methods (the exception to this would be handlers, but most handler code is cold), without the size penalties of using negative offsets.

yowl:
For my education, how does 3 - growing the stack down, allow us to get rid of a frame pointer?

SingleAccretion:
[...]

PR #2237, opened 2023-03-23
[NativeAOT-LLVM] Return complex structs by reference

### LINKED ISSUE #2190: [NativeAOT-LLVM] PI-related issues

SingleAccretion:
Using the return buffer pointer, like other targets. The pointer will always be unmanaged, with us pinning it if necessary.

Contributes to #2214.
Fixes #2190.

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -626,7 +626,7 @@ var_types Compiler::getArgTypeForStruct(CORINFO_CLASS_HANDLE clsHnd,
                                         unsigned             structSize)
 {
 #ifdef TARGET_WASM
-    return m_llvm->GetArgTypeForStructWasm(clsHnd, wbPassStruct, structSize);
+    return m_llvm->GetArgTypeForStructWasm(clsHnd, wbPassStruct);
 #else // !TARGET_WASM
 
     var_types         useType         = TYP_UNKNOWN;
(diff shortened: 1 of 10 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good and a nice cleanup

PR #2254, opened 2023-04-15
Remove redundant quote in browser platform condition

hez2010:
Fix build.


yowl (approved):
Thanks, didn't test the actual publishing, sorry.

hez2010:
Note that in the latest version we have to specify `IlcHostPackagePath` manually, and `NativeDebugSymbols` seems to be overridden somewhere so we have to pass it via command line directly.
And when use it with .NET 8 sdk, we also need to pass `/p:PublishTrimmed=true`, it's no longer implicitly included.

PR #2255, opened 2023-04-15
NativeAOT-LLVM: Merge to apr 15 

yowl:
This gets us nearly up to date.   Bit of a long one, notable changes:

- Add `m_RIP` to `PInvokeTransitionFrame` for `USE_PORTABLE_HELPERS`
- Move GetHijackedReturnAddress to `FEATURE_HIJACK` - upstream issue?
- Add `CORINFO_HELP_NEW_MDARR_RARE` with same sig as `CORINFO_HELP_NEW_MDARR`
- Replace `gtGetStructHandle` with `node->GetLayout(_compiler)->GetClassHandle();`
- Reolace `GT_OBJ` with `GT_BLK`
- Delete code handling `TYP_BLK`
- Replace `GT_LCL_VAR_ADDR` and `GT_LCL_FLD_ADDR` with `GT_LCL_ADDR`
- Add `ret void` for `BBJ_EHFAULTRET`
- Exclude tests `GenerateUnmanagedEntryPoints` and `CustomMain` although it might be possible to make at least `CustomMain` pass
- Remove linux builds, are failing to `build tests` with lots of
```
ld.lld : error : unable to find library -lz [/__w/1/s/src/tests/nativeaot/SmokeTests/FrameworkStrings/Baseline.csproj] [/__w/1/s/src/tests/build.proj]
```
I tried in WSL/Ubuntu and they are fine, any suggestions appreciated.

- Turn off `INCREMENTAL` for the linker.  It was erroring for me - not compatible with `/DEBUGTYPE`
- Exclude `PHASE_EXPAND_STATIC_INIT` Some snippets about that follow:
```
@__NonGCStaticBase_S_P_CoreLib_System_Environment__DATA = global [6 x ptr] [ptr @S_P_CoreLib_System_Environment___cctor, ptr null, ptr null, ptr null, ptr null, ptr null]
```
For reference this is the LLVM that was produced and caused a `memory access out of bounds`.  Haven't looked to closely except to note that there are a lot of nulls in the symbol.
```llvm
define void @S_P_CoreLib_System_Environment___cctor(ptr %0) !dbg !20386 {
BB00:
  br label %BB04

BB04:                                             ; preds = %BB00
  store ptr null, ptr %0, align 4
  br label %BB01, !dbg !20387

BB01:                                             ; preds = %BB04
  %1 = load i32, ptr inttoptr (i32 add (i32 ptrtoint (ptr @__NonGCStaticBase_S_P_CoreLib_System_Environment to i32), i32 -4) to ptr), align 4, !dbg !20387
  %2 = icmp ne i32 %1, 0, !dbg !20387
  br i1 %2, label %BB02, label %BB03.1, !dbg !20387

BB02:                                             ; preds = %BB01
  %3 = getelementptr i8, ptr %0, i32 4, !dbg !20387
  %4 = call ptr @__GetNonGCStaticBase_S_P_CoreLib_System_Environment(ptr %3), !dbg !20387
  br label %BB03.1, !dbg !20387

BB03.1:                                           ; preds = %BB02, %BB01
  store ptr inttoptr (i32 add (i32 ptrtoint (ptr @__NonGCStaticBase_S_P_CoreLib_System_Environment to i32), i32 4) to ptr), ptr %0, align 4, !dbg !20387
  %5 = getelementptr i8, ptr %0, i32 4, !dbg !20387
  call void @RhpSetShadowStackTop(ptr %5), !dbg !20387
  %6 = call i32 @RhGetProcessCpuCount(), !dbg !20387
  %7 = load ptr, ptr %0, align 4, !dbg !20387
  %8 = icmp ult ptr %7, inttoptr (i32 1024 to ptr), !dbg !20387
  br i1 %8, label %BB05, label %BB03.2, !dbg !20387

BB03.2:                                           ; preds = %BB03.1
  store i32 %6, ptr %7, align 4, !dbg !20387
  ret void, !dbg !20387

BB05:                                             ; preds = %BB03.1
  %9 = getelementptr i8, ptr %0, i32 4, !dbg !20387
  call void @S_P_CoreLib_Internal_Runtime_CompilerHelpers_ThrowHelpers__ThrowNullReferenceException(ptr %9), !dbg !20387
  unreachable, !dbg !20387
}
```

cc @dotnet/nativeaot-llvm


yowl:
@SingleAccretion A question about https://github.com/dotnet/runtime/blob/main/src/coreclr/jit/flowgraph.cpp#L2683  and the call to `lvaSetStruct`.  In 

https://github.com/dotnet/runtimelab/blob/cd6f3ffcabe63f7a9ce2cb9f47e7dead2ad538c8/src/coreclr/jit/llvmlower.cpp#L439-L456

We have this code where we decide the offsets based on, currently `varDsc->TypeGet() == TYP_BLK`, however that is no longer the case as we have `TYP_STRUCT`.  Should I be doing the same `padOffset`, etc, based on a non null  `classHandle`?

Thanks

SingleAccretion:
Yes - `TYP_BLK` locals are now struct locals with a block layout.

yowl:
Have a problem with this dump, 
[EGitHubruntimelabartifactsbincoreclrwindows.x64.Debugilcnet8.0ilc.exe.txt](https://github.com/dotnet/runtimelab/files/11256289/EGitHubruntimelabartifactsbincoreclrwindows.x64.Debugilcnet8.0ilc.exe.txt)

Funclet with index 1 is not generated as the starting block `BB14` has no immediate dominator in `bbIDom`.  Then when `BB21` is up for code gen it fails to find the LLVM function.

Edit: resovled.

src/coreclr/jit/compiler.cpp
```diff
@@ -5087,14 +5075,18 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
     optLoopsRequirePreHeaders = false;
 
 #ifdef DEBUG
+#ifndef TARGET_WASM
     DoPhase(this, PHASE_STRESS_SPLIT_TREE, &Compiler::StressSplitTree);
+#endif // !TARGET_WASM
```
SingleAccretion: Why `#ifdef` this out?

src/coreclr/jit/compiler.cpp
```diff
@@ -5087,14 +5075,18 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
...
 #endif
 
     // Expand runtime lookups (an optimization but we'd better run it in tier0 too)
     DoPhase(this, PHASE_EXPAND_RTLOOKUPS, &Compiler::fgExpandRuntimeLookups);
 
+#ifndef TARGET_WASM
     // Partially inline static initializations
     DoPhase(this, PHASE_EXPAND_STATIC_INIT, &Compiler::fgExpandStaticInit);
+#endif // !TARGET_WASM
```
SingleAccretion: We should file an issue to re-enable this.

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -325,10 +325,17 @@ void Llvm::generateBlocks()
...
-        // Walk all the exceptional code blocks and generate them, since they don't appear in the normal flow graph.
+        // Walk all the exceptional code blocks and generate them since they don't appear in the normal flow graph.
         for (Compiler::AddCodeDsc* add = _compiler->fgGetAdditionalCodeDescriptors(); add != nullptr; add = add->acdNext)
         {
-            generateBlock(add->acdDstBlk);
+            // if the LLVM function was not created due to the first block not being reachable
+            // or the exceptional code block is part of a try that was not reachable,
+            // then don't generate the exceptional code block
+            if (m_functions[getLlvmFunctionIndexForBlock(add->acdDstBlk)].LlvmFunction != nullptr &&
+                !(add->acdDstBlk->hasTryIndex() && m_EHDispatchLlvmBlocks[add->acdDstBlk->getTryIndex()] == nullptr))
```
SingleAccretion: Is it a correctness problem if we remove the second condition (pertaining to unreachable `try`s)?

It would be best to not leak the EH details too much.

src/coreclr/jit/llvmlower.cpp
```diff
@@ -61,7 +61,7 @@ void Llvm::lowerSpillTempsLiveAcrossSafePoints()
                 break;
             case TYP_STRUCT:
                 // This case should be **very** rare if at all possible. Just use a new local.
-                structHandle = _compiler->gtGetStructHandle(node);
+                structHandle = node->GetLayout(_compiler)->GetClassHandle();
```
SingleAccretion, line 65:
```suggestion
                // TODO-LLVM: switch this logic to use layouts instead of handles.
                structHandle = node->GetLayout(_compiler)->GetClassHandle();
```

src/coreclr/jit/llvmlower.cpp
```diff
@@ -437,7 +434,7 @@ void Llvm::assignShadowStackOffsets(std::vector<LclVarDsc*>& shadowStackLocals,
 
     unsigned offset = 0;
     auto assignOffset = [this, &offset](LclVarDsc* varDsc) {
-        if (varDsc->TypeGet() == TYP_BLK)
+        if (varDsc->TypeGet() == TYP_STRUCT && varDsc->GetLayout()->IsBlockLayout())
```
```suggestion
        if ((varDsc->TypeGet() == TYP_STRUCT) && varDsc->GetLayout()->IsBlockLayout())
```

src/coreclr/jit/llvmlower.cpp
```diff
@@ -448,12 +445,13 @@ void Llvm::assignShadowStackOffsets(std::vector<LclVarDsc*>& shadowStackLocals,
...
             CorInfoType corInfoType = toCorInfoType(varDsc->TypeGet());
-            CORINFO_CLASS_HANDLE classHandle = varTypeIsStruct(varDsc) ? varDsc->GetStructHnd() : NO_CLASS_HANDLE;
+            CORINFO_CLASS_HANDLE classHandle =
+                varTypeIsStruct(varDsc) ? varDsc->GetLayout()->GetClassHandle() : NO_CLASS_HANDLE;
 
             offset = padOffset(corInfoType, classHandle, offset);
             varDsc->SetStackOffset(offset);
             offset = padNextOffset(corInfoType, classHandle, offset);
-        }
+        }
```
```suggestion
        }
```

src/coreclr/jit/llvmlower.cpp
```diff
@@ -507,12 +505,12 @@ void Llvm::initializeLocalInProlog(unsigned lclNum, GenTree* value)
     LclVarDsc* varDsc = _compiler->lvaGetDesc(lclNum);
     JITDUMP("Adding initialization for V%02u, %s:\n", lclNum, varDsc->lvReason);
 
-    // TYP_BLK locals have to be handled specially as they can only be referenced indirectly.
+    // Block layout struct locals have to be handled specially as they can only be referenced indirectly.
```
```suggestion
```
They no longer need to be, we can simplify this code to always create `STORE_LCL_VAR` now.

src/coreclr/jit/llvmlower.cpp
```diff
@@ -930,7 +922,7 @@ void Llvm::lowerStoreBlk(GenTreeBlk* storeBlkNode)
         }
         else
         {
-            CORINFO_CLASS_HANDLE srcHandle = _compiler->gtGetStructHandleIfPresent(src);
+            CORINFO_CLASS_HANDLE srcHandle = src->GetLayout(_compiler)->GetClassHandle();
```
SingleAccretion, line 915:
```suggestion
            // TODO-LLVM: switch this logic to use layouts.
            CORINFO_CLASS_HANDLE srcHandle = src->GetLayout(_compiler)->GetClassHandle();
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
```diff
@@ -91,22 +56,13 @@
...
-      <IlcHostPackagePath Condition="'%(PackageDefinitions.Name)' == '$(IlcHostPackageName)'">%(PackageDefinitions.ResolvedPath)</IlcHostPackagePath>
-=======
   <Target Name="ImportRuntimeIlcPackageTarget" Condition="'$(BuildingFrameworkLibrary)' != 'true' and '$(PublishAot)' == 'true' and $(IlcCalledViaPackage) == 'true'" DependsOnTargets="$(ImportRuntimeIlcPackageTargetDependsOn)" BeforeTargets="Publish">
     <Error Condition="'@(ResolvedILCompilerPack)' == ''" Text="The ResolvedILCompilerPack ItemGroup is required for target ImportRuntimeIlcPackageTarget" />
 
     <PropertyGroup>
       <IlcHostPackagePath Condition="'@(ResolvedILCompilerPack)' == '$(_hostPackageName)'">@(ResolvedILCompilerPack->'%(PackageDirectory)')</IlcHostPackagePath>
       <RuntimePackagePath Condition="'@(ResolvedTargetILCompilerPack)' == '$(_targetPackageName)'">@(ResolvedTargetILCompilerPack->'%(PackageDirectory)')</RuntimePackagePath>
       <RuntimePackagePath Condition="'@(ResolvedTargetILCompilerPack)' == ''">@(ResolvedILCompilerPack->'%(PackageDirectory)')</RuntimePackagePath>
```
SingleAccretion: I think this will break publishing as we rely on the `PackageDefinitions`-based mechanism.

I would recommend explicitly overriding this target below s.t. we don't have to deal with the conflicts every time:
```
<Target Name="ImportRuntimeIlcPackageTarget" ... /> # Upstream target

<PropertyGroup>
  <!-- EmitLegacyAssetsFileItems tells the SDK to generate PackageDefinitions -->
  <EmitLegacyAssetsFileItems>true</EmitLegacyAssetsFileItems>
</PropertyGroup>

<Target Name="ImportRuntimeIlcPackageTarget" ... /> # Same target name, NativeAOT-LLVM version
```
Would also recommend to do the same with `GenerateILCompilerExplicitPackageReferenceWarning` (which would just be an empty target).

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -118,7 +118,7 @@ public static void EmitObject(string objectFilePath, IEnumerable<DependencyNode>
 
                     ObjectData nodeContents = node.GetData(factory);
 
-                    dumper?.DumpObjectNode(factory.NameMangler, node, nodeContents);
+                    dumper?.DumpObjectNode(factory, node, nodeContents);
 #if DEBUG
                     foreach (ISymbolNode definedSymbol in nodeContents.DefinedSymbols)
                     {
```
SingleAccretion, line 124: I would expect that we need to adjust our static cctor trigger code after [#83937](https://github.com/dotnet/runtime/pull/83937).

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -884,7 +884,11 @@ private LLVMValueRef OutputCodeForGetGenericDictionary(LLVMBuilderRef builder, L
             ReadyToRunGenericHelperNode node, GenericLookupResult lookup, LLVMValueRef dictionary)
         {
             // Find the generic dictionary slot
-            int dictionarySlot = factory.GenericDictionaryLayout(node.DictionaryOwner).GetSlotForEntry(lookup);
+            if (!factory.GenericDictionaryLayout(node.DictionaryOwner)
+                    .TryGetSlotForEntry(lookup, out int dictionarySlot))
+            {
+                throw new Exception($"Failed to find slot entry in generic dictionary layout ${node.DictionaryOwner.GetDisplayName()}, entry {lookup}");
+            }
```
SingleAccretion: If I am reading https://github.com/dotnet/runtime/pull/82591 right, we're supposed to return a `null` pointer in this case.

src/libraries/externals.csproj
```diff
@@ -72,11 +72,7 @@
   <Target Name="OverrideRuntimeCoreCLR"
           DependsOnTargets="ResolveRuntimeFilesFromLocalBuild"
           AfterTargets="AfterResolveReferences"
-<<<<<<< HEAD
-          Condition="'$(RuntimeFlavor)' != 'Mono' and '$(TestNativeAot)' != 'true' and '$(ClrNativeAot)' != 'true'">
-=======
           Condition="'$(RuntimeFlavor)' != 'Mono' and '$(UseNativeAotCoreLib)' != 'true'">
->>>>>>> runtime/main
```
SingleAccretion: I think we can delete `ClrNativeAot` now.

src/tests/nativeaot/SmokeTests/UnitTests/Delegates.cs
```diff
@@ -51,12 +51,10 @@ public static int Run()
             result = Fail;
         }
 
-<<<<<<< HEAD
 #if !CODEGEN_WASM // TODO-LLVM: fails trying to create a thread. We need correct TryEnsureSufficientExecutionStack.
-=======
->>>>>>> runtime/main
         TestLinqExpressions.Run();
         TestDefaultInterfaceMethods.Run();
```
SingleAccretion: Does `TestDefaultInterfaceMethods` not work? Bit surprising.

src/coreclr/jit/compiler.cpp
yowl: Done,

src/coreclr/jit/llvmcodegen.cpp
yowl: Without this check it fails at 
```
 	clrjit_browser_wasm32_x64.dll!llvm::BasicBlock::getParent() Line 105	C++
	clrjit_browser_wasm32_x64.dll!Llvm::emitCallOrInvoke(llvm::FunctionCallee callee, llvm::ArrayRef<llvm::Value *> args) Line 2526	C++
 	clrjit_browser_wasm32_x64.dll!Llvm::buildCall(GenTreeCall * call) Line 1871	C++
 	clrjit_browser_wasm32_x64.dll!Llvm::visitNode(GenTree * node) Line 1088	C++
 	clrjit_browser_wasm32_x64.dll!Llvm::generateBlock(BasicBlock * block) Line 361	C++
 	clrjit_browser_wasm32_x64.dll!Llvm::generateBlocks() Line 339	C++
 	clrjit_browser_wasm32_x64.dll!Llvm::Compile() Line 29	C++
 	clrjit_browser_wasm32_x64.dll!Compiler::compCompile::__l2::<lambda_5>::operator()() Line 5182	C++
 	clrjit_browser_wasm32_x64.dll!ActionPhase<`Compiler::compCompile'::`2'::<lambda_5>>::DoPhase() Line 65	C++
```
because 
```cpp
catchLlvmBlock = m_EHDispatchLlvmBlocks[getCurrentProtectedRegionIndex()];
```
returns null.  However we can just put a `if (catchLlvmBlock != nullptr)` around the code that sets `catchLlvmBlock` to `nullptr` and it is fine with that.  I'll put that in the next commit

src/coreclr/jit/compiler.cpp
yowl: Forgot to come back to this.  it was just an assert that used `codegen` that was a problem, have isolated that with `#if` instead in the next commit

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: I don't think so as this is the code to create the dictionary lookup (e.g from a type at runtime), its not called by the jit during compilation, @MichalStrehovsky what do you think?

src/tests/nativeaot/SmokeTests/UnitTests/Delegates.cs
yowl: it is working, reinstated.

src/libraries/externals.csproj
yowl: Removed, thanks.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: yes, that is odd.  Whether I change the test for initialized to 1 or 0 seems to make no difference.  Have changed it to 0 to match x64 and others.

Edit.  I also tried to create a test that would fail, but couldn't.  Left the test in anywaty

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
MichalStrehovsky: If I understand this correctly, this code is the equivalent of the assembly we emit in `ReadyToRunGenericHelperNode` - I would expect this to generate null lookup instead of throwing.

It might only be reachable when running with `--scan` but not `-O`. It's certainly a corner case. Adding a comment to the exception with a link to dotnet/runtime#82591 might be enough.

yowl: Thanks, I stand corrected.  Now returns null.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: I think this is ok.  It doesn't seem to make any difference because `CheckStaticClassConstruction` has its own check for 0 so will return immediately anyway.  This is just an earlier check to save that call, but if it is wrong the cctor is not run again anyway.  There is perhaps a small edge problem with using 1 as a marker that another thread is running the cctor, in that 1 is a valid method table index for wasm (as is 0).  Its possible that those indicies are valid `cctorMethodAddress` although in practice I think many other functions are going to be created before the compiler gets to a cctor.

src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
yowl: Done, thanks

src/coreclr/jit/llvmlower.cpp
yowl: Indeed, removed, thanks!

src/coreclr/jit/llvmcodegen.cpp
yowl: Now removed the second condition and added a guard to the nulling of `catchLlvmBlock` in `emitCallOrInvoke`

yowl:
This looks ok for a second review I think. The last commit is just the removal of a comment that was no longer correct, so this should go green.  Thanks

eng/native/configurecompiler.cmake
```diff
@@ -85,6 +85,8 @@ if (MSVC)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /sourcelink:${CLR_SOURCELINK_FILE_PATH}")
   endif(EXISTS ${CLR_SOURCELINK_FILE_PATH})
 
+  add_linker_flag(/INCREMENTAL:NO) # Not compatible with /DEBUGTYPE
```
SingleAccretion: Does this means `./build clr.jit` will always rebuild all the files? It is very worrying of so.

Why doesn't upstream hit incompatibility error?

src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
```diff
@@ -73,44 +43,54 @@
...
+    <PropertyGroup>
+      <HostOSIdentifier Condition="$([MSBuild]::IsOSPlatform('OSX'))">osx</HostOSIdentifier>
+      <HostOSIdentifier Condition="$([MSBuild]::IsOSPlatform('WINDOWS'))">win</HostOSIdentifier>
+      <HostOSIdentifier Condition="'$(HostOSIdentifier)' == ''">linux</HostOSIdentifier>
+
+      <!-- Default to host that matches OS architecture on Windows. It allows compilation using msbuild.exe that is x86 by default -->
+      <IlcHostArch Condition="'$(OS)' == 'Windows_NT' and '$(IlcHostArch)' == ''">$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().ToLowerInvariant)</IlcHostArch>
+      <!-- Default to host that matches SDK architecture on non-Windows -->
+      <IlcHostArch Condition="'$(IlcHostArch)' == ''">$([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant)</IlcHostArch>
+      <IlcHostPackageName>runtime.$(HostOSIdentifier)-$(IlcHostArch).Microsoft.DotNet.ILCompiler.LLVM</IlcHostPackageName>
```
SingleAccretion: Hmm, is there a need to duplicate `HostOSIdentifier` calculations instead of using `_hostOS`? I don't think we need to go quite that far in duplicating if not.

src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
```diff
@@ -73,44 +43,54 @@
...
+
+      <IlcHostPackagePath Condition="'%(PackageDefinitions.Name)' == '$(IlcHostPackageName)'">%(PackageDefinitions.ResolvedPath)</IlcHostPackagePath>
+      <RuntimePackagePath Condition="'%(PackageDefinitions.Name)' == 'runtime.browser-wasm.Microsoft.DotNet.ILCompiler.LLVM'">%(PackageDefinitions.ResolvedPath)</RuntimePackagePath>
```
SingleAccretion:
```suggestion
      <RuntimePackagePath Condition="'%(PackageDefinitions.Name)' == '$(_targetPackageName)'">%(PackageDefinitions.ResolvedPath)</RuntimePackagePath>
```

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -2514,17 +2518,21 @@ llvm::CallBase* Llvm::emitCallOrInvoke(llvm::FunctionCallee callee, ArrayRef<Val
...
-        // Protected region index that is set in the emit context refers to the "logical" enclosing
-        // protected region, i. e. the one before funclet creation. But we do not need to (in fact,
-        // cannot) emit an invoke targeting block inside a different LLVM function.
-        if (catchLlvmBlock->getParent() != getCurrentLlvmFunction())
-        {
-            catchLlvmBlock = nullptr;
-        }
-        // No need to invoke no-throw functions.
-        else if (llvm::isa<Function>(callee.getCallee()) && llvm::cast<Function>(callee.getCallee())->doesNotThrow())
+        if (catchLlvmBlock != nullptr)
```
SingleAccretion: Hmm, the state where we have a protected region without a dispatch block is arguably worse than checking that the protected region was unreachable...

I think I've come to a solution that lets us deal with dead throw helper blocks robustly:
1) In `emitJumpToThrowHelper` mark them: `throwBlock->bbFlags |= BBF_MARKED;`.
2) In the loop over additional code descriptors, check for the mark.

The we don't need to involve EH here at all.

src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
yowl: Thanks

eng/native/configurecompiler.cmake
yowl: I think this only happens with `-msbuild`.  I think this is the same: https://github.com/dotnet/runtime/issues/64082

src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
yowl: Indeed not, removed.

src/coreclr/jit/llvmcodegen.cpp
yowl: I'm not sure I've removed everything that can be, but have added the flag and checked it before `generateBlock`

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -325,10 +325,16 @@ void Llvm::generateBlocks()
...
 
-        // Walk all the exceptional code blocks and generate them, since they don't appear in the normal flow graph.
+        // Walk all the exceptional code blocks and generate them since they don't appear in the normal flow graph.
         for (Compiler::AddCodeDsc* add = _compiler->fgGetAdditionalCodeDescriptors(); add != nullptr; add = add->acdNext)
         {
-            generateBlock(add->acdDstBlk);
+            // if the LLVM function was not created due to the first block not being reachable
+            // then don't generate the exceptional code block
+            if ((add->acdDstBlk->bbFlags & BBF_MARKED) > 0 &&
+                m_functions[getLlvmFunctionIndexForBlock(add->acdDstBlk)].LlvmFunction != nullptr)
```
```suggestion
            if ((add->acdDstBlk->bbFlags & BBF_MARKED) != 0)
```

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -2522,7 +2528,8 @@ llvm::CallBase* Llvm::emitCallOrInvoke(llvm::FunctionCallee callee, ArrayRef<Val
             catchLlvmBlock = nullptr;
         }
         // No need to invoke no-throw functions.
-        else if (llvm::isa<Function>(callee.getCallee()) && llvm::cast<Function>(callee.getCallee())->doesNotThrow())
+        else if (llvm::isa<Function>(callee.getCallee()) &&
+                 llvm::cast<Function>(callee.getCallee())->doesNotThrow())
```
```suggestion
        else if (llvm::isa<Function>(callee.getCallee()) && llvm::cast<Function>(callee.getCallee())->doesNotThrow())
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
```diff
@@ -73,44 +43,50 @@
...
+    <PropertyGroup>
+      <!-- Default to host that matches OS architecture on Windows. It allows compilation using msbuild.exe that is x86 by default -->
+      <IlcHostArch Condition="'$(OS)' == 'Windows_NT' and '$(IlcHostArch)' == ''">$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().ToLowerInvariant)</IlcHostArch>
+      <!-- Default to host that matches SDK architecture on non-Windows -->
+      <IlcHostArch Condition="'$(IlcHostArch)' == ''">$([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant)</IlcHostArch>
+      <IlcHostPackageName>runtime.$(_hostOS)-$(IlcHostArch).Microsoft.DotNet.ILCompiler.LLVM</IlcHostPackageName>
+
+      <IlcHostPackagePath Condition="'%(PackageDefinitions.Name)' == '$(IlcHostPackageName)'">%(PackageDefinitions.ResolvedPath)</IlcHostPackagePath>
+      <RuntimePackagePath Condition="'%(PackageDefinitions.Name)' == '$(_targetPackageName)'">%(PackageDefinitions.ResolvedPath)</RuntimePackagePath>
+    </PropertyGroup>
```
SingleAccretion:
```suggestion
      <IlcHostPackagePath Condition="'%(PackageDefinitions.Name)' == '$(_hostPackageName)'">%(PackageDefinitions.ResolvedPath)</IlcHostPackagePath>
      <RuntimePackagePath Condition="'%(PackageDefinitions.Name)' == '$(_targetPackageName)'">%(PackageDefinitions.ResolvedPath)</RuntimePackagePath>
    </PropertyGroup>
```
(If there is no reason for it not to work)

yowl: Yes, that looks fine, thanks.

yowl:
LLVM install failed

SingleAccretion (approved):
Thank you!

jkotas:
Thank you all!

Issue #2235, opened 2023-03-23
[NativeAOT LLVM] Using together with .NET 8 SDK


kant2002:
Troubleshooting of recent issue uncover fact that .NET 8 SDK is not supported.
So my current proposal is advocate following way of consuming NativeAOT-LLVM
```xml
<ItemGroup Condition="$(RuntimeIdentifier) == 'browser-wasm'">
  <!-- Remove existing ILCompilers -->
  <KnownILCompilerPack Remove="Microsoft.DotNet.ILCompiler" />
  <!-- Add NativeAOT LLVM. -->
  <KnownILCompilerPack Include="Microsoft.DotNet.ILCompiler.LLVM"
    TargetFramework="net7.0"
    ILCompilerPackNamePattern="runtime.**RID**.Microsoft.DotNet.ILCompiler.LLVM"
    ILCompilerPackVersion="8.0.0"
    ILCompilerRuntimeIdentifiers="browser-wasm;win-x64" />

     <!-- This line needed only for wildcard consumption -->
    <PackageReference Include="Microsoft.DotNet.ILCompiler.LLVM" Version="8.0.0-*" />
</ItemGroup>
```

Now a bit about how that's working.
Detection of supported ILC package happens here by matching TFM  with KnownILCompilerPack
https://github.com/dotnet/sdk/blob/d8b549b85299046654ba8e9bd0caff0eed507ae6/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs#L613-L618
and then added as implicit package references here
https://github.com/dotnet/sdk/blob/d8b549b85299046654ba8e9bd0caff0eed507ae6/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs#L690-L697

I initially [thought](https://github.com/dotnet/sdk/pull/31398) that first place can be extended to perform ILC package selection based on the TFM + RID, but second place insert additioal reference on ILCompiler which causing issues.

So this led me to following solution where I remove existing KnownILCompilerPack and create new one in suited for NativeAOT-LLVM.

If  https://github.com/dotnet/sdk/pull/31398 lands then removal of existing ILCompiler does not needed.

I did not dig deeper why we cannot use NativeAOT-LLVM without PublishAot, maybe that would be more viable solution with less divergence from main.

SingleAccretion:
Thank you for looking into this! I'll share my thinking on the problem space.

At a high level, this can be approached from 2 angles:
[...]

PR #2261, opened 2023-04-21
NativeAOT-LLVM: Remove ILLink target

yowl:
This PR removes the `dependson` for `PrepareForILLink` and the `_RunILLink` .

I think this will fix: #2235

### DIFF
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -19,13 +19,13 @@ from the project's root directory. New package sources must be added after the `
 
 Once you have added the package sources, add a reference to the ILCompiler package either by running
 ```bash
-> dotnet add package Microsoft.DotNet.ILCompiler -v 7.0.0-*
+> dotnet add package Microsoft.DotNet.ILCompiler -v 8.0.0-*
 ```
 
 or by adding the following element to the project file:
 ```xml
   <ItemGroup>
-    <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="7.0.0-*" />
+    <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="8.0.0-*" />
   </ItemGroup>
 ```
 
@@ -64,7 +64,17 @@ For WebAssembly, it is always a cross-architecture scenario as the compiler runs
 <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="8.0.0-*" />
 ```
 
-Then, the required package reference is
+Then, remove 
+```xml
+<NativeAot>true</NativeAot>
+```
+from any `PropertyGroup` tags if you have it. Instead, add
+```xml
+<PropertyGroup>
+  <PublishTrimmed>true</PublishTrimmed>
+</PropertyGroup>
+```
+The required package reference is
 ```xml
 <PackageReference Include="Microsoft.DotNet.ILCompiler.LLVM; runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM" Version="8.0.0-*" />
 ```
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -66,7 +66,7 @@ For WebAssembly, it is always a cross-architecture scenario as the compiler runs
 
 Then, remove 
 ```xml
-<NativeAot>true</NativeAot>
+<PublishAot>true</PublishAot>
 ```
 from any `PropertyGroup` tags if you have it. Instead, add
 ```xml
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -118,7 +118,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <IlcCompileDependsOn Condition="'$(BuildingFrameworkLibrary)' != 'true'">Compile;ComputeIlcCompileInputs</IlcCompileDependsOn>
     <IlcCompileDependsOn Condition="'$(IlcMultiModule)' == 'true' and '$(BuildingFrameworkLibrary)' != 'true'">$(IlcCompileDependsOn);BuildFrameworkLib</IlcCompileDependsOn>
     <IlcCompileDependsOn>$(IlcCompileDependsOn);SetupOSSpecificProps</IlcCompileDependsOn>
-    <IlcCompileDependsOn Condition="'$(NativeCodeGen)' != 'llvm'">$(IlcCompileDependsOn);PrepareForILLink</IlcCompileDependsOn>
+    <IlcCompileDependsOn>$(IlcCompileDependsOn);PrepareForILLink</IlcCompileDependsOn>
   </PropertyGroup>
 
   <ItemGroup Condition="$(IlcSystemModule) == ''">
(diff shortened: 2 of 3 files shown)


yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -118,7 +118,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <IlcCompileDependsOn Condition="'$(BuildingFrameworkLibrary)' != 'true'">Compile;ComputeIlcCompileInputs</IlcCompileDependsOn>
     <IlcCompileDependsOn Condition="'$(IlcMultiModule)' == 'true' and '$(BuildingFrameworkLibrary)' != 'true'">$(IlcCompileDependsOn);BuildFrameworkLib</IlcCompileDependsOn>
     <IlcCompileDependsOn>$(IlcCompileDependsOn);SetupOSSpecificProps</IlcCompileDependsOn>
-    <IlcCompileDependsOn>$(IlcCompileDependsOn);PrepareForILLink</IlcCompileDependsOn>
+    <IlcCompileDependsOn Condition="'$(NativeCodeGen)' != 'llvm'">$(IlcCompileDependsOn);PrepareForILLink</IlcCompileDependsOn>
```
SingleAccretion: What is the original reason for the dependence on `PrepareForILLink`?

yowl: I think its an extension point for developers to hook in their own steps before Ilc runs.  Not great to remove it in that case.  I'll have another go at this.

yowl:
How about this, I've just added an empty `PrepareForILLink` target so that users can add before and after targets?  I think this is all we get out of this task for Naot.  @MichalStrehovsky would know better what we are losing, if anything with an empty target.

yowl:
From what I read, PrepareForILLink was added as part of a commit to make NAOT similar in behaviour to PublishTrimmed.

yowl:
The latest commit adds a call to `_ComputeManagedAssemblyForILLink` to get similar behaviour, I think for the `PublishTrimmed` modes (partial?).

MichalStrehovsky:
PrepareForILLink set up everything for trimming (what assemblies should be trimmed, how, etc.):

https://github.com/dotnet/runtime/blob/2b530d4abdab90801b9c3434d62e511d9f72e3e7/src/tools/illink/src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets#L172-L183

Why are we not including the trimming targets on WASM?

yowl:
Upstream for (e.g win-x64) will call `PrepareForILLink` as a `DependsOn` for `WriteIlcRspFileForCompilation`  .  `PrepareForILLink` is defined in `Microsoft.NET.ILLink.targets` from the package `microsoft.net.illink.tasks` which does not get included for NativeAOT-LLVM (its done via the SDK, not sure?).

Maybe we should just make it a requirement (and document it), that the user has to add
```
   <PackageReference Include="microsoft.net.illink.tasks" Version="8.0.0-*" />
```
to their csproj, that seems to work?

MichalStrehovsky:
> to their csproj, that seems to work?

Yep, that sounds better. Best would be to figure out why it's not getting included (are we setting PublishTrimmed too late?) but both are better than the empty target.

yowl:
> Best would be to figure out why it's not getting included

Basically I start with the fact that `ILLinkTargetsPath` is not set for LLVM.  That leads me to the `console2.csproj.nuget.dgspec.json` file where for winx64 there is ILLink as a dependency
```json
"frameworks": {
  "net8.0": {
    "targetAlias": "net8.0",
    "dependencies": {
      "Microsoft.DotNet.ILCompiler": {
        "suppressParent": "All",
        "target": "Package",
        "version": "[8.0.0-preview.3.23174.8, )",
        "autoReferenced": true
      },
      "Microsoft.NET.ILLink.Tasks": {
        "suppressParent": "All",
        "target": "Package",
        "version": "[8.0.0-preview.3.23174.8, )",
        "autoReferenced": true
      }
    },
```

Where as for LLVM that section is
```json
  "frameworks": {
    "net8.0": {
      "targetAlias": "net8.0",
      "dependencies": {
        "Microsoft.DotNet.ILCompiler.LLVM": {
          "target": "Package",
          "version": "[8.0.0-*, )"
        },
        "runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM": {
          "target": "Package",
          "version": "[8.0.0-*, )"
        }
      },
```
how does `.nuget.dgspec.json` get generated ?

MichalStrehovsky:
That would be question that has answers somewhere in the dotnet/sdk repo. Does it get referenced if you add `<PublishTrimmed>true</PublishTrimmed>` to the project?

yowl:
> Does it get referenced if you add `<PublishTrimmed>true</PublishTrimmed>` to the project?

Yes it does, will look at dotnet/sdk to see what I can find.  Thanks.

yowl:
So without `PublishAot==true` (https://github.com/dotnet/runtimelab/issues/2235#issuecomment-1481415589) this logic leaves 
`_IsTrimmingEnabled==false`

```
  <PropertyGroup>
    <DefaultCopyToPublishDirectoryMetadata Condition="'$(DefaultCopyToPublishDirectoryMetadata)' == ''">true</DefaultCopyToPublishDirectoryMetadata>
    <_GetChildProjectCopyToPublishDirectoryItems Condition="'$(_GetChildProjectCopyToPublishDirectoryItems)' == ''">true</_GetChildProjectCopyToPublishDirectoryItems>
    <IsPublishable Condition="'$(IsPublishable)' == ''">true</IsPublishable>
    <!-- PublishAot depends on PublishTrimmed. This must be set early enough for the KnownILLinkPack to be restored. -->
    <PublishTrimmed Condition="'$(PublishTrimmed)' == '' And '$(PublishAot)' == 'true'">true</PublishTrimmed>
    <_IsTrimmingEnabled Condition="'$(_IsTrimmingEnabled)' == '' And ('$(PublishTrimmed)' == 'true' Or '$(IsTrimmable)' == 'true')">true</_IsTrimmingEnabled>
    <_IsTrimmingEnabled Condition="'$(_IsTrimmingEnabled)' == ''">false</_IsTrimmingEnabled>
    <DynamicCodeSupport Condition="'$(DynamicCodeSupport)' == '' And '$(PublishAot)' == 'true'">false</DynamicCodeSupport>
  </PropertyGroup>
```

I'll go with `<PublishTrimmed>true</PublishTrimmed>`   as that is implied to be true for NativeAOT .

yowl:
cc @dotnet/nativeaot-llvm 

I think this is good as we'll get it.  Have not tested .net 7 but I'm not sure we really care, I don't.

docs/using-nativeaot/compiling.md
```diff
@@ -64,7 +64,17 @@ For WebAssembly, it is always a cross-architecture scenario as the compiler runs
 <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="8.0.0-*" />
 ```
 
-Then, the required package reference is
+Then, remove 
+```xml
+<NativeAot>true</NativeAot>
```
jkotas: Should this be `<PublishAot>true</PublishAot>`?

yowl: Yes, thanks.

jkotas (approved):
Thanks

Issue #2259, opened 2023-04-20
NativeAOT-LLVM: export for NativeAOT_StaticInitialization automatically


yowl:
We should probably do this automatically if the user specifies `/p:NativeLib=Static` as you cant do much without it.

I.e. to the linker command add 
```
-s EXPORTED_FUNCTIONS=_NativeAOT_StaticInitialization
```
Follows #2258

SingleAccretion:
Probably just update this line to `$(NativeLib) != ''`:
https://github.com/dotnet/runtimelab/blob/cd6f3ffcabe63f7a9ce2cb9f47e7dead2ad538c8/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L437

yowl:
Yes, that works, thanks

RReverser:
Or maybe better to change `NativeLib` to `Shared` in that project?

yowl:
The change will allow either.  I guess shared is the correct option as we are not linking the wasm library with anything at link time.

RReverser:
> The change will allow either. I guess shared is the correct option as we are not linking the wasm library with anything at link time.

Yeah I remember that's what I was told in the original PR (or one of the related discussions, don't remember where exactly), so I explicitly enabled it only for `Shared`.

I think it could be nice to reserve `Static` for potentially actual static libraries built for Wasm.

yowl:
WIll close this as we updated the documentation to reflect that `EXPORTED_FUNCTIONS=_NativeAOT_StaticInitialization` was not actually necessary.

PR #2262, opened 2023-04-21
NativeAOT-LLVM: Export NativeAOT_StaticInitialization implicitly

yowl:
This PR changes the condition for exporting `NativeAOT_StaticInitialization` to include static libraries and adds a bit of documentation.

Closes: #2259

### DIFF
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -99,6 +99,16 @@ public static int Answer()
 ```bash
 > dotnet publish /p:NativeLib=Static /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false /p:EmccExtraArgs="-s EXPORTED_FUNCTIONS=_Answer -s EXPORTED_RUNTIME_METHODS=cwrap" --self-contained
 ```
+From Javascript, before calling your library the runtime must be initialised explicitly.  This is done with the following Javascript
+```js
+// Initialise the .Net runtime
+const corertInit = Module.cwrap('NativeAOT_StaticInitialization', 'number', []);
+corertInit();
+
+// Call your function
+const answer = Module.cwrap('Answer', 'number', []);
+console.log(answer());
+```
 
 #### WebAssembly module imports
 Functions in other WebAssembly modules can be imported and invoked using `DllImport` e.g.
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -97,7 +97,7 @@ public static int Answer()
 }
 ```
 ```bash
-> dotnet publish /p:NativeLib=Static /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false /p:EmccExtraArgs="-s EXPORTED_FUNCTIONS=_Answer -s EXPORTED_RUNTIME_METHODS=cwrap --post-js=invokeLibraryFunction.js" --self-contained
+> dotnet publish /p:NativeLib=Shared /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false /p:EmccExtraArgs="-s EXPORTED_FUNCTIONS=_Answer -s EXPORTED_RUNTIME_METHODS=cwrap --post-js=invokeLibraryFunction.js" --self-contained
 ```
 Where `invokeLibraryFunction.js` is a Javascript file with the callback to call `Answer`, e.g.
 ```js
--- a/docs/using-nativeaot/compiling.md
+++ b/docs/using-nativeaot/compiling.md
@@ -97,7 +97,7 @@ public static int Answer()
 }
 ```
 ```bash
-> dotnet publish /p:NativeLib=Shared /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false /p:EmccExtraArgs="-s EXPORTED_FUNCTIONS=_Answer -s EXPORTED_RUNTIME_METHODS=cwrap --post-js=invokeLibraryFunction.js" --self-contained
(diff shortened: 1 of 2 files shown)


yowl:
cc @dotnet/nativeaot-llvm

docs/using-nativeaot/compiling.md
```diff
@@ -99,6 +99,16 @@ public static int Answer()
 ```bash
 > dotnet publish /p:NativeLib=Static /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false /p:EmccExtraArgs="-s EXPORTED_FUNCTIONS=_Answer -s EXPORTED_RUNTIME_METHODS=cwrap" --self-contained
 ```
+From Javascript, before calling your library the runtime must be initialised explicitly.  This is done with the following Javascript
+```js
+// Initialise the .Net runtime
+const corertInit = Module.cwrap('NativeAOT_StaticInitialization', 'number', []);
+corertInit();
+
```
SingleAccretion: Is it actually needed to call it? The function only exists to trigger the global constructor.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -118,7 +118,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <IlcCompileDependsOn Condition="'$(BuildingFrameworkLibrary)' != 'true'">Compile;ComputeIlcCompileInputs</IlcCompileDependsOn>
     <IlcCompileDependsOn Condition="'$(IlcMultiModule)' == 'true' and '$(BuildingFrameworkLibrary)' != 'true'">$(IlcCompileDependsOn);BuildFrameworkLib</IlcCompileDependsOn>
     <IlcCompileDependsOn>$(IlcCompileDependsOn);SetupOSSpecificProps</IlcCompileDependsOn>
-    <IlcCompileDependsOn>$(IlcCompileDependsOn);PrepareForILLink</IlcCompileDependsOn>
+    <IlcCompileDependsOn Condition="'$(NativeCodeGen)' != 'llvm'">$(IlcCompileDependsOn);PrepareForILLink</IlcCompileDependsOn>
```
```suggestion
    <IlcCompileDependsOn>$(IlcCompileDependsOn);PrepareForILLink</IlcCompileDependsOn>
```
Presumably unintended.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -437,7 +437,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Include="-s ERROR_ON_UNDEFINED_SYMBOLS=0" />
       <CustomLinkerArg Condition="'$(NativeDebugSymbols)' != 'true'" Include="-O2 -flto" />
       <CustomLinkerArg Condition="'$(NativeDebugSymbols)' == 'true'" Include="-g3" />
-      <CustomLinkerArg Condition="$(NativeLib) == 'Shared'" Include="-Wl,--export,NativeAOT_StaticInitialization" />
+      <CustomLinkerArg Condition="$(NativeLib) != ''" Include="-Wl,--export,NativeAOT_StaticInitialization" />
```
SingleAccretion: I am a bit ambivalent about this...

As https://github.com/dotnet/runtimelab/issues/2259#issuecomment-1520093017 mentions, we didn't originally enable it for all types of libraries because for static libraries the NAOT instructions say you should add it to the link manually.

yowl: Ok, as we don't need it at all maybe we just remove it and leave it to the user to do 

```
-s EXPORTED_FUNCTIONS=_NativeAOT_StaticInitialization
```

yowl: I've left it as it was.

docs/using-nativeaot/compiling.md
yowl: No it's not.  I've removed that from the documentation.  I though it was at one point.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: Would be great to find the document on NAOT static libraries and add the linker parameter for exporting there.

yowl: https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/samples/NativeLibrary/README.md  This one?

SingleAccretion: I guess this section: https://github.com/dotnet/samples/blob/main/core/nativeaot/NativeLibrary/README.md#building-static-libraries, but we can't update it so it's fine as-is.

docs/using-nativeaot/compiling.md
```diff
@@ -97,7 +97,15 @@ public static int Answer()
...
-> dotnet publish /p:NativeLib=Static /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false /p:EmccExtraArgs="-s EXPORTED_FUNCTIONS=_Answer -s EXPORTED_RUNTIME_METHODS=cwrap" --self-contained
+> dotnet publish /p:NativeLib=Static /p:SelfContained=true -r browser-wasm -c Debug /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false /p:EmccExtraArgs="-s EXPORTED_FUNCTIONS=_Answer -s EXPORTED_RUNTIME_METHODS=cwrap --post-js=invokeLibraryFunction.js" --self-contained
+```
+Where `invokeLibraryFunction.js` is a Javascript file with the callback to call `Answer`, e.g.
+```js
+Module['onRuntimeInitialized'] = function() { 
+  // Call your function
+  const answer = Module.cwrap('Answer', 'number', []);
+  console.log(answer())
+};
```
SingleAccretion: Would it be better to use `/p:NativeLib=Shared` here?

Static libraries are meant to be linked with something else (so they would, e. g., contain unresolved relocations), here this is more like a classic dynamic library (without any dynamic data relocations).

SingleAccretion: Also: `/p:SelfContained=true` looks redundant with `--self-contained`.

yowl: Yes, done. Have also been wanting to do more work on `wit-bindgen` which is another kind of linking between modules were the interop code is generated from and IDL.

yowl: > Also: `/p:SelfContained=true` looks redundant with `--self-contained`.

Thanks, removed.

SingleAccretion (approved):
LGTM!

jkotas:
Thank you!

PR #2263, opened 2023-04-21
NativeAOT-LLVM: map Ilc opimisation settings to emcc

yowl:
This PR maps `$(Optimize)` to `-O3` and when `$(IlcOptimizationPreference) == 'Size'` to `-Oz`  .  Emcc has no specific option for speed.  Applies to both compile and link step.

`-flto` is moved from a `NativeDebugSymbols` check to `Optimize`

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -342,7 +342,6 @@ The .NET Foundation licenses this file to you under the MIT license.
     <PropertyGroup>
       <CompileEmccArgs>-c</CompileEmccArgs>
       <CompileEmccArgs>$(CompileEmccArgs) -s DISABLE_EXCEPTION_CATCHING=0</CompileEmccArgs>
-      <CompileEmccArgs Condition="'$(NativeDebugSymbols)' != 'true'">$(CompileEmccArgs) -O2</CompileEmccArgs>
       <CompileEmccArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(CompileEmccArgs) -g3</CompileEmccArgs>
       <!-- TODO-LLVM: remove when we update to Emscripten with https://github.com/emscripten-core/emscripten/pull/18443 -->
       <CompileEmccArgs>$(CompileEmccArgs) -s USE_SDL=0 $(EmccOptimizationSetting)</CompileEmccArgs>
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -326,6 +326,11 @@ The .NET Foundation licenses this file to you under the MIT license.
 
   </Target>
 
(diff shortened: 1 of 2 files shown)


yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -326,6 +326,11 @@ The .NET Foundation licenses this file to you under the MIT license.
 
   </Target>
 
+  <PropertyGroup Condition="'$(NativeCodeGen)' == 'llvm'">
+    <EmccOptimizationSetting Condition="$(Optimize) == 'true' and $(IlcOptimizationPreference) != 'Size'">-O3</EmccOptimizationSetting>
```
SingleAccretion, line 330: Is `-O3` what we want for `-c Release` as the default?

In native compilers it usually implies aggressive speed-to-size tradeoff i. e. larger unrolling thresholds, etc.

What is the size impact on (say) `HelloWasm.wasm` for `-O1/-O2/-O3`?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -340,7 +345,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileEmccArgs Condition="'$(NativeDebugSymbols)' != 'true'">$(CompileEmccArgs) -O2</CompileEmccArgs>
       <CompileEmccArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(CompileEmccArgs) -g3</CompileEmccArgs>
```
SingleAccretion: The `O2` addition here should presumably be deleted now.

yowl: yes, indeed.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: I took off `-g3` at compile and `-g` at link time, `lto` is on and `binaryen` runs.  Added `Oz` for comparison
```
EXEC cannot compile inline asm [E:\GitHub\runtimelab3\src\tests\nativeaot\SmokeTests\HelloWasm\HelloWasm.csproj]
```
|Option | Size |
|--------|-----|
|O1|3,375,405|
|O2|2,952,589|
|O3|2,936,045|
|Oz|2,501,692|

SingleAccretion: So it looks like the right default after all. Interesting that `-Oz` is so much smaller, I wonder why (not necessary to investigate at this point, though).

SingleAccretion (approved):
LGTM!

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
jkotas, line 330: IlcOptimizationPreference was renamed to OptimizationPreference upstream. Something to keep in mind for next integration.

jkotas (approved):
Thank you!

PR #2272, opened 2023-05-14
[NativeAOT-LLVM] Prepare the compiler for multithreaded execution

SingleAccretion:
Encapsulate LLVM state dependent on single-threaded access into a "context". Current implementation will produce one bitcode file per one single-threaded context.

This change does not enable `--parallelism:<not 1>` by itself, as some significant problems still remain:
1) Determinism. Regular backends achieve this by sorting already-produced nodes.
2) Build integration. For proper MSBuild incrementality, we would like to have a list of inputs and outputs (bitcode files) that is known ahead of time. Not hard to fix.
3) Last I measured, the gains were underwhelming (~30% bitcode generation time reduction). This needs to be investigated in more depth.

So mostly this change is just making sure we are well set up for writing MT-safe code by default in the compiler.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -103,9 +103,8 @@ The .NET Foundation licenses this file to you under the MIT license.
   </PropertyGroup>
 
   <ItemGroup Condition="'$(NativeCodeGen)' == 'llvm'">
-    <!-- TODO-LLVM: remove the IL objects when the IL backend is gone -->
     <LlvmObjectNames Include="$(TargetName)" />
-    <LlvmObjectNames Include="$(TargetName)clrjit" />
+    <LlvmObjectNames Include="$(TargetName).data" />
     <LlvmObjectNames Include="$(TargetName).external" />
 
     <LlvmObjects Include="@(LlvmObjectNames->'$(NativeIntermediateOutputPath)%(Identity)$(LlvmObjectExt)')">
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
@@ -2,12 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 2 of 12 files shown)


yowl:
~~Realise this is draft, but can I ask what makes the LLVM generation (with perhaps https://releases.llvm.org/3.1/docs/ProgrammersManual.html#startmultithreaded) different to other code gens?~~

~~A more current link https://www.llvm.org/docs/ProgrammersManual.html#threads-and-llvm~~
Nevermind, I see this is what you are talking about having RTM'ed.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -103,9 +103,8 @@ The .NET Foundation licenses this file to you under the MIT license.
   </PropertyGroup>
 
   <ItemGroup Condition="'$(NativeCodeGen)' == 'llvm'">
-    <!-- TODO-LLVM: remove the IL objects when the IL backend is gone -->
     <LlvmObjectNames Include="$(TargetName)" />
-    <LlvmObjectNames Include="$(TargetName)clrjit" />
+    <LlvmObjectNames Include="$(TargetName).data" />
     <LlvmObjectNames Include="$(TargetName).external" />
```
SingleAccretion, line 108: Notable change: formerly the bitcode file produced by ILC (with data symbols) was the "main" one, while all the code was in `clrjit.bc`. Now the file with code is the "main" one and ILC-produced bitcode is named `.data.bc`.

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -25,9 +26,10 @@ namespace ILCompiler
 {
     public sealed class LLVMCodegenCompilation : RyuJitCompilation
     {
+        private ConcurrentDictionary<int, CorInfoImpl> _corinfos;
         private readonly Dictionary<TypeDesc, LLVMTypeRef> _llvmStructs = new();
-        private readonly ConditionalWeakTable<Thread, CorInfoImpl> _corinfos = new();
         private string _outputFile;
```
SingleAccretion, line 31: CWT is not suitable for us because we need to call the Jit after all of the threads have finished work. With CWT, some of the threads might have exited already and their `CorInfoImpl`s - collected.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
LGTM.  Sharing struct maps not worth the synchronisation I guess.

jkotas:
Thanks

Issue #2169, opened 2023-01-14
[NativeAOT-LLVM] Exception handling: outstanding tasks


SingleAccretion:
With #2145 we will have the basic runtime mechanism for dispatching exceptions in place, however, much work still remains for the implementation to reach non-prototype level of quality.

Correctness:
1) Inter-frame dispatch and filter ordering.
   - Capture the shadow stack of the throwing frame inside the native exception.
     - Call dispatchers "on" this shadow stack (will require some dispatch pieces to be moved to native code).
   - Capture the list of pending fault/finally handlers in the first pass and invoke them in the second.
     - Should the managed or native exception carry this list? Can we avoid heap allocations for this case by using the shadow stack somehow?
   - Hard problem: dynamic `stackalloc` live into handlers.
     - Current plan of action is to implement this with a custom stack-like allocator.
     - We should find/write an extensive set of tests for this scenario.
   - Fixed in #2284.
2) We need to be able to identify the "top frame", one after which the exception will go unhanded. This implies wrapping RPI methods inside catch-all `try/catch` (perhaps a native one):
   - Verify: generated (managed) `main`, `Thread.Start` entrypoints.
   - Fixed in #2274.
3) Related: we need to properly handle unhandled exceptions.
   - Invoke relevant SPCL callbacks.
   - Dump the stack trace.
   - Fixed in #2274.
4) I think there are also some first-chance handlers we need to invoke when throwing.
   - Fixes in #2287.
5) Exceptional exits from runtime imports will fail to restore the shadow stack, leading to shadow stack top corruption. This should be fixed by moving the shadow stack top restore logic into RPI frames.
    - Fixed in #2208.
6) `throw null` should be translated into `throw new NullReferenceException()`.
    - Fixed in #2218.
7) Rethrow currently recaptures the stack trace. It should not.
   - Fixed in #2284.
8) Throwing filters should return "continue search".
   - Fixed in #2287.
9) Throwing faults are not handled correctly, see `src\tests\JIT\Methodical\eh\nested\general\methodthrowsinfinally_d.csproj`. 
   - Fixed in #2374.
10) `wasi-wasm` needs to be done using "manual" unwind (in both passes). Fortunately, that is feasible because exceptions cannot propagate across any managed<->unmanaged boundaries, even when it comes the the native runtime (in other words, the runtime can only invoke managed code when in a tail position).
    - Fixed in #2387.

Performance / code size:
[...]

PR #2274, opened 2023-05-15
[NativeAOT-LLVM] Implement unhandled exception handling

SingleAccretion:
Via the following RyuJit transform for RPI methods:

```
void RPI() {
    { /* User code */ }
}
```

```
void RPI() {
    try {
        { /* User code */ }
    }
    when (object e) (HandleUnhandledException(e)) // Calls 'abort'
    catch { }
}
```

This now prints the exception stack trace nicely:
```
BEGIN EXECUTION
call C:\Users\Accretion\source\dotnet\diffs/../runtimelab/src/tests/Common/scripts/nativeaottest.cmd C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\ HelloWasm.dll
Unhandled Exception: System.Exception: Exception of type 'System.Exception' was thrown.
    at S.P.CoreLib.System.Runtime.EH.RhpThrowEx (<anonymous>:wasm-function[4150]:0x18d32a)
    at RhpThrowEx (<anonymous>:wasm-function[1873]:0x24a7f)
    at HelloWasm.Program.Main (<anonymous>:wasm-function[3930]:0x162142)
    at HelloWasm.Module..MainMethodWrapper (<anonymous>:wasm-function[4271]:0x1a50fe)
    at invoke.ii (C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:6807:36)
    at HelloWasm.Module..StartupCodeMain (<anonymous>:wasm-function[4260]:0x1a403b)
    at main (<anonymous>:wasm-function[16611]:0x1180e92)
    at C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:1226:22
    at callMain (C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:7515:15)
    at doRun (C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:7569:23)
    at run (C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:7584:5)
    at runCaller (C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:7492:19)
    at removeRunDependency (C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:1147:7)
    at receiveInstance (C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:1314:5)
    at receiveInstantiationResult (C:\Users\Accretion\source\dotnet\runtimelab\artifacts\tests\coreclr\browser.wasm.Debug\nativeaot\SmokeTests\HelloWasm\HelloWasm\native\HelloWasm.js:1332:5)

Aborted(native code called abort())
```

Unfortunately, the host may then reprint a messier stack from the exception that's used to implement `abort` in Emscripten, but nothing we can do about that (short of using a custom `abort`, which does not seem worth it).

Contributes to #2169.


src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1838,6 +1840,13 @@ void Llvm::buildIntegralConst(GenTreeIntConCommon* node)
 
 void Llvm::buildCall(GenTreeCall* call)
 {
+    if (isUnhandledExceptionHandler(call))
+    {
+        // Note: the exception object argument is already on the shadow stack since we're in a filter.
+        emitHelperCall(CORINFO_HELP_LLVM_EH_UNHANDLED_EXCEPTION, {}, /* doTailCall */ true);
+        return;
+    }
```
SingleAccretion, line 1848: This scheme is a hack, but one we will have to live with as we can only reference LLVM helpers in codegen.

src/coreclr/jit/flowgraph.cpp
```diff
@@ -3,6 +3,12 @@
 
 #include "jitpch.h"
 
+#ifdef TARGET_WASM
+#include "llvm.h"
+#define max(a, b) (((a) > (b)) ? (a) : (b))
+#define min(a, b) (((a) < (b)) ? (a) : (b))
+#endif // TARGET_WASM
```
SingleAccretion, line 10: We should figure out a `CodeGenInterface`-like solution at one point to avoid including all the LLVM headers into other compiler files.

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
```diff
@@ -117,6 +120,48 @@ private static void HandleExceptionWasmFault(void* pShadowFrame, DispatchData* p
             DispatchContinueSearch(exception, pDispatchData);
         }
 
+        // This handler is called by codegen for exceptions that escape from RPI methods (i. e. unhandled exceptions).
+        //
+        private static void HandleUnhandledException(object exception)
```
SingleAccretion, line 125: Note that this handler is not a "dispatcher" like the methods above it. This is an intentional simplification - it would have been possible to define a new dispatcher kind (essentially - filter without filter handler) and use it instead of this, and it would have been that little bit better size-wise. It would also have been possible to add the `try` wrapping much later and avoid pessimizations associated with having EH in code. I decided against both - RPI methods are rare, thus both CQ value and testability of such strategies is low.

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
```diff
@@ -117,6 +120,48 @@ private static void HandleExceptionWasmFault(void* pShadowFrame, DispatchData* p
...
+            if (pFailFastFunction == IntPtr.Zero)
+            {
+                FallbackFailFast(RhFailFastReason.UnhandledException, exception);
+            }
+
+            try
+            {
+                InternalCalls.RhpRawCalli_ViOII(pFailFastFunction, (int)RhFailFastReason.UnhandledException, exception, 0, 0);
+            }
+            catch when (true)
```
yowl, line 156: Is the `when (true)` a style guide?

SingleAccretion: No - it's required for correctness in the presense of multiple type universes. A catch-all `catch` is emitted as `catch (object)` by Roslyn, which would "burn" one `System.Object` into the EH table, and that `System.Object` may not be the same `System.Object` as the thrower observes.

(This capability is currently not used as the CoreLib and runtime are merged, but we keep it working)

yowl: Thanks!

yowl (approved):
LGTM, thanks.

PR #2275, opened 2023-05-16
NativeAOT-LLVM: Merge runtime to 17 May 2023

yowl:
This PR is the latest merge (to 83f71b53d7f08700fd059191859f7931cf5712f4) from `runtime`.

Set `RUNTIME_FLAVOR=CoreCLR` for `build-native.cmd` so that `STATIC_ICU` does not get set.  We need all the function renaming macros such as
```
#define u_charsToUChars(...) u_charsToUChars_ptr(__VA_ARGS__)
```
from `pal_icushim_internal.h` to avoid missing symbols.

Delete the assert from the beginning of `SsaBuilder::AddPhiArg`
```c++
    // For LLVM backed, every predecessor must have its own phi arg
    if (!block->IsLIR())
    {
        for (GenTreePhi::Use& use : phi->Uses())
        {
            assert(use.GetNode()->AsPhiArg()->GetSsaNum() != ssaNum);
        }
    }
```
Not sure that is ok, maybe it is fallout from the ASG refactoring that needs looking at?

Exclude `LowLevelLifoSemaphore` for Wasi, think its just for threading and we don't have the base  `LowLevelLifoSemaphoreBase`


yowl:
cc @dotnet/nativeaot-llvm

PR #2277, opened 2023-05-17
NativeAOT-LLVM: Only build the host for one target: browser_wasm_win.

yowl:
This PR attempts to address #2276 by just building the host (win-x64) package for one of the wasm targets, `browser_wasm_win`.

Thanks @SingleAccretion !


SingleAccretion (approved):

jkotas (approved):
Thanks!

PR #2283, opened 2023-05-19
NativeAOT-LLVM: Merge runtime to 17 May 2023 

yowl:
This PR is the latest merge (to 83f71b53d7f08700fd059191859f7931cf5712f4) from `runtime`.

Set `RUNTIME_FLAVOR=CoreCLR` for `build-native.cmd` so that `STATIC_ICU` does not get set.  We need all the function renaming macros such as
```
#define u_charsToUChars(...) u_charsToUChars_ptr(__VA_ARGS__)
```
from `pal_icushim_internal.h` to avoid missing symbols.

Delete the assert from the beginning of `SsaBuilder::AddPhiArg`
```c++
    // For LLVM backed, every predecessor must have its own phi arg
    if (!block->IsLIR())
    {
        for (GenTreePhi::Use& use : phi->Uses())
        {
            assert(use.GetNode()->AsPhiArg()->GetSsaNum() != ssaNum);
        }
    }
```
Not sure that is ok, maybe it is fallout from the ASG refactoring that needs looking at?

Exclude `LowLevelLifoSemaphore` for Wasi, think its just for threading and we don't have the base  `LowLevelLifoSemaphoreBase`


yowl:
Replaces https://github.com/dotnet/runtimelab/pull/2275
@dotnet/nativeaot-llvm

src/coreclr/jit/gentree.cpp
```diff
@@ -17721,19 +17713,12 @@ bool GenTree::IsPhiNode()
...
-    assert(!res || AsOp()->gtOp1->OperIs(GT_LCL_VAR) || OperGet() == GT_STORE_LCL_VAR);
 #else
-    bool res = OperIs(GT_ASG) && AsOp()->gtOp2->OperIs(GT_PHI);
-    assert(!res || AsOp()->gtOp1->OperIs(GT_LCL_VAR));
-#endif
-    return res;
-=======
     return OperIs(GT_STORE_LCL_VAR) && AsLclVar()->Data()->OperIs(GT_PHI);
->>>>>>> 83f71b53d7f08700fd059191859f7931cf5712f4
+#endif
```
SingleAccretion: This should not need the `ifdef TARGET_WASM` path anymore (no `ASG`s in SSA).

src/coreclr/jit/llvmlower.cpp
```diff
@@ -998,7 +997,7 @@ void Llvm::lowerIndir(GenTreeIndir* indirNode)
 
 void Llvm::lowerStoreBlk(GenTreeBlk* storeBlkNode)
 {
-    assert(storeBlkNode->OperIs(GT_STORE_BLK, GT_STORE_OBJ));
+    assert(storeBlkNode->OperIs(GT_STORE_BLK));
```
SingleAccretion, line 1000: There is a `TODO` below that can be addressed.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Environment.NativeAot.cs
```diff
@@ -63,5 +63,12 @@ internal static void ShutdownCore()
         }
 
         public static int TickCount => (int)TickCount64;
+
+#if TARGET_WASM
+        private static unsafe string[] GetCommandLineArgsNative()
```
SingleAccretion, line 68: Usual question - why didn't we need this before?

src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/RyuJitCompilationBuilder.cs
```diff
@@ -21,7 +21,8 @@ public class RyuJitCompilationBuilder : CompilationBuilder
         private ProfileDataManager _profileDataManager;
         private string _jitPath;
 
-        protected RyuJitCompilationBuilder(CompilerTypeSystemContext context, CompilationModuleGroup group, NodeMangler mangler)
+        protected RyuJitCompilationBuilder(CompilerTypeSystemContext context, CompilationModuleGroup group,
+            NodeMangler mangler)
```
SingleAccretion: Intentional formatting change?

src/libraries/System.Private.CoreLib/src/System/Environment.cs
```diff
@@ -88,7 +88,6 @@ public static void SetEnvironmentVariable(string variable, string? value, Enviro
 
 #if !MONO
         internal static string[]? s_commandLineArgs;
-
```
SingleAccretion: Intentional formatting change?

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
```diff
@@ -2607,4 +2607,4 @@
     <Compile Include="$(MSBuildThisFileDirectory)System\Numerics\IUnaryPlusOperators.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Numerics\IUnsignedNumber.cs" />
   </ItemGroup>
-</Project>
\ No newline at end of file
+</Project>
```
SingleAccretion: Unintentional formatting change?

src/native/libs/System.Globalization.Native/CMakeLists.txt
```diff
@@ -94,7 +89,7 @@ if (LOCAL_BUILD)
     # For minipal files
     include_directories(../../)
     include_directories(${CMAKE_CURRENT_BINARY_DIR})
-elseif (CLR_CMAKE_TARGET_MACCATALYST OR CLR_CMAKE_TARGET_IOS OR CLR_CMAKE_TARGET_TVOS OR CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI)
+elseif ((CLR_CMAKE_TARGET_MACCATALYST OR CLR_CMAKE_TARGET_IOS OR CLR_CMAKE_TARGET_TVOS OR CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI) AND NOT RUNTIME_FLAVOR STREQUAL "CoreCLR")
     add_definitions(-DSTATIC_ICU)
     set(NATIVEGLOBALIZATION_SOURCES ${NATIVEGLOBALIZATION_SOURCES} pal_icushim_static.c)
 else()
```
SingleAccretion: I am assuming this is the fallout from https://github.com/dotnet/runtime/pull/82393, what I don't quite understand is what the difference is between what's upstream and what's downstream. It sounds like https://github.com/dotnet/runtime/pull/82393 would be "just what we want", but clearly it isn't. Is the core difference that we build ICU from source here and upstream gets it as a binary from somewhere?

That aside, if we must have NAOT-LLVM be different, I don't think indirecting it through the runtime flavor is useful (it'll always be set to `CoreCLR` by the script anyway), we should just call this difference out and do something simpler, e. g. comment this whole block out.

SingleAccretion:
Thank you for pulling the conflicts forward!

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Environment.NativeAot.cs
yowl: https://github.com/dotnet/runtime/pull/84953/files#diff-145407db2c27bf3412239087b0d37526ff44f7f7a60b3dc50e1d5f192d619e55R89-R101

We had `GetCommandLineArgsNative` before in [Environment.CoreCLR.cs](https://github.com/dotnet/runtime/pull/84953/files#diff-01661fb4013eaaa2309e399d45cd56a0cc148594e2dcd8e10e7453492d320939) but that was in `src\coreclr\System.Private.CoreLib\System.Private.CoreLib.csproj`, but now it is referenced from `src/libraries/System.Private.CoreLib/src/System/Environment.cs` so we need a definition

src/native/libs/System.Globalization.Native/CMakeLists.txt
yowl: > upstream gets it as a binary from somewhere?

That looks to be it, mono is getting `libicuuc.a` and `libicui18n.a` ultimately from `$(PkgMicrosoft_NETCore_Runtime_ICU_Transport)` .  I can't see where that comes from, the SDK perhaps.
https://github.com/dotnet/runtime/blob/d7a30ba1f1c9ccce26a6a223898b7b9aba186a3a/src/mono/wasm/wasm.proj#L26

I will try to do the same and drop `-s USE_ICU=1` from the emcc build

src/libraries/System.Private.CoreLib/src/System/Environment.cs
yowl: reverted

src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/RyuJitCompilationBuilder.cs
yowl: reverted

src/coreclr/jit/llvmlower.cpp
yowl: Thanks, done.

src/coreclr/jit/gentree.cpp
yowl: Thanks, indeed, removed.

jkotas:
Ready to be merged?

Thank you for staying up to date with the upstream!

yowl:
No, not yet thanks.  Looking at if we can get inline with mono regards the ICU library.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Environment.NativeAot.cs
SingleAccretion: And the natural definition comes from `src/libraries/System.Private.CoreLib/System/Environment.Unix.cs`, not included for WASM (by design). It also returns `Array.Empty<string>();`, this should too I think.

src/native/libs/System.Globalization.Native/CMakeLists.txt
yowl: This isn't quite as straightforward as I thought because of https://github.com/emscripten-core/emscripten/issues/17540#issuecomment-1207463409  .  I was going to link `libicuuc.a` and friends into `libSystem.Globalization.Native.a` but that (linking one static library into another) seems hard with cmake as per that comment.  Mono solves this by linking into an executable (`dotnet.native`) which cmake does do.

Would it be acceptable therefore to do something similar to mono and link `libicuuc.a`, `libicui18n.a` and `libicudata.a` from the `microsoft.netcore.runtime.icu.transport` package at "user" link time.  I.e. bring in via a `<PackageReference Include="Microsoft.NETCore.Runtime.ICU.Transport" PrivateAssets="all" Version="$(MicrosoftNETCoreRuntimeICUTransportVersion)" GeneratePathProperty="true" />` from  a build integration file?

jkotas: The transport packages are meant for transport of bits between builds of different dotnet org repos. They are not meant to be used by "user" build.

You would want to repackage the ICU bits from the transport package into the `Microsoft.DotNet.ILCompiler` package.

yowl: I've not gone the repackaging route for now because I was able to just put a cmake post build event in:
```
if (CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI)
   add_custom_command(TARGET System.Globalization.Native-Static
       POST_BUILD 
       COMMAND ${CMAKE_AR} qL $<TARGET_FILE:System.Globalization.Native-Static> ${CMAKE_ICU_DIR}/lib/libicuuc.a ${CMAKE_ICU_DIR}/lib/libicui18n.a ${CMAKE_ICU_DIR}/lib/libicudata.a)
endif()
```
which resolves to a `llvm-ar` command appending the contents of these 3 archives to `libSystem.Globalization.Native.a`  This seemed simpler, but for it to work with mono, we would have to see if the mono team are ok with removing 
https://github.com/dotnet/runtimelab/blob/2f8f4a90ca6deb76ab2551ba4aca56713267a5cb/src/mono/wasm/runtime/CMakeLists.txt#L19-L20

And understand why we need the data file `libicudata.a` and mono does not (or maybe they do need it for the Aot version?)

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Environment.NativeAot.cs
yowl: Done, thanks

src/coreclr/runtime.proj
```diff
@@ -58,8 +58,8 @@
       <_CoreClrBuildArg Condition="'$(ClrCrossComponentsSubset)' == 'true'" Include="-component crosscomponents" />
     </ItemGroup>
 
-    <ItemGroup Condition="'$(TargetsAppleMobile)' == 'true'">
-      <_CoreClrBuildArg Include="-cmakeargs -DCMAKE_ICU_DIR=&quot;$(_IcuDir)&quot;" />
+    <ItemGroup Condition="'$(TargetsAppleMobile)' == 'true' or '$(TargetArchitecture)' == 'wasm'">
+      <_CoreClrBuildArg Include="-cmakeargs &quot;-DCMAKE_ICU_DIR=$(_IcuDir)&quot;" />
```
SingleAccretion, line 62: Is the quoting change intentional?

src/native/libs/System.Globalization.Native/CMakeLists.txt
```diff
@@ -94,7 +81,7 @@ if (LOCAL_BUILD)
     # For minipal files
     include_directories(../../)
     include_directories(${CMAKE_CURRENT_BINARY_DIR})
-elseif (CLR_CMAKE_TARGET_MACCATALYST OR CLR_CMAKE_TARGET_IOS OR CLR_CMAKE_TARGET_TVOS OR CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI)
+elseif ((CLR_CMAKE_TARGET_MACCATALYST OR CLR_CMAKE_TARGET_IOS OR CLR_CMAKE_TARGET_TVOS OR CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI))
```
SingleAccretion:
```suggestion
elseif (CLR_CMAKE_TARGET_MACCATALYST OR CLR_CMAKE_TARGET_IOS OR CLR_CMAKE_TARGET_TVOS OR CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI)
```

SingleAccretion (approved):
LGTM with two comments, thank you!

The ICU solution is very neat, great to see that.

src/coreclr/runtime.proj
yowl: Yes, it was.  I had problems with the value getting dropped, but wasn't sure why.  With the quoting as it was I get

```
Calling "E:\GitHub\runtimelab\eng\native\gen-buildsys.cmd" "E:\GitHub\runtimelab\src\coreclr" "E:\GitHub\runtimelab\artifacts\obj\coreclr\browser.wasm.Debug" vs2022 wasm browser "-DRUNTIME_FLAVOR=CoreCLR" "-DCMAKE_BUILD_TYPE=Debug" "-DCLR_CMAKE_KEEP_NATIVE_SYMBOLS=1" "-DCLR_CM
  AKE_TARGET_ARCH=wasm" "-DCLR_CMAKE_TARGET_OS=emscripten" "-DCLI_CMAKE_FALLBACK_OS=browser" "-DCLR_CMAKE_PGO_INSTRUMENT=0" "-DCLR_CMAKE_OPTDATA_PATH=" "-DCLR_CMAKE_PGO_OPTIMIZE=0" -DCMAKE_ICU_DIR
```
You can see `-DCMAKE_ICU_DIR` had no value, I wasn't sure if it was the backslash  in `c:\` or something else.

yowl: And as the other `-D` arguments were quoted to include the `-D...` I did the same.

PR #2284, opened 2023-05-20
[NativeAOT-LLVM] Implement two-pass exception handling

SingleAccretion:
Implements the two-pass system which #2145 laid the groundwork for.

The general idea is that we use native exception handling for the first pass, caching second-pass handlers in a linked list, to be invoked in the second pass. With this, our EH semantics should align fully with .NET proper modulo bugs.

One area where a two-pass scheme like this causes problems is dynamic stack allocation. We can no longer use the native stack in cases where the underlying data may escape into handlers, and so much of this change is focused on adding the implementation and tests for another kind of "stack" - "the dynamic stack", a specialized allocator, used only for `localloc`s in methods with EH. This allocator tracks the shadow stack to know how much state to release and is invoked by codegen on method return (handling normal flow) and by the dispatchers at the very end of the second pass (handling exceptional flow).

Contributes to #2169.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl:
> We can no longer use the native stack in cases where the underlying data may escape into handlers, and so much of this change is focused on adding the implementation and tests for another kind of "stack" - "the dynamic stack"

I understand why we can't use the native stack for `stackalloc` but I'm not clear on why a `stackalloc` variable cannot be accessed from an EH the same as any other variable on the shadow stack that escapes to a EH?

SingleAccretion:
It is a very good question - why not use the shadow stack for such `stackalloc`s? This is certainly a tradeoff, the considerations were:
1) Pro: the allocation path is fast. Faster than native in fact, as we don't do stack probing (and if/when we start to, it'll be relatively cheap).
2) Pro: no need to write a specialized allocator; more robust.
3) Con: this introduces dynamic shadow stack allocation. Dynamic stack allocation in general is a qualitative leap in complexity (see, e. g., #85548) and may make implementing some future features more difficult and/or make them more expensive. One such feature I had in mind is precise GC. More immediately, this would complicate the compiler.
4) Con: large native allocations on the shadow stack that are actually "not needed" (realized escape into a handler should be very rare, after all) slow down GC scanning.
5) Con: Having an efficient stack allocator like may be interesting for other purposes (like, say, allocating `FaultNode`s in the dispatch path...). This was a minor point.

I also considered allocating a pinned managed array, this would have been the simplest and most robust solution of all, but also the least performant (so I rejected it).

src/coreclr/jit/llvmlower.cpp
```diff
@@ -450,14 +455,28 @@ void Llvm::lowerLocals()
...
         }
     }
 
+    if ((shadowStackLocals.size() == 0) && m_lclHeapUsed && doUseDynamicStackForLclHeap())
+    {
+        // The dynamic stack is tied to the shadow one. If we have an empty shadow frame with a non-empty dynamic one,
+        // an ambiguity in what state must be released on return arises - our caller might have an empty shadow frame
+        // as well, but of course we don't want to release its dynamic state accidentally. To solve this, pad out the
+        // shadow frame in methods that use the dynamic stack if it is empty. The need to do this should be pretty rare
+        // so it is ok to waste a shadow stack slow here.
```
yowl: nit: typo slow->slot ?

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -2456,6 +2462,456 @@ private static void TestFilterNested()
...
+
+                result = stkTwo == stkThree && stkOne != stkThree && Volatile.Read(ref stkOne[StkAllocSize - 1]) == 2;
+                Volatile.Write(ref *(int*)null, 0);
+            }
+            catch (NullReferenceException)
+            {
+                Volatile.Read(ref stkThree[StkAllocSize - 1]);
+            }
+        }
+        EndTest(result);
```
yowl, line 2736: Should we be resetting `result` to false before each test, in case the code never gets to the assignment?

yowl (approved):
Thanks!

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
SingleAccretion: Did that in cases where `result` assignment wasn't the last thing a test would do.

yowl (approved):

PR #2288, opened 2023-05-25
NativeAOT-LLVM: Do not share memory for wasi

yowl:
This PR removes the sharing of memory for the wasi link by removing `-Wl,--import-memory,--export-memory`.  Many wasi tools are not yet ready for threads (where sharing of memory might be useful), nor is sharing applicable for components which use a shared nothing design.  The user can always add them back with  `LinkerArg` but they can't easily remove them


yowl:
cc @dotnet/nativeaot-llvm

SingleAccretion (approved):

PR #2289, opened 2023-05-25
[NativeAOT-LLVM] Fix a few issues

SingleAccretion:
The unhandled exception handling code needs to always reset the scratch block, since it'll put it into a protected region.

Some helpers changed signatures in https://github.com/dotnet/runtime/pull/85962.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl:
Interesting,  I wonder if this is the fix for the assert I am hitting at https://github.com/dotnet/runtime/blob/12f4ed6189486597352a5264bdc69a08d172d8ba/src/coreclr/jit/fgbasic.cpp#L323

yowl (approved):

jkotas (approved):
Thank you

Issue #2293, opened 2023-05-29
[NativeAOT-LLVM] Multi-threaded compiler


SingleAccretion:
This issues is a collection of notes so far collected in the process of thinking about how to make LLVM compilation amenable to multithreading.

Recall, for context, that we have two major contributors to compile time right now: the IL -> LLVM phase (RyuJit) and the LLVM -> WASM phase (Clang). For both of these, we are constrained by LLVM's design to work on `Module`s. For maximally optimal throughput, working on the same number of modules as the `parallelism` setting is required.

The hard problem to solve, then, is determinism. There are two possible levels:
1) Deterministic w.r.t. `parallelism` only, i. e. we allow builds with different `parallelism` settings to produce different results. This is believed to be achievable with a fairly simple scheme not too far from what we currently do. I don't think it is viable, however, as `parallelism`, if not specified, is implicitly set to the number of logical cores, so this only achieves a weak form of local determinism.
2) Fully deterministic, i. e. all builds, regardless of `parallelism`, produce identical results. This is what ILC proper has and what we should have as well.

Let's review what other toolchains do to achieve `2` and how it applies to us:
1) MonoAOT: compiles one module per one managed assembly. Downside: large assemblies like SPC would take a disproportionate amount of time. This is a variation of the "partition the input based on something inherent in the input itself". For ILC, partitioning by assembly is a bit unnatural.
2) ILC proper: sorts the outputs before writing them out to an object file. For us, that would mean sorting the resulting `.wasm`, as we otherwise appear to have no control on the order LLVM emits things in. This would allow using the most efficient `parallelism <-> Module` mapping during compilation, but would then be both "hard" (WASM binary format is a fragile one, as, for example, changing the function's index may invalidate all references to it from everywhere), expensive to maintain and adds to compile time by itself.
3) C++ compilers once again use a variant of partitioning the input based on itself (i. e. translation units). Translation units are generally smaller than .NET assemblies, and there are usually more of them than the `parallelism` setting. It is known that compiling larger `Module`s is better than many small ones (especially when doing this naively with a new compiler process for each one), though both are obviously better than compiling everything on a single thread.

[...]

PR #2297, opened 2023-06-04
[NativeAOT-LLVM] Initial version of a multi-threaded compiler

SingleAccretion:
This change implements the multi-threading scheme described in #2293.

We partition the incoming into `8` modules, regardless of the `--parallelism` setting, which achieves the desired determinism characteristics. These modules are compiled on multiple threads both when generating LLVM and when generating machine code (i. e. invoking `clang`).

The peformance results of this work can be summarized as "modest", though some gains, especially in Release builds, are quite visible (up to 2x speedups on my machine). For yet unclear reasons, parallelism inside ILC itself is not fully utilized, however, that is also not the bottleneck of the build, as especially in Debug a lot of time is spent in `wasm-esmcripten-finalize.exe`. Time is also spent with the disk interactions - ideally the next step would be to move the Clang part of the build in-process (or do something like thin-LTO).

I have put together a small sheet of the results as they apply to `HelloWasm` on my machine:

Module count | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |   | NAOT ILC / 8
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Total   compilation time (Release) | 01:03.3 | 42.0 | 37.8 | 33.7 | 33.1 | 32.8 | 30.9 | 30.3 |   | 26.6
  |   | 01:04.7 | 41.9 | 39.2 | 33.1 | 32.5 | 32.9 | 31.6 | 30.1 |   | 27.6
  |   |   |   |   |   |   |   |   |   |   |  
Bitcode   compilation time (Release) | 6.66 | 5.36 | 5.04 | 4.75 | 4.58 | 4.57 | 4.45 | 4.55 |   | 3.84
  |   | 7.72 | 5.34 | 4.84 | 4.78 | 5 | 4.6 | 4.46 | 4.53 |   | 3.71
  |   |   |   |   |   |   |   |   |   |   |  
WASM file   size | 4.672 | 4.537 | 4.5 | 4.483 | 4.467 | 4.451 | 4.457 | 4.455 |   | 
  |   |   |   |   |   |   |   |   |   |   |  
Total   compilation time (Debug) | 1:05.9 | 53.4 | 51.8 | 49.6 |   |   |   | 49.3 |   | 46.6
  |   | 1:04.9 | 53.1 | 52.9 |   |   |   |   |   |   | 44.7
  |   |   |   |   |   |   |   |   |   |   |  
Bitcode   compilation time (Debug) | 13.65 | 11.05 | 10.72 | 10.14 |   |   |   | 9.92 |   | 5.62
  |   | 13.82 | 11.12 | 10.65 |   |   |   |   |   |   | 5.31

Note that the measurement were done using a CoreCLR-based ILC (except for the `NAOT / 8` column).

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMCodegenCompilation.CodeGen.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMCodegenCompilation.CodeGen.cs
@@ -1,0 +1,392 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Numerics;
+
+using ILCompiler.DependencyAnalysis;
+
+using Internal.JitInterface;
+using Internal.TypeSystem;
+using Internal.TypeSystem.Ecma;
+
+using LLVMSharp.Interop;
+
+namespace ILCompiler
+{
+    public sealed partial class LLVMCodegenCompilation
+    {
+        private readonly Dictionary<TypeDesc, LLVMTypeRef> _llvmStructs = new();
+
+        // We define an alternative entrypoint for the runtime exports, one that has the (original) managed calling convention.
+        // This allows managed code as well as low-level runtime helpers to avoid the overhead of shadow stack save/restore
+        // when calling the export. Thus, the "mangling" we use here is effectively an ABI contract between the compiler and
+        // runtime.
+        public override string GetRuntimeExportManagedEntrypointName(MethodDesc method)
+        {
+            if (!method.HasCustomAttribute("System.Runtime", "RuntimeExportAttribute"))
+            {
+                return null;
+            }
+
+            return ((EcmaMethod)method).GetRuntimeExportName() + "_Managed";
+        }
+
+        public override ISymbolNode GetExternalMethodAccessor(MethodDesc method, ReadOnlySpan<TargetAbiType> sig)
+        {
+            Debug.Assert(!sig.IsEmpty);
+            string name = PInvokeILProvider.GetDirectCallExternName(method);
+
+            return NodeFactory.ExternSymbolWithAccessor(name, method, sig);
+        }
+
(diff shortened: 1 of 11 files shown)


src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilationBuilder.cs
```diff
@@ -72,7 +77,9 @@ internal void SetOptions(IEnumerable<KeyValuePair<string, string>> options)
         // S128 natural alignment of stack
         public string DataLayout { get; private set; } = "e-m:e-p:32:32-i64:64-n32:64-S128";
         public string Target { get; private set; } = "wasm32-unknown-emscripten";
-        public string ModuleName { get; private set; } = "netscripten";
         public CorInfoLlvmEHModel ExceptionHandlingModel { get; private set; } = CorInfoLlvmEHModel.Cpp;
+
+        // Below options are debug-only and not supported.
+        public int MaxLlvmModuleCount { get; private set; } = 8;
```
SingleAccretion, line 83: This is the hardcoded number. Ideally, it would be correlated with how "large" the input is instead.

`8` is a conservative value, but this is also an incomplete solution, so I don't think there is great point in tuning it.

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMCodegenCompilation.CodeGen.cs
```diff
@@ -0,0 +1,392 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
SingleAccretion, line 2: This has been split out from `LLVMCodegenCompilation.cs`; no code changes.

SingleAccretion:
Let me add a test first here.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

PR #2299, opened 2023-06-06
[NativeAOT-LLVM] Fix a determinism issue in `lowerSpillTempsLiveAcrossSafePoints`

SingleAccretion:
And stop spilling frozen objects.

Encountered in #2297.

### DIFF
--- a/src/coreclr/jit/llvmlower.cpp
+++ b/src/coreclr/jit/llvmlower.cpp
@@ -229,7 +229,7 @@ void Llvm::lowerSpillTempsLiveAcrossSafePoints()
             // Locals are handled by the general shadow stack lowering (already "spilled" so to speak).
             // Local address nodes always point to the stack (native or shadow). Constant handles will
             // only point to immortal and immovable (frozen) objects.
-            return !node->OperIsLocal() && !node->OperIs(GT_LCL_ADDR);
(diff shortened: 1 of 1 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm 

Required for #2297.

src/coreclr/jit/llvmlower.cpp
```diff
@@ -140,8 +140,22 @@ void Llvm::Lower()
...
+    struct DeterministicNodeHashInfo : public HashTableInfo<DeterministicNodeHashInfo>
+    {
+        static bool Equals(GenTree* left, GenTree* right)
+        {
+            return left == right;
+        }
+
+        static unsigned GetHashCode(GenTree* node)
+        {
+            return node->TypeGet() ^ node->OperGet();
```
yowl, line 153: For my benefit, how  is this value good enough to be deterministic, on the surface looks like you could have 2 nodes with the same type and operation that would not be deterministic?

SingleAccretion: It is deterministic in the sense that it always returns the same value for the same node.

It is ok for the iteration order to be "random" in the sense of not being related to insertion order, what was not ok is the it run-to-run variance introduced by the lower bits of node pointers being different.

yowl (approved):
Thanks

PR #2303, opened 2023-06-09
[NativeAOT-LLVM] Put LLVM helpers into the `CorInfoHelpFunc` enum directly

SingleAccretion:
This allows them to be referenced in RyuJit IR, which is a requirement for #2304.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
LGTM!

PR #2306, opened 2023-06-13
[NativeAOT-LLVM] Stub out finalization and enable some tests

SingleAccretion:
Finalization "support" available here is only that we'll run finalizers when explicitly requested - which is enough to enable the two tests I was looking to enable.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/wasm/FinalizerHelpers.SingleThreaded.cpp
+++ b/src/coreclr/nativeaot/Runtime/wasm/FinalizerHelpers.SingleThreaded.cpp
@@ -45,7 +45,7 @@ void RhEnableFinalization()
 
 EXTERN_C NATIVEAOT_API void __cdecl RhWaitForPendingFinalizers(UInt32_BOOL allowReentrantWait)
 {
-    // Must be called in cooperative mode as "ProcessFinalizersAndReturn" RPIs back into managed.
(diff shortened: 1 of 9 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/nativeaot/Runtime/wasm/FinalizerHelpers.SingleThreaded.cpp
```diff
@@ -0,0 +1,76 @@
...
+// This method is called at the end of GC in case finalizable objects were present.
+void RhEnableFinalization()
+{
+    // TODO: Implement automatic finalization. Cannot just call "ProcessFinalizersAndReturn"
+    // here as it will deadlock the GC.
+}
+
+EXTERN_C NATIVEAOT_API void __cdecl RhWaitForPendingFinalizers(UInt32_BOOL allowReentrantWait)
+{
+    // Must be called in cooperative mode as "ProcessFinalizersAndReturn" RPIs back into managed.
```
jkotas: Should this comment be other way around? The assert is verifying that it is not called in cooperative mode.

SingleAccretion: > Should this comment be other way around?

Yep. Fixed it.

yowl (approved):
Look good, thanks!

Issue #2307, opened 2023-06-13
[NativeAOT-LLVM] Enable tests and fix bugs


SingleAccretion:
1) We should never produce "naked" `undef`/`poison` values. These can be produced implicitly when doing things like `(int)float.NaN`. Open questions:
    - How many `freeze`s do we need?
      - On all operations that can produce undefined values? Note this set is quite broad.
    - How do we test this? Especially the runtime case.
      - Write a bespoke LLVM IR verifier that would check all potentially-undefined-value-producing operations are frozen before use?
2) Dynamically (and, technically, statically) created fat function pointers may be allocated at a 2GB+ address, making the masking we do (of a sign bit) ineffective. We need to detect this case and fail fast / OOM at runtime.
4) A few alignment-related things are not correct: unaligned loads/stores are not annotated correctly, plus block layout locals in `alloca`s are not aligned to pointer size.

SingleAccretion:
> Reflection smoke test fails on creating a delegate with byref-likes involved. Likely a codegen bug.

Turns out not. This is the `RhGetCodeTarget` NYI. It is non-trivial to implement `RhGetCodeTarget` on WASM with current LLVM features. Ideas:
1) Add an LLVM+linker feature that allows fine-grained control over how the indirect function table is constructed. This would allow for an efficient lookup scheme (e. g.: continuous range for unboxing stubs and their targets, stubs are even, targets are odd).
2) Same as `1`, but can be done today, at greater complexity cost, by rewriting and/or creating custom `.wasm` files.
3) Lookup data structures (similar to how native unwind info is used to store this information for instantiating unboxing stubs on Unix). Remains to be seen how expensive this would be size-wise. Note: `RhGetCodeTarget` is meant to be "cheap" as it is used in delegate equality comparisons.
4) Rewrite code such that `RhGetCodeTarget` is not needed, e. g. put both the unboxed and boxed entry points into the reflection invoke map. It is not clear how feasible that would be.

SingleAccretion:
> We definitely need `System.Runtime.Tests`

Small progress report: this is not going too smoothly. We have a few compiler bugs, with those fixed, the tests do compile in Release, but not in Debug because `wasm-emscripten-finalize.exe` consumes way too much memory (well - it did compile in ~5 hours, but only after OOMing the shell and making the rest of the system hardly usable for that time period).

yowl:
[...]

PR #2311, opened 2023-06-19
[NativeAOT-LLVM] Codegen fixes for libraries tests

SingleAccretion:
This is enough to compile S.R.Tests in Release and Debug. Unfortunately, Release crashes on thread creation (we need a Wasm-specific TP implementation like Mono has), while Debug crashes on functions having too many locals (perhaps we can tune our debug codegen to be less naive).

Contributes to #2307.

### DIFF
--- a/src/coreclr/jit/llvmlower.cpp
+++ b/src/coreclr/jit/llvmlower.cpp
@@ -454,7 +454,7 @@ void Llvm::lowerLocalsBeforeNodes()
                 else if (!_compiler->fgVarNeedsExplicitZeroInit(lclNum, /* bbInALoop */ false, /* bbIsReturn*/ false) ||
                          varDsc->HasGCPtr())
                 {
-                    var_types zeroType = (varDsc->TypeGet() == TYP_STRUCT) ? TYP_INT : genActualType(varDsc);
+                    var_types zeroType = (varDsc->TypeGet() == TYP_STRUCT) ? TYP_INT : varDsc->TypeGet();
                     initializeLocalInProlog(lclNum, _compiler->gtNewZeroConNode(zeroType));
                 }
             }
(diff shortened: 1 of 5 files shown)


src/coreclr/jit/llvmlower.cpp
```diff
@@ -687,6 +669,29 @@ void Llvm::insertProlog()
...
+                fieldVarDsc->lvIsParam = false;
+            }
+
+            varDsc->lvPromoted = false;
+            varDsc->lvFieldLclStart = BAD_VAR_NUM;
+            varDsc->lvFieldCnt = 0;
+        }
+    }
+}
+
```
SingleAccretion, line 694: We were not stripping all promotion annotations, which led to problems with SSA assuming parents use their (dependently promoted) fields. Since our model is that we don't have promotion after lowering, this strips all of it.

src/coreclr/jit/llvmlower.cpp
```diff
@@ -998,6 +1003,11 @@ void Llvm::lowerCall(GenTreeCall* callNode)
         {
             CurrentRange().Remove(CurrentRange().LastNode(), /* markOperandsUnused */ true);
         }
+
+        if (!CurrentBlock()->KindIs(BBJ_THROW))
+        {
+            _compiler->fgConvertBBToThrowBB(CurrentBlock());
+        }
```
SingleAccretion, line 1010: Some blocks weren't marked `BBJ_THROW` even as they had always-throwing calls in them.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -91,6 +91,30 @@ public static int getSignatureForMethodSymbol(IntPtr thisHandle, void* symbolHan
...
+                        TypeDesc pointerType = _this._compilation.TypeSystemContext.GetWellKnownType(WellKnownType.Void).MakePointerType();
+                        MethodSignatureFlags signatureFlags = MethodSignatureFlags.Static | MethodSignatureFlags.UnmanagedCallingConvention;
+                        MethodSignature signature = new MethodSignature(signatureFlags, 0, pointerType, new[] { pointerType });
+
+                        // We're technically leaking the signature object here, but we don't get here often, so it's ok.
+                        _this.Get_CORINFO_SIG_INFO(signature, pSig, scope: null);
+                        return 1;
+                    }
+                }
+            }
```
SingleAccretion, line 116: The compiler can feed codegen these allocator nodes as "bare" `ExternSymbolNode` when used as part of the `AllocatorOf<T>` intrinsic.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
LGTM, thanks!

PR #2313, opened 2023-06-20
[NativeAOT-LLVM] Enable JS BigInt integration by default

SingleAccretion:
This halves the build time of HelloWasm in Debug (~50s -> ~25s).

Even more significant gains are expected for larger projects (our libraries tests are literally unbuildable in Debug without this option).

Compatibility-wise, BigInt integration is universally older than exceptions, which we also enable by default, so we don't lose anything.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Well found, thanks

PR #2316, opened 2023-06-23
[NativeAOT-LLVM] Use the `__start`/`__stop` symbols to reference modules for WASM

SingleAccretion:
As it turns out, `wasm-ld` already supports emitting these special symbols, so all we need to do is make sure the section is kept alive via `@llvm.used`.

Discovered this accidentally while looking at linker source code.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.InteropServices;
 
 using ILCompiler.DependencyAnalysisFramework;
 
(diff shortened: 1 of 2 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -109,9 +113,6 @@ public static void EmitObject(string objectFilePath, IEnumerable<DependencyNode>
                         case ExternMethodAccessorNode accessorNode:
                             objectWriter.GetCodeForExternMethodAccessor(accessorNode);
                             continue;
-                        case ModulesSectionNode modulesSectionNode:
```
yowl, line 112: Is this just for multifile compilation?

SingleAccretion: Well, sort of. There is always one such node produced in a compilation (single-file too).

yowl: but we don't need it anymore because we are not using `RtRHeaderWrapper` ?

SingleAccretion: It is still needed and still generated. The only change is how the runtime gets to it - it uses the linked-generated symbols instead of `RtRHeaderWrapper`.

After the linker runs, the results end up looking something like this:
```
__start__modules
__modules (section with this node as the only member)
__stop__modules
```

yowl: oh, I think I see, its emitted now in `EmitKeepAliveList`

yowl (approved):
Looks good, thanks

Issue #2300, opened 2023-06-08
[NativeAOT-LLVM] ILC throughput improvements


SingleAccretion:
This issue is meant to capture ideas about ILC TP improvements. Initial list from a bit of looking at profiles of compiling debug code:

1) `fgFindBasicBlocks` is doing a prejit inlining scan of which the result will not be used. It looks not useful for ILC in general, currently, but I cannot find how is the "never" result propagated back into the image in CG2.
2) `emitHelperCall` is more expensive than it should be, mainly due to the EE calls. We should cache the `CORINFO_HELP_FUNC -> llvm::Function*` information in the compilation context (and use it when emitting ordinary calls too).
3) We emit bulky LLVM IR with a lot of no-op `inttoptr`s and `ptrtoint`s. These bloat things and make writing and reading of bitcode slower.

We should also create a tool (likely based on PIN) to reliably track TP regressions and improvements (just for the Jit for now).

EgorBo:
> but I cannot find how is the "never" result propagated back into the image in CG2.

isn't via `comp->setMethodAttribs(m_Callee, CORINFO_FLG_BAD_INLINEE);` ? (in inline.cpp)

SingleAccretion:
> isn't via `comp->setMethodAttribs(m_Callee, CORINFO_FLG_BAD_INLINEE)`; ? (in inline.cpp)

https://github.com/dotnet/runtime/blob/f1332ab0d82ee0e21ca387cbd1c8a87c5dfa4906/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs#L1171-L1176

PR #2319, opened 2023-06-25
[NativeAOT-LLVM] Eliminate gratituous allocations from the LLVM object writer

SingleAccretion:
Also make it one-pass, there is no reason for it not to be.

The object writer is still pretty slow (1.5s in Release build for HelloWasm on my machine), but the rest of issues are not as easy to deal with.

Contributes to #2300.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
@@ -287,9 +287,11 @@ namespace ILCompiler.DependencyAnalysis
 
 #if DEBUG
             _module.PrintToFile(Path.ChangeExtension(_objectFilePath, ".txt"));
-#endif
             _module.Verify(LLVMVerifierFailureAction.LLVMAbortProcessAction);
+
+            _module.PrintToFile(Path.ChangeExtension(_objectFilePath, "external.txt"));
             _moduleWithExternalFunctions.Verify(LLVMVerifierFailureAction.LLVMAbortProcessAction);
+#endif
 
             string dataLlvmObjectPath = _objectFilePath;
             _module.WriteBitcodeToFile(dataLlvmObjectPath);
@@ -706,7 +708,7 @@ namespace ILCompiler.DependencyAnalysis
 
         private void GetCodeForExternMethodAccessor(ExternMethodAccessorNode node)
         {
-            // TODO: use the Utf8 string directly here.
+            // TODO-LLVM: use the Utf8 string directly here.
             string externFuncName = node.ExternMethodName.ToString();
             LLVMTypeRef externFuncType;
 
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
@@ -6,11 +6,13 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
+using System.Text;
 
 using ILCompiler.DependencyAnalysisFramework;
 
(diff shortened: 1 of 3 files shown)


src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -39,18 +41,15 @@ internal sealed unsafe class LLVMObjectWriter
...
-        // Data to be emitted as LLVM bitcode after all of the object nodes have been processed.
-        private readonly List<ObjectNodeDataEmission> _dataToFill = new();
+        // Used for writing mangled names together with Utf8Name.
+        private readonly Utf8StringBuilder _utf8StringBuilder = new Utf8StringBuilder();
 
         // List of global values to be kept alive via @llvm.used.
         private readonly List<LLVMValueRef> _keepAliveList = new();
 
+        // Data emitted for the current object node. Initial capacity chosen to be the size of the largest node in a small program.
+        private LLVMValueRef[] _currentObjectData = new LLVMValueRef[100_000];
```
SingleAccretion, line 51: This largest object node was `__embedded_metadata` aka `MetadataNode` aka reflection info.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -377,11 +309,14 @@ private void EmitKeepAliveList()
         {
             // See https://llvm.org/docs/LangRef.html#the-llvm-used-global-variable.
             ReadOnlySpan<LLVMValueRef> llvmUsedSymbols = CollectionsMarshal.AsSpan(_keepAliveList);
-            LLVMTypeRef llvmUsedType = LLVMTypeRef.CreateArray(_ptrType, (uint)llvmUsedSymbols.Length);
-            LLVMValueRef llvmUsedGlobal = _module.AddGlobal(llvmUsedType, "llvm.used");
-            llvmUsedGlobal.Linkage = LLVMLinkage.LLVMAppendingLinkage;
-            llvmUsedGlobal.Section = "llvm.metadata";
-            llvmUsedGlobal.Initializer = LLVMValueRef.CreateConstArray(_ptrType, llvmUsedSymbols);
+            if (llvmUsedSymbols.Length != 0)
```
SingleAccretion, line 312: Not a correctness fix per se, since we always have something to keep alive, but it turns out a zero-sized `@llvm.used` is illegal.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -182,184 +155,143 @@ public static void EmitObject(string objectFilePath, IEnumerable<DependencyNode>
...
-            }
-
             EmitKeepAliveList();
 
 #if DEBUG
             _module.PrintToFile(Path.ChangeExtension(_objectFilePath, ".txt"));
-#endif
             _module.Verify(LLVMVerifierFailureAction.LLVMAbortProcessAction);
+
+            _module.PrintToFile(Path.ChangeExtension(_objectFilePath, "external.txt"));
```
jkotas, line 292: Is it intentional to change extension to `external.txt` here? `external.txt` does not look like an extension.

SingleAccretion: The intent is to transform `<path>/Name.bc` (`_objectFilePath`) into `<path>/Name.external.txt` (this pattern is used in a couple other places). Using `Path.ChangeExtension` is a little bit of an abuse, but the alternative is open-coding it.

jkotas (approved):
Thank you!

PR #2323, opened 2023-06-28
[NativeAOT-LLVM] Don't hook into LLVM error reporting in the Jit

SingleAccretion:
It overrides the CLR's, effectively swallowing what would otherwise be uncaught managed exceptions. This can turn simple ILC failures into cryptic errors like truncated LLVM output.

@dotnet/nativeaot-llvm


yowl (approved):
Thanks

PR #2325, opened 2023-06-30
ManagedZLib Experiment

VivianaDuenas:
This is intended to be the foundation for adding the new managed version, of the ZLib API, on top.

I already changed a couple of things for this to be aligned with a more managed approach: Mainly took off the pointers and used Span, Memory or byte arrays (in the meantime) instead. Added comments on doubts/annotations and things that need to be revised or re-evaluated later on, when having more idea on how exactly is going to be the new design.


carlossanlop:
>[First iteration of Inflate with first unit test passing](https://github.com/dotnet/runtimelab/pull/2325/commits/18431afbc0cc9d8a3b1838cbe2cfeeda45e12ef9)

Congratulations on having your first unit test passing! cc @jeffhandley

Microsoft.ManagedZlib.sln
```diff
@@ -1,11 +1,13 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26124.0
+# Visual Studio Version 17
+VisualStudioVersion = 17.6.33801.468
 MinimumVisualStudioVersion = 15.0.26124.0
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Experiment", "src\Experiment\src\Experiment.csproj", "{B7977360-6671-4707-9A1C-1C29D5BE2674}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ManagedZLib", "src\Microsoft.ManagedZLib\src\Microsoft.ManagedZLib.csproj", "{B7977360-6671-4707-9A1C-1C29D5BE2674}"
```
jkoritzinsky:
```suggestion
Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ManagedZlib", "src\Microsoft.ManagedZlib\src\Microsoft.ManagedZlib.csproj", "{B7977360-6671-4707-9A1C-1C29D5BE2674}"
```

Microsoft.ManagedZlib.sln
```diff
@@ -1,11 +1,13 @@
...
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Experiment.Tests", "src\Experiment\tests\Experiment.Tests.csproj", "{CE81B6BD-CCCC-4223-9069-B28435A4A5C1}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ManagedZLib.Tests", "src\Microsoft.ManagedZLib\tests\Microsoft.ManagedZLib.Tests.csproj", "{CE81B6BD-CCCC-4223-9069-B28435A4A5C1}"
```
jkoritzinsky:
```suggestion
Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ManagedZlib.Tests", "src\Microsoft.ManagedZlib\tests\Microsoft.ManagedZlib.Tests.csproj", "{CE81B6BD-CCCC-4223-9069-B28435A4A5C1}"
```

Microsoft.ManagedZlib.sln
```diff
@@ -1,11 +1,13 @@
...
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ManagedZLib.Benchmarks", "src\Microsoft.ManagedZLib\benchmarks\Microsoft.ManagedZLib.Benchmarks.csproj", "{56594674-14FE-4BCA-BFFE-F4FAFA4F2282}"
```
jkoritzinsky:
```suggestion
Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.ManagedZlib.Benchmarks", "src\Microsoft.ManagedZlib\benchmarks\Microsoft.ManagedZlib.Benchmarks.csproj", "{56594674-14FE-4BCA-BFFE-F4FAFA4F2282}"
```

src/Microsoft.ManagedZlib/benchmarks/Microsoft.ManagedZlib.Benchmarks.csproj
```diff
@@ -0,0 +1,15 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="0.13.5" />
+	<ProjectReference Include="..\src\Microsoft.ManagedZLib.csproj" />
```
jkoritzinsky:
```suggestion
    <PackageReference Include="BenchmarkDotNet" Version="0.13.5" />
    <ProjectReference Include="..\src\Microsoft.ManagedZlib.csproj" />
```

jkoritzinsky (commented):
It looks like there's some case sensitivity issues in the project files that's causing CI to fail on Linux/Mac. These suggestions should be enough to get the tests running there.

eng/Version.Details.xml
```diff
@@ -3,10 +3,6 @@
   <ProductDependencies>
     <!-- This VS redist package is used as a way to get a non-shipping version number
          when using a custom runtime version.  -->
```
VivianaDuenas:
```suggestion
```

Directory.Build.props
```diff
@@ -12,7 +12,7 @@
 
     <!-- Set this property to false if you don't want to use the runtime
           framework version defined in Versions.props -->
-    <UseCustomRuntimeVersion>true</UseCustomRuntimeVersion>
+    <UseCustomRuntimeVersion>false</UseCustomRuntimeVersion>
```
carlossanlop: @jkoritzinsky I noticed this was a suggestion in the readme when branching from standalone. Maybe this should be set by default to `false`, and then we just add an explanation of why.

eng/Versions.props
```diff
@@ -7,7 +7,7 @@
     <MajorVersion>1</MajorVersion>
     <MinorVersion>0</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <PreReleaseVersionLabel>alpha</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>mzlib</PreReleaseVersionLabel>
```
carlossanlop: @jkoritzinsky does using alpha clash with something from the official runtime?

src/Microsoft.ManagedZlib/benchmarks/CompressFiles.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
+using System.IO;
+using System.IO.Compression;
+using System.Text;
+
+namespace Microsoft.ManagedZLib.Benchmarks
+{
+    class CompressFiles
+    {
+
```
carlossanlop: Remove this space
```suggestion
```

src/Microsoft.ManagedZlib/benchmarks/CompressFiles.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
+using System.IO;
+using System.IO.Compression;
+using System.Text;
+
+namespace Microsoft.ManagedZLib.Benchmarks
```
carlossanlop: Consider converting to file-scoped namespaces in all your cs files. Saves you a bunch of tabs.

I think we can also make this a mandatory roslyn analyzer rule. Let me look for the rule ID and I'll give you the instructions.

src/Microsoft.ManagedZlib/benchmarks/CompressFiles.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    class CompressFiles
+    {
+        
+        private string zipName = @".\resultZip.zip";
```
carlossanlop: IDE0044: Add readonly modifier:
```suggestion
        private readonly string zipName = @".\resultZip.zip";
```

src/Microsoft.ManagedZlib/benchmarks/CompressFiles.cs
```diff
@@ -0,0 +1,76 @@
...
+        public void MakeFile(string filename, string message)
+        {
+            //Crea o appendea al final del archivo
+            FileStream fs = new FileStream(filename, FileMode.Create, FileAccess.ReadWrite);
```
carlossanlop: IDE0090: 'new' expression can be simplified:

```suggestion
            FileStream fs = new(filename, FileMode.Create, FileAccess.ReadWrite);
```

src/Microsoft.ManagedZlib/benchmarks/CompressFiles.cs
```diff
@@ -0,0 +1,76 @@
...
+
+namespace Microsoft.ManagedZLib.Benchmarks
+{
+    class CompressFiles
+    {
+        
+        private string zipName = @".\resultZip.zip";
+        public void MakeFile(string filename, string message)
+        {
+            //Crea o appendea al final del archivo
```
carlossanlop: Please translate to english the comments you wish to keep. Otherwise, please remove them.

src/Microsoft.ManagedZlib/benchmarks/CompressFiles.cs
```diff
@@ -0,0 +1,76 @@
...
+            FileStream fs = new FileStream(filename, FileMode.Create, FileAccess.ReadWrite);
+
+            if (fs.CanWrite)
+            {
+                //byte[] buffer = Encoding.Default.GetBytes("Hello World");
```
carlossanlop: Remove commented code.

src/Microsoft.ManagedZlib/benchmarks/CompressFiles.cs
```diff
@@ -0,0 +1,76 @@
...
+            //     File.Delete(zipName);
+            // }
+            ZipFile.CreateFromDirectory(directoryPath,zipName);
+        }
+        public void UnzipIt(string ZipFilePath, string ExtractDestination)
+        {
+
+            ZipFile.CreateFromDirectory(ZipFilePath, ExtractDestination);
+        }
+        /*
```
carlossanlop: Remove commented code.

src/Microsoft.ManagedZlib/benchmarks/CompressFiles.cs
```diff
@@ -0,0 +1,76 @@
...
+                byte[] buffer = Encoding.Default.GetBytes(message);
+                fs.Write(buffer,0,buffer.Length) ; //Desde donde hasta donde 
+                //fs.Write(buffer);
+            }
+            fs.Flush();
+            fs.Close();
+        }
+        public void ZipIt (string directoryPath)
+        {
+            //Si existe, llora
```
carlossanlop: Remove commented code.

src/Microsoft.ManagedZlib/benchmarks/ManagedZlibBenchmarks.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Microsoft.ManagedZLib.Benchmarks
+{
+    // Referring the dotnet/performance documentation (performance/docs/microbenchmark-design-guidelines.md)
+    /* 
+    BenchmarkDotNet creates a type which derives from type with benchmarks. 
+    So the type with benchmarks must not be sealed and it can NOT BE STATIC 
+    and it has to BE PUBLIC. It also has to be a class (no structs support).
+    */
+    public class ManagedZLibBenchmark{
```
carlossanlop:
```suggestion
    public class ManagedZLibBenchmark
    {
```

src/Microsoft.ManagedZlib/benchmarks/ManagedZlibBenchmarks.cs
```diff
@@ -0,0 +1,64 @@
+﻿using BenchmarkDotNet.Attributes;
+using System.IO; //para path
```
carlossanlop:
```suggestion
using System.IO;
```

src/Microsoft.ManagedZlib/benchmarks/ManagedZlibBenchmarks.cs
```diff
@@ -0,0 +1,64 @@
+﻿using BenchmarkDotNet.Attributes;
```
carlossanlop: I think we should add the license header to all your files. There's a roslyn analyzer for that which we can enable it. I'll look for it and will give you the suggestion.

src/Microsoft.ManagedZlib/benchmarks/ManagedZlibBenchmarks.cs
```diff
@@ -0,0 +1,64 @@
...
+    // Referring the dotnet/performance documentation (performance/docs/microbenchmark-design-guidelines.md)
+    /* 
+    BenchmarkDotNet creates a type which derives from type with benchmarks. 
+    So the type with benchmarks must not be sealed and it can NOT BE STATIC 
+    and it has to BE PUBLIC. It also has to be a class (no structs support).
+    */
+    public class ManagedZLibBenchmark{
+
+       
+        //private string message = "prueba Benchmarks";
```
carlossanlop: Remove commented code.

src/Microsoft.ManagedZlib/benchmarks/ManagedZlibBenchmarks.cs
```diff
@@ -0,0 +1,64 @@
...
+        [GlobalCleanup]
+        public void Cleanup() => Directory.Delete(_rootDirPath, recursive: true);
+
+        [Benchmark]
+        public void zipProccess() { 
+            //Sigue checando si el arhcivo existe dentro de ZipIt pero porque 
+            //Cuando lo puse en el GlobalSetUp, tiro una excepcion de que el zip ya
+            //existiaaaaa
+            //Como si no lo hubiera checado y en ese caso eliminado en el globalsetup
+            compressObj.ZipIt(_testDirPath);
```
carlossanlop: The `CompressedFiles` class is unnecessary. You could move all its logic to this one.

src/Microsoft.ManagedZlib/src/DeflateStream.cs
```diff
@@ -0,0 +1,1078 @@
...
+namespace Microsoft.ManagedZLib;
+
+public partial class DeflateStream : Stream
+{
+    private const int DefaultBufferSize = 8192;
+    private Stream _stream;
+    private Inflater? _inflater;
+    private Deflater? _deflater;
+    private byte[]? _buffer;
+    private int _activeAsyncOperation; // 1 == true, 0 == false
```
carlossanlop: Consider making this a boolean instead, and change the logic that consumes this variable so that it takes a boolean too.

src/Microsoft.ManagedZlib/src/DeflateStream.cs
```diff
@@ -0,0 +1,1078 @@
...
+    public override long Length
+    {
+        get { throw new NotSupportedException("NotSupported - This operation is not supported."); }
+    }
+
+    public override long Position
+    {
+        get { throw new NotSupportedException("NotSupported - This operation is not supported."); }
+        set { throw new NotSupportedException("NotSupported - This operation is not supported."); }
+    }
```
carlossanlop: 1. The strings are redundant. The exception class name is sufficiently descriptive.
2. IDE0027: Use expression body for accessors.

```suggestion
    public override long Length => throw new NotSupportedException();

    public override long Position
    {
        get => throw new NotSupportedException();
        set => throw new NotSupportedException();
    }
```

src/Microsoft.ManagedZlib/src/DeflateStream.cs
```diff
@@ -0,0 +1,1078 @@
...
+
+    public override long Seek(long offset, SeekOrigin origin)
+    {
+        throw new NotSupportedException("NotSupported - This operation is not supported.");
+    }
+
+    public override void SetLength(long value)
+    {
+        throw new NotSupportedException("NotSupported - This operation is not supported.");
+    }
```
carlossanlop:
```suggestion
    public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();

    public override void SetLength(long value) => throw new NotSupportedException();
```

src/Microsoft.ManagedZlib/src/Deflater.cs
```diff
@@ -0,0 +1,273 @@
...
+
+using ZErrorCode = Microsoft.ManagedZLib.ManagedZLib.ErrorCode;
+using ZFlushCode = Microsoft.ManagedZLib.ManagedZLib.FlushCode;
+
+namespace Microsoft.ManagedZLib
+{
+    /// <summary>
+    /// Provides a wrapper around the ZLib compression API.
+    /// </summary>
+    internal sealed class Deflater : IDisposable
```
carlossanlop: @jkoritzinsky Vivi mentioned that you suggested to keep `IDisposable`. For my own understanding, can you please elaborate why?

src/Microsoft.ManagedZlib/src/Deflater.cs
```diff
@@ -0,0 +1,273 @@
...
+                                                             windowBits, memLevel, strategy);
+            }
+            catch (Exception cause)
+            {
+                throw new ZLibException("ZLibErrorDLLLoadError - The underlying compression routine could not be loaded correctly.", cause);
+            }
+
+            switch (errC)
+            {
+                case ZErrorCode.Ok:
```
carlossanlop: As @jkoritzinsky suggested already, you could remove all the ZErrorCode usages and simplify the code by just adding asserts. The error codes were used because they came from native code and we needed to capture them and throw something different depending on the value. Not anymore.

src/Microsoft.ManagedZlib/src/Deflater.cs
```diff
@@ -0,0 +1,273 @@
...
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (!_isDisposed)
+            {
+                if (disposing) {
+                    //_zlibStream.Dispose(); //Vivi's note: DISPOSE  OF ZSTREAM *dispose method not yet imlemented
```
carlossanlop: Are we still going to need to dispose ZlibStream? It seems you might be able to find a way to remove that class altogether, so probably this Dispose method could end up being empty.

src/Microsoft.ManagedZlib/src/IHuffmanTree.cs
```diff
@@ -0,0 +1,325 @@
...
+                nextCode[len]++;
+            }
+        }
+        return code;
+    }
+
+    private void CreateTable()
+    {
+        uint[] codeArray = CalculateHuffmanCode();
+#if DEBUG
```
carlossanlop: Was this a translation from native code? Do we still need this?

src/Microsoft.ManagedZlib/src/Inflater.cs
```diff
@@ -0,0 +1,865 @@
...
+    private readonly OutputBuffer _output;
+    private readonly InputBuffer _input;
+
+    private IHuffmanTree? _literalLengthTree;// Literals
+    private IHuffmanTree? _distanceTree;     // Distance
+    private IHuffmanTree? _codeLengthTree;   // Length
+
+    private int _literalLengthCodeCount; // Literals
+    private int _distanceCodeCount;      // Distance
+    private int _codeLengthCodeCount;    // Length
```
carlossanlop: The variable names are sufficiently descriptive. You can remove these comments.

```suggestion
    private IHuffmanTree? _literalLengthTree;
    private IHuffmanTree? _distanceTree;
    private IHuffmanTree? _codeLengthTree;

    private int _literalLengthCodeCount;
    private int _distanceCodeCount;
    private int _codeLengthCodeCount;
```

src/Microsoft.ManagedZlib/src/Inflater.cs
```diff
@@ -0,0 +1,865 @@
...
+
+namespace Microsoft.ManagedZLib;
+
+
+/// <summary>
+/// Provides a wrapper around the ZLib decompression API.
+/// </summary>
+internal sealed class Inflater
+{
+    // ------- Buffers
```
carlossanlop: The variable names plus the space are sufficiently descriptive of what these two variables belong to. You can remove this comment.

```suggestion
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "asyoulik.txt") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "cp.html") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "fields.c") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "grammar.lsp") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "kennedy.xls") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "lcet10.txt") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "plrabn12.txt") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "ptt5") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "sum") };
+            //yield return new object[] { Path.Combine("UncompressedTestFiles", "xargs.1") };
```
carlossanlop: As I mentioned yesterday, you might have an easier way debugging if you use `InlineData` instead o `MemberData` in your unit test. That way, in VS, you can choose an individual data item to use, which you cannot do when using `MemberData` (it runs all).

carlossanlop (commented):
Left some comments, @VivianaDuenas  for you to consider. I have reviewed 2/3 of the files.

eng/Versions.props
jkoritzinsky: If dotnet/runtime and dotnet/runtimelab are both alpha at the same time, they could end up building the same package versions. If we were to assign said runtimelab branch to publish to a NuGet feed, then we'd have two packages with the same version with different contents. Changing the `PreReleaseVersionLabel` here is the right thing to do.

Directory.Build.props
jkoritzinsky: I think either `true` or `false` could be a good default here. `true` is good if you're starting an experiment during a release or are upstreaming features during the experiment (instead of only at the end). `false` is good if you're only going to build on released versions of .NET. I don't care what the default is.

src/Microsoft.ManagedZlib/src/Deflater.cs
jkoritzinsky: I suggested to keep `IDisposable` in case we decide to use `ArrayPool` for the `InputBuffer` and `OutputBuffer` types that are used in the `Inflater`, at which point we'd want to be able to dispose them (which would eventually propagate up to here). I didn't realize that the `IDisposable` was only on the `Deflater`.

sbomer:
The build is failing on linux because of inconsistencies in casing - I'd suggest renaming all files to use `ManagedZLib` instead of `ManagedZlib`, to follow the convention of existing .NET APIs.

src/Microsoft.ManagedZlib/src/Deflater.cs
VivianaDuenas: It was on both, but I didn't end up using `ArrayPool` in `Inflater`. `Deflater` is still in progress.

src/Microsoft.ManagedZlib/src/Deflater.cs
VivianaDuenas: Yes, so far, ZlibStream seems not to be useful anymore.

src/Microsoft.ManagedZlib/src/IHuffmanTree.cs
VivianaDuenas: This is used for decoding dynamic block headers (format of compression with Dynamic Huffman codes) in Inflate.

src/Microsoft.ManagedZlib/tests/StreamHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+public static partial class StreamHelpers
+{
+    public static async Task<MemoryStream> CreateTempCopyStream(string path)
+    {
+        var bytes = File.ReadAllBytes(path);
```
carlossanlop: nit: We prefer to avoid `var`, especially when invoking a method or property.

src/Microsoft.ManagedZlib/tests/StreamHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        var ms = new MemoryStream();
```
carlossanlop:
```suggestion
        MemoryStream ms = new();
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+
+namespace Microsoft.ManagedZLib.Tests;
+
+public class MyClassTests
+{
+    public Stream CreateStream(Stream stream, CompressionMode mode) => new DeflateStream(stream, mode);//For test1
+    public Stream CreateStream(Stream stream, CompressionMode mode, bool leaveOpen) => new DeflateStream(stream, mode, leaveOpen);
+    public Stream CreateStream(Stream stream, CompressionLevel level) => new DeflateStream(stream, level);
+    public Stream CreateStream(Stream stream, CompressionLevel level, bool leaveOpen) => new DeflateStream(stream, level, leaveOpen);
+    public Stream BaseStream(Stream stream) => ((DeflateStream)stream).BaseStream;
```
carlossanlop: We don't need these helper methods if we dedicate this test class exclusively to deflate stream tests.
I also didn't see any references to these methods (except in the commented method) so it's safe to remove them all.
```suggestion
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+
+    //    byte[] uncompressedStreamBytes = uncompressedStream.ToArray();
+    //    byte[] decompressorOutputBytes = decompressorOutput.ToArray();
+
+    //    Assert.Equal(uncompressedStreamBytes.Length, decompressorOutputBytes.Length);
+    //    for (int i = 0; i < uncompressedStreamBytes.Length; i++)
+    //    {
+    //        Assert.Equal(uncompressedStreamBytes[i], decompressorOutputBytes[i]);
+    //    }
+    //}
```
carlossanlop: Remove unused and commented code.
```suggestion
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+        using MemoryStream original = new MemoryStream();
+
+        byte[] originalBytes = new byte[] { 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41,
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41,
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41,
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41,
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41,
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41,
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41,
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41};
```
carlossanlop: This can become a class field.

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+        byte[] compressedBytes = new byte[10];
+        byte[] finalBytes = new byte[10];
+        original.Write(originalBytes);
+
+        MemoryStream originalData = new(); 
+        originalData.Write(originalBytes);
+        originalData.Position = 0;
+
+        MemoryStream compressedDestination = new();
+        using (System.IO.Compression.DeflateStream compressor = new System.IO.Compression.DeflateStream(compressedDestination, System.IO.Compression.CompressionMode.Compress, leaveOpen: true))
```
carlossanlop: This is wrong, we shouldn't be testing anything from System.IO.Compression. What this code currently doing is testing our official implementation, not yours.

Edit: Nevermind, you're using it only to compress, and your code to decompress. This is ok.

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+
+        byte[] originalBytes = new byte[] { 0x4C, 0x4C, 0x4F, 0x52, 0x41 }; //LLORA
+        byte[] compressedBytes = new byte[10];
+        byte[] finalBytes = new byte[10];
+        fs_originalData.Write(originalBytes); //ORIGINAL - filestream
+        //originalData.Write(originalBytes); //ORIGINAL
+        //originalData.Position = 0;
+        fs_originalData.Position = 0;
+
+        using (System.IO.Compression.DeflateStream compressor = new System.IO.Compression.DeflateStream(compressedDestination, System.IO.Compression.CompressionMode.Compress, leaveOpen: true))
```
carlossanlop, line 159:
```suggestion
        using (DeflateStream compressor = new DeflateStream(compressedDestination, CompressionMode.Compress, leaveOpen: true))
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+            Assert.Equal(uncompressedStreamBytes[i], decompressorOutputBytes[i]);
+        }
+    }
+
+    [Fact]
+    public void Test3()
+    {
+        using MemoryStream originalData = new MemoryStream(); //Original - non compressed
+        MemoryStream compressedDestination = new(); // Compressed with System.IO.Compresion - Stored here
+        //System.IO.Compression.DeflateStream compressor = new System.IO.Compression.DeflateStream(compressedDestination, System.IO.Compression.CompressionMode.Compress, leaveOpen: true);
```
carlossanlop, line 201:
```suggestion
        //DeflateStream compressor = new DeflateStream(compressedDestination, CompressionMode.Compress, leaveOpen: true);
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+
+        byte[] originalBytes = new byte[] { 0x4C, 0x4C, 0x4F, 0x52, 0x41 }; //LLORA
+        originalData.Write(originalBytes); //ORIGINAL
+        originalData.Position = 0;
+
+        using (System.IO.Compression.DeflateStream compressor = new System.IO.Compression.DeflateStream(compressedDestination, System.IO.Compression.CompressionMode.Compress, leaveOpen: true))
```
carlossanlop, line 207:
```suggestion
        using (DeflateStream compressor = new DeflateStream(compressedDestination, CompressionMode.Compress, leaveOpen: true))
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+    public void Test4() // Test3 with a filestream
+    {
+        Directory.CreateDirectory(_testDirPath2); // Creates all segments: root/inputdir/testdir
+        FileStream fs_originalData = File.Create(_testFilePath2);
+
+        //File.Create(_testFilePath).Dispose();
+        //File.Delete(_testDirPath);
+        using MemoryStream originalData = new MemoryStream(); //Original - non compressed
+        MemoryStream compressedDestination = new(); // Compressed with System.IO.Compresion - Stored here
+        //System.IO.Compression.DeflateStream compressor = new System.IO.Compression.DeflateStream(compressedDestination, System.IO.Compression.CompressionMode.Compress, leaveOpen: true);
```
carlossanlop, line 268:
```suggestion
        //DeflateStream compressor = new DeflateStream(compressedDestination, CompressionMode.Compress, leaveOpen: true);
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+
+        byte[] originalBytes = new byte[] { 0x4C, 0x4C, 0x4F, 0x52, 0x41 }; //LLORA
+        fs_originalData.Write(originalBytes); //ORIGINAL - filestream
+        originalData.Write(originalBytes); //ORIGINAL
+        originalData.Position = 0;
+        fs_originalData.Position = 0;
+
+        using (System.IO.Compression.DeflateStream compressor = new System.IO.Compression.DeflateStream(compressedDestination, System.IO.Compression.CompressionMode.Compress, leaveOpen: true))
```
carlossanlop, line 276:
```suggestion
        using (DeflateStream compressor = new DeflateStream(compressedDestination, CompressionMode.Compress, leaveOpen: true))
```

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+    //    byte[] decompressorOutputBytes = decompressorOutput.ToArray();
+
+    //    Assert.Equal(uncompressedStreamBytes.Length, decompressorOutputBytes.Length);
+    //    for (int i = 0; i < uncompressedStreamBytes.Length; i++)
+    //    {
+    //        Assert.Equal(uncompressedStreamBytes[i], decompressorOutputBytes[i]);
+    //    }
+    //}
+    [Fact]
+    public void Test1()
```
carlossanlop: Consider renaming the tests to something meaningful that explains the purpose of the test. Here and elsewhere.

src/Microsoft.ManagedZlib/tests/MyClassTests.cs
```diff
@@ -0,0 +1,339 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.ManagedZLib.Tests;
+
+public class MyClassTests
```
carlossanlop: I think this class and its filename should be renamed to `DeflateStreamTests` to clearly indicate that this is the only type were' testing.
```suggestion
public class DeflateStreamTests
```

src/Microsoft.ManagedZlib/src/Microsoft.ManagedZlib.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net7.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
+	<Nullable>enable</Nullable>
```
carlossanlop:
```suggestion
    <TargetFramework>net8.0</TargetFramework>
    <Nullable>enable</Nullable>
```

src/Microsoft.ManagedZlib/tests/LocalMemoryStream.cs
```diff
@@ -0,0 +1,40 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
carlossanlop: This whole file could be removed, it's not being used anywhere (except in the commented test, which we probably want to remove too).

src/Microsoft.ManagedZlib/src/OutputWindow.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+        else
+        {
+            // only one copy is needed if there is no wrap around.
+            copied = input.CopyTo(_window, _lastIndex, length);
+        }
+
+        _lastIndex = (_lastIndex + copied) & WindowMask; //To keep it withting the window sizw boundary
+        _bytesUsed += copied;
+        return copied; // Me imagino que esto es lo que espera el defalte stream como "bytesRead"
```
carlossanlop: Please translate the comments to english so you can get a proper review from all folks.

carlossanlop (commented):
Left some more comments.

carlossanlop:
Vivi I'll push a fix for the case-sensitivity of the files so the CI starts passing.

Edit: Actually, I can't, no permissions. I'll ask you to invite me, Jeremy and Sven as contributors of your fork.

src/Microsoft.ManagedZLib/tests/DeflateStreamTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        compressedDestination.Position = 0;
+
+        // Decompression
+        MemoryStream expectedStream = new();
+        using (DeflateStream decompressor = new DeflateStream(compressedDestination, CompressionMode.Decompress, leaveOpen: true))
+        {
+            decompressor.CopyTo(expectedStream);
+        }
+
+        using StreamReader readerOriginal = new StreamReader(originalStream);
```
sbomer: StreamReader is designed for reading text - I would suggest using the `Stream` APIs implemented by `MemoryStream` instead.

src/Microsoft.ManagedZLib/tests/DeflateStreamTests.cs
```diff
@@ -0,0 +1,96 @@
...
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41,
+                                            0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41, 0x4C, 0x4C, 0x4F, 0x52, 0x41} };
+    }
+
+    [Theory]
+    [MemberData(nameof(ByteArrayData))]
+    public void ReadSmallArrays(byte[] bytes)
+    {
+        using MemoryStream originalData = new MemoryStream();
+        originalData.Write(bytes);
```
sbomer: nit: I think you can use the MemoryStream constructor which takes a byte[] instead of calling Write and resetting the position.

src/Microsoft.ManagedZLib/tests/DeflateStreamTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    [Theory]
+    [MemberData(nameof(UncompressedTestFiles))]
+    public void ReadFile(string path)
+    {
+        using FileStream fileStream = File.OpenRead(path);
+        VerifyRead(fileStream);
+    }
+
+    private void VerifyRead (Stream originalStream)
```
sbomer:
```suggestion
    private void VerifyRead(Stream originalStream)
```

src/Microsoft.ManagedZLib/tests/DeflateStreamTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "asyoulik.txt") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "cp.html") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "fields.c") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "grammar.lsp") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "kennedy.xls") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "lcet10.txt") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "plrabn12.txt") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "ptt5") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "sum") };
+        yield return new object[] { Path.Combine("UncompressedTestFiles", "xargs.1") };
```
sbomer: nit: I'd separate out the directory name into a constant variable. You could also generate the file list by walking the directory (look at README.md inside System.IO.Compression.TestData for an example).

src/Microsoft.ManagedZLib/tests/TempDirectory.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.IO;
+
+namespace Microsoft.ManagedZLib;
+
+/// <summary>
+/// Represents a temporary directory.  Creating an instance creates a directory at the specified path,
+/// and disposing the instance deletes the directory.
+/// </summary>
+public class TempDirectory : IDisposable
```
sbomer: Please remove this unless it is used (it looks like it's unused to me)

BenchmarkDotNet.Artifacts/results/Microsoft.ManagedZLib.Benchmarks.ManagedZLibBenchmark-report-github.md
```diff
@@ -0,0 +1,21 @@
+``` ini
```
carlossanlop: These results need to be excluded from this PR. We will post the results later after doing the comparison.

src/Microsoft.ManagedZLib/tests/DeflateStreamTests.cs
carlossanlop: This helper method is used by both tests: the one that takes an array of bytes, and the one that takes a file.

The file ones were failing to pass because when comparing original vs expected as raw bytes. I suspect it has something to do with the way encoding works and how it gets compressed and then decompressed. But when we were comparing them as text strings, the tests passed.

What we could do is add an if statement here to decide if we want to compare the before and after as raw bytes or as text.

src/Microsoft.ManagedZLib/tests/DeflateStreamTests.cs
carlossanlop: Do this for all the lines:
`yield return new object[] { "<only keep the filename>" };`

The test parameters would now be only one (the filename, remove the folder name).

Then adjust the rest of the code that gets you the file path so that it also takes the "UncompressedTestFiles" argument in its internal Path.Combine.

carlossanlop: OK the current fix is good too. The method that forms the path was part of the benchmarks as you mentioned.

src/Microsoft.ManagedZLib/benchmarks/Microsoft.ManagedZLib.Benchmarks.csproj
```diff
@@ -7,9 +7,14 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="C:\Users\t-vivianadue\.nuget\packages\system.io.compression.testdata\8.0.0-beta.23326.1\contentFiles\any\any\DeflateTestData\~%24stDocument.doc" />
```
carlossanlop: This is hardcoded to your machine. Can you try setting this to a folder that would work for anyone?:

I think this should work:
```suggestion
    <Compile Remove="$(NuGetPackageRoot)\system.io.compression.testdata\8.0.0-beta.23326.1\contentFiles\any\any\DeflateTestData\~%24stDocument.doc" />
```

src/Microsoft.ManagedZLib/benchmarks/ManagedZLibBenchmarks.cs
```diff
@@ -2,23 +2,72 @@
...
+    public int DecompressManaged()
     {
+        CompressedFile!.CompressedDataStream.Position = 0;
+        MemoryStream expectedStream = new();
+        DeflateStream decompressor = new DeflateStream(CompressedFile.CompressedDataStream, CompressionMode.Decompress);
+        decompressor.CopyTo(expectedStream);
+        return 0;
     }
+
+    public class programRun
```
carlossanlop:
```suggestion
    public class ProgramRun
```

src/Microsoft.ManagedZLib/benchmarks/ManagedZLibBenchmarks.cs
```diff
@@ -2,23 +2,72 @@
...
+    {
+        static void Main()
+        {
+            BenchmarkRunner.Run<ManagedZLibBenchmark>();
+        }
```
carlossanlop:
```suggestion
        static void Main() => BenchmarkRunner.Run<ManagedZLibBenchmark>();
```

src/Microsoft.ManagedZLib/benchmarks/Microsoft.ManagedZLib.Benchmarks.csproj
```diff
@@ -7,9 +7,14 @@
     <Nullable>enable</Nullable>
```
carlossanlop, line 6: Can you remove line 6? `<ImplicitUsings>enable</ImplicitUsings>`

I'm a bit paranoid because I don't want System.IO.Compression to get autoresolved. Better to indicate it explicilty where needed.

carlossanlop: Note that you'll get build errors after removing that line, both in CompressedFile.cs and in ManagedZLibBenchmarks.cs.

src/Microsoft.ManagedZLib/benchmarks/ManagedZLibBenchmarks.cs
```diff
@@ -2,23 +2,72 @@
...
 namespace Microsoft.ManagedZLib.Benchmarks;
 
-public class ManagedZLibBenchmarks
+// Referring the dotnet/performance documentation (performance/docs/microbenchmark-design-guidelines.md)
+// BenchmarkDotNet creates a type which derives from type with benchmarks. 
+// So the type with benchmarks must not be sealed and it can NOT BE STATIC 
+// and it has to BE PUBLIC. It also has to be a class (no structs support).
+public class ManagedZLibBenchmark
 {
+    public IEnumerable<string> UncompressedTestFileNames()
```
carlossanlop:
```suggestion
    public static IEnumerable<string> UncompressedTestFileNames()
```

src/Microsoft.ManagedZLib/tests/DeflateStreamTests.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    [Theory]
+    [MemberData(nameof(UncompressedTestFiles))]
+    public void ReadFile(string path)
+    {
+        using FileStream fileStream = File.OpenRead(path);
+        VerifyRead(fileStream);
+    }
+
+    private void VerifyRead(Stream actualStream)
```
carlossanlop:
```suggestion
    private static void VerifyRead(Stream actualStream)
```

src/Microsoft.ManagedZLib/tests/TempDirectory.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.IO;
+
+namespace Microsoft.ManagedZLib;
+
+/// <summary>
+/// Represents a temporary directory.  Creating an instance creates a directory at the specified path,
+/// and disposing the instance deletes the directory.
+/// </summary>
+public class TempDirectory
```
carlossanlop: We're not using this anymore. Please remove the file.

carlossanlop (approved):
LGTM once the remaining feedback is addressed (Please merge with Squash & Merge).

src/Microsoft.ManagedZLib/benchmarks/Microsoft.ManagedZLib.Benchmarks.csproj
VivianaDuenas: It works.

PR #2327, opened 2023-06-30
[NativeAOT-LLVM] Debug break on assertion failure

SingleAccretion:
Evidently this is not the default CRT behavior, so #2323 lost this ability. Restore it manually.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks fine to me.  Thanks

PR #2333, opened 2023-07-06
[NativeAOT-LLVM] Add a portable implementation of RhGetCodeTarget

SingleAccretion:
In this implementation:
 1) The compiler generates an array of tuples { Stub, Target }. This is done for both simple and instantiating stubs.
 2) At runtime, this information is sorted, with lookups done via binary search.

Costs:
 1) The size of the array is ~5KB for the HelloWasm test, which has ~400 simple unboxing stubs and ~100 instantiating ones. For a medium-sized application that has ~50K stubs, that's 400K on 32 bit (and double that on 64 bit).
 2) Binary search is not "cheap" and, unfortunately, we will often be hit it on the startup path because of ComputeLdftnReverseLookup. Fortunately, we should rarely hit in steady state, as it will only be used for Delegate.MethodInfo and Delegate.CreateDelegate (the latter - in rare scenarios).
 3) Managed code that supports all of this clocks in at about 5K, which is rather large, but is mostly a reflection of how our codegen is not that great.

Code structure choices:
 1) ILC - layered under MetadataManager like the other maps used by reflection.
 2) Runtime - layered in CoreLib for consistency with the existing (non-portable) implementation.

Contributes to #2307.


src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/LLVMMethodCodeNode.cs
```diff
@@ -116,39 +122,24 @@ public void InitializeDebugInfo(MethodDebugInformation debugInfo)
         public void InitializeLocalTypes(TypeDesc[] localTypes)
         {
         }
-    }
-
-    internal sealed class LlvmMethodBodyNode : LLVMMethodCodeNode, IMethodBodyNode
```
SingleAccretion, line 121: I deleted the `LlvmMethodBodyNode/LLVMMethodCodeNode` fork, it did not look to serve any meaningful purpose. `MethodCodeNode` does not use such pattern either.

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/LLVMMethodCodeNode.cs
```diff
@@ -116,39 +122,24 @@ public void InitializeDebugInfo(MethodDebugInformation debugInfo)
...
-            : base(method)
-        {
-            Debug.Assert(!method.IsAbstract);
-        }
 
-        protected override string GetName(NodeFactory factory) => this.GetMangledName(factory.NameMangler);
+        public bool IsSpecialUnboxingThunk => ((CompilerTypeSystemContext)Method.Context).IsSpecialUnboxingThunk(_method);
 
-        public override IEnumerable<DependencyListEntry> GetStaticDependencies(NodeFactory factory)
+        public ISymbolNode GetUnboxingThunkTarget(NodeFactory factory)
```
SingleAccretion, line 127: `ISpecialUnboxThunkNode` implementation is C&P from `MethodCodeNode`.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/Target_Wasm/UnboxingStubTargetMappingNode.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        private static ISymbolNode GetUnboxingStubTarget(NodeFactory factory, IMethodNode stub) => stub switch
+        {
+            UnboxingStubNode simple => simple.GetUnderlyingMethodEntrypoint(factory),
+            ISpecialUnboxThunkNode { IsSpecialUnboxingThunk: true } instantiating => instantiating.GetUnboxingThunkTarget(factory),
+            _ => null
+        };
+    }
+
+    public sealed class UnboxingStubTargetMappingsNode : EmbeddedDataContainerNode
```
SingleAccretion: Arguably `UnboxingStubTargetMappingsNode` could live in its own file... I think the ILC style so far has been to pack related classes like these together, however.

SingleAccretion:
@dotnet/nativeaot-llvm

jkotas:
This may be interesting to upstream to address https://github.com/dotnet/runtime/issues/73906

src/coreclr/nativeaot/Runtime/RuntimeInstance.cpp
```diff
@@ -229,6 +237,65 @@ extern "C" bool __stdcall RegisterUnboxingStubs(PTR_VOID pvStartRange, uint32_t
...
+        return true;
+    };
+
+    MappingsBlob* allMappings = (MappingsBlob*)pMappings;
+    uint32_t allMappingsCount = mappingsSizeInBytes / sizeof(UnboxingStubTargetMapping);
+    uint32_t simpleStubCount = allMappings->SimpleUnboxingStubCount;
+    uint32_t instantiatingStubCount = allMappingsCount - simpleStubCount;
+
+    if (!initMap(m_unboxingStubTargetsMap, &allMappings->Data[0], simpleStubCount) ||
+        !initMap(m_unboxingAndInstantiatingStubTargetsMap, &allMappings->Data[simpleStubCount], instantiatingStubCount))
```
jkotas: The core runtime should not need this mapping for anything. Is that correct?

Would it make more sense to initialize this map lazily and do the whole thing in C#, like we do that for other maps that are needed by reflection?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/Target_Wasm/UnboxingStubTargetMappingNode.cs
```diff
@@ -0,0 +1,142 @@
...
+        private readonly IMethodNode _unboxingStubNode;
+
+        public UnboxingStubTargetMappingNode(IMethodNode unboxingStubNode)
+        {
+            Debug.Assert(unboxingStubNode is UnboxingStubNode or ISpecialUnboxThunkNode { IsSpecialUnboxingThunk: true });
+            _unboxingStubNode = unboxingStubNode;
+        }
+
+        public IMethodNode Stub => _unboxingStubNode;
+        public ISymbolNode GetTarget(NodeFactory factor) => GetUnboxingStubTarget(factor, Stub);
```
yowl:
```suggestion
        public ISymbolNode GetTarget(NodeFactory factory) => GetUnboxingStubTarget(factory, Stub);
```

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/Target_Wasm/UnboxingStubTargetMappingNode.cs
```diff
@@ -0,0 +1,142 @@
...
+                return stub is UnboxingStubNode ? -1 : 1;
+            }
+
+            return comparer.Compare(stub, otherStub);
+        }
+
+        public static void AddConditionalMappingDependency(ref CombinedDependencyList dependencies, NodeFactory factory, IMethodNode unboxingStubNode)
+        {
+            // The mapping is only needed if the target will actually be present in the output.
+            if (factory.UseUnboxingStubTargetMappings && GetUnboxingStubTarget(factory, unboxingStubNode) is ISymbolNode target)
```
yowl: I forget the flow of the node generation, are we guaranteed that the target dependency is added before this is called?

src/coreclr/nativeaot/Runtime/RuntimeInstance.cpp
SingleAccretion: > The core runtime should not need this mapping for anything. Is that correct?

Yes. It's only consumed at the CoreLib layer and above.

> Would it make more sense to initialize this map lazily and do the whole thing in C#, like we do that for other maps that are needed by reflection?

I considered it. The main problem is that the functionality is used by delegate equality, which seems low-level enough to be non-trivial to initialize robustly in managed code.

Additionally, the `SHash` closed hashing table implementation is very nicely fitted to this use case, a managed dictionary would use 3x+ more memory.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/Target_Wasm/UnboxingStubTargetMappingNode.cs
SingleAccretion: No - that is why conditional dependencies are used. The mapping node will only be marked if both the stub and its target get marked.

(Or, if you meant "Is `GetUnboxingStubTarget` guaranteed to return a valid node always", then yes, the factory will create a new one if needed, even if it ends up not being marked)

yowl (approved):

src/coreclr/nativeaot/Runtime/RuntimeInstance.cpp
```diff
@@ -229,6 +237,65 @@ extern "C" bool __stdcall RegisterUnboxingStubs(PTR_VOID pvStartRange, uint32_t
...
+    {
+        uint32_t SimpleUnboxingStubCount;
+        UnboxingStubTargetMapping Data[];
+    };
+
+    ASSERT(((mappingsSizeInBytes - sizeof(MappingsBlob)) % sizeof(UnboxingStubTargetMapping)) == 0);
+    auto initMap = [](UnboxingStubTargetsMap& map, UnboxingStubTargetMapping* mappings, uint32_t count) {
+        // Supporting multiple registrations would require synchronizing access to the tables.
+        ASSERT(map.GetCount() == 0);
+        map.Reallocate(count);
```
jkotas: `SHash::Reallocate` takes size, not a number of elements that the SHash is expected to hold. If we keep this, do we need to transform count into size here to avoid rehashing?

src/coreclr/nativeaot/Runtime/RuntimeInstance.cpp
jkotas: Does it need to be used by the delegate equality in the first place? It looks like that it is a left-over from some flavor of multi-module mode that allowed multiple unboxing stubs to exist for given method. It should not be a thing in current native AOT. You can try to upstream removing it from delegate equality.

Even if it still needed for some reason, delegate equality should not be used by low-level runtime code where the lazy initialization would lead to problems.

A reverse map like this one is a common problem for reflection-like functionality. I think it would be fine to implement this reverse map in a similar way. We have a few of those. For example, look for `FunctionPointersToOffsets`. (This one is also more complicated than it needs to be due to support for multimodule flavors that we do not care about anymore.)

In fact, it looks like this reverse map was implemented in managed code at some point. There are unused `UnboxingAndInstantiatingStubMap` and `struct UnboxingAndInstantiatingStubMapEntry` that one would need to implement reverse maps like this.

SingleAccretion: > Does it need to be used by the delegate equality in the first place? It looks like that it is a left-over from some flavor of multi-module mode that allowed multiple unboxing stubs to exist for given method. It should not be a thing in current native AOT. You can try to upstream removing it from delegate equality.

Doing some research, agree it looks unnecessary.

I will pursue its removal and a managed implementation for these maps.

SingleAccretion:
@dotnet/nativeaot-llvm PTAL

jkotas (approved):
Thank you!

PR #2337, opened 2023-07-07
[NativeAOT-LLVM] Stub out hybrid globalization

SingleAccretion:
We do not have the bandwidth to support it properly.

While ostensibly the upstream work could be reused, practically it is too tied to the Mono runtime internals.

Upstream ref: [#79989](https://github.com/dotnet/runtime/issues/79989). See also https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-hybrid-mode.md.

This will fix the (harmless) warnings about the missing internal calls we've been getting:
```
  ILC: Method '[S.P.CoreLib]Interop+JsGlobalization.CompareString(string&,char*,int32,char*,int32,CompareOptions,int32&,object&)' will always throw because: Invalid IL or CLR metadata in 'Int
  32 JsGlobalization.CompareString(System.String ByRef, Char*, Int32, Char*, Int32, System.Globalization.CompareOptions, Int32 ByRef, System.Object ByRef)'
  ILC: Method '[S.P.CoreLib]Interop+JsGlobalization.StartsWith(string&,char*,int32,char*,int32,CompareOptions,int32&,object&)' will always throw because: Invalid IL or CLR metadata in 'Boolea
  n JsGlobalization.StartsWith(System.String ByRef, Char*, Int32, Char*, Int32, System.Globalization.CompareOptions, Int32 ByRef, System.Object ByRef)'
  ILC: Method '[S.P.CoreLib]Interop+JsGlobalization.ChangeCaseInvariant(char*,int32,char*,int32,bool,int32&,object&)' will always throw because: Invalid IL or CLR metadata in 'Void JsGlobaliz
  ation.ChangeCaseInvariant(Char*, Int32, Char*, Int32, Boolean, Int32 ByRef, System.Object ByRef)'
  ILC: Method '[S.P.CoreLib]Interop+JsGlobalization.ChangeCase(string&,char*,int32,char*,int32,bool,int32&,object&)' will always throw because: Invalid IL or CLR metadata in 'Void JsGlobaliza
  tion.ChangeCase(System.String ByRef, Char*, Int32, Char*, Int32, Boolean, Int32 ByRef, System.Object ByRef)'
  ILC: Method '[S.P.CoreLib]Interop+JsGlobalization.EndsWith(string&,char*,int32,char*,int32,CompareOptions,int32&,object&)' will always throw because: Invalid IL or CLR metadata in 'Boolean
  JsGlobalization.EndsWith(System.String ByRef, Char*, Int32, Char*, Int32, System.Globalization.CompareOptions, Int32 ByRef, System.Object ByRef)'
  ILC: Method '[S.P.CoreLib]Interop+JsGlobalization.IndexOf(string&,char*,int32,char*,int32,CompareOptions,bool,int32&,object&)' will always throw because: Invalid IL or CLR metadata in 'Int3
  2 JsGlobalization.IndexOf(System.String ByRef, Char*, Int32, Char*, Int32, System.Globalization.CompareOptions, Boolean, Int32 ByRef, System.Object ByRef)'
```


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas:
Would forcing GlobalizationMode.Hybrid to false using ILLinker substitutions work better?

SingleAccretion:
> Would forcing GlobalizationMode.Hybrid to false using ILLinker substitutions work better?

I checked and it does not because the compiler still gets to see the invalid internal calls when not eliminating dead code (empirically, with `-c Debug`).

jkotas:
It is really the case? I thought that `FeatureSwitchManager` kicks in unconditionally, independent on the optimizations.

SingleAccretion:
> It is really the case? I thought that `FeatureSwitchManager` kicks in unconditionally, independent on the optimizations.

It does indeed. There was a certain amount of operator error: I substituted the inner `Settings.Hybrid` value, which did not work in Debug, but did in Release. Substituting `GlobalizationMode.Hybrid` works fine both ways. Have updated the change accordingly.

jkotas (approved):
Thank you!

PR #2340, opened 2023-07-13
[NativeAOT-LLVM] GCDesc computation fix

SingleAccretion:
Please refer to the commit message for the description of this change.

The change is intended for upstreaming (naturally).

Contributes to #2307.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/EETypeCreator.cs
+++ b/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/EETypeCreator.cs
@@ -526,17 +526,17 @@ namespace Internal.Runtime.TypeLoader
             int numSeries = 0;
             int i = 0;
 
-            bool first = true;
+            int first = -1;
             int last = 0;
             short numPtrs = 0;
             while (i < bitfield.Count)
             {
                 if (bitfield[i])
                 {
-                    if (first)
+                    if (first == -1)
                     {
-                        baseOffset += i;
-                        first = false;
+                        first = i;
+                        baseOffset += first;
                     }
                     else if (gcdesc != null)
                     {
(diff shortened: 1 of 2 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas:
Regression introduced by https://github.com/dotnet/runtime/pull/86877. @VSadov Could you please review this change? 

I assume that we will want to upstream this fix to dotnet/runtime.

src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/EETypeCreator.cs
```diff
@@ -565,7 +565,7 @@ private static unsafe int CreateArrayGCDesc(LowLevelList<bool> bitfield, int ran
             {
                 if (numSeries > 0)
                 {
-                    *ptr-- = (short)((bitfield.Count - last) * IntPtr.Size);
+                    *ptr-- = (short)((first + bitfield.Count - last) * IntPtr.Size);
```
VSadov, line 568: Right. The skip after the last pointer in an array element is not just to the end of the element, it is to the first pointer in the next element. 

This will matter when the first pointer in the element type is not at 0. It is rare, since the auto layout puts pointers to the start of the struct, but definitely possible - when layout is not auto or when structs are nesting.

This is a very good catch!! Needs to be upstreamed.

VSadov (approved):
LGTM!

src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/EETypeCreator.cs
VSadov: Interestingly, native `CreateArrayMethodTable` also adds `pElemMT->GetBaseSize() - OBJECT_BASESIZE`, although I do not think a struct can have a base size that is not == `OBJECT_BASESIZE`.

https://github.com/dotnet/runtime/blob/9915e9825edf174df73bc6db9ab59f167e92ab3c/src/coreclr/vm/array.cpp#L623-L624

PR #2341, opened 2023-07-15
[NativeAOT-LLVM] Implement exception dispatch info support

SingleAccretion:
This is used by the task infrastructure to capture stack traces; without this we lose stack traces for exceptions captured in tasks.

The support is a bit best-effort because of the difference in how stack traces are captured, but it did work well for the cases I was debugging.

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Exception.NativeAot.LLVM.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Exception.NativeAot.LLVM.cs
@@ -2,37 +2,87 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Diagnostics;
-using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
+using System.Threading;
 
 namespace System
 {
     public partial class Exception
     {
-        internal static void DispatchExLLVM(object exception)
-        {
-            AppendExceptionStackFrameLLVM(exception, new StackTrace(1).ToString());
-            //RhpThrowEx(exception); can't as not handling the transition unmanaged->managed in the landing pads.
-        }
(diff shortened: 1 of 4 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Exception.NativeAot.LLVM.cs
```diff
@@ -2,38 +2,88 @@
...
+                // 1. Don't clear if we're rethrowing with `throw;`.
+                // 2. Don't clear if we're throwing through ExceptionDispatchInfo.
+                //    This is done through invoking RestoreDispatchState which sets "_dispatchStateRestored" followed by throwing normally using `throw ex;`.
+                if (isFirstRethrowFrame || ex._dispatchStateRestored)
+                    return;
+
+                // If out of memory, avoid any calls that may allocate.  Otherwise, they may fail
+                // with another OutOfMemoryException, which may lead to infinite recursion.
+                if (ex == PreallocatedOutOfMemoryException.Instance)
+                    return;
```
SingleAccretion, line 48: Made the code look more like `AppendExceptionStackFrame` while I was here. We are now counting exceptions and guarding against OOM recursion.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks.  I guess there is no smoke test for this?

SingleAccretion:
>  I guess there is no smoke test for this?

Yes. I didn't add any because this is not really a fully functional implementation (as managed stack traces are in general NYI). There are some tests in `src/libraries/System.Runtime/tests` which exercise EDI (and which will have to be disabled once we start running them).

jkotas:
Thank you

PR #2342, opened 2023-07-15
[NativeAOT-LLVM] Fix array allocators

SingleAccretion:
There were two bugs:
1) The overflow detection wasn't working properly due to C++ rules.
   The "align up" arithmetic was being performed in 32 bit integers.
2) When we overflow on 32 bit, we are supposed to OOM, not actually
   throw an overflow exception.

Fix both issues by simply delegating to the slow path helpers when we have large (65K+ elements) arrays.

With this fix, a small libraries test suite runs to completion on my machine:
```
Finished:    System.Collections.Specialized.Tests

=== TEST EXECUTION SUMMARY ===
Total: 2132, Errors: 0, Failed: 0, Skipped: 8, Time: 24.0773544s
```

### DIFF
--- a/src/coreclr/nativeaot/Runtime/wasm/AllocFast.cpp
+++ b/src/coreclr/nativeaot/Runtime/wasm/AllocFast.cpp
@@ -100,28 +100,18 @@ COOP_PINVOKE_HELPER(Array*, RhpNewArray, (void* pShadowStack, MethodTable* pArra
         ThrowOverflowException(pShadowStack, pArrayEEType);
     }
 
-    size_t size;
 #ifndef HOST_64BIT
     // if the element count is <= 0x10000, no overflow is possible because the component size is
     // <= 0xffff, and thus the product is <= 0xffff0000, and the base size is only ~12 bytes
     if (numElements > 0x10000)
     {
-        // Perform the size computation using 64-bit integeres to detect overflow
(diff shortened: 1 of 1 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas (approved):
Thank you!

jkotas:
It would be nice to upstream these changes to https://github.com/dotnet/runtime/blob/main/src/coreclr/nativeaot/Runtime/portable.cpp

src/coreclr/nativeaot/Runtime/wasm/AllocFast.cpp
```diff
@@ -242,36 +232,26 @@ COOP_PINVOKE_HELPER(Array*, RhpNewArrayAlign8, (void* pShadowStack, MethodTable*
...
     uint8_t* result = acontext->alloc_ptr;
     int requiresAlignObject = ((uint32_t)result) & 7;
     size_t paddedSize = size;
     if (requiresAlignObject)
     {
         if (paddedSize > SIZE_MAX - 12)
         {
-            ASSERT_UNCONDITIONALLY("NYI");  // TODO: Throw overflow
+            ThrowOverflowException(pShadowStack, pArrayEEType);
         }
```
jkotas, line 255: This check is dead code now that the large element sizes are handled as early out above.

PR #2344, opened 2023-07-18
[NativeAOT-LLVM] Fix an EH bug

SingleAccretion:
The code was assuming that if a handler has no "catchret"s, the associated filter(s) will always return "true", which is not true.

With this fix, all `System.Collections` tests run to completion (in Release).

Recommended to review with the "hide whitespaces" option.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks

PR #2345, opened 2023-07-18
[NativeAOT-LLVM] Remove size overflow checks from Align8 helpers

SingleAccretion:
These would only be hit if we had an object approaching 4GB in size. C# disallows even specifying such sizes with StructLayout(Size).

The slow path in the GC does not check this condition, so I do not see the point of checking it on the fast path.

https://github.com/dotnet/runtime/blob/ee2a4574830d3a7a350d141af4a6c8f434d9082c/src/coreclr/gc/gc.cpp#L48895-L48957


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks

jkotas (approved):
Thank you

PR #2352, opened 2023-07-21
[NativeAOT-LLVM] Delete the no-longer-needed casts in `cpobj` code

SingleAccretion:
We no longer normalize structs with GC pointers to `i32`s.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Nice to be able to delete `castIfNecessary`, thanks!

PR #2363, opened 2023-07-29
NativeAOT-LLVM: write externals module to external.txt

yowl:
This PR writes the extenals module to the externals.txt file, which looked like a copy/paste typo.


yowl:
cc @dotnet/nativeaot-llvm

SingleAccretion (approved):

Issue #2359, opened 2023-07-25
NativeAOT-LLVM: runtime initialization workarounds 


yowl:
For Wasm components, there is no mechanism currently for a component's module, to have its initialization called, i.e for static c++ class constructors to run.  Wasm tool chains put these in a `__wasm_call_ctors` method, which is invoked (for libraries) via `_initialize`.  However the component model does not export `_initialize` and the wasmtime (I've not investigated other hosts) host cannot/does not invoke it.  We therefore have a problem when an exported method is called as the runtime is not initialized.  We can take advantage of the fact that Wasm memory is zero initialized to do something like 

```c++
#ifdef TARGET_WASI
static int m_initializeCalled = 0;
EXTERN_C void _initialize();
#endif

COOP_PINVOKE_HELPER(void, RhpReversePInvoke, (ReversePInvokeFrame * pFrame))
{
#ifdef TARGET_WASI
    if (!m_initializeCalled)
    {
        _initialize();
        m_initializeCalled = 1;
    }
#endif
```

As a hack for components.  It could probably be extended to cover `wasm-wasi` programs , but  as a temporary workaround until https://github.com/bytecodealliance/preview2-prototyping/issues/99 is resolved, could there be a better solution?

Thanks

SingleAccretion:
Three comments:
1) What global constructors do we rely on? The runtime initialization only needs a tiny piece of code to set the callback, the value of which is then used a sentinel. This looks fixable, but if there are others, it may be getting more challenging (e. g. does ICU have global constructors).
2) Invoking `_initialize` like that, if it doesn't have a "have I been initialized already" guard inside seems rather unsafe, considering we may run foreign code from unknown places.
3) How do other people run any useful code? Do they not use C++ at all (and other languages with global constructors)?

yowl:
1.You are right, I thought there was something, but its the runtime intitialization callback that is the blocker.  

2. This was just my first idea to make something callable as a wasm component.  In the reactor model of wasm components, only exports are callable which would all go through `RhpReversePInvoke` so it seemed ok for this single use case.  I'm fine with nothing making into the branch at the moment, but I'm not sure there is a clear path to how we can get the runtime initialised for wasi components where the first call to dotnet code, including the runtime will be the exported symbol.   

[...]

PR #2364, opened 2023-07-30
NativeAOT-LLVM: Init callback declaration

yowl:
This PR accepts the recommendation from [NativeAOT_StaticInitialization](https://github.com/dotnet/runtimelab/issues/2359#issuecomment-1651236811) and teherby removes `NativeAOT_StaticInitialization` and its initialisation.  This enables the runtime to be initialised without the use of static c++ ctors which are not invoked presently for WASI reactor components.

https://github.com/WebAssembly/wasi-libc/pull/74
https://github.com/WebAssembly/WASI/issues/13

Instead the initialisation from a WASI reactor component is done via the wasi-libc `_initialize` function.  The call to this function is done in `/Bootstrap/main.cpp` which is`#if def`ed in for WASI and `NATIVEAOT_DLL`

WASI reactor components can now be compiled and executed (although the WIT codegen is not ready yet).

This change does introduce the problem of intialisation of `CustomNativeMain` from a WASI program as it is not distnguishable from a wasi reactor compoent, and the crt1.o for the program case does not contain `_initialize`, `_intialize` is only present for the reactor version of crt1 https://github.com/WebAssembly/wasi-libc/blob/9f51a7102085ec6a6ced5778f0864c9af9f50000/libc-bottom-half/crt/crt1-reactor.c#L7-L27

To solve the problem of no `_initialize` for WASI + `NATIVEAOT_DLL` from a `CustomNativeMain` I added a `__weak` dummy function.


src/coreclr/nativeaot/Bootstrap/main.cpp
```diff
@@ -169,10 +169,17 @@ extern "C" int __managed__Main(int argc, char* argv[]);
 #else
 #define NATIVEAOT_ENTRYPOINT __managed__Startup
 extern "C" void __managed__Startup();
+#ifdef HOST_WASM
+extern "C" void _initialize();
```
jkotas, line 176: Add a comment about where `_initialize` comes from and what it does?

yowl:
Worth mentioning here is what node.js does.  

At the bottom of this document https://nodejs.org/api/wasi.html  it describes how the node.js runtime initializes WASI modules/compoents.  Ive not tested this because I have been focusing on wasmtime because it looks to be leading the way in terms of WIT support and  wasm component linking and running.  However on the surface it looks like with this change, if node.js was used as the host, then `_initialize` would be called twice resulting in failure.

src/coreclr/nativeaot/Bootstrap/main.cpp
```diff
@@ -169,10 +168,28 @@ extern "C" int __managed__Main(int argc, char* argv[]);
 #else
 #define NATIVEAOT_ENTRYPOINT __managed__Startup
 extern "C" void __managed__Startup();
+// _initialize is a function generated by the WASI SDK libc that calls the LLVM synthesized __wasm_call_ctors function for reactor components:
+// https://github.com/WebAssembly/wasi-libc/blob/9f51a7102085ec6a6ced5778f0864c9af9f50000/libc-bottom-half/crt/crt1-reactor.c#L7-L27
+// We define and call it for NATIVEAOT_DLL and TARGET_WASI to call all the global c++ static constructors.  This ensures the runtime is intialized
```
jkotas:
```suggestion
// We define and call it for NATIVEAOT_DLL and TARGET_WASI to call all the global c++ static constructors.  This ensures the runtime is initialized
```

src/coreclr/nativeaot/Bootstrap/main.cpp
```diff
@@ -169,10 +168,28 @@ extern "C" int __managed__Main(int argc, char* argv[]);
...
+// when calling into WebAssembly Component Model components
+#if defined(NATIVEAOT_DLL) && defined(TARGET_WASI)
+extern "C" void _initialize();
+
+// CustomNativeMain programs are built using the same libbootstrapperdll as NATIVEAOT_DLL but wasi-libc will not provide an _initialize implemeentation,
```
jkotas:
```suggestion
// CustomNativeMain programs are built using the same libbootstrapperdll as NATIVEAOT_DLL but wasi-libc will not provide an _initialize implementation,
```

src/coreclr/nativeaot/Bootstrap/main.cpp
```diff
@@ -169,10 +168,28 @@ extern "C" int __managed__Main(int argc, char* argv[]);
...
+// so create a dummy one here and make it weak to allow wasi-libc to provide the real implmentation for WASI reactor components.
```
jkotas:
```suggestion
// so create a dummy one here and make it weak to allow wasi-libc to provide the real implementation for WASI reactor components.
```

jkotas (approved):
This looks good to me. @SingleAccretion ?

SingleAccretion (commented):
The code changes LGTM, but I do have a question about wasmtime's behavior w.r.t. `_initialize`.

On the surface of things, https://github.com/bytecodealliance/wasmtime/pull/1565 claims to execute for it reactors. Why does it not work in this case?

yowl:
Yes, a good question, I will try to get a more definitive answer and update here, but my current understanding is that wasmtime will call `_initialize` when it is invoking a WASI reactor module, but that is not the same thing as a WASM component (components are a lower level and dont require WASI).  To invoke the c# wasm component it is passed to the instantiation of another module, as in the example https://docs.rs/wasmtime/latest/wasmtime/component/macro.bindgen.html .

PR #2371, opened 2023-08-13
[NativeAOT-LLVM] Add support for `JitDisasm`

SingleAccretion:
The usage is the same as `JitDump`, but this switch is usable in Release (upstream and now here).

### DIFF
--- a/src/coreclr/jit/llvmcodegen.cpp
+++ b/src/coreclr/jit/llvmcodegen.cpp
@@ -37,21 +37,8 @@ void Llvm::Compile()
 
     generateAuxiliaryArtifacts();
 
-#if DEBUG
-    JITDUMP("\n===================================================================================================================\n");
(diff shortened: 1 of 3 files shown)


src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -3290,3 +3310,9 @@ llvm::Intrinsic::ID Llvm::getLlvmIntrinsic(NamedIntrinsic intrinsicName) const
             return llvm::Intrinsic::not_intrinsic;
     }
 }
+
+void Llvm::displayValue(Value* value)
+{
+    // TODO-LLVM: support JitStdOutFile here.
+    value->print(llvm::outs());
+}
```
SingleAccretion: It turns out that `value->dump()` prints to `stderr` by default, which breaks the common pattern of redirecting output to a file.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
LGTM, thanks.

PR #2373, opened 2023-08-14
[NativeAOT-LLVM] Always use `memset` in `buildStoreLocalField`

SingleAccretion:
It turns out there is no benefit to complicating the code with null value stores.

No diffs in optimized code.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks!  Looks good.

PR #2374, opened 2023-08-16
[NativeAOT-LLVM] Rewrite exception handling to support nested dispatch

SingleAccretion:
See the added `nativeaot-wasm-exception-handling.md` for an explanation of how the new EH scheme works. Feedback on what to improve in it is most welcome (e. g. some parts are not detailed enough, or unclear, or some terms are not explained, etc)!

Contributes to #2169.


SingleAccretion:
@dotnet/nativeaot-llvm

docs/design/coreclr/botr/nativeaot-wasm-exception-handling.md
```diff
@@ -0,0 +1,293 @@
...
+- Abandonment. As we have seen above, exceptions can "replace" those thrown up the call stack, and this must be handled correctly.
+
+For the storage location of the dispatch information, we choose managed thread-static storage, mainly by method of exclusion:
+- We need something thread-local.
+- Thus, it is either the shadow stack, native thread-local storage or managed thread-local storage.
+- The shadow stack is unwound as the second pass progresses, and multiple exceptions can target the same catching frame, so it is difficult to make it work well in this case.
+- Between native and managed TLS, we need GC reporting, so we choose managed. This is reinforced by the desire to have dispatch code be managed.
+
+Virtual unwind stack maintainance is trickier. We observe the following:
+- `NOT_IN_TRY` frames must be unlinked "in advance" as control will not be reach their native counterparts.
```
yowl:
```suggestion
- `NOT_IN_TRY` frames must be unlinked "in advance" as control will not reach their native counterparts.
```

docs/design/coreclr/botr/nativeaot-wasm-exception-handling.md
```diff
@@ -0,0 +1,293 @@
...
+ ^         ^
+ |         |
+ |         |
+ |        [throw E0]                         ; Will cause the abandonment of E1
+ |
+[throw E1]
+```
+Indeed, even in the first example, the nested exception can itself be abandoned mid-flight via another nested throw.
+
+To correctly handle all of this, we must know when to unlink a given exception from the thread-local list of active ones. It turns out we can do so at the very end of the second pass, before transferring control to the catch handler. Consider that for an exception to not cause abandonment, its catch **must** lie below that of its predecessor's next one and that the oppossite is, crucially, also true:
```
yowl:
```suggestion
To correctly handle all of this, we must know when to unlink a given exception from the thread-local list of active ones. It turns out we can do so at the very end of the second pass, before transferring control to the catch handler. Consider that for an exception to not cause abandonment, its catch **must** lie below that of its predecessor's next one and that the opposite is, crucially, also true:
```

src/coreclr/jit/llvmlssa.cpp
```diff
@@ -501,14 +511,14 @@ class ShadowStackAllocator
...
-            // Funclets (especially filters) will be called by the dispatcher while live state still exists
-            // on shadow frames below (in the tradional sense, where stacks grow down) them. For this reason,
-            // funclets will access state from the original frame via a dedicated shadow stack pointer, and
-            // use the actual shadow stack for calls.
-            unsigned shadowStackLclNum =
-                m_llvm->CurrentBlock()->hasHndIndex() ? m_llvm->_originalShadowStackLclNum : m_llvm->_shadowStackLclNum;
-            GenTree* lclAddress =
-                m_llvm->insertShadowStackAddr(lclNode, varDsc->GetStackOffset() + lclNode->GetLclOffs(), shadowStackLclNum);
+            // Filters will be called by the first pass while live state still exists on shadow frames above (in the
+            // tradional sense, where stacks grow down) them. For this reason, filters will access state from the
```
yowl:
```suggestion
            // traditional sense, where stacks grow down) them. For this reason, filters will access state from the
```

yowl (approved):
Qudos for the extensive write up and code.  I don't pretend to have gone through the logic, but I see a good set of tests so this looks great.  Thanks!

jkotas:
Thank you for the hard work and dedication!

PR #2379, opened 2023-08-19
NativeAOT-LLVM: Update Emcc to 3.1.47 and LLVM to 17.0.4

yowl:
This PR updates our version of emcc to the latest, 3.1.47.  This includes the stack size change and the default 64KB fails in Linq expressions, so reset via a flag back to 1MB.  Also allows 0 as a valid value for `m_pStackLow`.

Emscripten 3.1.47 is using the top of tree for LLVM at the time of its release which is actually LLVM 18.


yowl:
Seem to have a problem with the latest clang so have opened https://github.com/llvm/llvm-project/issues/65428

yowl:
Cc @dotnet/nativeaot-llvm

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -66,26 +66,29 @@ For the runtime libraries:
...
-- Download the LLVM 15.0.6 source from https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.6/llvm-15.0.6.src.tar.xz
-- Extract it and create a subdirectory in the `llvm-15.0.6.src` folder. Do not create at the root of a drive.
-- Download the CMake 15.0.6 source from https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.6/cmake-15.0.6.src.tar.xz
-- Extract it to a folder called cmake at the same level as `llvm-15.0.0.src`.
+- Download the LLVM 17.0.4 source from https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.4/llvm-17.0.4.src.tar.xz
+- Extract it and create a subdirectory in the `llvm-17.0.4.src` folder. Do not create at the root of a drive.
+- Download the CMake 17.0.4 source from https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.4/cmake-17.0.4.src.tar.xz
+- Extract it to a folder called cmake at the same level as `llvm-17.0.4.src`.
+- Download the third-party 17.0.4 source from https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.4/third-party-17.0.4.src.tar.xz
+- Extract it to a folder called third-party at the same level as `llvm-17.0.4.src`.
```
SingleAccretion, line 82: Building LLVM is not getting easier, is it...

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -7,18 +7,24 @@ Invoke-WebRequest -Uri https://tukaani.org/xz/xz-5.2.5-windows.zip -OutFile xz.z
 Expand-Archive -LiteralPath xz.zip -DestinationPath .
 copy bin_i686\xz.exe . # get it in the path for tar
 
-Invoke-WebRequest -Uri https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.6/llvm-15.0.6.src.tar.xz -OutFile llvm-15.0.6.src.tar.xz
+Invoke-WebRequest -Uri https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.4/llvm-17.0.4.src.tar.xz -OutFile llvm-17.0.4.src.tar.xz
 
-./xz -d --force llvm-15.0.6.src.tar.xz
-tar -xf llvm-15.0.6.src.tar
+./xz -d --force llvm-17.0.4.src.tar.xz
```
SingleAccretion, line 12: Possible future enhancement: use a variable for the LLVM version.

src/coreclr/jit/CMakeLists.txt
```diff
@@ -692,6 +692,9 @@ endif (CLR_CMAKE_TARGET_ARCH_RISCV64)
 # Note as well that we need this "CLR_CMAKE_BUILD_LLVM_JIT", defined by the build scripts,
 # because there is no way in CMake to ask "what --target's' am I being asked to configure?".
 if (CLR_CMAKE_BUILD_LLVM_JIT)
+    # LLVM headers require C++17.
+    set(CMAKE_CXX_STANDARD 17)
```
SingleAccretion: Can this be an LLVM-Jit-target-specific property instead: `set_target_properties(${TARGETDETAILS_TARGET} PROPERTIES CXX_STANDARD 17)`?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -532,6 +535,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Include="-s TOTAL_STACK=$(IlcWasmStackSize)" />
       <CustomLinkerArg Condition="'$(WasmEnableJSBigIntIntegration)' == 'true'" Include="-s WASM_BIGINT=1" />
       <CustomLinkerArg Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'" Include="-s DISABLE_EXCEPTION_CATCHING=0" />
+      <CustomLinkerArg Include="-s GLOBAL_BASE=1024" />
+      <CustomLinkerArg Include="-s STACK_SIZE=1MB" />
```
```suggestion
```
Set above.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -517,6 +517,9 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Condition="'$(NativeDebugSymbols)' == 'true'" Include="-g3" />
       <CustomLinkerArg Condition="'$(NativeDebugSymbols)' != 'true'" Include="-Wl,--strip-all" />
       <CustomLinkerArg Condition="'$(NativeDebugSymbols)' != 'true'" Include="-Wl,--compress-relocations" />
+      <!-- see https://github.com/emscripten-core/emscripten/issues/16836 for the issue that means we have to force the linker to include these symbols before LTO -->
+      <CustomLinkerArg Condition="'$(Optimize)' != 'true'" Include="-Wl,-u,htons" />
+      <CustomLinkerArg Condition="'$(Optimize)' != 'true'" Include="-Wl,-u,ntohs" />
```
SingleAccretion: Can we not pass `-flto` as the workaround instead? I suspect the set of these symbols is application-specific, and we don't actually use LTO anyway.

src/coreclr/nativeaot/Runtime/wasm/PalRedhawkWasm.cpp
```diff
@@ -25,15 +25,15 @@
...
 //
-extern "C" unsigned char __data_end;
-extern "C" unsigned char __heap_base;
+extern "C" unsigned char __stack_low;
+extern "C" unsigned char __stack_high;
 void PalGetMaximumStackBounds_SingleThreadedWasm(void** ppStackLowOut, void** ppStackHighOut)
 {
     // See https://github.com/emscripten-core/emscripten/pull/18057 and https://reviews.llvm.org/D135910.
     // TODO-LLVM: update to use "__stack_low" and "__stack_high" when a recent enough linker becomes
     // available (which should be Emscripten 3.1.26, TBD WASI SDK).
```
SingleAccretion:
```suggestion
```
Now done.

src/coreclr/jit/CMakeLists.txt
yowl: Yes, moved it around a bit.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: Yes, looks like we can just drop that.  Thanks

SingleAccretion (approved):
Great to see this land, thank you!

PR #2380, opened 2023-08-19
Inflate improvements

VivianaDuenas:
Fixing benchmark's bug by consuming proper test files and having the most updated version of benchmarkDotNet.

### DIFF
--- a/src/Microsoft.ManagedZLib/benchmarks/ManagedZLibBenchmarks.cs
+++ b/src/Microsoft.ManagedZLib/benchmarks/ManagedZLibBenchmarks.cs
@@ -2,7 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Configs;
+using BenchmarkDotNet.Jobs;
 using BenchmarkDotNet.Running;
+using Perfolizer.Horology;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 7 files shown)


jkoritzinsky (approved):

src/Microsoft.ManagedZLib/benchmarks/ManagedZLibBenchmarks.cs
```diff
@@ -62,15 +63,25 @@ public void DecompressManaged()
     {
         CompressedFile!.CompressedDataStream.Position = 0;
         _outputStream!.Position = 0;
-
-
```
stephentoub, line 66: `decompressor` below should also be disposed, which also means it should be created with leaveOpen: true.

Same in DecompressNative above.

And same in CompressedFile's ctor.

src/Microsoft.ManagedZLib/benchmarks/ManagedZLibBenchmarks.cs
```diff
@@ -62,15 +63,25 @@ public void DecompressManaged()
...
         decompressor.CopyTo(_outputStream);
     }
 
     public class ProgramRun
     {
-        public static void Main(string[] args) => BenchmarkSwitcher.FromAssembly(typeof(ProgramRun).Assembly).Run(args);
+        public static void Main(string[] args)
+        {
+            var job = Job.Default
+                .WithWarmupCount(1) // 1 warmup is enough for our purpose
```
stephentoub, line 81: It's enough to ensure that all methods are promoted to tier 1?

VivianaDuenas: Adam Sitnik suggested the change because it's the recommended configuration from the perf repo: [(Link to source)](https://github.com/dotnet/performance/blob/main/src/harness/BenchmarkDotNet.Extensions/RecommendedConfig.cs#L34). He said the vast majority of the methods get promoted during Pilot stage.

src/Microsoft.ManagedZLib/benchmarks/ManagedZLibBenchmarks.cs
VivianaDuenas: I'll add the leaveOpen parameter, but I had the doubt: If the instantiation (there's no global variable) of the decompressor is done inside the context of the benchmark, wouldn't it be disposed anyways at the end?

stephentoub: No, there's no implicit destruction at the end of a scope as in C++. That's what `using` is for.

VivianaDuenas: I didn't want to add overhead for measuring the algorithm's work, but I think I'll do 2 different benchmark's categories for it: One with instantiation+dispose and one just with the copyTo.

PR #2386, opened 2023-08-22
[NativeAOT-LLVM] WASI fixes

### LINKED ISSUE #2384: [NativeAOT LLVM] Produce reactor when compiling OutputType Library on wasi-wasm
As per discussion in https://github.com/dotnet/runtimelab/issues/2235#issuecomment-1688374322, there are currently two issues making it hard to produce Wasm libraries with custom imports/exports:

1. `<OutputType>Library</OutputType>` should pass `-mexec-model=reactor` to the linker so that it produces a reactor / library instead of WASI executable.
2. The generated Wasm should allow to invoke `_initialize` multiple times - right now it doesn't expect it to be called before any exports.

### LINKED ISSUE #2385: [NativeAOT LLVM] Generate exports list for the linker on wasi-wasm
As per https://github.com/dotnet/runtimelab/issues/2235#issuecomment-1688298619, right now you need to pass explicit

```xml
<CustomLinkerArg Include="-Wl,--export,test:numbers/test!roundtrip-s8" />
```

to get symbols exported from Wasm on wasi-wasm target.

It needs a solution similar to https://github.com/dotnet/runtimelab/pull/2232 to automatically generate export list and pass it to the linker.

SingleAccretion:
Following up on the thread in #2235.

Fixes #2385.
Fixes #2384.

Also upgrades the "missing SDKs" handling to errors and adds a test for a WASI reactor using NodeJS.


src/coreclr/nativeaot/Bootstrap/main.cpp
```diff
@@ -181,13 +182,20 @@ __attribute__((weak)) void _initialize()
...
+struct WasiInitializationFlag { WasiInitializationFlag() { *(volatile bool*)&g_CalledInitialize = true; } };
+WasiInitializationFlag g_WasiInitializationFlag;
 #endif // TARGET_WASI
 #endif // !NATIVEAOT_DLL
 
 static int InitializeRuntime()
 {
 #if defined(NATIVEAOT_DLL) && defined(TARGET_WASI)
-    _initialize();
+    if (!g_CalledInitialize)
```
RReverser, line 195: Perhaps it would be simpler to have `static bool alreadyInitialized = false;` inside this function instead of separate struct that relies on static constructor? (assuming this function is called from inside `_initialize`)

src/tests/nativeaot/SmokeTests/SharedLibrary/SharedLibraryDriver.mjs
```diff
@@ -0,0 +1,39 @@
...
+    process.exit(3);
+
+// As long as no unmanaged exception is thrown managed class loaders were initialized successfully.
+instance.exports.EnsureManagedClassLoaders();
+
+// #if !CODEGEN_WASI - for some reason tries to create a background thread?
+// if (instance.exports.CheckSimpleGCCollect() != 100)
+//     process.exit(4);
+
+// #if !CODEGEN_WASI - enable when we support exception handling
```
RReverser, line 35: I actually tried some simple exception handling examples and somehow they worked, but in some other scenarios it didn't. I'm not sure what it depends on - LLVM optimisations?

src/coreclr/nativeaot/Bootstrap/main.cpp
SingleAccretion: > assuming this function is called from inside _initialize

It isn't - it's called by the runtime on thread attach. So we need something from inside `_initialize` itself to notify us it has already run. Explicit static constructor is then that something.

src/tests/nativeaot/SmokeTests/SharedLibrary/SharedLibraryDriver.mjs
RReverser: For example, I had something like this in my sample and it compiled and ran fine:

```cs
	[UnmanagedCallersOnly(EntryPoint = "hello")]
	public static void Hello() {
		try {
			Console.WriteLine("Hello from exception");
		} catch (Exception e) {
			Console.WriteLine(e);
		}
		Console.WriteLine("Hello from C#");
		MyMethod("lores ipsum");
	}
```

src/coreclr/nativeaot/Bootstrap/main.cpp
RReverser: Ah ok, sorry.

src/tests/nativeaot/SmokeTests/SharedLibrary/SharedLibraryDriver.mjs
SingleAccretion: What did and what did not work?

There is a dependency on `emscripten_get_callstack` inside exception handling (it's called on each `throw`), so I would _expect_ things to fail relatively deterministically.

Edit: that example doesn't `throw`, so I suppose it is expected to work.

RReverser: > Edit: that example doesn't `throw`, so I suppose it is expected to work.

Oh my god, this is ridiculous. I need sleep 😬

src/coreclr/nativeaot/Bootstrap/main.cpp
SingleAccretion: No problem, thank you for the review!

SingleAccretion:
@dotnet/nativeaot-llvm

src/tests/nativeaot/SmokeTests/SharedLibrary/SharedLibraryDriver.mjs
```diff
@@ -0,0 +1,39 @@
...
+
+if (instance.exports.ReturnsPrimitiveChar() != 97) // 'a'
+    process.exit(3);
+
+// As long as no unmanaged exception is thrown managed class loaders were initialized successfully.
+instance.exports.EnsureManagedClassLoaders();
+
+// #if !CODEGEN_WASI - for some reason tries to create a background thread?
+// if (instance.exports.CheckSimpleGCCollect() != 100)
+//     process.exit(4);
```
SingleAccretion, line 33: I have the identified the root cause for this and will PR the fix separately.

src/coreclr/nativeaot/Bootstrap/main.cpp
```diff
@@ -168,11 +168,12 @@ extern "C" int __managed__Main(int argc, char* argv[]);
 #else
 #define NATIVEAOT_ENTRYPOINT __managed__Startup
 extern "C" void __managed__Startup();
+
+#ifdef TARGET_WASI
```
yowl, line 172: Why is `NATIVEAOT_DLL` dropped here but still included in `InitializeRuntime` ?   If this is intended then I suppose the comment on 175 should be changed?

yowl (commented):
Thanks, just a question on `NATIVEAOT_DLL`

src/coreclr/nativeaot/Bootstrap/main.cpp
SingleAccretion: It's not really dropped, it just wasn't needed, since we're in the `#else` part of `#ifndef NATIVEAOT_DLL`:
```cpp
#ifndef NATIVEAOT_DLL
#define NATIVEAOT_ENTRYPOINT __managed__Main
#if defined(_WIN32)
extern "C" int __managed__Main(int argc, wchar_t* argv[]);
#else
extern "C" int __managed__Main(int argc, char* argv[]);
#endif
#else // NATIVEAOT_DLL guaranteed to be defined below
#define NATIVEAOT_ENTRYPOINT __managed__Startup
extern "C" void __managed__Startup();

#ifdef TARGET_WASI
```

yowl (approved):

src/coreclr/nativeaot/Bootstrap/main.cpp
```diff
@@ -181,13 +182,20 @@ __attribute__((weak)) void _initialize()
 {
 }
 
+// Guard the "_initialize" call so that well-behaving hosts do not get affected by this workaround.
+bool g_CalledInitialize = false;
```
jkotas:
```suggestion
static bool g_CalledInitialize = false;
```
?

jkotas (approved):
Thank you!

Issue #2269, opened 2023-05-05
NativeAOT-LLVM: Wasi tracking


yowl:
This PR is to track issues that are known to be problems with the wasi-wasm RID.

- [ ] System.Globalization (solve by building ICU for wasi-sdk presumably)  
- [ ] thread stack size, `PalGetMaximumStackBounds` / `pthread_attr_get`
- [ ] mmap/munmap - need to track the blocks ourselves, in a double linked list might be enough initially, even if prone to fragmentation and sub optimal performance
- [x] Build integration - WIP in https://github.com/dotnet/runtimelab/pull/2268

mhmd-azeez:
Hi, do you have a small sample for `wasi-wasm`? I have trouble building it:

```csproj
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <OutputType>Exe</OutputType>
    <PublishTrimmed>true</PublishTrimmed>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.ILCompiler.LLVM" Version="9.0.0-dev" />
    <PackageReference Include="runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM" Version="9.0.0-dev" />
    <PackageReference Include="runtime.browser-wasm.Microsoft.DotNet.ILCompiler.LLVM" Version="9.0.0-dev" />
    <PackageReference Include="runtime.wasi-wasm.Microsoft.DotNet.ILCompiler.LLVM" Version="9.0.0-dev" />
  </ItemGroup>
</Project>
```

```
PS D:\x\dotnet\aotwasm> dotnet publish -r wasi-wasm /p:TargetArchitecture=wasm /p:PlatformTarget=AnyCPU /p:MSBuildEnableWorkloadResolver=false --self-contained            
MSBuild version 17.8.0+6cdef4241 for .NET
  Determining projects to restore...
C:\Program Files\dotnet\sdk\8.0.100-rc.2.23502.2\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(1
48,5): error NETSDK1084: There is no application host available for the specified RuntimeIdentifier 'wasi-wasm'. [D:\x\dotnet\aotwasm\ao 
twasm.csproj]
[...]

PR #2387, opened 2023-08-23
[NativeAOT-LLVM] Support emulated unwinding for WASI

SingleAccretion:
The idea here is very simple. It rests on two premises:
1) We can "unwind" using early returns conditioned on a global "exception thrown" variable.
2) We will only need to "unwind" through managed code.

It turns out, crucially, that `2` is true: all native runtime code has the constraint that it will only tail-call managed code. On native platforms this is due to the lack of unwind information for this native code, for us it means we don't need to modify anything upstream for things to "just work". We do still need to ensure our own "manually managed" code such as the helpers generated in pure LLVM and allocators work in alignment with the tail-calling assumption.

The end result is that we now support all exception handling features except stack traces on non-WasmEH-enabled non-browser platforms with only ~200 lines of code changed.


src/tests/Common/dirs.proj
```diff
@@ -30,17 +30,9 @@
...
-      <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\MultiModule\*.csproj" />
       <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\DynamicGenerics\*.csproj" />
-      <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\Exceptions\*.csproj" />
-      <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\UseSystemResourceKeys\*.csproj" />
-      <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\FrameworkStrings\*.csproj" />
       <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\Reflection\*.csproj" />
-      <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\TrimmingBehaviors\*.csproj" />
       <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\UnitTests\*.csproj" />
-      <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\HelloWasm\ExceptionHandlingTests.Cpp.csproj" />
     </ItemGroup>
```
SingleAccretion, line 36: The remaining three tests are failing on some seemingly unrelated issue with the interface dispatch always on stack. I wonder what will it turn out to be, there is nothing WASI-specific about CID.

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvm.h
```diff
@@ -108,6 +109,7 @@ enum HelperFuncInfoFlags
     HFIF_SS_ARG = 1, // The helper has shadow stack arg.
     HFIF_VAR_ARG = 1 << 1, // The helper has a variable number of args and must be treated specially.
     HFIF_NO_RPI_OR_GC = 1 << 2, // The helper will not call (back) into managed code or trigger GC.
+    HFIF_NO_RPI_OR_GC_OR_THROW = 1 << 3, // The helper either throws or does not call into managed code / trigger GC.
```
yowl: The names of these are a little odd, but I presume they are following a convention.  For `HFIF_NO_RPI_OR_GC_OR_THROW` its !RPI || !GC || THROW, but the ! is sort of lost after the first `NO`.  HFIF_THROW_OR_NO_RPI_OR_GC` would make the THROW positive I suppose.  Not a big thing, the comment explains well enough.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -452,6 +462,7 @@ public enum TargetAbiType : byte
     public enum CorInfoLlvmEHModel
```
yowl, line 462: do these enum values warrant a one line comment or the name is enough?

yowl (commented):
This is pretty cool, thanks!

src/tests/Common/dirs.proj
SingleAccretion: Found out the issue, should have been more obvious in retrospect since it is mentioned in #2269 -- there is partial `munmap`ping going on in `PalVirtualAlloc` that is used to allocate the CID caches.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
SingleAccretion: The enum is duplicated in `llvm.h`, where there are tiny comments about the meanings. I don't think it's pointful to duplicate Jit-EE interface comment at this point in time (while the interface is still constantly evolving).

src/coreclr/jit/llvm.h
SingleAccretion: Agree this name is not the best, have changed it to your suggestion.

yowl (approved):

jkotas:
Thank you!

Issue #2357, opened 2023-07-24
[NativeAOT-LLVM] Optimization setting is not passed to the linker


SingleAccretion:
There is a bug currently in the targets that causes us to not tell the linker we're optimizing (notice the `!=`):

https://github.com/dotnet/runtimelab/blob/10183cf8c03a6c0b204cd12ce30c0e7f240f8dfa/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L510

Fixing the bug results in:
1) `wasm-opt` being run, which saves 8%+ in code size for HelloWasm, mostly from compressing the 5-byte LEBs left by the compiler in the original object files (linker has `--compress-relocations` which does the same thing, but doesn't preserve debug info).
2) Debug info being completely removed, including the `names` section.

We don't want the default `-c Release` build to end with no debug info, so we need something in the middle - ideally, with both the code size saving (for `wasmjit-diff.ps1` purposes) and debug info. We can strip the latter when `NativeDebugSymbols=false`.

SingleAccretion:
Some notes and numbers on how linking and running `wasm-opt` affects the output. All tests were done on HelloWasm as usual.

1\) Just `-g3` - this will not run Binaryen. The big issue with this mode is that _all_ relocations remain 5 bytes long (so e. g. an address of a global is 6 bytes, `i32.const <reloc>`). Baseline `Code` section size is 3087 KBytes, note this includes 700KB of ICU.
2\) `--compress-relocations` - this will tell the linker to compress relocations and invalidate debug info.
```
Total bytes of base: 3161482
Total bytes of diff: 2891113
Total bytes of delta: -270369 (-8.55% % of base)
Average relative delta: -12.95%
    diff is an improvement
    average relative diff is an improvement
```
3\) `-g3` + `-O3` - this will run Binaryen in a "handicapped" mode, because it cannot update DWARF in all cases. Importantly, no inlining will be performed. This is the only mode where we can sort of have our debug info cake and eat it. Even so, using this adds 10 seconds to the build (a 35% increase of the overall build time, including `csc` and everything else!), which will scale with larger binaries, and is not be compatible with `-gseparate-dwarf`.
```
Total bytes of base: 3161482
Total bytes of diff: 2795974
Total bytes of delta: -365508 (-11.56% % of base)
Average relative delta: -14.09%
    diff is an improvement
    average relative diff is an improvement
```
[...]

PR #2388, opened 2023-08-24
[NativeAOT-LLVM] More miscellaneous fixes

SingleAccretion:
Fixing various bugs and improving the documentation:

1) Disable background GC.
2) Add in some missed `pthread` shims.
3) Add various linker settings for stripping debug information.
4) Specify the global base and stack size.
5) Add notes about `NativeDebugSymbols` and `UnmanagedEntryPointsAssembly` to the documentation.
6) Remove `HFIF_NO_RPI_OR_GC` from the RPI helper.


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -510,6 +512,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Include="@(NativeObjects->'&quot;%(Identity)&quot;'->Replace(&quot;\&quot;, &quot;/&quot;))" />
       <CustomLinkerArg Include="-o &quot;$(NativeBinary.Replace(&quot;\&quot;, &quot;/&quot;))&quot;" />
       <CustomLinkerArg Condition="'$(NativeDebugSymbols)' == 'true'" Include="-g3" />
+      <CustomLinkerArg Condition="'$(NativeDebugSymbols)' != 'true'" Include="-Wl,--strip-all" />
+      <CustomLinkerArg Condition="'$(NativeDebugSymbols)' != 'true'" Include="-Wl,--compress-relocations" />
       <CustomLinkerArg Condition="'$(Optimize)' != 'true'" Include="$(WasmOptimizationSetting) -flto" />
```
RReverser, line 517: Random question, not related to this PR but noticed because it was nearby: why is LTO enabled and `WasmOptimizationSetting` passed to the linker only when `$(Optimize) != 'true'`? Shouldn't it be the opposite?

SingleAccretion: > why is LTO enabled

I believe that's more or less a leftover from a previous time (it used to be that the linker was passed bitcode files directly without the `.o` step). It is not used currently.

> Shouldn't it be the opposite?

It is a bug, yes, but it's not abundantly clear what the right setting actually is. See #2357.

SingleAccretion:
The GC test still fails, now with heap corruption. Will take a while to debug, most likely.

Edit: turned out to be a GC hole due to a codegen bug.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks for adding the memory layout options.

src/coreclr/gc/gcpriv.h
```diff
@@ -184,6 +184,11 @@ inline void FATAL_GC_ERROR()
 
 #define BACKGROUND_GC   //concurrent background GC (requires WRITE_WATCH)
 
+// It is no longer possible to build without BACKGROUND_GC defined, so we have to use this duplicate
```
jkotas, line 187: @cshung You have been fixing the build without `BACKGROUND_GC` defined in the past: https://github.com/dotnet/runtime/pull/77328 . Is building without `BACKGROUND_GC` defined still something you care about?

jkotas (approved):
Thank you!

src/coreclr/gc/gcpriv.h
jkotas: https://github.com/dotnet/runtime/pull/91445 should fix the build with undefined `BACKGROUND_GC`

PR #2396, opened 2023-09-06
NativeAOT-LLVM Merge sep 23

yowl:
This PR moves forward to fe88e1bc5419196ae079077b35f28818efa568fa.  There's 2 resolving of merge conflicts as we added unaligned relocs as part of this merge.

- remove `assert (use_frozen_segments_p);` and `ETW::GCLog::WalkHeap()` - upstreamed
- remove `CORINFO_TYPE_BOOL`  and add `CORINFO_HELP_THROW_ENTRYPOINT_NOT_FOUND_EXCEPTION`
- remove `StackTraces` test
- #if out `PalCreateDumpInitialize` and add stub for `PalCreateCrashDumpIfEnabled`.  Not sure I've solved this the cleanest way possible.
- disable the new test `StackTraceMetadata` for WASI.   For browser `-Wl,--strip-all` allows this to pass.
- take the `lowMemoryP ? Interop.BOOL.TRUE : Interop.BOOL.FALSE` from upstream.

cc @dotnet/nativeaot-llvm


src/coreclr/jit/CMakeLists.txt
```diff
@@ -713,8 +713,7 @@ if (CLR_CMAKE_BUILD_LLVM_JIT)
     list(FILTER JIT_SOURCES EXCLUDE REGEX regalloc\.cpp)
 
     # We'll be linking against LLVM built without /guard:ehcont, so disable it.
-    string(REPLACE "/guard:ehcont" "" CMAKE_SHARED_LINKER_FLAGS ${CMAKE_SHARED_LINKER_FLAGS})
-    string(REPLACE "/guard:ehcont" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+    set_property(GLOBAL PROPERTY CLR_EH_CONTINUATION OFF)
```
SingleAccretion: Why `GLOBAL`? I would expect you only need to turn it off for the LLVM Jit targets:
```
set_target_properties(clrjit_browser_wasm32_${ARCH_HOST_NAME} PROPERTIES CLR_EH_CONTINUATION OFF)
# I don't know if CMake errors out on no-existent targets, so this may need to be commented.
set_target_properties(clrjit_browser_wasm64_${ARCH_HOST_NAME} PROPERTIES CLR_EH_CONTINUATION OFF)
```

src/coreclr/tools/aot/ILCompiler/ILCompilerRootCommand.cs
```diff
@@ -315,89 +155,18 @@ <<<<<<< HEAD
             new("--singlemethodgenericarg") { Description = "Single method compilation: generic arguments to the method" };
         public CliOption<string> MakeReproPath { get; } =
             new("--make-repro-path") { Description = "Path where to place a repro package" };
+        public CliOption<string[]> WasmImport { get; } =
+            new("--wasmimport") { Description = "WebAssembly import module names for PInvoke functions"};
+        public CliOption<string[]> WasmImportList { get; } =
+            new("--wasmimportlist") {Description = "File with list of WebAssembly import module names for PInvoke functions"};
```
SingleAccretion, line 161: Could you manually check if these continue to work? Just to protect against changes in the meaning of the type argument...

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -517,6 +517,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Condition="'$(Optimize)' != 'true'" Include="$(WasmOptimizationSetting) -flto" />
       <CustomLinkerArg Condition="'$(IlcLlvmTarget)' != ''" Include="-target $(IlcLlvmTarget)" />
       <CustomLinkerArg Condition="'$(IlcLlvmExceptionHandlingModel)' == 'wasm'" Include="-fwasm-exceptions" />
+      <CustomLinkerArg Condition="'$(StackTraceSupport)' == 'false'" Include="-Wl,--strip-all" />
```
SingleAccretion: Hmm, this will strip all debug info, not just the stack trace data, which is not very satisfactory. Let's bundle this work into #2404 and disable the tests.

src/coreclr/nativeaot/Runtime/unix/PalCreateDump.cpp
```diff
@@ -618,3 +619,8 @@ PalCreateDumpInitialize()
 
     return true;
 }
+#else
+void PalCreateCrashDumpIfEnabled()
+{
+}
+#endif // !defined(HOST_WASM)
```
SingleAccretion, line 626: It would be cleaner to create a no-op `PalCreateDump.cpp` under the `/wasm` directory and so avoid modifying upstream code.

WASM does have the concept of a core dump, but it'll likely take years before we will be able to collect one from within WASM itself (or even JS).

src/coreclr/nativeaot/Runtime/unix/PalCreateDump.h
```diff
@@ -7,5 +7,8 @@
 
 extern bool PalCreateDumpInitialize();
 extern void PalCreateCrashDumpIfEnabled();
+
+#if !defined(HOST_WASM)
 extern void PalCreateCrashDumpIfEnabled(int signal, siginfo_t* siginfo = nullptr, void* exceptionRecord = nullptr);
+#endif // !defined(HOST_WASM)
```
SingleAccretion: I recall CI was seeing failures to the effect of undefined `siginfo_t`; you could forward-declare it to avoid the `#ifdef`.

src/coreclr/jit/CMakeLists.txt
yowl: Moved to where we actually create the target which removes the need for commenting the`wasm64`

src/coreclr/tools/aot/ILCompiler/ILCompilerRootCommand.cs
yowl: Tried both `wasmimport` and `wasmimportlist` and they look fine, correctly adding the attribute in the LLVM (I just picked some imports to play with):
```
attributes #0 = { "wasm-import-module"="SomeModule" "wasm-import-name"="GetNativeFunctionToCall" }
attributes #1 = { "wasm-import-module"="SomeModule" "wasm-import-name"="NativeIntToDouble" }
```

src/coreclr/nativeaot/Runtime/unix/PalCreateDump.cpp
```diff
@@ -13,10 +13,11 @@
 #include <unistd.h>
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
+#include <signal.h>
```
SingleAccretion, line 16: Presumably all changes in this file can now be reverted.

src/coreclr/nativeaot/Runtime/unix/PalCreateDump.h
```diff
@@ -3,9 +3,16 @@
...
+// Emscripten's headers.
+#if defined(TARGET_WASI)
+typedef struct siginfo_s siginfo_t;
+#endif
+
 #include <signal.h>
 
 extern bool PalCreateDumpInitialize();
 extern void PalCreateCrashDumpIfEnabled();
+
```
SingleAccretion:
```suggestion
```
Whitespace change.

src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
```diff
@@ -432,10 +432,12 @@ REDHAWK_PALEXPORT bool REDHAWK_PALAPI PalInit()
 
     ConfigureSignals();
 
+#ifndef HOST_WASM
```
SingleAccretion:
```suggestion
```

src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
```diff
@@ -432,10 +432,12 @@ REDHAWK_PALEXPORT bool REDHAWK_PALAPI PalInit()
...
     if (!PalCreateDumpInitialize())
     {
         return false;
     }
+#endif
```
SingleAccretion:
```suggestion
```

src/coreclr/nativeaot/Runtime/wasm/PalCreateDump.cpp
```diff
@@ -0,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// We do not support crash dumps in Wasm yet, so just stub.
+
+#include "../unix/PalCreateDump.h"
```
SingleAccretion:
```suggestion
#include "PalCreateDump.h"
```
I think that should work...

SingleAccretion (approved):
LGTM modulo remaining feedback. Thank you!

src/coreclr/nativeaot/Runtime/unix/PalCreateDump.cpp
yowl: yes, thanks

src/coreclr/nativeaot/Runtime/wasm/PalCreateDump.cpp
yowl: Yes, it does, thanks.

jkotas:
Thank you!

PR #2401, opened 2023-09-12
NativeAOT-LLVM: Emit object nodes as structs not arrays to allow for unaligned relocs.

yowl:
This PR enables relocs to be written unaligned in anticipation of merging upstream and `StackTraceMethodMappingNode` where it writes a "command" byte before the relocs.  This is achieved by using a struct for the object node instead of an array.  In many cases this struct will only contain LLVM `ptr`s and here the reading from the node is unchanged.  Reading from `StackTraceMethodMappingNode` when it is tackled will be different...

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
@@ -205,7 +205,7 @@ namespace ILCompiler.DependencyAnalysis
             {
                 if (!useStruct && nextRelocValid && nextRelocOffset % pointerSize != 0)
                 {
-                    // Switch from array to struct.
+                    // Switch from array to struct. This will need more elements because binary data is output byte-by-byte.
                     useStruct = true;
                     dataSizeInElements = nodeContents.Relocs.Length + dataSizeInBytes - (nodeContents.Relocs.Length * pointerSize);
 
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
@@ -49,6 +49,7 @@ namespace ILCompiler.DependencyAnalysis
 
         // Data emitted for the current object node. Initial capacity chosen to be the size of the largest node in a small program.
         private LLVMValueRef[] _currentObjectData = new LLVMValueRef[100_000];
+        private LLVMTypeRef[] _currentObjectTypes = new LLVMTypeRef[100_000];
 
 #if DEBUG
         private static readonly Dictionary<string, ISymbolNode> _previouslyWrittenNodeNames = new();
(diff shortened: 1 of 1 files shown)


yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -166,65 +167,66 @@ private unsafe void EmitObjectNode(ObjectNode node, ObjectData nodeContents)
...
+            int relocIndex = 0;
+            int relocLength = nodeContents.Relocs.Length;
 
-            // Emit binary data.
             ReadOnlySpan<byte> data = nodeContents.Data.AsSpan();
             for (int i = 0; i < dataElements.Length; i++)
             {
                 ref LLVMValueRef dataElementRef = ref dataElements[i];
-                if (dataElementRef == default)
+                ref LLVMTypeRef typeElementRef = ref typeElements[i];
```
SingleAccretion: Nit: this pattern of saving `ref array[i]` is a micro-optimization I used to save one bounds check, but here we don't save anything with it anymore, so it would be clearer to assign to `array[i]` directly.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -166,65 +167,66 @@ private unsafe void EmitObjectNode(ObjectNode node, ObjectData nodeContents)
...
+
+                if (relocIndex < relocLength && currOffset == nodeContents.Relocs[relocIndex].Offset)
                 {
-                    ulong value = 0;
-                    int offset = i * pointerSize;
-                    int size = Math.Min(dataSizeInBytes - offset, pointerSize);
-                    data.Slice(offset, size).CopyTo(new Span<byte>(&value, size));
+                    Relocation reloc = nodeContents.Relocs[relocIndex];
+                    typeElementRef = _ptrType;
```
SingleAccretion: Nit: move down to `dataElementRef = ...` for symmetry.

SingleAccretion (commented):
My main concern is how well does LLVM cope with large struct types. Concretely:
1) What (if any) is the time difference for creating LLVM before/after (excluding the time it takes to write the file)?
2) What (if any) is the size difference of the generated bitcode?

yowl:
With release builds:

There is no noticeable difference in time.  The last commit shows where I added the StopWatch.  Both before and after took between 313 and 370 ms 

Before:
320 to 370 ms

After:
313 to 364 ms

However there is a noticeable increase in size of the bitcode for `HelloWasm.bc` :

Before 4,108,260 bytes
After 4,783,664 bytes

image

We could have both arrays and structs, switching on the presence of anything of size % ptrsize != 0.   If we default to arrays, but switch as soon as we see the first non factor `ptrsize` element, that should be nearly as fast as what we have, seeing as the problematic symbol starts with a byte?

SingleAccretion:
Hmm, it would be great to avoid fallbacks, however, I don't have any ideas on how. So the proposed mitigation sounds good to me.

yowl:
Another option might be to write a byte array instead of an i32 array, I could try that to see if it performs ok, I would expect the size to be as now.  Do you think that is worth a try?

SingleAccretion:
How would a byte array represent relocs? With some sort of embedded shifts/`ptrtoint`s?

yowl:
Oh yes, oops, that would be complicated.

yowl:
Latest commit figures:

Times: 325 to 372 ms  unchanged basically
Size After: 4,108,588 (328 bytes larger)

yowl:
Have left the stopwatch in, incase you want to check what I've measured, but will take it out if we are happy with these figures.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -168,63 +174,122 @@ private unsafe void EmitObjectNode(ObjectNode node, ObjectData nodeContents)
...
-            // Emit relocations. We assume these are always aligned.
-            foreach (Relocation reloc in nodeContents.Relocs)
+            while (dataOffset < dataSizeInBytes)
             {
-                long delta;
-                fixed (void* location = &nodeContents.Data[reloc.Offset])
+                if (!useStruct && nextRelocValid && nextRelocOffset % pointerSize != 0)
                 {
-                    delta = Relocation.ReadValue(reloc.RelocType, location);
+                    // Switch from array to struct.
```
SingleAccretion:
```suggestion
                    // Switch from array to struct. This will need more elements because binary data is output byte-by-byte.
```
Took me a little to realize why the size recalculation is needed.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -168,63 +174,122 @@ private unsafe void EmitObjectNode(ObjectNode node, ObjectData nodeContents)
...
+                    {
+                        Array.Resize(ref _currentObjectData, Math.Max(_currentObjectData.Length * 2, dataSizeInElements));
+                        Array.Resize(ref _currentObjectTypes, Math.Max(_currentObjectTypes.Length * 2, dataSizeInElements));
+                    }
+
+                    dataElements = _currentObjectData.AsSpan(0, dataSizeInElements);
+                    typeElements = _currentObjectTypes.AsSpan(0, dataSizeInElements);
+
+                    // Restart zipping while loop.
+                    dataOffset = elementOffset = relocIndex = 0;
```
SingleAccretion: Nit: I think it's a little unconventional for runtime code style to use assignment expressions.

SingleAccretion (approved):
Numbers and code look good, thank you!

yowl:
Thanks, don't know why the previous build failed.  Some blip somewhere.

jkotas (approved):
Thanks!

PR #2402, opened 2023-09-16
[NativeAOT-LLVM] Pop fewer virtual unwind frames explicitly

SingleAccretion:
When we are unwinding past a virtual frame, we must unlink it. For top-level faults this is done via an explicit helper call in codegen and we assumed that all faults could use the virtual unwind frame of their parent function.

This change refines this so that only faults with nested catch handlers need to do this while for others, the responsibility for releasing the frame is "pushed down" to the inner handler.

This is achieved by giving top-level EH regions protected by faults `NOT_IN_TRY` unwind index instead of `NOT_IN_TRY_CATCH`.

An example of where this optimization helps is when we have the common `try { } catch { } finally { }` pattern in C#, which is represented as `try { try { } catch { } } finally { }` in IL.

Diffs are as expected:
```
Top method improvements (percentages):
         -15 (-3.42% of base) : 1004.dasm - HelloWasm_Program__TryFinallyWithCatch
         -15 (-3.40% of base) : 1002.dasm - S_P_TypeLoader_Internal_Runtime_TypeLoader_TypeLoaderEnvironment__EnsureTypeHandleForType
          -8 (-2.86% of base) : 1003.dasm - HelloWasm_Program__TryFinally
         -22 (-2.82% of base) : 1011.dasm - S_P_CoreLib_System_Reflection_DynamicInvokeInfo__InvokeWithManyArguments
          -8 (-2.65% of base) : 1012.dasm - HelloWasm_Program___TestIntraFrameFilterOrderDeep_g__InnerInnerFilterAndFinally_223_2
          -8 (-2.63% of base) : 1006.dasm - HelloWasm_Program___TestIntraFrameFilterOrderDeep_g__InnerFilterAndFinally_223_1
          -8 (-2.61% of base) : 1005.dasm - HelloWasm_Program___TestIntraFrameFilterOrderBasic_g__InnerFilterAndFinally_222_1
         -22 (-2.48% of base) : 1013.dasm - S_P_CoreLib_System_Runtime_CompilerServices_ClassConstructorRunner__EnsureClassConstructorRun
         -15 (-2.42% of base) : 1010.dasm - S_P_CoreLib_System_Globalization_CompareInfo_SortHandleCache__GetCachedSortHandle
         -10 (-1.96% of base) : 1014.dasm - S_P_CoreLib_System_Gen2GcCallback__Finalize
         -17 (-1.30% of base) : 1001.dasm - S_P_CoreLib_System_Reflection_Runtime_TypeInfos_RuntimeTypeInfo__GetPropertyImpl
          -8 (-0.67% of base) : 1009.dasm - S_P_CoreLib_System_Reflection_DynamicInvokeInfo__Invoke
          -8 (-0.35% of base) : 1007.dasm - S_P_TypeLoader_Internal_Runtime_TypeLoader_TypeLoaderEnvironment__RegisterDynamicGenericTypesAndMethods
```

### DIFF
--- a/src/coreclr/jit/llvm.h
+++ b/src/coreclr/jit/llvm.h
@@ -389,15 +389,21 @@ public:
     PhaseStatus AddVirtualUnwindFrame();
 
 private:
+    static const unsigned UNWIND_INDEX_NONE = -1;
     static const unsigned UNWIND_INDEX_NOT_IN_TRY = 0;
     static const unsigned UNWIND_INDEX_NOT_IN_TRY_CATCH = 1;
     static const unsigned UNWIND_INDEX_BASE = 2;
 
     void computeBlocksInFilters();
+    ArrayStack<unsigned>* computeUnwindIndexMap();
     CORINFO_GENERIC_HANDLE generateUnwindTable();
 
     bool mayPhysicallyThrow(GenTree* node);
     bool isBlockInFilter(BasicBlock* block) const;
+
+#ifdef DEBUG
+    void printUnwindIndex(ArrayStack<unsigned>* indexMap, unsigned index);
+#endif // DEBUG
 
     // ================================================================================================================
     // |                                           Shadow stack allocation                                            |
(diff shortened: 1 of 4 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

Issue #2398, opened 2023-09-08
Green Thread Experiment Results


davidwrighton:
### Update
**For the runtime-async feature currently in development, see https://github.com/dotnet/runtime/issues/109632**

---

Our goal with the green thread experiment   was to understand the basic costs and benefits of introducing green threads to the .NET Runtime environment.

## Why green threads 

The .NET asynchronous programming model makes it a breeze   to write application asynchronous code, which is crucial to achieve scalability of I/O bound scenarios.

I/O bound code spends most of its time waiting, for example waiting for data to be returned from another service over the network.   The scalability benefits of asynchronous code come from reducing the cost of requests waiting for I/O by several orders of magnitude  . For reference, the baseline cost of a request  that is waiting for I/O operation to complete is in the 100 bytes range for async C# code. The cost of the same request is in the 10 kilobytes range with synchronous code since the entire operating system thread is blocked. Async C# code allows a server of the  given size to handle several orders of magnitude more requests in parallel.

The downside of async C# code is in that developers must decide which methods need to be async. It is not viable to simply make all methods in the program async. Async methods have lower performance, limitations on the type of operations that they can perform and  async methods can only be called from other async methods . It makes the programming model complicated. What color is your function is a great description of this problem.

The key benefit of green threads is that it  makes function colors disappear and simplifies the programming model. The green threads should be cheap enough to allow all code to be written as synchronous, without giving up on scalability and performance.  Green threads have been proven to be a viable model in other programming environments. We wanted to see if it is viable with C# given the existence of async/await and the need to coexist with that model.

## What we have done

As part of this experiment, we prototyped green threads implementation within the .NET runtime exposed by new APIs to schedule/yield green thread based tasks. We also updated sockets and ASP.NET Core to use the new API to validate a basic webapi scenario end-to-end.

The prototype proved that implementing green threads in .NET and ASP.NET Core would be viable.

Async style:

```csharp
var response = context.Response;
response.StatusCode = 200;
response.ContentType = "text/plain";
response.ContentLength = payload.Length;
// Call async method for I/O explicitly
return response.Body.WriteAsync(payload).AsTask();
```

Green thread style:

```csharp
var response = context.Response;
response.StatusCode = 200;
response.ContentType = "text/plain";
response.ContentLength = payload.Length;
// Call sync method. It does async I/O under the covers! 
response.Body.Write(payload);
```

The performance of the green threads prototype was competitive with the current async/await. 

[...]

PR #2407, opened 2023-09-19
Green Threads Report

davidwrighton:
Technical details report on green threads effort

Related: https://github.com/dotnet/runtimelab/issues/2398

[Markdown](https://github.com/davidwrighton/runtimelab/blob/report/docs/design/features/greenthreads.md)


docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+We want to investigate the behavior of green threads in .NET. The goal here isn’t really to show that green threads are faster/slower than async/await, but really to investigate the programming model that would appear with green threads. The performance numbers that we gathered were done with the goal of proving that we CAN implement a reasonable green threads implementation in the current .NET Runtime, not because we expect to see major performance benefits from this approach.
+
+##  Implementation constraints and details
+As part of the experiment, we developed an implementation of green threads in the runtimelab repo. See dotnet/runtimelab at feature/green-threads (github.com) and Commits · dotnet/aspnetcore (github.com)
+The general outcome was a version of green threads that uses the segmented stack model, and provides a small number of new low level apis to work with green threads, and a fairly small number of higher level apis were changed to take advantage of green threads.
+### Fundamental api changes
+-	Thread.IsGreenThread – Used to identify if the current thread is a green thread
```
jkotas:
```suggestion
-	`Thread.IsGreenThread` – Used to identify if the current thread is a green thread
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+-	Task.RunAsGreenThread(Action) – Similar to Task.Run, but runs code on a green thread, using the green thread threadpool.
```
jkotas:
```suggestion
-	`Task.RunAsGreenThread(Action)` – Similar to `Task.Run`, but runs code on a green thread, using the green thread threadpool.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+-	Task.Wait() – When called on a green thread, will yield the green thread until the task waited on is complete, then will schedule the green thread to continue
```
jkotas:
```suggestion
-	`Task.Wait()` – When called on a green thread, will yield the green thread until the task waited on is complete, then will schedule the green thread to continue
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+1. Support for byref parameters
+  *	Known to be problematic, as they cannot simply be copied to the GC heap, and can currently only be reported to the GC as part of the stack walking logic
+2. Support for conversion of byrefs to pointers
+  * Certain byrefs, if they are known to point to the stack can safely be converted into pointers in the product today
+  * As pointers are not trackable by the runtime (and this is not fixable, as it is legal to export pointers to a non-process visible location, and then bring them back), it is not possible to safely implement a scheme which allows stacks to be moved during GC in all circumstances
+  * This complicates the design of a green thread system, as it cannot use a contiguous stack for green threads, as it would need to be prohibitively large
+  * Other systems that have successfully transitioned to green threads such as Go/Java do not have this capability, and are able to implement green thread approach which do not have the Hot-Split problem
+3.	Support for function pointers
```
jkotas:
```suggestion
  - Known to be problematic, as they cannot simply be copied to the GC heap, and can currently only be reported to the GC as part of the stack walking logic
2. Support for conversion of byrefs to pointers
  - Certain byrefs, if they are known to point to the stack can safely be converted into pointers in the product today
  - As pointers are not trackable by the runtime (and this is not fixable, as it is legal to export pointers to a non-process visible location, and then bring them back), it is not possible to safely implement a scheme which allows stacks to be moved during GC in all circumstances
  - This complicates the design of a green thread system, as it cannot use a contiguous stack for green threads, as it would need to be prohibitively large
  - Other systems that have successfully transitioned to green threads such as Go/Java do not have this capability, and are able to implement green thread approach which do not have the Hot-Split problem
3. Support for function pointers
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+5.	OS Security features
+   - In particular, shadow stack features, such as Intel CET are not simple to make compatible with green threads.
+   - The current Windows implementation of CET is not capable of working with green threads; however, it may be possible to change the Windows Kernel/API surface to make it compatible.
+6.	OS injection of code to the thread
+   - Under some circumstances, Windows will run code on the thread stack. Notably, user APCs and hardware exceptions will trigger running arbitrary code on the thread stack. As this code was not built for running on a condensed stack, there are potential problems here.
+7.	The existing Async/Await pattern
+   - The existing async/await pattern is both ubiquitous throughout the libraries and not entirely compatible with the green threading model. While it isn’t technically an implementation problem, it provides significant opportunity for customer confusion.
+   - We chose to integrate the api for these two constructs which ended up working out.
+8.	Support for thread static variables
```
jkotas:
```suggestion
4. P/Invoke
   - P/Invoked code must execute on the normal OS stack as it is not specialized to work with green threads.
   - In addition, P/Invoke is an opportunity for extremely problematic faults. In particular, if a p/invoke exposes OS thread specific state to the green thread, it may not behave correctly, as future calls to native code on the same green thread may not occur on the same OS thread.
5. OS Security features
   - In particular, shadow stack features, such as Intel CET are not simple to make compatible with green threads.
   - The current Windows implementation of CET is not capable of working with green threads; however, it may be possible to change the Windows Kernel/API surface to make it compatible.
6. OS injection of code to the thread
   - Under some circumstances, Windows will run code on the thread stack. Notably, user APCs and hardware exceptions will trigger running arbitrary code on the thread stack. As this code was not built for running on a condensed stack, there are potential problems here.
7. The existing Async/Await pattern
   - The existing async/await pattern is both ubiquitous throughout the libraries and not entirely compatible with the green threading model. While it isn’t technically an implementation problem, it provides significant opportunity for customer confusion.
   - We chose to integrate the api for these two constructs which ended up working out.
8. Support for thread static variables
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+   - In .NET we have more usage of these variables, so we implemented a re-use model for thread statics. This mirrors how our ThreadPool gets the best performance.
+
+### Implementation Challenges
+During the implementation, we encountered numerous challenges.
+-	The Windows OS is not designed to allow green threads. Making this work relies on somewhat undocumented behavior
+-	Various diagnostic tools fail to work reliably (ETW, windbg, IDNA)
+-	SEH exception handling does not safely handle cross stack segments
+-	Diagnosing implementation issues has been very slow and difficult, as they tend to cause memory corruption across multiple threads
+-	Every OS/Architecture pair needs quite a bit of custom hand-written assembly. While not yet known it is possible that MUSL based Linux may need different logic than glibc based Linux
+-	The prototype implementation was built on Windows due to team member familiarity with the platform, and therefore implementation challenges discovered were focussed on the problems encountered in development on Windows. However, there are non-Windows issues that were discovered
```
jkotas:
```suggestion
- The Windows OS is not designed to allow green threads. Making this work relies on somewhat undocumented behavior
- Various diagnostic tools fail to work reliably (ETW, windbg, IDNA)
- SEH exception handling does not safely handle cross stack segments
- Diagnosing implementation issues has been very slow and difficult, as they tend to cause memory corruption across multiple threads
- Every OS/Architecture pair needs quite a bit of custom hand-written assembly. While not yet known it is possible that MUSL based Linux may need different logic than glibc based Linux
-	The prototype implementation was built on Windows due to team member familiarity with the platform, and therefore implementation challenges discovered were focused on the problems encountered in development on Windows. However, there are non-Windows issues that were discovered
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+### Ways in which green threads are a leaky abstraction
+There are a number of details about the green thread design with are “leaky” abstractions
+-	If a green thread calls into native code and then back into managed, we cannot yield the thread, and any attempts will simply fall back to a standard thread blocking wait
+-	Interaction with native code which has thread static state is risky. If a bit of interop exposes native thread state to managed code, and then the thread changes out underneath that code interesting behaviors may occur
+
+### Example of change to plaintext demo app
+
+Before
+```
```
jkotas:
```suggestion
```csharp
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+var response = context.Response;
+response.StatusCode = 200;
+response.ContentType = "text/plain";
+response.ContentLength = payload.Length;
+return response.Body.WriteAsync(payload).AsTask();
+```
+After
+```
```
jkotas:
```suggestion
```csharp
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+1.	Implementation constraints and details that should be considered should we decide to implement this for actual production scenarios
+2.	Investigate the usage of green threads in .NET and investigate what customer and class library code would change to use the feature
+3.	Measure the performance of the product with green threads to evaluate if this is feasible from a performance standpoint
+We want to investigate the behavior of green threads in .NET. The goal here isn’t really to show that green threads are faster/slower than async/await, but really to investigate the programming model that would appear with green threads. The performance numbers that we gathered were done with the goal of proving that we CAN implement a reasonable green threads implementation in the current .NET Runtime, not because we expect to see major performance benefits from this approach.
+
+##  Implementation constraints and details
+As part of the experiment, we developed an implementation of green threads in the runtimelab repo. See dotnet/runtimelab at feature/green-threads (github.com) and Commits · dotnet/aspnetcore (github.com)
```
jkotas:
```suggestion
As part of the experiment, we developed an implementation of green threads in the runtimelab repo. See https://github.com/dotnet/runtimelab/tree/feature/green-threads and https://github.com/dotnet/aspnetcore/tree/feature/green-threads
```

jkotas (commented):
Formatting nits

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+ 
+
+## Impact on the programming model
+The green threads prototype has several implications on public API surface area.  Our approach has involved adding a minimal set of APIs directly related to green threads: a property to check whether you’re currently running on a green thread (Thread.IsGreenThread) and a method that queues work to run on a green thread and gives you back a Task to represent that work (Task.RunOnGreenThread).  All other details are handled by the implementation. A critical detail then is how do synchronous APIs participate in being handled asynchronously under the covers.  Any synchronous API that wants to be implemented in terms of a counterpart asynchronous API requires a small amount of stub code to be included.  For example, the synchronous:
+
+```
+public int Receive(byte[] buffer, int offset, int size, SocketFlags socketFlags)
+```
+method gets this bit of boilerplate added to the beginning of its implementation:
+```
```
jkotas:
```suggestion
```csharp
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+  - Known to be problematic, as they cannot simply be copied to the GC heap, and can currently only be reported to the GC as part of the stack walking logic
+2. Support for conversion of byrefs to pointers
+  - Certain byrefs, if they are known to point to the stack can safely be converted into pointers in the product today
+  - As pointers are not trackable by the runtime (and this is not fixable, as it is legal to export pointers to a non-process visible location, and then bring them back), it is not possible to safely implement a scheme which allows stacks to be moved during GC in all circumstances
+  - This complicates the design of a green thread system, as it cannot use a contiguous stack for green threads, as it would need to be prohibitively large
+  - Other systems that have successfully transitioned to green threads such as Go/Java do not have this capability, and are able to implement green thread approach which do not have the Hot-Split problem
+3. Support for function pointers
+   - Given the presence of function pointers, it is more difficult to implement a scheme where code that executes on a green thread is distinct from code which executes on normal threads.
```
jkotas: Nit: The bullet points under 1 and 2 are still rendered differently than bullet points under 3 and beyond caused by inconsistencies in indentation.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
+# Green Threads Technical Report
+
+In this experiment, we sought to explore green threads on .NET and discover
```
richlander:
```suggestion
In this experiment, we sought to explore green threads on .NET to discover:
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+1.	Implementation constraints and details that should be considered should we decide to implement this for actual production scenarios
+2.	Investigate the usage of green threads in .NET and investigate what customer and class library code would change to use the feature
+3.	Measure the performance of the product with green threads to evaluate if this is feasible from a performance standpoint
```
richlander:
```suggestion
1.	Implementation constraints and details that should be considered for a production implementation.
2.	Investigate the usage of green threads in .NET, with a focus on customer and class library code that would need to adapt.
3.	Measure the performance of the product with green threads to evaluate if this is feasible from a performance standpoint.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,231 @@
...
+We want to investigate the behavior of green threads in .NET. The goal here isn’t really to show that green threads are faster/slower than async/await, but really to investigate the programming model that would appear with green threads. The performance numbers that we gathered were done with the goal of proving that we CAN implement a reasonable green threads implementation in the current .NET Runtime, not because we expect to see major performance benefits from this approach.
```
richlander: This is written in a different tense as the previous paragraphs.

```suggestion
The primary goal of the experiment was to investigate and prototype the programming model for .NET green threads. Performance information was collected with the goal of proving that we CAN implement a reasonable green threads implementation in the current .NET Runtime. Any additional findings were considered as secondary, such as performance wins or whether green threads are faster/slower than async/await, per the prototype implementation.

The report provides a detailed description of our findings. At the current time, there is no plan to move forward with the prototype, per [dotnet/runtimelab #2398](https://github.com/dotnet/runtimelab/issues/2398).
```

agocke (approved):

docs/design/features/greenthreads.md
davidwrighton: I've fixed this issue

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+2.	Investigate the usage of green threads in .NET, with a focus on customer and class library code that would need to adapt.
+3.	Measure the performance of the product with green threads to evaluate if this is feasible from a performance standpoint.
+The primary goal of the experiment was to investigate and prototype the programming model for .NET green threads. Performance information was collected with the goal of proving that we CAN implement a reasonable green threads implementation in the current .NET Runtime. Any additional findings were considered as secondary, such as performance wins or whether green threads are faster/slower than async/await, per the prototype implementation.
+
+The report provides a detailed description of our findings. At the current time, there is no plan to move forward with the prototype, per [dotnet/runtimelab #2398](https://github.com/dotnet/runtimelab/issues/2398).
+
+##  Implementation constraints and details
+As part of the experiment, we developed an implementation of green threads in the runtimelab repo. See https://github.com/dotnet/runtimelab/tree/feature/green-threads and https://github.com/dotnet/aspnetcore/tree/feature/green-threads
```
richlander:
```suggestion
As part of the experiment, we developed an implementation of green threads in [dotnet/runtimelab:feature/green-threads](https://github.com/dotnet/runtimelab/tree/feature/green-threads) and [dotnet/aspnetcore:feature/green-threads](https://github.com/dotnet/aspnetcore/tree/feature/green-threads) repos/branches.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+The general outcome was a version of green threads that uses the segmented stack model, and provides a small number of new low level apis to work with green threads, and a fairly small number of higher level apis were changed to take advantage of green threads.
```
richlander:
```suggestion
The general outcome is a targeted implementation based on the segmented stack model. It required a small number of new low level apis to work with green threads and to adapt a small number of higher level apis to take advantage of the new functionality and to provide expected behavior with green threads. 
```

probably be best to write "APIs" throughput.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+-	`Task.RunAsGreenThread(Action)` – Similar to `Task.Run`, but runs code on a green thread, using the green thread threadpool.
+-	`Task.Wait()` – When called on a green thread, will yield the green thread until the task waited on is complete, then will schedule the green thread to continue
+### Higher level api changes
+-	Socket and file apis were changed to support green threads by transitioning synchronous apis to async apis when called on a green thread.
+-	Kestrel in ASP.NET was modified to support writing a response via a green thread.
+
+### Implementation constraints discovered during research
+A general principle of the green thread effort was that normal, already existing .NET code should be able to work when run on a green thread. There are a number of specific features of .NET which complicate implementation of green threads, and this document will attempt to describe the .NET features which are known to be problematic, and why.
+1. Support for byref parameters
+  - Known to be problematic, as they cannot simply be copied to the GC heap, and can currently only be reported to the GC as part of the stack walking logic
```
richlander:
```suggestion
  - Known to be problematic, as they cannot simply be copied to the GC heap, and can currently only be reported to the GC as part of the stack walking logic.
```

These long points are begging for a full stop.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+2. Support for conversion of byrefs to pointers
+  - Certain byrefs, if they are known to point to the stack can safely be converted into pointers in the product today
```
richlander:
```suggestion
  - Certain byrefs, if they are known to point to the stack can safely be converted into pointers.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+  - As pointers are not trackable by the runtime (and this is not fixable, as it is legal to export pointers to a non-process visible location, and then bring them back), it is not possible to safely implement a scheme which allows stacks to be moved during GC in all circumstances
+  - This complicates the design of a green thread system, as it cannot use a contiguous stack for green threads, as it would need to be prohibitively large
+  - Other systems that have successfully transitioned to green threads such as Go/Java do not have this capability, and are able to implement green thread approach which do not have the Hot-Split problem
```
richlander:
```suggestion
  - As pointers are not trackable by the runtime (and this is not fixable, as it is legal to export pointers to a non-process visible location, and then bring them back), it is not possible to safely implement a scheme which allows stacks to be moved during GC in all circumstances.
  - This complicates the design of a green thread system, as it cannot use a contiguous stack for green threads, as it would need to be prohibitively large.
  - Other systems that have successfully transitioned to green threads such as Go/Java do not have this capability, and are able to implement green thread approach which do not have the Hot-Split problem.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+3. Support for function pointers
+  - Given the presence of function pointers, it is more difficult to implement a scheme where code that executes on a green thread is distinct from code which executes on normal threads.
+  - This will make it more difficult to specialize jitted code for green threads.
+4. P/Invoke
+  - P/Invoked code must execute on the normal OS stack as it is not specialized to work with green threads.
```
richlander, line 35: Might be useful to offer some insight on other dev stacks here.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+  - In addition, P/Invoke is an opportunity for extremely problematic faults. In particular, if a p/invoke exposes OS thread specific state to the green thread, it may not behave correctly, as future calls to native code on the same green thread may not occur on the same OS thread.
+5. OS Security features
+  - In particular, shadow stack features, such as Intel CET are not simple to make compatible with green threads.
```
richlander:
```suggestion
  - In particular, shadow stack features, such as [Intel CET](https://github.com/dotnet/designs/blob/main/accepted/2021/runtime-security-mitigations.md#intel-control-flow-enforcement-technology-cet) are not simple to make compatible with green threads.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+8. Support for thread static variables
+  - Thread statics are commonly used within the BCL to achieve great performance. However, the green thread model in many languages assumes that each green thread has a unique set of thread statics.
+  - In .NET we have more usage of these variables, so we implemented a re-use model for thread statics. This mirrors how our ThreadPool gets the best performance.
+
+### Implementation Challenges
+During the implementation, we encountered numerous challenges.
+- The Windows OS is not designed to allow green threads. Making this work relies on somewhat undocumented behavior
+- Various diagnostic tools fail to work reliably (ETW, windbg, IDNA)
+- SEH exception handling does not safely handle cross stack segments
+- Diagnosing implementation issues has been very slow and difficult, as they tend to cause memory corruption across multiple threads
```
richlander: I think you need to elaborate on this one. It isn't clear what is causing the  memory corruption.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+- Every OS/Architecture pair needs quite a bit of custom hand-written assembly. While not yet known it is possible that MUSL based Linux may need different logic than glibc based Linux
```
richlander:
```suggestion
- Every OS/Architecture pair needs quite a bit of custom hand-written assembly. It is possible that MUSL-based Linux may need different logic than glibc-based Linux.
```

docs/design/features/greenthreads.md
richlander:
```suggestion
During the implementation, we encountered numerous challenges.
- The Windows OS is not designed to allow green threads. Making this work relies on somewhat undocumented behavior.
- Various diagnostic tools fail to work reliably (ETW, windbg, IDNA)
- SEH exception handling does not safely handle cross stack segments.
- Diagnosing implementation issues has been very slow and difficult, as they tend to cause memory corruption across multiple threads.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+-	The prototype implementation was built on Windows due to team member familiarity with the platform, and therefore implementation challenges discovered were focused on the problems encountered in development on Windows. However, there are non-Windows issues that were discovered
```
richlander, line 56: This isn't saying a lot.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+  - Thread local storage access on non-Windows platforms may utilize an unbounded amount of stack as it requires a call to a libc provided function in some cases
```
richlander:
```suggestion
  - Thread local storage access on Unix platforms may utilize an unbounded amount of stack as it requires a call to a libc provided function in some cases.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+  - Integration of Linux Perf with split stack is not supported as far as research has shown
```
richlander:
```suggestion
  - Integration of Linux Perf with split stack is not supported as far as research has shown.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Design of underlying runtime changes
+The underlying runtime was changed to support allocating stack segments and transitioning onto those segments when running on the thread pool. The implementation chosen was a fairly simple segmented stack approach, which is known to have various hot-split problems, but is generally considered the most straightforward way to implement a custom thread stack system.
+In the implementation, we chose to implement a unified stack with both local variables and the control stack together. This was chosen as it is somewhat simple to implement and did not violate any of the implementation constraints found. In order to support injected OS code onto the stack, as well as FCalls/HCalls, a Red Zone of about 4KB was added to each stack segment, bringing the minimum stack allocation to 12KB for a stack segment.
```
richlander: "FCalls/HCalls" 

Would be good to point to a doc where these are defined/described.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+2.	When calling native code
+  -	When calling native code, either through a p/invoke, FCall, or HCall, the logic will transition back to the OS thread to allow arbitrary C++ code to execute.
+3.	When the needed stack space exceeds the amount on the current stack segment.
+  -	The segmented stack model for green threads allows for arbitrarily deep callstacks. This is done by forming a linked list of stack segments
+  -	The presence of this linked list creates the Hot Split problem. The Hot Split problem is when an application has a need to grow to a new stack segment on each iteration of a hot loop. This can cause extremely significant unexpected performance degradation. This experiment does not attempt to work around the problem.
+
+#### Pre-prolog
+To support growing the stack, a pre-prolog was added to each managed function. While the prototype adds this to all functions, in a more production implementation, the pre-prolog would only be needed for non-leaf functions which do not use more the X bytes of stack space. (X chose empirically to provide best performance). The X bytes would be allowed to allocate within the red zone of the stack segment.
```
richlander:
```suggestion
To support growing the stack, a pre-prolog was added to each managed function. While the prototype adds this to all functions, in a more production implementation, the pre-prolog would only be needed for non-leaf functions which do not use more the X bytes of stack space. (X was chosen empirically to provide best performance). The X bytes would be allowed to allocate within the red zone of the stack segment.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+
+#### Stack layout diagrams
+##### Diagram of what the stack may look like when using multiple stack segments and a pinvoke.
+![ExperimentalStackLayout](greenthreads_diagrams/ExperimentalStackLayout.svg)
+ 
+
+##### Before yield/after yield
+
+![YieldOperation](greenthreads_diagrams/YieldOperation.svg)
+![YieldOperation2](greenthreads_diagrams/YieldOperation2.svg)
```
richlander:
```suggestion

Diagram of what the stack may look like when using multiple stack segments and a p/invoke:

![ExperimentalStackLayout](greenthreads_diagrams/ExperimentalStackLayout.svg)
 

Diagram of before yield:

![YieldOperation](greenthreads_diagrams/YieldOperation.svg)

Diagram of after yield:

![YieldOperation2](greenthreads_diagrams/YieldOperation2.svg)
```

richlander: Are these images actually different? It seems like two screen caps of the same slide.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+ 
+
+## Impact on the programming model
+The green threads prototype has several implications on public API surface area.  Our approach has involved adding a minimal set of APIs directly related to green threads: a property to check whether you’re currently running on a green thread (Thread.IsGreenThread) and a method that queues work to run on a green thread and gives you back a Task to represent that work (Task.RunOnGreenThread).  All other details are handled by the implementation. A critical detail then is how do synchronous APIs participate in being handled asynchronously under the covers.  Any synchronous API that wants to be implemented in terms of a counterpart asynchronous API requires a small amount of stub code to be included.  For example, the synchronous:
```
richlander:
```suggestion
The green threads prototype has several implications on public API surface area.

The following APIs were added:

- `Thread.IsGreenThread` -- A property to check whether you’re currently running on a green thread.
- `Task.RunOnGreenThread` -- A method that queues work to run on a green thread and gives you back a Task to represent that work.  

All other details are handled by the implementation. A critical detail then is how do synchronous APIs participate in being handled asynchronously under the covers.  Any synchronous API that wants to be implemented in terms of a counterpart asynchronous API requires a small amount of stub code to be included.  

For example, the synchronous:
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+```csharp
+if (Thread.IsGreenThread)
+{
+    ValueTask<int> vt = ReceiveAsync(buffer.AsMemory(offset, size), socketFlags);
+    return vt.IsCompleted ?
+        vt.GetAwaiter().GetResult() :
+        vt.AsTask().GetAwaiter().GetResult();
+}
+```
+In other words, the synchronous Receive checks whether it’s on a green thread, and if it is, instead delegates to the corresponding ReceiveAsync method.  That method returns a ValueTask<int>.  If that operation completes synchronously, or at least has completed by the time we check, then this “synchronous” operation just returns the result of the completed operation (or throws an exception if it failed).  If, however, the operation didn’t complete yet, the implementation blocks in the equivalent of Task.Wait().  The implementation of Task.Wait has been updated with the necessary logic to suspend this green thread until the Task completes, hooking up a continuation to the Task that will resume the green thread when the Task completes, unblocking the wait.
```
richlander:
```suggestion
In other words, the synchronous Receive checks whether it’s on a green thread, and if it is, instead delegates to the corresponding `ReceiveAsync` method.  That method returns a `ValueTask<int>`.  If that operation completes synchronously, or at least has completed by the time we check, then this “synchronous” operation just returns the result of the completed operation (or throws an exception if it failed).  If, however, the operation didn’t complete yet, the implementation blocks in the equivalent of `Task.Wait()`.  The implementation of `Task.Wait` has been updated with the necessary logic to suspend this green thread until the `Task` completes, hooking up a continuation to the Task that will resume the green thread when the `Task` completes, unblocking the wait.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+  - The existing async/await pattern is both ubiquitous throughout the libraries and not entirely compatible with the green threading model. While it isn’t technically an implementation problem, it provides significant opportunity for customer confusion.
+  - We chose to integrate the api for these two constructs which ended up working out.
+8. Support for thread static variables
+  - Thread statics are commonly used within the BCL to achieve great performance. However, the green thread model in many languages assumes that each green thread has a unique set of thread statics.
+  - In .NET we have more usage of these variables, so we implemented a re-use model for thread statics. This mirrors how our ThreadPool gets the best performance.
+
+### Implementation Challenges
+During the implementation, we encountered numerous challenges.
+- The Windows OS is not designed to allow green threads. Making this work relies on somewhat undocumented behavior
+- Various diagnostic tools fail to work reliably (ETW, windbg, IDNA)
```
jkotas:
```suggestion
- Various diagnostic tools fail to work reliably (ETW, windbg, TTD)
```
IDNA is MS-internal acronym, I do not see it documented anywhere https://learn.microsoft.com/en-us/search/?terms=IDNA
TTD is the public TLA for this tech. https://learn.microsoft.com/en-us/search/?terms=TTD

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+{
+    ValueTask<int> vt = ReceiveAsync(buffer.AsMemory(offset, size), socketFlags);
+    return vt.IsCompleted ?
+        vt.GetAwaiter().GetResult() :
+        vt.AsTask().GetAwaiter().GetResult();
+}
+```
+In other words, the synchronous Receive checks whether it’s on a green thread, and if it is, instead delegates to the corresponding ReceiveAsync method.  That method returns a ValueTask<int>.  If that operation completes synchronously, or at least has completed by the time we check, then this “synchronous” operation just returns the result of the completed operation (or throws an exception if it failed).  If, however, the operation didn’t complete yet, the implementation blocks in the equivalent of Task.Wait().  The implementation of Task.Wait has been updated with the necessary logic to suspend this green thread until the Task completes, hooking up a continuation to the Task that will resume the green thread when the Task completes, unblocking the wait.
+This approach makes it fairly straightforward for us to instrument any synchronous API with an asynchronous counterpart, having the former delegate to the latter, albeit with a variety of implications:
```
richlander, line 171:
```suggestion

This approach makes it fairly straightforward for us to instrument any synchronous API with an asynchronous counterpart, having the former delegate to the latter, albeit with a variety of implications:
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
...ated and a synchronization primitive (e.g. manual reset event) to be created as part of the wait.  Some of this is inherent to the operation, and some of it could likely be optimized in the future should it be necessary.  There are other costs that will be larger and more challenging.  For example, the previous Receive example accepted a byte[] into which to store the received data. What about the overload that takes a Span<byte> instead?  Delegating to the asynchronous implementation requires something that can be stored on the heap.  The prototype currently implements this by renting an array pool array, receiving into that, and then copying back into the span, but that has measurable overhead, in particular for synchronous operations that complete synchronously due to data already being available.  There are other possibilities to explore, such as pinning and utilizing a pooled heap-based object that can be passed around, or pinning and using some other non-Memory<byte>-based asynchronous implementation implemented to work directly in terms of a pointer/length.  Employing such options, however, will increase the complexity and scope of the work required to roll this out broadly.
```
richlander:
```suggestion
-	There is now overhead involved in the synchronous operations that we didn’t exert before and that doesn’t exist in the corresponding asynchronous method. This includes needing to check the current thread to determine whether it is a green thread, and also to pay for a `Task` to be created and a synchronization primitive (e.g. manual reset event) to be created as part of the wait.  Some of this is inherent to the operation, and some of it could likely be optimized in the future should it be necessary.  There are other costs that will be larger and more challenging.  For example, the previous Receive example accepted a `byte[]` into which to store the received data. What about the overload that takes a `Span<byte>` instead?  Delegating to the asynchronous implementation requires something that can be stored on the heap.  The prototype currently implements this by renting an array pool array, receiving into that, and then copying back into the span, but that has measurable overhead, in particular for synchronous operations that complete synchronously due to data already being available.  There are other possibilities to explore, such as pinning and utilizing a pooled heap-based object that can be passed around, or pinning and using some other non-`Memory<byte>`-based asynchronous implementation implemented to work directly in terms of a pointer/length.  Employing such options, however, will increase the complexity and scope of the work required to roll this out broadly.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+The inverse is also relevant to consider.  While the majority of our functionality is exposed as at least a synchronous method and then potentially also an asynchronous method, there are a few libraries where only asynchronous APIs have been exposed for the relevant functionality, e.g. System.Threading.Channels, System.IO.Pipelines, and to a large extent, System.Net.Http (although it gained a minimal set of golden-path synchronous APIs in a few years back).  We would need to consider whether to add synchronous APIs to these libraries.  If we didn’t add them, developers that wanted to consume them in a scalable manner from green threads would need to wrap them themselves; we could potentially provide helpers to make that boilerplate easier.  If we did add them, we’d need to think through the ramifications for when they’re used _not_ on green threads.  This kind of sync-over-async API is something we actively discourage developers from adding today, and choose to never do ourselves in new surface area.
```
richlander:
```suggestion
The inverse is also relevant to consider.  While the majority of our functionality is exposed as at least a synchronous method and then potentially also an asynchronous method, there are a few libraries where only asynchronous APIs have been exposed for the relevant functionality, e.g. System.Threading.Channels, System.IO.Pipelines, and to a large extent, System.Net.Http (although it gained a minimal set of golden-path synchronous APIs a few years back).  We would need to consider whether to add synchronous APIs to these libraries.  If we didn’t add them, developers that wanted to consume them in a scalable manner from green threads would need to wrap them themselves; we could potentially provide helpers to make that boilerplate easier.  If we did add them, we’d need to think through the ramifications for when they’re used _not_ on green threads.  This kind of sync-over-async API is something we actively discourage developers from adding today, and choose to never do so ourselves in new surface area.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+### Ways in which green threads are a leaky abstraction
+There are a number of details about the green thread design with are “leaky” abstractions
+-	If a green thread calls into native code and then back into managed, we cannot yield the thread, and any attempts will simply fall back to a standard thread blocking wait
+-	Interaction with native code which has thread static state is risky. If a bit of interop exposes native thread state to managed code, and then the thread changes out underneath that code interesting behaviors may occur
```
richlander:
```suggestion
-	Interaction with native code which has thread static state is risky. If a bit of interop exposes native thread state to managed code, and then the thread changes out underneath that code and interesting behaviors may occur
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+7. The existing Async/Await pattern
+  - The existing async/await pattern is both ubiquitous throughout the libraries and not entirely compatible with the green threading model. While it isn’t technically an implementation problem, it provides significant opportunity for customer confusion.
+  - We chose to integrate the api for these two constructs which ended up working out.
+8. Support for thread static variables
+  - Thread statics are commonly used within the BCL to achieve great performance. However, the green thread model in many languages assumes that each green thread has a unique set of thread statics.
+  - In .NET we have more usage of these variables, so we implemented a re-use model for thread statics. This mirrors how our ThreadPool gets the best performance.
+
+### Implementation Challenges
+During the implementation, we encountered numerous challenges.
+- The Windows OS is not designed to allow green threads. Making this work relies on somewhat undocumented behavior
```
jkotas:
```suggestion
- The Windows OS is not designed to allow green threads
```
We should be either more specific or drop this.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Example of change to plaintext demo app
+
+Before
+```csharp
+var response = context.Response;
+response.StatusCode = 200;
+response.ContentType = "text/plain";
+response.ContentLength = payload.Length;
+return response.Body.WriteAsync(payload).AsTask();
```
richlander, line 189:
```suggestion
// Call async method for I/O explicitly
return response.Body.WriteAsync(payload).AsTask();
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+
+Yielding a green thread and resuming it combined takes about 500 ns. By comparison, two context switches of a regular thread take about 8 µs. This may be interesting for reducing the CPU overhead of spin-waiting, and custom thread scheduling. Awaiting an async operation and continuing it takes only 110 ns, so there is a bit of a gap. However, async/await can become viral and longer await chains would each incur the overhead whereas green thread yield/resume would not.
+
+![AwaitVsWait](greenthreads_diagrams/AwaitVsWait.png)
+
+- [Deep await on regular thread VS Deep await on green thread] Awaits have more overhead currently when they happen on green threads. Since the continuation need to be marshaled to run on a green thread, they miss the fast path, involving more allocation, and some more overhead in switching to a green thread. There may be ways to reduce this overhead, but there will probably be some overhead that is a necessary consequence of enabling both green and non-green threads to interoperate. For instance, if thread pool worker threads ran as green threads by default, some of this overhead would likely go away.
```
richlander:
```suggestion
- **Deep await on regular thread vs Deep await on green thread** -- Awaits have more overhead currently when they happen on green threads. Since the continuation need to be marshaled to run on a green thread, they miss the fast path, involving more allocation, and some more overhead in switching to a green thread. There may be ways to reduce this overhead, but there will probably be some overhead that is a necessary consequence of enabling both green and non-green threads to interoperate. For instance, if thread pool worker threads ran as green threads by default, some of this overhead would likely go away.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+- [Deep wait on regular dedicated thread VS Deep wait on green thread] Both appear to be similar, with slightly worse scaling on green threads. At lower depths perf is better on green threads, a theory may be that the yield/resume latency is better than spin-waiting latency. At higher depths, the extra cost may be a result of stack switching and less efficient use of memory.
+- [Deep await on regular thread VS Deep wait on green thread] It seems clear that longer await chains don’t scale well. They don’t happen in TechEmpower benchmarks, I have seen very long await chains in Bing SNR. Await on a regular thread is cheaper at low depths, this is a similar gap to what was mentioned before, perhaps there would be ways to reduce it but likely some of this gap will remain and yield tradeoffs.
```
richlander:
```suggestion
- **Deep wait on regular dedicated thread vs Deep wait on green thread** -- Both appear to be similar, with slightly worse scaling on green threads. At lower depths perf is better on green threads, a theory may be that the yield/resume latency is better than spin-waiting latency. At higher depths, the extra cost may be a result of stack switching and less efficient use of memory.
- **Deep await on regular thread vs Deep wait on green thread** -- It seems clear that longer await chains don’t scale well. They don’t happen in TechEmpower benchmarks, I have seen very long await chains in Bing SNR. Await on a regular thread is cheaper at low depths, this is a similar gap to what was mentioned before, perhaps there would be ways to reduce it but likely some of this gap will remain and yield tradeoffs.
```

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+-	User research – Validate our hypotheses about the green threads programming model with customers
+- Build a prototype that works on Linux
+-	Broader benchmarks of more real-world code
+-	Investigate impact of green threads infrastructure on non-green thread code
+-	Enable support for exceptions, and validate that the performance of exceptions on green threads is substantially better than performance on exceptions over async/await
+-	Work with Windows kernel team to develop a model that would allow .NET to manage its own thread stacks without abandoning the security posture that Windows provides
+-	Enable non-conservative stack scanning of green thread stack segments, and measure performance
+-	Evaluate scalability of green thread system, and evaluate reasonable performant limits
+-	Investigate where to put IsGreenThread checks. Should they be scattered through code, or only in I/O?
```
richlander:
```suggestion
-	Investigate where to put `IsGreenThread` checks. Should they be scattered through code, or only in I/O?
```

docs/design/features/greenthreads.md
davidwrighton: That's not easy, we had a series of about different memory corruption issues generally caused by multiple physical thread interaction with the green thread model, and there was nothing I can clearly describe at this point. I recall thinking that many of the issues were TLS related, but not all of them.

docs/design/features/greenthreads.md
davidwrighton: I'm not trying to say a lot here other than we don't know what we'll find on Unix.

kant2002:
This is outstanding!

docs/design/features/greenthreads.md
agocke: These are runtime-internal concepts, but there is a section in BOTR on them.

docs/design/features/greenthreads.md
```diff
@@ -0,0 +1,233 @@
...
+  - Thread local storage access on Unix platforms may utilize an unbounded amount of stack as it requires a call to a libc provided function in some cases.
+  - Integration of Linux Perf with split stack is not supported as far as research has shown.
+
+### Design of underlying runtime changes
+The underlying runtime was changed to support allocating stack segments and transitioning onto those segments when running on the thread pool. The implementation chosen was a fairly simple segmented stack approach, which is known to have various hot-split problems, but is generally considered the most straightforward way to implement a custom thread stack system.
+In the implementation, we chose to implement a unified stack with both local variables and the control stack together. This was chosen as it is somewhat simple to implement and did not violate any of the implementation constraints found. In order to support injected OS code onto the stack, as well as FCalls/HCalls, a Red Zone of about 4KB was added to each stack segment, bringing the minimum stack allocation to 12KB for a stack segment.
```
jkotas:
```suggestion
In the implementation, we chose to implement a unified stack with both local variables and the control stack together. This was chosen as it is somewhat simple to implement and did not violate any of the implementation constraints found. In order to support injected OS code onto the stack, as well as [FCalls/HCalls](https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/botr/corelib.md#calling-from-managed-to-native-code), a Red Zone of about 4KB was added to each stack segment, bringing the minimum stack allocation to 12KB for a stack segment.
```

docs/design/features/greenthreads.md
davidwrighton: We'll drop this. The issue that I found was around access to fields of the TEB, and I found external resources which described how to modify the reserved fields of the TEB to allow code to generally work on a green thread, but I do not have confidence that those external resources had a complete understanding of what they were doing and how changes there are handled by the OS. I didn't work with the Microsoft OS team on this issue and do not know what sorts of issues they would have with this approach.

docs/design/features/greenthreads.md
davidwrighton: They are different now.

docs/design/features/greenthreads.md
davidwrighton: I've appliedthese changes manually.

docs/design/features/greenthreads.md
davidwrighton: Link has been added to the appropriate BOTR chapter

davidwrighton:
Suggested edits have either been taken or explicitly rejected.

PR #2419, opened 2023-11-12
NativeAOT-LLVM: Merge nov 23

yowl:
This PR merges to commit 76aeefb8230c897ff83eaaa7fd66adbe9c7e6b5f.

Of note:

- Add `FUNC` defintions for `CORINFO_HELP_GETCLASSFROMMETHODPARAM` and `CORINFO_HELP_GETSYNCFROMCLASSHANDLE`
- DIsable standalone GC
- Enable `Hybrid { get; }` in `GlobalizationMode` for Wasi
- Reinstate the Javascript version of `dotnet_browser_entropy` but rename to `mono_wasm_browser_entropy` to avoid bringing in a Typescript dependency.
- Disable the new `Determinism` smoke test as it uses file IO.
- Copied `toCorInfoType` to compiler.hpp - Any suggestion how I should avoid that? (I want to set the LLVM only `gtCorInfoType` field in `gtNewHelperCallNode`)
- Add cases for `SCK_NULL_REF_EXCPN` in `flowgraph.cpp` and add a call to LLVM lower the helper in `acdHelper`, exposing `lowerRange`


yowl:
@jkotas There are a lot of yml changes with this merge and obviously I've got somethings wrong.  Can you see what the problem is with the yml that is stopping the jobs loading?  Thanks.

jkotas:
The error is:
```
/eng/pipelines/runtimelab.yml (Line: 126, Col: 15): 'buildArgs' is already defined
473417
```

yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/jit/compiler.hpp
```diff
@@ -1447,6 +1487,10 @@ inline GenTreeCall* Compiler::gtNewHelperCallNode(
         result->gtFlags |= arg1->gtFlags & GTF_ALL_EFFECT;
     }
 
+#ifdef TARGET_WASM
+    result->gtCorInfoType = toCorInfoType(type);
```
SingleAccretion: Why is this needed? Lowering should be able to find the right type by itself.

src/coreclr/jit/compiler.hpp
```diff
@@ -3117,12 +3161,7 @@ inline bool Compiler::fgIsThrowHlpBlk(BasicBlock* block)
         if (block == add->acdDstBlk)
         {
             return add->acdKind == SCK_RNGCHK_FAIL || add->acdKind == SCK_DIV_BY_ZERO || add->acdKind == SCK_OVERFLOW ||
-<<<<<<< HEAD
-                   add->acdKind == SCK_ARG_EXCPN || add->acdKind == SCK_ARG_RNG_EXCPN ||
-                   add->acdKind == SCK_NULL_REF_EXCPN;
-=======
                    add->acdKind == SCK_ARG_EXCPN || add->acdKind == SCK_ARG_RNG_EXCPN || add->acdKind == SCK_FAIL_FAST;
```
SingleAccretion: `SCK_NULL_REF_EXCPN` missing.

src/coreclr/jit/emit.h
```diff
@@ -1,10 +1,13 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 /*****************************************************************************/
-#ifndef TARGET_WASM
 #ifndef _EMIT_H_
 #define _EMIT_H_
 
+#ifdef TARGET_WASM
+#define BITS_IN_BYTE (8) // TODO-LLVM: required for switchrecoginition.cpp SWITCH_MAX_DISTANCE
```
SingleAccretion: It would be better to switch `switchrecoginition.cpp` to use `BITS_PER_BYTE` (in this change) and delete `BITS_IN_BYTE` (upstream).

src/coreclr/jit/llvm.h
```diff
@@ -365,6 +365,7 @@ class Llvm
     void lowerStoreBlk(GenTreeBlk* storeBlkNode);
     void lowerStoreDynBlk(GenTreeStoreDynBlk* storeDynBlkNode);
     void lowerDivMod(GenTreeOp* divModNode);
+    void LowerArrLength(GenTreeArrCommon* node);
```
SingleAccretion: Nit: `LowerArrLength` -> `lowerArrLength`.

src/coreclr/jit/llvmlower.cpp
```diff
@@ -590,12 +596,79 @@ void Llvm::lowerDivMod(GenTreeOp* divModNode)
...
+        // If the array is NULL, then we should get a NULL reference
+        // exception when computing its length.  We need to maintain
+        // an invariant where there is no sum of two constants node, so
+        // let's simply return an indirection of NULL.
+
+        addr = arr;
     }
+    else
+    {
+        GenTree* con = _compiler->gtNewIconNode(lenOffset, TYP_I_IMPL);
```
SingleAccretion, line 662:
```suggestion
        // TODO-LLVM-CQ: use LEAs here.
        GenTree* con = _compiler->gtNewIconNode(lenOffset, TYP_I_IMPL);
```

src/coreclr/jit/llvmlower.cpp
```diff
@@ -590,12 +596,79 @@ void Llvm::lowerDivMod(GenTreeOp* divModNode)
...
+        addr         = _compiler->gtNewOperNode(GT_ADD, TYP_BYREF, arr, con);
+
+        CurrentRange().InsertAfter(arr, con, addr);
+    }
+
+    _compiler->fgAddCodeRef(CurrentBlock(), SCK_NULL_REF_EXCPN);
+
+    // Change to a GT_IND.
+    node->ChangeOper(GT_IND);
+    node->AsIndir()->Addr() = addr;
```
SingleAccretion: This should call `lowerIndir` after changing the node instead of `fgAddCodeRef` above.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
```diff
@@ -99,17 +99,14 @@
     <Copy SourceFiles="$(NativeOutputPath)$(TargetName)$(NativeBinaryExt)" DestinationFolder="$(PublishDir)" />
   </Target>
 
-<<<<<<< HEAD
+  <Target Name="_CopyAotSymbols" AfterTargets="Publish"
+          Condition="'$(CopyOutputSymbolsToPublishDirectory)' == 'true'">
     <Delete Files="$(PublishDir)\$(TargetName).js" Condition="'$(_targetOS)' == 'browser'"/>
     <Copy SourceFiles="$(NativeOutputPath)$(TargetName).js" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser'"/>
```
SingleAccretion: This target is copying symbols; copying `.js`/`.wasm` should be included in the preceding target I think (otherwise when the user sets `CopyOutputSymbolsToPublishDirectory`, they won't get any output).

src/coreclr/nativeaot/Runtime/Portable/CMakeLists.txt
```diff
@@ -7,6 +7,9 @@ add_definitions(-DUSE_PORTABLE_HELPERS)
 
 add_library(PortableRuntime STATIC ${COMMON_RUNTIME_SOURCES} ${PORTABLE_RUNTIME_SOURCES})
 
+add_library(standalonegc-disabled STATIC ${STANDALONEGC_DISABLED_SOURCES})
+add_library(standalonegc-enabled STATIC ${STANDALONEGC_ENABLED_SOURCES})
```
SingleAccretion: I don't think we need to build `standalonegc-enable`, do we?

src/coreclr/nativeaot/Runtime/clrgc.enabled.cpp
```diff
@@ -12,6 +12,10 @@
 #include "holder.h"
 #include "RhConfig.h"
 
+#if defined(TARGET_WASM)
+#define BYTE unsigned char
```
SingleAccretion: If we don't build `standalonegc-enable`, we will not need this.

src/coreclr/nativeaot/Runtime/wasm/PalRedhawkWasm.cpp
```diff
@@ -197,9 +197,9 @@ extern "C" int __cxa_thread_atexit(Dtor dtor, void* obj, void*)
 // runtime does not require this functionality either, and so we can implement this function in terms of simple
 // "malloc".
 //
-REDHAWK_PALEXPORT _Ret_maybenull_ _Post_writable_byte_size_(size) void* REDHAWK_PALAPI PalVirtualAlloc(_In_opt_ void* pAddress, size_t size, uint32_t allocationType, uint32_t protect)
+REDHAWK_PALEXPORT _Ret_maybenull_ _Post_writable_byte_size_(size) void* REDHAWK_PALAPI PalVirtualAlloc(uintptr_t size, uint32_t protect)
 {
-    if ((pAddress != nullptr) || (allocationType != MEM_COMMIT) || (protect != PAGE_READWRITE))
+    if ((protect != PAGE_READWRITE))
```
SingleAccretion:
```suggestion
    if (protect != PAGE_READWRITE)
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System.Private.CoreLib.csproj
```diff
@@ -25,12 +25,8 @@
   </ItemGroup>
 
   <PropertyGroup>
-<<<<<<< HEAD
-    <DefineConstants>SYSTEM_PRIVATE_CORELIB;FEATURE_MANAGED_ETW_CHANNELS;FEATURE_MANAGED_ETW;$(DefineConstants)</DefineConstants>
-    <DefineConstants Condition="'$(DebugWasmEH)' == 'true'">ENABLE_NOISY_WASM_EH_LOG;$(DefineConstants)</DefineConstants>
-=======
     <DefineConstants>SYSTEM_PRIVATE_CORELIB;$(DefineConstants)</DefineConstants>
->>>>>>> origin/runtime-main
+    <DefineConstants Condition="'$(DebugWasmEH)' == 'true'">ENABLE_NOISY_WASM_EH_LOG;$(DefineConstants)</DefineConstants>
```
SingleAccretion: If it's causing merge conflicts, let's just delete this. I was handing in bringing up new EH, as you could specify it on the command line (`./build clr.nativeaotlibs /p:DebugWasmEH=true`).
```suggestion
```

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
```diff
@@ -136,6 +137,7 @@ private void CompileMethods(List<LLVMMethodCodeNode> methodsToCompile)
                 }
             }
 
+            _parallelism = 1;
```
SingleAccretion: Why is this needed?

src/libraries/System.Private.CoreLib/src/System/Globalization/GlobalizationMode.cs
```diff
@@ -17,7 +17,7 @@ private static partial class Settings
 #else
             internal static bool Invariant { get; } = AppContextConfigHelper.GetBooleanConfig("System.Globalization.Invariant", "DOTNET_SYSTEM_GLOBALIZATION_INVARIANT");
 #endif
-#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS || TARGET_BROWSER
+#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS || TARGET_BROWSER || TARGET_WASI
```
SingleAccretion: Why does WASI needs this defined? Hybrid globalization does not make sense on WASI.

src/native/libs/System.Native/pal_random.lib.js
```diff
@@ -28,13 +28,17 @@ const DotNetEntropyLib = {
             }
         }
     },
-    dotnet_browser_entropy: function (buffer, bufferLength) {
+
+    // TODO-LLVM: Mono uses a typescript file for mono_wasm_browser_entropy and we probably  don't want to introduce typescript as a dependency
+    // so we stay with the plain Javascript implementation but rename to avoid changing random.c
+    // Revist if and when the Wasi HTTP or CLI proposal progresses (which contain a secure RNG in wasi-random)
```
SingleAccretion: For my education: are browsers expected to implement `wasi-random`?

(P1 of WASI I believe already has some facility for getting these)

src/tests/Common/dirs.proj
```diff
@@ -27,12 +26,12 @@
       <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\StackTraceMetadata\StackTraceMetadata.csproj" Condition="'$(TargetOS)' == 'wasi'" />
       <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\StackTraceMetadata\StackTraceMetadata_Stripped.csproj" Condition="'$(TargetOS)' == 'browser'" />
       <DisabledProjects Include="$(TestRoot)nativeaot\GenerateUnmanagedEntryPoints\GenerateUnmanagedEntryPoints.csproj" />
+      <!--Determinism uses File.Open read which Wasm can't support without Wasi or Emscripten file system support which we don't enable by default. -->
+      <DisabledProjects Include="$(TestRoot)nativeaot\SmokeTests\Determinism\Determinism.csproj" />
```
SingleAccretion: We could rewrite this test upstream to compare the files in MSBuild, and then delete our own test that does the same thing... I may look into it.

src/coreclr/jit/llvm.h
yowl: Thanks, and I also capitilised `LowerRange` as it is now public.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
yowl: Thanks, wasn't concentrating.

src/coreclr/nativeaot/Runtime/wasm/PalRedhawkWasm.cpp
yowl: Thanks

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/LLVMCodegenCompilation.cs
yowl: oops, thanks

src/libraries/System.Private.CoreLib/src/System/Globalization/GlobalizationMode.cs
yowl: Whatever the reason was it's gone now.

src/native/libs/System.Native/pal_random.lib.js
yowl: In my head, yes!  Let me see if I can get something better for educational purposes.

src/coreclr/nativeaot/Runtime/Portable/CMakeLists.txt
```diff
@@ -29,3 +35,9 @@ add_custom_command(
 )
 
 install_static_library(PortableRuntime aotsdk nativeaot)
+install_static_library(standalonegc-disabled aotsdk nativeaot)
+
+if (NOT CLR_CMAKE_TARGET_ARCH_WASM)
+install_static_library(standalonegc-enabled aotsdk nativeaot)
```
SingleAccretion:
```suggestion
  install_static_library(standalonegc-enabled aotsdk nativeaot)
```

src/coreclr/jit/flowgraph.cpp
```diff
@@ -3846,6 +3693,9 @@ PhaseStatus Compiler::fgCreateThrowHelperBlocks()
                 case SCK_FAIL_FAST:
                     msg = " for FAIL_FAST";
                     break;
+                case SCK_NULL_REF_EXCPN:
+                    msg = " for NULL_REF_EXCPN3";
```
SingleAccretion:
```suggestion
                    msg = " for NULL_REF_EXCPN";
```

src/coreclr/tools/aot/ILCompiler.RyuJit/Compiler/RyuJitCompilation.cs
```diff
@@ -23,14 +23,9 @@ public class RyuJitCompilation : Compilation
...
         private readonly ProfileDataManager _profileDataManager;
-<<<<<<< HEAD
         protected readonly MethodImportationErrorProvider _methodImportationErrorProvider;
-        protected readonly int _parallelism;
-=======
-        private readonly MethodImportationErrorProvider _methodImportationErrorProvider;
         private readonly ReadOnlyFieldPolicy _readOnlyFieldPolicy;
-        private readonly int _parallelism;
->>>>>>> origin/runtime-main
+        protected int _parallelism;
```
SingleAccretion: I suspect you can restore the `readonly` now:
```suggestion
        protected readonly int _parallelism;
```

src/libraries/System.Private.CoreLib/src/System/Globalization/GlobalizationMode.cs
SingleAccretion:
```suggestion
#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS || TARGET_BROWSER
```
?

yowl:
Thanks for picking these up.

src/native/libs/System.Native/pal_random.lib.js
yowl: I've not been able to find anything about this, nor get any response from a browser team.  I wouldn't be suprised if it happens, it would make sense to me.  However as Typescript is going to be installed for Mono users anyway, perhaps it is only a small ask to NAOT-LLVM to make sure that workload is available also.  Happy to change to the typescript version (and add whatever that means to our build targets) if you'd prefer?

SingleAccretion: No, I think it's completely fine to leave as-is for now. We may revisit this depending on how #2434 goes.

> I've not been able to find anything about this, nor get any response from a browser team. I wouldn't be suprised if it happens, it would make sense to me. 

SingleAccretion (approved):
Looks great, thank you!

jkotas:
Thank you!

PR #2421, opened 2023-11-16
Add measurement of async2 logic with Execution and SynchronizationContext capture/restore

davidwrighton:
Add measurement of async2 logic with Execution and SynchronizationContext capture/restore
- Attempt to determine if we could fully simulate the current async semantics while still achieving significant performance wins
- TLDR: Yes, we can get a pretty substantial performance gain, but we lose somewhere between 20-30% of what we could gain


jakobbotsch (approved):

davidwrighton:
CI failures are expected. This repo has not yet been cleaned up such that it will build nicely.

PR #2425, opened 2023-11-17
[Async-2]: Tasklet aging

VSadov:
- Added a test that measures durations of trivial gen-0 GCs in the presence of 10000 suspended tasks, with 100 frames each.
Both async-1 and async-2 are measured.

- Introduced tasklet aging to avoid reporting old stacks in younger GCs. Since a suspended stack cannot suddenly start pointing to younger objects, we can avoid reporting some suspended stacks when they cannot possibly have anything young enough to be of interest to the current GC cycle. 
Tasklet aging considerably reduces GC pauses due to root reporting. 

- Added a switch to turn aging on/off - `DOTNET_TaskletAging`.
The default state is `on`

- Changed the shape of tasklet storage. Instead of one big list, it is a list of lists. The inner lists belong to the same stack. 
This is to enable aging of an entire stack. Also to reduce locking when tasklets are resumed (only resuming/unlinking the head needs a lock)

### DIFF
--- a/docs/design/features/runtime-handled-tasks.md
+++ b/docs/design/features/runtime-handled-tasks.md
@@ -618,3 +618,55 @@ In the current experiment we have looked into two implementing strategies:
 ## On unmanaged pointers.
 We cannot possibly track unmanaged pointers that refer to the stack across suspensions. Disallowing such scenarios would not be a take-back though since currently C# does not allow `await` in unsafe context, therefore await and unmanaged pointers cannot mix.
 
+## Impact on GC (heap size, pauses).
+
+We have examined the behavior of experimental async2 implementations on GC. For simplicity we looked into behavior of a simple benchmark that ensures 10000 suspended tasks at some point, each containing 100 stack frames with at least 1 object and 1 integer alive across await and thus requiring capture. At the time the tasks are suspended, the benchmark forces a number of Gen0 GCs and measures average time taken by GCs as well as the managed heap size.
+To further reduce the number of involved variables, the benchmark is single-threaded and Workstation GC was used.
+
+The benchmark produces very little garbage when the masurements are performed as the main interest is not on how GC handles allocations, but on how the presence of captured tasks impacts GC.
+
+Our observations (on x64, AMD 5950X):
+
+**===== Async1**:
+
+There is some noise, but a typical Gen0 pause in the above scenario was: 120 microseconds.
+The managed heap size (as reported by `GetGCMemoryInfo.HeapSizeBytes`) is stable at 130 Mb.
+The peak working set (as reported by `Task Manager`) is 148 Mb. 
+
+**=====Async2 (unwinding/stack-capturing based)**:
+
+When GC pauses were measured as-is, the results were very disappointing - **48690 microsecond**. (**~400x worse** than async1 and generally unacceptable for a Gen0 pause)
+
+The reasons for such timings is that the implementation would go through every one of the 1000000 tasklets and report referenced objects as roots. That is time consuming and is also very redundant when GC generational model is considered. Since suspended stacks can't see assignments of new objects, once GC performs en-masse promotion, all the roots owned by these tasks become too old to be of interest for collections that target younger generations until the containing task is resumed.  
+This is a unique challenge of stack-capturing implementation. In async1 suspended tasks are regular heap objects and do not require any special reporting in GC STW phases.
+
+As a solution to this issue, we implemented tasklet aging mechanism, similar to approach used by GC handles. We make note of GC promotions, and increase "min age" of suspended tasks, so that, for example, tasks that only refer to tenured objects do not need to report roots in ephemeral collections.
+
+The taklet aging was an enormous optimization and the pause time in the above benchmark went down to **480 microseconds**. That is **100x improvement** and is not too bad, considering that this is close to the worst case scenario.
+It was still **4x worse** than async1.
+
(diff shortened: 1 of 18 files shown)


VSadov:
For the async-1, I get on my machine:

```
async-1 =====================
..........Time per Gen0 GC (microseconds): 255
..........Time per Gen0 GC (microseconds): 232
..........Time per Gen0 GC (microseconds): 171
..........Time per Gen0 GC (microseconds): 228
..........Time per Gen0 GC (microseconds): 225
..........Time per Gen0 GC (microseconds): 287
..........Time per Gen0 GC (microseconds): 389
..........Time per Gen0 GC (microseconds): 371
```

For the async-2 (unwinder based), I get
With aging:
```
async-2 =====================
..........Time per Gen0 GC (microseconds): 2490
..........Time per Gen0 GC (microseconds): 2319
..........Time per Gen0 GC (microseconds): 1552
..........Time per Gen0 GC (microseconds): 2349
```

With aging turned off:
```
async-2 =====================
..........Time per Gen0 GC (microseconds): 55181
```

VSadov:
@jakobbotsch -  I expect the state-machine flavor to behave similar to async-1 on this test, but could not measure due to unexpected exceptions.  
Possible some issue with keeping continuations alive?

```
async-2 ===================== 
..........Time per Gen0 GC (microseconds): 377
System.AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not set to an instance of an object.) (Object reference not . . . .
```

src/tests/Loader/async/async2gcRootsScan.cs
```diff
@@ -0,0 +1,159 @@
...
+    static async2 Task<int> Recursive2(int n)
+    {
+        Task<int> cTask = cs.Task;
+
+        // make some garbage
+        object o1 = n;
+
+// THIS CAUSES CRASHES IN UNWINDER FLAVOR
+        //object o2 = n;
+        //object o3 = n;
```
VSadov, line 62: I'd like to uncomment these to have a more real benchmark, but it causes crashes in unwinder-based implementation. 
I think it can be looked at separately from this change.

jakobbotsch:
I'll take a look @VSadov. ~~In the mean time, it looks like it works with `DOTNET_TieredCompilation=0`, so you can try that to get the measurements~~ opened a PR with the fix.

src/coreclr/vm/runtimesuspension.cpp
```diff
@@ -611,35 +642,174 @@ void UnregisterTasklet(Tasklet* pTasklet)
...
+{
+    if (!g_pConfig->TaskletAging())
+    {
+        return;
+    }
+
+    CrstHolder crstHolder(&g_taskletCrst);
+    for (Tasklet* pCurStack = g_pTaskletSentinel->pTaskletNextInLiveList; pCurStack != g_pTaskletSentinel; pCurStack = pCurStack->pTaskletNextInLiveList)
+    {
+        if (pCurStack->minGeneration > condemned)
```
davidwrighton, line 706: How does this handle cases such as having a pointer to a GC object which has been pinned into Gen0, such that it stays in Gen0 from GC to GC?

VSadov: > having a pointer to a GC object which has been pinned into Gen0

I think you are referring to "demotion" scenario, which is really a promotion that does not promote all survivors due to pinning.   
Sometimes GC does not want to move generation boundaries past a bunch of pinned objects with lots of empty space in between as it wants to keep that space available for user allocations. If that happens GC calls `GcDemote` meaning that some objects got older, while a few stayed young. In such case we could either:

1. just make min age of everything affected in this GC to be `0` for simplicity. They will re-age in later GCs.
1. dig through objects and figure the actual minimum age in each tasklet. 

I am doing `#2`, since it is easy to re-walk the objects in a tasklet, so we can do that, but no real preference between `#1` and `#2`  
([with real stacks I'd do `#1` as enumerating objects on a real stack is more involved](https://github.com/dotnet/runtime/pull/95221/files#r1409921150))

This is also the strategy used by handles - on demotion reexamine every referenced object and figure the new min age in clumps. Demotions could be somewhat expensive, but they are also rare.

VSadov:
I have added GC pause/allocation observations to the write-up as a part of this PR:

https://github.com/dotnet/runtimelab/pull/2425/files#r1412337061

docs/design/features/runtime-handled-tasks.md
```diff
@@ -618,3 +618,55 @@ In the current experiment we have looked into two implementing strategies:
 ## On unmanaged pointers.
 We cannot possibly track unmanaged pointers that refer to the stack across suspensions. Disallowing such scenarios would not be a take-back though since currently C# does not allow `await` in unsafe context, therefore await and unmanaged pointers cannot mix.
 
+## Impact on GC (heap size, pauses).
```
VSadov, line 621: The writeup on GC pauses/allocation impact.

davidwrighton (approved):

VSadov:
Thanks!

PR #2429, opened 2023-11-17
[Async-2] Null out TLS head continuation reference

jakobbotsch:
Avoid permanently holding on to a reference to the continuation chain from the previous suspension in TLS.

(I don't think this particularly affects the results of #2425, but seems like we shouldn't be doing it in any case.)

### DIFF
--- a/src/coreclr/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreCLR.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreCLR.cs
@@ -466,7 +466,9 @@ namespace System.Runtime.CompilerServices
         {
             ref RuntimeAsyncAwaitState state = ref t_runtimeAsyncAwaitState;
             awaitableProxy._notifier = state.Notifier;
-            return state.SentinelContinuation!.Next!;
+            Continuation head = state.SentinelContinuation!.Next!;
+            state.SentinelContinuation.Next = null;
+            return head;
         }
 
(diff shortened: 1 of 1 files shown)


src/coreclr/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreCLR.cs
```diff
@@ -466,7 +466,10 @@ private static Continuation GetHeadContinuation(AwaitableProxy awaitableProxy)
         {
             ref RuntimeAsyncAwaitState state = ref t_runtimeAsyncAwaitState;
             awaitableProxy._notifier = state.Notifier;
-            return state.SentinelContinuation!.Next!;
+            Continuation sentinelContinuation = state.SentinelContinuation!;
+            Continuation head = sentinelContinuation.Next!;
+            sentinelContinuation.Next = null;
```
VSadov, line 471: Perhaps rename the function to something like `UnlinkHeadContinuation` - to imply this cannot be done twice.

VSadov (approved):
LGTM

davidwrighton (approved):

src/coreclr/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreCLR.cs
jakobbotsch: Good idea, renamed!

jakobbotsch:
Ugh, accidentally merged this as a merge instead of squash...

PR #2431, opened 2023-11-18
[Async-2] A few small tweaks on the managed side of state-machine based implementation.

VSadov:
A few  tweaks:

- made AwaitableProxy a struct
because we can
- unlink Notifier  
to not keep it rooted in TLS longer than necessary
- use ExceptionDispatchInfo to rethrow
to preserve stacks/watson bucketing info etc...

### DIFF
--- a/src/coreclr/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreCLR.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreCLR.cs
@@ -504,13 +504,10 @@ namespace System.Runtime.CompilerServices
             {
                 Continuation headContinuation = UnlinkHeadContinuation(out var awaitableProxy);
                 await awaitableProxy;
-                Continuation? finalResult = DispatchContinuations(headContinuation, out Exception? ex);
+                Continuation? finalResult = DispatchContinuations(headContinuation);
                 if (finalResult != null)
                 {
                     Debug.Assert(finalResult == finalContinuation);
-                    if (ex != null)
-                        ExceptionDispatchInfo.Capture(ex).Throw();
-
                     if (IsReferenceOrContainsReferences<T>())
                     {
                         return (T)finalResult.GCData![0];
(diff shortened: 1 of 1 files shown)


VSadov:
I did not see any impact on benchmarks, just a few things that I noticed while getting more familiar with the state-machine based implementation.

src/coreclr/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.CoreCLR.cs
```diff
@@ -492,18 +500,16 @@ private static async Task<T> FinalizeTaskReturningThunk<T>(Continuation continua
...
-                Continuation headContinuation = UnlinkHeadContinuation(awaitableProxy);
+                Continuation headContinuation = UnlinkHeadContinuation(out var awaitableProxy);
                 await awaitableProxy;
                 Continuation? finalResult = DispatchContinuations(headContinuation, out Exception? ex);
                 if (finalResult != null)
                 {
                     Debug.Assert(finalResult == finalContinuation);
                     if (ex != null)
-                        throw ex;
+                        ExceptionDispatchInfo.Capture(ex).Throw();
```
jakobbotsch: Instead of this I think we can clean things up a bit more by rethrowing (with `throw;`) from `DispatchContinuations`. It should also avoid this one `ExceptionDispatchInfo`.

VSadov: It also seems a bit redundant to throw - just so that async machine would catch ex and put in the task, but I am not sure there is a way around that unless this method implements async completely by hand. I think that would be too far into diminishing returns for a prototype.

VSadov: rethrowing from within `DispatchContinuations` seems interesting. I will take a look, since I am touching this anyways.

jakobbotsch: > It also seems a bit redundant to throw - just so that async machine would catch ex and put in the task, but I am not sure there is a way around that unless this method implements async completely by hand. I think that would be too far into diminishing returns for a prototype.

Yeah, presumably we would end up with some more custom fitted code to handle this... originally I just used a `TaskCompletionSource`, but it didn't handle flowing all the contexts correctly.
I think we can definitely squeeze some more perf out here in the future -- we currently also seem to pay quite a bit of cost for async machinery that is optimized for cases we never hit here.

jakobbotsch (approved):

VSadov:
I assume this is ok to merge?

jakobbotsch:
> I assume this is ok to merge?

Yes, it looks fine to me (already approved it)

Issue #2434, opened 2023-11-22
[NativeAOT LLVM] Support dotnet.js


maraf:
- [x] Building naot flavored dotnet.js
- [x] Enable dotnet.js for naot-llvm apps targeting `browser-wasm`
- [x] JS interop
  - [x] `Interop.Runtime.cs` as wasm imports (instead of icalls)
  - [x] Marshaling
    - [x] JS marshallers should expect wasm import/export instead of Mono reflection found methods
    - [x] Copy string to heap instead of passing MonoString ref to JS (already as feature flag in upstream)
    - [x] Tasks
    - [x] Delegates
    - [x] JSObject
  - [x] JSImport
  - [x] JSExport
    - [x] Generated `__Register_` as wasm export
    - [x] Generated wrapper as wasm export
  - [x] Deliver the interop generator outside of app.ref pack
- [ ] Main
  - [x] Export user written Main (possibly async) as wasm export (we need to do JS marshaling for it). Generated `managed_main` is also responsible for storing `args` and `tearingdown` the runtime
  - [ ] Store args, so they are available from `Environment.GetCommandLineArgs()`
- [x] Yield to JS event loop
- [x] Timers
- [x] Allow to call UCO exports before Main (aka library mode)
- [ ] Add support for environment variables
- [ ] Add support for runtime options
- [x] Move runtime JS files to `/aotsdk` instead of `/framework`
- [ ] Find a proper solution for `Interop.Runtime.NativeAOT.cs` and other switches in libs (libs can't use switch based on runtime flavor)
- [ ] Run library tests
  - [ ] System.Runtime.InteropServices.JavaScript

Smoke test with currently supported feature set
https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/src/tests/nativeaot/SmokeTests/DotnetJs

Original prototype 
https://github.com/maraf/MinimalDotNetWasmNativeAOT/tree/dotnetjs/DotnetJsHack

maraf:
cc @dotnet/nativeaot-llvm @pavelsavara @ivanpovazan

SingleAccretion:
A few comments on the overall architecture.

>  Buidling naot flavored dotnet.js

The current NAOT-LLVM build is very simple (pass a few things to `emcc`, get `.js` back). It would be very valuable to ensure it remains so.

[...]

PR #2436, opened 2023-11-23
[NativeAOT LLVM] Mark all `JSExport` wrapper methods as `UnmanagedCallersOnly`

maraf:
- Add `UnmanagedCallersOnly` to every generated for `JSExport`
- Generate unique entrypoint name in form of `[assembly]namespace/class:method_args_hash` with replaced special chars (the pattern comes from the mono variant)
- Generate registration method with `UnmanagedCallersOnly`

<details>
  <summary>Example of generated class</summary>

```c#
// <auto-generated/>
namespace System.Runtime.InteropServices.JavaScript
{
    [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute]
    unsafe class __GeneratedInitializer
    {
        [global::System.ThreadStaticAttribute]
        static bool initialized;
        [global::System.Runtime.CompilerServices.ModuleInitializerAttribute]
        static internal void __Net7SelfInit_()
        {
            if (Environment.Version.Major == 7)
            {
                __Register_();
            }
        }

        [global::System.Diagnostics.CodeAnalysis.DynamicDependencyAttribute("__Wrapper_Greeting", "Xyz.Interop.MyClass", "BrowserConsoleApp")]
        static void __Register_()
        {
            if (initialized || global::System.Runtime.InteropServices.RuntimeInformation.OSArchitecture != global::System.Runtime.InteropServices.Architecture.Wasm)
                return;
            initialized = true;
            global::System.Runtime.InteropServices.JavaScript.JSFunctionBinding.BindManagedFunction("[BrowserConsoleApp]Xyz.Interop/MyClass:Greeting", 490867262, new global::System.Runtime.InteropServices.JavaScript.JSMarshalerType[] { global::System.Runtime.InteropServices.JavaScript.JSMarshalerType.Int32 });
        }

        [global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute(EntryPoint = "BrowserConsoleApp__GeneratedInitializer__Register_")]
        static void __Register__Export()
        {
            __Register_();
        }
    }
}
namespace Xyz
{
    public unsafe partial class Interop
    {
        public unsafe partial class MyClass
        {
            [global::System.Diagnostics.DebuggerNonUserCode]
            [global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute(EntryPoint = "_5B_BrowserConsoleApp_5D_Xyz_Interop_2F_MyClass_3A_Greeting_490867262")]
            internal static unsafe void __Wrapper_Greeting(global::System.Runtime.InteropServices.JavaScript.JSMarshalerArgument* __arguments_buffer)
            {
                ref global::System.Runtime.InteropServices.JavaScript.JSMarshalerArgument __arg_exception = ref __arguments_buffer[0];
                ref global::System.Runtime.InteropServices.JavaScript.JSMarshalerArgument __arg_return = ref __arguments_buffer[1];
                int __retVal = default;
                try
                {
                    __retVal = Xyz.Interop.MyClass.Greeting();
                    __arg_return.ToJS(__retVal);
                }
                catch (global::System.Exception ex)
                {
                    __arg_exception.ToJS(ex);
                }

                // Marshal - Convert managed data to native data.
                __arg_return.ToJS(__retVal);
            }
        }
    }
}

```
</details>

### DIFF
--- a/src/libraries/System.Runtime.InteropServices.JavaScript/gen/JSImportGenerator/Constants.cs
+++ b/src/libraries/System.Runtime.InteropServices.JavaScript/gen/JSImportGenerator/Constants.cs
@@ -17,6 +17,7 @@ namespace Microsoft.Interop.JavaScript
         public const string ModuleInitializerAttributeGlobal = "global::System.Runtime.CompilerServices.ModuleInitializerAttribute";
         public const string CompilerGeneratedAttributeGlobal = "global::System.Runtime.CompilerServices.CompilerGeneratedAttribute";
         public const string DynamicDependencyAttributeGlobal = "global::System.Diagnostics.CodeAnalysis.DynamicDependencyAttribute";
+        public const string UnmanagedCallersOnlyAttributeGlobal = "global::System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute";
         public const string ThreadStaticGlobal = "global::System.ThreadStaticAttribute";
         public const string TaskGlobal = "global::System.Threading.Tasks.Task";
         public const string SpanGlobal = "global::System.Span";
--- a/src/libraries/System.Runtime.InteropServices.JavaScript/gen/JSImportGenerator/JSExportGenerator.cs
+++ b/src/libraries/System.Runtime.InteropServices.JavaScript/gen/JSImportGenerator/JSExportGenerator.cs
@@ -335,7 +335,7 @@ namespace Microsoft.Interop.JavaScript
                                     AttributeArgument(
                                         LiteralExpression(
                                             SyntaxKind.StringLiteralExpression,
-                                            Literal(assemblyName + initializerClass + initializerName)
+                                            Literal(FixupSymbolName(assemblyName + initializerClass + initializerName))
                                         )
                                     )
                                     .WithNameEquals(
--- a/src/libraries/System.Runtime.InteropServices.JavaScript/gen/JSImportGenerator/JSExportGenerator.cs
+++ b/src/libraries/System.Runtime.InteropServices.JavaScript/gen/JSImportGenerator/JSExportGenerator.cs
@@ -35,6 +35,8 @@ namespace Microsoft.Interop.JavaScript
 
         public void Initialize(IncrementalGeneratorInitializationContext context)
         {
+            var assemblyName = context.CompilationProvider.Select(static (c, _) => c.AssemblyName);
+
             // Collect all methods adorned with JSExportAttribute
             var attributedMethods = context.SyntaxProvider
                 .ForAttributeWithMetadataName(Constants.JSExportAttribute,
@@ -103,7 +105,8 @@ namespace Microsoft.Interop.JavaScript
                 .Collect();
 
             IncrementalValueProvider<string> registration = regSyntax
-                .Select(static (data, ct) => GenerateRegSource(data))
+                .Combine(assemblyName)
+                .Select(static (data, ct) => GenerateRegSource(data.Left, data.Right))
                 .Select(static (data, ct) => data.NormalizeWhitespace().ToFullString());
 
             IncrementalValueProvider<ImmutableArray<(string, string)>> generated = generateSingleStub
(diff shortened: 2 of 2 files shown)


pavelsavara:
Motivation is that NAOT can't use reflection to discover the wrapper, right ?

Could you please explain how would be the UCO symbol called ?

maraf:
> Could you please explain how would be the UCO symbol called ?

Example here https://github.com/maraf/MinimalDotNetWasmNativeAOT/blob/dotnetjs/DotnetJsHack/dotnet.runtime.js#L4939

pavelsavara (approved):

SingleAccretion (approved):
LGTM (to the limited extent I can review this code).

I have a question, for my education: why is the `__Register_` method set up the way it is (i. e. why not call it from the generated wrapper under an initialization check)?

Related: `[ModuleInitializerAttribute]` is anti-pattern in NAOT and should be avoided if at all possible.

Another consideration: the UCOs are only exported by default for input assembly with ILC. What is the plan for making `JSExport`s from (C#) libraries work?

pavelsavara:
> I have a question, for my education: why is the `__Register_` method set up the way it is (i. e. why not call it from the generated wrapper under an initialization check)?

I'm not sure I understand the "under an initialization check", do you mean lazily ?

The reason is that JS side doesn't even know that the namespace and the function exist until they are bound by C# side.

> Related: `[ModuleInitializerAttribute]` is anti-pattern in NAOT and should be avoided if at all possible.

Yes, we learned that lesson and fixed it for Net8 and later, where we only initialize it when JS calls `const exports = await getAssemblyExports("assemblyName");` for particular assembly. But Net7 is not that smart and so we have this fallback.

The late bound approach avoids running many static constructors until you actually need it.

pavelsavara:
@maraf looking at `BrowserConsoleApp__GeneratedInitializer__Register` UCO, I think we need to support assembly name with dots in it.

pavelsavara:
> Related: `[ModuleInitializerAttribute]` is anti-pattern in NAOT and should be avoided if at all possible.

Perhaps we could teach some tool to trim it for Net8 and above.

SingleAccretion:
> I'm not sure I understand the "under an initialization check", do you mean lazily ?

(Yes, `void __Wrapper_Greeting() { if (!initialized) __Register_(); ... }`)

> The reason is that JS side doesn't even know that the namespace and the function exist until they are bound by C# side.

And this is because `getAssemblyExports` doesn't ask for a particular export, but for all of them from a particular assembly, and you want to support "nice" JS syntax (`assembly.ClassName.MyExport()` instead of `assembly.GetExport("ClassName", "MyExport").Invoke()`-sort of thing), and you cannot "hook" JS object member resolution?

pavelsavara:
> you want to support "nice" JS syntax

Yes.

> you cannot "hook" JS object member resolution?

Eventually we could use [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to do that hooking. It's known to be slow, but I'm not sure how much slow.

pavelsavara:
We can gradually improve it. Problem is backward compatibility. To support code which people already generated with old version of the generator and it's inside of some nuget package. Or find the courage to break it for them.

SingleAccretion:
Makes sense.

I don't personally see any big problem with this scheme. It does mean you have to eagerly initialize all the exports for a particular assembly, but perhaps there shouldn't be that many of them in the regular case.

pavelsavara:
I wonder if `runtimelab` would self-consume this generator or keep using the `runtime` version (in SDK).
Do you plan for merge this to `runtime` repo @maraf  ?

SingleAccretion:
> I wonder if runtimelab would self-consume this generator or keep using the runtime version (in SDK).

There is also the question of how would you consume it outside `runtime[lab]`. Some sort of package reference override?

pavelsavara:
> > I wonder if runtimelab would self-consume this generator or keep using the runtime version (in SDK).
> 
> There is also the question of how would you consume it outside `runtime[lab]`. Some sort of package reference override?

In Blazor we support non-AOT, non-workload dev loop (and also deployment). In that mode UCO would not show up at all as a native symbol in the .wasm file. So we need mono reflection there.

My concern is about codebase consistency.

jkotas:
> Perhaps we could teach some tool to trim it for Net8 and above.

`System.Runtime.InteropServices.JavaScript` ships "inbox" with netcoreapp, so the generator should be only concerned around with the current netcoreapp versions. It should not run for downlevel versions - the downlevel versions should use the generator that shipped with them. It means that the current (net9) version of the generator does not need to worry about targeting net7 or net8 runtime. Thoughts?

pavelsavara:
> Thoughts?

A) user code which used the generator in Net7 and the compiled code still lives in some nuget package. And should run with latest runtime.

B) new code generated with latest SDK to produce nuget package targeting Net7 need also work.

Are you saying that for B) the Net7 generator will be used and I don't need to worry about it anymore ?

pavelsavara:
If so, this is worth generator patch for Net8, to not generate that `ModuleInitializerAttribute`. 

We will need to find some alternative way how to keep this initializer code from always being trimmed. Because ILLink has no idea this will be called from JS. Currently `ModuleInitializerAttribute` keeps it alive.

jkotas:
> Are you saying that for B) the Net7 generator will be used and I don't need to worry about it anymore ?

Yes.

Try building Net7 app using .NET 8 SDK with `/v:diag` option. You should see `JSImportGenerator` or `LibraryImportGenerator` coming from Microsoft.NETCore.App.Ref 7.0.* .

jkotas:
> If so, this is worth generator patch for Net8, to not generate that ModuleInitializerAttribute.

We should do this for Net9. I am not sure whether it would meet the servicing bar for Net8.

pavelsavara:
And it could be done here too. Because UCO will keep it alive, right ?

yowl:
UCO will root the method as far as the Ilc method tree walking goes.  I don't have any other concerns here, LGTM.

maraf:
> I wonder if `runtimelab` would self-consume this generator or keep using the `runtime` version (in SDK). Do you plan for merge this to `runtime` repo @maraf?

We should be able to override `KnownFrameworkReference` to get `JSImportGenerator` and `S.R.IS.Javascript.dll` (aka `TargetingPack`) built from runtimelab

Issue #2240, opened 2023-03-26
[NativeAOT-LLVM] Implement finalization


SingleAccretion:
Currently, finalization does not work on WASM as we cannot use the finalizer thread mechanism.

In principle, it should be possible to run the finalizers synchronously, on the same thread that started the GC. In practice, a naive implementation will result in a deadlock because `RhEnableFinalization` is called before the "more space lock" is released by the GC on some code paths (`gc_heap::allocate`). Presumably, this is the same reason GC stress is disabled on the finalizer thread.

Still looking for a place to thread things through...

jkotas:
> it should be possible to run the finalizers synchronously, on the same thread that started the GC. 

Is it what Mono does currently?

SingleAccretion:
> Is it what Mono does currently?

I looked and it is not. Mono schedules the finalizers to be run on a small queue with [setTimeout](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout), with the delay parameter set to zero, along with a number of other "background" tasks.

jkotas:
I think it is the only way to it reliably. It is risky to start running finalizers on a thread in arbitrary state. For example, the thread may be in the middle of updating a datastructure that the finalizer will try to access as well. It will result either into a deadlock or a data corruption.

SingleAccretion:
> I think it is the only way to it reliably. It is risky to start running finalizers on a thread in arbitrary state. For example, the thread may be in the middle of updating a datastructure that the finalizer will try to access as well. It will result either into a deadlock or a data corruption.

Well, I am not sure how the event loop-driven scheme would ameliorate this issue. Canonically, control would yield to the event loop (and these "background" tasks would run) when code hits an `await` suspension point.

The problem with `setTimeout` is that it is not portable to synchronous code, where we don't yield to the event loop. This is the case for WASI and a certain subset of code one could write for NodeJS-like environments.

I am considering pulsing the finalizer queue on the exit of the slow path to allocation helpers right now.

pavelsavara:
To make the runtime useful in browser, most of interop with JavaScript requires that you don't block the browser's event loop.
That is, in order to run any UI or make HTTP and WebSocket requests, you have to `yield` somehow from your `Task` queue.

I my opinion finalizer is "small fish" compared to those problems.

PR #2440, opened 2023-11-24
[NativeAOT LLVM] Enable JS interop

maraf:
- The goal is not to enable all JS import/export scenarios, but start with some simple ones
- Updated smoke tests
- The `mono_wasm_bind_js_function` and `mono_wasm_bind_cs_function` are duplicated for NativeAOT case
- Due to `DisableImplicitFrameworkReferences=true` in tests, generators are added as explicit references

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
@@ -25,6 +25,7 @@
     <EmccExportedRuntimeMethod Include="addFunction" />
     <EmccExportedRuntimeMethod Include="UTF8ToString" />
     <EmccExportedRuntimeMethod Include="UTF8ArrayToString" />
+    <EmccExportedRuntimeMethod Include="UTF16ToString" />
     <EmccExportedRuntimeMethod Include="stringToUTF8Array" />
     <EmccExportedRuntimeMethod Include="FS_createPath" />
     <EmccExportedRuntimeMethod Include="FS_createDataFile" />
@@ -53,7 +54,7 @@
       <_DotNetJsLinkerFlag Include="-Wl,--export,__main_argc_argv" /><!-- Export main, a temporal solution until we have a way to export async main -->
       <_DotNetJsLinkerFlag Include="-s EXPORT_ES6=1" /><!-- Produce ES6 module as expented by hosting API -->
       <_DotNetJsLinkerFlag Include="-s MODULARIZE=1" /><!-- Produce ES6 module as expented by hosting API -->
-      <_DotNetJsLinkerFlag Include="-s INVOKE_RUN=0" /><!-- Don't automatically run main, it is called explicitly by ser -->
+      <_DotNetJsLinkerFlag Include="-s INVOKE_RUN=0" /><!-- Don't automatically run main, it is called explicitly by user -->
       <_DotNetJsLinkerFlag Include="-s EXPORT_NAME=&quot;'createDotnetRuntime'&quot;" /><!-- Export emscripten factory as 'createDotnetRuntime', as expented by hosting API -->
       <_DotNetJsLinkerFlag Include="-s ENVIRONMENT=&quot;'web,webview,worker,node,shell'&quot;" /><!-- Add support for all enviroments (default list is smaller) -->
       <_DotNetJsLinkerFlag Condition="'$(EmccEnvironment)' != ''" Include="-s ENVIRONMENT=&quot;$(EmccEnvironment)&quot;" />
--- a/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
@@ -25,7 +25,6 @@
     <EmccExportedRuntimeMethod Include="addFunction" />
     <EmccExportedRuntimeMethod Include="UTF8ToString" />
     <EmccExportedRuntimeMethod Include="UTF8ArrayToString" />
-    <EmccExportedRuntimeMethod Include="UTF16ToString" />
     <EmccExportedRuntimeMethod Include="stringToUTF8Array" />
     <EmccExportedRuntimeMethod Include="FS_createPath" />
     <EmccExportedRuntimeMethod Include="FS_createDataFile" />
--- a/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <LinkNativeDependsOn>$(LinkNativeDependsOn);PrepareDotNetJsApiForLinking</LinkNativeDependsOn>
     <NativeBinary>$(NativeOutputPath)dotnet.native.js</NativeBinary>
-    <IlcExportUnmanagedEntrypoints>true</IlcExportUnmanagedEntrypoints>
+    <IlcExportUnmanagedEntrypoints>true</IlcExportUnmanagedEntrypoints><!-- For JSExports -->
   </PropertyGroup>
   <ItemGroup>
     <EmccExportedFunction Include="_free" />
@@ -85,7 +85,7 @@
     <ItemGroup>
       <_FilesToCopyToNative Include="$(IlcFrameworkNativePath)\dotnet*.js" />
       <_FilesToCopyToNative Include="$(IlcFrameworkNativePath)\dotnet*.map" Condition="'$(WasmEmitSourceMap)' == 'true'" />
-      <_FilesToCopyToNative Include="@(WasmExtraFilesToDeploy)" /><!-- Use defined extra files to be included in the build output -->
+      <_FilesToCopyToNative Include="@(WasmExtraFilesToDeploy)" /><!-- User defined extra files to be included in the build output -->
     </ItemGroup>
     <Copy SourceFiles="@(_FilesToCopyToNative)" DestinationFolder="$(NativeOutputPath)" />
   </Target>
--- a/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
@@ -79,7 +79,14 @@
     <ItemGroup>
       <CustomLinkerArg Include="@(_DotNetJsLinkerFlag)" />
     </ItemGroup>
-    <WriteLinesToFile File="$(ExportsFile)" Lines="@(EmccExportedFunction)" Overwrite="false" Encoding="utf-8" Condition="'$(ExportsFile)' != ''" />
+    <ReadLinesFromFile File="$(ExportsFile)" Condition="'$(ExportsFile)' != ''">
+      <Output TaskParameter="Lines" ItemName="_ExistingExports" />
+    </ReadLinesFromFile>
+    <ItemGroup Condition="'$(ExportsFile)' != ''">
+      <_ExportsToAddToExportsFile Include="@(EmccExportedFunction)" />
+      <_ExportsToAddToExportsFile Remove="@(_ExistingExports)" />
+    </ItemGroup>
+    <WriteLinesToFile File="$(ExportsFile)" Lines="@(_ExportsToAddToExportsFile)" Overwrite="false" Encoding="utf-8" Condition="'@(_ExportsToAddToExportsFile->Count())' != '0'" />
   </Target>
   <Target Name="CopyDotnetJsAfterLinking" AfterTargets="LinkNativeLlvm">
     <ItemGroup>
--- a/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
@@ -1,7 +1,8 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <LinkNativeDependsOn Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' == 'true'">$(LinkNativeDependsOn);PrepareDotNetJsApiForLinking</LinkNativeDependsOn>
+    <LinkNativeDependsOn>$(LinkNativeDependsOn);PrepareDotNetJsApiForLinking</LinkNativeDependsOn>
     <NativeBinary>$(NativeOutputPath)dotnet.native.js</NativeBinary>
+    <IlcExportUnmanagedEntrypoints>true</IlcExportUnmanagedEntrypoints>
   </PropertyGroup>
   <ItemGroup>
     <EmccExportedFunction Include="_free" />
@@ -72,12 +73,13 @@
       <!-- Exported functions required by user or runtime API -->
       <_DotNetJsLinkerFlag Include="-s DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=$(_EmccExportedLibraryFunction)" Condition="'$(_EmccExportedLibraryFunction)' != ''" />
       <_DotNetJsLinkerFlag Include="-s EXPORTED_RUNTIME_METHODS=$(_EmccExportedRuntimeMethods)" />
-      <_DotNetJsLinkerFlag Include="-s EXPORTED_FUNCTIONS=$(_EmccExportedFunctions)" />
+      <_DotNetJsLinkerFlag Include="-s EXPORTED_FUNCTIONS=$(_EmccExportedFunctions)" Condition="'$(ExportsFile)' == ''" /><!-- Multiple args in the same rsp file is not support, add functions to app exports -->
       <_DotNetJsLinkerFlag Include="$(EmccExtraLDFlags)" />
     </ItemGroup>
     <ItemGroup>
       <CustomLinkerArg Include="@(_DotNetJsLinkerFlag)" />
     </ItemGroup>
+    <WriteLinesToFile File="$(ExportsFile)" Lines="@(EmccExportedFunction)" Overwrite="false" Encoding="utf-8" Condition="'$(ExportsFile)' != ''" />
   </Target>
   <Target Name="CopyDotnetJsAfterLinking" AfterTargets="LinkNativeLlvm">
     <ItemGroup>
--- a/src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
+++ b/src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
@@ -10,30 +10,28 @@ internal static partial class Interop
 {
     internal static unsafe partial class Runtime
     {
-#pragma warning disable SYSLIB1054
         [MethodImplAttribute(MethodImplOptions.InternalCall)]
         internal static extern void ReleaseCSOwnedObject(IntPtr jsHandle);
-        [DllImport("*", EntryPoint = "mono_wasm_bind_js_function")]
-        public static extern unsafe void BindJSFunction(string function_name, int function_name_length, string module_name, int module_name_length, void* signature, out IntPtr bound_function_js_handle, out int is_exception);
+        [LibraryImport("*", EntryPoint = "mono_wasm_bind_js_function", StringMarshalling = StringMarshalling.Utf16)]
+        public static unsafe partial void BindJSFunction(string function_name, int function_name_length, string module_name, int module_name_length, void* signature, out IntPtr bound_function_js_handle, out int is_exception);
         [MethodImpl(MethodImplOptions.InternalCall)]
         public static extern void InvokeJSFunction(IntPtr bound_function_js_handle, void* data);
-        [DllImport("*", EntryPoint = "mono_wasm_invoke_import")]
-        public static extern unsafe void InvokeImport(IntPtr fn_handle, void* data);
-        [DllImport("*", EntryPoint = "mono_wasm_bind_cs_function")]
-        public static extern unsafe void BindCSFunction(string fully_qualified_name, int fully_qualified_name_length, int signature_hash, void* signature, out int is_exception);
+        [LibraryImport("*", EntryPoint = "mono_wasm_invoke_import", StringMarshalling = StringMarshalling.Utf16)]
+        public static unsafe partial void InvokeImport(IntPtr fn_handle, void* data);
+        [LibraryImport("*", EntryPoint = "mono_wasm_bind_cs_function", StringMarshalling = StringMarshalling.Utf16)]
+        public static unsafe partial void BindCSFunction(string fully_qualified_name, int fully_qualified_name_length, int signature_hash, void* signature, out int is_exception);
         [MethodImpl(MethodImplOptions.InternalCall)]
         public static extern void ResolveOrRejectPromise(void* data);
         [MethodImpl(MethodImplOptions.InternalCall)]
         public static extern IntPtr RegisterGCRoot(IntPtr start, int bytesSize, IntPtr name);
         [MethodImpl(MethodImplOptions.InternalCall)]
         public static extern void DeregisterGCRoot(IntPtr handle);
-#pragma warning restore SYSLIB1054
 
         public static unsafe void BindJSFunction(string function_name, string module_name, void* signature, out IntPtr bound_function_js_handle, out int is_exception, out object result)
         {
             BindJSFunction(function_name, function_name.Length, module_name, module_name.Length, signature, out bound_function_js_handle, out is_exception);
             if (is_exception != 0)
-                result = "Runtime.BindJSFunction failed";
+                result = "Runtime.BindJSFunction failed"; // TODO-LLVM-JSInterop: Marshal exception message
             else
                 result = "";
         }
@@ -42,7 +40,7 @@ internal static partial class Interop
         {
             BindCSFunction(fully_qualified_name, fully_qualified_name.Length, signature_hash, signature, out is_exception);
             if (is_exception != 0)
-                result = "Runtime.BindCSFunction failed";
+                result = "Runtime.BindCSFunction failed"; // TODO-LLVM-JSInterop: Marshal exception message
             else
                 result = "";
         }
--- a/src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
+++ b/src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.JavaScript;
 
 internal static partial class Interop
 {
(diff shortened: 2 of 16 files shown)


elringus:
@maraf Sorry if I'm asking something irrelevant here, but do I understand correctly that there are plans to move WASM target from Mono to NativeAOT/LLVM? If that's the case, will the existing "high-level" JS interop API (the `JSImport/Export` stuff from `System.Runtime.InteropServices.JavaScript`) still work with the new backend or it'll be something different?

maraf:
> @maraf Sorry if I'm asking something irrelevant here, but do I understand correctly that there are plans to move WASM target from Mono to NativeAOT/LLVM? 

There is no such plan at the moment. Anyway NativeAOT/LLVM has come a long way and we are looking at comparison 

> If that's the case, will the existing "high-level" JS interop API (the `JSImport/Export` stuff from `System.Runtime.InteropServices.JavaScript`) still work with the new backend or it'll be something different?

The goal of the https://github.com/dotnet/runtimelab/issues/2434 is to have the same/close as possible API for both Mono and NativeAOT/LLVM in the context of the browser, so that is possible try to wasmbrowser based apps on NativeAOT/LLVM.

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -122,6 +123,119 @@ export function mono_wasm_bind_cs_function(fully_qualified_name: MonoStringRef,
     }
 }
 
+function mono_wasm_bind_cs_function_naot(fully_qualified_name: CharPtr, fully_qualified_name_length: number, signature_hash: number, signature: JSFunctionSignature, is_exception: Int32Ptr): void {
```
maraf: @SingleAccretion This is the best backward compatible approach I can think of. Do you have any better suggestion?

The function has a different signature, differently reads strings and does different thing in catch and finally clauses

src/libraries/System.Runtime.InteropServices.JavaScript/src/System.Runtime.InteropServices.JavaScript.csproj
```diff
@@ -22,7 +22,8 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetPlatformIdentifier)' == 'browser'">
-    <Compile Include="$(CommonPath)Interop\Browser\Interop.Runtime.cs" Link="System\Runtime\InteropServices\JavaScript\Interop\Interop.Runtime.cs" />
+    <Compile Include="$(CommonPath)Interop\Browser\Interop.Runtime.cs" Link="System\Runtime\InteropServices\JavaScript\Interop\Interop.Runtime.cs" Condition="'$(RuntimeFlavor)' != 'CoreCLR'" />
```
jkotas, line 26: libraries build system is not setup to build different binaries per runtime flavor. We have all libraries (except CoreLib) 100% shared between all runtimes flavors.

This change is ok as a quick hack for runtimelab, but it would not be upstreamable. Add a comment about it?

jkotas: (The proper way to do this would be to come up with a scheme that works for all runtime flavors and/or move any runtime specific parts into CoreLib.)

SingleAccretion: I would expect this can be solved similar to Linq.Expressions, with an internal feature switch (always off for Mono, always on for NativeAOT).

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
```diff
@@ -1,15 +1,36 @@
...
+    {
+        [JSImport("interop.math", "main.js")]
+        internal static partial int Math(int a, int b, int c);
+
+        [JSExport]
+        internal static int Square(int x)
+        {
+            Console.WriteLine($"Computing square of {x}");
+            return x^2;
+        }
```
SingleAccretion: Could you write more tests? For `JSExport`, you can take inspiration (aka copy code) from the `SharedLibrary` test - the main items of interest are EH and GC exercising methods.

For both `JSExport` and `JSImport`, it would be good to have some number of methods that exercise the actual marshaling (strings, arrays, tasks, delegates, basically,  challening cases from the table in https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/import-export-interop?view=aspnetcore-8.0). Perhaps these tests are already written somewhere and too could be copied...

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Runtime.InteropServices.JavaScript;
+
+internal static partial class Interop
+{
+    // WARNING: until https://github.com/dotnet/runtime/issues/37955 is fixed
+    // make sure that the native side always sets the out parameters
+    // otherwise out parameters could stay un-initialized, when the method is used in inlined context
+    internal static unsafe partial class Runtime
+    {
+#pragma warning disable SYSLIB1054
```
SingleAccretion: Any particular reason to not use `LibraryImport`?

There is a sense in which these low-level functions shouldn't be using marshalling anyway, but at least there is nothing expensive here.

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
pavelsavara: I guess it will take some time yet before you are able to run this https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.InteropServices.JavaScript/tests/System.Runtime.InteropServices.JavaScript.UnitTests/System/Runtime/InteropServices/JavaScript/JSImportExportTest.cs

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
```diff
@@ -0,0 +1,77 @@
...
+        public static extern void DeregisterGCRoot(IntPtr handle);
+#pragma warning restore SYSLIB1054
+
+        public static unsafe void BindJSFunction(string function_name, string module_name, void* signature, out IntPtr bound_function_js_handle, out int is_exception, out object result)
+        {
+            BindJSFunction(function_name, function_name.Length, module_name, module_name.Length, signature, out bound_function_js_handle, out is_exception);
+            if (is_exception != 0)
+                result = "Runtime.BindJSFunction failed";
+            else
+                result = "";
```
SingleAccretion: The original Mono functions were internal calls and would run in cooperative GC mode (i. e. as if they were managed code). `DllImport` is PInvoke, which puts the code into preemptive mode. I wonder if this will cause problems.

(Code in preemptive mode moslty cannot touch managed state).

src/mono/wasm/runtime/invoke-cs.ts
SingleAccretion: It is a lot of code. Let me see what all of it does.

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
SingleAccretion: Right. We don't need super-comprehensive testing here, but something that would fail if some basic aspect of the system gets broken.

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
SingleAccretion: > but at least there is nothing expensive here.

Actually, I take that back, there is UTF16 -> UTF8 marshalling going on.

It is quite inefficient and not really needed, since both C# and JS operate on 2-byte `char`s. Not something to be fixed downstream, though.

pavelsavara: This is passed as `MonoString*` in the mono icall. So I guess we should configure the LibraryImport/DllImport to do something similar or to use UTF16.

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
pavelsavara: what's the plan for running GC for NAOT on browser ? What would trigger it and what would schedule finalizers ?

SingleAccretion: > what's the plan for running GC for NAOT on browser ? What would trigger it and what would schedule finalizers ?

Same as on other platforms - you GC whenever GC decides its time to GC. Practically, this will happens on allocations, or explicit `GC.Collect`s.

Finalization is a little bit NYI right now (#2240), but it could follow the same plan (nice because WASI would _have_ to do that), or be hooked into the event loop.

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
maraf: I would rather use standard JS interop tests when we get there. I don't see value in duplicating them here. This is just a smoke test.

Also, a lot of JS interop scenarios don't work yet and are out of scope for this PR

src/mono/wasm/runtime/invoke-cs.ts
maraf: You look at the initial commit in this branch. That's the minimal change

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
SingleAccretion: Right, we should at least use UTF16, so that it pins the data instead of transcoding and copying it.

Ideally, upstream code would use the same approach, however, absent that we'll have to managed with `if (NativeAOT)` for now, i. e. something in the spirit of (@maraf):
```
; Warning: very wrong TS syntax
function mono_wasm_bind_js_function(function_name: any, ...)
    assert_bindings();
    if (!NativeAOT) {
    const function_name_root = mono_wasm_new_external_root<MonoString>(function_name),
        module_name_root = mono_wasm_new_external_root<MonoString>(module_name),
        resultRoot = mono_wasm_new_external_root<MonoObject>(result_address);

        const js_function_name = monoStringToString(function_name_root)!;
        const mark = startMeasure();
        const js_module_name = monoStringToString(module_name_root)!;
   } else {
        const js_function_name = utf16toString(function_name_root)!;
        const js_module_name = utf16toString(module_name_root)!;
   }
```

There is always a difficult act of touching less upstream code, but not so little that merging becomes conflict-free, yet broken functionality-wise. I think with these functions there is too much code (that we don't have expertise in, and that is fairly likely to change in the future) for them to be copied.

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
SingleAccretion: > Also, a lot of JS interop scenarios don't work yet and are out of scope for this PR

Right. I would expect that as they are enabled, they get added to this test. Could you add a few simple ones that already work then (like I would expect strings to work)?

One of the reasons why it is valuable to have this testing is that even when you enable the full libraries test suite, it will take > minute to compile, while this test will take much less. Libraries tests are not suitable for inner loop.

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
pavelsavara: Let's merge this here and see how it evolves. Once the dust settles, we could upstream some refactoring. For how I guess we don't really know how those "extension points" should look like in general and all the places we need them in.

SingleAccretion: Right, not suggesting we do any kind of refactoring right now, hence the `if (NativeAOT)` path.

pavelsavara: Maraf just tried to make nice diff as you asked him earlier.

src/mono/wasm/runtime/invoke-cs.ts
SingleAccretion: Alright, I looked at it and both `mono_wasm_bind_cs_function_naot` and `mono_wasm_get_assembly_exports_naot` look good to me. I would kindly ask to collapse `mono_wasm_bind_js_function` back with `if`s.

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -345,12 +488,30 @@ export async function mono_wasm_get_assembly_exports(assembly: string): Promise<
...
+async function mono_wasm_get_assembly_exports_naot(assembly: string): Promise<any> {
+    assert_bindings();
+    const result = exportsByAssembly.get(assembly);
+    if (!result) {
+        const mark = startMeasure();
+        
+        const register = (Module as any)["_" + assembly + "__GeneratedInitializer" + "__Register_"];
+        mono_assert(register, `Missing wasm export for JSExport registration function in assembly ${assembly}`);
+
+        endMeasure(mark, MeasuredBlock.getAssemblyExports, assembly);
```
SingleAccretion: > const mark = startMeasure();

Does it make sense to measure what is effectively a dictionary access, or is it still better for consistency?

src/mono/wasm/runtime/invoke-js.ts
```diff
@@ -9,7 +10,7 @@ import { get_signature_argument_count, bound_js_function_symbol, get_sig, get_si
 import { setI32, setI32_unchecked, receiveWorkerHeapViews } from "./memory";
 import { monoStringToString, stringToMonoStringRoot } from "./strings";
 import { MonoObject, MonoObjectRef, MonoString, MonoStringRef, JSFunctionSignature, JSMarshalerArguments, WasmRoot, BoundMarshalerToJs, JSFnHandle, BoundMarshalerToCs, JSHandle, MarshalerType } from "./types/internal";
-import { Int32Ptr } from "./types/emscripten";
+import { CharPtr, Int32Ptr } from "./types/emscripten";
```
SingleAccretion: Can additional TS imports be on separate lines to reduce conflicts, or does the language prohibit that?
```
import { Int32Ptr } from "./types/emscripten";
import { CharPtr } from "./types/emscripten";
```

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -134,6 +248,10 @@ function bind_fn_0V(closure: BindingClosure) {
         try {
             const args = alloc_stack_frame(2);
             // call C# side
+            if (NativeAOT) {
+                (method as any)(args); // TODO MF: handle exception
```
SingleAccretion:
```suggestion
                (method as any)(args); // TODO-LLVM-JSInterop: handle exception
```
Unique `TODO` postfixes make it easy to search for them.

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -157,6 +275,10 @@ function bind_fn_1V(closure: BindingClosure) {
             marshaler1(args, arg1);
 
             // call C# side
+            if (NativeAOT) {
+                (method as any)(args); // TODO MF: handle exception
```
SingleAccretion:
```suggestion
                (method as any)(args); // TODO-LLVM-JSInterop: handle exception
```

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -181,6 +303,12 @@ function bind_fn_1R(closure: BindingClosure) {
...
```
SingleAccretion:
```suggestion
                (method as any)(args); // TODO-LLVM-JSInterop: handle exception
```

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -210,6 +338,12 @@ function bind_fn_2R(closure: BindingClosure) {
             marshaler2(args, arg2);
 
             // call C# side
+            if (NativeAOT) {
+                (method as any)(args); // TODO MF: handle exception
```
SingleAccretion:
```suggestion
                (method as any)(args); // TODO-LLVM-JSInterop: handle exception
```

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -244,6 +378,15 @@ function bind_fn(closure: BindingClosure) {
             }
 
             // call C# side
+            if (NativeAOT) {
+                (method as any)(args); // TODO MF: handle exception
```
SingleAccretion:
```suggestion
                (method as any)(args); // TODO-LLVM-JSInterop: handle exception
```

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
pavelsavara: I guess we need to make decision about yielding to browser event loop. 
Big part of JS interop is moot without it (promises, network events, UI).

SingleAccretion: Well, I would expect the yielding to happen naturally in async code, as you hit `await`s and such. GC shouldn't really play much into it? We would of course have the Browser-specific thread pool implementation.

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
maraf: Strings for example don't work

SingleAccretion: Interesting that they don't. Well, let's leave this be until they do.

src/mono/wasm/runtime/invoke-cs.ts
maraf: It calls the generated register method which calls back to JS for binding individual exported methods

src/mono/wasm/runtime/invoke-js.ts
maraf: Good idea. It works. Thanks!

src/mono/wasm/runtime/invoke-cs.ts
SingleAccretion: Ahh, hmm, so it is this line that calls it:
```
const register = (Module as any)["_" + assembly + "__GeneratedInitializer" + "__Register_"];
```
?

maraf: No, the actual call was missing 😀
Sorry for confusion

maraf:
JSExport needs more work...

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
```diff
@@ -72,12 +73,13 @@
       <!-- Exported functions required by user or runtime API -->
       <_DotNetJsLinkerFlag Include="-s DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=$(_EmccExportedLibraryFunction)" Condition="'$(_EmccExportedLibraryFunction)' != ''" />
       <_DotNetJsLinkerFlag Include="-s EXPORTED_RUNTIME_METHODS=$(_EmccExportedRuntimeMethods)" />
-      <_DotNetJsLinkerFlag Include="-s EXPORTED_FUNCTIONS=$(_EmccExportedFunctions)" />
+      <_DotNetJsLinkerFlag Include="-s EXPORTED_FUNCTIONS=$(_EmccExportedFunctions)" Condition="'$(ExportsFile)' == ''" /><!-- Multiple args in the same rsp file is not support, add functions to app exports -->
```
SingleAccretion, line 76:
```suggestion
      <_DotNetJsLinkerFlag Include="-s EXPORTED_FUNCTIONS=$(_EmccExportedFunctions)" Condition="'$(ExportsFile)' == ''" /> <!-- Multiple EXPORTED_FUNCTIONS in the same rsp file are not support, add functions to app exports -->
```
What does "app exports" mean here - `$(EmccExportedFunctions)`?

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
```diff
@@ -72,12 +73,13 @@
...
       <_DotNetJsLinkerFlag Include="$(EmccExtraLDFlags)" />
     </ItemGroup>
     <ItemGroup>
       <CustomLinkerArg Include="@(_DotNetJsLinkerFlag)" />
     </ItemGroup>
+    <WriteLinesToFile File="$(ExportsFile)" Lines="@(EmccExportedFunction)" Overwrite="false" Encoding="utf-8" Condition="'$(ExportsFile)' != ''" />
```
SingleAccretion: Modifying an existing file is a bit scary. Does this work correctly when the inputs haven't changed (i. e. in incremental build)?

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
```diff
@@ -0,0 +1,77 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.JavaScript;
+
+internal static partial class Interop
+{
+    // WARNING: until https://github.com/dotnet/runtime/issues/37955 is fixed
+    // make sure that the native side always sets the out parameters
+    // otherwise out parameters could stay un-initialized, when the method is used in inlined context
```
SingleAccretion:
```suggestion
```
Not applicable to RyuJit.

src/libraries/Common/src/Interop/Browser/Interop.Runtime.NativeAOT.cs
```diff
@@ -0,0 +1,77 @@
...
+    internal static unsafe partial class Runtime
+    {
+#pragma warning disable SYSLIB1054
+        [MethodImplAttribute(MethodImplOptions.InternalCall)]
+        internal static extern void ReleaseCSOwnedObject(IntPtr jsHandle);
+        [DllImport("*", EntryPoint = "mono_wasm_bind_js_function")]
+        public static extern unsafe void BindJSFunction(string function_name, int function_name_length, string module_name, int module_name_length, void* signature, out IntPtr bound_function_js_handle, out int is_exception);
```
SingleAccretion: To come back to the marshalling point: this should use `[LibraryImport("*", EntryPoint = "mono_wasm_bind_js_function", StringMarshalling = StringMarshalling.Utf16)]`, with the JS side using `UTF16ToString` (which will need to be included into `EmccExportedRuntimeMethod`, I guess).

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -345,12 +412,35 @@ export async function mono_wasm_get_assembly_exports(assembly: string): Promise<
             // it doesn't have the __GeneratedInitializer class
             cwraps.mono_wasm_runtime_run_module_cctor(asm);
         }
+
```
SingleAccretion:
```suggestion
```
Unintentional whitespace change.

SingleAccretion (commented):
Just a few comments, we are quite close. Thank you for collapsing back TS changes!

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
maraf: UCO

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
maraf: Since `ExportsFile` is not input for `LinkNativeLlvm`, it's noop. But I could write to the file only if it doesn't contain `@(EmccExportedFunction)` already, for future enhancements

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
maraf: We need https://github.com/dotnet/runtime/pull/95180 for strings

src/mono/wasm/runtime/dotnet.d.ts
```diff
@@ -48,6 +48,7 @@ declare interface EmscriptenModule {
     getValue(ptr: number, type: string, noSafe?: number | boolean): number;
     UTF8ToString(ptr: CharPtr, maxBytesToRead?: number): string;
     UTF8ArrayToString(u8Array: Uint8Array, idx?: number, maxBytesToRead?: number): string;
+    UTF16ToString(ptr: CharPtr, maxBytesToRead?: number): string;
```
SingleAccretion: Not need with the `strings.ts` change?

src/mono/wasm/runtime/types/emscripten.ts
```diff
@@ -57,6 +57,7 @@ export declare interface EmscriptenModule {
     getValue(ptr: number, type: string, noSafe?: number | boolean): number;
     UTF8ToString(ptr: CharPtr, maxBytesToRead?: number): string;
     UTF8ArrayToString(u8Array: Uint8Array, idx?: number, maxBytesToRead?: number): string;
+    UTF16ToString(ptr: CharPtr, maxBytesToRead?: number): string;
```
SingleAccretion: Not needed?

SingleAccretion (approved):
LGTM modulo two questions and assuming green CI. Thank you!

I would also ask to add a note about the JSInterop library build workaround to #2434.

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
yowl: > We need [dotnet/runtime#95180](https://github.com/dotnet/runtime/pull/95180) for strings

Not in time for this PR, but I will start the next merge soon.

src/mono/wasm/runtime/dotnet.d.ts
maraf: Good catch!

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
maraf: It's ok, no rush on that one, I have other things to do :)

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -134,6 +173,14 @@ function bind_fn_0V(closure: BindingClosure) {
         try {
             const args = alloc_stack_frame(2);
             // call C# side
+            if (NativeAOT) {
```
pavelsavara: we have multiple copies of this code. Could we introduce `invoke_method_and_handle_exception_naot` ?

src/mono/wasm/runtime/roots.ts
```diff
@@ -60,6 +61,15 @@ export function mono_wasm_new_root_buffer_from_pointer(offset: VoidPtr, capacity
  * Releasing this root will not de-allocate the root space. You still need to call .release().
  */
 export function mono_wasm_new_external_root<T extends MonoObject>(address: VoidPtr | MonoObjectRef): WasmRoot<T> {
+    if (NativeAOT) {
```
pavelsavara, line 64: Since we don't need GC roots for `MonoString*` and we will delete legacy interop soon, I hope that we could figure out way how to get rid of all roots too (on main). That would make your life easier.

pavelsavara (approved):

src/mono/wasm/runtime/roots.ts
maraf: I'm looking forward to it

Issue #2414, opened 2023-10-12
NativeAOT-LLVM: Question: Should we drop WasmImport and use DllImport


yowl:
We have been discussing the `WasmImport` msbuild property for Wasm Components and think it would be a good time to revisit this.  `WasmImport` is used to generate the following LLVM:

```llvm
attributes #2 = { "wasm-import-module"="hellowasi" "wasm-import-name"="reverse" }
```
which is an attribute added to a function that generates the following wasm (WAT)
```wat
(import "hellowasi" "reverse" (func $reverse (type 4)))
```

Previous and open issues/PRs for this:
https://github.com/dotnet/runtimelab/issues/1390
https://github.com/dotnet/runtimelab/pull/1845
https://github.com/dotnet/runtimelab/pull/2410 

I don't know that the original decision to not use `DllImport` was based on anything other than thinking this was not exactly the same as `DllImport` so rather than cause any problem, I just created something new.

However now is good time, with Wasm component work starting, and wanting to support Mono, to revisit this and in particular look at whether we should switch to `DllImport` `Value` for the module name and `EntryPoint` for the import name (function name)

cc @pavelsavara , @AaronRobinsonMSFT , @dotnet/nativeaot-llvm , @silesmo

pavelsavara:
Only observation I have about this is that type system is quite limited for the WASM export/import
- no strings
- just part of numeric types
- no buffers, no structs
- but you can pass opaque JS object reference. Mono on browser doesn't use it right now.
- and pointers 32/64 bit question. Mono is now mostly assuming 32bit only.

I'm not sure that using DLLImport would lead users to assume it's business as usual talking to something that lives in the same address space.

cc @SteveSandersonMS @maraf @lambdageek

AaronRobinsonMSFT:
@yowl Good question. I think ideally we would use `DllImport`, but it does come with non-trivial UX cost. Source usage boils down to one of the following:

```csharp
#if WASM
[DllImport("module", EntryPoint = "func")]
#else
[DllImport("libc", EntryPoint = "func")]
#endif
static extern void Func();
```

or 

```csharp
#if WASM
[WasmImport("module", EntryPoint = "func")]
#else
[DllImport("libc", EntryPoint = "func")]
#endif
static extern void Func();
```

We already have `LibraryImport` and `JSImport`, both of which are what we could call the "modern" way of calling an external component, so a `WasmImport` does align with that approach.

[...]

PR #2444, opened 2023-11-29
[NativeAOT-LLVM] Implement WasmImportlinkage

yowl:
This PR implements the new `WasmImportlinkageAttribute` and remove the old policy based mechanism.  Added a test for the browser target as it is easier to fulfill the import in Javascript.

Added a key `ExternSymbolKey` to `_externSymbolsWithAccessors` to distinguish between Wasm imports and normal `DllImport`s, Wasm imports of the same function from different modules, and imports of the same name but with different signatures.

When writing the object files, added a new dictionary `_wasmImportLinkages` to track emitted functions with the wasm import attributes.  This allows us to remove duplicate imports as they behaviour for that is not defined.


yowl:
cc @dotnet/nativeaot-llvm

src/tests/nativeaot/SmokeTests/HelloWasm/shell.js
```diff
@@ -0,0 +1,130 @@
+// No copyright header as this is largely a cut and paste from the standard
+// Emscripten produced shell
```
SingleAccretion: Is all of the JS needed? It would be nice to trim it down to a more readable state.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -814,10 +814,13 @@ private void GetCodeForExternMethodAccessor(ExternMethodAccessorNode node)
...
+            foreach (MethodDesc methodDesc in node.EnumerateMethods())
             {
-                externFunc.AddFunctionAttribute("wasm-import-name", externFuncName);
-                externFunc.AddFunctionAttribute("wasm-import-module", wasmModuleName);
+                if (methodDesc.HasCustomAttribute("System.Runtime.InteropServices", "WasmImportLinkageAttribute"))
+                {
+                    externFunc.AddFunctionAttribute("wasm-import-name", externFuncName);
+                    externFunc.AddFunctionAttribute("wasm-import-module", methodDesc.GetPInvokeMethodMetadata().Module);
+                }
             }
```
SingleAccretion: I think more is needed to implement the behavior in https://github.com/dotnet/runtimelab/issues/2414#issuecomment-1772551025.

`WasmImportLinkage`-attributed methods create what is effectively a second namespace to the static symbols, so, for example:
```
[DllImport("xyz", EntryPoint = "xyz_func"), WasmImportLinkage]
void func_1();

[DllImport("*", EntryPoint = "xyz_func")]
void func_2();
```
Here, we will link in `func_2` statically, and `func_1` as an `xyz_func` import from the `xyz` module. It means signature mismatch and `AddFunctionAttribute("wasm-import-name")` must be segregated for these two categories too. For example:
```
[DllImport("xyz", EntryPoint = "xyz_func"), WasmImportLinkage]
void func_2(int arg);
[DllImport("xyz", EntryPoint = "xyz_func"), WasmImportLinkage]
void func_3();
```
Should report a warning (and choose which signature to import? Or import none?).

While this:
```
[DllImport("not_xyz", EntryPoint = "xyz_func"), WasmImportLinkage]
void func_2(int arg);
[DllImport("xyz", EntryPoint = "xyz_func"), WasmImportLinkage]
void func_3();
```
Should not report a warning, etc.

Naturally, a number of tests is needed to verify all these interactions.

docs/using-nativeaot/compiling.md
```diff
@@ -119,18 +119,13 @@ Note that assemblies other than the one being published (e. g. those from refere
 ```
 
 #### WebAssembly module imports
-Functions in other WebAssembly modules can be imported and invoked using `DllImport` e.g.
+Functions in other WebAssembly modules can be imported and invoked using `DllImport` and `WasmImportLinkage` e.g.
```
SingleAccretion, line 122: > `WasmImportLinkage`

Probably worth mentioning that:
```
`WasmImportLinkage` is only available when you build against the latest nightly SDK, although you can define it manually as well:
<copy paste minimal definition of WasmImportLinkage>
```

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: > Or import none?

I think that would be safer.

yowl: > must be segregated

I was thinking that I would change https://github.com/dotnet/runtimelab/blob/7c00d39843bfd00a5deb707701225cc5f033aff7/src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/LLVMCodegenNodeFactory.cs#L16 to have a key of a struct with the necessary fields to do this segregation.  Does that sound correct?

SingleAccretion: SGTM.

src/tests/nativeaot/SmokeTests/HelloWasm/shell.js
yowl: Have removed most of it.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: > Naturally, a number of tests is needed to verify all these interactions.

There are just 3 tests:

1. The original which just tests we place the import in the correct module
2. A test for not importing anything if there are duplciate module/func names
3. A test for the same function in different modules (a minor variation on 1 I suppose)

Do we want some others?

SingleAccretion: 1) Simple test that it works: calls its respective method.
   a) Lazy (i. e. just `[DllImport]`): throws PNSE.
      - It is fine to defer the work of actually throwing the exception from this PR.

   b) Static/WASM import call their respective methods.
2) Behavior in presence of overloading.
   a) Two static PIs for the same symbol with identical signatures, but different module names, work.
      - This will use multiple `DirectPInvoke`s.

   b) Two WASM imports for the same `EntryPoint`, but different modules, call different functions.
   c) Two PIs for the same symbol with ABI compatible signatures (`uint func(uint arg)`, `int func(int arg)`) work.
   d) Two PIs for the same symbol with ABI-incompatible signatures (`int func()`, `int func(int arg)`) warn.
      - Static behavior: the correct PI declaration still works.
      - WASM import behavior: all PI declarations throw `EntryPointNotFoundException`.
        - It is fine to defer the work of actually throwing the exception from this PR.
3) Interactions:
```
[DllImport("not_xyz", EntryPoint="func")]
void func_1(); // Throws PNSE

[DllImport("xyz", EntryPoint="func")] // <DirectPinvoke Include="xyz!func" />
void func_2(); // Calls the static function "func"

[DllImport("xyz", EntryPoint="func"), WasmImportLinkage]
void func_3(); // Calls the WASM import
```
I think tests are missing for `1`, `2a`, `2c`, `3`.

yowl: Thanks,

For 2a, are we going to synthesize the export name like `GlobalizationNative_ChangeCase ?

SingleAccretion: No, it'll be the same `EntryPoint`, i. e. for both:
```
[DllImport("xyz", EntryPoint = "func")
void func_1();

[DllImport("not_xyz", EntryPoint = "func")
void func_2();

<DirectPInvoke Include="xyz" />
<DirectPInvoke Include="not_xyz" />
```
They'll resolve to the same `func` static symbol.

yowl: Thanks, I'd completely forgotten about the changes we were going to do for DirectPInvoke, sorry.  I've added a few more tests but I don't think I've got the tests for PNSE, or signature warnings quite right.  We suppress `IL3049` for HelloWasm and I didn't really understand what is happening with the `switch` in `TestNativeCallsWithMismatchedSignatures`

docs/using-nativeaot/compiling.md
yowl: updated the docs with this, thanks

docs/using-nativeaot/compiling.md
```diff
@@ -119,21 +119,27 @@ Note that assemblies other than the one being published (e. g. those from refere
...
-<ItemGroup>
-  <WasmImport Include="wasi_snapshot_preview1!random_get" />
-</ItemGroup>
-```
-Will cause the above `random_get` to create this WebAssembly:
+This will create an import from the `wasi_snapshot_preview1` module with the function name `random_get`.  The `import` in the WAT would look like this:
 ```
 (import "wasi_snapshot_preview1" "random_get" (func $random_get (type 3)))
 ```
+Note: `WasmImportLinkageAttribute` is currently only available in the nightly SDK.  You can either build against the nightly SDK or you can defined this attribute in your code;
```
SingleAccretion:
```suggestion
Note: `WasmImportLinkageAttribute` is currently only available in the nightly SDK.  You can either build against the nightly SDK or you can define this attribute in your code:
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -127,6 +127,11 @@ The .NET Foundation licenses this file to you under the MIT license.
     <IlcLlvmExceptionHandlingModel Condition="'$(IlcLlvmExceptionHandlingModel)' == ''">emulated</IlcLlvmExceptionHandlingModel>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(NativeCodeGen)' == 'llvm'">
+    <DirectPInvoke Include="libSystem.Globalization.Native" />
+    <DirectPInvoke Include="libSystem.Native" />
+  </ItemGroup>
```
SingleAccretion: I think it's time for us to start using `Microsoft.NETCore.Native.Wasm.props` (which should be renamed to `.targets`), and add this there, similar to how it is for `Unix.targets`.

src/coreclr/tools/Common/TypeSystem/Ecma/EcmaMethod.cs
```diff
@@ -498,6 +498,9 @@ public override PInvokeMetadata GetPInvokeMethodMetadata()
             if ((ImplAttributes & MethodImplAttributes.PreserveSig) != 0)
                 attributes |= PInvokeAttributes.PreserveSig;
 
+            if(HasCustomAttribute("System.Runtime.InteropServices", "WasmImportLinkageAttribute"))
+                attributes |= PInvokeAttributes.WasmImportLinkage;
```
SingleAccretion: `EcmaMethod.cs` is core type system code, representing only ECMA-defined concepts. The attribute checking will have be done elsewhere else.

src/coreclr/tools/aot/ILCompiler.Compiler/ILCompiler.Compiler.csproj
```diff
@@ -449,7 +448,6 @@
     <Compile Include="Compiler\DependencyAnalysis\FieldMetadataNode.cs" />
     <Compile Include="Compiler\DependencyAnalysis\ILScanNodeFactory.cs" />
     <Compile Include="Compiler\DictionaryLayoutProvider.cs" />
-    <Compile Include="Compiler\DirectPInvokePolicy.cs" />
```
SingleAccretion: I just checked, `DirectPInvokePolicy.cs` is actually (unused) upstream code. So the conflict-free strategy here would be to not touch it in this change, merge it, and then delete it upstream.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -15,6 +15,7 @@
 using Internal.Text;
 using Internal.TypeSystem;
 using LLVMSharp.Interop;
+using static System.Net.Mime.MediaTypeNames;
```
SingleAccretion:
```suggestion
```

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/ExternMethodAccessorNode.cs
```diff
@@ -86,6 +92,88 @@ public override int CompareToImpl(ISortableNode other, CompilerComparer comparer
         protected override void EmitCode(NodeFactory factory, ref LoongArch64Emitter instructionEncoder, bool relocsOnly) => throw new NotImplementedException();
         protected override void EmitCode(NodeFactory factory, ref WasmEmitter instructionEncoder, bool relocsOnly) { }
 
-        protected override string GetName(NodeFactory context) => $"ExternMethodAccessor {ExternMethodName}";
+        protected override string GetName(NodeFactory context) => $"ExternMethodAccessor {ExternSymbolKey.ExternMethodName}";
+    }
+
+    // Wasm imports are different if any of module name, function name, ABI signature are different.
+    // Direct externs are different if the extern function name is different.
```
SingleAccretion: > Wasm imports are different if any of module name, function name, ABI signature are different

The WASM file format does support "overloaded" imports, but we do not (since LLVM does not, I believe). So the signature is not part of the symbol resolution.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -810,14 +814,30 @@ private void GetCodeForExternMethodAccessor(ExternMethodAccessorNode node)
             }
 
             LLVMModuleRef externFuncModule = _moduleWithExternalFunctions;
-            Debug.Assert(externFuncModule.GetNamedFunction(externFuncName).Handle == IntPtr.Zero);
-            LLVMValueRef externFunc = externFuncModule.AddFunction(externFuncName, externFuncType);
+            WasmImportKey wasmImportKey = new WasmImportKey(node.ExternSymbolKey);
+            bool externExists = _wasmImportLinkages.TryGetValue(wasmImportKey, out LLVMValueRef externFunc);
```
SingleAccretion: This I am struggling to understand the need for `_wasmImportLinkages`. The node factory should unique accessor nodes for the same WASM import, so why would we be adding the same one twice?

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/LLVMCodegenNodeFactory.cs
```diff
@@ -49,17 +49,19 @@ public override IMethodNode RuntimeExportManagedEntrypoint(string name)
 
         internal ExternMethodAccessorNode ExternSymbolWithAccessor(string name, MethodDesc method, ReadOnlySpan<TargetAbiType> sig)
         {
-            Dictionary<string, ExternMethodAccessorNode> map = _externSymbolsWithAccessors;
+            Dictionary<ExternSymbolKey, ExternMethodAccessorNode> map = _externSymbolsWithAccessors;
+            PInvokeMetadata pInvokeMethodMetadata = method.GetPInvokeMethodMetadata();
+            ExternSymbolKey key = new(pInvokeMethodMetadata.Module, name, pInvokeMethodMetadata.Flags.WasmImportLinkage, sig);
```
SingleAccretion: The layering here no longer quite makes sense. Originally, this only served static symbols, so we would go like so:
```
codegen asks LLVMCodegenCompilation
  for an accessor to a static PInvoke (via 'GetExternalMethodAccessor')
    LLVMCodegenCompilation asks ConfigurablePInvokePolicy
      for a direct symbol name for that static PI
    LLVMCodegenCompilation asks LLVMCodegenNodeFactory
      to produce a node representing the accessor, for that symbol name
```
Somewhat non-obviously, but importantly, `method/sig` here are auxiliary information essentially only used for diagnostic purposes. Now we are using `method` to essentially "skip over" the PI policy, which controls symbol resolution.

I think we can use the PI policy mechanism to our advantage and make this a lot cleaner:
1) Create a new `WasmConfigurablePInvokePolicy : ConfigurablePInvokePolicy` that would do override `GenerateDirectCall` somewhat like so:
```
public override bool GenerateDirectCall(MethodDesc method, out string externName)
{
    if (IsWasmImport(method))
    {
        externName = "__wasm_import_" + moduleName + "_" + externName;
        return true;
    }
}
```
2) Leave all of the accessor code essentially unchanged.
3) In LLVM object writer, ask the policy whether the accessor node represents a WASM import. Could be simple as asking whether the symbol name starts with `__wasm_import_`.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
SingleAccretion: >  We suppress IL3049 for HelloWasm and I didn't really understand what is happening with the switch in `TestNativeCallsWithMismatchedSignatures`

The basic idea of the test is that invalid PIs do not fail the compilation and do not fail the binding of valid PIs (the latter will not be true for WASM imports, but, oh well). So it's not really testing the warning directly in any special way.

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -1556,6 +1569,102 @@ private static void _CallMe(int x)
...
+
+    private static void LazyDllImportThrows()
+    {
+        StartTest("Lazy DllImport fails");
+        try
+        {
+            LazyMethod();
+            FailTest("Lazy linked DllImport did not throw");
+        }
+        catch(PlatformNotSupportedException)
```
SingleAccretion:
```suggestion
        catch (PlatformNotSupportedException)
```

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.csproj
```diff
@@ -15,13 +15,19 @@
     <CMakeProjectReference Include="CMakeLists.txt" />
   </ItemGroup>
 
+  <ItemGroup>
+    <DirectPInvoke Include="Program!CallMe" />
```
SingleAccretion: It is surprising that this is needed. Isn't it already imported via `*`?

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -1556,6 +1569,102 @@ private static void _CallMe(int x)
...
+    private static extern int CallAbiCompatFunctionWithInt(int arg);
+
+    [DllImport("StaticModule2", EntryPoint = "CommonStaticFunctionName")]
+    private static extern uint CallAbiCompatFunctionWitUint(uint arg);
+
+    private static void TestStaticAbiCompatibleSignatures()
+    {
+        StartTest("Static imports with ABI compatible signatures");
+        EndTest(CallAbiCompatFunctionWithInt(456) == 456 && CallAbiCompatFunctionWitUint(789) == 789);
+    }
```
SingleAccretion: Could you add the same kind of test to the `WasmImport` section above?

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -1556,6 +1569,102 @@ private static void _CallMe(int x)
...
+
+    [DllImport("ModuleName", EntryPoint = "DupImportTest"), WasmImportLinkage]
+    private static extern int WasmImportFuncDup2();
+
+    [System.Runtime.InteropServices.UnmanagedCallersOnly(EntryPoint = "JustForRooting")]
+    private static void RootFuncDup(int x)
+    {
+        WasmImportFuncDup1(0);
+        WasmImportFuncDup2();
+    }
```
SingleAccretion: This should technically be a `try { } catch (EntryPointNotFoundException) { }` test.

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -1556,6 +1569,102 @@ private static void _CallMe(int x)
...
+    private static void TestStaticAbiCompatibleSignatures()
+    {
+        StartTest("Static imports with ABI compatible signatures");
+        EndTest(CallAbiCompatFunctionWithInt(456) == 456 && CallAbiCompatFunctionWitUint(789) == 789);
+    }
+
+    [DllImport("StaticModule1", EntryPoint = "StaticIncompatFunctionName")]
+    private static extern int CallAbiIncompatFunctionWithInt(int arg);
+
+
```
SingleAccretion:
```suggestion
```

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -1556,6 +1569,102 @@ private static void _CallMe(int x)
...
+    [DllImport("StaticModule2", EntryPoint = "StaticIncompatFunctionName")]
+    private static extern void CallAbiIncompatFunctionWithFloat(float arg);
+
```
SingleAccretion, line 1714:
```suggestion
```

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -1556,6 +1569,102 @@ private static void _CallMe(int x)
...
+
+    // Compilation should produce a warning for CallAbiIncompatFunctionWithInt and CallAbiIncompatFunctionWithFloat.
+    [System.Runtime.InteropServices.UnmanagedCallersOnly(EntryPoint = "JustForRooting2")]
+    private static void RootFuncDup()
+    {
+        CallAbiIncompatFunctionWithInt(0);
+        CallAbiIncompatFunctionWithFloat(1);
+    }
+#endif
```
SingleAccretion, line 1722: Could you also add the scenario `3` test?

src/coreclr/tools/aot/ILCompiler.Compiler/ILCompiler.Compiler.csproj
yowl: ok, will do.

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/LLVMCodegenNodeFactory.cs
yowl: sounds good, thanks.

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/ExternMethodAccessorNode.cs
yowl: Ok, this was related to getting the warning out in LLVMObjectWriter, I'll redo that in line with the other comments

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: > * It is fine to defer the work of actually throwing the exception from this PR.

I'm missing something here.  If we don't do the generation of the PNSE here how do we do the test for 3:
```
[DllImport("not_xyz", EntryPoint="func")]
void func_1(); // Throws PNSE
```

SingleAccretion: Just `#if 0` it out for now?

src/coreclr/tools/Common/TypeSystem/Ecma/EcmaMethod.cs
yowl: Reverted

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: Done

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: ok, a few are like this that throw unmanaged exceptions in the wasm runtime.

src/coreclr/tools/aot/ILCompiler.LLVM/Compiler/DependencyAnalysis/ExternMethodAccessorNode.cs
yowl: removed now.

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: You are right, it can all be done the same as direct P/Invokes.

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.csproj
yowl: Removed.

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
yowl: Done.

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
yowl: Have implemented like this, but is `#if` out for now as we dont throw `EntryPointNotFoundException` yet, same situation as PNSE I suppose.

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
yowl: Added the missing tests.

yowl:
Think the last commit addresses the remaining review comments, thanks.

src/coreclr/tools/aot/ILCompiler/WasmConfigurablePInvokePolicy.cs
```diff
@@ -0,0 +1,45 @@
...
+        public WasmConfigurablePInvokePolicy(TargetDetails target, IReadOnlyList<string> directPInvokes, IReadOnlyList<string> directPInvokeLists)
+            : base(target, directPInvokes, directPInvokeLists)
+        {
+        }
+
+        public override bool GenerateDirectCall(MethodDesc method, out string externName)
+        {
+            if (IsWasmImport(method, out PInvokeMetadata pInvokeMetadata))
+            {
+                externName = "__wasm_import_" + pInvokeMetadata.Module + "_" + pInvokeMetadata.Name;
```
SingleAccretion, line 21: > pInvokeMetadata.Name

Does this work correctly with PIs that don't specify `EntryPoint` (if not, it also needs a test)?

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -814,10 +814,17 @@ private void GetCodeForExternMethodAccessor(ExternMethodAccessorNode node)
             LLVMValueRef externFunc = externFuncModule.AddFunction(externFuncName, externFuncType);
 
             // Add import attributes if specified.
-            if (_compilation.ConfigurableWasmImportPolicy.TryGetWasmModule(externFuncName, out string wasmModuleName))
+            // Copy the prefix from WasmConfigurablePInvokePolicy here to avoid making the policy classes public.
```
SingleAccretion: It would be better to consolidate this logic in the policy.

You can add a virtual `bool GenerateWasmImportCall(MethodDesc method, out string name, out string moduleName)` to `PInvokeILEmitterConfiguration`, and expose it through `PInvokeILProvider` as, e. g., `GetWasmImportCallInfo`

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
```diff
@@ -814,10 +814,17 @@ private void GetCodeForExternMethodAccessor(ExternMethodAccessorNode node)
...
+            if (node.Signature != null && externFuncName.StartsWith("__wasm_import_"))
             {
-                externFunc.AddFunctionAttribute("wasm-import-name", externFuncName);
-                externFunc.AddFunctionAttribute("wasm-import-module", wasmModuleName);
+                foreach (MethodDesc method in node.EnumerateMethods())
```
SingleAccretion: Perf nit: adding a `GetSingleMethod()` to the accessor node, like:
```
_methods switch { MethodDesc methods => method, _ => /* foreach loop over HashSet<MethodDesc> */ }
```
will avoid allocating the enumerator.

yowl: Thanks, I did the `switch` more like the one earlier in the file as I couldn't see how to do the foreach with the other style as I think it needs an expression.
```cs
switch (_methods)
{
    case MethodDesc oneMethod:
        return oneMethod;

    default:
        foreach (MethodDesc method in (HashSet<MethodDesc>)_methods)
        {
            return method;
        }

        return null;
}
```

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs
yowl: Thanks, yes that's cleaner.

src/coreclr/tools/aot/ILCompiler/WasmConfigurablePInvokePolicy.cs
yowl: It seems to yes, added the test:
```cs
private static void TestNamedModuleCallWithoutEntryPoint()
{
    StartTest("Wasm import from named module test");
    EndTest(ModuleFuncWithoutEntryPoint(77) == 77);
}

[DllImport("ModuleName"), WasmImportLinkage]
private static extern int ModuleFuncWithoutEntryPoint(int x);
```

Generates
```llvm
declare i32 @__wasm_import_ModuleName_ModuleFuncWithoutEntryPoint(i32) #4

define ptr @get.__wasm_import_ModuleName_ModuleFuncWithoutEntryPoint() {
GetExternFunc:
  ret ptr @__wasm_import_ModuleName_ModuleFuncWithoutEntryPoint
}

attributes #4 = { "wasm-import-module"="ModuleName" "wasm-import-name"="ModuleFuncWithoutEntryPoint" }
```

SingleAccretion (approved):
Thank you for making the changes, this LGTM!

Could you open an issue detailing what is still missing (PNSE/EPNF)?

yowl:
Thanks for the extended review!

jkotas (approved):
Looks great! Thank you for adding a lot of tests.

PR #2452, opened 2023-12-06
[NativeAOT-LLVM] Enable libraries testing

SingleAccretion:
And run our first libraries test in CI. The workflow here is the same as upstream; the change itself consists of the infrastructure updates to run the tests using non-native hosts.

This is just the "it's functional" cut; there are many pieces that are suboptimal or just don't work (e. g. the XML reporting). #2455 details the follow-up work.

Contributes to #2307.


eng/Subsets.props
```diff
@@ -413,7 +413,7 @@
 
   <!-- Mono sets -->
   <ItemGroup Condition="$(_subset.Contains('+mono.llvm+')) or $(_subset.Contains('+mono.aotcross+')) or '$(TargetOS)' == 'ios' or '$(TargetOS)' == 'iossimulator' or '$(TargetOS)' == 'tvos' or '$(TargetOS)' == 'tvossimulator' or '$(TargetOS)' == 'maccatalyst' or '$(TargetOS)' == 'android' or '$(TargetOS)' == 'browser' or '$(TargetOS)' == 'wasi' or '$(TargetsLinuxBionic)' == 'true'">
-    <ProjectToBuild Include="$(MonoProjectRoot)llvm\llvm-init.proj" Category="mono" />
+    <ProjectToBuild Include="$(MonoProjectRoot)llvm\llvm-init.proj" Category="mono" Condition="'$(RuntimeFlavor)' == 'mono'" />
```
SingleAccretion, line 416: This used to unconditionally copy GBs from the LLVM package that we don't use.

eng/liveBuilds.targets
```diff
@@ -28,6 +28,12 @@
     <CoreCLRAotSdkDir>$([MSBuild]::NormalizeDirectory('$(CoreCLRArtifactsPath)', 'aotsdk'))</CoreCLRAotSdkDir>
     <CoreCLRBuildIntegrationDir>$([MSBuild]::NormalizeDirectory('$(CoreCLRArtifactsPath)', 'build'))</CoreCLRBuildIntegrationDir>
 
+    <!-- NativeAOT-LLVM: we set up cross-compilation for WASM differently from upstream. The upstream cross-compilation scheme
+         does not scale for local development because it requires you to build the same host tools for each target configuration.
+         E. g. you need to build the same Debug compiler 3 times if you want to compile code in Debug, Checked and Release. -->
+    <IlcConfig Condition="'$(IlcConfig)' == ''">$(CoreCLRConfiguration)</IlcConfig>
+    <CoreCLRCrossILCompilerDir Condition="'$(CoreCLRCrossILCompilerDir)' != '' and '$(TargetsWasm)' == 'true'">$([MSBuild]::NormalizeDirectory('$(RepoRoot)', 'artifacts', 'bin', 'coreclr', '$(HostOS).$(BuildArchitecture).$(IlcConfig)', 'ilc'))</CoreCLRCrossILCompilerDir>
+
```
SingleAccretion, line 38: Initially, I wanted to set things up using the upstream cross-compilation scheme, but, as the comment explains, that would put us in a worse place workflow-wise.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Another milestone, thanks!

(looks like there is a conflict)

PR #2458, opened 2023-12-11
[NativeAOT-LLVM] Fix a corner-case bug with explicit tail calls

SingleAccretion:
Explicit (`.tail`) void-returning tailcalls suppress the importation of `GT_RETURN`. Compensate for this IR quirk in lowering.

`.tail` is used in the `MainMethodWrapper` to compensate for `NoOptimization`, where this bug was hit by @maraf. We haven't hit it before because all the runtime tests use error codes to communicate success/failure, hence all the main methods are `int`-returning.

Adapted the C++ EH test, which doesn't return normally from `Main`, to test this scenario.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
LGTM!

Issue #2461, opened 2023-12-15
[NativeAOT-LLVM] Publishing broken


yowl:
@jkotas, Please can you tell me what the error is again?  Last successful publish was

![image](https://github.com/dotnet/runtimelab/assets/2720041/ca3021d4-d460-4e33-9591-0c4abcd8fc28)

Thanks.

jkotas:
`Build tests` is failing in `windows-x64 Release AllConfigurations`. Log: [buildlog.txt](https://github.com/dotnet/runtimelab/files/13686243/buildlog.txt)

jkotas:
Let me know if you need anything else.

yowl:
> `Build tests` is failing in `windows-x64 Release AllConfigurations`. Log: [buildlog.txt](https://github.com/dotnet/runtimelab/files/13686243/buildlog.txt)

I see, yes looks like it only builds the `libs` subset, will see why that is.

yowl:
Can you post the error from the latest build please?

jkotas:
`windows-x64 Release AllConfigurations` - `Run tests in single file mode` is failing with:
```
2023-12-15T21:24:50.2971736Z ##[section]Starting: Run tests in single file mode
2023-12-15T21:24:50.3203885Z ==============================================================================
2023-12-15T21:24:50.3204024Z Task         : Command line
2023-12-15T21:24:50.3204095Z Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
2023-12-15T21:24:50.3204200Z Version      : 2.231.1
2023-12-15T21:24:50.3204257Z Author       : Microsoft Corporation
2023-12-15T21:24:50.3204333Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
2023-12-15T21:24:50.3204439Z ==============================================================================
2023-12-15T21:24:52.5536962Z Generating script.
2023-12-15T21:24:52.5605890Z Script contents: shell
2023-12-15T21:24:52.5611394Z D:\a\_work\1\s/src/tests/run.cmd runnativeaottests Release x64
2023-12-15T21:24:52.5840672Z ========================== Starting Command Output ===========================
2023-12-15T21:24:52.6013664Z ##[command]"C:\Windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_work\_temp\63c6cceb-158e-4b5a-8b3c-1c81b7e9b335.cmd""
[...]

PR #2462, opened 2023-12-15
[NativeAOT-LLVM] package publishing fixes

yowl:
Not really convinced this is the problem with #2461, but it does look wrong.  

Indents the parameters for `runtimelab-post-build-steps.yml`


yowl:
cc @dotnet/nativeaot-llvm

SingleAccretion (approved):

PR #2463, opened 2023-12-15
[NativeAOT-LLVM] Official windows build does not build tests

yowl:
This PR removes the `Build tests` step from the official windows build.  My logic here is

1. At https://github.com/dotnet/runtimelab/blob/c0547d1c13a46c2dee1be023e75c8eca0a0730e2/eng/pipelines/runtimelab.yml#L152 we need the libraries built

2. That happens here https://github.com/dotnet/runtimelab/blob/c0547d1c13a46c2dee1be023e75c8eca0a0730e2/eng/pipelines/runtimelab.yml#L133 where only the subset `libs` is built

3. Therefor we can't/shouldn't run the tests.

Incidentally why is `runtimelab.yml` included in the `runtime` repo?

### DIFF
--- a/eng/pipelines/runtimelab/coreclr-tests.yml
+++ b/eng/pipelines/runtimelab/coreclr-tests.yml
@@ -17,8 +17,8 @@ jobs:
       postBuildSteps:
         - template: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
           parameters:
-          testFilter: ''
-          runSingleFileTests: false
+            testFilter: ''
+            runSingleFileTests: false
 
 
 #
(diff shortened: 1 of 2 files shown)


SingleAccretion (approved):
LGTM. It is hard to debug official build, so it makes sense for it to not do too many things.

> Incidentally why is runtimelab.yml included in the runtime repo?

I presume it is so that new `runtimelab` branches don't have to write their own CI from scratch.

jkotas:
> > Incidentally why is runtimelab.yml included in the runtime repo?

> I presume it is so that new runtimelab branches don't have to write their own CI from scratch.

Yes, that's correct.

yowl:
Hopefully this will fix #2461

PR #2464, opened 2023-12-16
[NativeAOT-LLVM] Pass runSingleFileTests: false for official build

yowl:
This PR turns off runSingleFileTests for the official build.  Apologies I missed this one.


jkotas (approved):
Let's see whether it is going to fix the publishing.

PR #2465, opened 2023-12-17
[NativeAOT-LLVM] Revert addition of allconfigurations steps

yowl:
This PR removes the merge/addition of the allConfigurations (x64) steps which we don't need (as the host compiler packages are done in the `runtlimelab-post-build-steps.yml`


yowl:
cc @dotnet/nativeaot-llvm

SingleAccretion (approved):

PR #2469, opened 2023-12-20
Benchmark of interaction between async1 and async2

davidwrighton:


docs/design/features/runtime-handled-tasks.md
```diff
@@ -462,13 +462,40 @@ Major identified concerns are
...
+![MinCallCostGraph](async-mincallcost.svg)
+
+In this graph, you can see that the performance of interaction between the async1 and async2 models are not problematically slow.
+Indeed, the performance of async2 code calling async1 functions is actually greater than the performance of async1 calling async1.
+In addition, the cost to call an async2 function from another async2 function is substantially less (even if the async2 function being called needs to save the context in order to maintain full similarity of behavior with async1.)
+Here we can see the cost of inlining vs not inlining, as there is a meaningful reduction in iteration count.
+In addition, the async2 method with Context Save is also unable to inline at the current time, and also performs work simulating the additional work that the async2 logic would need to implement to fully replicate async1 behavior around async locals, and synchronization context handling.
+
+NOTES: 
+- Cost of calls from async1 method to async2 method are shown for the JIT state machine model only.
```
VSadov, line 486: It looks like the graph and table contain data for both state machine and unwinder. Am I looking at it incorrectly?

VSadov: Also in the graphs it looks like unwinder impl has roughly the same performance as state machine. Is the `"calls from async1 to async2 for the unwinder model were substantially slower"` part about something else?

davidwrighton: The graph does not include that data point. (See the note that calls from async1 to async2 were not shown for the unwinder model). Data for that is in the table.

PR #2470, opened 2023-12-21
Bump the roslyn commit to a version supporting top-level async2 functions. Add a test.

VSadov:
Bumps the Roslyn reference to a version that supports async2 in local and top-level functions.

Top level functions produce quite convoluted IL, since they are local functions with possible state captures, etc...
But runtime seems to have no problems with async2 in such context.


VSadov:
since this is just bumping Roslyn reference, I think it is ok to merge

PR #2483, opened 2024-01-11
[NativeAOT-LLVM] Remove USE_SDL option as no longer needed

yowl:
This small PR removes ths `USE_SDL` option as we have upgraded the emscripten version and updates a missed doc change for 8 -> 9

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -402,8 +402,6 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'">$(CompileWasmArgs) -s DISABLE_EXCEPTION_CATCHING=0</CompileWasmArgs>
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'wasm'">$(CompileWasmArgs) -fwasm-exceptions</CompileWasmArgs>
       <CompileWasmArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(CompileWasmArgs) -g3</CompileWasmArgs>
-      <!-- TODO-LLVM: remove -s USE_SDL=0 when we update to Emscripten with https://github.com/emscripten-core/emscripten/pull/18443 -->
-      <CompileWasmArgs>$(CompileWasmArgs) -s USE_SDL=0 $(WasmOptimizationSetting)</CompileWasmArgs>
 
       <ScriptExt Condition="'$(OS)' == 'Windows_NT'">.bat</ScriptExt>
       <WasmCompilerPath>&quot;$(EMSDK)/upstream/emscripten/emcc$(ScriptExt)&quot;</WasmCompilerPath>
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -402,6 +402,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'">$(CompileWasmArgs) -s DISABLE_EXCEPTION_CATCHING=0</CompileWasmArgs>
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'wasm'">$(CompileWasmArgs) -fwasm-exceptions</CompileWasmArgs>
       <CompileWasmArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(CompileWasmArgs) -g3</CompileWasmArgs>
+      <CompileWasmArgs>$(CompileWasmArgs) $(WasmOptimizationSetting)</CompileWasmArgs>
 
       <ScriptExt Condition="'$(OS)' == 'Windows_NT'">.bat</ScriptExt>
       <WasmCompilerPath>&quot;$(EMSDK)/upstream/emscripten/emcc$(ScriptExt)&quot;</WasmCompilerPath>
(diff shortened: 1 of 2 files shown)


yowl:
cc @dotnet/nativeaot-llvm

jkotas:
Smoke tests are failing:
```
2024-01-11T03:29:04.5377315Z   "D:\a\_work\1\s\wasm-tools\node-v20.2.0-win-x64\node.exe" --stack-trace-limit=100 System.Collections.Specialized.Tests.js -notrait category=OuterLoop -notrait category=failing 
2024-01-11T03:29:04.9428896Z   D:\a\_work\1\s\artifacts\bin\System.Collections.Specialized.Tests\Release\net9.0\publish\System.Collections.Specialized.Tests.js:128
2024-01-11T03:29:04.9430985Z         throw ex;
2024-01-11T03:29:04.9431133Z         ^
2024-01-11T03:29:04.9431282Z   
2024-01-11T03:29:04.9431466Z   RuntimeError: memory access out of bounds
2024-01-11T03:29:04.9432276Z       at System.Collections.Specialized.Tests.wasm.S_P_CoreLib_System_Environment___cctor (wasm://wasm/System.Collections.Specialized.Tests.wasm-0eff111e:wasm-function[3782]:0x249480)
2024-01-11T03:29:04.9433090Z       at System.Collections.Specialized.Tests.wasm.S_P_CoreLib_System_Runtime_CompilerServices_ClassConstructorRunner__EnsureClassConstructorRun (wasm://wasm/System.Collections.Specialized.Tests.wasm-0eff111e:wasm-function[38161]:0x1d278e1)
2024-01-11T03:29:04.9434116Z       at System.Collections.Specialized.Tests.wasm.S_P_CoreLib_System_Runtime_CompilerServices_ClassConstructorRunner__CheckStaticClassConstructionReturnNonGCStaticBase (wasm://wasm/System.Collections.Specialized.Tests.wasm-0eff111e:wasm-function[3783]:0x2494c1)
2024-01-11T03:29:04.9434925Z       at System.Collections.Specialized.Tests.wasm.__GetGCStaticBase_S_P_CoreLib_System_Environment (wasm://wasm/System.Collections.Specialized.Tests.wasm-0eff111e:wasm-function[41246]:0x1ea9ad4)
2024-01-11T03:29:04.9435678Z       at System.Collections.Specialized.Tests.wasm.S_P_CoreLib_Internal_Runtime_CompilerHelpers_StartupCodeHelpers__InitializeCommandLineArgs (wasm://wasm/System.Collections.Specialized.Tests.wasm-0eff111e:wasm-function[3776]:0x248d98)
2024-01-11T03:29:04.9436398Z       at System.Collections.Specialized.Tests.wasm.System_Collections_Specialized_Tests__Module___StartupCodeMain (wasm://wasm/System.Collections.Specialized.Tests.wasm-0eff111e:wasm-function[36558]:0x1c19611)
2024-01-11T03:29:04.9436876Z       at System.Collections.Specialized.Tests.wasm.main (wasm://wasm/System.Collections.Specialized.Tests.wasm-0eff111e:wasm-function[58]:0x2973c)
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -402,8 +402,6 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'">$(CompileWasmArgs) -s DISABLE_EXCEPTION_CATCHING=0</CompileWasmArgs>
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'wasm'">$(CompileWasmArgs) -fwasm-exceptions</CompileWasmArgs>
       <CompileWasmArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(CompileWasmArgs) -g3</CompileWasmArgs>
-      <!-- TODO-LLVM: remove -s USE_SDL=0 when we update to Emscripten with https://github.com/emscripten-core/emscripten/pull/18443 -->
-      <CompileWasmArgs>$(CompileWasmArgs) -s USE_SDL=0 $(WasmOptimizationSetting)</CompileWasmArgs>
```
SingleAccretion, line 406: This also deletes `$(WasmOptimizationSetting)`. Why would that make the tests fail in such a way, though...

yowl: Thanks, I'll put that back.  Put looking at why it crashes, this is the wat
```wat
  (func $S_P_CoreLib_System_Environment___cctor (;839;) (param $var0 i32)
    (local $var1 i32)
    (local $var2 i32)
    (local $var3 i32)
    (local $var4 i32)
    i32.const -4
    local.set $var1
    local.get $var1
    i32.load offset=909728
```

I can't find in the spec where it says the dynamic arg `-4` is not valid, but it appears so playing with 
```wat
(module
  (memory $memory 1)
  (export "memory" (memory $memory))

  (func (export "load_first_item_in_mem") (param $num i32) (result i32)
    i32.const -4

    ;; load first item in memory and return the result
    i32.load offset=4
  )

)
```
in https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Memory/Load.  An emscripten bug somewhere (binaryen/clang maybe)?  Strangely, locally the debug tests are ok, only the release fail.

SingleAccretion: It sure does look like some strange LLVM bug, but bad (or just slightly unorthodox) codegen on our part is also not out of the question. Even more strangely, in `browser-wasm windows Debug` a libraries test fails, even as for `-c Debug` this change shouldn't matter.

> I can't find in the spec where it says the dynamic arg -4 is not valid, 

I think loads/stores are meant to take the argument as unsigned, so `-4` is basically trying to access the last few bytes of all possible linear memory (even not taking into account that with the large offset, the final address computation would overflow).

I think we can keep this PR scoped to `USE_SDL` and open a follow-up issue if restoring the optimization setting restores the CI.

jkotas:
Failure is known issue.

jkotas (approved):

jkotas:
Thank you!

yowl:
ah ok, thanks, I tried windows/debug locally  and it didn't reproduce.

Issue #2454, opened 2023-12-07
[NativeAOT-LLVM] PInvoke exception throwing


yowl:
To round off #2444 we need to codegen exception throwing in a few places

1. `PlatformNotSupportedException` for lazy (not Direct PInvoke) `DllImport`'s
2. `EntryPointNotFoundException` When an attempt is made to WasmImport different functions using the same WasmImport module and function name, no import is produced.

Then uncomment the tests for these in `HelloWasm.cs`.

SingleAccretion:
A few more items:
  - After making the PIs lazy by default, remove `-sERROR_ON_UNDEFINED_SYMBOLS=0` from our `emcc` options.
  - Likewise with `--unresolved-symbols=ignore-all` with WASI.

PR #2489, opened 2024-01-16
[NativeAOT-LLVM] Implement lazy PI

SingleAccretion:
Add a WASM-specific `ResolvePInvoke` that simply throws.

Add a small document explaining the two PI linkage options and link to it from the PNSE exception message.

Contributes to #2454. It turns out that https://github.com/dotnet/runtimelab/issues/2454#issuecomment-1846093232 requires work not directly related to that issue:
1) `-sERROR_ON_UNDEFINED_SYMBOLS=0` will have to stay because Emscripten warns on (expectedly) unresolved WASM imports. It will have to be fixed upstream.
2) WASI needs some ICU fixups.

Even once the above two are fixed, we should do something like enable warnings by default, and escalate them to errors only if the relevant MSBuildism is set. We would not want people's builds to fail due to a random missing symbol somewhere (...in the BCL...).

### DIFF
--- a/src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
+++ b/src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
@@ -303,6 +303,8 @@ internal unsafe partial class Program
 
         TestNativeCallback();
 
+        LazyDllImportThrows();
+
         TestDirectPInvoke();
 
 #if false // TODO-LLVM: Should throw an EntryPointNotFoundException, but fails in the Wasm runtime (RuntimeError: null function or function signature mismatch)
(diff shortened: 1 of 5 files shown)


src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -1705,20 +1700,6 @@ private static void TestStaticPInvokeOverloadedInDifferentModules()
...
-    [DllImport("StaticModule2", EntryPoint = "StaticIncompatFunctionName")]
-    private static extern void CallAbiIncompatFunctionWithFloat(float arg);
-
-    // Compilation should produce a warning for CallAbiIncompatFunctionWithInt and CallAbiIncompatFunctionWithFloat.
-    [System.Runtime.InteropServices.UnmanagedCallersOnly(EntryPoint = "JustForRooting2")]
-    private static void RootFuncDup()
-    {
-        CallAbiIncompatFunctionWithInt(0);
-        CallAbiIncompatFunctionWithFloat(1);
-    }
```
SingleAccretion, line 1721: It turns out this part duplicates what's tested below with `TestNativeCallsWithMismatchedSignatures`, so I took it out.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

PR #2490, opened 2024-01-18
[NativeAOT-LLVM] Implement most of globalization

SingleAccretion:
How this works:
1) At runtime build time, we receive a transport package with the ICU libraries and data (`icudt.dat` and friends).
2) We package the data file into a static library with some inline assembly magic. We add a bit of code to the shim to locate the symbol in this static library, and tweak the CoreLib code to invoke that path on NativeAOT+WASM.
3) We package all of the resulting ICU static libraries into the runtime package, just like `libSystem.Native.a` and friends.
4) We link in the necessary pieces with the usual mechanism.

Overall few changes are required, but this is notably different from what Mono does, which runs some code at runtime startup that locates the ICU data (either by downloading it, or from the file system, or from the single-file bundle). It means that our `DotNetJsApi` case will still use the static data, but do that startup thing too. We can cross the bridge of reconciling these two approaches once `DotNetJsApi` gets a little more mature.

Another part not yet working is the time zone data, which uses Mono-specific bundling mechanism. It is the last undefined symbol for `HelloWasm` targeting WASI, I will tackle it separately (in #2491).

### DIFF
--- a/eng/Subsets.props
+++ b/eng/Subsets.props
@@ -63,7 +63,8 @@
     <DefaultCoreClrSubsets Condition="'$(PrimaryRuntimeFlavor)' != 'CoreCLR'">clr.iltools+clr.packages</DefaultCoreClrSubsets>
 
     <DefaultNativeAotSubsets>clr.alljits+clr.tools+clr.nativeaotlibs+clr.nativeaotruntime</DefaultNativeAotSubsets>
-    <DefaultNativeAotSubsets Condition="'$(TargetArchitecture)' == 'wasm'">clr.nativeaotlibs+clr.nativeaotruntime+nativeaot.build+mono.wasmruntime</DefaultNativeAotSubsets>
+    <DefaultNativeAotSubsets Condition="'$(TargetsWasm)' == 'true'">clr.nativeaotlibs+clr.nativeaotruntime+nativeaot.build</DefaultNativeAotSubsets>
+    <DefaultNativeAotSubsets Condition="'$(TargetsBrowser)' == 'true'">$(DefaultNativeAotSubsets)+mono.wasmruntime</DefaultNativeAotSubsets>
 
     <DefaultMonoSubsets Condition="'$(MonoEnableLLVM)' == 'true' and '$(MonoLLVMDir)' == ''">mono.llvm+</DefaultMonoSubsets>
     <DefaultMonoSubsets Condition="'$(MonoAOTEnableLLVM)' == 'true' and '$(MonoLLVMDir)' == ''">mono.llvm+</DefaultMonoSubsets>
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -506,9 +506,16 @@ The .NET Foundation licenses this file to you under the MIT license.
 
     <MakeDir Directories="$([System.IO.Path]::GetDirectoryName($(NativeBinary)))" />
 
+    <ItemGroup Condition="'$(InvariantGlobalization)' != 'true'">
+      <NativeLibrary Include="$(IlcFrameworkNativePath)libicuuc.a" />
+      <NativeLibrary Include="$(IlcFrameworkNativePath)libicui18n.a" />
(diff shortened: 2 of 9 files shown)


eng/Subsets.props
```diff
@@ -63,7 +63,8 @@
     <DefaultCoreClrSubsets Condition="'$(PrimaryRuntimeFlavor)' != 'CoreCLR'">clr.iltools+clr.packages</DefaultCoreClrSubsets>
 
     <DefaultNativeAotSubsets>clr.alljits+clr.tools+clr.nativeaotlibs+clr.nativeaotruntime</DefaultNativeAotSubsets>
-    <DefaultNativeAotSubsets Condition="'$(TargetArchitecture)' == 'wasm'">clr.nativeaotlibs+clr.nativeaotruntime+nativeaot.build+mono.wasmruntime</DefaultNativeAotSubsets>
+    <DefaultNativeAotSubsets Condition="'$(TargetsWasm)' == 'true'">clr.nativeaotlibs+clr.nativeaotruntime+nativeaot.build</DefaultNativeAotSubsets>
+    <DefaultNativeAotSubsets Condition="'$(TargetsBrowser)' == 'true'">$(DefaultNativeAotSubsets)+mono.wasmruntime</DefaultNativeAotSubsets>
```
SingleAccretion, line 67: This is actually necessary for build correctness - `browser.proj` copies Browser-targeting ICU files to the `libs.native` output directory (duplicatively with `build-native.proj`), and they are not suitable for WASI linking.

src/native/libs/build-native.proj
```diff
@@ -30,8 +30,21 @@
...
+    <ItemGroup>
+      <_IcuArtifacts Condition="'$(_IcuDir)' != '' and '$(TargetsAppleMobile)' != 'true'"
+                     Include="$(_IcuDir)/lib/libicuuc.a;
+                              $(_IcuDir)/lib/libicui18n.a;
+                              $(_IcuDir)/lib/libicudata.a;
+                              $(_IcuDir)/lib/*.dat" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(_IcuArtifacts)" DestinationFolder="$([MSBuild]::NormalizeDirectory('$(ArtifactsBinDir)', 'native', '$(_BuildNativeOutConfig)'))" SkipUnchangedFiles="true" />
+  </Target>
```
SingleAccretion, line 44: I hope to upstream changes to this file, and perhaps remove the duplicate copying done by `wasm.proj`/`wasi.proj`.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -555,8 +562,8 @@ The .NET Foundation licenses this file to you under the MIT license.
     <ItemGroup Condition ="'$(_targetOS)' == 'wasi'" >
       <CustomLinkerArg Include="--sysroot=&quot;$(WASI_SDK_PATH.Replace(&quot;\&quot;, &quot;/&quot;))/share/wasi-sysroot&quot;" />
       <CustomLinkerArg Include="@(CustomLinkerArgExport->'-Wl,--export=%(Identity)')" />
-      <!-- Wasi has lots of undefined symbols currently, mostly globalization -->
-      <CustomLinkerArg Include="-Wl,--unresolved-symbols=ignore-all -Wl,--error-limit=0" />
+      <!-- We want to put 'warn-unresolved-symbols' here, but wasm-ld crashs on that. Oops. -->
+      <CustomLinkerArg Include="-Wl,--unresolved-symbols=ignore-all" />
```
SingleAccretion: Will have to debug and file a bug on LLVM about this (if it hasn't been fixed already)...

Edit: there is one already - https://github.com/llvm/llvm-project/issues/53987.

SingleAccretion:
@dotnet/nativeaot-llvm

jkotas:
FYI: Likely conflict with https://github.com/dotnet/runtime/pull/97216

eng/Subsets.props
yowl: Should line 66 be TargetsWasi?

SingleAccretion: I don't think it should, the subsets it sets are common to both WASI and Browser.

yowl (approved):
Nice, thanks

PR #2503, opened 2024-02-02
[NativeAOT-LLVM] Fix the `Math.Round` intrinsic

SingleAccretion:
We want `@llvm.roundeven`, not `@llvm.round`.

Found by the time span division test: https://github.com/dotnet/runtimelab/issues/2307#issuecomment-1922035041.


yowl (approved):
Thanks

PR #2504, opened 2024-02-02
[NativeAOT-LLVM] Fix the missing time zone data index

SingleAccretion:
The "zone.tab" file is used to index system time zones and was missing.

Another fix for a bug from https://github.com/dotnet/runtimelab/issues/2307#issuecomment-1922035041.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks!

Issue #2294, opened 2023-05-30
[NativeAOT-LLVM] Efficient shadow stack allocation


SingleAccretion:
#2510 added an initial precise LSSA implementation. There are still many obvious areas for improvement:

1\) Variables that otherwise have disjoint lifetimes need to be shuffled between stack locations. Consider:
```
object a = { ... };
// Using 'a'
object b = a; // Last use of 'a'.
```
Allocation `a` and `b` to the same slot would allow make the `b = a` assignment effectively zero-cost.

2\) "Related" variables do not need to be double-spilled, at least in the conservative model we have. I. e. in the following example:
```
object a = { ... };
SafePoint(); ; 'a' needs to be spilled to not be collected

ref T b = ref a.Field;
SafePoint(); ; 'b' does not need to be spilled since it's already rooted by 'a'
```
This idea can be extended to many situations, e. g.:
1) Array access, span access.
2) Pre-pinned locations, e. g. implicit byrefs.
3) PHIs of all of the above.

3\) Structs of which fields are accessed are treated very conservatively right now because they don't participate in SSA.
4\) Variables spilled to keep SDSUs alive are always reloaded (also due to them not being in SSA).

PR #2509, opened 2024-02-06
[NativeAOT-LLVM] Zero-diff refactoring of LSSA

SingleAccretion:
This paves the way for changes that will introduce more precise analysis.

No diffs.
Contributes to #2294.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
I really just read this and haven't followed the logic 100%.  As often the case, as it's passing and subsequent PRs are using it, then it's likely all good!

PR #2512, opened 2024-02-08
Add initial documentation for the projection tooling

kotlarmilos:
## Description

This PR adds initial documentation for the tooling focusing on static methods and P/Invoke declarations.


docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+| ------------------------------- | -------- |
+| `Swift.Int64`                   | `long`   |
+| `Swift.UInt64`                  | `ulong`  |
+| `Swift.Int32`                   | `int`    |
+| `Swift.UInt32`                  | `uint`   |
+| `Swift.Int16`                   | `short`  |
+| `Swift.UInt16`                  | `ushort` |
+| `Swift.Int8`                    | `sbyte`  |
+| `Swift.UInt8`                   | `byte`   |
+| `Swift.UnsafeRawPointer`        | `IntPtr` |
```
AaronRobinsonMSFT: The `IntPtr` / `UIntPtr` are aliases to `nint` / `nuint`. We should be using actual pointers like `void*` in C# when expressing the pointer concept.

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+| `Swift.UnsafeMutableRawPointer` | `IntPtr` |
+| `Int`                           | `nint`   |
+| `UInt`                          | `nuint`  |
+| `Bool`                          | `bool`   |
+| `Float`                         | `float`  |
+| `Double`                        | `double` |
+
+Swift primitive types are implemented as frozen structs that conform to Swift-specific lowering processes handled by the runtime. However, such mappingcan fit within the underlying calling convention as these types are below the size limit for being passed by reference.
```
AaronRobinsonMSFT:
```suggestion
Swift primitive types are implemented as frozen structs that conform to Swift-specific lowering processes handled by the runtime. However, such mapping can fit within the underlying calling convention as these types are below the size limit for being passed by reference. 
```

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+| `UInt`                          | `nuint`  |
+| `Bool`                          | `bool`   |
+| `Float`                         | `float`  |
+| `Double`                        | `double` |
+
+Swift primitive types are implemented as frozen structs that conform to Swift-specific lowering processes handled by the runtime. However, such mappingcan fit within the underlying calling convention as these types are below the size limit for being passed by reference. 
+
+### Structs
+
+Swift structs are projected into C# as `IDisposable` classes, implementing `ISwiftStruct`, which inherit from `ISwiftNominalType` and `ISwiftDisposable`. This approach is chosen due to the semantic differences in lifecycle management between Swift and C# types. Bindings are generated with `SwiftStructAttribute` that inherits `SwiftNominalTypeAttribute`. This attribute contains information about the Swift type.
```
AaronRobinsonMSFT: All these interfaces and public APIs must go through API Review. We are fine to work out the details in this repo, but prior to being official we will need sign-off. Just making sure that is clear for now.

docs/README.md
```diff
@@ -1,3 +1,201 @@
-# Documentation
+# .NET Swift interop tooling documentation
 
-This document outlines the functional design of the tooling and projections between Swift and .NET.
\ No newline at end of file
+This document provides a detailed overview of the .NET Swift interop tooling, focusing on the projections between Swift and .NET, and the functional design of the tooling.
```
AaronRobinsonMSFT: Might be worth explicit stating these tools will work with C# only and any other .NET language is a non-goal.

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+            public nint X
+            {
+                get { }
+                set { }
+            }
+        }
+    }
+```
+
+There is a payload `SwiftData` along with two constructors. The first maps onto the `init` method inside the swift class. The second is an internal constructor that gets used to define an uninitialized type. This constructor gets used by the marshaler when a type needs to be allocated before it gets used, for example, as a return value because Swift semantics don’t allow to explicitly have variables in an uninitialized state.
```
AaronRobinsonMSFT: > This constructor gets used by the marshaler when a type needs to be allocated before it gets used, for example, as a return value because Swift semantics don’t allow to explicitly have variables in an uninitialized state.

We aren't going to be able to pass `class` types down to Swift so I don't see why this is needed, perhaps it makes the source generation easier? For example, I would consider the two following scenarios - either the type has an official default constructor or doesn't and we create an `internal` default constructor.

What does the constructor with a `SwiftNominalCtorArgument` get us?

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+    {
+        [SwiftStruct("libNewClassCompilerTests.dylib",
+            "$s21NewClassCompilerTests6BarIntVMn", 
+            "$s21NewClassCompilerTests6BarIntVN", "")]
+        public class BarInt : ISwiftStruct
+        {
+            public BarInt(nint x)
+            {
+            }
+            internal BarInt(SwiftNominalCtorArgument unused)
```
jkotas: The internal constructor won't work well if the bindings are spanning multiple assemblies. I assume that we will eventually want to have one assembly per component for pre-generated bindings.

docs/README.md
jkotas: I assume that the attributes that support projections are going to be part of SwiftRuntimeLibrary, similar to how attributes that support Objective-C or WinRT projections are in their own respective support libraries/repos.

Strictly speaking, the projection support libraries are not under jurisdictions of dotnet/runtime API reviews, but it does not hurt to seek feedback for the API shape.

docs/README.md
jkoritzinsky: I agree with Aaron here. We should map `UnsafeRawPointer` to `void*`.

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+    using System;
+    using System.Runtime.InteropServices;
+    using SwiftRuntimeLibrary;
+    using SwiftRuntimeLibrary.SwiftMarshal;
+    
+    namespace NewClassCompilerTests
+    {
+        [SwiftStruct("libNewClassCompilerTests.dylib",
+            "$s21NewClassCompilerTests6BarIntVMn", 
+            "$s21NewClassCompilerTests6BarIntVN", "")]
```
jkoritzinsky: Can you add an explanation of what these parameters mean to this doc?

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+| `Swift.UInt32`                  | `uint`   |
+| `Swift.Int16`                   | `short`  |
+| `Swift.UInt16`                  | `ushort` |
+| `Swift.Int8`                    | `sbyte`  |
+| `Swift.UInt8`                   | `byte`   |
+| `Swift.UnsafeRawPointer`        | `IntPtr` |
+| `Swift.UnsafeMutableRawPointer` | `IntPtr` |
+| `Int`                           | `nint`   |
+| `UInt`                          | `nuint`  |
+| `Bool`                          | `bool`   |
```
rolfbjarne, line 51: While I agree that Swift's `Bool` should be mapped to C#'s `bool`, just a heads up that C# `bool` isn't a blittable type (unlike all the other types in this list), so in P/Invokes and reverse callbacks we should probably use `byte` as an intermediate type.

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+
+The projection tooling will generate the following C#, with function bodies left empty for simplicity:
+```csharp
+    using System;
+    using System.Runtime.InteropServices;
+    using SwiftRuntimeLibrary;
+    using SwiftRuntimeLibrary.SwiftMarshal;
+    
+    namespace NewClassCompilerTests
+    {
```
rolfbjarne: IMHO the namespace isn't relevant to this documentation, so it can be removed to keep the sample code simpler.

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
```
rolfbjarne: Same wrt namespace.

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+            }
+            ~FooECTIA()
+            {
+            }
+            public static FooECTIA NewA(nint value0)
+            {
+            }
+            public static FooECTIA NewB(double value0)
+            {
+            }
```
rolfbjarne: The [naming conventions](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods) says to use verbs for methods:

```suggestion
            public static FooECTIA CreateA(nint value0)
            {
            }
            public static FooECTIA CreateB(double value0)
            {
            }
```

rolfbjarne: Although this might be a potential alternative design?

```csharp
public class FooECTIA : ISwiftEnum {
    public FooECTIA (FooECTIACases case, nint value) {}
    public FooECTIA (FooECTIACases case, double value) {}
}
```

This could be simplified to something like this if the value type defines the case as well:

```csharp
public class FooECTIA : ISwiftEnum {
    public FooECTIA (nint value) {}
    public FooECTIA (double value) {}
}
```

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+- Closures
+- Passing a struct by value in more registers than P/Invoke will allow
+- Exporting .NET into Swift
+
+## Functional outline
+
+The tooling consists of the following components:
+- **CLIInterface**: A command-line interface that orchestrates the tooling workflow.
+- **SwiftReflector**: A component that aggregates public API definitions from Swift modules.
+- **Dynamo**: A component that provides API for C# code generation.
```
rolfbjarne: Any particular reason Roslyn wouldn't work?

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+            public void Dispose()
+            {
+            }
+            void Dispose(bool disposing)
+            {
+            }
+            ~BarInt()
+            {
+            }
+            public byte[] SwiftData
```
rolfbjarne: What happens if code modifies the array directly?

```csharp
var value = new BarInt ();
value.SwiftData [1] = 42;
```

I'm assuming nothing good... so maybe we should figure out a way to expose this so that users can't shoot themselves in the foot (as easily at least)?

One simple solution might be to mark it as EditorBrowseable.Never.

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+            "$s21NewClassCompilerTests6BarIntVN", "")]
+        public class BarInt : ISwiftStruct
+        {
+            public BarInt(nint x)
+            {
+            }
+            internal BarInt(SwiftNominalCtorArgument unused)
+            {
+            }
+            public static SwiftMetatype GetSwiftMetatype()
```
vitek-karas: Do we have another doc describing the bases of the type system mapping? All these `ISwiftNominalType` and so on, along with for example this method and what are the expectations.

For this specific method, I'm curious what it's for. And if it's part of some contract, maybe it should be a static interface method?

docs/README.md
vitek-karas: Is there a description somewhere how exactly this works and what is it for?

docs/README.md
vitek-karas: I agree with Rolf that we should use Roslyn. But I think it's OK to stick with the existing code for now and change this as a subsequent PR - but Roslyn should be the plan (unless there's some technical reason not to)

docs/README.md
```diff
@@ -1,3 +1,201 @@
...
+
+The `SwiftRuntimeLibrary` This component is a library that provides basic support for Swift interop that the generated code builds on. It includes bindings for common built-in types like arrays and dictionaries. It provides a single source of truth for Swift types so that they can be exposed across an assembly boundary. 
+
+### Workflow
+
+The general workflow for generating C# bindings from Swift code is as follows:
+
+1. Process the Swift library interface (`.swiftinterface`) with the SwiftReflector and generate `TLDefinition` with mappings between entry points and mangled names.
+2. Aggregate a public API based on `.swiftinterface` with the SwiftReflector.
+3. Generate surce code for C# bindings with Dynamo and compile the code into a managed library.
```
vitek-karas:
```suggestion
3. Generate source code for C# bindings with Dynamo and compile the code into a managed library.
```

docs/README.md
kotlarmilos: Ok, thanks. Added to the implementation plan in https://github.com/dotnet/runtimelab/pull/2512.

docs/README.md
jkotas: FWIW, Roslyn is not a particularly great API for generating source code. Most source generators generate the source by printing it one line at time, without much help from Roslyn. Random example: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Emitter.cs#L461-L485

docs/README.md
kotlarmilos: Correct. A lightweight wrapper will be needed to convert bool to byte before the P/Invoke. I'll add this to the task list as a low-priority feature. For the initial version, we might choose to use the byte mapping.

docs/README.md
kotlarmilos: Added.

docs/README.md
kotlarmilos: Simplified.

docs/README.md
kotlarmilos: Good point, added a section to illustrate the projections. We will seek feedback when migrating the `SwiftRuntimeLibrary`.

docs/README.md
AaronRobinsonMSFT: Another facet of this that has been mentioned is that source generators aren't the best solution when C# isn't the "source of truth". When the C# is the "source of truth", the source generators are appropriate - see `LibraryImport` or the COM source generator.

docs/README.md
kotlarmilos: > The [naming conventions](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods) says to use verbs for methods:

Added to the implementation plan. 



> ```cs
> public class FooECTIA : ISwiftEnum {
>     public FooECTIA (FooECTIACases case, nint value) {}
>     public FooECTIA (FooECTIACases case, double value) {}
> }
> ```

Your suggestion is sensible and appears to simplify usage but It is hard to make a decision at this stage. The existing factory-like approach is easier for maintenance and further extensibility. I will add it to the list of tasks and we can consider implementing it at some later stage.

docs/README.md
kotlarmilos: > We aren't going to be able to pass class types down to Swift so I don't see why this is needed, perhaps it makes the source generation easier?

How are we supposed to interop with functions that accept `class` types?

> What does the constructor with a `SwiftNominalCtorArgument` get us?

According to the docs, this constructor should be used by marshaling; it allocates the `SwiftData` member to the correct size but does not initialize it.

docs/README.md
kotlarmilos: What do you consider a component? Should we split them by Swift namespaces? 

@stephen-hawley Is there any particular reason the constructor has been set to internal?

docs/README.md
```diff
@@ -1,3 +1,210 @@
...
+
+1. Process the Swift library interface (`.swiftinterface`) with the SwiftReflector and generate `TLDefinition` with mappings between entry points and mangled names.
+2. Aggregate a public API based on `.swiftinterface` with the SwiftReflector.
+3. Generate source code for C# bindings with Dynamo and compile the code into a managed library.
+
+### Shipping
+
+This subsection will be expanded during the design review process.
+
+### User expirience
```
rolfbjarne:
```suggestion
### User experience
```

docs/README.md
kotlarmilos: Here is the current definition:
```csharp
	[AttributeUsage (AttributeTargets.Class, AllowMultiple = false)]
	public sealed class SwiftStructAttribute : SwiftValueTypeAttribute {

		public SwiftStructAttribute (string libraryName, string nominalTypeDescriptor, string metadata, string witnessTable)
			: base (libraryName, nominalTypeDescriptor, metadata, witnessTable)
		{
		}
	}
```

It seems to be intended for future extensibility. If that's the case, I would removing it to simplify and narrow the scope.

docs/README.md
```diff
@@ -1,3 +1,210 @@
...
+
+The tooling provides an implementation of runtime Swift types in C#. The code snippet below illustrates the general mechanism by which the types are projected.
+
+```csharp
+namespace SwiftRuntimeLibrary {
+	public interface ISwiftValueType : IDisposable {
+		byte [] SwiftData { get; set; }
+	}
+
+    public interface ISwiftObject : IDisposable {
```
rolfbjarne: Fix indentation (tab instead of spaces, since the rest of the code has tabs):

```suggestion
	public interface ISwiftObject : IDisposable {
```

docs/README.md
vitek-karas: It would make sense to have a way to identify the Swift type - specifically for the Projection B relies on Projection A scenario - B will need to be able to find which C# type is a specific Swift type from A.

kotlarmilos: Is the implemented interface sufficient for identifying the Swift type?

docs/README.md
kotlarmilos: Added to the list in https://github.com/dotnet/runtime/issues/95633.

docs/README.md
kotlarmilos: Here is the existing docs: https://github.com/xamarin/binding-tools-for-swift/blob/main/docs/Runtime-Public-API.md

I suggest to reduce the scope of the public API needed at this stage and migrate only the subset to this repository. Also, creating a contract (ABI) interface is a good idea, adding it to the list of tasks.

docs/README.md
kotlarmilos: According to the docs, it is an interface that represents Swift named value types.

> - `byte[] SwiftData { get; set }` - gets an array of opaque data that represents the swift value type. The size of the array can be found by calling `StructMarshal.Marshal.Strideof()`.

@stephen-hawley Could you please provide additional context on this?

kotlarmilos: Does it need to be a public member?

docs/README.md
AaronRobinsonMSFT: > How are we supposed to interop with functions that accept class types?

They will be reduced to a raw pointer at some point. There is no way in CoreCLR to marshal a `class` instance across an interop boundary without some kind of projection. I agree that the user facing API surface will accept a `class` instance, but the generated code will convert that into a blittable form - likely a raw pointer.

> According to the docs, this constructor should be used by marshaling; it allocates the SwiftData member to the correct size but does not initialize it.

If we need to allocate a managed object projection in an uninitialized state the marshaller should use something akin to [`RuntimeHelpers.GetUninitializedObject()`](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.runtimehelpers.getuninitializedobject).

I could be missing nuance here on how this all comes together so if this doesn't align with the design let's discuss it with some concrete examples we can walk though.

docs/README.md
jkoritzinsky: I would consider each `.framework` a separate component.

docs/README.md
jkoritzinsky: If it needs to be public, could we use `ReadOnlySpan` or `ReadOnlyMemory`?

kotlarmilos:
This PR has been updated to align with the implementation in https://github.com/dotnet/runtimelab/pull/2517. The scope has been narrowed to only include static methods and P/Invoke calls.

rolfbjarne (approved):

PR #2521, opened 2024-02-26
[NativeAOT-LLVM] Implement timers on Browser

### LINKED ISSUE #2505: [NativeAOT-LLVM] Implement timers on Browser
https://github.com/dotnet/runtimelab/pull/2502 takes the PNSE version for `TimerQueue` with the expectation that it will be replaced by its functional `Browser` cousin. It is a matter of deciding where/how to put the little bit of JS that sets the 'native' timer.

Mono reference: [ `sheduling.ts`](https://github.com/dotnet/runtime/blob/0d8c1e298bb6a896e9b9796832073b8fa9f73ecd/src/mono/browser/runtime/scheduling.ts#L73-L93). Point of note: prevent callbacks from executing if the runtime has already exited (via `abort` or otherwise).

SingleAccretion:
We can take a slightly different approach to Mono to avoid JS.

Contributes to #2505.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

PR #2525, opened 2024-03-05
Swift bindings generator with simple parser and emitter

kotlarmilos:
## Description

This PR introduces a toolset for generating C# bindings from Swift interface files. This is the MVP aiming to support the generation of C# bindings for [CryptoKit dev templates](https://github.com/dotnet/runtime/issues/95636).

The tooling consists of the following components:
- **SwiftBindings**: A command-line interface with Swift ABI parser, marshalling types, and string-based C# emitter.
- **SwiftRuntimeLibrary**: A library providing projections of common Swift types.

The parser and emitter are modular and implement `ISwiftParser` and `ICSharpEmitter` interfaces. The Swift ABI parser can be replaced with a swiftinterface parser, while the string-based emitter can be replaced with an object model-based emitter.

**Workflow:**
1. Parse Swift ABI files to aggregate public ABI.
2. Generate marshalling information if needed.
3. Emit C# source code using a string-based emitter.

The unit tests verify the XML layout of the type database and various P/Invoke cases. Additionally, CryptoKit dev templates with unsafe (buffer) pointers are implemented.


src/SwiftRuntimeLibrary/src/ISwiftObject.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace SwiftRuntimeLibrary
```
rolfbjarne: What about this instead:

```suggestion
namespace Swift.Runtime
```

src/SwiftRuntimeLibrary/src/SwiftRuntimeLibrary.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Library</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>disable</Nullable>
```
rolfbjarne: Why not nullable?

src/SwiftRuntimeLibrary/tests/SwiftRuntimeLibrary.Tests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>disable</Nullable>
```
rolfbjarne: Same

src/SwiftRuntimeLibrary/src/SwiftCore.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xamtypedatabase version="1.0">
```
rolfbjarne:
```suggestion
<swifttypedatabase version="1.0">
```

src/SwiftRuntimeLibrary/src/SwiftCore.xml
```diff
@@ -0,0 +1,50 @@
...
+    <entities>
+        <entity sharpNameSpace="System" sharpTypeName="byte" entityType="Scalar" blitable="true" size="1" stride="1">
```
rolfbjarne: Technically we're creating managed bindings, not C# bindings, so use `managed` instead of `sharp` in the attribute names:

```suggestion
        <entity managedNamespace="System" managedTypeName="byte" entityType="Scalar" blitable="true" size="1" stride="1">
```

src/SwiftRuntimeLibrary/src/SwiftCore.xml
```diff
@@ -0,0 +1,50 @@
...
+            <typedeclaration kind="struct" name="Swift.UInt8" module="Swift" />
+        </entity>
+        <entity sharpNameSpace="System" sharpTypeName="sbyte" entityType="Scalar" blitable="true" size="1" stride="1">
```
rolfbjarne: `blittable` instead of `blitable` matches how we write it everywhere else:

```suggestion
        <entity sharpNameSpace="System" sharpTypeName="sbyte" entityType="Scalar" blittable="true" size="1" stride="1">
```

src/SwiftRuntimeLibrary/src/SwiftCore.xml
```diff
@@ -0,0 +1,50 @@
...
+            <typedeclaration kind="struct" name="Swift.Int8" module="Swift" />
+        </entity>
+        <entity sharpNameSpace="System" sharpTypeName="short" entityType="Scalar" blitable="true" size="2" stride="2">
```
rolfbjarne: Is `sharpTypeName` the C# keyword for the type? If so, IMHO it would make more sense to use the BCL type name instead (otherwise there's a weird mix of C#-isms here):

```suggestion
        <entity sharpNameSpace="System" sharpTypeName="Int16" entityType="Scalar" blitable="true" size="2" stride="2">
```

Then somewhere else there can be a mapping between BCL types and the corresponding built-in C# type if so desired.

src/SwiftRuntimeLibrary/src/ISwiftObject.cs
kotlarmilos: Sounds good. Is it compatible with ObjC namespace naming?

src/SwiftRuntimeLibrary/src/SwiftCore.xml
kotlarmilos: Thanks, updated.

src/SwiftRuntimeLibrary/src/ISwiftObject.cs
rolfbjarne: Yes, I don't think that will clash with anything.

src/SwiftRuntimeLibrary/src/ISwiftObject.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Swift.Runtime
+{
+    public unsafe interface ISwiftObject : IDisposable
```
AaronRobinsonMSFT: I suggest getting into the [habit of documenting public APIs](https://learn.microsoft.com/dotnet/csharp/language-reference/xmldoc/) during development. It helps with reviews and eventual public documentation efforts.

src/SwiftRuntimeLibrary/src/SwiftCore.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<swifttypedatabase version="1.0">
+    <entities>
+        <entity managedNameSpace="System" managedTypeName="Byte" entityType="Scalar" blittable="true" size="1" stride="1">
```
AaronRobinsonMSFT: Is this schema documented somewhere?

src/SwiftRuntimeLibrary/src/ISwiftObject.cs
rolfbjarne: If you set `GenerateDocumentationFile=true` in the csproj, you'll get warnings if you have undocumented APIs. Then setting `TreatWarningsAsErrors=true` and you'll never forget to document anyhing.

kotlarmilos: Updated, thanks.

src/SwiftRuntimeLibrary/src/SwiftCore.xml
kotlarmilos: I don't believe the schema is standardized. I've updated the docs to document the schema.

docs/README.md
```diff
@@ -52,9 +52,40 @@ The table below lists the Swift types and their corresponding C# types.
...
+##### `entities`
+- **Description:** Container for individual data type entities.
+- **Child Elements:**
+    - `entity`: Represents a specific data type in Swift.
+        - **Attributes:**
+        - `managedNameSpace`: Specifies the managed namespace of the data type.
+        - `managedTypeName`: Specifies the managed type name of the data type.
+        - `entityType`: Specifies the type of entity.
+        - `blittable`: Specifies whether the data type is blittable (`true` or `false`).
+        - `size`: Specifies the size of the data type in bytes.
```
jkotas: Does this mean that this database file is architecture specific? Or are sizes of all types that can show up here assumed to be architecture neutral?

src/SwiftRuntimeLibrary/src/SwiftCore.xml
```diff
@@ -0,0 +1,50 @@
...
+        <entity managedNameSpace="System" managedTypeName="UIntPtr" entityType="Scalar" blittable="true" size="8" stride="8">
+            <typedeclaration kind="struct" name="Swift.UInt" module="Swift" />
+        </entity>
+        <entity managedNameSpace="System" managedTypeName="bool" entityType="Scalar" blittable="false" size="4" stride="4">
+            <typedeclaration kind="struct" name="Swift.Bool" module="Swift" />
+        </entity>
+        <entity managedNameSpace="SwiftRuntimeLibrary" managedTypeName="UnsafeRawPointer" entityType="Scalar" blittable="false" size="0" stride="0">
+            <typedeclaration kind="struct" name="Swift.UnsafeRawPointer" module="Swift" />
+        </entity>
+        <entity managedNameSpace="SwiftRuntimeLibrary" managedTypeName="UnsafeMutableRawPointer" entityType="Scalar" blittable="false" size="0" stride="0">
```
AaronRobinsonMSFT: Can Swift have zero sized data types? If so, are they exposed in the stable ABI? I don't recall standard C or C++ permitting zero sized data types, but could be mistaken. Either way, are these really size 0?

docs/README.md
kotlarmilos: I wouldn't go with architecture-specific database if possible. At this stage, we don't use these fields and they can be excluded.

src/SwiftRuntimeLibrary/src/SwiftCore.xml
kotlarmilos: Look at the previous comment, I suggest excluding it until it becomes necessary.

src/SwiftRuntimeLibrary/src/SwiftCore.xml
```diff
@@ -0,0 +1,50 @@
...
+        <entity managedNameSpace="System" managedTypeName="Double" entityType="Scalar" blittable="true" size="8" stride="8">
+            <typedeclaration kind="struct" name="Swift.Double" module="Swift" />
+        </entity>
+        <entity managedNameSpace="System" managedTypeName="IntPtr" entityType="Scalar" blittable="true" size="8" stride="8">
+            <typedeclaration kind="struct" name="Swift.Int" module="Swift" />
+        </entity>
+        <entity managedNameSpace="System" managedTypeName="UIntPtr" entityType="Scalar" blittable="true" size="8" stride="8">
+            <typedeclaration kind="struct" name="Swift.UInt" module="Swift" />
+        </entity>
+        <entity managedNameSpace="System" managedTypeName="bool" entityType="Scalar" blittable="false" size="4" stride="4">
```
AaronRobinsonMSFT: Is this for .NET, Swift, or the marshaled type? Both .NET and Swift have `bool` as 1 byte - `MemoryLayout<Bool>.size`. The built in marshaller defaults to the `BOOL` size, which is 4. However, if we are documenting that here then I don't understand this file's purpose. Are these for .NET, Swift or interop types? What are we doing about [`DisableDefaultMarshalling`](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.disableruntimemarshallingattribute), since in that case the marshalled `bool` would indeed be 1.

src/SwiftRuntimeLibrary/src/SwiftCore.xml
stephen-hawley: Just as an FYI, Swift supports 0 length value types.

src/SwiftRuntimeLibrary/src/SwiftCore.xml
stephen-hawley: This file is meant for mapping swift types to .NET types. In a quick look-over of the code, size, stride and blittable are artifacts and can go away entirely.
In the dim dark ages of BTfS, I tried to create two classes of struct: blittable (and fully representable in C# with no wrappers) and non-blittable. This was an endless source of bugs and I removed it.

docs/README.md
kotlarmilos: The `entityType`, `blittable`, `size`, and `stride` attributes have been removed for simplicity.

kotlarmilos:
This PR has been updated with a bindings generator that includes a simple parser and emitter (refer to the updated description and docs). Please note that this is an MVP for CryptoKit dev templates.

docs/README.md
```diff
@@ -105,38 +132,29 @@ In the example, the user's code references the `HelloLibraryBindings` namespace
...
 
-namespace SwiftReflector.Parser
-{
-    public interface ISwiftParser
-    {
-        public ModuleDeclaration GetModuleDeclaration(string filePath, ErrorHandling errors);
-    }
-}
-```
+The aggregation of the public ABI is done through the `ISwiftParser` interface. This interface defines the layout for concrete implementations responsible for parsing and collecting ABI information. Two implementations exist: Swift ABI parser and Swift interface parser. The Swift ABI parser aggregates ABI information based on an ABI json file. The Swift interface parser is designed to handle `.swiftinterface` files. The `.swiftinterface` file doesn't contain mangled names, and the parser should consumes the dynamic library (`.dylib`) to generate declarations. Currently, the tooling only implements the Swift ABI parser.
```
rolfbjarne:
```suggestion
The aggregation of the public ABI is done through the `ISwiftParser` interface. This interface defines the layout for concrete implementations responsible for parsing and collecting ABI information. Two implementations exist: Swift ABI parser and Swift interface parser. The Swift ABI parser aggregates ABI information based on an ABI json file. The Swift interface parser is designed to handle `.swiftinterface` files. The `.swiftinterface` file doesn't contain mangled names, and the parser should consume the dynamic library (`.dylib`) to generate declarations. Currently, the tooling only implements the Swift ABI parser.
```

src/SwiftRuntimeLibrary/src/TypeDatabase.cs
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Xml;
+
+namespace SwiftRuntimeLibrary
```
rolfbjarne:
```suggestion
namespace Swift.Runtime
```

src/SwiftRuntimeLibrary/src/TypeDatabase.cs
```diff
@@ -0,0 +1,116 @@
...
+                        throw new Exception("Invalid XML structure: Missing attributes.");
+
+                    _swiftToCSharpMapping.Add(swiftName, csharpName);
+                }
+            }
+            else
+            {
+                throw new Exception("Invalid XML structure: 'entities' node not found.");
+            }
+        }
```
rolfbjarne: This can be turned around to reduce indentation:

```suggestion
            if (entitiesNode == null)
            	throw new Exception("Invalid XML structure: 'entities' node not found.");

            foreach (XmlNode? entityNode in entitiesNode.ChildNodes)
            {
                XmlNode? typeDeclarationNode = entityNode?.SelectSingleNode("typedeclaration");
                if (typeDeclarationNode == null)
                    throw new Exception("Invalid XML structure: 'typedeclaration' node not found.");
                    
                string? swiftName = typeDeclarationNode?.Attributes?["name"]?.Value;
                string? csharpName = entityNode?.Attributes?["managedTypeName"]?.Value;
                if (swiftName == null || csharpName == null)
                    throw new Exception("Invalid XML structure: Missing attributes.");

                _swiftToCSharpMapping.Add(swiftName, csharpName);
            }
        }
```

src/SwiftRuntimeLibrary/src/TypeDatabase.cs
```diff
@@ -0,0 +1,116 @@
...
+            if (!ValidateXmlSchema(xmlDoc))
+                throw new Exception("Invalid XML schema.");
+
+            switch (xmlDoc.DocumentElement?.Attributes?["version"]?.Value)
+            {
+                case "1.0":
+                    ReadVersion1_0(xmlDoc);
+                    break;
+                default:
+                    throw new Exception("Unsupported XML version.");
```
rolfbjarne:
```suggestion
                    throw new Exception("Unsupported database version {0} in {1}.", version, file);
```

src/SwiftRuntimeLibrary/src/TypeDatabase.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// </summary>
+        /// <param name="file">The path to the XML file containing the type mappings.</param>
+        public TypeDatabase(string file)
+        {
+            XmlDocument xmlDoc = new XmlDocument();
+            xmlDoc.Load(file);
+            if (!ValidateXmlSchema(xmlDoc))
+                throw new Exception("Invalid XML schema.");
+
+            switch (xmlDoc.DocumentElement?.Attributes?["version"]?.Value)
```
rolfbjarne:
```suggestion
            var version = xmlDoc.DocumentElement?.Attributes?["version"]?.Value;
            switch (version)
```

src/SwiftRuntimeLibrary/src/TypeDatabase.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TypeDatabase"/> class.
+        /// </summary>
+        /// <param name="file">The path to the XML file containing the type mappings.</param>
+        public TypeDatabase(string file)
+        {
+            XmlDocument xmlDoc = new XmlDocument();
+            xmlDoc.Load(file);
+            if (!ValidateXmlSchema(xmlDoc))
+                throw new Exception("Invalid XML schema.");
```
rolfbjarne:
```suggestion
                throw new Exception("Invalid XML schema in {0}.", file);
```

src/SwiftRuntimeLibrary/src/TypeDatabase.cs
kotlarmilos: Do you think we should change the .csproj name to `SwiftRuntime.csproj`/`SwiftRuntime.Tests.csproj`?

rolfbjarne: Maybe Swift.Runtime[.Test].csproj?

kotlarmilos: Changed to Swift.Bindings and Swift.Runtime for clarity.

src/Swift.Runtime/src/TypeDatabase.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        /// <summary>
+        /// Checks if a Swift type name corresponds to a known C# value type.
+        /// </summary>
+        /// <param name="swiftName">The Swift type name.</param>
+        /// <returns><c>true</c> if the Swift type name maps to a C# value type; otherwise <c>false</c>.</returns>
+        public bool IsValueType(string swiftName)
+        {
+            return _swiftToCSharpMapping.TryGetValue(swiftName, out string? csharpName);
```
matouskozak: Do we expect the mapping in `src/Swift.Runtime/src/TypeDatabase.xml` to grow in the future so that it would contain also other types that don't match to C# value types? For example, strings?

src/Swift.Bindings/src/Model/MethodDecl.cs
```diff
@@ -0,0 +1,26 @@
...
+        
+        /// <summary>
+        /// Mangled name of the method.
+        /// </summary>
+        public required string MangledName { get; init; }
+        
+        /// <summary>
+        /// Signature of the method.
+        /// </summary>
+        public required IEnumerable<TypeDecl> Signature { get; set; }
```
matouskozak: I wonder if having signature represented just by a simple `IEnumerable<TypeDecl>` will be enough in the future or if it would be better to declare `SignatureDecl` and split return and params into separate fields. 

My concerns here is the possibility of implicit `self`/`this` but I'm unsure if that will concern us in the future. Nonetheless, for the prototype we can keep this for simplicity.

src/Swift.Bindings/src/Parser/BindingsGenerator.SwiftABIParser.cs
```diff
@@ -0,0 +1,130 @@
...
+        [JsonProperty("ABIRoot")]
+        public required RootNode ABIRoot { get; set; }
+    }
+
+    /// <summary>
+    /// Represents the root node of a module.
+    /// </summary>
+    public sealed class RootNode
+    {
+        public required string Kind { get; set; }
```
matouskozak, line 23: Maybe enum would be a better long term type for `Kind` but I suppose we will move away from parsing Swift ABI so it is probably okay to use string for now.

src/Swift.Runtime/src/ISwiftObject.cs
matouskozak, line 1: Is the ISwiftObject used for something? I couldn't find its usage in the PR.

matouskozak (approved):
Good job! Looking good and much easier to read through than the original big PR. I put some comments but they are more questions about the future so no need to address them in this PR.

src/Swift.Runtime/src/TypeDatabase.cs
kotlarmilos: The `TypeDatabase.xml` should provide mappings for common C# types. It can be expanded in the future, and we need to project each type pattern individually. For non-trivial mappings, a projection will be created within the `Swift.Runtime` namespace (i.e. unsafe buffer pointers). Currently, the type database attempts to find a known mapping. If not found, the type is then searched for within `Swift.Runtime` namespace:
```csharp
/// <summary>
/// Gets the C# type name corresponding to the specified Swift type name. 
/// The method first tries to find a known mapping, and if that fails, it looks for a type in Swift.Runtime.
/// </summary>
/// <param name="swiftTypeName">The Swift type name.</param>
/// <returns>The corresponding C# type name.</returns>
public string[] GetCSharpTypeName(string swiftTypeName)
{
    // Try to find a known mapping
    if (_swiftToCSharpMapping.TryGetValue(swiftTypeName, out string? csharpTypeName))
    {
        return new string [] { "System", csharpTypeName };
    }

    // Try to find a type in Swift.Runtime
    Type? swiftRuntimeType = Type.GetType($"Swift.Runtime.{swiftTypeName}");
    if (swiftRuntimeType != null)
        return new string [] { "Swift.Runtime", swiftRuntimeType.Name };
    
    throw new Exception($"No mapping for {swiftTypeName} type found.");
}
```

src/Swift.Bindings/src/Model/MethodDecl.cs
kotlarmilos: It's a good idea, and we can consider implementing it in the future. Swift ABI signature is presented in this way (return type + params).

src/Swift.Bindings/src/Parser/BindingsGenerator.SwiftABIParser.cs
kotlarmilos: If `Kind` was implemented as enum, it would impact the JSON deserialization process, expecting `int` type instead of `string`:
```csharp
var abiRoot = JsonConvert.DeserializeObject<ABIRootNode>(jsonContent);
```

> will move away from parsing Swift ABI so it is probably okay to use string for now.

I think it is a good idea to wait until parser becomes more stable.

src/Swift.Runtime/src/ISwiftObject.cs
kotlarmilos: Not needed, removed for now.

kotlarmilos:
This PR includes projection tooling skeleton and implements CryptoKit dev templates with unsafe (buffer) pointers. We will move forward with surfacing https://developer.apple.com/documentation/cryptokit/chachapoly enum and https://developer.apple.com/documentation/cryptokit/symmetrickey struct in a subsequent PR.

Please review this PR and provide feedback.

.gitignore
```diff
@@ -190,3 +190,15 @@ Session.vim
...
+*.swiftinterface*
+*.swiftdoc*
+*.swiftmodule*
+*.abi*
+*.swiftsourceinfo*
+*.dylib
+
+# Testing artifacts
+testing/
+src/samples/**/*Bindings.cs
```
AaronRobinsonMSFT, line 204: These should all go under `artifacts/`. In the current experiment form, this seems okay, but prior to merging into dotnet/runtime, all outputs should be under `artifacts/`.

AaronRobinsonMSFT: Oops. Sorry, I thought we were working in a dotnet/runtime fork. Never mind.

src/Swift.Runtime/src/TypeDatabase.cs
matouskozak: Thank you for the explanation

docs/functional-outline.png
AaronRobinsonMSFT, line 1: This would ideally be an SVG instead of PNG.

src/Swift.Bindings/src/Model/MethodDecl.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace BindingsGeneration
+{
+    /// <summary>
+    /// Represents a method declaration.
+    /// </summary>
+    public sealed record MethodDecl
```
AaronRobinsonMSFT, line 9: Using `record`s in public APIs hasn't been confirmed as yet. This is likely okay for the current state. Just be aware that if this is going to be publicly consumable, the API review team may push back.

AaronRobinsonMSFT (approved):
LGTM

docs/README.md
```diff
@@ -44,6 +44,8 @@ The table below lists the Swift types and their corresponding C# types.
 | `Swift.UInt16`                  | `ushort` |
 | `Swift.Int8`                    | `sbyte`  |
 | `Swift.UInt8`                   | `byte`   |
+| `Swift.UnsafePointer`           | `void*`  |
```
jkotas: I know we touched on this before - it does not feel right to drop the type information for the strongly typed pointers. Strong typing is an extra level of protection against programming bugs that we are giving away here.

src/Swift.Runtime/src/Library/UnsafeBufferPointer.cs
```diff
@@ -0,0 +1,72 @@
...
+        public nint Count => (nint)((byte*)_End - (byte*)_Position);
+    }
+
+    // <summary>
+    // Represents Swift UnsafeBufferPointer in C#.
+    // </summary>
+    public unsafe readonly struct UnsafeBufferPointer
+    {
+        private readonly void* _Position;
+        public readonly nint Count;
```
jkotas: Swift does not seem to have fields as first-class concept. Their docs only talk about properties: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/properties/

I am wondering whether mapping stored properties to fields in C# is conceptually correct.

Is it a breaking change in Swift to change a stored property to a computed property in non-frozen type? I would expect that it is not a breaking change - can you confirm that?

If computed properties and stored properties are interchangeable in Swift, we should map both to properties in C#. Ie make this `public nint Count { get; };` or `private nint _count; public nint Count => _count;`.

docs/README.md
kotlarmilos: Ok, as we discussed, a thin layer would be useful here.

rolfbjarne (approved):

src/Swift.Runtime/src/Library/UnsafeBufferPointer.cs
kotlarmilos: Good catch. This is not a breaking change in Swift. I confirmed it by consuming a library with struct that evolved from
```swift
public struct MyType {
    public var myProperty: Int = 0
    public init() {}
}
```
to
```swift
public struct MyType {
    private var _myProperty: Int = 0
    public var myProperty: Int {
        get { _myProperty }
        set { _myProperty = newValue }
    }

    public init() {}
}
```

docs/functional-outline.png
kotlarmilos: Updated.

docs/README.md
kotlarmilos: Types for unsafe pointers have been added. Mutable pointers are projected as structs with generics in C# to address methods overload issues and enable future extensibility.

src/Swift.Runtime/src/Library/UnsafeBufferPointer.cs
kotlarmilos: Buffer pointers are updated with `private readonly nint _count; public nint Count => _count;`.

src/Swift.Runtime/src/Library/UnsafeBufferPointer.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        private readonly void* _position;
+        private readonly nint _count;
+        public UnsafeBufferPointer(void* start, nint count)
+        {
+            _position = start;
+            _count = count;
+        }
+
+        public void* BaseAddress => _position;
```
jkotas:
```suggestion
        public T* BaseAddress => _position;
```

src/Swift.Runtime/src/Library/UnsafeBufferPointer.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        private readonly void* _position;
+        private readonly nint _count;
+        public UnsafeMutableBufferPointer(void* start, nint count)
+        {
+            _position = start;
+            _count = count;
+        }
+
+        public void* BaseAddress => _position;
```
jkotas:
```suggestion
        public T* BaseAddress => _position;
```

jkotas: All `void*` in the strongly typed wrappers should be `T*`.

src/Swift.Runtime/src/Library/UnsafePointer.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Swift.Runtime
+{
+    // <summary>
+    // Represents Swift UnsafePointer in C#.
+    // </summary>
+    public unsafe readonly struct UnsafePointer<T>
+    {
+        private readonly void* _rawValue;
+        public UnsafePointer(void* _rawValue)
```
jkotas: The argument name should not have underscore. Also, should the argument name match the property name that it corresponds to `UnsafePointer(void* poin_rawValue)`?

src/Swift.Runtime/src/Library/UnsafePointer.cs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Swift.Runtime
+{
+    // <summary>
+    // Represents Swift UnsafePointer in C#.
+    // </summary>
+    public unsafe readonly struct UnsafePointer<T>
```
jkotas: I think it would make sense to have implicit conversions to/from T* for these types.

jkotas: (And to/from `void*` for the non-generic variant.)

src/Swift.Runtime/src/Library/UnsafeBufferPointer.cs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    // <summary>
+    // Represents Swift UnsafeRawBufferPointer in C#.
+    // </summary>
+    public unsafe readonly struct UnsafeRawBufferPointer
+    {
+        private readonly void* _position;
+        private readonly void* _end;
+        public UnsafeRawBufferPointer(void* start, nint count)
```
jkotas:
```suggestion
        public UnsafeRawBufferPointer(void* baseAddress, nint count)
```
Match property name?

src/Swift.Runtime/src/Library/UnsafeBufferPointer.cs
stephen-hawley: IIRC, if a property is stored, you still get getters and setter generated by the compiler. Since BTfS doesn't deal with struct lowering, it maps properties (stored or otherwise) as C# properties.

stephen-hawley (approved):

PR #2514, opened 2024-02-10
NativeAOT-LLVM: Merge feb24

yowl:
This PR merges us to b5411e3a110cdbc643be3a922a150a9e083cd240 

Some changes included in this merge:

- Return type for `RhpCidResolveWasm_Managed` and `RhpResolveInterfaceDispatch` to `PCODE`
- Add some `FEATURE_EVENT_TRACE` types to ` TARGET_WASM`
- Remove `dotnet-legacy.d.ts`
- `Exchange`/`CompareExchange` for `byte` and `short`
- Add `CancelPromise`
- New helpers:
```c++
FUNC(CORINFO_HELP_READYTORUN_THREADSTATIC_BASE_NOCTOR) CORINFO_TYPE_PTR, { }, HFIF_SS_ARG
{ FUNC(CORINFO_HELP_VALUEPROFILE32) },
{ FUNC(CORINFO_HELP_VALUEPROFILE64) },
```

PR #2527, opened 2024-03-07
[NativeAOT-LLVM] GC stress and faster checked builds

SingleAccretion:
Adds a `JitGcStress` compiler knob that will cause codegen to insert calls to a GC-triggering helper after each safe point in the method. This makes GC holes dramatically easier to reproduce. I've now used this successfully to root cause the problem in #2514. We should also consider enabling this on some of our tests, though this change does not do that.

I am also making two tweaks to make the Checked build faster here:
1) Parallelize the writing of bitcode. This will also help large compiles.
2) _Disable_ writing out of textual IR by default (can still be enabled via `--codegenopt:JitCheckLlvmIR=1`). I've found that the tradeoff this makes in making the checked build slower means I am almost always using a Release build + `llvm-dis` instead.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -174,14 +174,10 @@ bool Llvm::callRequiresShadowStackSave(const GenTreeCall* call) const
     // back into managed code, we need to save the shadow stack pointer, so that the RPI frame can pick it up.
     // Another case where the save/restore is required is when calling into native runtime code that can trigger
     // a GC (canonical example: allocators), to communicate shadow stack bounds to the roots scan.
-    // TODO-LLVM-CQ: optimize the GC case by using specialized helpers which would sink the save/restore to the
-    // unlikely path of a GC actually happening.
-    // TODO-LLVM-CQ: we should skip the managed -> native -> managed transition for runtime imports implemented
-    // in managed code as runtime exports.
     //
     if (call->IsHelperCall())
     {
-        return helperCallRequiresShadowStackSave(_compiler->eeGetHelperNum(call->gtCallMethHnd));
+        return helperCallRequiresShadowStackSave(call->GetHelperNum());
     }
 
     // SPGCT calls are assumed to never RPI by contract.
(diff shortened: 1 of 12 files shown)


src/coreclr/jit/llvm.cpp
```diff
@@ -174,14 +174,10 @@ bool Llvm::callRequiresShadowStackSave(const GenTreeCall* call) const
     // back into managed code, we need to save the shadow stack pointer, so that the RPI frame can pick it up.
     // Another case where the save/restore is required is when calling into native runtime code that can trigger
     // a GC (canonical example: allocators), to communicate shadow stack bounds to the roots scan.
-    // TODO-LLVM-CQ: optimize the GC case by using specialized helpers which would sink the save/restore to the
-    // unlikely path of a GC actually happening.
-    // TODO-LLVM-CQ: we should skip the managed -> native -> managed transition for runtime imports implemented
-    // in managed code as runtime exports.
```
SingleAccretion, line 180: Both of these have been fixed.

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/nativeaot/Runtime/thread.cpp
```diff
@@ -392,13 +392,26 @@ void Thread::Destroy()
...
+extern OBJECTREF* GetShadowStackTop();
 
-void GcScanWasmShadowStack(ScanFunc * pfnEnumCallback, ScanContext * pvCallbackData)
+void Thread::GcScanWasmShadowStack(ScanFunc * pfnEnumCallback, ScanContext * pvCallbackData)
 {
     // Wasm does not permit iteration of stack frames so is uses a shadow stack instead
     EnumGcRefsInRegionConservatively(GetShadowStackBottom(), GetShadowStackTop(), pfnEnumCallback, pvCallbackData);
+
+    // TODO-LLVM-Upstream: unify this method with the general "GcScanRootsWorker" below.
+    for (GCFrameRegistration* pCurGCFrame = m_pGCFrameRegistrations; pCurGCFrame != NULL; pCurGCFrame = pCurGCFrame->m_pNext)
```
SingleAccretion, line 405: These GC frames are usually only used for dynamic/reflection invokes (that need to preserve a dynamic number of GC pointers).

yowl (approved):
Thanks, this is very useful.

PR #2531, opened 2024-03-22
[NativeAOT-LLVM] Fix a GC hole in the new LSSA implementation 

SingleAccretion:
We need to be careful about not depending on stale shadow stack state when making decisions not to spill a def derived from an already spilled value.

This has some cost, but an overly large one, about 0.15% in terms of code size for HelloWasm.

Ref: https://github.com/dotnet/runtimelab/pull/2514#issuecomment-1979808170.

### DIFF
--- a/src/coreclr/jit/llvmlssa.cpp
+++ b/src/coreclr/jit/llvmlssa.cpp
@@ -226,8 +226,21 @@ private:
             }
         };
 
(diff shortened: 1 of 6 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
I'm not claiming to have a good understanding of this, but the coverage looks good.

jkotas:
Thank you!

PR #2535, opened 2024-03-29
[NativeAOT-LLVM] Consolidate HelloWasm IL projects

SingleAccretion:
We don't need separate IL projects for all IL. Consolidating them into one slightly improves build time.

Also deletes the unused `HelloWasm.sln`.

No functional changes.


SingleAccretion:
`windows-x64 Debug` failure is a know upstream problem.

@dotnet/nativeaot-llvm

yowl (approved):
Thanks!

PR #2547, opened 2024-04-15
[NativeAOT-LLVM] Do not force sections for 'ordinary' data

SingleAccretion:
Forcing sections disables the default behavior of `-fdata-sections`, inhibiting linker GC, and also forces strict alignment packing.

This does not save a lot, due to how the ILC-produced data is structured, but does save a bit: 5K for WasmDebugging, 12K for HelloWasm.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

PR #2551, opened 2024-04-19
[NativeAOT-LLVM] Enable Task, JSObject and delegate JS interop

maraf:
- Enable task marshalling
- Enable delegate marshalling
- Clean up legacy interop methods removed upstream
- Fix binding JSExport without namespace
- Update smoke test to include Task, JSObject and delegates


maraf:
cc @dotnet/nativeaot-llvm

src/tests/nativeaot/SmokeTests/DotnetJs/wwwroot/main.js
```diff
@@ -34,5 +35,54 @@ if (concat != "AaaBbb") {
     result = 15;
 }
 
+let isPromiseResolved = false;
+let promise = new Promise(resolve => setTimeout(() => { console.log("Promise resolved"); isPromiseResolved = true; resolve(); }, 2000));
```
SingleAccretion: 2s is quite a bit of time to wait for a test. Could it be lowered (to something like 100ms)? (Same below).

src/tests/nativeaot/SmokeTests/DotnetJs/wwwroot/main.js
```diff
@@ -34,5 +35,54 @@ if (concat != "AaaBbb") {
...
+}
+if (asyncResult != 87) {
+    result = 17;
+}
+
+try {
+    isPromiseResolved = false;
+    promise = new Promise(resolve => setTimeout(() => { console.log("Promise resolved"); isPromiseResolved = true; resolve(); }, 2000));
+    asyncResult = await exports.DotnetJsApp.Program.Interop.Async(promise, true);
+    console.log(`Async result: ${asyncResult}`);
```
SingleAccretion:
```suggestion
    console.log(`Unexpected async result: ${asyncResult}`);
```
?

src/tests/nativeaot/SmokeTests/DotnetJs/wwwroot/main.js
```diff
@@ -34,5 +35,54 @@ if (concat != "AaaBbb") {
...
+    if (!isPromiseResolved) {
+        result = 18;
+    }
```
SingleAccretion: Shouldn't this be in the catch branch?

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
```diff
@@ -44,5 +49,65 @@ internal static string Concat(string a, string b)
...
+            return 87;
+        }
+
+        [JSExport]
+        internal static async Task<int> AsyncWithCancel()
+        {
+            Console.WriteLine("Fire HTTP");
+            using var cts = new CancellationTokenSource();
+            using var http = new HttpClient();
+            var responseTask = http.GetStringAsync("https://github.com", cts.Token);
```
SingleAccretion: Isn't this test racy?

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
```diff
@@ -44,5 +49,65 @@ internal static string Concat(string a, string b)
...
+            Console.WriteLine($"Property x is '{x}'");
+            jsObject.SetProperty("y", x + 1);
+            return x == 42;
+        }
+        
+        [JSExport]
+        internal static void Func([JSMarshalAs<JSType.Function<JSType.String>>()] Func<string> getLocation, [JSMarshalAs<JSType.Function<JSType.String>>()] Action<string> log)
+        {
+            Console.WriteLine($"Got Func '{getLocation}' and '{log}'");
+            log($"Running at '{getLocation()}'");
```
SingleAccretion: It would be good to test the output of `getLocation`. E. g. make it a simple `() => "Some magic string"` on the JS side, `log(getLocation())` here and verify that the same string comes back in the messages on the JS side.

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
```diff
@@ -44,5 +49,65 @@ internal static string Concat(string a, string b)
...
+        {
+            Console.WriteLine($"Got JSObject '{jsObject}'");
+            var x = jsObject.GetPropertyAsInt32("x");
+            Console.WriteLine($"Property x is '{x}'");
+            jsObject.SetProperty("y", x + 1);
+            return x == 42;
+        }
+        
+        [JSExport]
+        internal static void Func([JSMarshalAs<JSType.Function<JSType.String>>()] Func<string> getLocation, [JSMarshalAs<JSType.Function<JSType.String>>()] Action<string> log)
```
SingleAccretion:
```suggestion
        internal static void TestDelegateMarshalling([JSMarshalAs<JSType.Function<JSType.String>>()] Func<string> getLocation, [JSMarshalAs<JSType.Function<JSType.String>>()] Action<string> log)
```

src/mono/browser/runtime/cwraps.ts
```diff
@@ -319,12 +319,15 @@ function cwrap (name: string, returnType: string | null, argTypes: string[] | un
...
-
+    const fns = [...fn_signatures];
     for (const sig of fns) {
         const wf: any = wrapped_c_functions;
         const [lazyOrSkip, name, returnType, argTypes, opts] = sig;
         const maybeSkip = typeof lazyOrSkip === "function";
+        if (NativeAOT) {
+            wf[name] = () => name;
+            continue;
+        }
```
SingleAccretion: I am confused by this. Doesn't this basically turn all functions on the interface to `return <function mame>;` no-ops?

src/mono/browser/runtime/managed-exports.ts
```diff
@@ -389,6 +386,15 @@ function bind_assembly_exports_naot (assembly: string) {
 export const bind_assembly_exports: (assemblyName: string) => Promise<void> = NativeAOT ? bind_assembly_exports_naot : bind_assembly_exports_mono;
 
 function get_method (method_name: string): MonoMethod {
+    if (NativeAOT) {
+        const fqn = runtimeHelpers.runtime_interop_namespace + "." + runtimeHelpers.runtime_interop_exports_classname + "." + method_name;
+        const exportName = `_${fqn.replace(/\./g, "_")}`;
+        const exportFunc = (Module as any)[exportName];
+        return exportFunc ?? (() => {
+            throw new Error(`Wasm export ${fqn} not found`);
+        });
```
SingleAccretion, line 395: Is `CompleteTask` the only export that's relevant for NAOT?

src/mono/browser/runtime/managed-exports.ts
```diff
@@ -389,6 +386,15 @@ function bind_assembly_exports_naot (assembly: string) {
 export const bind_assembly_exports: (assemblyName: string) => Promise<void> = NativeAOT ? bind_assembly_exports_naot : bind_assembly_exports_mono;
 
 function get_method (method_name: string): MonoMethod {
+    if (NativeAOT) {
+        const fqn = runtimeHelpers.runtime_interop_namespace + "." + runtimeHelpers.runtime_interop_exports_classname + "." + method_name;
+        const exportName = `_${fqn.replace(/\./g, "_")}`;
+        const exportFunc = (Module as any)[exportName];
```
SingleAccretion, line 392:
```suggestion
        const exportName = `_${runtimeHelpers.runtime_interop_namespace}_${runtimeHelpers.runtime_interop_exports_classname}_${method_name}`;
        const exportFunc = (Module as any)[exportName];
```
?

SingleAccretion (commented):
How far are we from running S.R.IS.JavaScript.Tests in CI?

maraf:
> How far are we from running S.R.IS.JavaScript.Tests in CI?

Just async main, but I think we can tweak the test runner to actually run in JSExport. I would like to stand-up more tests in general. Let's discuss on discord and continue in follow-up

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
maraf: We have similar test upstream. The `BrowserHttpHandler` always does the async path

SingleAccretion: So even if the `Task.Delay` below is increased to something unreasonable like a minute, this is still expected to succeed?

src/mono/browser/runtime/managed-exports.ts
maraf: We would have to modify `runtime_interop_namespace` to replace `.` by `_`

SingleAccretion: Ah, indeed.

src/mono/browser/runtime/managed-exports.ts
maraf: And `CallDelegate` (which don't have tested yet). My motivation was to minify the diff
EDIT: Added test for `CallDelegate`

src/mono/browser/runtime/cwraps.ts
maraf: Yes. It allows for `cwraps` to be called and returns "a value". At JS functions we are using it is enough. My motivation was to minify the diff

SingleAccretion: 👍. Can it be more obviously no-op, like return an empty object or some such?

In any case, it needs a comment explaining this.

src/tests/nativeaot/SmokeTests/DotnetJs/Program.cs
maraf: I'm fine removing the test. We don't support marshalling cancellation tokens and wiring them to `AbortController` ATM. So only HTTP and JS module import currently supports cancellation.

maraf: Or we can drop the delay and let the C# code run synchronously to be sure

SingleAccretion: > So only HTTP and JS module import currently supports cancellation.

Ahh, so that's how things are.

My preference would be to find some HTTP endpoint that never responds and make the test robust via that. Next option is keeping it as is, next is removing. Up to you.

maraf: Found a ~~super~~ quite slow one

SingleAccretion (approved):
LGTM. Thank you!

Issue #2455, opened 2023-12-08
[NativeAOT-LLVM] Problems with libraries tests


SingleAccretion:
Follow-ups to #2452:

1) Running one test twice for different arches uses stale artifacts.
   -- Publish directory doesn't include arch postfix.
2) XML output is useless (floods the console).
3) XML output is not collected anywhere (probably fine as long as we're not using Helix).
4) Larger tests fail to run when compiled in Debug.
   -- This is a codegen issue.
5) Even larger tests will fail to reserve static memory in the linker.
    - This will be fixed upstream: https://github.com/emscripten-core/emscripten/issues/20888.

yowl:
`CreateSaturatingFromNFloatTest`s and `CreateTruncatingFromIntPtrTest `s are `#if`'ed out.  An implementation is needed for LLVM.

PR #2552, opened 2024-04-20
[NativeAOT-LLVM] Implement saturating semantics for float to int conversions

SingleAccretion:
Using the same intrinsics Rust uses, and which conveniently correspond exactly both to .NET semantics _and_ WASM's non-trapping conversion instructions, which we now enable by default via `-mnontrapping-fptoint`.

Ref: https://github.com/dotnet/runtimelab/issues/2455#issuecomment-2067534258.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (commented):
Thanks!

yowl (approved):

PR #2560, opened 2024-04-23
Add basic support for frozen structs projections

kotlarmilos:
## Description

This PR implements basic support for frozen structs. It refactors the tooling, including the type database and declarations, to manage recursive dependencies. It introduces Swift metadata projections and type registrar system for mapping between Swift and C#. Additionally, it updates the parser and emitter to interact with the type database and registrar, and implement basic type filtering. Tests are added for non-generic nested frozen structs with primitive properties.


stephen-hawley (commented):
One thing you're going to run into in your code is that you don't have a distinction between a type declaration and a type specification. This will cause you problems later on.

Here's what the distinction is:
a type declaration is a nominal type in swift. It is anything that has a name and its public facing API. This includes top-level functions, structs, enums, classes, and actors.

A type specification is the annotation that is used to specify the type of a parameter, an associated type, a type alias target, or the return type of a method. In swift, a type specification can include nominal types (fully or partially specified), tuples, closure types, and protocol list types. In addition, parts of these types may include syntactic sugars such as option, implicitly unwrapped options, and meta information such as attributes. All of those elements need to be handled correctly.

Indeed, the representation of these types is, unto itself, a little language.

I strongly recommend that you take `TypeSpecParser`, `TypeSpecTokenizer`, `TypeSpecToken`, and `TypeSpec (and its subclasses). This represents a very compact recursive descent parser than can handle whatever the swift compiler can generate.

kotlarmilos:
I've converted this PR to draft as we plan to have additional iterations before another review. @stephen-hawley thank you for your contribution. Here are some thoughts:
- I like your approach with the conductor class and factories, but I don't see the value of having an additional layer on top of the declarations, such as `TypeEnv`, `MethodEnv`, and `ArgumentEnv`. I think we should operate with a single source of truth, which should be the declarations. Do you see anything missing in the declarations?
- I suggest moving handlers to emitters, as they actually emit code. Perhaps we can name them `{Name}Emitter`?
- Can we use `FieldDecl` instead of `ArgumentDecl`?
- It looks like the factory should replace `ModuleEmitter` and be invoked from `Program.cs`.
- I wonder if we can simplify `IFactory`. The codegen might bloat with such generics usage, and we may try to simplify it. I imagine that in most cases, we would rely on string representation and not actual generic types in the `IFactory` implementations.

Let's integrate your changes and add comments, documentation, and more test cases.

kotlarmilos:
> I strongly recommend that you take TypeSpecParser, TypeSpecTokenizer, TypeSpecToken, and `TypeSpec (and its subclasses). This represents a very compact recursive descent parser than can handle whatever the swift compiler can generate.

Let's add them in a follow-up PR as this one is already ~+5k loc. What is the difference between `TypeDecl` and `TypeSpec`?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
```diff
@@ -0,0 +1,189 @@
...
+    /// </summary>
+    public class ClassHandlerFactory : IFactory<BaseDecl, ITypeHandler>
+    {
+        /// <summary>
+        /// Determines if the factory handles the specified declaration.
+        /// </summary>
+        /// <param name="decl">The base declaration.</param>
+        public bool Handles(BaseDecl decl)
+        {
+            return decl is TypeDecl && decl is ClassDecl;
```
stephen-hawley: `decl is TypeDecl &&` isn't needed. If the type is a `ClassDecl` it is always a `TypeDecl`

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
```diff
@@ -0,0 +1,189 @@
...
+    /// </summary>
+    public class StructHandlerFactory : IFactory<BaseDecl, ITypeHandler>
+    {
+         /// <summary>
+        /// Determines if the factory handles the specified declaration.
+        /// </summary>
+        /// <param name="decl">The base declaration.</param>
+        public bool Handles(BaseDecl decl)
+        {
+            return decl is TypeDecl && decl is StructDecl;
```
stephen-hawley: `decl is TypeDecl && ` isn't needed.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
```diff
@@ -0,0 +1,189 @@
...
+                // TODO: Fix memory access violation
+                // // Verify field against Swift type information
+                // if (swiftTypeInfo.HasValue && !VerifyFieldRecord(swiftTypeInfo.Value, structDecl.Fields.IndexOf(fieldDecl), fieldDecl))
+                // {
+                //     Console.WriteLine("Field record does not match the field declaration");
+                // }
+            }
+            writer.WriteLine();
+
+            foreach (BaseDecl baseDecl in structDecl.Declarations)
```
stephen-hawley: This should be done through the conductor, otherwise we will have identical code for classes, structs, enums, and actors when it could be in one place.

stephen-hawley (changes requested):

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
kotlarmilos: I created `BaseHandler` that implements `HandleBaseDecl` which calls into `Conductor`. Do you have any better ideas here?

stephen-hawley: How about putting an `EmitInnerTypes` method into `BaseHandler` which looks like:
```
protected EmitInnerTypes(IEnumerable<TypeDecl> types, IndentedTextWriter writer, Conductor conductor, TypeDatabase typeDatabase) {
    writer.Indent++;
    foreach (var type in types) {
        HandleBaseDecl(writer, baseDecl, conductor, typeDatabase);
    }
    writer.Indent--;
}
```

kotlarmilos: Moved the loop to the `HandleBaseDecl`. I didn't change the name of the method as it handles other declarations as well. We can rename it later.

stephen-hawley (approved):

kotlarmilos:
Merging this PR. Any additional feedback will be addressed in follow-up work.

Issue #2558, opened 2024-04-23
[NativeAOT-LLVM] Re-enable the `DotNetJsApi` smoke test


SingleAccretion:
We disabled it in #2543 to get things merging with the expectation of re-enabling it soon.

maraf:
```
EXEC : warning : undefined symbol: mono_wasm_bind_cs_function (referenced by root reference (e.g. compiled C/C++ code)) [D:\a\_work\1\s\src\tests\nativeaot\SmokeTests\DotnetJs\DotnetJs.csproj] [D:\a\_work\1\s\src\tests\build.proj]
```

PR #2561, opened 2024-04-24
[NativeAOT-LLVM] Re-enable DotnetJs smoke test

maraf:
Fixes https://github.com/dotnet/runtimelab/issues/2558

### DIFF
--- a/src/mono/browser/runtime/exports-binding.ts
+++ b/src/mono/browser/runtime/exports-binding.ts
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+import NativeAOT from "consts:nativeAOT";
 import WasmEnableThreads from "consts:wasmEnableThreads";
 
 import { mono_wasm_debugger_log, mono_wasm_add_dbg_command_received, mono_wasm_set_entrypoint_breakpoint, mono_wasm_fire_debugger_agent_message_with_data, mono_wasm_fire_debugger_agent_message_with_data_to_pause } from "./debug";
@@ -33,6 +34,7 @@ import {
 } from "./pthreads";
 import { mono_wasm_dump_threads } from "./pthreads/ui-thread";
 import { mono_wasm_schedule_synchronization_context } from "./pthreads/shared";
+import { mono_wasm_bind_cs_function } from "./invoke-cs";
 
 // the JS methods would be visible to EMCC linker and become imports of the WASM module
 
@@ -113,6 +115,10 @@ export const mono_wasm_imports = [
     mono_wasm_get_first_day_of_week,
     mono_wasm_get_first_week_of_year,
 ];
+
+if (NativeAOT) {
+    mono_wasm_imports.push(mono_wasm_bind_cs_function as any);
+}
 
 const wasmImports: Function[] = [
     ...mono_wasm_imports,
--- a/src/mono/browser/runtime/exports-binding.ts
+++ b/src/mono/browser/runtime/exports-binding.ts
@@ -33,7 +33,6 @@ import {
 } from "./pthreads";
 import { mono_wasm_dump_threads } from "./pthreads/ui-thread";
 import { mono_wasm_schedule_synchronization_context } from "./pthreads/shared";
-import { mono_wasm_bind_cs_function } from "./invoke-cs";
 
 // the JS methods would be visible to EMCC linker and become imports of the WASM module
 
--- a/src/mono/browser/runtime/exports-binding.ts
+++ b/src/mono/browser/runtime/exports-binding.ts
@@ -1,7 +1,6 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-import NativeAOT from "consts:nativeAOT";
 import WasmEnableThreads from "consts:wasmEnableThreads";
 
 import { mono_wasm_debugger_log, mono_wasm_add_dbg_command_received, mono_wasm_set_entrypoint_breakpoint, mono_wasm_fire_debugger_agent_message_with_data, mono_wasm_fire_debugger_agent_message_with_data_to_pause } from "./debug";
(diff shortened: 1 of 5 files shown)


src/mono/browser/runtime/managed-exports.ts
```diff
@@ -374,21 +374,16 @@ function bind_assembly_exports_mono (assemblyName: string): Promise<void> {
...
-        let assemblyWithoutExtension = assembly;
-        if (assemblyWithoutExtension.endsWith(".dll")) {
-            assemblyWithoutExtension = assemblyWithoutExtension.substring(0, assembly.length - 4);
-        }
-        const register = (Module as any)["_" + assemblyWithoutExtension + "__GeneratedInitializer" + "__Register_"];
-        mono_assert(register, `Missing wasm export for JSExport registration function in assembly ${assembly}`);
-        register();
+    loaderHelpers.assert_runtime_running();
+    let assemblyWithoutExtension = assembly;
+    if (assemblyWithoutExtension.endsWith(".dll")) {
```
pavelsavara, line 380: no webcil ?

jkotas:
@dotnet/nativeaot-llvm

src/mono/browser/runtime/exports-binding.ts
```diff
@@ -114,6 +116,10 @@ export const mono_wasm_imports = [
     mono_wasm_get_first_week_of_year,
 ];
 
+if (NativeAOT) {
+    mono_wasm_imports.push(mono_wasm_bind_cs_function as any);
```
SingleAccretion: We discussed yesterday why we need this: because of the `Interop.Runtime.BindCSFunction` call in the original fix that is/was a statically linked function.

Now the code calls `JavaScriptImports.BindCSFunction` that is a list of JSImports which are (presumably) dynamically resolved. So why is this still needed?

maraf: This is the of functions that are passed to emcc to be linked as wasm imports. The definition of the list is "static" array

src/mono/browser/runtime/managed-exports.ts
maraf: What are mono users passing as assembly extension for JSExports?

src/mono/browser/runtime/exports-binding.ts
SingleAccretion: I understand - the question is what code references `mono_wasm_bind_cs_function` statically now (it used to be line 286)?

pavelsavara: I'm not sure if I understand the question. But will try to answer it :)
We are never trimming any TS code based on emcc linker. It's separate `dotnet.runtime.js` module which is part of runtime pack. It is not part of the emcc compilation/linker on the customer dev machine. That's just `dotnet.native.js`

SingleAccretion: Here's my question. First, we observe the failure mode we saw when merging: https://github.com/dotnet/runtimelab/issues/2558#issuecomment-2074576866.
Q: what does this warning mean?
A: it means that emcc failed to resolve the `mono_wasm_bind_cs_function` symbol.
Q: why was that?
A: because it was removed from `exports-binding.ts`, which is the mechanism used to connect (via a rather complex scheme) TS code to Emscripten-produced code.
Q: but where was it referenced in the first place?
A: the original fix called `Interop.Runtime.BindCSFunction` (on that line 286).

This change adds `mono_wasm_bind_cs_function` to `exports-binding.ts`, but _deletes_ the `Interop.Runtime.BindCSFunction` call. So, is `mono_wasm_bind_cs_function` referenced somewhere else as well?

maraf: Oh, great catch! The import is not needed now and I'll remove it. It was needed before I unified the call on C# side and didn't remove it afterwards

src/mono/browser/runtime/managed-exports.ts
pavelsavara: I have not realized this is coming from the user. Perhaps we need similar logic for mono too ?

src/mono/browser/runtime/exports-binding.ts
pavelsavara: so what is calling `mono_wasm_bind_cs_function` now ? something still has to do the marshaling on the JS side, right ?

src/mono/browser/runtime/managed-exports.ts
maraf: Here we need to strip the extension for get a name for wasm export. Whether extension should be optional or not on mono is a different question. But based on lazy loading, we probably should be supporting extension agnostic assembly names

src/mono/browser/runtime/exports-binding.ts
maraf: It's JSImport now

pavelsavara: so it will be on `INTERNAL` js object rather than C symbol.

SingleAccretion: By the way - is this JSImport binding relative to `globalThis` or something else? In other words - will it work for the multiple-runtimes-on-the-same-page (aka native-library) scenario?

maraf: JSImports are matched with objects passed `setModuleImports` with same shortcuts for matchin internal API. Nothing global involved

pavelsavara: there is `globalThis.getDotnetRuntime(0).INTERNAL` where runtime `runtimeId` is `0` if you are the first runtime installed in that page. 
You can get that id when you start the runtime. 

```
const runtime = await dotnet.create();
// runtime.runtimeId
// runtime.INTERNAL
setModuleImports("main.js", {
        Sample: {
            Test: {
                displayMeaning
            }
        }
});
```

You can have multiple runtimes on the same page, but you have to create them from separate ES6 modules.
```
const { dotnet: dotnet2 } = await import('./_framework/dotnet.js?instance=2-' + guid);
const runtime2 = await dotnet2.create();
```

The same ES6 module can't have multiple runtimes because emscripten/runtime instance state is not encapsulated enough.
For multi-threaded runtime it probably doesn't work, because there are `postMessage` with workers, but I'm not sure.

SingleAccretion (approved):
LGTM, thank you!

PR #2562, opened 2024-04-24
[NativeAOT-LLVM] Resolve some warnings, replace some `std` types with Jit types

### LINKED ISSUE #2554: [NativeAOT-LLVM] remove pragma disabled warnings
`llvm.h` contains several warnings disabled via `pragma`.  Some are in LLVM headers files which can remain with a `push` and `pop` .  Those in clrjit source files need to be fixed.

SingleAccretion:
The first commit fixes #2554.

The following few commits replace the usage of various `std` types, mostly `std::vector` with their JIT equivalents. This is for a couple reasons:
1) Jit types plug into the Jit allocator, which is both much faster and integrated into the Jit's memory usage diagnostics (`CMK_XYZ`).
2) We cannot really use the standard collections (see https://github.com/dotnet/runtime/pull/101088).
3) Consistency with the rest of the Jit.

The totality of the changes are zero-diff, except for the last commit, where I switched to `jitstd::sort` and some diffs were expected.

I will note that with regards to `2`, we are still not functionally correct, since LLVM code (and some of our own code) may still call standard library functions which may throw. The primary concern is OOM, since, fortunately, LLVM itself doesn't use exceptions, and there are a couple solutions which could be employed. Luckily, this is not a problem that needs to be solved here and now.

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -27,10 +27,6 @@ extern ICorJitHost* g_jitHost;
 
 #ifdef TARGET_WASM
 #include "llvm.h"
-
-#define max(a, b) (((a) > (b)) ? (a) : (b))
-#define min(a, b) (((a) < (b)) ? (a) : (b))
-
 #endif // TARGET_WASM
 
 unsigned Compiler::jitTotalMethodCompiled = 0;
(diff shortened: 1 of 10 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

jkotas (approved):
Thanks!

PR #2563, opened 2024-04-25
Add new official build for runtimelab

agocke:
This brings in the same 1ES official build templates that are in runtime and are now required for all internal builds.


eng/common/templates-official/job/publish-build-assets.yml
```diff
@@ -68,7 +68,7 @@ jobs:
     - checkout: self
       fetchDepth: 3
       clean: true
-
```
jkotas: Revert changes in this file. The extra whitespace are from dotnet/runtime main

eng/common/templates-official/post-build/post-build.yml
```diff
@@ -39,7 +39,7 @@ parameters:
     displayName: Enable NuGet validation
```
jkotas: Same here

eng/pipelines/official.yml
```diff
@@ -0,0 +1,77 @@
+# This Yaml Document has been converted by ESAI Yaml Pipeline Conversion Tool.
+# Please make sure to check all the converted content, it is your team's responsibility to make sure that the pipeline is still valid and functions as expected.
```
jkotas: Delete this comment?

eng/pipelines/official.yml
```diff
@@ -0,0 +1,77 @@
...
+    exclude:
+    - '**.md'
+    - eng/Version.Details.xml
+    - .devcontainer/*
+    - .github/*
+    - docs/*
+    - LICENSE.TXT
+    - PATENTS.TXT
+    - THIRD-PARTY-NOTICES.TXT
+pr:
```
jkotas: Is this meant to have pr: trigger?

eng/pipelines/official.yml
```diff
@@ -0,0 +1,77 @@
+# This Yaml Document has been converted by ESAI Yaml Pipeline Conversion Tool.
```
jkotas: Should this file be named runtimelab-official.yml to make it possible to checkin to dotnet/runtime main, like [runtimelab.yml](https://github.com/dotnet/runtime/blob/main/eng/pipelines/runtimelab.yml) is checked in today?

This existing setup is meant to allow every runtimelab project to get the CI and build support automatically.

eng/common/templates-official/job/publish-build-assets.yml
agocke: Thanks, I was reviewing with whitespace changes hidden.

eng/pipelines/official.yml
agocke: Nope, fixed

agocke:
FYI for msft employees, internal run ID is 2438214

eng/pipelines/official.yml
agocke: Makes sense. It will still require customization for each branch, since it looks like a lot of the branches have custom build steps (e.g. NativeAOT-LLVM is different from swift)

jkotas: Yes, customization is expected.

jkotas:
> internal run ID is 2438214

Succeeded, but it did not publish the packages (e.g. https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-experimental/NuGet/runtime.browser-wasm.Microsoft.DotNet.ILCompiler.LLVM/)

eng/pipelines/runtimelab-official.yml
```diff
@@ -0,0 +1,60 @@
...
+  parameters:
+    templatePath: 'templates-official'
+
+- ${{ if and(ne(variables['System.TeamProject'], 'public'), ne(variables['Build.Reason'], 'PullRequest')) }}:
+  - name: TeamName
+    value: dotnet-core
+extends:
+  template: /eng/pipelines/common/templates/pipeline-with-resources.yml@self
+  parameters:
+    isOfficialBuild: true
```
SingleAccretion, line 30: Below in the file, there are places that check `isOfficialBuild` instead of hardcoding it. What is the intent behind that?

eng/pipelines/runtimelab-official.yml
```diff
@@ -0,0 +1,60 @@
...
+          platforms:
+          - windows_arm64
+          - OSX_x64
+          - Browser_wasm_win
+          - wasi_wasm_win
+          jobParameters:
+            templatePath: 'templates-official'
+            timeoutInMinutes: 300
+            isOfficialBuild: ${{ variables.isOfficialBuild }}
+            testGroup: innerloop
```
SingleAccretion, line 48: > testGroup: innerloop

Where is this parameter defined? I don't see it in https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/eng/pipelines/common/global-build-job.yml.

eng/pipelines/runtimelab-official.yml
```diff
@@ -0,0 +1,60 @@
...
+            postBuildSteps:
+            - template: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+              parameters:
+                isOfficialBuild: ${{ variables.isOfficialBuild }}
+                uploadLibrariesTests: ${{ eq(variables.isOfficialBuild, false) }}
+                uploadIntermediateArtifacts: ${{ variables.isOfficialBuild }}
+                librariesConfiguration: Release
+            ${{ if eq(variables.isOfficialBuild, true) }}:
+              buildArgs: -s clr.aot+libs+nativeaot.packages -c $(_BuildConfig)
```
SingleAccretion: This is an odd placement for `buildArgs`. I wonder if the indentation works out correctly.

SingleAccretion (commented):
Some high level questions.

What is it exactly that is required of the "official" `.yml` (that wasn't previously)?

Is it _required_ to have a separate `.yml` instead of using `runtimelab.yml` with `isOfficialBuild: true`, or is it a code organization choice? Can/should we delete `isOfficialBuild == true` paths from `runtimelab.yml` now (downstream and upstream)?

eng/pipelines/runtimelab-official.yml
```diff
@@ -0,0 +1,60 @@
...
+    stages:
+    - stage: Build
+      jobs:
+      - template: /eng/pipelines/common/platform-matrix.yml
+        parameters:
+          templatePath: 'templates-official'
+          jobTemplate: /eng/pipelines/common/global-build-job.yml
+          buildConfig: Release
+          platforms:
+          - windows_arm64
```
yowl: `windows_x64` ?

agocke:
> > internal run ID is 2438214
> 
> Succeeded, but it did not publish the packages (e.g. https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-experimental/NuGet/runtime.browser-wasm.Microsoft.DotNet.ILCompiler.LLVM/)

Yup, expected -- it's branch gated so we won't get any new packages until this is merged.

agocke:
> What is it exactly that is required of the "official" `.yml` (that wasn't previously)?

In short, we need to import an internal template that sets policy.

> 
> Is it _required_ to have a separate `.yml` instead of using `runtimelab.yml` with `isOfficialBuild: true`, or is it a code organization choice?

It's possible to share, but hard. The internal template requires certain tasks to be used, and those tasks aren't available in the public template. The current guidance is to split to internal and external.

>  Can/should we delete `isOfficialBuild == true` paths from `runtimelab.yml` now (downstream and upstream)?

Yes, although that's not strictly necessary.

eng/pipelines/runtimelab-official.yml
agocke: It is likely that some are auto-generated from the conversion script, and some were just written by me by hand. Thus you have me hardcoding things to true (because this is the official build) and the script not knowing what to do and pulling the variable.

eng/pipelines/runtimelab-official.yml
agocke: I will add that.

eng/pipelines/runtimelab-official.yml
```diff
@@ -0,0 +1,60 @@
...
+          - OSX_x64
+          - Browser_wasm_win
+          - wasi_wasm_win
+          jobParameters:
+            templatePath: 'templates-official'
+            timeoutInMinutes: 300
+            isOfficialBuild: ${{ variables.isOfficialBuild }}
```
SingleAccretion:
```suggestion
            isOfficialBuild: true
```

eng/pipelines/runtimelab-official.yml
```diff
@@ -0,0 +1,60 @@
...
+            testGroup: innerloop
+            postBuildSteps:
+            - template: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
+              parameters:
+                isOfficialBuild: ${{ variables.isOfficialBuild }}
+                uploadLibrariesTests: ${{ eq(variables.isOfficialBuild, false) }}
+                uploadIntermediateArtifacts: ${{ variables.isOfficialBuild }}
+                librariesConfiguration: Release
```
SingleAccretion:
```suggestion
                isOfficialBuild: true
                uploadIntermediateArtifacts: true
                librariesConfiguration: Release
```

eng/pipelines/runtimelab-official.yml
```diff
@@ -0,0 +1,60 @@
...
+            ${{ if eq(variables.isOfficialBuild, true) }}:
+              buildArgs: -s clr.aot+libs+nativeaot.packages -c $(_BuildConfig)
+    - template: /eng/pipelines/official/stages/publish.yml
+      parameters:
+        isOfficialBuild: ${{ variables.isOfficialBuild }}
```
SingleAccretion:
```suggestion
        isOfficialBuild: true
```

eng/pipelines/runtimelab-official.yml
SingleAccretion: Note that we don't really need `windows_arm64` here (and probably don't want it, since it doesn't get tested on PR CI).

eng/pipelines/runtimelab-official.yml
SingleAccretion:
```suggestion
```

eng/pipelines/runtimelab-official.yml
```diff
@@ -0,0 +1,60 @@
...
+          - windows_arm64
+          - OSX_x64
+          - Browser_wasm_win
+          - wasi_wasm_win
+          jobParameters:
+            templatePath: 'templates-official'
+            timeoutInMinutes: 300
+            isOfficialBuild: ${{ variables.isOfficialBuild }}
+            testGroup: innerloop
+            postBuildSteps:
```
SingleAccretion, line 50:
```suggestion
            buildArgs: -s clr.aot+libs+nativeaot.packages -c $(_BuildConfig)
            postBuildSteps:
```

SingleAccretion (commented):
Thank you for the explanation, it makes sense to me.

I have added a couple suggestions that hardcode the official build parameter and make the code a bit simpler.

Modulo that, LGTM. We can see to the `runtimelab.yml` changes separately.

I am assuming a version of this will go upstream. Is that correct?

jkotas (approved):

yowl (approved):
LGTM, thanks

jkotas:
> Also build tests in official build

I do not think we should be doing that. It makes official build flakier and the failures in official build are impossible for non-Microsoft people to see and debug.

agocke:
The current config is running tests, so I can instead just skip testing entirely.

jkotas:
> skip testing entirely.

Yes, that would be better.

yowl:
> The current config is running tests, so I can instead just skip testing entirely.

Sorry, my mistake.

agocke:
np, simple change. Looks like the internal build passed.

jkotas:
> I am assuming a version of this will go upstream. Is that correct?

https://github.com/dotnet/runtime/pull/101655

Issue #2539, opened 2024-04-01
NativeAOT-LLVM: package publishing broken


yowl:
HI @jkotas, could you please post the current package publishing error, the merge has broken it again?

Thanks,

jkotas:
It is failing with

![image](https://github.com/dotnet/runtimelab/assets/6668460/23135f23-0bdc-4e6b-a711-6f861bd70d7c)

I do not think that it was broken by the merge. 

We are required to switch to standardized YAML templates, and it looks like that the branches that have not switched yet fail the official builds now.

@agocke Could you please confirm that the above error is caused by missing official templates?

If it is the case, the fix is going to be to take another merge to pick up changes like https://github.com/dotnet/runtime/pull/99433 and then fix things up.

yowl:
I've started the next merge, as mentioned above, there are yml changes.  Will we be able to swtich to these new jobs do you think:
```
          - template: /eng/pipelines/common/platform-matrix.yml
            parameters:
              jobTemplate: /eng/pipelines/common/global-build-job.yml
              helixQueuesTemplate: /eng/pipelines/coreclr/templates/helix-queues-setup.yml
              buildConfig: checked
              platforms:
              - linux_x64
              - windows_x64
              - Browser_wasm_win
              - wasi_wasm_win
              jobParameters:
                timeoutInMinutes: 300
                buildArgs: -s clr.aot+libs+nativeaot.packages -c debug -rc $(_BuildConfig)
                postBuildSteps:
                  - template: /eng/pipelines/common/templates/runtimes/build-runtime-tests-and-send-to-helix.yml
                    parameters:
                      creator: dotnet-bot
                      testRunNamePrefixSuffix: CoreCLR_$(_BuildConfig)
```
i.e. the `helix-queues-setup.yml` and `build-runtime-tests-and-send-to-helix.yml`?  I will try `global-build-job.yml` to see how it goes.

jkotas:
`global-build-job.yml` and `helix-queues-setup.yml` look reusable I am not 100% sure about `build-runtime-tests-and-send-to-helix.yml`

SingleAccretion:
Note: we don't use helix for tests (the tests are run on the build machines).

agocke:
Yup the above issue is caused by the switch to internal builds.

I'm happy to fix up the internal pipeline, but we need to get the public pipeline swapped over. Most important is to just get a merge from runtime-main. I can probably take it from there.

yowl:
[...]

PR #2566, opened 2024-04-27
FIx publish in official builds

jkotas:
Update the platform matrix and build command line for official builds to match what it was before the template change

### DIFF
--- a/eng/pipelines/runtimelab-official.yml
+++ b/eng/pipelines/runtimelab-official.yml
@@ -37,6 +37,7 @@ extends:
           jobTemplate: /eng/pipelines/common/global-build-job.yml
           buildConfig: Release
           platforms:
+#          - linux_x64
           - windows_x64
           jobParameters:
             templatePath: 'templates-official'
--- a/eng/pipelines/runtimelab-official.yml
+++ b/eng/pipelines/runtimelab-official.yml
@@ -38,15 +38,12 @@ extends:
           buildConfig: Release
           platforms:
           - windows_x64
-          - OSX_x64
-          - Browser_wasm_win
-          - wasi_wasm_win
           jobParameters:
             templatePath: 'templates-official'
             timeoutInMinutes: 300
             isOfficialBuild: true
             testGroup: innerloop
-            buildArgs: -s clr.aot+libs+nativeaot.packages -c $(_BuildConfig)
+            buildArgs: -s clr+libs+hosts+packs -c $(_BuildConfig)
             postBuildSteps:
             - template: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
               parameters:
--- a/eng/pipelines/runtimelab-official.yml
+++ b/eng/pipelines/runtimelab-official.yml
@@ -37,14 +37,14 @@ extends:
           jobTemplate: /eng/pipelines/common/global-build-job.yml
           buildConfig: Release
           platforms:
-#          - linux_x64
-          - windows_x64
+          - Browser_wasm_win
+          - wasi_wasm_win
           jobParameters:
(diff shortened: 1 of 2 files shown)


eng/pipelines/runtimelab.yml
```diff
@@ -50,95 +50,64 @@ extends:
...
-          # Build the whole product with Release CoreCLR
-          #
-          - template: /eng/pipelines/common/platform-matrix.yml
-            parameters:
-              jobTemplate: /eng/pipelines/common/global-build-job.yml
-              helixQueuesTemplate: /eng/pipelines/libraries/helix-queues-setup.yml
-              buildConfig: release
-              platforms:
-#              - linux_x64
-              - windows_x64
```
jkotas, line 129: I am updating the official builds to this

eng/pipelines/runtimelab.yml
```diff
@@ -50,95 +50,64 @@ extends:
...
-              jobParameters:
-                isOfficialBuild: true
-                timeoutInMinutes: 380
-                buildArgs: -s clr+libs+hosts+packs -c $(_BuildConfig)
```
jkotas, line 133: And this

jkotas:
@dotnet/nativeaot-llvm

akoeplinger (approved):

yowl:
Thanks, I defer to others with better knowledge!

eng/pipelines/runtimelab-official.yml
```diff
@@ -37,16 +37,14 @@ extends:
...
-          - OSX_x64
-          - Browser_wasm_win
-          - wasi_wasm_win
           jobParameters:
             templatePath: 'templates-official'
             timeoutInMinutes: 300
             isOfficialBuild: true
             testGroup: innerloop
-            buildArgs: -s clr.aot+libs+nativeaot.packages -c $(_BuildConfig)
+            buildArgs: -s clr+libs+hosts+packs -c $(_BuildConfig)
```
SingleAccretion: I don't think this will work - and I don't think the merged version of `runtimelab.yml` `isOfficialBuild` branch was quite correct either. The `host` subset is gated by `NativeAotSupported` (`false` for WASM).

Here's how this looked before the merge:

https://github.com/dotnet/runtimelab/blob/143b29c91ff4b62a0a93af6af4457f6127c7f8a1/eng/pipelines/runtimelab.yml#L110-L128

It is meant to work as follows:
1) We produce the target (WASM) packages in the first part of the build. Note: arch-independent packages are also produced here, but there is a later condition somewhere that [picks](https://github.com/dotnet/runtimelab/blob/143b29c91ff4b62a0a93af6af4457f6127c7f8a1/eng/pipelines/official/stages/publish.yml#L2) one platform out of which it is published.
2) We produce the host (`runtime.win-x64`) packages in the post-build steps. We only do this for `browser_wasm_win` to avoid the conflict.

The problem in https://github.com/dotnet/runtimelab/issues/2539#issuecomment-2080367183 is that we switched `windows_arm64` to `windows_x64` and so one of the target packages collided with one of the host packages. That is something I failed to consider with https://github.com/dotnet/runtimelab/pull/2563#discussion_r1580151005.

The correct fix is to probably just remove all but `Browser_wasm_win` and `wasi_wasm_win` from the platform list here (and assume that whatever breakage of upstream we cause will be caught by PR validation).

jkotas: Thank you - this makes sense. I was puzzled by how this can ever work, but I did not consider that runtimelab.yml can have incorrectly resolved conflicts.

agocke (commented):
Retroactive LGTM

PR #2571, opened 2024-05-02
[NativeAOT-LLVM] Cleanup dotnet.js API integration with ExportsFile

maraf:
- `ExportsFile` is always set
- Use UTF8 without BOM when overriding exports file

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
@@ -5,14 +5,12 @@
     <IlcExportUnmanagedEntrypoints>true</IlcExportUnmanagedEntrypoints><!-- For JSExports -->
   </PropertyGroup>
   <ItemGroup>
-    <EmccNativeExportedFunction Include="free" />
-    <EmccNativeExportedFunction Include="htons" />
-    <EmccNativeExportedFunction Include="main" />
-    <EmccNativeExportedFunction Include="malloc" />
-    <EmccNativeExportedFunction Include="memalign" />
-    <EmccNativeExportedFunction Include="memset" />
-    <EmccNativeExportedFunction Include="ntohs" />
-    <EmccExportedFunction Include="_%(EmccNativeExportedFunction)" />
+    <EmccExportedFunction Include="_free" />
+    <EmccExportedFunction Include="_htons" />
+    <EmccExportedFunction Include="_malloc" />
+    <EmccExportedFunction Include="_memalign" />
+    <EmccExportedFunction Include="_memset" />
+    <EmccExportedFunction Include="_ntohs" />
     <EmccExportedFunction Include="stackAlloc" />
     <EmccExportedFunction Include="stackRestore" />
     <EmccExportedFunction Include="stackSave" />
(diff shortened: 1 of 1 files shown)


maraf:
@dotnet/nativeaot-llvm

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
```diff
@@ -5,12 +5,14 @@
...
-    <EmccExportedFunction Include="_memset" />
-    <EmccExportedFunction Include="_ntohs" />
+    <EmccNativeExportedFunction Include="free" />
+    <EmccNativeExportedFunction Include="htons" />
+    <EmccNativeExportedFunction Include="main" />
+    <EmccNativeExportedFunction Include="malloc" />
+    <EmccNativeExportedFunction Include="memalign" />
+    <EmccNativeExportedFunction Include="memset" />
+    <EmccNativeExportedFunction Include="ntohs" />
+    <EmccExportedFunction Include="_%(EmccNativeExportedFunction)" />
```
SingleAccretion: Here's a snag I have not considered: `EmccExportedFunction` is a public API documented as being directly fed into Emscripten, i. e. you're expected to mangle it beforehand by adding the underscore to C symbols, but _not_ symbols defined in JS libraries. So it's hard to feed that into ILC as-is: no matter what you do, JS symbols will come out mangled on the other side.

In light of this, I think we have to keep the exports file rewriting for now (see e. g. https://github.com/emscripten-core/emscripten/issues/8380 for what the future might look like).

We could add another option to ILC that says "don't mangle symbols", but it doesn't seem worth it at this point in time.

BTW, there is a bit of code in the file under `Condition="'$(ExportsFile)' == ''` that could be deleted because the exports file is now always created so all of the exports can be routed through it.

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
```diff
@@ -36,8 +38,15 @@
 
     <!-- Pseudo-namespace for the JS interop symbols. -->
     <DirectPInvoke Include="System.Runtime.InteropServices.JavaScript" />
+    <UnmanagedEntryPointsAssembly Include="System.Runtime.InteropServices.JavaScript" />
```
SingleAccretion: Unrelated change.
```suggestion
```

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
maraf: Aye aye

maraf:
@SingleAccretion Based on your comment, I have recycled the PR to fix the problem with UTF8-BOM

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
```diff
@@ -76,20 +76,19 @@
       <!-- Exported functions required by user or runtime API -->
       <_DotNetJsLinkerFlag Include="-s DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=$(_EmccExportedLibraryFunction)" Condition="'$(_EmccExportedLibraryFunction)' != ''" />
       <_DotNetJsLinkerFlag Include="-s EXPORTED_RUNTIME_METHODS=$(_EmccExportedRuntimeMethods)" />
-      <_DotNetJsLinkerFlag Include="-s EXPORTED_FUNCTIONS=$(_EmccExportedFunctions)" Condition="'$(ExportsFile)' == ''" /><!-- Multiple args in the same rsp file is not support, add functions to app exports -->
```
SingleAccretion, line 79: `_EmccExportedFunctions` is now dead code.

SingleAccretion (approved):
LGTM modulo one comment and assuming green CI. Thank you!

PR #2574, opened 2024-05-05
[NativeAOT-LLVM] Build on Linux

### LINKED ISSUE #1713: [NativeAOT-LLVM] Building for Win/Linux fails
While the branch is aimed to building LLVM/Wasm, it ought to be possible to still build for win-x64 and linux-x64 but attempting to do so fails with
```
E:\tmp\aot\pkg\microsoft.dotnet.ilcompiler.llvm\7.0.0-dev\build\Microsoft.NETCore.Native.Publish.targets(75,5): error : The PrivateSdkAssemblies ItemGroup is required for _ComputeAssembliesToCompileToNative [E:\tmp\aot\aot.csproj]
```
This is a similar problem to https://github.com/dotnet/runtimelab/issues/260#issuecomment-728161671 , the names are
```
Message:  PackageDefinitions Microsoft.DotNet.ILCompiler.LLVM/7.0.0-dev PackageDefinitions.Name Microsoft.DotNet.ILCompiler.LLVM RuntimeIlcPackageName runtime.win-x64.Microsoft.DotNet.ILCompiler
```

and can be worked around by modifying `Microsoft.DotNet.ILCompiler.targets` to append `.LLVM` to the package name, for example, just handling both names

```
      <RuntimePackagePath Condition="'%(PackageDefinitions.Name)' == '$(RuntimeIlcPackageName)'">%(PackageDefinitions.ResolvedPath)</RuntimePackagePath>
      <RuntimePackagePath Condition="'%(PackageDefinitions.Name)' == '$(RuntimeIlcPackageName).LLVM'">%(PackageDefinitions.ResolvedPath)</RuntimePackagePath>
      <IlcHostPackagePath Condition="'%(PackageDefinitions.Name)' == '$(IlcHostPackageName)'">%(PackageDefinitions.ResolvedPath)</IlcHostPackagePath>
      <IlcHostPackagePath Condition="'%(PackageDefinitions.Name)' == '$(IlcHostPackageName).LLVM'">%(PackageDefinitions.ResolvedPath)</IlcHostPackagePath>
```

yowl:
This PR reenables the build of the product on Linux, including the LLVM JIT.  Does not build the Wasm runtime or libraries on Linux (to be done in #2569 )

Fixes #1713


src/mono/browser/runtime/package-lock.json
```diff
@@ -1,8 +1,2756 @@
 {
   "name": "@microsoft/dotnet-runtime",
```
jkotas: Are the changes in this file intentional - why do we need them?

yowl: No, reverted.

yowl:
cc @dotnet/nativeaot-llvm

eng/native/gen-buildsys.sh
```diff
@@ -95,6 +95,9 @@ fi
 
 if [[ "$host_arch" == "wasm" ]]; then
     if [[ "$target_os" == "browser" ]]; then
+        if [[ -n "$NATIVEAOT_CI_WASM_BUILD_EMSDK_PATH" ]]; then
+             source $NATIVEAOT_CI_WASM_BUILD_EMSDK_PATH/emsdk_env.sh
+        fi
```
SingleAccretion: The Windows equivalent is in `eng/common/build.ps1`. Can we do the same with `eng/common/build.sh` to keep the scripts similar?

eng/pipelines/common/global-build-job.yml
```diff
@@ -206,15 +206,25 @@ jobs:
...
 
+      # Install Wasm dependencies on Linux: emscripten, LLVM, NodeJS.
+    - ${{ if and(eq(parameters.hostedOs, 'linux'), and(eq(parameters.runtimeFlavor, 'coreclr'), eq(parameters.archType, 'wasm'))) }}:
+      - script: $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-emscripten.sh $(Build.SourcesDirectory)/wasm-tools
+        displayName: Install/activate emscripten
+      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 -Configs ${{ parameters.buildConfig }} -CI
+        workingDirectory: $(Build.SourcesDirectory)/wasm-tools
+        displayName: Install/build LLVM
+      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-nodejs.ps1 $(Build.SourcesDirectory)/wasm-tools
+        displayName: Install NodeJS
```
SingleAccretion: Would it work to unify the both Windows and Linux on `pwsh`?

eng/pipelines/common/global-build-job.yml
```diff
@@ -206,15 +206,25 @@ jobs:
...
       - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.cmd $(Build.SourcesDirectory)\wasm-tools $(Build.SourcesDirectory) ${{ parameters.buildConfig }}
         displayName: Install/build LLVM
       - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-nodejs.cmd $(Build.SourcesDirectory)\wasm-tools
         displayName: Install NodeJS
 
+      # Install Wasm dependencies on Linux: emscripten, LLVM, NodeJS.
+    - ${{ if and(eq(parameters.hostedOs, 'linux'), and(eq(parameters.runtimeFlavor, 'coreclr'), eq(parameters.archType, 'wasm'))) }}:
+      - script: $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-emscripten.sh $(Build.SourcesDirectory)/wasm-tools
+        displayName: Install/activate emscripten
+      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 -Configs ${{ parameters.buildConfig }} -CI
```
SingleAccretion: `parameters.buildConfig` can be `Checked`, to the normalization of of `Config` in `install-llvm.cmd` needs to be moved into `install-llvm.ps1`.

eng/pipelines/runtimelab.yml
```diff
@@ -59,7 +59,7 @@ extends:
             helixQueuesTemplate: /eng/pipelines/coreclr/templates/helix-queues-setup.yml
             buildConfig: debug
             platforms:
-#            - linux_x64
+            - linux_x64
```
SingleAccretion: These `linux_x64` jobs only test upstream code - I think we can do with just one to exercise our changes to the build scripts (say, only Release).

eng/pipelines/runtimelab/install-emscripten.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
SingleAccretion: Perhaps we can use the same powershell sharing strategy as with the the other scripts?

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -35,23 +41,45 @@ foreach ($Config in $Configs)
...
     $BuildDirName = "build-$($Config.ToLower())"
-    mkdir $BuildDirName -Force
+    if ($IsWindows)
+    {
+        mkdir $BuildDirName -Force
+    }
+    else
+    {
+        mkdir $BuildDirName --parents
+    }
```
SingleAccretion:
```suggestion
    New-Item $BuildDirName -Force
```
`mkdir` is a PS alias; not using should avoid the need for the condition.

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -35,23 +41,45 @@ foreach ($Config in $Configs)
...
+    if ($IsWindows)
+    {
+        $generator="Visual Studio 17 2022"
+    }
+    else
+    {
+        $generator="Unix Makefiles"
+    }
+
+    $CmakeConfigureCommandLine = "-G", "$generator", "-DLLVM_INCLUDE_BENCHMARKS=OFF"
```
SingleAccretion:
```suggestion
    if ($IsWindows)
    {
        $CmakeGenerator = "Visual Studio 17 2022"
    }
    else
    {
        $CmakeGenerator = "Unix Makefiles"
    }

    $CmakeConfigureCommandLine = "-G", "$CmakeGenerator", "-DLLVM_INCLUDE_BENCHMARKS=OFF"
```
Nit: pascal case variables.

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -6,6 +6,12 @@ param(
     [switch]$NoBuild
 )
 
+# Set IsWindows if the version of Powershell does not already have it.
+if (!(Test-Path variable:global:IsWindows)) 
+{
+    $IsWindows=[environment]::OSVersion.Platform -eq [PlatformID]::Win32NT
+}
```
SingleAccretion:
```suggestion
# Set IsWindows if the version of Powershell does not already have it.
if (!(Test-Path variable:global:IsWindows)) 
{
    $IsWindows = [Environment]::OSVersion.Platform -eq [PlatformID]::Win32NT
}
```

eng/pipelines/runtimelab/install-nodejs.ps1
```diff
@@ -1,7 +1,21 @@
 $InstallPath = $Args[0]
 $NodeJSVersion = "v20.2.0"
-$NodeJSInstallName = "node-$NodeJSVersion-win-x64"
-$NodeJSZipName = "$NodeJSInstallName.zip"
+
+if (!(Test-Path variable:global:IsWindows))
+{
+    $IsWindows=[environment]::OSVersion.Platform -eq [PlatformID]::Win32NT
```
SingleAccretion:
```suggestion
    $IsWindows = [Environment]::OSVersion.Platform -eq [PlatformID]::Win32NT
```

eng/pipelines/runtimelab/install-nodejs.ps1
```diff
@@ -33,9 +47,17 @@ if ($RetryCount -le 0)
     exit 1
 }
 
-Expand-Archive -LiteralPath "$InstallPath\$NodeJSInstallName.zip" -DestinationPath $InstallPath -Force
+if ($IsWindows)
+{
+    Expand-Archive -LiteralPath "$InstallPath\$NodeJSZipName" -DestinationPath $InstallPath -Force
+    $NodeJSExePath = "$InstallPath\$NodeJSInstallName\node.exe"
```
SingleAccretion, line 53: Does `Expand-Archive` not work on Linux (it's a bit surprising)?

src/coreclr/build-runtime.sh
```diff
@@ -171,6 +171,26 @@ if [[ -n "$__RequestedBuildComponents" ]]; then
     __CMakeTarget=" $__RequestedBuildComponents "
     __CMakeTarget="${__CMakeTarget// paltests / paltests_install }"
 fi
+
+if [[ "$__CMakeTarget" == *"wasmjit"* ]]; then
+  __ExtraCmakeArgs="$__ExtraCmakeArgs -DCLR_CMAKE_BUILD_LLVM_JIT=1"
```
SingleAccretion: Do we need `__ExtraCmakeArgs` in addition to `__CMakeArgs`?

src/coreclr/jit/CMakeLists.txt
```diff
@@ -91,7 +91,9 @@ function(create_standalone_jit)
     # We'll be linking against LLVM built without /guard:ehcont, so disable it.
     set_target_properties(${TARGETDETAILS_TARGET} PROPERTIES CLR_EH_CONTINUATION OFF)
 
-    find_package(LLVM REQUIRED CONFIG PATHS $ENV{LLVM_CMAKE_CONFIG})
+    # Prefer the version specified in LLVM_CMAKE_CONFIG over the "system" installed version
+    find_package(LLVM REQUIRED CONFIG PATHS $ENV{LLVM_CMAKE_CONFIG} NO_DEFAULT_PATH)
+    find_package(LLVM REQUIRED CONFIG)
```
SingleAccretion: Does this mean we will search for the system LLVM install too? That would not be correct.

src/native/libs/build-native.sh
```diff
@@ -60,15 +60,15 @@ source "$__RepoRootDir"/eng/native/build-commons.sh
 
 # Set cross build
 if [[ "$__TargetOS" == browser ]]; then
-    if [[ -z "$EMSDK_PATH" ]]; then
+    if [[ -z "$EMSDK" ]]; then
```
SingleAccretion: It would be better to just set `EMSDK_PATH` to `EMSDK` somewhere above s.t. these lines do not need to be modified.

src/tests/Common/tests.targets
```diff
@@ -6,7 +6,7 @@
   </ItemGroup>
 
   <PropertyGroup>
-      <TestAssemblyDir Condition="'$(TestAssemblyDir)' == ''">$(BaseOutputPathWithConfig)\tests\</TestAssemblyDir>
+      <TestAssemblyDir Condition="'$(TestAssemblyDir)' == ''">$(BaseOutputPathWithConfig)\Tests\</TestAssemblyDir>
```
SingleAccretion: Why does upstream not run into the same problem (in other words - is this an upstream bug)?

eng/native/gen-buildsys.sh
yowl: Yes, moved.

eng/pipelines/common/global-build-job.yml
yowl: yes, removed the `.cmd`

eng/pipelines/common/global-build-job.yml
yowl: Done

eng/pipelines/runtimelab.yml
yowl: RIght, have left just Release.

eng/common/build.sh
```diff
@@ -87,6 +87,8 @@ prepare_machine=false
 verbosity='minimal'
 runtime_source_feed=''
 runtime_source_feed_key=''
+host_arch=''
+target_os=''
```
SingleAccretion: Are these only required for `NATIVEAOT_CI_WASM_BUILD_EMSDK_PATH`? It's not clear what they're for.

eng/pipelines/runtimelab/install-emscripten.ps1
```diff
@@ -0,0 +1,23 @@
...
+git clone https://github.com/emscripten-core/emsdk.git
+
+Set-Location -Path emsdk
+
+# Checkout a specific commit to avoid unexpected issues
+git checkout 37b85e9
+
+python ./emsdk.py install 3.1.47
+
+./emsdk activate 3.1.47
```
SingleAccretion: We have `install-emscripten.cmd` referenced in `docs/using-nativeaot/prerequisites.md` and also `docs/workflow/building/coreclr/nativeaot.md`, these need to be updated now.

Also, it is odd that the first line calls `.py` directly while the second goes through the wrapper script.

eng/pipelines/runtimelab/install-emscripten.sh
yowl: right, have deletes the .cmd/.sh and just left the `ps1`

eng/pipelines/runtimelab/install-llvm.ps1
yowl: Thanks

eng/pipelines/runtimelab/install-llvm.ps1
yowl: Thanks, done with `-ItemType Directory`

eng/pipelines/runtimelab/install-nodejs.ps1
yowl: It does, but doesn't support `gz` which is gzip, or `xz` which is some LZMA I think.

yowl: And those are our options for linux https://nodejs.org/dist/v9.9.0/

src/coreclr/jit/CMakeLists.txt
yowl: It does, removed.

src/tests/Common/tests.targets
yowl: Have reverted this one, it looked wrong, but wasn't the problem I thought it was.

src/native/libs/build-native.sh
yowl: Done, thanks

yowl:
Lots of little commits, but think that's all the comments above addressed.

src/coreclr/build-runtime.sh
yowl: missed this one, but is in now.

src/coreclr/build-runtime.sh
```diff
@@ -185,7 +205,7 @@ fi
 
 eval "$__RepoRootDir/eng/native/version/copy_version_files.sh"
 
-build_native "$__HostOS" "$__HostArch" "$__ProjectRoot" "$__IntermediatesDir" "$__CMakeTarget" "$__CMakeArgs" "CoreCLR component"
+build_native "$__HostOS" "$__HostArch" "$__ProjectRoot" "$__IntermediatesDir" "$__CMakeTarget" "$__CMakeArgs $__ExtraCmakeArgs" "CoreCLR component"
```
SingleAccretion:
```suggestion
build_native "$__HostOS" "$__HostArch" "$__ProjectRoot" "$__IntermediatesDir" "$__CMakeTarget" "$__CMakeArgs" "CoreCLR component"
```
Now removed.

src/coreclr/jit/CMakeLists.txt
```diff
@@ -91,7 +91,8 @@ function(create_standalone_jit)
     # We'll be linking against LLVM built without /guard:ehcont, so disable it.
     set_target_properties(${TARGETDETAILS_TARGET} PROPERTIES CLR_EH_CONTINUATION OFF)
 
-    find_package(LLVM REQUIRED CONFIG PATHS $ENV{LLVM_CMAKE_CONFIG})
+    # Prefer the version specified in LLVM_CMAKE_CONFIG over the "system" installed version
+    find_package(LLVM REQUIRED CONFIG PATHS $ENV{LLVM_CMAKE_CONFIG} NO_DEFAULT_PATH)
```
SingleAccretion:
```suggestion
    # Only look at "our" LLVM with LLVM_CMAKE_CONFIG.
    find_package(LLVM REQUIRED CONFIG PATHS $ENV{LLVM_CMAKE_CONFIG} NO_DEFAULT_PATH)
```

eng/pipelines/runtimelab/install-emscripten.ps1
```diff
@@ -0,0 +1,23 @@
+param(
+    [string]$InstallDir
+)
+
+New-Item -ItemType Directory -Force -ErrorAction SilentlyContinue -Path (Split-Path -Path $InstallDir -Parent) -Name (Split-Path -Path $InstallDir -Leaf)
```
SingleAccretion: Does this really need to be this elaborate?

I just tried:
```
/home/singleaccretion> New-Item /home/singleaccretion/test/xyz/dir -Force -ItemType Directory
```
And it created the nested directory as expected.

eng/pipelines/common/global-build-job.yml
```diff
@@ -206,13 +206,23 @@ jobs:
...
+      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-emscripten.ps1 $(Build.SourcesDirectory)/wasm-tools
         displayName: Install/activate emscripten
       - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.cmd $(Build.SourcesDirectory)\wasm-tools $(Build.SourcesDirectory) ${{ parameters.buildConfig }}
         displayName: Install/build LLVM
-      - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-nodejs.cmd $(Build.SourcesDirectory)\wasm-tools
+      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-nodejs.ps1 $(Build.SourcesDirectory)\wasm-tools
+        displayName: Install NodeJS
+
+      # Install Wasm dependencies on Linux: emscripten, LLVM, NodeJS.
+    - ${{ if and(eq(parameters.hostedOs, 'linux'), and(eq(parameters.runtimeFlavor, 'coreclr'), eq(parameters.archType, 'wasm'))) }}:
```
SingleAccretion: Can we collapse the Windows/Linux into one section now (will require putting Emscripten on the same `pwsh` plan)?

eng/common/build.sh
yowl: Its to avoid 
```
/home/scott/github/runtimelab/eng/common/build.sh: line 245: host_arch: unbound variable
```
`${host_arch-''}` might be another way to solve it, or make it a mandatory parameter.

eng/pipelines/runtimelab/install-emscripten.ps1
```diff
@@ -0,0 +1,23 @@
+param(
+    [string]$InstallDir
+)
+
+New-Item -ItemType Directory -Force -ErrorAction SilentlyContinue -Name $InstallDir
+
+$ErrorActionPreference="Stop"
+
+Set-Location -Path $InstallDir
```
SingleAccretion:
```suggestion
$ErrorActionPreference="Stop"

New-Item -Path $InstallDir -ItemType Directory -Force
Set-Location -Path $InstallDir
```
I think this ought to work...

eng/pipelines/common/global-build-job.yml
```diff
@@ -206,13 +206,13 @@ jobs:
...
         displayName: Disk Usage before Build
 
-      # Install Wasm dependencies: emscripten, LLVM, NodeJS
+      # Install Wasm dependencies on Windows: emscripten, LLVM, NodeJS
     - ${{ if and(eq(parameters.runtimeFlavor, 'coreclr'), eq(parameters.archType, 'wasm')) }}:
-      - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-emscripten.cmd $(Build.SourcesDirectory)\wasm-tools
+      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-emscripten.ps1 $(Build.SourcesDirectory)/wasm-tools
         displayName: Install/activate emscripten
-      - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.cmd $(Build.SourcesDirectory)\wasm-tools $(Build.SourcesDirectory) ${{ parameters.buildConfig }}
+      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 $(Build.SourcesDirectory)\wasm-tools $(Build.SourcesDirectory) ${{ parameters.buildConfig }}
```
SingleAccretion:
```suggestion
      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 $(Build.SourcesDirectory)/wasm-tools $(Build.SourcesDirectory) -Configs ${{ parameters.buildConfig }} -CI
```
The cmake detection and `cd $(Build.SourcesDirectory)/wasm-tools` need to be migrated from `install-llvm.cmd` to `install-llvm.ps1`.

eng/common/build.sh
SingleAccretion: But the only uses of `host_arch` that I see look to be new (added in this change). My suggestion is delete them - or are they needed for something?

eng/pipelines/runtimelab/install-emscripten.ps1
yowl: Its ok on Linux, but with a drive letter it fails.  We could strip the drive letter and assume we are "in" that drive, but it probably isn't much easier to read.
```
PS C:\Users\scott> new-item -ItemType Directory -Force -Name c:/tmp/1/2
new-item : The given path's format is not supported.
At line:1 char:1
+ new-item -ItemType Directory -Force -Name c:/tmp/1/2
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [New-Item], NotSupportedException
    + FullyQualifiedErrorId : System.NotSupportedException,Microsoft.PowerShell.Commands.NewItemCommand

PS C:\Users\scott> new-item -ItemType Directory -Force -Name /tmp/1/2


    Directory: C:\Users\scott\tmp\1


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----        11/05/2024     16:54                2
```

eng/pipelines/common/global-build-job.yml
yowl: Yes, done

yowl:
"But the only uses of host_arch that I see look to be new (added in this change). My suggestion is delete them - or are they needed for something?"

You are right, I don't actually need it here, but I will for `./build.sh clr.aot -c Debug -a wasm -os browser`

eng/pipelines/runtimelab/install-emscripten.ps1
SingleAccretion: > `new-item -ItemType Directory -Force -Name c:/tmp/1/2`

I think this should use `-Path`. I tried:
```
PS C:\Users\Accretion\AppData\Local\Temp> new-item -Force -ItemKind Directory -Path C:\Users\Accretion\AppData\Local\Temp/xyz/rot/tot

    Directory: C:\Users\Accretion\AppData\Local\Temp\xyz\rot

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----          13.05.2024    00:22                tot
```

SingleAccretion:
My preference would be introduce them in the WASM change. In general, it looks unexpected to me that we need to change `eng/common/build.sh`, since these scripts are used upstream for `-a wasm -os browser` builds too.

eng/pipelines/common/global-build-job.yml
yowl: The cmake detection, i.e. executing `init-vs-env.cmd` as we do, doesn't seem necessary, and puts the `CMakePath` at the end anyway so doesn't precede anything already in the path.  It is possible to do from powershell, but would involve a wrapper `cmd` to `echo` the `CMakePath` back to the powershell do you scoping of environment variables and the lack or an equivalent `call` in powershell.  Question is, do we really need it?

SingleAccretion: I had assumed it was needed. If that's not the case - great!

yowl: We do need `cmake.exe` in the path, so have added this back, which means we need a `cmd` for the windows side.

eng/pipelines/runtimelab/install-emscripten.ps1
yowl: Yes, indeed that does.

yowl:
> My preference would be introduce them in the WASM change. In general, it looks unexpected to me that we need to change `eng/common/build.sh`, since these scripts are used upstream for `-a wasm -os browser` builds too.

OK, removed here.

yowl:
Was on a break last week, so did this with lots of trail and error, but have now rebased to remove that noise and is green.

eng/pipelines/common/global-build-job.yml
```diff
@@ -222,10 +224,18 @@ jobs:
...
+        - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/set-cmake-path.cmd
+          displayName: Set CMake path
+        - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 -CI -InstallDir $(Build.SourcesDirectory)/wasm-tools -Configs ${{ parameters.buildConfig }}
+          displayName: Install/build LLVM
+      - ${{ if eq(parameters.osGroup, 'linux') }}:
+        - script: $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-pwsh.sh $(Build.SourcesDirectory)/wasm-tools
+          displayName: Install Powershell 7
+        # Install LLVM for the win-x64 and linux-x64 build as it will build the clrjit for browser_wasm cross compilation
+        - script: $(Build.SourcesDirectory)/wasm-tools/powershell7/pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 -CI -InstallDir $(Build.SourcesDirectory)/wasm-tools -Configs ${{ parameters.buildConfig }}
+          displayName: Install/build LLVM
```
SingleAccretion: This will install tools for host configurations (`windows_x64`, `linux_x64`) that do not actually target WASM. I think the LLVM stuff should be deleted for `windows_x64` and `Install Powershell 7` folded into the above block (under `{{ if and(eq(parameters.runtimeFlavor, 'coreclr'), eq(parameters.archType, 'wasm')) }}:`) under `if ne(parameters.hostedOs, 'windows')` (this way it will, hopefully, also work for macOS if/when that comes).

eng/pipelines/common/global-build-job.yml
```diff
@@ -222,10 +224,18 @@ jobs:
...
+      - ${{ if eq(parameters.osGroup, 'windows') }}:
+        - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/set-cmake-path.cmd
+          displayName: Set CMake path
+        - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 -CI -InstallDir $(Build.SourcesDirectory)/wasm-tools -Configs ${{ parameters.buildConfig }}
+          displayName: Install/build LLVM
+      - ${{ if eq(parameters.osGroup, 'linux') }}:
+        - script: $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-pwsh.sh $(Build.SourcesDirectory)/wasm-tools
+          displayName: Install Powershell 7
+        # Install LLVM for the win-x64 and linux-x64 build as it will build the clrjit for browser_wasm cross compilation
+        - script: $(Build.SourcesDirectory)/wasm-tools/powershell7/pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 -CI -InstallDir $(Build.SourcesDirectory)/wasm-tools -Configs ${{ parameters.buildConfig }}
```
SingleAccretion: `install-pwsh.sh` sets `PATH`. Does a plain `pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.ps1 ...` not work?

yowl: I think that will work, yes.  Its not actually exercised with this PR but should be fine.

eng/pipelines/common/global-build-job.yml
yowl: Thanks, done.  The Linux build isn't pulling in these tools with this change but that's fine, they will be with #2569

eng/pipelines/common/global-build-job.yml
```diff
@@ -208,11 +208,17 @@ jobs:
 
       # Install Wasm dependencies: emscripten, LLVM, NodeJS
     - ${{ if and(eq(parameters.runtimeFlavor, 'coreclr'), eq(parameters.archType, 'wasm')) }}:
-      - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-emscripten.cmd $(Build.SourcesDirectory)\wasm-tools
+      - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-emscripten.ps1 $(Build.SourcesDirectory)/wasm-tools
         displayName: Install/activate emscripten
-      - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/install-llvm.cmd $(Build.SourcesDirectory)\wasm-tools $(Build.SourcesDirectory) ${{ parameters.buildConfig }}
+      - script: call $(Build.SourcesDirectory)/eng/pipelines/runtimelab/set-cmake-path.cmd
```
SingleAccretion, line 213: I suspect this will need to be put under `if eq(parameters.hostedOs, 'windows')`, but let's leave that for the next change.

SingleAccretion (approved):
LGTM, thank you!

jkotas (approved):
Thank you!

sardonic1:
What's the state of affairs here?  I'd like to be able to automate wasi-wasm compilations using Linux (docker on Mac).  I see that @yowl has done a lot of work here, but I'm unsure how to leverage it if it is indeed completed.  For example, I can't see any artifacts published like runtime.linux-x64.Microsoft.DotNet.ILCompiler.LLVM.

SingleAccretion:
@sardonic1 there is still one more task left here, which is to enable the official build, which produces the packages.

sardonic1:
Thanks for that.  Is there a rough idea of when this feature might be available?  Any idea if any of this experimental branch will make it to a .NET 9 release?

SingleAccretion:
> Thanks for that. Is there a rough idea of when this feature might be available?

I think we can make the official build work pretty soon ('few days' soon). It is not a large change.

> Any idea if any of this experimental branch will make it to a .NET 9 release?

The compiler and runtime - no. Note: .NET 9 is pretty much "done" at this point - upstream `main` will soon switch to .NET 10. A certain amount of recent WASI-related libraries work will be in .NET 9.

PR #2589, opened 2024-05-20
Enable CodeQL unconditionally

agocke:


jkotas (changes requested):
runtime-main is automatically synchronized fork for dotnet/runtime:main. It should not be committed to directly. Doing so will break the synchronization.

Could you please send this change to dotnet/runtime:main instead?

PR #2591, opened 2024-05-22
[NativeAOT-LLVM] Merge may24

yowl:
This PR merges to 301604e35ccf41f53e01893e08cf7f33d5256c1a

- Some `#ifdef BACKGROUND_GC` in `gc.cpp` upstreamable.
- Adds a `WasmClassifier`
- Remove use of `REG_R0/REG_F0` to `REG_STK` to avoid asserts arising from the classifier.
- `#if` out the `assert(dscStackOffset == expected.GetStackOffset());` assert as "stack" offsets not aligned.  If aligned we hit another assert as per the comment.
- Reorder `REG_STK` and `REG_LLVM` to avoid an `assert`
- A workaround for mono's use of `emsdk_env.cmd` where we have `emsdk_env.bat` , didn't look to see why mono has a different extension.
- comment out macos debug build: upstream issue.


yowl:
cc @dotnet/nativeaot-llvm - should go green from here.

src/coreclr/gc/gc.cpp
```diff
@@ -18431,6 +18431,7 @@ bool gc_heap::should_retry_other_heap (int gen_number, size_t size)
     }
 }
 
+#ifdef BACKGROUND_GC
```
SingleAccretion, line 18434: This should go upstream (probably the declarations should be ifdefed out too).

src/coreclr/jit/abi.h
```diff
@@ -274,6 +293,8 @@ typedef Arm32Classifier PlatformClassifier;
 typedef RiscV64Classifier PlatformClassifier;
 #elif defined(TARGET_LOONGARCH64)
 typedef LoongArch64Classifier PlatformClassifier;
+#elif defined(TARGET_WASM32)
```
SingleAccretion:
```suggestion
#elif defined(TARGET_WASM)
```
I think we'll have pretty much the same code for WASM64 too.

src/coreclr/jit/jitconfigvalues.h
```diff
@@ -810,22 +810,18 @@ CONFIG_INTEGER(JitSaveFpLrWithCalleeSavedRegisters, W("JitSaveFpLrWithCalleeSave
...
-CONFIG_INTEGER(JitCheckLlvmIR, W("JitCheckLlvmIR"), 0)
-CONFIG_INTEGER(JitRunLssaTests, W("JitRunLssaTests"), 0)
-CONFIG_INTEGER(JitGcStress, W("JitGcStress"), 0)
+RELEASE_CONFIG_INTEGER(JitCheckLlvmIR, W("JitCheckLlvmIR"), 0)
+RELEASE_CONFIG_INTEGER(JitRunLssaTests, W("JitRunLssaTests"), 0)
+RELEASE_CONFIG_INTEGER(JitGcStress, W("JitGcStress"), 0)
 
 #ifdef DEBUG
 CONFIG_STRING(JitEnableLssaRange, W("JitEnableLssaRange"))
 #endif // DEBUG
```
SingleAccretion:
```suggestion
CONFIG_STRING(JitEnableLssaRange, W("JitEnableLssaRange"))
```
Now redundant.

src/coreclr/jit/llvm.cpp
```diff
@@ -396,7 +396,7 @@ bool Llvm::helperCallMayPhysicallyThrow(CorInfoHelpFunc helperFunc) const
         { FUNC(CORINFO_HELP_ASSIGN_BYREF) }, // Not used on WASM.
 
         // Not used in NativeAOT (or at all in some cases).
-        { FUNC(CORINFO_HELP_ASSIGN_STRUCT) },
+        { FUNC(CORINFO_HELP_BULK_WRITEBARRIER) },
```
SingleAccretion: This one should go to the above group now with `// Not used on WASM.`. We could start using it too.

src/coreclr/jit/register.h
```diff
@@ -149,13 +149,11 @@ REGDEF(STK,    8+KBASE,    0x0000,       "STK"  )
...
 //
-REGDEF(R0,                          0, 0x01, "R0")
-REGDEF(F0,                          1, 0x02, "F0")
-REGDEF(LLVM,                        2, 0x04, "LLVM")
+REGDEF(STK,                         2, 0x04, "SS")
 REGDEF(STK_CANDIDATE_UNCONDITIONAL, 3, 0x08, "SS_UNCONDITIONAL")
 REGDEF(STK_CANDIDATE_TENTATIVE,     4, 0x10, "SS_TENTATIVE")
 REGDEF(STK_CANDIDATE_COMMITED,      5, 0x20, "SS_COMMITED")
-REGDEF(STK,                         6, 0x40, "SS")
+REGDEF(LLVM,                        6, 0x40, "LLVM")
```
SingleAccretion: `REG_STK` should be the last register:
```
; target.h

// Each register list in register.h must declare REG_STK as the last value.
```

src/mono/browser/browser.proj
```diff
@@ -37,6 +37,8 @@
     <_EmccCompileRspPath>$(NativeBinDir)src\emcc-compile.rsp</_EmccCompileRspPath>
     <_EmccLinkRspPath>$(NativeBinDir)src\emcc-link.rsp</_EmccLinkRspPath>
     <EmSdkLLVMAr>$(EMSDK_PATH)\bin\llvm-ar</EmSdkLLVMAr>
+    <!-- NativeAOT-LLMVM does not use Emscripten tools built from source. -->
```
SingleAccretion:
```suggestion
    <!-- NativeAOT-LLVM does not use the dotnet/emsdk EMSDK, which differs in layout from EMSDK proper. -->
```
I do wish a bit it didn't differ...

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -35,12 +35,12 @@ public static void addCodeReloc(IntPtr thisHandle, void* handle)
...
-        private static byte[] AppendNullByte(byte[] inputArray)
+        private static byte[] AppendNullByte(ReadOnlySpan<byte> inputArray)
         {
-            byte[] nullTerminated = new byte[inputArray.Length + 1];
-            inputArray.CopyTo(nullTerminated, 0);
+            Span<byte> nullTerminated = new Span<byte>(new byte[inputArray.Length + 1]);
+            inputArray.CopyTo(nullTerminated);
             nullTerminated[inputArray.Length] = 0;
-            return nullTerminated;
+            return nullTerminated.ToArray();
```
SingleAccretion: This will allocate two arrays, while only one is needed.

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -63,7 +63,7 @@ private static byte[] AppendNullByte(byte[] inputArray)
             node.AppendMangledName(_this._compilation.NameMangler, sb);
 
             sb.Append("\0");
-            return (byte*)_this.GetPin(sb.UnderlyingArray);
+            return (byte*)_this.GetPin(sb.AsSpan().ToArray());
```
SingleAccretion: This will double the allocations on this path, which is not great. It would be better to re-expose `UnderlyingArray` instead.

src/coreclr/jit/lclvars.cpp
```diff
@@ -1889,7 +1889,9 @@ void Compiler::lvaClassifyParameterABI()
 // first stack arg is not considered to be at offset 0.
 // TODO-Cleanup: Unify things so that x86 is consistent with other platforms
 // here and change fgMorphExpandStackArgForVarArgs to account for that.
-#ifndef TARGET_X86
+// LLVM: staock offsets are not aligned in the classifier, if they are then this passes and
+// assert(segment.Offset + segment.Size <= lvaLclExactSize(lclNum)) below fails.
```
SingleAccretion, line 1893: Why does the second assert fail?

src/coreclr/jit/importercalls.cpp
```diff
@@ -7545,6 +7545,20 @@ bool Compiler::IsTargetIntrinsic(NamedIntrinsic intrinsicName)
...
+    switch(intrinsicName)
+    {
+        case NI_System_Math_MultiplyAddEstimate:
+        case NI_System_Math_ReciprocalEstimate:
+        case NI_System_Math_ReciprocalSqrtEstimate:
+            return false;
+        default:
+            break;
+    }
+#endif //TARGET_WASM
```
SingleAccretion: I have done some time thinking about this, and looking at https://github.com/dotnet/runtime/pull/102098, and I think the better fix for this issue is to augment `IsTargetIntrinsic` to return `true` for these intrinsics.

`return nullptr` from `impEstimateIntrinsic` requires a corresponding `ifdef` in CoreLib, see https://github.com/dotnet/runtime/pull/102098/files#r1597779222. We have the `sqrt` intrinsic, so `ReciprocalSqrtEstimate` is expandable inline for WASM, and others have trivial expansions.

The actual CQ item here is to enable the better CQ for these intrinsic by actually exploiting the non-determinism, but it's a bit hard because LLVM doesn't have the same notion of intrinsics deterministic on "process granularity".

src/coreclr/jit/targetwasm.h
```diff
@@ -44,12 +44,12 @@
   #define FEATURE_EH_CALLFINALLY_THUNKS 1  // Generate call-to-finally code in "thunks" in the enclosing EH region, protected by "cloned finally" clauses.
   #define CSE_CONSTS               1       // Enable if we want to CSE constants
 
-  #define RBM_ALLFLOAT             RBM_F0
+  #define RBM_ALLFLOAT             REG_STK
```
SingleAccretion: If we are deleting fake registers, it would be better to replace them with `REG_NA` instead of `REG_STK`.

`REG_STK` is the half-way point between real and fake registers itself :).

src/coreclr/jit/lclvars.cpp
yowl: Good question, I can change the classifier to 
```c++
    if (type == TYP_STRUCT)
    {
        structPassingKind wbPassStruct;
        type = comp->m_llvm->GetArgTypeForStructWasm(structLayout->GetClassHandle(), &wbPassStruct);
    }

    assert(type != TYP_STRUCT);

    unsigned typeSize = genTypeSize(type);

    ABIPassingSegment segment = ABIPassingSegment::OnStack(m_stackArgSize, 0, typeSize);
    m_stackArgSize += AlignUp(typeSize, TARGET_POINTER_SIZE);

    return ABIPassingInformation::FromSegment(comp, segment);
```
But that doesn't match the values passed to `SetStackOffset`
For example, when the first arg is not a "user arg", e.g. `lvaInitThisPtr`, then we dont set the stack offset, as its assumed to be `allocRegArg(TYP_INT)` I suppose.  Maybe this is where I've gone wrong in attempting to remove the use of the registers and I should look for another solution?

SingleAccretion: There is an assumption in the compiler that `this` is always passed in a register. The new classifier can accommodate this assumption like so:
```
if (wellKnownParam == WellKnownArg::ThisPointer)
{
    // Produce a register segment (by bringing back R0).
}
```
We can delete this later - when the old classification is phased out.

yowl: Thanks, `this` is solved with that.  For `TYP_STRUCT` and implicit by refs I dont see how these 2 uses of the size are working:

Here the size is used as is, ignoring byref structs, and the value used to set the stack offset.

https://github.com/dotnet/runtime/blob/3eb94d1a816f941dd66f01e62c1af25c108014bc/src/coreclr/jit/lclvars.cpp#L696

However here

https://github.com/dotnet/runtime/blob/main/src/coreclr/jit/lclvars.cpp#L1904

the segment is checked against the size adusted for storing byrefs as pointers.  How do we handle this to both pass this assert and the one that checks the classifer offsets with the stack offsets ?

SingleAccretion: We would need to modify `eeGetArgSize` to harmonize these. But at this point I agree that commenting out the assert is simpler.

src/coreclr/jit/targetwasm.h
yowl: Ok, reverted basically.  I added 
```c
  #define RBM_R0                   REG_NA
  #define RBM_F0                   REG_NA
  #define REG_F0                   REG_NA
```
Maybe I should just collapse the uses of `RBM/REG_F0` ?

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.cs
```diff
@@ -74,7 +75,7 @@ private static byte[] AppendNullByte(byte[] inputArray)
             _this.GetMangledFilterFuncletName(sb, index);
 
             sb.Append("\0");
-            return (byte*)_this.GetPin(sb.UnderlyingArray);
+            return (byte*)_this.GetPin(sb.AsSpan().ToArray());
```
SingleAccretion:
```suggestion
            return (byte*)_this.GetPin(sb.UnderlyingArray);
```

src/coreclr/jit/compiler.h
```diff
@@ -8280,8 +8280,8 @@ class Compiler
             reg     = REG_R11;
             regMask = RBM_R11;
 #elif defined(TARGET_WASM)
-            reg     = REG_R0;
-            regMask = RBM_R0;
+            reg     = REG_STK;
+            regMask = RBM_STK;
```
```suggestion
            reg     = REG_NA;
            regMask = REG_NA;
```
If that runs into asserts, `R0` will need to be physically brought back.

src/coreclr/jit/lclvars.cpp
```diff
@@ -1710,6 +1710,12 @@ void Compiler::lvaClassifyParameterABI(Classifier& classifier)
         {
             wellKnownArg = WellKnownArg::RetBuffer;
         }
+#ifdef TARGET_WASM
+        else if (i == info.compThisArg)
+        {
+            wellKnownArg = WellKnownArg::ThisPointer;
+        }
+#endif //TARGET_WASM
```
SingleAccretion:
```suggestion
```
No longer needed now that we've decided on silencing out the assert.

src/coreclr/jit/targetwasm.cpp
```diff
@@ -8,14 +8,61 @@
...
 const Target::ArgOrder Target::g_tgtUnmanagedArgOrder = ARG_ORDER_R2L;
 
-const regNumber intArgRegs [] = { REG_R0 };
-const regMaskTP intArgMasks[] = { RBM_R0 };
-const regNumber fltArgRegs [] = { REG_F0 };
-const regMaskTP fltArgMasks[] = { RBM_F0 };
 // clang-format on
+
+const regNumber intArgRegs[] = { REG_STK };
+const regNumber fltArgRegs[] = { REG_STK };
```
SingleAccretion, line 21:
```suggestion
const regNumber intArgRegs[] = { REG_NA };
const regNumber fltArgRegs[] = { REG_NA };
```

src/coreclr/jit/targetwasm.h
```diff
@@ -44,7 +44,11 @@
   #define FEATURE_EH_CALLFINALLY_THUNKS 1  // Generate call-to-finally code in "thunks" in the enclosing EH region, protected by "cloned finally" clauses.
   #define CSE_CONSTS               1       // Enable if we want to CSE constants
 
-  #define RBM_ALLFLOAT             RBM_F0
+  #define RBM_R0                   REG_NA
+  #define RBM_F0                   REG_NA
+  #define REG_F0                   REG_NA
```
SingleAccretion: `RBM_<...>` is a register mask, i. e. `1 << REG_<...>`. So the right "nothing" value for it is zero, i. e. `RBM_NONE`.
```suggestion
  #define RBM_R0                   RBM_NONE
  #define RBM_F0                   RBM_NONE
  #define REG_F0                   REG_NA
```

src/coreclr/jit/targetwasm.h
```diff
@@ -44,7 +44,11 @@
...
+
+  #define RBM_ALLFLOAT             REG_F0
```
SingleAccretion:
```suggestion
  #define RBM_ALLFLOAT             RBM_F0
```

src/coreclr/jit/targetwasm.h
```diff
@@ -72,26 +76,29 @@
...
   #define CNT_CALLEE_TRASH_FLOAT   4
+  #define CNT_CALL_GC_REGS         (CNT_CALLEE_SAVED) // This is how targetx86 is defined.
 
-  #define REG_CALLEE_SAVED_ORDER   REG_R0
+  #define REG_CALLEE_SAVED_ORDER   RBM_R0
   #define RBM_CALLEE_SAVED_ORDER   RBM_R0
 
   // GenericPInvokeCalliHelper VASigCookie Parameter
-  #define REG_PINVOKE_COOKIE_PARAM REG_R0
+  #define REG_PINVOKE_COOKIE_PARAM RBM_R0
```
SingleAccretion: There are a number of cases where `REG_` values have been changed to be defined as `RBM_R0`, which are masks. They should be changed back to register values (`REG_NA` if that works, brought back `R0/F0` otherwise).

src/coreclr/jit/importercalls.cpp
yowl: Changed `Math` to 
```cs
        public static double ReciprocalSqrtEstimate(double d)
        {
#if MONO || TARGET_RISCV64 || TARGET_LOONGARCH64 || TARGET_WASM
            return 1.0 / Sqrt(d);
#else
            return ReciprocalSqrtEstimate(d);
#endif
        }
```

src/coreclr/jit/lclvars.cpp
yowl: Ok, I left the `this` code in but can take it out as you say.

src/coreclr/jit/importercalls.cpp
SingleAccretion: Why was it needed? It doesn't look expected to me.

src/coreclr/jit/lclvars.cpp
yowl: Right, wasn't sure if it was "correct".  Now removed.

src/coreclr/jit/targetwasm.h
yowl: Reinstated `R0`

src/coreclr/jit/compiler.h
yowl: Have brought it back, just didn't seem to be a great idea, at least not in this PR.

src/coreclr/jit/importercalls.cpp
yowl: Its not needed, but I thought that' what you refered to in "requires a corresponding ...".  I will revert it

yowl:
I'm assuming the git hub failure is an infrastructure problem and this is actually green.

SingleAccretion (approved):
LGTM, thank you!

jkotas:
> I'm assuming the git hub failure is an infrastructure problem and this is actually green.

Looking into it

jkotas:
Thank you!

PR #2592, opened 2024-05-23
tweaks to enable NativeAOT-LLVM on Linux with WASI-SDK 22

dicej:
Some of these changes are borrowed from https://github.com/dotnet/runtimelab/pull/2569.

Note that I had to manually copy pthread.h from the wasi-sdk-22/share/wasi-sysroot/include/wasm32-wasi-threads directory to the wasi-sdk-22/share/wasi-sysroot/include/wasm32-wasi directory as a workaround until https://github.com/WebAssembly/wasi-libc/issues/501 is addressed.

### DIFF
--- a/eng/native/gen-buildsys.sh
+++ b/eng/native/gen-buildsys.sh
@@ -93,11 +93,17 @@ if [[ "$scan_build" == "ON" && -n "$SCAN_BUILD_COMMAND" ]]; then
     cmake_command="$SCAN_BUILD_COMMAND $cmake_command"
 fi
 
+cmake_extra_defines_wasm=()
 if [[ "$host_arch" == "wasm" ]]; then
     if [[ "$target_os" == "browser" ]]; then
         cmake_command="emcmake $cmake_command"
     elif [[ "$target_os" == "wasi" ]]; then
-        true
+        if [[ -z $WASI_SDK_PATH ]]; then
+            echo "Error: Should set WASI_SDK_PATH environment variable pointing to WASI SDK root."
+            exit 1
+        fi
+
+        cmake_extra_defines_wasm=("-DCLR_CMAKE_TARGET_OS=wasi" "-DCLR_CMAKE_TARGET_ARCH=wasm" "-DWASI_SDK_PREFIX=$WASI_SDK_PATH" "-DCMAKE_TOOLCHAIN_FILE=$WASI_SDK_PATH/share/cmake/wasi-sdk.cmake" "-DCMAKE_CROSSCOMPILING_EMULATOR=node --experimental-wasm-bigint --experimental-wasi-unstable-preview1")
     else
         echo "target_os was not specified"
         exit 1
@@ -110,6 +116,7 @@ $cmake_command \
   "-DCMAKE_BUILD_TYPE=$buildtype" \
   "-DCMAKE_INSTALL_PREFIX=$__CMakeBinDir" \
   $cmake_extra_defines \
+  "${cmake_extra_defines_wasm[@]}" \
   $__UnprocessedCMakeArgs \
   -S "$1" \
   -B "$2"
--- a/eng/pipelines/runtimelab/install-wasi-sdk.ps1
+++ b/eng/pipelines/runtimelab/install-wasi-sdk.ps1
@@ -1,6 +1,13 @@
-Invoke-WebRequest -Uri https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-20/wasi-sdk-20.0.m-mingw.tar.gz -OutFile wasi-sdk-20.0.m-mingw.tar.gz
+Invoke-WebRequest -Uri https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-22/wasi-sdk-22.0.m-mingw64.tar.gz -OutFile wasi-sdk-22.0.m-mingw64.tar.gz
 
-tar -xzf wasi-sdk-20.0.m-mingw.tar.gz
+tar -xzf wasi-sdk-22.0.m-mingw64.tar.gz
 
-mv wasi-sdk-20.0+m wasi-sdk
+mv wasi-sdk-22.0+m wasi-sdk
 
+# Temporary WASI-SDK 22 workaround: Until
+# https://github.com/WebAssembly/wasi-libc/issues/501 is addressed, we copy
+# pthread.h from the wasm32-wasi-threads include directory to the wasm32-wasi
+# include directory.  See https://github.com/dotnet/runtimelab/issues/2598 for
+# the issue to remove this workaround once WASI-SDK 23 is released.
+
+cp wasi-sdk/share/wasi-sysroot/include/wasm32-wasi-threads/pthread.h wasi-sdk/share/wasi-sysroot/include/wasm32-wasi/
--- a/src/coreclr/build-runtime.sh
+++ b/src/coreclr/build-runtime.sh
@@ -177,11 +177,11 @@ if [[ "$__CMakeTarget" == *"wasmjit"* ]]; then
 
   if [[ "$__BuildType" == "Release" ]]; then
     if [[ -n $LLVM_CMAKE_CONFIG_RELEASE ]]; then
-      LLVM_CMAKE_CONFIG="$LLVM_CMAKE_CONFIG_RELEASE"
+      export LLVM_CMAKE_CONFIG="$LLVM_CMAKE_CONFIG_RELEASE"
     fi
   else
     if [[ -n $LLVM_CMAKE_CONFIG_DEBUG ]]; then
-      LLVM_CMAKE_CONFIG="$LLVM_CMAKE_CONFIG_DEBUG"
+      export LLVM_CMAKE_CONFIG="$LLVM_CMAKE_CONFIG_DEBUG"
     fi
   fi
 
--- a/src/coreclr/jit/CMakeLists.txt
+++ b/src/coreclr/jit/CMakeLists.txt
@@ -706,16 +706,11 @@ else()
   create_standalone_jit(TARGET clrjit_win_x86_${ARCH_HOST_NAME} OS win ARCH x86 DESTINATIONS .)
 endif (CLR_CMAKE_TARGET_ARCH_RISCV64)
 
-# Note that we currently do not support building the LLVM Jit on Unix.
-# Note as well that we need this "CLR_CMAKE_BUILD_LLVM_JIT", defined by the build scripts,
+# Note that we need this "CLR_CMAKE_BUILD_LLVM_JIT", defined by the build scripts,
 # because there is no way in CMake to ask "what --target's' am I being asked to configure?".
 if (CLR_CMAKE_BUILD_LLVM_JIT)
     # The LLVM clrjit needs to be the last clrjit to use create_standalone_jit as it modifies some cmake variables.
     # LLVM clrjit has an extra export - registerLlvmCallbacks.
-    set(CLRJIT_EXPORTS ${CMAKE_CURRENT_LIST_DIR}/ClrJit.Llvm.exports)
-    set(JIT_EXPORTS_FILE ${CMAKE_CURRENT_BINARY_DIR}/ClrJit.Llvm.exports.def)
-    preprocess_file (${CLRJIT_EXPORTS} ${JIT_EXPORTS_FILE})
-    set(JIT_DEF_FILE ${JIT_EXPORTS_FILE})
 
     # Exclude cpp files that are not required when not processing beyond rationalized LIR.
     # Use REGEX as this list contains the absolute paths.
--- a/src/coreclr/jit/llvmcodegen.cpp
+++ b/src/coreclr/jit/llvmcodegen.cpp
@@ -1118,6 +1118,7 @@ void Llvm::visitNode(GenTree* node)
             break;
         case GT_JMP:
             NYI("LLVM/GT_JMP"); // Requires support for explicit tailcalls.
+            break;
         default:
             unreached();
     }
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -388,9 +388,11 @@ The .NET Foundation licenses this file to you under the MIT license.
 
   <Target Name="CheckWasmSdks">
     <Error Text="Emscripten not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Emscripten and ensure the EMSDK environment variable points to the directory containing upstream/emscripten/emcc.bat"
-           Condition="'$(EMSDK)' == ''" />
+           Condition="'$(EMSDK)' == '' and '$(_targetOS)' == 'browser'" />
     <Error Text="Wasi SDK not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Wasi SDK and ensure the WASI_SDK_PATH environment variable points to the directory containing share/wasi-sysroot"
            Condition="'$(WASI_SDK_PATH)' == '' and '$(_targetOS)' == 'wasi'" />
+    <Warning Text="The WASI SDK version is too low. Please use WASI SDK 22 or newer with a 64 bit Clang."
+             Condition="!Exists('$(WASI_SDK_PATH)/VERSION')" />
   </Target>
 
   <Target Name="CompileWasmObjects"
@@ -420,8 +422,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs>$(CompileWasmArgs) -fvisibility=default -mllvm -combiner-global-alias-analysis=false -mllvm -disable-lsr --sysroot=&quot;$(WASI_SDK_PATH)/share/wasi-sysroot&quot; -target $(IlcLlvmTarget)</CompileWasmArgs>
 
       <ExeExt Condition="'$(OS)' == 'Windows_NT'">.exe</ExeExt>
-      <!-- using Emscripten's clang++ because of a crash in wasi-sdk's clang++ (https://github.com/WebAssembly/wasi-sdk/issues/326) -->
-      <WasmCompilerPath>&quot;$(EMSDK)/upstream/bin/clang++$(ExeExt)&quot;</WasmCompilerPath>
+      <WasmCompilerPath>&quot;$(WASI_SDK_PATH)/bin/clang++$(ExeExt)&quot;</WasmCompilerPath>
       <WasmLinkerPath>&quot;$(WASI_SDK_PATH)/bin/clang&quot;</WasmLinkerPath>
     </PropertyGroup>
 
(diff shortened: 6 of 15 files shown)


dicej:
@dotnet-policy-service agree company="Fermyon Technologies"

jkotas:
cc @dotnet/nativeaot-llvm

eng/native/gen-buildsys.sh
```diff
@@ -59,7 +59,7 @@ for i in "${@:6}"; do
     esac
 done
 
-cmake_extra_defines=
+cmake_extra_defines=()
```
SingleAccretion: This is an upstream (dotnet/runtime) file. Why are these changes needed?

(Formulated another way: why did it work before and doesn't work now?)

src/coreclr/build-runtime.sh
```diff
@@ -189,6 +189,8 @@ if [[ "$__CMakeTarget" == *"wasmjit"* ]]; then
     echo The LLVM_CMAKE_CONFIG environment variable pointing to llvm-build-dir/lib/cmake/llvm must be set. 1>&2
     exit 1
   fi
+
+  export LLVM_CMAKE_CONFIG
```
SingleAccretion: It would be better to add `export` to the above `LLVM_CMAKE_CONFIG="$LLVM_CMAKE_CONFIG_[RELEASE|DEBUG]"` to make the script look similar to its `.cmd` version.

src/coreclr/jit/CMakeLists.txt
```diff
@@ -558,15 +558,24 @@ endif()
...
 if(CLR_CMAKE_TARGET_WIN32)
-  set(CLRJIT_EXPORTS ${CMAKE_CURRENT_LIST_DIR}/ClrJit.exports)
-  set(JIT_EXPORTS_FILE ${CMAKE_CURRENT_BINARY_DIR}/ClrJit.exports.def)
+  if(CLR_CMAKE_BUILD_LLVM_JIT)
+    set(CLRJIT_EXPORTS ${CMAKE_CURRENT_LIST_DIR}/ClrJit.Llvm.exports)
+    set(JIT_EXPORTS_FILE ${CMAKE_CURRENT_BINARY_DIR}/ClrJit.Llvm.exports.def)
+  else()
+    set(CLRJIT_EXPORTS ${CMAKE_CURRENT_LIST_DIR}/ClrJit.exports)
+    set(JIT_EXPORTS_FILE ${CMAKE_CURRENT_BINARY_DIR}/ClrJit.exports.def)
+  endif()
```
SingleAccretion: This will modify the exports for all Jits (and is cause of the current CI failure).

There are two ways to fix it:
1) Continue with the current approach of a separate `ClrJit.Llvm.exports` that is "set last".
2) Add `registerLlvmCallbacks` to the general exports file, deleting `ClrJit.Llvm.exports`, and add a stub no-op implementation of it under `#ifndef TARGET_WASM` somewhere (`ee_il_dll.cpp`).

My preference is the second approach.

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -1118,6 +1118,7 @@ void Llvm::visitNode(GenTree* node)
             break;
         case GT_JMP:
             NYI("LLVM/GT_JMP"); // Requires support for explicit tailcalls.
+            LLVM_FALLTHROUGH;
```
```suggestion
            break;
```
(Let's see if this angers MSVC's code analyzer... It can be `FALLTHROUGH;` if it does.)

We don't use LLVM macros in Jit source code.

src/native/libs/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ if (CLR_CMAKE_TARGET_UNIX OR CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI)
     endif ()
 
     set(CMAKE_INSTALL_PREFIX $ENV{__CMakeBinDir})
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -Wno-declaration-after-statement")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -Wno-declaration-after-statement -Wno-switch-default")
```
SingleAccretion:
```suggestion
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -Wno-declaration-after-statement")

    // TODO-LLVM: remove once upstream moves to WASI SDK 22 as well (or otherwise fixes these warnings).
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-switch-default")
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -420,8 +420,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs>$(CompileWasmArgs) -fvisibility=default -mllvm -combiner-global-alias-analysis=false -mllvm -disable-lsr --sysroot=&quot;$(WASI_SDK_PATH)/share/wasi-sysroot&quot; -target $(IlcLlvmTarget)</CompileWasmArgs>
 
       <ExeExt Condition="'$(OS)' == 'Windows_NT'">.exe</ExeExt>
-      <!-- using Emscripten's clang++ because of a crash in wasi-sdk's clang++ (https://github.com/WebAssembly/wasi-sdk/issues/326) -->
-      <WasmCompilerPath>&quot;$(EMSDK)/upstream/bin/clang++$(ExeExt)&quot;</WasmCompilerPath>
+      <WasmCompilerPath>&quot;$(WASI_SDK_PATH)/bin/clang++$(ExeExt)&quot;</WasmCompilerPath>
```
SingleAccretion, line 425: This will need corresponding CI changes in `eng\pipelines\runtimelab\install-wasi-sdk.ps1` so that we test with the new SDK.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -388,7 +388,7 @@ The .NET Foundation licenses this file to you under the MIT license.
 
   <Target Name="CheckWasmSdks">
     <Error Text="Emscripten not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Emscripten and ensure the EMSDK environment variable points to the directory containing upstream/emscripten/emcc.bat"
-           Condition="'$(EMSDK)' == ''" />
+           Condition="'$(EMSDK)' == '' and '$(_targetOS)' == 'browser'" />
     <Error Text="Wasi SDK not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Wasi SDK and ensure the WASI_SDK_PATH environment variable points to the directory containing share/wasi-sysroot"
            Condition="'$(WASI_SDK_PATH)' == '' and '$(_targetOS)' == 'wasi'" />
   </Target>
```
SingleAccretion: It would be good to add a warning for older SDKs that we know will likely not work.

SingleAccretion (commented):
The pthread situation will need to be resolved, s.t. an unmodified SDK can be used, by pulling the stubs from `PalRedhawkWasm.cpp` to `PalRedhawkWasm.h`.

dicej:
Thanks for the detailed feedback, @SingleAccretion.  I'm planning to address it when I get a chance early next week.

eng/native/gen-buildsys.sh
dicej: The key additions are on line 100, which is a translation of the WASI-specific part of gen-buildsys.cmd.  Also, incrementally accumulating arguments like we do in `cmake_extra_defines` requires some extra care in `bash` when some of those arguments have space characters in them.  Previously, we didn't need to worry about that, but the `-DCMAKE_CROSSCOMPILING_EMULATOR=...` argument _does_ have spaces, so in order to make that work I changed `cmake_extra_defines` from a string into an array.

eng/native/gen-buildsys.sh
```diff
@@ -97,7 +97,7 @@ if [[ "$host_arch" == "wasm" ]]; then
     if [[ "$target_os" == "browser" ]]; then
         cmake_command="emcmake $cmake_command"
     elif [[ "$target_os" == "wasi" ]]; then
-        true
+        cmake_extra_defines+=("-DCLR_CMAKE_TARGET_OS=wasi" "-DCLR_CMAKE_TARGET_ARCH=wasm" "-DWASI_SDK_PREFIX=$WASI_SDK_PATH" "-DCMAKE_TOOLCHAIN_FILE=$WASI_SDK_PATH/share/cmake/wasi-sdk.cmake" "-DCMAKE_CROSSCOMPILING_EMULATOR=node --experimental-wasm-bigint --experimental-wasi-unstable-preview1")
```
yowl:
```suggestion
        if [[ -z $WASI_SDK_PATH ]]; then
            echo "Error: Should set WASI_SDK_PATH environment variable pointing to WASI SDK root."
            exit 1
        fi
        cmake_extra_defines+=("-DCLR_CMAKE_TARGET_OS=wasi" "-DCLR_CMAKE_TARGET_ARCH=wasm" "-DWASI_SDK_PREFIX=$WASI_SDK_PATH" "-DCMAKE_TOOLCHAIN_FILE=$WASI_SDK_PATH/share/cmake/wasi-sdk.cmake" "-DCMAKE_CROSSCOMPILING_EMULATOR=node --experimental-wasm-bigint --experimental-wasi-unstable-preview1")
```

Consistent with the windows version.

dicej:
> The pthread situation will need to be resolved, s.t. an unmodified SDK can be used, by pulling the stubs from `PalRedhawkWasm.cpp` to `PalRedhawkWasm.h`.

I tried pursuing that, but it got ugly fast.  I had to edit code under e.g. src/coreclr/gc and sprinkle new `#ifdef`s in a lot of places, which felt like overkill for a temporary workaround.

Then I took a step back and decided to add the following to `install-wasi-sdk.ps1`:

```
# Temporary WASI-SDK 22 workaround: Until
# https://github.com/WebAssembly/wasi-libc/issues/501 is addressed, we copy
# pthread.h from the wasm32-wasi-threads include directory to the wasm32-wasi
# include directory:

cp wasi-sdk/share/wasi-sysroot/include/wasm32-wasi-threads/pthread.h wasi-sdk/share/wasi-sysroot/include/wasm32-wasi/
```

If that's not acceptable, I'd vote we stick with WASI-SDK 21 for the time being, then upgrade to WASI-SDK 23, which should include a fix for https://github.com/WebAssembly/wasi-libc/issues/501.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
dicej: Can you elaborate on how we could check this?  Ideally we'd extract the version number from the first line of `${WASI_SDK_PATH}/VERSION` and check it against a "known-to-be-compatible" list, but I'm new to the MSBuild project file DSL, so not sure how do to that here.

src/coreclr/jit/CMakeLists.txt
```diff
@@ -565,7 +565,6 @@ if(CLR_CMAKE_TARGET_WIN32)
   set(JIT_DEF_FILE ${JIT_EXPORTS_FILE})
 else()
   set(CLRJIT_EXPORTS ${CMAKE_CURRENT_LIST_DIR}/ClrJit.PAL.exports)
-
```
SingleAccretion: Whitespace diff - should be reverted.

src/coreclr/jit/CMakeLists.txt
```diff
@@ -706,16 +705,15 @@ else()
...
 if (CLR_CMAKE_BUILD_LLVM_JIT)
     # The LLVM clrjit needs to be the last clrjit to use create_standalone_jit as it modifies some cmake variables.
     # LLVM clrjit has an extra export - registerLlvmCallbacks.
-    set(CLRJIT_EXPORTS ${CMAKE_CURRENT_LIST_DIR}/ClrJit.Llvm.exports)
-    set(JIT_EXPORTS_FILE ${CMAKE_CURRENT_BINARY_DIR}/ClrJit.Llvm.exports.def)
-    preprocess_file (${CLRJIT_EXPORTS} ${JIT_EXPORTS_FILE})
-    set(JIT_DEF_FILE ${JIT_EXPORTS_FILE})
+    if (CLR_CMAKE_TARGET_LINUX)
+      add_compile_options(-Wno-switch)
+      add_compile_options(-Wno-format)
```
SingleAccretion: These warnings should be either disabled in source, addressed, or an issue filed for the removal of these suppressions.

eng/pipelines/runtimelab/install-wasi-sdk.ps1
```diff
@@ -1,6 +1,12 @@
...
 
-mv wasi-sdk-20.0+m wasi-sdk
+mv wasi-sdk-22.0+m wasi-sdk
 
+# Temporary WASI-SDK 22 workaround: Until
+# https://github.com/WebAssembly/wasi-libc/issues/501 is addressed, we copy
+# pthread.h from the wasm32-wasi-threads include directory to the wasm32-wasi
+# include directory:
+
+cp wasi-sdk/share/wasi-sysroot/include/wasm32-wasi-threads/pthread.h wasi-sdk/share/wasi-sysroot/include/wasm32-wasi/
```
SingleAccretion, line 13: It is ok for this hack to stay here to enable the runtime build, but:
1) There should be an issue filed tracking removing it.
2) It should be added as a step in `docs/workflow/building/coreclr/nativeaot.md`, under `download and install the Wasi SDK from...`.

dicej:
FYI, this should address the missing pthread.h issue, once merged and included in a release: https://github.com/WebAssembly/wasi-libc/pull/504

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: Is `${WASI_SDK_PATH}/VERSION` a new thing - I don't see it in the v20 layout?

eng/native/gen-buildsys.sh
SingleAccretion: I looked at this a bit closer, and it looks like this script is only used in upstream build by `build-native.sh`, which does the same thing manually.

https://github.com/dotnet/runtime/blob/64b93517eb4bea5de51aa5bd7e5b32076198f1f9/src/native/libs/build-native.sh#L68-L79

Do we need `-DCMAKE_CROSSCOMPILING_EMULATOR=node ...`? I suspect we don't.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
dicej: Yeah, looks like it's new: https://github.com/WebAssembly/wasi-sdk/commit/0409839729fb4348764905a7e8c2bb53000087a3.  I guess the fallback option would be to run `${WASI_SDK_PATH}/bin/clang --version` and infer the WASI-SDK version from the LLVM version, assuming the latter maps uniquely to the former.

eng/native/gen-buildsys.sh
dicej: I wasn't able to get the build to work without `-DCMAKE_CROSSCOMPILING_EMULATOR=node ...`.  I believe it's used as part of CMake platform feature detection (i.e. the little test programs that CMake builds and runs to verify that the platform supports various features).

eng/pipelines/runtimelab/install-wasi-sdk.ps1
dicej: I've updated nativeaot.md and opened https://github.com/dotnet/runtimelab/issues/2598

dicej:
Hm, I see `cl.exe` is not happy with the `#pragma clang ...` directives I added to suppress warnings in llvmlssa.cpp.  Is there a better way to approach that?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: Since it looks like the version file is only present in SDK 22, we can just key off of it existing (for now).

I. e. something like:
```
<Warning Condition="!Exists('$(WASI_SDK_PATH)/VERSION')" Text="The WASI SDK version is too low. Please use WASI SDK 22 or newer with a 64 bit Clang." />
```
(I added the 64 bit breadcrumb since it looks like the 22 releases have both 32 bit and 64 bit versions.)

eng/native/gen-buildsys.sh
SingleAccretion: Interesting, in my local experiment, omitting this from the Windows build still succeeded. Let's leave investigating that for another day.

(Edit: after a bit more digging, I see where this is coming from.)

Changing the type of `cmake_extra_defines` has a big diff. It would be better to add a new local (`cmake_extra_defines2`...) that has the argument with spaces in it and add that to the final cmake command line.

SingleAccretion:
>  Is there a better way to approach that?

It would be ideal if there was a way to propagate the warning upwards, but it looks like there isn't... Presumably, the directives can be guarded with `#ifdef __clang__`.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -388,9 +388,11 @@ The .NET Foundation licenses this file to you under the MIT license.
 
   <Target Name="CheckWasmSdks">
     <Error Text="Emscripten not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Emscripten and ensure the EMSDK environment variable points to the directory containing upstream/emscripten/emcc.bat"
-           Condition="'$(EMSDK)' == ''" />
+           Condition="'$(EMSDK)' == '' and '$(_targetOS)' == 'browser'" />
     <Error Text="Wasi SDK not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Wasi SDK and ensure the WASI_SDK_PATH environment variable points to the directory containing share/wasi-sysroot"
            Condition="'$(WASI_SDK_PATH)' == '' and '$(_targetOS)' == 'wasi'" />
+    <Warning Text="The WASI SDK version is too low. Please use WASI SDK 22 or newer with a 64 bit Clang."
+             Condition="!Exists('$(WASI_SDK_PATH)/VERSION')" />
```
```suggestion
           Condition="'$(WASI_SDK_PATH)' != '' and '$(_targetOS)' == 'wasi' and !Exists('$(WASI_SDK_PATH)/VERSION')" />
```

SingleAccretion (approved):
LGTM modulo one comment and assuming green CI. Thank you!

jkotas:
Could you please resolve the conflict?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl:
```suggestion
             Condition="!Exists('$(WASI_SDK_PATH)/VERSION') and '$(_targetOS)' == 'wasi'" />
```
You probably worked this one out already.

dicej:
I've resolved the conflict and everything's green.  Are we good to merge this?

jkotas:
Thanks!

PR #2600, opened 2024-05-30
Update official build to generate and upload NuGet packages to the feed

kotlarmilos:
## Description

This PR updates the `prepare-signed-artifacts.yml` file to use Arcade's templates and modifies the publish script to generate and upload NuGet packages to the feed.

Official build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2463272&view=results

### DIFF
--- a/eng/pipelines/common/templates/build-job.yml
+++ b/eng/pipelines/common/templates/build-job.yml
@@ -45,7 +45,6 @@ jobs:
 
       steps:
       - script: $(_buildScript)
-                -ci
                 -pack
                 -c $(_BuildConfig)
                 $(_testBuildArg)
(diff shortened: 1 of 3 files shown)


eng/pipelines/common/templates/build-job.yml
```diff
@@ -45,7 +45,6 @@ jobs:
 
       steps:
       - script: $(_buildScript)
-                -ci
```
akoeplinger: why?

kotlarmilos: The `-ci` flag adds a suffix to the version in the maestro build promotion pipeline. Without it, the suffix is `-dev`, which might be more appropriate for standalone experiments.

akoeplinger: -dev is for local scenarios, we shouldn't be using that in CI

kotlarmilos: Ok, reverting this change.

akoeplinger (approved):

PR #2595, opened 2024-05-26
[NativeAOT-LLVM] Update llvm to 18.1.3 and emscripten to 3.1.54

yowl:
Updates the version of LLVM and emscripten.

depends #2591 

Moving emscripten later than 3.1.54 causes C/C++ version standard problems, i.e. need to go from gnu99 to c11:

E.g.
```
C:/github/runtimelab/src/native/libs/System.Native/pal_networking.c(171,1): error GB0A439F3: '_Static_assert' is incompatible with C standards before C11 [-Wer
ror,-Wpre-c11-compat] [C:\github\runtimelab\src\native\libs\build-native.proj]
```

PR #2602, opened 2024-06-05
[NativeAOT-LLVM] Support using emscripten from Mono workload

maraf:
Soon version of emscripten that uses NativeAOT-LLVM and mono will align. These changes allows to use emscripten from mono workload. For the moment it unblocks smoke tests on runtime repo.

- Allow to override `WasmCompilerPath` for browser
- Pass `EmscriptenEnvVars` when running the emscripten compiler

Preview of upstream integration bits
```xml
<Target Name="SetupEmscriptenEnvVars" BeforeTargets="CompileSingleWasmObject;LinkNativeLlvm">
  <!-- Environment variables required for running emsdk commands like `emcc` -->
  <ItemGroup Condition="'$(EMSDK)' == '$(EmscriptenUpstreamEmscriptenPath)'">
    <EmscriptenEnvVars Include="EMSDK_PYTHON=$(EmscriptenPythonToolsPath)python.exe" Condition="'$(OS)' == 'Windows_NT'" />
    <EmscriptenEnvVars Include="PYTHONPATH=$(EmscriptenPythonToolsPath)" Condition="'$(OS)' == 'Windows_NT'" />
    <EmscriptenEnvVars Include="PYTHONHOME=" Condition="'$(OS)' == 'Windows_NT'" />
    <EmscriptenEnvVars Include="DOTNET_EMSCRIPTEN_LLVM_ROOT=$(EmscriptenSdkToolsPath)bin" />
    <EmscriptenEnvVars Include="DOTNET_EMSCRIPTEN_BINARYEN_ROOT=$(EmscriptenSdkToolsPath)" />
    <EmscriptenEnvVars Include="DOTNET_EMSCRIPTEN_NODE_JS=$([MSBuild]::NormalizePath($(EmscriptenNodeToolsPath), 'bin', 'node$(_ExeExt)'))" />
    <EmscriptenEnvVars Include="EM_CACHE=$(WasmCachePath)" Condition="'$(WasmCachePath)' != ''" />
    <EmscriptenEnvVars Include="EM_FROZEN_CACHE=True" Condition="'$(WasmCachePath)' == '$(EmscriptenCacheSdkCacheDir)'" />
  </ItemGroup>
</Target>
```

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -387,12 +387,12 @@ The .NET Foundation licenses this file to you under the MIT license.
   </Target>
 
   <Target Name="CheckWasmSdks">
+    <!-- The EMSDK is typically set by user and points to a standard emscripten layout -->
+    <!-- The EmscriptenUpstreamEmscriptenPath comes from upstream mono workload and points to an upstream folder of emscripten. The EmscriptenUpstreamBinPath points to bin folder -->
     <Error Text="Emscripten not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Emscripten and ensure the EMSDK environment variable points to the directory containing upstream/emscripten/emcc.bat"
-           Condition="'$(EMSDK)' == '' and '$(_targetOS)' == 'browser'" />
+           Condition="'$(EMSDK)' == '' and '$(EmscriptenUpstreamEmscriptenPath)' == '' and '$(_targetOS)' == 'browser'" />
     <Error Text="Wasi SDK not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Wasi SDK and ensure the WASI_SDK_PATH environment variable points to the directory containing share/wasi-sysroot"
            Condition="'$(WASI_SDK_PATH)' == '' and '$(_targetOS)' == 'wasi'" />
-    <Warning Text="The WASI SDK version is too low. Please use WASI SDK 22 or newer with a 64 bit Clang."
-             Condition="!Exists('$(WASI_SDK_PATH)/VERSION') and '$(_targetOS)' == 'wasi'" />
   </Target>
 
   <Target Name="CompileWasmObjects"
@@ -414,7 +414,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'">$(CompileWasmArgs) -s DISABLE_EXCEPTION_CATCHING=0</CompileWasmArgs>
 
       <ScriptExt Condition="'$(OS)' == 'Windows_NT'">.bat</ScriptExt>
-      <WasmCompilerPath Condition="'$(WasmCompilerPath)' == ''">&quot;$(EMSDK)/upstream/emscripten/emcc$(ScriptExt)&quot;</WasmCompilerPath>
+      <WasmCompilerPath Condition="'$(EMSDK)' != ''">&quot;$(EMSDK)/upstream/emscripten/emcc$(ScriptExt)&quot;</WasmCompilerPath>
+      <WasmCompilerPath Condition="'$(EMSDK)' == ''">&quot;$(EmscriptenUpstreamEmscriptenPath)emcc$(ScriptExt)&quot;</WasmCompilerPath>
       <WasmLinkerPath>$(WasmCompilerPath)</WasmLinkerPath>
     </PropertyGroup>
 
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -393,6 +393,8 @@ The .NET Foundation licenses this file to you under the MIT license.
            Condition="'$(EMSDK)' == '' and '$(EmscriptenUpstreamEmscriptenPath)' == '' and '$(_targetOS)' == 'browser'" />
     <Error Text="Wasi SDK not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Wasi SDK and ensure the WASI_SDK_PATH environment variable points to the directory containing share/wasi-sysroot"
            Condition="'$(WASI_SDK_PATH)' == '' and '$(_targetOS)' == 'wasi'" />
+    <Warning Text="The WASI SDK version is too low. Please use WASI SDK 22 or newer with a 64 bit Clang."
+             Condition="!Exists('$(WASI_SDK_PATH)/VERSION') and '$(_targetOS)' == 'wasi'" />
   </Target>
 
   <Target Name="CompileWasmObjects"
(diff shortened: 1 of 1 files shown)


maraf:
cc @dotnet/nativeaot-llvm

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -439,7 +439,7 @@ The .NET Foundation licenses this file to you under the MIT license.
   </Target>
 
   <Target Name="CompileSingleWasmObject">
-    <Exec Command="$(ExecCommand)" />
+    <Exec Command="$(ExecCommand)" EnvironmentVariables="@(EmscriptenEnvVars)" />
```
SingleAccretion, line 445: What are these environment variables expected to affect?

This step is compiling bitcode files produces by ILC into object files. In the fullness of time, this step will be folded into ILC itself.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -414,7 +414,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'">$(CompileWasmArgs) -s DISABLE_EXCEPTION_CATCHING=0</CompileWasmArgs>
 
       <ScriptExt Condition="'$(OS)' == 'Windows_NT'">.bat</ScriptExt>
-      <WasmCompilerPath>&quot;$(EMSDK)/upstream/emscripten/emcc$(ScriptExt)&quot;</WasmCompilerPath>
+      <WasmCompilerPath Condition="'$(WasmCompilerPath)' == ''">&quot;$(EMSDK)/upstream/emscripten/emcc$(ScriptExt)&quot;</WasmCompilerPath>
```
SingleAccretion: Is this override needed due to the layout difference between dotnet/emsdk and upstream emsdk?

SingleAccretion:
> Soon version of emscripten that uses NativeAOT-LLVM and mono will align

Which version will move, considering https://github.com/dotnet/runtimelab/pull/2595? I believe the current upstream emsdk is a bit older than 3.1.54.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
maraf: Exactly

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
maraf: We have a different `.emscripten` file. It relies on env vars to find other tools. We also provide a path to immutable prebuilt cache.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: It is unfortunate that the layouts are different.

We will have to support both. It means creating an interface to the outside world that enables us to reason which layout is being used. There are two ways to do it:
1) Use a different environment variable (runtime build uses `EMSDK_PATH`) for the dotnet layout.
    - This would also need a precedence rule - which EMSDK wins if both are available.
2) Probe the layout, e. g. `UsingDotnetEmsdkLayout = Exists($(EMSDK)/upstream)`.

I see the upstream targets use `1`, and I prefer that approach too. I think the precedence should be that `EMSDK` wins, but with a warning (i. e. `if $(EMSDK) != '' && $(EMSDK_PATH) != '' warn("Both EMSDK and EMSDK_PATH set. Using EMSDK.")`.

maraf: At the moment, I have it exactly like that. `EMSDK` wins and if `EMSDK` is empty, a `EmscriptenWorkloadPath` is used and `WasmCompilerPath` changed.

maraf:
> Which version will move, considering https://github.com/dotnet/runtimelab/pull/2595? I believe the current upstream emsdk is a bit older than 3.1.54.

Upstream is moving to 3.1.56

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
maraf: I have added a preview of upstream changes to the PR description

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: What I am saying is that `WasmCompilerPath` should not be the interface to the SDK. `WasmCompilerPath` is an internal property (like `CompileWasmArgs`), just not named like that.

It's coincidental that we only use `emcc` from the SDK right now. We may need to use other tools (e. g. `llvm-ar` to create static libraries, or `clang` for #2603). The logic that would select which layout to use should not need to be aware of what tools the targets happen to want to use.

What I am proposing is that we make `EMSDK_PATH` that interface property.

maraf: Okay, makes sense

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -387,12 +387,12 @@ The .NET Foundation licenses this file to you under the MIT license.
...
+    <!-- The EmscriptenUpstreamEmscriptenPath comes from upstream mono workload and points to an upstream folder of emscripten. The EmscriptenUpstreamBinPath points to bin folder -->
     <Error Text="Emscripten not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Emscripten and ensure the EMSDK environment variable points to the directory containing upstream/emscripten/emcc.bat"
-           Condition="'$(EMSDK)' == '' and '$(_targetOS)' == 'browser'" />
+           Condition="'$(EMSDK)' == '' and '$(EmscriptenUpstreamEmscriptenPath)' == '' and '$(_targetOS)' == 'browser'" />
     <Error Text="Wasi SDK not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Wasi SDK and ensure the WASI_SDK_PATH environment variable points to the directory containing share/wasi-sysroot"
            Condition="'$(WASI_SDK_PATH)' == '' and '$(_targetOS)' == 'wasi'" />
-    <Warning Text="The WASI SDK version is too low. Please use WASI SDK 22 or newer with a 64 bit Clang."
-             Condition="!Exists('$(WASI_SDK_PATH)/VERSION') and '$(_targetOS)' == 'wasi'" />
```
SingleAccretion: Why delete the warning?

maraf: I'm sorry, merge damage. My bad

SingleAccretion (approved):
LGTM. Thank you!

PR #2604, opened 2024-06-05
[NativeAOT-LLVM] Move dotnet.js to aotsdk

maraf:


maraf:
cc @dotnet/nativeaot-llvm

eng/liveBuilds.targets
```diff
@@ -229,15 +227,15 @@
...
+        $(CoreCLRAotSdkDir)dotnet.runtime.js;
+        $(CoreCLRAotSdkDir)dotnet.runtime.js.map;
+        $(CoreCLRAotSdkDir)dotnet.d.ts;
+        $(CoreCLRAotSdkDir)package.json;"
         IsNative="true" />
       <LibrariesRuntimeFiles Condition="'$(TargetOS)' == 'browser' and '$(RuntimeFlavor)' == 'CoreCLR'"
-                             Include="$(CoreCLRNativeArtifactsPath)src\es6\*.js"
+                             Include="$(CoreCLRAotSdkDir)src\es6\*.js"
                              NativeSubDirectory="src\es6"
                              IsNative="true" />
```
SingleAccretion: This should not be needed anymore. Also, `Microsoft.DotNet.ILCompiler.LLVM.pkgproj` needs to be fixed to not copy these files into `framework/`.

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
```diff
@@ -71,7 +71,7 @@
       <JSFileType Include="lib.js"     Kind="js-library" />
       <JSFileType Include="post.js"    Kind="post-js" />
       <JSFileType Include="extpost.js" Kind="extern-post-js" />
-      <_WasmExtraJSFile Include="$(IlcFrameworkNativePath)src\es6\*.%(JSFileType.Identity)" Kind="%(JSFileType.Kind)" />
+      <_WasmExtraJSFile Include="$(IlcSdkPath)src\es6\*.%(JSFileType.Identity)" Kind="%(JSFileType.Kind)" />
```
SingleAccretion: I suspect this needs tweaks to `Microsoft.DotNet.ILCompiler.LLVM.pkgproj` to preserve the `src/es6`. Alternatively, it needs fixes here to work with the flat hierarchy (which would be simpler if it's possible).

SingleAccretion (commented):
Please also verify manually that the package still works and has the expected layout before merging.

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
maraf: Good catch! I forgot changes to pkgproj in stash

maraf:
image

SingleAccretion (approved):
Changes LGTM assuming the package works. Thank you!

PR #741, opened 2021-02-24
Disable objwriter dependency on terminfo

jkotas:
terminfo dependency is common source of issues (#699 is recent example). It is optional dependency only used by pretty printing code that we do not need.

PR #2605, opened 2024-06-06
[NativeAOT-LLVM] add support for building WASI and browser debug & release on Linux

yowl:
This PR adds the scripts and YML for building WASI and browser on Linux.

Takes some of the changes from Joel's work at #2614 to build on Linux.  I changed some cases of `Browser` to `browser` which seemed like the more consistent direction.  

Decided to pass `TestWrapperTargetsWindows` down from `runtimelab-post-build-steps.yml` to avoid troublesome inferencing.


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -22,7 +22,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <_SymbolPrefix Condition="'$(_IsApplePlatform)' == 'true'">_</_SymbolPrefix>
     <LinkerFlavor Condition="'$(LinkerFlavor)' == '' and '$(_targetOS)' == 'freebsd'">lld</LinkerFlavor>
     <LinkerFlavor Condition="'$(LinkerFlavor)' == '' and '$(_linuxLibcFlavor)' == 'bionic'">lld</LinkerFlavor>
-    <LinkerFlavor Condition="'$(LinkerFlavor)' == '' and '$(_targetOS)' == 'linux'">bfd</LinkerFlavor>
+    <LinkerFlavor Condition="'$(LinkerFlavor)' == '' and '$(_targetOS)' == 'linux'">lld</LinkerFlavor>
```
jkotas: This should propagate from the top-level LDFLAGS set by the init-compiler.sh script. Is it getting lost somewhere?

yowl: I assume from the fact that changing this line to lld makes no difference, that it is arriving as bfd.

yowl: Something up with args to init-compiler.sh maybe.

eng/pipelines/runtimelab/install-wasi-sdk.ps1
```diff
@@ -1,8 +1,27 @@
...
+    $WasiTar = "wasi-sdk-22.0.m-mingw64.tar.gz"
+    $WasiFolder = "wasi-sdk-22.0+m"
+}
+else
+{
+    $WasiTar = "wasi-sdk-22.0-linux.tar.gz"
+    $WasiFolder = "wasi-sdk-22.0"
+}
+
+Invoke-WebRequest -Uri https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-22/$WasiTar -OutFile $WasiTar
```
pavelsavara: please add `$ProgressPreference = 'SilentlyContinue'` otherwise `Invoke-WebRequest` is very slow.
See https://stackoverflow.com/questions/28682642/powershell-why-is-using-invoke-webrequest-much-slower-than-a-browser-download/43477248#43477248

eng/pipelines/runtimelab/install-wasmer.ps1
```diff
@@ -1,5 +1,33 @@
...
+if ($IsWIndows)
+{
+    $WasmerTar="wasmer-windows-amd64.tar.gz"
+}
+else
+{
+    $WasmerTar="wasmer-linux-amd64.tar.gz"
+}
+
+Invoke-WebRequest -Uri https://github.com/wasmerio/wasmer/releases/download/v3.3.0/$WasmerTar -OutFile $WasmerTar
```
pavelsavara: please add `$ProgressPreference = 'SilentlyContinue'`

eng/pipelines/runtimelab/install-wasi-sdk.ps1
yowl: Thanks, done.  I will squash all this when I've finished cleaning things up.

yowl:
cc @dotnet/nativeaot-llvm

eng/pipelines/common/platform-matrix.yml
```diff
@@ -479,6 +479,58 @@ jobs:
         buildConfig: ${{ parameters.buildConfig }}
         ${{ insert }}: ${{ parameters.jobParameters }}
 
+# Browser WebAssembly Linux X64
+
+- ${{ if containsValue(parameters.platforms, 'Browser_wasm_linux_x64') }}:
```
SingleAccretion: My preference would be to name these in a way that makes it clear that they're downstream-specific. E. g. `Browser_wasm_linux_x64_naot_llvm`; and comment we're adding these because the "stock" `wasi-wasm` doesn't have the build tools (that sanitizer image) we need.

eng/pipelines/common/platform-matrix.yml
```diff
@@ -479,6 +479,58 @@ jobs:
...
+      archType: wasm
+      targetRid: browser-wasm
+      platform: Browser_wasm_linux_x64
+      shouldContinueOnError: ${{ parameters.shouldContinueOnError }}
+      ${{ if eq(parameters.container, '') }}:
+        container: linux_x64
+      ${{ if ne(parameters.container, '') }}:
+        container:
+          image: ${{ parameters.container }}
+          registry: mcr
```
SingleAccretion: I think it would make sense to hardcode the `azurelinux-3.0-cross-amd64-net9.0-sanitizer` here instead of `runtimelab*` YMLs.

eng/pipelines/runtimelab.yml
```diff
@@ -111,3 +114,23 @@ extends:
                 - template: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
                   parameters:
                     librariesConfiguration: Release
+
+        #
+        # Build and test wasi linux with Release libraries and Release runtime with sanitzer container
+        #
```
SingleAccretion: Why not fold this into `Build and test with Release libraries and Release runtime` above?

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -90,8 +107,12 @@ foreach ($Config in $Configs | % { if ($_ -eq "Checked") { "Debug" } else { $_ }
         }
     }
     $CmakeConfigureCommandLine += "-DCMAKE_BUILD_TYPE=$LlvmConfig"
+    
+    if (!$IsWindows)
+    {
+        $CmakeConfigureCommandLine += "-DCMAKE_SYSROOT=/crossrootfs/x64", "-DCMAKE_SYSTEM_NAME=Linux", "-DCMAKE_INSTALL_PREFIX=/usr/local/llvm-cross", "-DLLVM_HOST_TRIPLE=x86_64-pc-linux-gnu", "-DLLVM_TARGETS_TO_BUILD=WebAssembly"
+    }
 
-    Write-Host "Invoking CMake configure: 'cmake $CmakeConfigureCommandLine'"
```
SingleAccretion, line 94: Logging deleted intentionally?

eng/testing/FindWasmHostExecutableAndRun.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+
+source $SCRIPT_DIR/FindWasmHostExecutable.sh "$1"
+
+if [ -n "${WASM_HOST_EXECUTABLE}" ]; then
+  shift
+  $WASM_HOST_EXECUTABLE "$WASM_BINARY_TO_EXECUTE" $WASM_HOST_ARGS_SEPERATOR "$@"
```
SingleAccretion, line 10: `.cmd` also logs the command it is about to run.

eng/testing/FindWasmHostExecutable.sh
```diff
@@ -0,0 +1,31 @@
...
+dirname=$(dirname "$1")
+
+node="node --stack_trace_limit=100"
+WASM_HOST_ARGS_SEPERATOR=""
+
+ls $dirname
+
+if [ -e "${dirname}/${exename}.js" ]; then
+  echo using node
+  $node --version
```
SingleAccretion:
```suggestion
```

eng/pipelines/runtimelab/install-wasi-sdk.ps1
```diff
@@ -1,8 +1,28 @@
...
+    $WasiTar = "wasi-sdk-22.0.m-mingw64.tar.gz"
+    $WasiFolder = "wasi-sdk-22.0+m"
+}
+else
+{
+    $WasiTar = "wasi-sdk-22.0-linux.tar.gz"
+    $WasiFolder = "wasi-sdk-22.0"
+}
+
+$ProgressPreference=-SilentlyContinue
```
SingleAccretion:
```suggestion
$ProgressPreference = 'SilentlyContinue'
```

eng/pipelines/runtimelab/install-wasi-sdk.ps1
```diff
@@ -1,8 +1,28 @@
...
+Invoke-WebRequest -Uri https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-22/$WasiTar -OutFile $WasiTar
+
+tar -xzf $WasiTar
+
```
SingleAccretion:
```suggestion
```

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -90,8 +107,12 @@ foreach ($Config in $Configs | % { if ($_ -eq "Checked") { "Debug" } else { $_ }
         }
     }
     $CmakeConfigureCommandLine += "-DCMAKE_BUILD_TYPE=$LlvmConfig"
+    
+    if (!$IsWindows)
+    {
+        $CmakeConfigureCommandLine += "-DCMAKE_SYSROOT=/crossrootfs/x64", "-DCMAKE_SYSTEM_NAME=Linux", "-DCMAKE_INSTALL_PREFIX=/usr/local/llvm-cross", "-DLLVM_HOST_TRIPLE=x86_64-pc-linux-gnu", "-DLLVM_TARGETS_TO_BUILD=WebAssembly"
```
SingleAccretion: Why is `DLLVM_HOST_TRIPLE`/`DCMAKE_SYSTEM_NAME` needed? Same question for `DLLVM_TARGETS_TO_BUILD` (we don't build target-specific libraries)?

src/coreclr/jit/CMakeLists.txt
```diff
@@ -92,6 +92,8 @@ function(create_standalone_jit)
     target_include_directories(${TARGETDETAILS_TARGET} PRIVATE "../tools/Common/JitInterface")
 
     # Only look at "our" LLVM with LLVM_CMAKE_CONFIG.
+    set(LLVM_DIR $ENV{LLVM_CMAKE_CONFIG})
+    set(CMAKE_CXX_STANDARD 17)
```
SingleAccretion: Does the previous approach not work? It's fine if it doesn't, but then it needs to be deleted and this line also converted to be target-based:
```
set_target_properties(${TARGETDETAILS_TARGET} PROPERTIES CXX_STANDARD 17)
```

src/coreclr/jit/compiler.cpp
```diff
@@ -1275,8 +1275,12 @@ void DisplayNowayAssertMap()
             fout = _wfopen(strJitMeasureNowayAssertFile, W("a"));
             if (fout == nullptr)
             {
+#if !defined(HOST_WINDOWS)
+                // TODO: how do we print a `const char16_t*` portably?
```
SingleAccretion, line 1266: Do these changes need for this to compile in CI/locally? There is an upstream change pending that will resolve them; it would be better to omit them here.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -418,6 +418,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'">$(CompileWasmArgs) -mllvm -enable-emscripten-cxx-exceptions</CompileWasmArgs>
 
       <ScriptExt Condition="'$(OS)' == 'Windows_NT'">.bat</ScriptExt>
+      <ScriptExt Condition="'$(OS)' != 'Windows_NT'"></ScriptExt>
```
SingleAccretion: Is `ScriptExt` set by something upstream? If so, this variable should be renamed to something less generic (e. g. `_IlcScriptExt`, or something like that).

eng/pipelines/runtimelab-official.yml
```diff
@@ -35,9 +35,12 @@ extends:
         parameters:
           jobTemplate: /eng/pipelines/common/global-build-job.yml
           buildConfig: Release
+          container: azurelinux-3.0-cross-amd64-net9.0-sanitizer
           platforms:
           - Browser_wasm_win
           - wasi_wasm_win
+          - Browser_wasm_linux_x64
+          - wasi_wasm_linux_x64
```
SingleAccretion: I suspect this will not work as-is.

We have two things being published:
1) Target packages: `browser-wasm` / `wasi-wasm`.
   - In the current setup, these are always published and so we'll get package conflicts, since both platforms will be producing the (hopefully) identical target packages.
2) Host packages.
   - In the current setup, only `Browser_wasm_win` publishes host packages.

To fix this up, we'll need to adjust the conditions that prevent package conflicts (e. g. only build target packages on Windows, only build host packages on `Browser` platforms). But I would suggest leaving the official build for a follow-up change.
```suggestion
          - Browser_wasm_win
          - wasi_wasm_win
```

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -18,7 +18,9 @@ steps:
   # Now we need to build the cross-targeting compilers, RyuJit and ILC. Likewise with packages.
 
   - ${{ if eq(parameters.archType, 'wasm') }}:
-    - script: $(Build.SourcesDirectory)/build$(scriptExt) clr.wasmjit+clr.aot -c $(buildConfigUpper) $(_officialBuildParameter) -ci
+    - script: $(Build.SourcesDirectory)/build$(scriptExt) clr.wasmjit+clr.aot -c $(buildConfigUpper) $(_officialBuildParameter) -ci -cross
+      env:
+        ROOTFS_DIR: /crossrootfs/x64
```
SingleAccretion: I think if you switch the container in the platform definition to `linux_x64_sanitizer`, this explicit `ROOTFS_DIR` can be dropped.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -30,12 +32,22 @@ steps:
...
+        displayName: Build WebAssembly tests
+    - ${{ elseif eq(parameters.platform, 'Browser_wasm_linux_x64') }}:
+      - script: |
+          source $(Build.SourcesDirectory)/wasm-tools/emsdk/emsdk_env.sh
+          LD_LIBRARY_PATH=/usr/local/lib:/opt/lib:/lib:/usr/lib:/crossrootfs/x64/lib/x86_64-linux-gnu/ $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) -arch:${{ parameters.archType }} -target-os:browser tree nativeaot /p:LibrariesConfiguration=${{ parameters.librariesConfiguration }}
+        displayName: Build WebAssembly tests
+    - ${{ elseif eq(parameters.platform, 'wasi_wasm_linux_x64') }}:
+      - script: |
+          source $(Build.SourcesDirectory)/wasm-tools/emsdk/emsdk_env.sh
+          LD_LIBRARY_PATH=/usr/local/lib:/opt/lib:/lib:/usr/lib:/crossrootfs/x64/lib/x86_64-linux-gnu/ $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) -arch:${{ parameters.archType }} -target-os:wasi tree nativeaot /p:LibrariesConfiguration=${{ parameters.librariesConfiguration }}
```
SingleAccretion: Why is a `LD_LIBRARY_PATH` like this needed?

eng/pipelines/runtimelab/install-llvm.ps1
yowl: I think we can drop `DLLVM_HOST_TRIPLE` and `DLLVM_TARGETS_TO_BUILD `.  I was following https://llvm.org/docs/HowToCrossCompileLLVM.html but this is not full cross compiling, so guess we don't need them all.  Attempting to drop `DCMAKE_SYSTEM_NAME ` fails:

```
../../../bin/llvm-min-tblgen: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
include/llvm/TargetParser/CMakeFiles/RISCVTargetParserTableGen.dir/build.make:118: recipe for target 'include/llvm/TargetParser/RISCVTargetParserDef.inc' failed
```

src/coreclr/jit/compiler.cpp
yowl: Yes, locally in docker, so presumably in CI docker as well.  Without these changes we cannot compile on the image:

```
  /runtime/src/coreclr/jit/compiler.cpp:1278:83: error: length modifier 'w' results in undefined behavior or no effect with 's' conversion specifier [-Werror,-Wformat]
   1278 |                 fprintf(jitstdout(), "Failed to open JitMeasureNowayAssertFile \"%ws\"\n",
        |                                                                                  ~^~
  /runtime/src/coreclr/jit/compiler.cpp:1279:25: error: format specifies type 'wchar_t *' but the argument has type 'LPCWSTR' (aka 'const char16_t *') [-Werror,-Wformat]
   1278 |                 fprintf(jitstdout(), "Failed to open JitMeasureNowayAssertFile \"%ws\"\n",
        |                                                                                  ~~~
        |                                                                                  %s
   1279 |                         strJitMeasureNowayAssertFile);
        |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
  2 errors generated.
  make[3]: *** [jit/CMakeFiles/clrjit_universal_wasm32_x64.dir/build.make:189: jit/CMakeFiles/clrjit_universal_wasm32_x64.dir/compiler.cpp.o] Error 1
```

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
yowl: Because `libtinfo.so.5` is only in the rootfs.  The host has `libtinfo.so.6`

jkotas: Do we build the LLVM with libtinfo dependency ourselves or is that for pre-built LLVM?

If we build it ourselves, we can get rid of libtinfo dependency like this: https://github.com/dotnet/runtimelab/pull/741/files#diff-8bb555c9ef0d91e66546d2fa904a310f4d22f129e191dff6b12e4fbdfe65e321R33

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -100,7 +122,6 @@ foreach ($Config in $Configs | % { if ($_ -eq "Checked") { "Debug" } else { $_ }
 
     if (!$NoBuild)
     {
-        Write-Host "Invoking CMake --build"
         cmake --build $BuildDirPath --config $LlvmConfig --target LLVMCore LLVMBitWriter
```
SingleAccretion:
```suggestion
        Write-Host "Invoking CMake --build"
        cmake --build $BuildDirPath --config $LlvmConfig --target LLVMCore LLVMBitWriter
```

eng/pipelines/runtimelab/install-llvm.ps1
```diff
@@ -114,10 +135,11 @@ foreach ($Config in $Configs | % { if ($_ -eq "Checked") { "Debug" } else { $_ }
         $LlvmCmakeConfigEnvVarName = if ($Config -eq "Release") {"LLVM_CMAKE_CONFIG_RELEASE"} else {"LLVM_CMAKE_CONFIG_DEBUG"}
     }
 
-    Write-Host "Setting $LlvmCmakeConfigEnvVarName to '$LlvmCmakeConfigPath'"
     if ($CI)
```
SingleAccretion, line 140:
```suggestion
    Write-Host "Setting $LlvmCmakeConfigEnvVarName to '$LlvmCmakeConfigPath'"
    if ($CI)
```

eng/testing/FindWasmHostExecutable.sh
```diff
@@ -0,0 +1,29 @@
...
+
+exename=$(basename "$1" .wasm)
+exename=$(basename "$exename" .dll)
+exename=$(basename "$exename" .js)
+dirname=$(dirname "$1")
+
+node="node --stack_trace_limit=100"
+WASM_HOST_ARGS_SEPERATOR=""
+
+ls $dirname
```
SingleAccretion:
```suggestion
```

src/tests/Common/scripts/nativeaottest.sh
```diff
@@ -10,6 +10,13 @@
...
-$_DebuggerFullPath $1/native/$exename "${@:3}"
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+
+source $SCRIPT_DIR/../../../../eng/testing/FindWasmHostExecutable.sh "$1/native/$2"
+
+if [ -n "${WASM_HOST_EXECUTABLE}" ]; then
+  $WASM_HOST_EXECUTABLE "$WASM_BINARY_TO_EXECUTE" "${@:3}"
+else
+  chmod +x $1/native/$exename
+  $_DebuggerFullPath $1/native/$exename "${@:3}"
```
SingleAccretion, line 21: What sets `exename` now?

yowl:
@jkotas  It seems as though `libz` is no longer part of the docker image, does that ring any bells?  I tried the fix from https://github.com/dotnet/runtime/issues/106566 but we still need `libz` at linking

jkotas:
I think it will work if you apply the full delta from https://github.com/dotnet/runtime/pull/106569.

The earlier commit was missing this part of the change: https://github.com/dotnet/runtime/pull/106569/files#diff-a8c821001409a0c4d6589a59e663d4d07e53b0127b345f17640410828c580b37L153-L154

yowl:
Thanks for the help and feedback.  Have addressed it and finally got it green.  As noted there are some changes to make it compile that are included in #2614 .  Whether they go here or there, not sure, both require the changes.

src/tests/Common/scripts/nativeaottest.sh
yowl: reverted the deleting of setting `exename`

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
yowl: Yes, that works, thanks

eng/pipelines/runtimelab-official.yml
yowl: ok, reverted

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: renamed.

yowl:
Will try to remove `DCMAKE_SYSTEM_NAME ` now we aren't using terminfo

yowl:
nuget error

eng/pipelines/common/platform-matrix.yml
yowl: renamed

eng/pipelines/common/platform-matrix.yml
yowl: Done

eng/pipelines/runtimelab.yml
yowl: Done.

eng/pipelines/runtimelab/install-llvm.ps1
yowl: no, thanks

eng/testing/FindWasmHostExecutableAndRun.sh
yowl: added log, thanks.

src/coreclr/jit/CMakeLists.txt
yowl: reverted

yowl:
Green, think I've got all the feedback addressed.

eng/pipelines/runtimelab-official.yml
```diff
@@ -35,6 +35,7 @@ extends:
         parameters:
           jobTemplate: /eng/pipelines/common/global-build-job.yml
           buildConfig: Release
+          container: azurelinux-3.0-cross-amd64-net9.0-sanitizer
```
SingleAccretion:
```suggestion
```

eng/pipelines/runtimelab/install-wasi-sdk.cmd
```diff
@@ -3,7 +3,7 @@ cd /D "%1"
 
 echo Installing Wasi SDK
 
-powershell -NoProfile -NoLogo -ExecutionPolicy ByPass -File "%~dp0install-wasi-sdk.ps1"
+powershell -NoProfile -NoLogo -ExecutionPolicy ByPass -File "%~dp0install-wasi-sdk.ps1" -InstallDir .
```
SingleAccretion: Now unused and can be deleted?

eng/pipelines/runtimelab/install-wasmer.ps1
```diff
@@ -0,0 +1,34 @@
+[CmdletBinding(PositionalBinding=$false)]
```
SingleAccretion: Now unused and can be deleted.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -30,12 +30,22 @@ steps:
     - ${{ if eq(parameters.platform, 'browser_wasm_win') }}:
       - script: |
           call $(Build.SourcesDirectory)\wasm-tools\emsdk\emsdk_env
-          $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) ${{ parameters.archType }} tree nativeaot /p:LibrariesConfiguration=${{ parameters.librariesConfiguration }}
+          $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) ${{ parameters.archType }} tree nativeaot /p:LibrariesConfiguration=${{ parameters.librariesConfiguration }} /p:TestWrapperTargetsWindows=true
```
SingleAccretion: It is concerning that `/p:TestWrapperTargetsWindows=true` is needed to make things work _on Windows_. Does that mean the local test build is broken?

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -51,18 +61,30 @@ steps:
             call $(Build.SourcesDirectory)\wasm-tools\emsdk\emsdk_env
             $(Build.SourcesDirectory)/src/tests/run$(scriptExt) runnativeaottests $(buildConfigUpper) ${{ parameters.archType }} ${{ parameters.osGroup }}
           displayName: Run WebAssembly tests in single file mode
+      - ${{ elseif in(parameters.platform, 'wasi_wasm_linux_x64_naot_llvm') }}:
+        - script: |
+            $(Build.SourcesDirectory)/src/tests/run$(scriptExt) --runnativeaottests ${{ parameters.archType }} $(buildConfigUpper) wasi
+          displayName: Run WebAssembly tests in single file mode
```
SingleAccretion: Why does need to be separate? AFAICT it is just the same as the `${{ else }}:` below, just different parameter order.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -51,18 +61,30 @@ steps:
...
+      - ${{ if in(parameters.platform, 'browser_wasm_win', 'wasi_wasm_win') }}:
+        - script: |
+              call $(Build.SourcesDirectory)\wasm-tools\emsdk\emsdk_env
+              $(Build.SourcesDirectory)/build$(scriptExt) libs.tests -test -a ${{ parameters.archType }} -os ${{ parameters.osGroup }} -lc ${{ parameters.librariesConfiguration }} -rc $(buildConfigUpper) /p:TestNativeAot=true /p:RunSmokeTestsOnly=true
+          displayName: Build and run WebAssembly libraries tests
+      - ${{ else }}:
+        - script: |
+              source $(Build.SourcesDirectory)/wasm-tools/emsdk/emsdk_env.sh
+              $(Build.SourcesDirectory)/build$(scriptExt) libs.tests -test -a ${{ parameters.archType }} -os ${{ parameters.osGroup }} -lc ${{ parameters.librariesConfiguration }} -rc $(buildConfigUpper) /p:TestNativeAot=true /p:RunSmokeTestsOnly=true
+          displayName: Build and run WebAssembly libraries tests
```
SingleAccretion: Why does this need to be separate? `emsdk_env.sh` is already part of `build.sh`.

src/tests/Directory.Build.props
```diff
@@ -178,8 +178,8 @@
 
   <!-- Set Test Wrapper running host OS -->
   <PropertyGroup>
-    <TestWrapperTargetsWindows>false</TestWrapperTargetsWindows>
-    <TestWrapperTargetsWindows Condition=" ('$(TargetsWindows)' != '' And '$(TargetsWindows)' ) OR ('$(TargetOS)' == 'android' And '$(TargetArchitecture)' == 'arm64' ) OR ($(WindowsHost) And '$(TargetArchitecture)' == 'wasm')">true</TestWrapperTargetsWindows>
+    <TestWrapperTargetsWindows Condition="'$(TestWrapperTargetsWindows)' == ''">false</TestWrapperTargetsWindows>
+    <TestWrapperTargetsWindows Condition=" ('$(TargetsWindows)' != '' And '$(TargetsWindows)' ) OR ('$(TargetOS)' == 'android' And '$(TargetArchitecture)' == 'arm64' )">true</TestWrapperTargetsWindows>
```
SingleAccretion: It is not clear to me any changes to `TestWrapperTargetsWindows` logic are required.

Here's how the upstream condition will evaluate on Windows-targeting-WASM:
- ` ('$(TargetsWindows)' != '' And '$(TargetsWindows)' ) OR ('$(TargetOS)' == 'android' And '$(TargetArchitecture)' == 'arm64' ) OR ($(WindowsHost) And '$(TargetArchitecture)' == 'wasm')` => `true` due to `(($(WindowsHost) And '$(TargetArchitecture)' == 'wasm')`.
On Not-Windows-targeting-WASM:
- => `""` (none of the Windows conditions are `true`).

Which is how it is supposed to be.

src/tests/build.sh
```diff
@@ -318,6 +320,30 @@ handle_arguments_local() {
             fi
             ;;
 
+        target-os*|-target-os*)
```
SingleAccretion: I think it would be better to structure this the same as the Windows and `run.sh` script: Add `wasi`/`browser` arguments (possibly also `wasm`, though it's not strictly required). It would make the command lines more consistent and simplify CI.

src/tests/Common/scripts/nativeaottest.sh
```diff
@@ -13,3 +13,13 @@
...
 $_DebuggerFullPath $1/native/$exename "${@:3}"
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
+source $SCRIPT_DIR/../../../../eng/testing/FindWasmHostExecutable.sh "$1/native/$2"
+
+if [ -n "${WASM_HOST_EXECUTABLE}" ]; then
+  $WASM_HOST_EXECUTABLE "$WASM_BINARY_TO_EXECUTE" "${@:3}"
+else
+  chmod +x $1/native/$exename
+  $_DebuggerFullPath $1/native/$exename "${@:3}"
```
SingleAccretion:
```suggestion
SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
source $SCRIPT_DIR/../../../../eng/testing/FindWasmHostExecutable.sh "$1/native/$2"

if [ -n "${WASM_HOST_EXECUTABLE}" ]; then
  $WASM_HOST_EXECUTABLE "$WASM_BINARY_TO_EXECUTE" "${@:3}"
else
  exename=$(basename $2 .dll)
  chmod +x $1/native/$exename
  $_DebuggerFullPath $1/native/$exename "${@:3}"
```
It looks to be running `$_DebuggerFullPath $1/native/$exename "${@:3}"` twice, which can't be right...

src/tests/Directory.Build.props
yowl: You are right, it's not required. Thanks.

eng/pipelines/runtimelab/install-wasmer.ps1
yowl: deleted

eng/pipelines/runtimelab/install-wasi-sdk.cmd
yowl: Right

eng/pipelines/runtimelab-official.yml
yowl: Done

src/tests/build.sh
yowl: thanks, done.

eng/pipelines/runtimelab/install-llvm.ps1
yowl: Added

yowl:
A problem with our timers perhaps

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.cs
```diff
@@ -10,7 +10,8 @@
 using System.Threading;
 using System.Threading.Tasks;
 using System.Diagnostics.Metrics;
-#if TARGET_WASI
+// TODO-LLVM: This is not upstreamable and should be deleted when https://github.com/dotnet/runtimelab/pull/2614 is merged
+#if TARGET_WASI && !NATIVE_AOT
```
SingleAccretion, line 14: Could you open an issue about reverting these changes?

src/tests/build.proj
```diff
@@ -615,6 +615,7 @@
       <GroupBuildCmd Condition="'$(IlcMultiModule)' == 'true'">$(GroupBuildCmd) "/p:IlcMultiModule=true"</GroupBuildCmd>
       <GroupBuildCmd Condition="'$(IlcUseServerGc)' == 'false'">$(GroupBuildCmd) "/p:IlcUseServerGc=false"</GroupBuildCmd>
       <GroupBuildCmd Condition="'$(BuildNativeAotFrameworkObjects)' == 'true'">$(GroupBuildCmd) "/p:BuildNativeAotFrameworkObjects=true"</GroupBuildCmd>
+      <GroupBuildCmd Condition="'$(TestWrapperTargetsWindows)' ==  'true'">$(GroupBuildCmd) "/p:TestWrapperTargetsWindows=true"</GroupBuildCmd>
```
SingleAccretion:
```suggestion
```

src/tests/build.sh
```diff
@@ -167,6 +167,8 @@ usage_list+=("-test:xxx - Only build the specified test project ^(relative or ab
 usage_list+=("-dir:xxx - Build all test projects in the given directory ^(relative or absolute directory under src\tests^).");
 usage_list+=("-tree:xxx - Build all test projects in the given subtree ^(relative or absolute directory under src\tests^).");
 usage_list+=("-log:xxx - Base file name to use for log files (used in lab pipelines that build tests in multiple steps to retain logs for each step).")
+usage_list+=("-broser - Target Wasm/browser.")
```
SingleAccretion:
```suggestion
usage_list+=("-browser - Target Wasm/browser.")
```

src/tests/Common/CLRTest.Execute.Bash.targets
```diff
@@ -341,7 +341,7 @@ fi
       <WatcherRunFile>"$CORE_ROOT/watchdog" $_WatcherTimeoutMins</WatcherRunFile>
 
       <!-- Note that this overwrites CLRTestBashPreCommands rather than adding to it. -->
-      <CLRTestBashPreCommands Condition="'$(CLRTestKind)' == 'BuildAndRun' and '$(TargetArchitecture)' == 'wasm'"><![CDATA[
+      <CLRTestBashPreCommands Condition="'$(CLRTestKind)' == 'BuildAndRun' and '$(TargetArchitecture)' == 'wasm' And '$(TargetOS)' == 'browser' And '$(RuntimeFlavor)' == 'mono'"><![CDATA[
```
SingleAccretion: Is the `'$(TargetOS)' == 'browser'` check necessary?

SingleAccretion (approved):
LGTM modulo feedback.

Thank you 🚀!

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.cs
yowl: Yep, https://github.com/dotnet/runtimelab/issues/2666

src/tests/Common/CLRTest.Execute.Bash.targets
yowl: no, removed.

yowl:
Thanks for the reviews!

jkotas:
Great work! Thanks a lot

Issue #2606, opened 2024-06-09
[NativeAOT-LLVM] [Question]: Should GCPointerMap have a provision for adding optional fields to GCStaticEETypeNode


yowl:
WIth this code
```cs
    public static class InteropReturnArea
    {
        [InlineArray(3)]
        internal struct ReturnArea
        {
            private ulong buffer;
            
            internal unsafe nint AddressOfReturnArea()
            {
                return (nint)Unsafe.AsPointer(ref buffer);
            }
        }
        
        [ThreadStatic]
        [FixedAddressValueType]
        internal static ReturnArea returnArea = default;
```

`ReturnArea` should be align 8, so `GCStaticEETypeNode` should be written with an optional field `OptionalFieldsFlag` I assume.  The mechanism seems to be to do with `GCPointerMap` but there is nothing present to do this.  Am I off track here or is there something missing?

Thanks

jkotas:
> Should GCPointerMap have a provision for adding optional fields to GCStaticEETypeNode 

I do not think we want to have a generic provision for arbitrary optional fields in GCPointerMap. 

We may want to track the `RequiresAlign8` in GCPointerMap (more convenient); or `RequiresAlign8` can be tracked independently from `GCPointerMap`.

`GCStaticEETypeNode` for regular statics has the same problem, but it is masked by the thread statics being allocated on `PINNED_OBJECT_HEAP` that happens to guarantee 8-byte alignment even on 32-bit platforms. A quick work around for thread statics would be to allocate all thread statics conservatively with 8-byte alignment.

Linux arm32 is probably affected by this bug as well. We may want to fix it in upstream. cc @filipnavara

jkotas:
I have opened https://github.com/dotnet/runtime/issues/103234 for the upstream bug on Linux arm32

filipnavara:
> Linux arm32 is probably affected by this bug as well. We may want to fix it in upstream. cc @filipnavara

Sounds like it. I am not sure I fully understand the scope of the issue yet. Generally, Align8 was one of the last items on the https://github.com/dotnet/runtime/issues/97729 list.

PR #2609, opened 2024-06-10
[NativeAOT-LLVM] Workaround to align thread statics to 8 byte boundary

yowl:
This PR is a workaround to align all thread statics to 8 byte boundary to ensure those statics that do require 8 byte alignment are aligned.

Follows https://github.com/dotnet/runtimelab/issues/2606

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/ThreadStatics.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/ThreadStatics.cs
@@ -138,7 +138,7 @@ namespace Internal.Runtime
             }
 
 #if TARGET_WASM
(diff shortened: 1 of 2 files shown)


yowl:
cc @dotnet/nativeaot-llvm

SingleAccretion:
Can we add a test that reliable fails without this change and reliably passes with this change?

Idea: since objects are allocated next to each other, usually, and thread static bases are allocated lazily, check that if we access thread statics of three types, `A`, `B` and `C`, where `A` and `C` need to be 8-aligned, while `B` has an 'odd' size (three pointers, so, just an `int` thread static field, for example), both `A` and `C` get aligned properly.

yowl:
Not sure this is the simplest test, but it was failing.

SingleAccretion (approved):
LGTM. Thank you!

PR #2610, opened 2024-06-10
[NativeAOT-LLVM] JSExport-based async main

SingleAccretion:
The idea is simple:
1) Source generate a JSExport thunk for the entry point with a well-known name.
2) Invoke that JSExport - the existing host code already does this.

The complications are in the finer details. For example, this change utilizes a hidden file injected into user compilations to achieve the JSExport bit, to simplify the code generation. Even with that, due to C# rules, we cannot generate wrappers for all possible entry point definitions.


SingleAccretion:
@dotnet/nativeaot-llvm

maraf (approved):
Looks good to me 👍
Thank you!

PR #2612, opened 2024-06-11
[NativeAOT-LLVM] Don't use the `emcc` wrapper for bitcode compilation

### LINKED ISSUE #2603: [NativeAOT-LLVM] Use clang from EMSDK directly to compile bitcode
We currently use `emcc`, which involves a number of wrapper layers that are unnecessary when compiling bitcode.

SingleAccretion:
The wrapper just adds overhead.

Also remove unnecessary clang arguments. Also fix a warning.

Fixes #2603.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -127,6 +127,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <WasmOptimizationSetting Condition="$(Optimize) == 'true' and $(OptimizationPreference) != 'Size' and '$(_targetOS)' == 'wasi'">-O2</WasmOptimizationSetting>
     <WasmOptimizationSetting Condition="$(Optimize) == 'true' and $(OptimizationPreference) == 'Size'">-Oz</WasmOptimizationSetting>
 
+    <IlcLlvmTarget Condition="'$(_targetOS)' == 'browser'">wasm32-unknown-emscripten</IlcLlvmTarget>
     <IlcLlvmTarget Condition="'$(_targetOS)' == 'wasi'">wasm32-unknown-wasi</IlcLlvmTarget>
     <IlcWasmStackSize>1048576</IlcWasmStackSize> <!-- 1MB -->
     <IlcWasmStackSize Condition="'$(EmccStackSize)' != ''">$(EmccStackSize)</IlcWasmStackSize>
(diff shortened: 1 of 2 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
LGTM, thanks!

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -387,10 +388,11 @@ The .NET Foundation licenses this file to you under the MIT license.
   </Target>
 
   <Target Name="CheckWasmSdks">
-    <!-- The EMSDK is typically set by user and points to a standard emscripten layout -->
-    <!-- The EmscriptenUpstreamEmscriptenPath comes from upstream mono workload and points to an upstream folder of emscripten. The EmscriptenUpstreamBinPath points to bin folder -->
-    <Error Text="Emscripten not found, not compiling to WebAssembly. To enable WebAssembly compilation, install Emscripten and ensure the EMSDK environment variable points to the directory containing upstream/emscripten/emcc.bat"
-           Condition="'$(EMSDK)' == '' and '$(EmscriptenUpstreamEmscriptenPath)' == '' and '$(_targetOS)' == 'browser'" />
+    <!-- $(EMSDK) is typically set by the user and points to a standard emscripten layout. -->
+    <!-- $(EmscriptenSdkToolsPath) comes from upstream mono workload and points to the dotnet/emsdk-style SDK layout. -->
```
maraf, line 392: The `EmscriptenSdkToolsPath` is used to build mono/browser.proj in upstream, whereas `EmscriptenUpstreamEmscriptenPath` is defined by mono workload that is used on user's machine to build mono-based wasm apps.

The goal is to be able to use mono workload

maraf: https://github.com/dotnet/emsdk/blob/main/eng/sdk_files/Emscripten.Sdk.props

maraf (changes requested):

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: > https://github.com/dotnet/emsdk/blob/main/eng/sdk_files/Emscripten.Sdk.props

Ehhm, doesn't this define the same `EmscriptenSdkToolsPath` property?

The goal with the property switchover is to have one knob to override the SDK from the command line if desired.

maraf: I forgot that, but the question is whether to rely on internal structure or use provided properties... We can possibly calculate with the idea that the layout won't change..

maraf (approved):

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: There is indeed a tradeoff in it. Using a single property as the source of truth I think has more practical value.

In the long term, I would also like to converge the two layouts back together. It makes development-time scenarios (testing new EMSDKs / LLVMs) significantly easier and has no downside. It would also make the system simpler

PR #2613, opened 2024-06-12
[NativeAOT-LLVM] Don't always build the runtime and fix static graph restore for BuildIntegration.proj

SingleAccretion:
The subset condition as written leads to always invoking runtime build, which is undesirable.

Also change the workaround for static graph restore in `BuildIntegration.proj` to a better working option.


SingleAccretion:
@dotnet/nativeaot-llvm

maraf (approved):
Looks good to me 👍 Thank you!

PR #2615, opened 2024-06-17
[NativeAOT-LLVM] Remove the LLVMSharp dependency

SingleAccretion:
Use the LLVM buit into the Jit instead, with a thin interop layer.

This has a number of benefits:
1) No dependency on external build processes.
2) "By construction" LLVM version coherency.
3) Smaller package size.
4) Ability to expose anything we may need from the LLVM's C++ API.

The downside is that we need to write a bit more code.

This is a no-diff change.


SingleAccretion:
@dotnet/nativeaot-llvm

maraf (approved):
AFAIU it looks good to me 👍

yowl (approved):
Thanks, this is a nice simplification of the architecture.

PR #2634, opened 2024-07-17
[NativeAOT-LLVM] Implement automatic finalization

SingleAccretion:
Drain the finalizer queue on exit from allocation helpers on their slow path, as well as in `GC.Collect()`.

Fixes https://github.com/dotnet/runtimelab/issues/2240.

### DIFF
--- a/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/__Finalizer.SingleThreaded.cs
+++ b/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/__Finalizer.SingleThreaded.cs
@@ -12,7 +12,7 @@ namespace System.Runtime
 {
     internal static class __Finalizer
     {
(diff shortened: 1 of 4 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

maraf (approved):
Looks good to me. Thank you!

yowl:
Sorry, bit late to this, is it possible to add a test for this in `HelloWasm` ?

SingleAccretion:
> Sorry, bit late to this, is it possible to add a test for this in HelloWasm ?

It seems challenging to test without assuming the "synchronous" semantics implemented here, and without `WaitForPendingFinalizers` (for which we already have a test).

pavelsavara:
Naive question: what behavioral/semantics expectations we are breaking by not running finalizers on a dedicated thread ? Are there scenarios which break ?

SingleAccretion:
> Naive question: what behavioral/semantics expectations we are breaking by not running finalizers on a dedicated thread ? Are there scenarios which break ?

I think we'll have to find out through testing and usage experience. One can certainly imagine code that assumes threads exists would be problematic, but that's not a finalization-specific problem.

PR #2636, opened 2024-07-18
Added TypeSpec hierarchy to swift-bindings

stephen-hawley:


src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text;
+
+namespace BindingsGeneration;
+
+public abstract class TypeSpec
```
kotlarmilos, line 19: Remind me, what is the difference between TypeSpec and TypeDecl/ArgumentDecl?

kotlarmilos: Since we use these types for parsing, is the `src/Model` a good path for these types? Do we need a new directory?

src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
...
+	protected TypeSpec(TypeSpecKind kind)
+	{
+		Kind = kind;
+		GenericParameters = new List<TypeSpec>();
+		Attributes = new List<TypeSpecAttribute>();
+	}
+
+	public TypeSpecKind Kind { get; private set; }
+	public List<TypeSpec> GenericParameters { get; private set; }
+	public bool ContainsGenericParameters { get { return GenericParameters.Count != 0; } }
```
kotlarmilos, line 43: I know that these getters are needed later, but this may be a good opportunity to review/refactor them. They look too complex to me. Are they frequently used?

src/Swift.Bindings/src/Model/TypeSpec.cs
kotlarmilos, line 19: Since the codebase is still small, it may be good to add/follow doc annotations for these types.

src/Swift.Bindings/src/Model/TupleTypeSpec.cs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Text;
+
+namespace BindingsGeneration;
+
+public class TupleTypeSpec : TypeSpec
```
kotlarmilos, line 11: I would expect to have a declaration for TupleType as well. Is it going to be added in the next PR?

src/Swift.Bindings/tests/TypeSpecTests/TypeSpecTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+        }
+
+        private static void InitializeResources()
+        {
+        }
+    }
+
+    [Fact]
+    public static void TestNamedBasicName()
```
kotlarmilos, line 26: Thank you for adding these tests!

kotlarmilos (commented):
Thank you for your contribution! Overall, it looks good. I have two questions:
1. The changes contain unused helper functions, and I am concerned about accumulating "unnecessary" code. Can we adopt a more test-driven approach and add only the necessary functions for this stage? I know that it takes more time but this is great opportunity to review and refactor them.
2. How can we ensure that the changes are effective without integration and E2E test cases? Do you think these tests can be added here with our help, or would you prefer including them in a follow-up PR?

src/Swift.Bindings/src/Model/TypeSpec.cs
stephen-hawley: If you want, I'm happy to move it, but these are parts of the actual model since the next step for me is to refactor the *Decl code to use `TypeSpec`. In terms of the difference, a `*Decl` is a declaration of a type which includes all of its members. A `TypeSpec` is a specification of a type with no information about members (etc). You would use A `TypeSpec` within a `*Decl` to label the types of its members.
For example, right now an `ArgumentDecl` has a `TypeIdentifier` which is a `TypeDecl`, but that really should be a `TypeSpec`.

From a modeling point of view:
``` swift
public struct Foo { // Foo is a StructDecl because it *declares* the type Foo
    public var X: (Int, Bool) // and x is a FieldDecl because it *declares* the field X,
                                            // however (Int, Bool) is a type specification because it specifies the type of the field,
                                            // but includes no further information
}

File: src/Swift.Bindings/src/Model/TypeSpec.cs (hunk 2 above)
Comment (stephen-hawley) [reply src/Swift.Bindings/src/Model/TypeSpec.cs:43]: Yes, they are frequently used.
Especially for example if I have:
```
public func Foo<T>(a: SomeContainer<Int, T>) { }
```
I care very much that the TypeSpec `SomeContainer<Int, T>` is a reference to a partially bound generic type and I care very much to be able to ask if `Int` and `T` are bound or not.

File: src/Swift.Bindings/src/Model/TupleTypeSpec.cs (hunk 3 above)
Comment (stephen-hawley) [reply src/Swift.Bindings/src/Model/TupleTypeSpec.cs:11]: I think it's important to have a clear distinction between type declarations and the specifications (see above) of members within those types. I can understand how you might think that tuples, protocols lists, and closures belong in the `*Decl` but `Swift.String` as a `TypeSpec` is a **reference** to the `StructDecl` `Swift.String`. Besides being a compact representation, using a `TypeSpec` avoids circular references in the *Decl and makes referencing a type that outside your module cheap up front.

For example if I have this:
```swift
public class Foo {
     public init (other: Foo) { /* copy constructor */ }
     public init (externData: SomeModule.ExternData) { }
}
```
The first ctor has a reference to `Foo` in the argument list and if represented by a `ClassDecl` will make a circular reference and the second ctor has a reference to an external module, the loading of which can be deferred until it's needed.

src/Swift.Bindings/src/Model/TypeSpecKind.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+namespace BindingsGeneration;
+
+public enum TypeSpecKind
```
kotlarmilos, line 9: Should we move this to the TypeSpec?

src/Swift.Bindings/src/Model/TypeSpecAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Linq;
+
+namespace BindingsGeneration;
+
+public class TypeSpecAttribute
+{
+    public TypeSpecAttribute(string name)
+    {
+        Name = name;
+        Parameters = new List<string>();
```
kotlarmilos, line 21: Could you provide some examples here? Is the string a good type here?

src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
...
+	public List<TypeSpecAttribute> Attributes { get; private set; }
+	public bool HasAttributes { get { return Attributes.Count != 0; } }
+	public bool IsInOut { get; set; }
+	public bool IsAny { get; set; }
+	public virtual bool IsEmptyTuple { get { return false; } }
```
kotlarmilos, line 69: Is it possible to move IsEmptyTuple to TupleTypeSpec?

src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
...
+	public static bool BothNullOrEqual(TypeSpec? one, TypeSpec? two)
+	{
+		if (one is null && two is null)
+			return true;
+		if (one is null || two is null)
+			return false;
+		return one.Equals(two);
+	}
+
+	public bool ContainsBoundGenericClosure()
```
kotlarmilos: Perhaps we can move this to ClosureTypeSpec?

src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
...
+
+	public TypeSpecKind Kind { get; private set; }
+	public List<TypeSpec> GenericParameters { get; private set; }
+	public bool ContainsGenericParameters { get { return GenericParameters.Count != 0; } }
+	public List<TypeSpecAttribute> Attributes { get; private set; }
+	public bool HasAttributes { get { return Attributes.Count != 0; } }
+	public bool IsInOut { get; set; }
+	public bool IsAny { get; set; }
+	public virtual bool IsEmptyTuple { get { return false; } }
+	protected abstract string LLToString(bool useFullName);
```
kotlarmilos, line 76: Why LLToString is needed?

src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
...
+			return false;
+		return one.Equals(two);
+	}
+
+	public bool ContainsBoundGenericClosure()
+	{
+		return ContainsBoundGenericClosure(0);
+	}
+
+	bool ContainsBoundGenericClosure(int depth)
```
kotlarmilos: Perhaps we can move this to ClosureTypeSpec?

src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
...
+				builder.Append(GenericParameters[i].ToString(useFullNames));
+			}
+			builder.Append('>');
+		}
+		builder.Append(LLFinalStringParts());
+
+		return builder.ToString();
+	}
+
+	static string[] intNames = {
```
kotlarmilos: I would move this to the type database.

src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
...
+			"Swift.Int", "Swift.UInt", "Swift.Int8", "Swift.UInt8",
+			"Swift.Int16", "Swift.UInt16", "Swift.Int32", "Swift.UInt32",
+			"Swift.Int64", "Swift.UInt64", "Swift.Char"
+		};
+
+	public static bool IsIntegral(TypeSpec ts)
```
kotlarmilos: Should these getters be in TypeDecl?

src/Swift.Bindings/src/Model/TypeSpec.cs
```diff
@@ -0,0 +1,380 @@
...
+
+	public TypeSpec ReplaceName(string toFind, string replacement)
+	{
+		var result = this;
+		if (!String.IsNullOrEmpty(replacement))
+			ReplaceName(this, toFind, replacement, ref result);
+		return result;
+	}
+
+	static bool ReplaceName(TypeSpec original, string toFind, string replacement, ref TypeSpec result)
```
kotlarmilos, line 327: What ReplaceName functions do?

src/Swift.Bindings/src/Model/TypeSpecAttribute.cs
stephen-hawley: It isn't, but it's not the worst. In this case, the contents of attributes are incredibly varied. In the spec, it's described as a Dyck language (balanced brackets, braces, etc) but not a lot of other detail. It looks like some of the common attributes are preparsed with the information that we care about set aside.  Mostly we care about the attributes for a couple really important things: `@available` and `@escaping`. The latter has no parameters, while the former has very predictable ones.

src/Swift.Bindings/src/Model/TypeSpec.cs
stephen-hawley: No - this is a convenience method that can be used on any `TypeSpec` without asking if it's a tuple.

src/Swift.Bindings/src/Model/TypeSpec.cs
stephen-hawley: No - this drills down into all the components of all type specs. A `NamedTypeSpec` can contain a bound generic closure in its generic parameters, for example. I don't see an immediate need for this one, so I think I'll take it out entirely.

src/Swift.Bindings/src/Model/TypeSpec.cs
stephen-hawley: ToString in `TypeSpec` takes care of all the decor around a TypeSpec (`inout`, `any` and generic components). Subclasses only need to concentrate on the bulk of writing. This avoids code duplication.

src/Swift.Bindings/src/Model/TypeSpec.cs
stephen-hawley: It should probably be in the type database as `IsIntegral (TypeDecl td)` and `IsIntegral(TypeSpec ts)`. I don't see it used immediately, so I'll remove it entirely.

kotlarmilos (approved):

PR #2639, opened 2024-07-29
Infrastructure for swift entry point demangling

stephen-hawley:
This PR adds infrastructure for demangling Swift entry points.
A couple things to note: the demangler itself is a port of the Apple demangler to C#. The actual demangling entry point has been
commented out until the rest of the code has been refactored. In reviewing this PR, I wouldn't spend a lot of time looking at Swift5Demangler.cs as it is the actual ported code. As a result of the port it contains code which is less than idiomatic in C# but the cost of changing it will be costly when it needs to get updated in the future.

### DIFF
--- a/src/Swift.Bindings/src/Demangler/PunyCode.cs
+++ b/src/Swift.Bindings/src/Demangler/PunyCode.cs
@@ -9,6 +9,10 @@ using System.Text;
 namespace BindingsGeneration.Demangling {
 	/// <Summary>
 	/// A class for decoding puny code. Encoder has been removed.
+	/// Note that this is not true puny code but a variant where _ is used as the
+	/// delimeter and ent symbology is for encoding is [a-zA-J].
+	/// Also non-symbol ASCII caracters (except [$_a-zA-Z0=9]) are mapped to
+	/// the code range d800-d880 and are encoded like non-ascii characters.
 	/// </Summary>
 	public class PunyCode {
 		const string kEncodingStr = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJ";
--- a/src/Swift.Bindings/src/Demangler/PunyCode.cs
+++ b/src/Swift.Bindings/src/Demangler/PunyCode.cs
@@ -1,0 +1,107 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace BindingsGeneration.Demangling {
+	/// <Summary>
+	/// A class for decoding puny code. Encoder has been removed.
+	/// </Summary>
+	public class PunyCode {
+		const string kEncodingStr = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJ";
+
+		static Dictionary<char, int> decodeTable = MakeDecodeTable(kEncodingStr);
+		const int kBase = 36;
+		const int tMin = 1;
+		const int tMax = 26;
+		const int skew = 38;
+		const int damp = 700;
+		const int initialBias = 72;
+		const int initialN = 0x80;
+		const char delimiter = '_';
+
+		public PunyCode ()
+		{
+		}
+
+		static Dictionary<char, int> MakeDecodeTable (string s)
+		{
+			var table = new Dictionary<char, int> ();
+			for (int i = 0; i < s.Length; i++) {
+				table [s [i]] = i;
+			}
+			return table;
+		}
+
+		static int Adapt (int delta, int numPoints, bool firstTime)
+		{
+			delta = delta / (firstTime ? damp : 2);
+
+			delta += delta / numPoints;
+			int k = 0;
+			while (delta > ((kBase - tMin) * tMax) / 2) {
+				delta = delta / (kBase - tMin);
+				k = k + kBase;
+			}
+			k += ((kBase - tMin + 1) * delta) / (delta + skew);
+			return k;
+		}
+
+		/// <Summary>
+		/// Decode a puny coded string
+		/// </Summary>
+		public string Decode (string input)
+		{
+
+			int n = initialN;
+			int i = 0;
+			int bias = initialBias;
+			var output = new StringBuilder ();
+
+			int pos = 0;
+			var delim = input.LastIndexOf ('_');
+			if (delim > 0) {
+				output.Append (input.Substring (0, delim));
+				pos = delim + 1;
+			}
+
+			int outputLength = output.Length;
+			int inputLength = input.Length;
+			while (pos < inputLength) {
+				int oldi = i;
+				int w = 1;
+				for (int k = kBase; ; k += kBase) {
+					int digit = decodeTable [input [pos++]];
+					i = i + (digit * w);
+					int t = Math.Max (Math.Min (k - bias, tMax), tMin);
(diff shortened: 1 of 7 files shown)


kotlarmilos:
Thanks for adding the docs annotations. I just want to clarify the change: Are we planning to use the demangler to parse the ABI JSON mangled names? Additionally, could you add some interfaces/abstract classes as integration points? 

It's not clear at this stage where and how it will be used. Also, please include basic test coverage for these changes.

stephen-hawley:
> Thanks for adding the docs annotations. I just want to clarify the change: Are we planning to use the demangler to parse the ABI JSON mangled names? Additionally, could you add some interfaces/abstract classes as integration points?
> 
> It's not clear at this stage where and how it will be used. Also, please include basic test coverage for these changes.

No, this is for demangling symbols from the binary output so you can get thunks and protocol witness tables which are not present in the abi json. That will sort out the problems.

I can add tests, but they won't help at this stage since there is no functioning entry point (yet). That will come in the next PR. I'm trying to stage things in in pieces to make things easier to review.

src/Swift.Bindings/src/Demangler/PunyCode.cs
```diff
@@ -0,0 +1,107 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace BindingsGeneration.Demangling {
+	/// <Summary>
+	/// A class for decoding puny code. Encoder has been removed.
+	/// </Summary>
+	public class PunyCode {
```
matouskozak, line 17: Why is this needed demangling? Are the Swift entry point represented using punycode encoding?

src/Swift.Bindings/src/Demangler/StringSlice.cs
```diff
@@ -0,0 +1,215 @@
...
+
+using System;
+using System.Text;
+using System.Collections.Generic;
+
+namespace BindingsGeneration.Demangling {
+	/// <Summary>
+	/// Represents a a slice which includes facilities similar to StringStream, but
+	/// simplified and more performant. This is build to work specifically with the
+	/// demangler.
```
matouskozak, line 12: I'm not sure if it's worth having a separate `StringSlice` class instead of using for example `(ReadOnly)Span` to represent a string slice. Span is heavily optimized inside .NET 9. Do you know what benefits this class has in the bigger picture of the tooling?

src/Swift.Bindings/src/Demangler/StringSlice.cs
```diff
@@ -0,0 +1,215 @@
...
+			return Current == c;
+		}
+
+		/// <Summary>
+		/// Returns true if the slice starts with the given string
+		/// </Summary>
+		public bool StartsWith (string s)
+		{
+			if (s == "")
+				return true; // I guess?
```
matouskozak, line 39: What does this "special" case translate into in real-world scenarios?

src/Swift.Bindings/src/Demangler/Swift5Demangler.cs
```diff
@@ -0,0 +1,2961 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+#nullable disable
+
+// This is a port of the Apple Swift 5 demangler
```
matouskozak, line 10: Is this ported from the old BTfS? 

You mentioned in the PR description that the cost of changing this will be high. Based on the name I suspect the demangler is Swift5 specific. With the Swift6 release, do you know if there are any necessary changes that we will need to do?

matouskozak (commented):
I'm not confident I can review this thoroughly since I haven't worked on the tooling yet. I put a few comments just to clarify a few points.

src/Swift.Bindings/src/Demangler/PunyCode.cs
stephen-hawley: Yes. Since these are entry points in a standard .dylib, they have to be ascii encoded so they adopt puny code for anything weird.

src/Swift.Bindings/src/Demangler/StringSlice.cs
stephen-hawley: StringSlice was written and in place before Span was ready. I'll look at replacing it.

src/Swift.Bindings/src/Demangler/StringSlice.cs
stephen-hawley: None other than it seemed to make sense in a mathematical sense and, as it turns out `"foo".StartsWith("")` evaluates to true.

src/Swift.Bindings/src/Demangler/Swift5Demangler.cs
stephen-hawley: This is moved from BTfS, yes. I believe I read a few years back that Apple has committed to this format moving forward. That doesn't mean that there won't be changes, but they need to be backwards compatible to this because of library evolution. And, as it turns out, since part of the reason why this is being used is to be able to get the thunks for instance methods, if they changed the way those work, they would be breaking everything pre 6 with library evolution.

src/Swift.Bindings/src/Demangler/PunyCode.cs
matouskozak: Do you think we could re-use the .NET `IdnMapping` implemenation for this https://learn.microsoft.com/en-us/dotnet/api/system.globalization.idnmapping?view=net-9.0 ? It has both `GetAscii` and `GetUnicode` encode and decode functions.

stephen-hawley: The problem with this is that it's not 100% puny code. From the Apple repository:
```
// These functions implement a variant of the Punycode algorithm from RFC3492,
// originally designed for encoding international domain names, for the purpose
// of encoding Swift identifiers into mangled symbol names. This version differs
// from RFC3492 in the following respects:
// - '_' is used as the encoding delimiter instead of '-'.
// - Encoding digits are represented using [a-zA-J] instead of [a-z0-9], because
//   symbol names are case-sensitive, and Swift mangled identifiers cannot begin
//   with a digit.
// - Optionally, non-symbol ASCII characters (characters except [$_a-zA-Z0-9])
//   are mapped to the code range 0xD800 - 0xD880 and are also encoded like
//   non-ASCII unicode characters.
```

matouskozak: Thank you for the detailed info. I would add a comment to the `<Summary>` that this is Swift specific punycode.

src/Swift.Bindings/src/Demangler/PunyCode.cs
```diff
@@ -0,0 +1,107 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace BindingsGeneration.Demangling {
+	/// <Summary>
+	/// A class for decoding puny code. Encoder has been removed.
```
matouskozak:
```suggestion
	/// A class for decoding Swift variant of puny code (https://github.com/swiftlang/swift/blob/main/docs/ABI/Mangling.rst). Encoder has been removed.
```

kotlarmilos (approved):

PR #2643, opened 2024-08-01
[NativeAOT-LLVM] Add a native WASM object writer

SingleAccretion:
This adds an object writer that directly emits WASM object files (see https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md) and uses it to emit data symbols. It is anticipated that future changes will use it for code (compiler-generated helpers) as well.

We have also decided to delete the LLVM-based object writer, since doing otherwise would imply either maintaining it (duplication of effort) or letting it bitrot. It means that ILC will no longer produce all output as LLVM bitcode.

There are two motivations for this change:
1) (Primary) https://github.com/dotnet/runtimelab/issues/2404 - we need to be able to emit relocations LLVM doesn't support.
2) (Secondary) Performance: the new object writer will, of course, skip all of the LLVM steps, which have been observed to be rather lengthy on larger projects due to some algorithmic issues in LLVM.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMCodegenCompilation.CodeGen.cs
+++ b/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMCodegenCompilation.CodeGen.cs
@@ -6,6 +6,7 @@ using System.Diagnostics;
 using System.Numerics;
 
 using ILCompiler.DependencyAnalysis;
+using ILCompiler.ObjectWriter;
 
 using Internal.JitInterface;
 using Internal.JitInterface.LLVMInterop;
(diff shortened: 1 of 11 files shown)


yowl:
This looks great!

SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMCodegenCompilation.CodeGen.cs
```diff
@@ -47,6 +48,116 @@ public override ISymbolNode GetExternalMethodAccessor(MethodDesc method, ReadOnl
 
         public override CorInfoLlvmEHModel GetLlvmExceptionHandlingModel() => Options.ExceptionHandlingModel;
 
+        internal WasmFunctionType GetWasmFunctionTypeForMethod(MethodSignature signature, bool hasHiddenParam)
+        {
+            WasmValueType wasmPointerType = _nodeFactory.Target.PointerSize == 4 ? WasmValueType.I32 : WasmValueType.I64;
+
+            WasmValueType GetWasmTypeForTypeDesc(TypeDesc type)
```
kg, line 55: Just curious, has struct scalarization (from the WASM C ABI) been performed at this point? Or do you have to do it?

SingleAccretion: Yes, it has - all of the callers of this function first check `GetPrimitiveTypeForTrivialWasmStruct`. So here we only expect primitives.

yowl: Is this related to multi value return types or does LLVM never produce those?

SingleAccretion: We don't support the multi-value calling convention. It is not standardized yet either.

yowl: https://webassembly.org/features/  suggests it is standardized, but if we don't support it then no problem.

yowl (approved):
This looks good to me.  I'm not familiar with the object file layout, so I'm going by the fact that as it works, it is correct.  Do we have any before/after numbers for total wasm size ?

SingleAccretion:
> Do we have any before/after numbers for total wasm size ?

Yes - this is a no-diff change, as expected.

jkotas:
Thank you!

PR #2645, opened 2024-08-04
[NativeAOT-LLVM] Port (most of) compiler-generated helpers to WASM

SingleAccretion:
Continuation of https://github.com/dotnet/runtimelab/pull/2643.

The handwritten helpers offer a minor (-1K code size on HelloWasm) CQ improvement, mostly because LLVM is quite aggressive in inlining `i32.const` relocations, which are 5 bytes large without compression.

The accessor nodes will require some additions to the object writer and so have been left for the next change. After that is done, interop added in https://github.com/dotnet/runtimelab/pull/2615/files can also be deleted.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Quite a big change!  We still have some LLVMSharp after this change I suppose?

SingleAccretion:
> Quite a big change! We still have some LLVMSharp after this change I suppose?

Just that little bit :). After the extern method accessor nodes are rewritten in the same way, all LLVM interop can be deleted from ILC.

jkotas:
Thank you!

PR #2650, opened 2024-08-09
Integrate TypeSpec into TypeDecls and ABI parsing.

stephen-hawley:
Added the TypeSpecParser and associated support to the code to be able to parse in Swift type names.

Renamed members that are in C# space to have a `CS` prefix.
Named members that are in Swift space to have a `Swift` prefix.
Added a ton of unit tests.

### DIFF
--- a/src/Swift.Bindings/src/Model/ArgumentDecl.cs
+++ b/src/Swift.Bindings/src/Model/ArgumentDecl.cs
@@ -11,7 +11,12 @@ namespace BindingsGeneration
         /// <summary>
         /// Type of the argument.
         /// </summary>
-        public required TypeDecl TypeIdentifier { get; set; }
+        public required TypeDecl CSTypeIdentifier { get; set; }
+
+        /// <summary>
+        /// Type of the argument
+        ///
+        public required TypeSpec SwiftTypeSpec { get; set; }
         
         /// <summary>
         /// The private name of the argument.
--- a/src/Swift.Bindings/src/Model/MethodDecl.cs
+++ b/src/Swift.Bindings/src/Model/MethodDecl.cs
@@ -28,8 +28,6 @@ namespace BindingsGeneration
         /// Signature of the method.
         /// </summary>
         public required List<ArgumentDecl> CSSignature { get; set; }
-
-        public required List<TypeSpec> SwiftSignature { get; set; }
     }
 
     /// <summary>
--- a/src/Swift.Bindings/src/Model/MethodDecl.cs
+++ b/src/Swift.Bindings/src/Model/MethodDecl.cs
@@ -27,7 +27,9 @@ namespace BindingsGeneration
         /// <summary>
         /// Signature of the method.
         /// </summary>
-        public required List<ArgumentDecl> Signature { get; set; }
+        public required List<ArgumentDecl> CSSignature { get; set; }
+
+        public required List<TypeSpec> SwiftSignature { get; set; }
     }
 
     /// <summary>
--- a/src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
+++ b/src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
@@ -283,7 +283,7 @@ public class TypeSpecParser {
     /// Given an existing TypeSpec type, create a new NamedTypeSpec with name
     /// and type as the generic parameter, e.g. name&lt;type&rt;
     /// </summary>
-    TypeSpec WrapAsBoundGeneric(TypeSpec type, string name)
+    NamedTypeSpec WrapAsBoundGeneric(TypeSpec type, string name)
     {
         var result = new NamedTypeSpec(name);
         result.GenericParameters.Add(type);
--- a/src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
+++ b/src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
@@ -1,0 +1,358 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.IO;
+using System.Collections.Generic;
+
+
+namespace BindingsGeneration;
+
+/// <summary>
+/// A class to parse swift type specifications into the TypeSpec type
+/// The TypeSpec language is a little language to represent a swift type.
+/// It will lool like:
+/// typespec: [attribute]*[inout][any][typelable]typespec[generic-clause]['?']
+/// typespec: tuple-typespec|named-typespec|func-typespec|protocol-list-type-spec
+/// tuple-typespec:'('typespec-list')'
+/// named-typespec: name
+/// func-type-spec: ypespec [async][throws]t '->' typespec
+/// protocol-list-type-spec: named-typespec &amp; named-typespec [&amp; named-type-spec]*
+/// generic-clause: '<' typespec-list '>'
+/// typespeclist: typespec [',' typespec]*
+/// attribute @name['(' parameters ')']
+/// </summary>
+public class TypeSpecParser {
+    TextReader reader;
+    TypeSpecTokenizer tokenizer;
+
+    /// <summary>
+    /// Private constructor for a TypeSpecParser using the supplied TextReader as a source
+    /// </summary>
+    TypeSpecParser(TextReader reader)
+    {
+        this.reader = reader;
+        tokenizer = new TypeSpecTokenizer(reader);
+    }
+
+    /// <summary>
+    /// Private utility method to parse and return a TypeSpec
+    /// </summary>
+    TypeSpec? Parse()
+    {
+        // This is a very simple recursive descent parser with 1 token look-ahead
+        // The typespec language is very simple and doesn't need anything more
+        var token = tokenizer.Peek();
+        TypeSpec? type = null;
+        List<TypeSpecAttribute> attrs = new ();
+        var inout = false;
+        var isAny = false;
+        string? typeLabel = null;
+        var throwsClosure = false;
+        var asyncClosure = false;
+        var expectClosure = false;
+
+        // Prefix
+
+        // parse any attributes
+        if (token.Kind == TypeTokenKind.At) {
+            attrs = ParseAttributes();
+            token = tokenizer.Peek();
+        }
+
+        // looks like it's inout, consume and continue
+        if (token.Kind == TypeTokenKind.TypeName && token.Value == "inout") {
+            inout = true;
+            tokenizer.Next ();
+            token = tokenizer.Peek ();
+        }
+
+        // any, consume and continute
+        if (token.Kind == TypeTokenKind.TypeName && token.Value == "any") {
+            isAny = true;
+            tokenizer.Next ();
+            token = tokenizer.Peek ();
+        }
+
+        // label, consume and continue
+        if (token.Kind == TypeTokenKind.TypeLabel) {
+            typeLabel = token.Value;
+            tokenizer.Next ();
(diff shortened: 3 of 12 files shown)


src/Swift.Bindings/tests/TypeSpecTests/TypeSpecTests.cs
```diff
@@ -171,4 +171,87 @@ public static void TestGenericChallenging()
...
+            new NamedTypeSpec ("Dfoo"),
+            new NamedTypeSpec ("Cfoo"),
+            new NamedTypeSpec ("Efoo"),
+            new NamedTypeSpec ("Bfoo")
+        };
+
+        var protos1 = new ProtocolListTypeSpec (specs1);
+        var protos2 = new ProtocolListTypeSpec (specs2);
+
+        Assert.False (protos1.Equals (protos2));
```
matouskozak: Curious, why should protos1 and protos2 be equal?

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecTokenizer.cs
```diff
@@ -0,0 +1,238 @@
...
+            sb.Append (c);
+        }
+        for (char c = '['; c < '_'; c++) {
+            sb.Append (c);
+        }
+        sb.Append ('`');
+        for (char c = '{'; c <= (char)127; c++) {
+            sb.Append (c);
+        }
+        invalidNameChars = sb.ToString();
```
matouskozak: Would it make sense to rewrite this to use RegEx https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expressions instead?

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecTokenizer.cs
```diff
@@ -0,0 +1,238 @@
...
+            }
+            state = State.InName;
+            return null;
+        }
+    }
+
+    /// <summary>
+    /// Return true if and only if c is invalid for a name
+    /// <summary>
+    static bool InvalidNameCharacter (char c)
```
matouskozak: Not sure what is the expected naming convention here but maybe something like `IsInvalidNameCharacte` would be better in my eyes so that it doesn't visually mix up with `invalidNameChars` variable.

src/Swift.Bindings/src/Model/ArgumentDecl.cs
```diff
@@ -11,7 +11,12 @@ public record ArgumentDecl : BaseDecl {
         /// <summary>
         /// Type of the argument.
         /// </summary>
-        public required TypeDecl TypeIdentifier { get; set; }
+        public required TypeDecl CSTypeIdentifier { get; set; }
```
matouskozak, line 14: Should this be named CSTypeDecl instead of CSTypeIdentifier?

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
matouskozak, line 1: I wonder if there is a standardize way how to do the tokenizer/parser in C#/.NET? I haven't written that much C# code myself but it would be nice knowing that we are following the "correct" practises for this. I can check out if I can find any popular examples that we could use to validate this.

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
```diff
@@ -0,0 +1,358 @@
...
+
+        if (token.Kind == TypeTokenKind.LeftParenthesis) { // tuple
+            tokenizer.Next ();
+            TupleTypeSpec tuple = ParseTuple ();
+            type = tuple.Elements.Count == 1 ? tuple.Elements [0] : tuple;
+            typeLabel = type.TypeLabel;
+            type.TypeLabel = null;
+        } else if (token.Kind == TypeTokenKind.TypeName) { // name
+            tokenizer.Next ();
+            var tokenValue = token.Value.StartsWith ("ObjectiveC.", StringComparison.Ordinal) ?
```
matouskozak, line 94: I'm unsure whether this is the right place where we should handle Swift/ObjectiveC declarations. I would expect Parser just to return next token (`TypeSpec` in this case) and don't handle prepending `Foundation` lib.

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
```diff
@@ -0,0 +1,358 @@
...
+
+        if (tokenizer.Peek().Kind == TypeTokenKind.Arrow) {
+            tokenizer.Next();
+            type = ParseClosure(type, throwsClosure, asyncClosure);
+            expectClosure = false;
+            throwsClosure = false;
+            asyncClosure = false;
+        } else if (expectClosure) {
+            var errorCase = asyncClosure && throwsClosure ? "'async throws'" : asyncClosure ? "'async'" : "'throws'";
+            throw new Exception($"Unexpected token {tokenizer.Peek ().Value} after {errorCase} in a closure.");
```
matouskozak, line 127: I'm not sure if I missed it but we could also add test cases for the `throw` branches of the parser so that we are sure the errors are caught correctly.

matouskozak (commented):
Overall looking good to me, good job! I left a few comments with questions and suggestions. 

Also, with this kind of support already implemented, what kind of bindings is the tooling able to generate now? Are we able to have an e2e test case for with the current state? Having an example of e2e support this PR brings would help me to better understand the scope of this PR changes.

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
stephen-hawley: In this case, this was a bug/feature from the Swift compiler in that it was emitting ObjC types that should have been Foundation types.

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
stephen-hawley: That's a great question and the answer usually falls to using a generalized solution generated by a scanner/parser generator or writing your own. A typical scanner (aka tokenizer) is usually built around a deterministic finite state automaton (FSA), which is exactly what I did. Since there are only three states and most of the tokens are single character tokens, it made sense to do this by had since generated scanners tend to bloat in this case.

In terms of parsing, this is a very similar trade-off. Since parser generators are typically built for compiling larger languages, a lot of the support code is built around error handling, error recovery and handing tricky things like left recursion in production rules. Again, this is a very small language and the things it parses are going to be strings on the order of 50 characters (ie, 5 to 500) not multiline files with complex languages. It's not necessary to do so much work.

In terms of the actual structure, you're seeing a very common pattern in parsing where there is source of tokens with the ability to peek ahead one or more (in our case 1) and get the next. This is a very common pattern. Beyond that, are there are no standard or correct practices for this other than what I've done.

src/Swift.Bindings/src/Model/ArgumentDecl.cs
stephen-hawley: Maybe. Ask @kotlarmilos - I was following his naming.

src/Swift.Bindings/tests/TypeSpecTests/TypeSpecTests.cs
stephen-hawley: They aren't (hence `Assert.False`). Will change to an `Assert.NotEqual`

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecTokenizer.cs
stephen-hawley: Maybe? This is pretty quick though, although I will look at replacing with a set so that the time to access is constant.

src/Swift.Bindings/tests/TypeSpecTests/TypeSpecTests.cs
matouskozak: Ah, right. Sorry, my bad. I've misread it completely. Changing to `Assert.NotEqual` sounds good. Thank you.

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecParser.cs
matouskozak: Interesting, I haven't known about this. Does it mean that e.g., instead of `Foundation.NSString` it would be `ObjectiveC.NSString`, right? 

You mentioned it is a bug/feature in Swift compiler. Do you know if it is still valid and we need this?

src/Swift.Bindings/src/Model/ArgumentDecl.cs
kotlarmilos: We can rename it.

src/Swift.Bindings/src/Model/ArgumentDecl.cs
```diff
@@ -11,7 +11,12 @@ public record ArgumentDecl : BaseDecl {
...
+
+        /// <summary>
+        /// Type of the argument
+        ///
+        public required TypeSpec SwiftTypeSpec { get; set; }
```
kotlarmilos, line 19: Let's align the naming to be more descriptive.

src/Swift.Bindings/src/Model/MethodDecl.cs
```diff
@@ -27,7 +27,9 @@ public sealed record MethodDecl : BaseDecl
         /// <summary>
         /// Signature of the method.
         /// </summary>
-        public required List<ArgumentDecl> Signature { get; set; }
+        public required List<ArgumentDecl> CSSignature { get; set; }
+
+        public required List<TypeSpec> SwiftSignature { get; set; }
```
kotlarmilos: Why does the list item type differ here?

kotlarmilos (commented):
Thanks, Steve. I would also like to see E2E tests as they will strengthen our path forward. I have a feeling that while we are adding functional tests, we lack clarity on their usability. For example, what are the current capabilities of the tooling, including the recent changes?

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecTokenizer.cs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /// <summary>
+    /// Return true if and only if c is invalid for a name
+    /// <summary>
+    static bool IsInvalidNameCharacter (char c)
+    {
+        return invalidNameChars.Contains(c);
+    }
+}
```
jkurdek, line 236: NIT: I noticed that some files end with a newline. And some don't. Do we have any preferences for style to follow in future?

jkurdek (commented):
In my opinion both the code and the tests look good. As discussed on a meeting I do realise that we still miss some pieces which would enable us to write a e2e test. 

I'd like to suggest adding more detail to our PR descriptions and documentation. For those of us who are still climbing the learning curve (me 😄  ), it would be very helpful to understand the reason behind certain decisions, such as the need for TypeSpecs in the first place.

src/Swift.Bindings/src/Model/TypeSpecParsing/TypeSpecTokenizer.cs
stephen-hawley: Is it important since the tooling accepts it either way?

kotlarmilos (approved):
Thanks, Steve, and sorry for the delay!

The changes look good to me.

I would like to see more end-to-end tests in following work. Let's work together on them.

PR #2651, opened 2024-08-11
[NativeAOT-LLVM] Set the runtime flavor for 'nativeaot' subsets to CoreCLR

SingleAccretion:
The before-last merge introduced a problem that @dicej reported a while ago, where things like `./build nativeaot.build` would pick Mono as the default runtime flavor for `$(TargetsMobile) == true`. This change fixes it.


SingleAccretion:
@dotnet/nativeaot-llvm

maraf (approved):

PR #2656, opened 2024-08-12
[NativeAOT-LLVM] Remove the Align8 workaround

### LINKED ISSUE #2606: [NativeAOT-LLVM] [Question]: Should GCPointerMap have a provision for adding optional fields to GCStaticEETypeNode

SingleAccretion:
Fixes #2606.

Upstream fix: https://github.com/dotnet/runtime/pull/105931.


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas (approved):
Thanks

PR #2614, opened 2024-06-13
[NativeAOT-LLVM] support `System.Net.Http.HttpClient` on WASIp2

dicej:
This adds `WasiHttpHandler`, a new implementation of `HttpMessageHandler` based on
[wasi:http/outgoing-handler](https://github.com/WebAssembly/wasi-http/blob/v0.2.0/wit/handler.wit), plus tweaks to `System.Threading` to allow async `Task`s to work in a single-threaded context, with `ThreadPool` work items dispatched from an application-provided event loop.

WASIp2 supports asynchronous I/O and timers via `wasi:io/poll/pollable` resource handles.  One or more of those handles may be passed to `wasi:io/poll/poll`, which will block until at least one of them is ready.  In order to make this model play nice with C#'s `async`/`await` and `Task` features, we need to reconcile several constraints:

- WASI is currently single-threaded, and will continue to be that way for a while.
- C#'s `async`/`await` and `Task` features require a working `ThreadPool` implementation capable of deferring work.
- A WASI component can export an arbitrary number of functions to the host, and though they will always be called synchronously from the host, they need to be able to perform asynchronous operations before returning.
- WASIp3 (currently in the design and prototype phase) will support asynchronous exports, with the top level event loop running in the host instead of the guest, and `wasi:io/poll/pollable` will no longer exist.  Therefore, we don't want to add any temporary public APIs to the .NET runtime which will become obsolete when WASIp3 arrives.

The solution we arrived at looks something like this:

- Tweak the existing `ThreadPool` implementation for WASI so that methods such as `RequestWorkerThread` don't throw `PlatformNotSupportedException`s (i.e. allow work items to be queued even though the "worker thread" is always the same one that is queuing the work)
- Add two new methods to `Thread`:
    - `internal static void Dispatch`: Runs an iteration of the event loop, draining the `ThreadPool` queue of ready work items and calling `wasi:io/poll/poll` with any accumulated `pollable` handles
    - `internal static Task Register(int pollableHandle)`: Registers the specified `pollable` handle to be `poll`ed during the current or next call to `Dispatch`
    - Note that these methods are `internal` because they're temporary and should not be part of the public API, but they are intended to be called via `UnsafeAccessor` by application code (or more precisely, code generated by `wit-bindgen` for the application)

[...]

PR #2667, opened 2024-08-23
[NativeAOT-LLVM] disable SocketsHttpHandler support for WASI

bacongobbler:
Disabling this until `System.Net.Sockets` support comes to WASI via [wasi-sockets](https://github.com/WebAssembly/wasi-sockets). Many libraries (such as [grpc-dotnet](https://github.com/grpc/grpc-dotnet)) check the value of `SocketsHttpHandler.IsSupported` to see if it [needs to fall back](https://github.com/grpc/grpc-dotnet/blob/e9cc7e15796d39f1d2656178f56a45c09147d0fe/src/Shared/HttpHandlerFactory.cs#L32) from a sockets-based handler to a basic HTTP handler.

If this is not the correct location to update, let me know and I'd be happy to amend the PR.


bacongobbler:
related to #2614.

yowl (approved):
Thanks!  This is also in #2614 now, but seems reasonable to me that it can go independently here.  CI failure is known (and should be green if this is rerun, e.g. if it is closed and reopened)

jkotas (approved):

PR #2672, opened 2024-08-27
[NativeAOT-LLVM] Move thread stack top and bottom from thread locals to `thread` class

yowl:
This PR removes the thread locals for the shadow stack top and bottom and replaces them with members in the `thread` class.  This opens the opportunity for one thread to initiate the scan of the shadow stack of all threads unblocking one thing to make NAOT-LLVM runtime for Wasm multi threaded.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/thread.cpp
+++ b/src/coreclr/nativeaot/Runtime/thread.cpp
@@ -35,6 +35,29 @@ static Thread* g_RuntimeInitializingThread;
 
 #endif //!DACCESS_COMPILE
 
+#ifdef HOST_WASM
+void* Thread::GetShadowStackBottom()
+{
+    return m_pShadowStackBottom;
+}
+
+void Thread::SetShadowStackBottom(void *pShadowStack)
+{
+    ASSERT(m_pShadowStackBottom == nullptr);
+    m_pShadowStackBottom = pShadowStack;
+}
+
+void* Thread::GetShadowStackTop()
+{
+    return m_pShadowStackTop;
+}
+
+void Thread::SetShadowStackTop(void* pShadowStack)
+{
+    m_pShadowStackTop = pShadowStack;
+}
+#endif
+
 PInvokeTransitionFrame* Thread::GetTransitionFrame()
 {
     if (ThreadStore::GetSuspendingThread() == this)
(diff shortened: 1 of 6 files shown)


yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/nativeaot/Runtime/thread.h
```diff
@@ -162,6 +166,11 @@ class Thread : private RuntimeThreadLocals
                                                     // On Unix this is an optimization to not queue up more signals when one is
                                                     // still being processed.
     };
+#ifdef HOST_WASM
+    PTR_VOID GetShadowStackTop();
+    void SetShadowStackBottom(PTR_VOID pShadowStack);
+    void SetShadowStackTop(PTR_VOID pShadowStack);
+#endif
```
SingleAccretion: I think it would make sense to move these declarations a bit further down (e. g. after the group `InlinePInvokeReturn` is in).

Also, I would get rid of the DACy `PTR_VOID`s and just use `void*`. We're not going to have a DAC for WASM (or NAOT for that matter).

src/coreclr/nativeaot/Runtime/thread.h
```diff
@@ -196,6 +205,7 @@ class Thread : private RuntimeThreadLocals
     PInvokeTransitionFrame* GetTransitionFrame();
 
 #ifdef HOST_WASM
+    PTR_VOID GetShadowStackBottom();
```
SingleAccretion: Delete? `GcScanWasmShadowStack` can use `m_pShadowStackBottom` directly.

src/coreclr/nativeaot/Runtime/thread.cpp
```diff
@@ -35,6 +35,29 @@ static Thread* g_RuntimeInitializingThread;
 
 #endif //!DACCESS_COMPILE
 
+#ifdef HOST_WASM
+void* Thread::GetShadowStackBottom()
+{
+    return m_pShadowStackBottom;
+}
```
SingleAccretion: It's pretty important that these get inlined and since we don't have LTO, they should be moved to `thread.inl`.

yowl: ok, made them `FORCEINLINE`

yowl:
I haven't put the shadow stack top in the transition frame here, should I do that?

SingleAccretion:
> I haven't put the shadow stack top in the transition frame here, should I do that?

It can be a separate change.

A scheme I've been imagining is allocating the frame itself on the shadow stack s. t. the transition is as minimal as possible.

SingleAccretion (approved):
LGTM, thank you!

PR #2673, opened 2024-08-27
[NativeAOT-LLVM] Enable the official build for Linux

SingleAccretion:
This requires reconfiguring how we build packages a little. See the first commit.

Also, a little bit of YAML cleanup has been snuck in.


SingleAccretion:
@dotnet/nativeaot-llvm

PR #2675, opened 2024-08-28
[NativeAOT-LLVM] Update documentation

SingleAccretion:
To reflect the fact Linux x64 is now a supported platform for compiling on.

@dotnet/nativeaot-llvm


yowl:
There's a few other places where we refer to windows, e.g. https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/docs/workflow/building/coreclr/nativeaot.md?plain=1#L50  Should we prefix those, "e.g. on windows, " or similar?

SingleAccretion:
>  Should we prefix those, "e.g. on windows, " or similar?

No reason not to. Reworked that part a bit.

yowl (approved):
Thanks!

PR #2677, opened 2024-09-04
[NativeAOT-LLVM] Initial version of the managed stack traces on Browser

SingleAccretion:
We use the JS stack trace API to get a list of frames, either WASM or JS, and use this data to construct the various managed objects representing frames. The JS stack trace format provides us with the WASM function index, which we also store in the stack trace metadata, as our "IP". The result is full fidelity support for managed stack at no additional size cost.

The main execution cost is that the JS stack trace API only supports snapshotting the whole stack, which forces us to both spend more cycles parsing this trace and allocating more memory for it.

Missing features:
1) Safari/WebKit/JSC support (https://bugs.webkit.org/show_bug.cgi?id=278991). JSC will fall back to using the JS frames.
2) Proper stack traces inside funclets (both filter and finally).
3) Method name display for native methods. We could easily do this if we were ok with relying on the implementation detail of how wasm-ld lays out functions, but that's (obviously) not something we can really do.

<details>
<summary>What implementation detail?</summary>

Wasm-ld lays out code in the exact same order as specified on the command line for the passed-in object files (important note: only actual object files, not LTO bitcode). So we could surround the ILC-produced files by marker objects that define one function. With that, determining whether a given WASM function index is "managed" becomes as simple as a range check.
</details>

Contributes to https://github.com/dotnet/runtimelab/issues/2404.


SingleAccretion:
@dotnet/nativeaot-llvm

pavelsavara:
FYI @kg

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -546,21 +546,24 @@ The .NET Foundation licenses this file to you under the MIT license.
     </ItemGroup>
 
     <ItemGroup>
+      <_WasiComponentImports Include="$(IlcFrameworkNativePath)*.wit" />
```
yowl, line 549: Is this required for this PR, or just general goodness?

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
```diff
@@ -310,6 +313,9 @@ private static void RhpPopUnwoundVirtualFrames()
         [RuntimeExport("RhpHandleUnhandledException")]
         private static void HandleUnhandledException(object exception)
         {
+            // Append JS frames to this unhandled exception for better diagnostics.
```
yowl, line 316: What happens in Wasi here, or does it never get here?

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
```diff
@@ -360,71 +373,62 @@ private unsafe struct EHClause
...
-            private readonly nuint _format;
+            private readonly byte* _pEHTable;
+            private readonly bool _isLargeFormat;
 
             public EHTable(void* pUnwindTable)
             {
-                _pEHTable = (void*)((nuint)pUnwindTable & ~FormatMask);
-                _format = (nuint)pUnwindTable & FormatMask;
+                _pEHTable = (byte*)pUnwindTable;
+                _isLargeFormat = (*(byte*)pUnwindTable & MetadataLargeFormat) != 0;
```
yowl, line 386: I suppose using `_pEHTable` would remove a cast?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.NativeAot.Browser.cs
```diff
@@ -1,65 +1,93 @@
...
+        /// </summary>
+        private void InitializeForIpAddressArray(IntPtr[] eips, int skippedEipCount, int skipFrames, bool needFileInfo)
         {
-            var backtraceBuffer = new byte[8192];
-            int callstackLen;
-            // skip these 2:
-            // at S_P_CoreLib_System_Diagnostics_StackTrace__InitializeForCurrentThread (wasm-function[12314]:275)
-            // at S_P_CoreLib_System_Diagnostics_StackTrace___ctor_0(wasm-function[12724]:118)
-            skipFrames += 2; // METHODS_TO_SKIP is a constant so just change here
+            // Our callers may pass us values that have overflown.
```
yowl, line 42: when does it overflow, not sure I have understood how that happens?

yowl (commented):
Impressive piece of work!

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: The latter. I tried changing a few things while chasing down the bug that would end up being references below (https://github.com/dotnet/msbuild/issues/5937).

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
SingleAccretion: This eventually calls down into the WASI version of `AppendStack`, which is just a no-op for now.

Edit: in the eventual work that will bring stack traces to WASI, it will be just filtered out with a null IP check, or I'll ifdef it out for browser only. Not sure yet what will look better.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Diagnostics/StackTrace.NativeAot.Browser.cs
SingleAccretion: It turns out that it is legal to call `new StackTrace(skipFrames: int.MinValue)` and such, and there are even tests verifying the behavior of that. Such values are probably the result of overflow.

Also, the initial version of this code just did `skipFrames += SystemDiagnosticsStackDepth;` (like in `StackFrame.NativeAot.Browser.cs`) instead of `SkipSystemFrames`, which would overflow for `new StackTrace(skipFrames: int.MaxValue)`.

(Q: should the stack frame logic should also use `SkipSystemFrames`? A: probably not, since it would mean that in case of JSC or modified binaries it would start yielding system frames only. Which is not too bad for `new StackTrace().ToString()`, but makes `new StackFrame()` basically useless.)

src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/ExceptionHandling.wasm.cs
SingleAccretion: Yes, but at the cost of a field load (which is likely to get removed, but I like 'efficiency by construction' in system code).

yowl (approved):
Thanks!

PR #2678, opened 2024-09-04
Added documentation for binding most of the Swift types

stephen-hawley:
This PR consists of documentation for how to bind the major Swift types in C#, calling out issues that we may run into and possible ways around them.

At present, this is missing the documentation for protocols with associated types, actors, and closures. This PR was taking a long time to write so I felt it best to get what I have under scrutiny.

Further, protocols with associated types will take some time to explain.


src/docs/binding-overview.md
```diff
@@ -0,0 +1,105 @@
...
+- For value types, they allow more registers to be used for passing the type.
+- For instance methods and closures, there is a separate register reserved outside of the regular argument registers used for passing the instance pointer
+- For functions that throw, there is a separate register reserved for exceptions (errors)
+- For returning value types an extra register is available
+
+## Runtime Differences
+The main difference in the Swift runtime is that heap allocated types are reference counted rather than using another memory management scheme. The reference counting is not done with a single count but instead is managed with two counts: a strong count and a weak count. Weak references are available in Swift for making (potentially) circular data structures.
+
+## Idiomatic Differences
+For the most part, these are minor. C# prefers camel case starting with a capital letter for all types and entry points, using lowercase starts for private fields and argument names. C# adopts that conventions that interfaces start with 'I'. Swift starts types with captials but public variables will start with lowercase.
```
rolfbjarne:
```suggestion
For the most part, these are minor. C# prefers camel case starting with a capital letter for all types and entry points, using lowercase starts for private fields and argument names. C# adopts that conventions that interfaces start with 'I'. Swift starts types with capitals but public variables will start with lowercase.
```

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+
+Classes function very similarly to C# classes with some important semantic and runtime differences which affect how they are exposed in C#, but generally speaking, they can exist as a 1:1 mapping. It should be noted that while Swift provides the ability to create traditional OO code, Apple is eschewing the use of classes in their own code and preferring value types instead. It is telling that in the main Swift runtime library, libswiftCore, there are no exported classes.
+
+Swift classes can be either native Swift or Objective-C. The latter is in place strictly for compatability with legacy code and in fact, if a Swift class is declared to be `@ObjC`, it must inherit from `NSObject` and the swift compiler will generate a selector dispatcher for all methods marked as `@ObjC` but these just go to the Swift implementation, therefore Objective-C compatability is just a thin veneer on top of the Swift implementation.
+
+## Language Parity Mismatches
+Swift provides a different set of accessibility of methods that C#:
+- open - callable by all and overridable (effectively public virtual)
+- public - callable by all not overridable
+- private - callable only by members of the class
```
kotlarmilos, line 11: I don't think private members are even exposed in the ABI file.

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+// original swift class
+open class YourAge {
+    public init () { }
+    open func getAge() -> Int {
+        return 7
+    }
+}
+
+// simulated vtable:
+fileprivate struct YourAge_VTable {
```
kotlarmilos, line 69: Why do we need the vtable? Can't we just expose public members?

src/docs/binding-enums.md
```diff
@@ -0,0 +1,195 @@
+# Binding Enums
+
+Swift enums are a hybrid value type between enumeraions (names associated with cardinal values) and discriminated unions. Swift enums can be broken down into three categories:
```
jkotas:
```suggestion
Swift enums are a hybrid value type between enumerations (names associated with cardinal values) and discriminated unions. Swift enums can be broken down into three categories:
```

src/docs/binding-classes.md
stephen-hawley: Recall that purpose of this is to make it so that Swift can call into C# for the actual implementation of the members. For Swift to get the function pointer, C# must first call into Swift to set them. For n virtual members, we could either (1) make n calls to set the function pointers, (2) 1 call with n arguments, or (3) 1 call taking 1 argument (a vtable). While 2 and 3 are more or less equivalent, it won't work in the case of generic classes with virtual members, for which each class has to keep a dictionary of bound generic types mapped to functions implementing the virtual methods.
Because of efficiency and uniformity, having a vtable (per unique generic binding) makes the most sense.

src/docs/binding-classes.md
stephen-hawley: I was worried about structs and just checked this example:
```swift
@frozen
public struct Foo {
    private var name: String
    public init(to: String) {
        name = to
    }
    public func Ima() -> String {
        return "I am a(n) " + name
    }
}
```
and the good news is that private members are present in the abi.json file:
```json
{
  "ABIRoot": {
    "kind": "Root",
    "name": "TopLevel",
    "printedName": "TopLevel",
    "children": [
      {
        "kind": "TypeDecl",
        "name": "Foo",
        "printedName": "Foo",
        "children": [
          {
            "kind": "Var",
            "name": "name",
            "printedName": "name",
            "children": [
              {
                "kind": "TypeNominal",
                "name": "String",
                "printedName": "Swift.String",
                "usr": "s:SS"
              }
            ],
            "declKind": "Var",
            "usr": "s:20GenericTestFramework3FooV4name33_30D99E5814F18B95AFD2006B0E3423D7LLSSvp",
            "mangledName": "$s20GenericTestFramework3FooV4name33_30D99E5814F18B95AFD2006B0E3423D7LLSSvp",
            "moduleName": "GenericTestFramework",
            "isInternal": true,
            "declAttributes": [
              "HasStorage",
              "AccessControl"
            ],
            "fixedbinaryorder": 0,
            "hasStorage": true
          },
     ... extra removed
```

src/docs/binding-overview.md
```diff
@@ -0,0 +1,105 @@
...
+- ABI (application binary interface) mismatches
+- Runtime differences
+- Idiomatic differences
+- Accessibility
+
+## Language Parity Mismatches
+Swift is a hybrid programming language leaning towards functional. It has all the features that you would expect in a traditional object oriented programming language, but from Apple's point of view, these are not features that are encouraged. Instead, Apple is encouraging using value types for most everything using copy-on-write semantics for the collection types. In fact, the main runtime for swift, swiftCore, contains 0 published classes as of a few years ago. In addition, Apple encourages "aspect oriented" rather than object oriented coding.
+
+In addition, Swift has a number of language features that do not exist in C#:
+- Discriminated Unions - Swift enums are build so that any case in the set may have payloads and the payloads need not be the same type
```
rolfbjarne:
```suggestion
- Discriminated Unions - Swift enums are built so that any case in the set may have payloads and the payloads need not be the same type.
```

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+- Init methods get called with the type metadata in the first argument.
+- Methods that throw return the error in the error register.
+- Virtual methods in swift get called via a vtable. Since the ordering of the vtable is undefined, methods can't be called using the vtable from C#. In addition, the vtable is consider fragile and may change between versions of any given library. Fortunately, with enable-library-evolution set, the compiler will inject a dispatch thunk for each virtual method.
+
+
+## Runtime Differences
+Swift classes are reference counted rather than garbage collected. This can create some interesting conditions when an object from Swift surfaced in C# gets garbage collected, but the object may still be live in Swift, so it's necessary to manage this. The way this is handled in Binding Tools for Swift is to add it to an object registry which is a `Dictionary&lt;NativeHandle, GCHandle&gt;` The native handle is the Swift object instance and the GC Handle is made from the corresponding C# instance. When the pair is added to the registry, the registry also takes a weak reference to the Swift
```
rolfbjarne:
```suggestion
Swift classes are reference counted rather than garbage collected. This can create some interesting conditions when an object from Swift surfaced in C# gets garbage collected, but the object may still be live in Swift, so it's necessary to manage this. The way this is handled in Binding Tools for Swift is to add it to an object registry which is a `Dictionary<NativeHandle, GCHandle>` The native handle is the Swift object instance and the GC Handle is made from the corresponding C# instance. When the pair is added to the registry, the registry also takes a weak reference to the Swift
```

src/docs/binding-functions.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+When compiled, this function actual takes 3 arguments, not two:
+```swift
+public func Identity(a *T, type: T.Self, ProtoWitness(T.Self, SomeProtocol.Self)) -> T { }
+```
+
+## Idiomatic Differences
+Functions in swift start with a lowercase letter. We should auto-capitalize the first letter of a function name.
+Since property getters and setters map naturally.
```
rolfbjarne: Seems like there's a missing second part to this sentence?

src/docs/binding-structs.md
```diff
@@ -0,0 +1,100 @@
...
+
+Blittable structs are structs that can be correctly copied with `memcpy`. Essentially, if a struct contains a field which is heap allocated or contains a non-blittable data type, it will have issues with regards to reference counting. In addition, generic structs need to be considered non-blittable since they *may* contain non-blittable types, but we can't know that ahead of time.
+
+Structs in Swift are encouraged to be immutable. To help along with this, the compiler enforces a number of rules to prevent you from trying to mutate an immutable value. All arguments to a method are considered to be `let` bindings. The compiler won't let you modify public fields of struct that is defined with a `let` binding. Similarly, you cannot call a `mutating` method on a struct that is bound with `let`.
+
+Every value type in swift comes with a compiler generate Value Witness Table. Unlike other function tables in Swift (Protocol witness tables, vtables etc), the layout of value witness tables are well-defined. The value witness table is documented [here](binding-value-witness-table.md).
```
rolfbjarne:
```suggestion
Every value type in swift comes with a compiler generated Value Witness Table. Unlike other function tables in Swift (Protocol witness tables, vtables etc), the layout of value witness tables are well-defined. The value witness table is documented [here](binding-value-witness-table.md).
```

src/docs/binding-structs.md
```diff
@@ -0,0 +1,100 @@
...
+
+// ...
+// Consuming code, which seems totally reasonable for C#
+var n1 = new NamedHolder(SwiftString.FromString("Corey"));
+var n2 = new NamedHolder(SwiftString.FromString("Ian));
+n1 = n2; // memory leak.
+```
+
+## Accessibility
+The main problem that we have is with non-blittable structs. We would either need our users to manually destroy structs when they're no longer needed (this a really bad idea - people are awful at memory management - that's why we have garbage collection and )
```
rolfbjarne:
```suggestion
The main problem that we have is with non-blittable structs. We would either need our users to manually destroy structs when they're no longer needed (this a really bad idea - people are awful at memory management - that's why we have garbage collection and )
```

and what?

src/docs/binding-tuples.md
```diff
@@ -0,0 +1,60 @@
...
+```swift
+public func foo(T: (a: Int, b: Bool, (c: SomeStruct, d: SomeClass))) { // func of 1 argument
+    // ...
+}
+
+public func foo(a: Int, b: Bool, c: SomeStruct, d: SomeClass) { // func of 4 arguments
+    // ...
+}
+```
+It's unclear why apply would do this except that it might be a vestige of when Swift had the ability to have multiple
```
rolfbjarne:
```suggestion
It's unclear why Apple would do this except that it might be a vestige of when Swift had the ability to have multiple
```

rolfbjarne (approved):

kotlarmilos (approved):

dalexsoto (approved):

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+- private - callable only by members of the class
+- internal (default) - callable only by members of the module
+- fileprivate - callable only members defined in the same file
+
+For the most part, we only care about public and open as these two represent the public facing API of the class.
+There is no `protected` (private but for inheritors) access which will create an issue for exposing C# types to Swift.
+
+Swift has very specific rules for initialization. Under the hood, Swift breaks up the process of making an instance of a class into two pieces: allocation and initialization. Early on these were very much separate, but in the current swift runtime, the allocator will also call the initializer.
+
+The allocating initializer takes the type metadata for the class in the `self` register. We can call this directly, however under enable-library-evolution, the swift compiler generates a dispatch thunk for this method and we should probably call that.
```
AaronRobinsonMSFT: "enable-library-evolution" - Link to documentation on this concept?

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+
+Swift divides initializers into two categories: designated and convenience. Designated initializers are initializers which fully initialize a class. If a subclass inherits from base class, it must implement all designated initializers and call the parent class. A convenience intializer will have a different signature than all desinated initializers and **must** call a designated initializer.
```
AaronRobinsonMSFT:
```suggestion
Swift divides initializers into two categories: designated and convenience. Designated initializers are initializers which fully initialize a class. If a subclass inherits from base class, it must implement all designated initializers and call the parent class. A convenience intializer will have a different signature than all designated initializers and **must** call a designated initializer.
```

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+
+The generic programming model in Swift does specialization at runtime. It does this by creating the type metadata for the specialized type
+using the Metadata Accessor function. The Metadata Accessor function has one type metadata argument for each specialized type in the generic class. It is **very important** that the binding code calls the Metadata Accessor rather than trying to synthesize the type metadata object. This is because from Swift's point of view, every type metadata object is a singleton and the runtime will cache generic specializations. 
+
+Swift initializers can fail - this is something that doesn't have an exact analog in C#. There are two ways to handle this:
+- make the corresponding C# constructor throw
+- write a factory method that returns `BoundType?`
+
+## ABI differences
```
AaronRobinsonMSFT, line 31: Do we have a primary document we can reference here?

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+- Init methods get called with the type metadata in the first argument.
+- Methods that throw return the error in the error register.
+- Virtual methods in swift get called via a vtable. Since the ordering of the vtable is undefined, methods can't be called using the vtable from C#. In addition, the vtable is consider fragile and may change between versions of any given library. Fortunately, with enable-library-evolution set, the compiler will inject a dispatch thunk for each virtual method.
+
+
+## Runtime Differences
+Swift classes are reference counted rather than garbage collected. This can create some interesting conditions when an object from Swift surfaced in C# gets garbage collected, but the object may still be live in Swift, so it's necessary to manage this. The way this is handled in Binding Tools for Swift is to add it to an object registry which is a `Dictionary<NativeHandle, GCHandle>` The native handle is the Swift object instance and the GC Handle is made from the corresponding C# instance. When the pair is added to the registry, the registry also takes a weak reference to the Swift
```
AaronRobinsonMSFT, line 41: Would a [`ConditionalWeakTable`](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2) be more appropriate here?

src/docs/binding-functions.md
```diff
@@ -0,0 +1,103 @@
...
+    public init (x: Int) {
+        self.x = x
+    }
+    public static prefix -+- (c: Container)
+    {
+        return Container(c.x + c.x)
+    } 
+}
+```
+We run into two problems: C# doesn't have an operator `-+-` nor can we define one. If implemented as a method, `-+-` is unpronounceable in C#. Taking the approach of using a prefix like "Op_" or "Oper_" or "Operator_" we can make it clear that this was an operator. Then the operator name would need to be run through a sanitizer to turn it into something pronounceable. The best we could hope for is getting a name out the other side like "Op_PlusMinusPlus".
```
AaronRobinsonMSFT, line 102: Perhaps a link to a C# issue? This seems like a concept that is going to get lost in this document rather than trackable. In this case linking to something on github helps avoid losing this.

src/docs/binding-overview.md
```diff
@@ -0,0 +1,105 @@
...
+- Swift has only 1 namespace, the module name, with no nesting.
+- Swift has top level entities, i.e., global functions and variables.
+- In Swift, all members in a type *must* be initialized. Similarly, unsafe pointer types have a notion of 4 possible states that they can be in:
+  - allocated
+  - initialized
+  - deinitialized
+  - deallocated
+Effectively, all types follow this pattern, but the 4 states are managed by the compiler and the runtime and unlike pointer types, there is no explicit control of it.
+
+## ABI Differences
```
AaronRobinsonMSFT, line 34: I would like to see some primary document references. There are a lot of intricate details described here and understanding their provenance will help us move forward with addressing/accepting them as Swift interop evolves.

src/docs/binding-overview.md
```diff
@@ -0,0 +1,105 @@
...
+For bit fields, C# prefers using enums with the `[Flags]` attribute. In Swift, the pattern is to use a struct that implements `OptionSet`.
+
+## Accessibility
+First and foremost, our customers matter the most. They are the ones who will be using the code and everything they use should "feel" like C# to them rather than Swift. As much as possible, this should be automated.
+
+That said, there will be hard choices that need to be made. For example, consider the following Swift function:
+```swift
+public func VowelCount(text: String) -> Int { /* ... */ }
+```
+`String` in Swift is a vastly different and completely incompatible type from a C# string. In C# a string is an array of UTF16 characters. In Swift, the layout of `String` is completely different. It is a union of several different possible representations. As such, it is always the case that when you want to make a Swift string accessible in C# you will need to allocate memory. Similarly, when you want a C# string accessible in Swift, you will need to allocate memory. Consequently, you have to care about the memory lifetime of the object.
```
AaronRobinsonMSFT, line 59: For this particular issue, I strong recommend looking at how COM interop handled these sorts of issues. It seems to me that adding similar APIs for this sort of interop is completely appropriate [`Marshal.GetComObjectData()`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.marshal.getcomobjectdata) and would make the ergonomics work nicely for both the user and how this could integrate with our custom marshallers. I could imagine a new `SwiftStringMarshaller` that handles all of this. Since C# `string` types are immutable this approach works, but I do see how other scenarios involving mutable types may not. Something to consider.

/cc @jkoritzinsky

src/docs/binding-structs.md
```diff
@@ -0,0 +1,100 @@
+# Binding Structs
+
+Structs in Swift are value types that can have a number of forms each of which requires a certain amount of consideration: frozen/non-frozen and blitable and non-blitable.
```
AaronRobinsonMSFT:
```suggestion
Structs in Swift are value types that can have a number of forms each of which requires a certain amount of consideration: frozen/non-frozen and blittable and non-blittable.
```

AaronRobinsonMSFT: You can add a reference to the .NET documentation - https://learn.microsoft.com/dotnet/framework/interop/blittable-and-non-blittable-types.

src/docs/binding-structs.md
```diff
@@ -0,0 +1,100 @@
...
+    let n2:NamedHolder = NamedHolder(name: "Ian")
+    n1 = n2
+}
+runIt()
+print("all done")
+```
+It will generate the following output:
+```
+you killed Corey.
+you killed Ian.
```
AaronRobinsonMSFT: I would avoid this sort of phrasing in public documentation. "Corey has left the building"?

src/docs/binding-overview.md
rolfbjarne: One thing to have in mind that this will run on platforms where both the trimmer and AOT are necessary. That makes any kind of dynamic implementation a lot trickier to implement.

AaronRobinsonMSFT: There is nothing dynamic about the type information though. This is merely associating already allocated typed memory with an already allocated type instance. All the types are statically known and so I'm unclear how this specific `string` 
 example would be difficult to enable for trimming. If we were storing some dynamic type, I get it, but I don't see how this approach would impact trimming.

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+Swift initializers can fail - this is something that doesn't have an exact analog in C#. There are two ways to handle this:
+- make the corresponding C# constructor throw
+- write a factory method that returns `BoundType?`
+
+## ABI differences
+- Instance methods get called with the `self` register pointing to the instance.
+- Class methods get called with the `self` register pointing to the type metadata.
+- Init methods get called with the type metadata in the first argument.
+- Methods that throw return the error in the error register.
+- Virtual methods in swift get called via a vtable. Since the ordering of the vtable is undefined, methods can't be called using the vtable from C#. In addition, the vtable is consider fragile and may change between versions of any given library. Fortunately, with enable-library-evolution set, the compiler will inject a dispatch thunk for each virtual method.
```
jkoritzinsky: Link to the documentation of the dispatch tunks and how they're exposed for use?

src/docs/binding-classes.md
jkoritzinsky: Would it be possible for the generated code to expose symbols/methods following the same conventions as the Swift language (ie using dispatch thunks instead of a hard-coded vtable) such that the regular Swift runtime logic for discovering the overloads in enable-library-evolution scenarios works automatically without us having to shim the code with our own vtable?

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
...
+public func someOtherFunc(y: YourAge) {
+    if let ssp = y as? SomeSwiftProtocol { // as called from C# this will never work since y is YourAgeWrapper
+        ssp.ProtocolAction()
+    }
+}
+```
+
+The way to get this to work 100% in both directions is to have a reverse binding step that exposes the new C# members and protocol compliance in Swift.
+
+It may seem tempting to synthesize the protocol witness table, a type object and a synthetic swift object, but this would require knowing the layout of the protocol witness tables as well as the layout of the actual vtables which we don't and can't know. In addition, both tables are fragile and susceptible to changes in the class and/or protocol(s).
```
jkoritzinsky: What information is available here in enable-library-evolution/stable ABI scenarios? We can limit ourselves to scenarios that can be exposed in stable ABI scenarios.

Even if we can't know this information at compile time, could we construct it together at runtime without pre-generating Swift code? Or is the structure even unknownable at runtime? (If so, how does Swift implement this in stable ABI scenarios?)

I'm not concerned if a library says a type is frozen and then makes a breaking change. That's out of scope for .NET/Swift interop.

src/docs/binding-enums.md
```diff
@@ -0,0 +1,195 @@
...
+
+1. Construction
+2. Discriminator Access
+3. Payload Access
+
+It makes sense to have an outboard C# enum that defines the possible cases and then a type to host the enum.
+
+In Binding tools for Swift, the type was implemented with a class implementing `IDisposable` with an opaque payload. This was done for the same reason as [structs](binding-structs.md), and because of that we need take into consideration whether or not the payloads are non-blittable.
+
+We should not, under any circumstances, try to synthesize Swift enums ourselves. The memory layout of enums is not only opaque, it is fragile and the tools available in the value witness table are undocumented.
```
jkoritzinsky, line 44: To be able to pass Swift enums by value we can use the documented rules for ABI layout (the layout rules for frozen enums are actually documented today, just with really obtuse wording in TypeLayout.rst). The necessary runtime work to support these layouts is tracked by https://github.com/dotnet/runtime/issues/100896.

In https://github.com/dotnet/runtimelab/pull/2678/files#diff-22ab1f4e1a9ae7455a15769655d746aa2390369e9c06f6eb0e1be56a33b5f7d4R13-R14, it seems that the value witness table members needed to retrieve the enum payload are documented. Could we use those (combined with the ABI information above to avoid copies)?

src/docs/binding-enums.md
```diff
@@ -0,0 +1,195 @@
...
+There are a number of ways that we can improve the experience somewhat. For example, we can define a common interface for
+all enums:
+```csharp
+public interface ISwiftEnum<T> where T : System.Enum {
+    T Case { get; }
+}
+```
+Which at the very least couples the cases and the enum.
+
+I've been in contact with the C# team to encourage that whatever form unions take we can hook this into as cleanly as possible. To this end, I feel it's important to have an approach which is "fast track/abstract" which means that there will be a native C# implementation that works with C# syntax in a way that's performant, but also a way to make a non-native implementation operate with the supporting C# syntax. This precedent already exists in C# in `foreach`: if you use `foreach` on an array you get a fast-track implementation. If you use `foreach` on an `IEnumerable<T>` you get a functional, but less efficient version, but from the user's point of view it operates the same.
```
jkoritzinsky, line 168: Can we update this section to add a reference to the [Type Unions proposal](https://github.com/dotnet/csharplang/blob/main/proposals/TypeUnions.md#custom-unions) that is the current direction for unions in C# (and specifically the "custom unions" section for types that would be our own structure but with the APIs for C# to enable the language features).

src/docs/binding-functions.md
```diff
@@ -0,0 +1,103 @@
...
+- Throws - throws a type that implements the `Error` protocol. The error ends up in a dedicated register which is nil is there was no error. It should be noted that `throws` is not the same as throwing in other languages in that an error that is thrown **must** be handled or rethrown. Swift has no notion of an exception automatically propagating up the call stack. This is no doubt because reference count updating under those circumstances would be non-trivial at best.
+- Async - function runs asynchronously
+- Mutating - function allows a value type to be changed within the function. This can only be used in the context of a value type.
+- Homonyms - swift allows functions to be distinct only on return type with all else the same
+
+
+## Language Parity
+In terms of language parity issues, the biggest one are Swift class methods as C# doesn't have this notion. One approach that could be taken is to implement a base class for all type metadata objects and represent class methods in that class. For example, given this swift:
```
jkoritzinsky, line 19: What makes "class methods" different from "static methods" in C#?

Could we make the "type metadata" an implementation detail of the projection and not expose the metadata parameter?

Do `open class` methods exist? If so, how are they called virtually in Swift?

src/docs/binding-functions.md
```diff
@@ -0,0 +1,103 @@
...
+
+Like C#, Swift has keyword escaping, but uses backticks as delimeters instead of the at-sign prefix in C#.
+
+Operators are problematic as C# has only a limited set of available of operators whereas Swift has a very flexible set of operators.
+
+## ABI Differences
+The runtime team is taking care of both the self and return pointers as well as value type passing and returning.
+
+## Runtime Differences
+There are likely issues in mapping async functions.
```
jkoritzinsky: We can add a note here that Swift Async is currently out of scope and would likely require runtime work to implement the `swiftasync` calling convention.

src/docs/binding-protocols.md
```diff
@@ -0,0 +1,247 @@
+# Binding Protocols
+
+Protocols in Swift fall into two categories that are handled similarly but are different enough to warrant their own sections.
+
+Protocols are a vestige of Objective C and work very much like interfaces in C#. Protocols are a contract that a type has a set of methods available. Unlike ObjC, Swift implements protocols through a vtable called a protocol witness table. The layout of protocol witness tables is fragile and undefined.
```
jkoritzinsky, line 5: Is the layout undefined even in stable ABI scenarios?

src/docs/binding-protocols.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Language Parity Mismatches
+
+The only language parity mismatch is that Swift offers protocol composition through the `&` operator. For anything but return types, this can be handled with generic types with multiple constraints. If a swift function takes an argument of type `P1 & P2 & P3` this can be handled in C# by making the argument generic and added `where T: P1, P2, P3`. This cannot be applied to the return value though because when you have `public func foo() -> P1 & P2` in Swift you are saying that the function (callee) can choose the return type as long as it conforms to the protocols, whereas in C# if you write `T Foo<T>() : where T: P1 & P2` you are saying that the *caller* decides the type. This is problematic since there is no analog to this in C#. Solutions we can apply include returning `object` or picking one of the protocols and using that. Finally we could return a representation of the existential container in which we could put a method to extract the type as a given interface type. This essentially leaves the unboxing to the user. All of these solutions will have unique side effects, which will be discussed later.
```
jkoritzinsky, line 12: Are there any usages of this pattern (especially in the return type case) in libraries that we are planning on supporting Swift interop for in .NET 10+?

src/docs/binding-structs.md
```diff
@@ -0,0 +1,100 @@
+# Binding Structs
+
+Structs in Swift are value types that can have a number of forms each of which requires a certain amount of consideration: frozen/non-frozen and blitable and non-blitable.
+
+Structs that are frozen are guaranteed not to change. They have a fixed layout and are passed by value lowered up until the point when the size of the struct exceeds the number of registers dedicated for value types in Swift (currently 4 and unlikely to change).
```
jkoritzinsky: Frozen structs with `weak` fields are passed by value. I don't believe that non-bitwise-copyable structs are passed by value, but I may need to check.

```suggestion
Structs that are frozen are guaranteed not to change. They have a fixed layout and can be passed by value lowered up until the point when the size of the struct exceeds the number of registers dedicated for value types in Swift (currently 4 and unlikely to change).
```

src/docs/binding-structs.md
```diff
@@ -0,0 +1,100 @@
...
+```
+It will generate the following output:
+```
+you killed Corey.
+you killed Ian.
+all done
+```
+This shows how the destructor in the class gets executed when a struct gets overwritten ("you killed Corey") and when it goes out of scope. C# will do neither of this things.
+
+Swift allows 0-length structs, whereas C# does not.
```
jkoritzinsky, line 59: How are 0-length structs treated at the ABI layer? Are they expanded to 1-byte or are they omitted?

src/docs/binding-structs.md
```diff
@@ -0,0 +1,100 @@
...
+Swift allows structs to be non-copyable by adding the pseudo inheritance `: ~Copyable` to the type declaration. These types may present issues when
+being bound. The value witness table contains functions that will execute an illegal instruction if they are called to copy the type and all mechanisms
+within swift to get the address of an non-copyable instance are forbidden.
+
+## ABI Differences
+Swift has very specific rules for packing structs which Apple has laid out [here](https://github.com/swiftlang/swift/blob/main/docs/ABI/TypeLayout.rst).
+
+Structs are passed to functions in one of two ways depending on whether or not they are frozen under `enable-module-evolution` rules.
+If the struct is frozen, it is lowered into up to 4 registers, otherwise it is passed by reference. This is usually done by copying it onto the stack and taking the address into a register. If the func is an instance method it will be passed by reference in the self register.
```
jkoritzinsky:
```suggestion
If the struct is frozen and bitwise-copyable, it is lowered into up to 4 registers, otherwise it is passed by reference. This is usually done by copying it onto the stack and taking the address into a register. If the func is an instance method it will be passed by reference in the self register.
```

src/docs/binding-tuples.md
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+It's unclear why Apple would do this except that it might be a vestige of when Swift had the ability to have multiple
+argument lists and could do partial function application. It also might be because it enables very low-friction usage of tuples in that when used outside of the context of nominal types, tuples don't actually need to exist. They can be implemented as an artifical grouping of scoped variable names.
+
+## Language Parity Mistmatch
+
+Since Tuples became available in C# they have had two varieties: `System.Tuple` and `System.ValueTuple`. The latter is what has been adopted into C# as a first class citizen in the language and since it matches closely to Swift tuples, we should follow that.
+
+One potential problem we need to investigate is that the C# `ValueTuple` is generic and the size of the struct can't be put into a `StructLayout` attribute. Furthermore, we need to know if we need to change `ValueTuple` to have `[StructLayout(LayoutKind.Sequential)]` added to the types. If not, it might be necessary to create our own type that can easily be converted back and forth to `ValueTuple`.
```
jkoritzinsky, line 20: For tuple fields in bitwise-copyable frozen structs, the Swift projection may need to define its own types, using the features in https://github.com/dotnet/runtime/issues/100896, to get correct layout rules (unless Swift separates out tuples fields like it does for parameters).

src/docs/binding-tuples.md
```diff
@@ -0,0 +1,60 @@
...
+public func foo(T: (a: Int, b: Bool, (c: SomeStruct, d: SomeClass))) { // func of 1 argument
+    // ...
+}
+
+public func foo(a: Int, b: Bool, c: SomeStruct, d: SomeClass) { // func of 4 arguments
+    // ...
+}
+```
+It's unclear why Apple would do this except that it might be a vestige of when Swift had the ability to have multiple
+argument lists and could do partial function application. It also might be because it enables very low-friction usage of tuples in that when used outside of the context of nominal types, tuples don't actually need to exist. They can be implemented as an artifical grouping of scoped variable names.
```
jkoritzinsky, line 14: Handling tuples in this manner allows Swift to enregister more parameters (as each element of the tuple can use up to 4 registers after physical lowering until registers are exhausted instead of the tuple as a whole only being able to use up to 4 registers).

src/docs/binding-tuples.md
```diff
@@ -0,0 +1,60 @@
...
+
+This accessor appears to be thread safe.
+
+## Runtime Mismatches
+Tuples that contain heap allocated types will create issues due to reference counting issues. This is the same as with Structs.
+
+For calling generic functions that are being passed a tuple from C#, we will need to pass in the appropriate type metadata. Fortunately, this is relatively straightforward.
```
jkoritzinsky: Can we add some information at the top about how tuples are passed in generic Swift code (not expanded, passed as a struct)?

src/docs/binding-value-witness-table.md
```diff
@@ -0,0 +1,26 @@
...
+- Size - machine word representing the size of the type in bytes
+- Stride - machine word representing the stride of the type in bytes
+- Flags - 32 bit unsigned int - contains flags that describe how to work with the time including the memory alignment
+- ExtraInhabitantCount - 32 bit unsigned int - the number of extra inhabitants (free bits) in the type
+
+## Getting the value witness table
+
+There are two ways to get the value witness table for a type. The first is to get the address of memory using `dlsym` with the entry point for the value witness table. The second is get it relative to the type metadata. In Swift there is extended metadata information that always precedes the type metadata. One machine word back from the type metadata is a pointer to the ValueWitnessTable.
+
+Probably the most efficient way to get the value witness table is to always go from the Swift Type Metadata.
```
jkoritzinsky, line 24: Is this mechanism reliable/documented? Or is the only documented manner using `dlsym`/`NativeLibrary.GetSymbol`/symbol lookup?

src/docs/binding-variables.md
```diff
@@ -0,0 +1,31 @@
...
+- global
+- static
+- instance
+- class (computed only)
+- mutating set (value types only)
+- open (instance in classes only)
+
+## Language Parity Mismatches
+- Naming - Swift can have variable names that are unpronouncable in C#
+- Class - the notion of class variables doesn't exist in C#
```
jkoritzinsky, line 17: Similar question to my question about class functions earlier, what about static properties? How do they differ?

src/docs/binding-variables.md
```diff
@@ -0,0 +1,31 @@
...
+- willSet/didSet - while these can be conceptually modeled with events in C#, these are problematic for binding as there is no real way to override/chain these
```
jkoritzinsky, line 19: Could these events be represented with the `INotifyPropertyChanging`/`INotifyPropertyChanged` interfaces in `System.ComponentModel` for instance-level (mutating set/open) properties?

src/docs/binding-variables.md
```diff
@@ -0,0 +1,31 @@
...
+
+## ABI Differences
+- Accessors have the same issue as member functions in that they use a self register for the instance (if any) and an error register if they can throw.
+- Accessors can be async
```
jkoritzinsky, line 121: I think this should also be listed in the "Language Parity" section as property accessors can't be async (though get-only properties could be represented as such without too much issue).

src/docs/binding-structs.md
jkurdek:
```suggestion
If the struct is frozen, it is lowered into up to 4 registers, otherwise it is passed by reference. This is usually done by copying it onto the stack and taking the address into a register. If the func is an instance method the struct will be passed by reference in the self register.
```

src/docs/binding-structs.md
```diff
@@ -0,0 +1,100 @@
...
+
+// ...
+// Consuming code, which seems totally reasonable for C#
+var n1 = new NamedHolder(SwiftString.FromString("Corey"));
+var n2 = new NamedHolder(SwiftString.FromString("Ian));
+n1 = n2; // memory leak.
+```
+
+## Accessibility
+The main problem that we have is with non-blittable structs. We would either need our users to manually destroy structs when they're no longer needed (this a really bad idea - people are awful at memory management - that's why we have garbage collection and `IDisposable`)
```
jkurdek: I think the `or` part is missing from `either...or`

src/docs/binding-classes.md
```diff
@@ -0,0 +1,397 @@
+# Binding Classes
+
+Classes function very similarly to C# classes with some important semantic and runtime differences which affect how they are exposed in C#, but generally speaking, they can exist as a 1:1 mapping. It should be noted that while Swift provides the ability to create traditional OO code, Apple is eschewing the use of classes in their own code and preferring value types instead. It is telling that in the main Swift runtime library, libswiftCore, there are no exported classes.
+
+Swift classes can be either native Swift or Objective-C. The latter is in place strictly for compatability with legacy code and in fact, if a Swift class is declared to be `@ObjC`, it must inherit from `NSObject` and the swift compiler will generate a selector dispatcher for all methods marked as `@ObjC` but these just go to the Swift implementation, therefore Objective-C compatability is just a thin veneer on top of the Swift implementation.
+
+## Language Parity Mismatches
+Swift provides a different set of accessibility of methods that C#:
```
jkurdek:
```suggestion
Swift provides a different set of accessibility of methods than C#:
```

src/docs/binding-overview.md
rolfbjarne: That's great!

I merely wanted to bring trimming/AOT up in case it could cause a problem (I have no idea how the COM interop is implemented, but the [method](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getcomobjectdata?view=net-8.0) you linked to doesn't look strongly typed at all - which is very often a sign something isn't trimmable), because it's better to catch these problems early on. But if it's all statically known, then no problem whatsoever.

src/docs/binding-variables.md
```diff
@@ -0,0 +1,31 @@
+# Binding Variables
+
+Variables in Swift come in two general forms: computed and stored.  Unlike C#, Swift allows variables to exist at the top level of a module being effectively global.
+
+With enable-module-evolution set, all `let` variables will have a getter. Variables that are declared with `var` will have a getter as well as a setter. It is an error to make a set-only variable.
```
jkurdek: Is enable-module-evolution something different than enable-library-evolution?

jkurdek (approved):
Great write-up!

src/docs/binding-classes.md
stephen-hawley: The challenge is that Swift needs to know a priori how to call C# from Swift method overrides. There's effectively a circular dependency here. Are you suggesting that C# exports symbols that Swift will have access to? I can see how this might work, but how is it in an improvement? Most of the actual work will be in marshaling since Swift really does not like passing reference counted types to C function pointers.

src/docs/binding-classes.md
stephen-hawley: Yes/no. The key type here is the address of a Swift instance. So `TKey` is effectively `IntPtr` and `TValue` is a handle to a C# class. This is used in the case of "I got an instance from Swift, have we seen this already in C#? If so, recover the C# instance. If not, store the C# instance. `ConditionalWeakTable` is built such that both `TKey` and `TValue` are constrained to `class`. How would that work?

src/docs/binding-classes.md
stephen-hawley: I mean technically, yes, we could generate it at runtime in the sense that it's just a block of data in memory, but the ordering is unknown. In theory, if you know how big the protocol witness table is, you could line up the address of the entries with methods in the dylib to figure out the ordering. This gets more challenging when the type is generic because really wants to own the creation process of, say, the protocol witness table for a type conforming to a protocol with associated types.

src/docs/binding-functions.md
stephen-hawley: Class methods are different from static methods in that they have an instance and that instance is the type metadata object for the type which is passed in the `self` register. Yes, you can have `open class` methods and they are dispatched through either the vtable or in the case of enable-library-evolution, through a dispatch thunk.

src/docs/binding-protocols.md
stephen-hawley: I know of know documentation of the layout of protocol witness tables.

src/docs/binding-structs.md
stephen-hawley: 0 length structs are omitted IIRC

src/docs/binding-tuples.md
stephen-hawley: Swift does not separate out tuple fields. Swift tuples get packed like structs using "Universal" layout. It's mentioned [here](https://github.com/swiftlang/swift/blob/main/docs/ABI/TypeLayout.rst), but this doc was last edited 6 years ago.

src/docs/binding-tuples.md
stephen-hawley: Can you explain more what you're looking for here @jkoritzinsky? I think I'm getting hung up on what you mean by 'generic Swift code'.

src/docs/binding-value-witness-table.md
stephen-hawley: This method is documented (although the docs are years old).

src/docs/binding-variables.md
stephen-hawley: The problem isn't in C# though, it's in Swift. I had investigated this ~6 years ago and dismissed it as impractical and it created some problems because of (1) implicit chaining of observers and (2) different ordering of the of the chaining (head vs tail). I'll dig out my old notes and add them to the doc

src/docs/binding-tuples.md
jkoritzinsky: I'm looking for an example here for swift code like the following (where `foo<T>` is instantiated with a tuple):

```swift
func bar() -> Int
{
	return foo((1, 45));
}

func foo<T>(x: T) -> Int { return 0; }
```

src/docs/binding-classes.md
jkoritzinsky: There's two scenarios I'm addressing here:

1. Passing an instance of a C# class that inherits from a Swift base class as a reference to the base class:

In this case, the projection tooling should generate the type metadata pointer in the Swift object wrapper for the C# object such that the type metadata and vtables are set up with pointers to the C# overrides. These functions will have the **Swift** calling convention, not the C calling convention.

Therefore, when the Swift compiler generates code to call methods on the vtables of (possibly-derived) instances of the Swift base class, the metadata is correctly constructed such that the call goes to the `UnmanagedCallersOnly` wrapper with the Swift calling convention that the projection tooling generates.

This would remove any requirements to emit Swift code for this scenario, as the projection could construct the code into the correct structure from the start.

This would require us to know how the vtable is sorted. However, the vtable must be sorted in some sort of stable fashion (or have a mechanism to look up members by name) to support the `enable-library-evolution` scenario. We can work with the folks at Apple to document the existing behavior as it is set in stone until they explicitly break ABI.

2. Exposing a C# class to be instantiated from Swift (not currently in scope)

In this case, when using a toolchain like NativeAOT, the correct symbols (pointing to the data produced in scenario 1) are exported as per the Swift docs so Swift code that links to the NativeAOT shared library correctly discovers all of the expected symbols to link and the Swift compiler can treat the NativeAOT binary as a Swift binary (as it exposed the same symbols as if the library had been authored in Swift).

src/docs/binding-classes.md
jkoritzinsky: As mentioned in another comment, can we work with Apple to document the existing ordering of methods in protocol witness tables in enable-library-evolution scenarios? Then the projection tooling could use those rules to order the members in its generated code.

Apple can't change the algorithm of the vtable ordering in the `enable-library-evolution` scenarios, even for `@frozen` types between compilations (or they have to have some other stable mechanism), so we can use that ordering/mechanism reliably (as non-enable-library-evolution scenarios are out of scope).

src/docs/binding-functions.md
jkoritzinsky: Could we model non-open class methods as static methods (and in the marshalling stub pass the type metadata)?

Could we model open class methods as static abstracts on interfaces?

src/docs/binding-tuples.md
jkoritzinsky: In that case, the Swift projection may need to define its own tuple fields to get the correct layout experience. However, those types could have a `Deconstruct` method defined on them and an implicit conversion from/to ValueTuple to enable the C# features around value tuples.

src/docs/binding-value-witness-table.md
jkoritzinsky: Years-old docs are fine if they're still valid. Can you add a link to the docs for this?

src/docs/binding-variables.md
jkoritzinsky: If it's too difficult to do, I'm fine with it being marked as out-of-scope.

I think we should mention that these interfaces could be used to model them, but other implementation difficulties make it very expensive (and if we don't need it for SwiftUI or another library that's in scope, then it's out of scope)

src/docs/binding-tuples.md
stephen-hawley: Sure - in that case, the tuple gets constructed by the calling code, gets passed by reference along with the type metadata that describes the tuple.
Given the following code:
```swift
public func foo<T>(x: T) {
    print(x)
}
public func printTuple() {
    foo(x: (true, 15))
}
```
We get the following assembly for printTuple:
```
push rbp
mov rbp, rsp
sub rsp, 10 // stack frame for tuple
mov byte [rbp+var_10], 0x1 // true
mov qword [rbp+var_8], 0xf // 15
lea rdi, qword [_$sSb_SitMD] // demangling cache variable for type metadata for (Swift.Bool, Swift.Int)
call ___swift_instantiateConcreteTypeFromMangledName // make/get cached metadata
mov rsi, rax // type metadata in arg 1
lea rdi, [rbp+var10] // address of tuple in arg 0
call foo<T>(x: T)
add rsp, 10
pop rbp
ret
```
Fun side note (for some definition of fun): the public sources for swift do not include `__swift_instantiateConcreteTypeFromMangledName`. There is one reference to it in C++ code and a bunch of references to it in test code but no actual source. Turns out that function gets injected by the Swift compiler into the compiled code which them calls `swift_getTypeByMangledNameInContext2` after doing some basic sanity checking on the input. The sanity checking includes a check for null on the input string, which is odd because (1) the compiler already knows that it will never get null and (2) the code returns null if the input is null which will cause later code to crash.
At any rate, I find this interesting because in the past, the compiler would inject a call to `swift_getTupleTypeMetadata` with a request, a count of elements, some flags, and  a pointer to an array of type metadata pointers for each element in the tuple.
I wonder if the decision to use one vs the other was because even though the cost of decoding the mangled name is higher, the cost of using a string as a key into a cache is lower. At any rate, interesting change.

src/docs/binding-value-witness-table.md
stephen-hawley: Done.

src/docs/binding-tuples.md
stephen-hawley: I should also point out that this is entirely consistent with swifts argument passing for generics: pass by reference and include type metadata implicitly.

src/docs/binding-classes.md
stephen-hawley: Enable-library-evolution is designed so that witness tables ordering can change and not break consuming code. Apple lays that out [here](https://www.swift.org/blog/library-evolution/) and are explicit in saying that the layout must be completely opaque:

> When a framework publishes a protocol, client code can declare types conforming to this protocol. The compiler generates a table of function pointers known as the protocol witness table to describe each protocol conformance. Calling a protocol requirement on a generic parameter requires loading the right function pointer from a protocol witness table. Since protocol requirements can be re-ordered, and new protocol requirements with default implementations can be added, the layout of a protocol witness table must be completely opaque across a resilience boundary.

> This is accomplished in two steps. First, for every protocol requirement, the binary framework exports a special function called a dispatch thunk. The dispatch thunk is part of the framework itself, and so it can directly hard-code the offset of the protocol requirement in the witness table. If the protocol’s declaration is changed to re-order requirements, the order of entries in the witness table is changed, but the symbol names of the dispatch thunks remain the same. Since client code calls all protocol methods via dispatch thunks, binary compatibility with future versions of the framework can be maintained.

> Finally, to cope with adding new protocol requirements, protocol witness tables require runtime instantiation. Instead of emitting a witness table in the client code directly, the compiler emits a symbolic description of the conformance. The instantiation process places the protocol requirements in the correct order and fills missing entries to point to their default implementation, to produce a well-formed witness table which can be passed off to a dispatch thunk.

src/docs/binding-classes.md
stephen-hawley: OK - great. Thanks for explaining what you're thinking of. Scenario 1 is effectively impossible. `enable-library-evolution` dictates that the order of the vtable is undefined. (see the doc I linked from protocols). Resiliency is managed through dispatch thunks. The relevant paragraph is:
> The key takeaway here is that method dispatch to Swift-native methods on a resilient class is performed by calling a dispatch thunk; as with protocols, this allows methods on the class to be re-ordered and new methods added without disturbing callers. This mechanism also allows the superclass to add or remove methods, without disturbing subclasses.

Now, is it possible for us to write code that will generate a Swift type that will dispatch into C#? Yes, and in order to avoid having to implement our own code to generate type metadata etc, we let the Swift compiler do all of that for us.

But - here is an interesting thought for an optimization to that process - what if we wrote a Swift implementation in a way that had really obvious code for each of the method implementations and then rewrote the compiled code at link time or runtime to do go where we wanted.  So think about it like this - suppose I have the following code and wrapper:
```swift
open class OverrideMe { // assume that this came from someone else's library
    public init () { }
    open func multiName(text: String) -> String {
        return text + text
    }
}

open class OurWrapper : OverrideMe { // we generate this
    var gcHandle: OpaquePointer
    public init(gcHandle: OpaquePointer) {
        super()
        self.gcHandle = gcHandle
    }
    public func multiNameSuper(text: String, handle: OpaquePointer) -> String
    {
         return super.multiName(text: text)
    }
    open override multiName(text: String) -> String {
        return multiNameSuper(text: text, handle: gcHandle) // this is the important line
    }
}
```

So now imagine that we root through the binary, find the override and change the call site from `multiNameSuper` to a reference to the C# `[UnmanagedCallersOnly]` callback function. What this does is turn the responsibility of setting the callback pointer to a post-aot step.

I do want to stress that I'm softpedaling the challenges of rewriting the binaries post AOT and ignoring the cases when there is interpreter/JIT.

src/docs/binding-classes.md
jkoritzinsky: In this case, we should be able to emit the symbolic description of the conformance and the dispatch thunk as Swift expects to call them. Then, because the Swift code is compiled with `enable-library-evolution`, the Swift compiler will emit the code to call the protocol implementations through the dispatch thunks.

src/docs/binding-tuples.md
jkoritzinsky: Can we add a little of this into the doc (mainly that tuples aren't treated specially in generics as you mentioned in your second comment)?

PR #2681, opened 2024-09-09
[NativeAOT-LLVM] Add _requiresAlign8 to the compare method and name for GCStaticEETypeNode to avoid asserts when sorting.

yowl:
This PR fixes an assert in `SortableDependencyNode` and an `InvalidOperationException` in `WasmObjectWriter` when we have 2 `GCStaticEETypeNode` with the same GC bit map but different align 8 requirements.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GCStaticEETypeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GCStaticEETypeNode.cs
@@ -44,6 +44,10 @@ namespace ILCompiler.DependencyAnalysis
         public void AppendMangledName(NameMangler nameMangler, Utf8StringBuilder sb)
         {
             sb.Append("__GCStaticEEType_"u8).Append(_gcMap.ToString());
+            if (_requiresAlign8)
+            {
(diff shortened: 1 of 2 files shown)


yowl:
cc @dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/GCStaticEETypeNode.cs
```diff
@@ -44,6 +44,10 @@ protected override ObjectNodeSection GetDehydratedSection(NodeFactory factory)
         public void AppendMangledName(NameMangler nameMangler, Utf8StringBuilder sb)
         {
             sb.Append("__GCStaticEEType_"u8).Append(_gcMap.ToString());
+            if (_requiresAlign8)
+            {
+                sb.Append("_align8");
```
SingleAccretion:
```suggestion
                sb.Append("_align8"u8);
```

SingleAccretion (approved):
LGTM modulo one nit.

This should be upstreamed.

yowl:
> This should be upstreamed.

Right, will do.

pavelsavara (approved):

Issue #2663, opened 2024-08-22
[Swift Bindings] Integrate MachO.cs from xamarin-macios


stephen-hawley:
Code exists [here](https://github.com/xamarin/xamarin-macios/blob/main/tools/common/MachO.cs).
Note that there exists a version of this in [Binding Tools for Swift](https://github.com/xamarin/binding-tools-for-swift/blob/main/SwiftReflector/MachOHawley.cs) which is more complete in terms of support for symbols but is probably out of date for some of the architectures.

This code will be used for consuming delis and extracting the entry points.

What is needed in addition to this:
- code to get a list of all architectures in the file, the operating system, the min/max OS, and whether or not it's a simulator
- If using macios code, code to add `SymbolTableCommand` which inherits from `LoadCommand`. This class should include a collection of the 
- If using macios code, code to extend MachOFile.Read to read in SymbolTableCommands
- code to read a particular architecture and return an `IEnumerable<string>` which is a sequence of all public entry points

The enumeration code might look like this:
```csharp
static IEnumerable<string> GetPublicSymbols (this MachOFile file) {
    var symbols = file.load_commands.OfType<SymTabLoadCommand> ().First().Where(nle => nle.IsPublic && nle.EntryType == NListEntryType.InSection);
    return symbols.Select(sym => sym.str);
}
```

Bonus points: augment the code to run async

Testing:
Create a dylib (or consume an existing one on the OS) and assert that:
- it correctly excludes private symbols
- it correctly returns known public symbols
- it correctly reports the architecture
- it correctly reports the OS
- it correctly reports the minimum and maximum OS

PR #2686, opened 2024-09-12
Add MachO to read symbols

stephen-hawley:
This pulls in MachO.cs from xamarin-macios, addressing issue [2663](https://github.com/dotnet/runtimelab/issues/2663). I've made two fundamental changes to it:

1. I made all exceptions thrown conditional on the code being in `MTOUCH` or not. This is probably not the right switch. @rolfbjarne - what would you do?
2. I added symbol table support.

With these two changes in mind, @rolfbjarne - do we want to cross-pollinate this back into macios?

I made a minor refactor of BindingsGenerator in order to make it easier to access the path to the Swift dylib file. We'll need this later and it makes writing the tests easier as I can leverage the existing code.

Here's the why for all of this:
We need to get certain symbols from the Swift output that are not in the abi.json file. These include protocol witness tables, protocol descriptors, and dispatch thunks (and probably other things as well).

The choice to use MachO to do this comes down to these reasons:
1. MachO is small and reasonably efficient. 
2. MachO is extremely well tested and exercised
3. Getting the information from otool and nm involves launching executables that are each more than 100K in size and then will require parsing.

### DIFF
--- a/src/Swift.Bindings/src/Demangler/MachO.cs
+++ b/src/Swift.Bindings/src/Demangler/MachO.cs
@@ -6,12 +6,6 @@ using System.Text;
 
 #nullable enable
 
-#if MLAUNCH
-using Xamarin.Launcher;
-#elif XAMARIN_HOSTING
-using Xamarin.Hosting;
-#endif
-
 namespace Xamarin {
 	[Flags]
 	public enum Abi {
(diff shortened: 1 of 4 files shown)


rolfbjarne:
> 2\. 
> 
> With these two changes in mind, @rolfbjarne - do we want to cross-pollinate this back into macios?

No, I think it's fine if this file diverges.

I would just rip out all the code in the MTOUCH sections.

> The choice to use MachO to do this comes down to these reasons:

4. The Mach-O file format is quite stable, with only a few additions over the years. Those are always documented in header files though, so it's easy to update the code.
5. The code was easy to write, and turned out bug free pretty much from the get go (I don't think there as ever been any bugs reported against code from this file).

rolfbjarne (approved):

mcumming (approved):

Issue #2648, opened 2024-08-08
[NativeAOT-LLVM] Re-enable the `SharedLibrary` test for WASI P2


SingleAccretion:
It will be disabled in #2647 because it needs a rewritten driver, which is being done in #2614.

pavelsavara:
Fixed by https://github.com/dotnet/runtimelab/pull/2778

PR #2688, opened 2024-09-17
[NativeAOT-LLVM] Merge to Sep 24

yowl:
This PR merges from upstream to Sep 2024.

I made `acdHelper` `public` so we can access it from llvm codegen

See the comments at the end re non-running tests and failure on linux to find `libz`


SingleAccretion:
> remove ThrowMisalign

This helper is NAOT-LLVM specific and should be kept (it is used for atomics).

yowl:
Thanks, yes realised that a bit later.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

yowl:
Compiling XUnitLogChecker is failing on `linux-x64`, first on `bfd`/`ld` decision, fixing that moves it on to missing `-lz` library.  Wouldn't have expected this from a merge.  Perhaps the container is not correct for some reason.

yowl:
cc @dotnet/nativeaot-llvm 

There's a few strange things here:

1. The tests for browser and wasi claim to pass, but in fact they are not running.  I think the merging `nativeaot` wrapper should be compiled for the host so it can write files, but maybe I misunderstood the discussion in https://github.com/dotnet/runtimelab/issues/2695
2. I've had to remove the use of `-lz` on linux for the building of tests.  I can't see why this works upstream, but not here.

src/coreclr/jit/utils.cpp
```diff
@@ -1626,24 +1626,13 @@ void HelperCallProperties::init()
...
                 isPure        = true;
                 noThrow       = true;
                 nonNullReturn = true;
                 break;
 
-            case CORINFO_HELP_READYTORUN_GENERIC_HANDLE:
-                // Typically this is a no-throw call, but may become always-throw when fetching an unavailable handle.
-                isPure        = true;
-                nonNullReturn = true;
-                break;
```
SingleAccretion: This should be restored. `CORINFO_HELP_READYTORUN_GENERIC_HANDLE` may throw. This should fail one of the tests (`UnitTests` if memory serves me right).

src/libraries/System.Net.Http/src/System.Net.Http.csproj
```diff
@@ -457,20 +457,15 @@
              Link="Common\System\Net\Http\HttpHandlerDefaults.cs" />
   </ItemGroup>
 
-<<<<<<< HEAD
   <!-- TODO-LLVM: This is not upstreamable because it makes the build runtime-specific. -->
   <ItemGroup Condition="'$(TargetPlatformIdentifier)' == 'wasi' and '$(RuntimeFlavor)' != 'CoreCLR'">
-    <Reference Include="System.Threading.Thread" />
-=======
-  <ItemGroup Condition="'$(TargetPlatformIdentifier)' == 'wasi'">
```
SingleAccretion: Can we delete these ` '$(RuntimeFlavor)' != 'CoreCLR'` special cases now and take upstream verbatim? It would really like to avoid NAOT-LLVM conditionals outside of core runtime.

src/tests/nativeaot/SmokeTests/SharedLibrary/SharedLibrary.csproj
```diff
@@ -5,26 +5,18 @@
...
-    <CLRTestTargetUnsupported Condition="'$(TargetsAppleMobile)' == 'true'">true</CLRTestTargetUnsupported>
-
-    <!-- https://github.com/dotnet/runtimelab/issues/2648 -->
-    <CLRTestTargetUnsupported Condition="'$(TargetOS)' == 'wasi'">true</CLRTestTargetUnsupported>
-=======
     <BuildAsStandalone>false</BuildAsStandalone>
-    <!-- Unable to compile a project with the Library output type for apple mobile devices -->
-    <CLRTestTargetUnsupported Condition="'$(TargetsAppleMobile)' == 'true'">true</CLRTestTargetUnsupported>
+    <!-- Unable to compile a project with the Library output type for apple mobile devices or WASI -->
+    <CLRTestTargetUnsupported Condition="'$(TargetsAppleMobile)' == 'true' or '$(TargetOS)' == 'wasi'">true</CLRTestTargetUnsupported>
```
SingleAccretion: This should stay as a distinct condition pointing to `https://github.com/dotnet/runtimelab/issues/2648`.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/SocketsHttpHandler.cs
```diff
@@ -213,5 +213,26 @@ public Func<SocketsHttpPlaintextStreamFilterContext, CancellationToken, ValueTas
             get => throw new PlatformNotSupportedException();
             set => throw new PlatformNotSupportedException();
         }
+
+        public static bool SupportsAutomaticDecompression
```
SingleAccretion: Why do we have these differences from upstream?

src/tests/nativeaot/StartupHook/StartupHook.csproj
```diff
@@ -18,5 +18,6 @@
 
   <PropertyGroup>
     <CLRTestTargetUnsupported Condition="'$(TargetsWasm)' == 'true'">true</CLRTestTargetUnsupported>
+    <CLRTestExitCode>100</CLRTestExitCode>
```
SingleAccretion: Why is this needed?

src/tests/nativeaot/nativeaot.csproj
```diff
@@ -1,8 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
-    <ItemGroup>
-      <MergedWrapperProjectReference Include="*/**/*.??proj" Exclude="*/**/Library.csproj" />
-    </ItemGroup>
-  
-    <Import Project="$(TestSourceDir)MergedTestRunner.targets" />
-  </Project>
+  <PropertyGroup>
+    <CLRTestExitCode Condition="'$(CLRTestExitCode)' == ''">100</CLRTestExitCode>
```
SingleAccretion: This should already be set by `Common\CLRTest.Execute.targets`, why does it need to be duplicated?

src/tests/nativeaot/nativeaot.csproj
```diff
@@ -1,8 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
-    <ItemGroup>
-      <MergedWrapperProjectReference Include="*/**/*.??proj" Exclude="*/**/Library.csproj" />
-    </ItemGroup>
-  
-    <Import Project="$(TestSourceDir)MergedTestRunner.targets" />
-  </Project>
```
SingleAccretion: It would be good to reduce the diff on this file by following the upstream (wrong) whitespace formatting.

src/tests/nativeaot/nativeaot.csproj
```diff
@@ -1,8 +1,13 @@
...
+  <PropertyGroup>
+    <CLRTestExitCode Condition="'$(CLRTestExitCode)' == ''">100</CLRTestExitCode>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <MergedWrapperProjectReference Include="*/**/*.??proj" Exclude="*/**/Library.csproj;*/**/HelloWasm/*.??proj" Condition="'$(TargetsWasm)' != 'true'" />
+    <MergedWrapperProjectReference Include="*/**/*.??proj" Exclude="*/**/Library.csproj"  Condition="'$(TargetsWasm)' == 'true'" />
```
SingleAccretion: IMO it would be better to "include everything" (basically upstream include), then exclude the WASM projects we care about, basically replicating the list we had in `src/tests/Common/dirs.proj`.

Also, `dirs.proj` itself needs to be updated to be synced to upstream version now that no changes are needed for it.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
```diff
@@ -7,7 +7,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PublishAot Condition="'$(NativeAotSupported)' == 'true'">true</PublishAot>
     <SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>
-    <LinkerFlavor Condition="'$(CrossBuild)' == 'true' and '$(TargetsLinux)' == 'true'">lld</LinkerFlavor>
+    <LinkerFlavor Condition="('$(__TestBuildMode)' == 'nativeaot' or '$(CrossBuild)' == 'true') and '$(TargetsLinux)' == 'true'">lld</LinkerFlavor>
```
SingleAccretion: This indicates our test build isn't passing `CrossBuild` down correctly. I think we are missing `$(crossArg)` in the `./src/tests/build.sh` invocations in the templates. I think this explains the `-lz` mystery issue as well.

SingleAccretion (commented):
Obviously, runtime tests need to be fixed to run. I think for this merge we should use the escape hatches, i. e. set `BuildAllTestsAsStandalone` somewhere globally, and make the merged model work separately.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
yowl: Done.

src/tests/nativeaot/nativeaot.csproj
yowl: We can't use this for Wasm as we are not using the merged wrapper, but have changed for non -wasm

src/tests/nativeaot/nativeaot.csproj
yowl: Done

src/tests/nativeaot/nativeaot.csproj
yowl: removed.

src/tests/nativeaot/StartupHook/StartupHook.csproj
yowl: reverted

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/SocketsHttpHandler.cs
yowl: wasi doesn't compile without them, let me check why

src/tests/nativeaot/StartupHook/StartupHook.csproj
yowl: reverted

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/SocketsHttpHandler.cs
yowl: Reverted and remove some workarounds that were pending this merge

src/tests/nativeaot/SmokeTests/SharedLibrary/SharedLibrary.csproj
yowl: THanks, missed that.

src/libraries/System.Net.Http/src/System.Net.Http.csproj
yowl: Yes , we can

src/coreclr/jit/utils.cpp
yowl: restored

yowl:
Got there finally.  Can be reviewed again, thanks!

src/coreclr/jit/utils.cpp
```diff
@@ -1629,7 +1629,6 @@ void HelperCallProperties::init()
             case CORINFO_HELP_READYTORUN_GENERIC_HANDLE:
                 // logging helpers are not technically pure but can be optimized away
                 isPure        = true;
-                noThrow       = true;
                 nonNullReturn = true;
                 break;
```
SingleAccretion:
```suggestion
                noThrow       = true;
                nonNullReturn = true;
                break;

            case CORINFO_HELP_READYTORUN_GENERIC_HANDLE:
                // Typically this is a no-throw call, but may become always-throw when fetching an unavailable handle.
                isPure        = true;
                nonNullReturn = true;
                break;
```
And take out `case CORINFO_HELP_READYTORUN_GENERIC_HANDLE` from above. There is no need to pessimize the other helpers.

src/libraries/System.Net.Http/src/System/Net/Http/WasiHttpHandler/WasiHttpHandler.cs
```diff
@@ -167,10 +167,8 @@ protected internal override async Task<HttpResponseMessage> SendAsync(HttpReques
 
         #region PlatformNotSupported
 #pragma warning disable CA1822
-
         internal ClientCertificateOption ClientCertificateOptions;
 
-
```
SingleAccretion: Whitespace changes should be reverted.

src/libraries/System.Net.Http/src/System.Net.Http.csproj
```diff
@@ -24,13 +24,8 @@
...
   </PropertyGroup>
 
-  <!-- TODO-LLVM: This is not upstreamable because it makes the build runtime-specific. -->
-  <PropertyGroup>
-    <DefineConstants Condition="'$(RuntimeFlavor)' == 'CoreCLR'">$(DefineConstants);NATIVE_AOT</DefineConstants>
-  </PropertyGroup>
-  
   <ItemGroup>
-    <ILLinkSubstitutionsXmls Include="$(ILLinkDirectory)ILLink.Substitutions.xml" />
+    <ILLinkSubstitutionsXmls Include="$(ILLinkDirectory)ILLink.Substitutions.xml" Condition="'$(TargetPlatformIdentifier)' != 'wasi'"/>
```
SingleAccretion: `Condition="'$(TargetPlatformIdentifier)' != 'wasi'"` still needed?

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -8,6 +8,7 @@ parameters:
   testFilter: tree nativeaot
   isOfficialBuild: false
   librariesConfiguration: Debug
+  crossArg: $(crossArg)
```
SingleAccretion: Why indirect through the `parameters` template arguments, instead of using `$(crossArg)` directly?

src/tests/Common/mergedrunnermobile.targets
```diff
@@ -39,8 +39,26 @@
...
+  
   <!-- Don't use the usual script generation, the scripts it generates are not useful for these tests. -->
-  <Target Name="GenerateExecutionScriptsInternal" />
+  <Target Name="GenerateExecutionScriptsInternal"
+    Condition="$(_CLRTestNeedsToRun) or $(_CLRTestBuildsExecutable)"
+    DependsOnTargets="FetchExternalProperties;Generate@(ExecutionScriptKind, 'ExecutionScript;Generate')ExecutionScript;" />
+
   <Target Name="GenerateLocalRunnerScript" DependsOnTargets="GenerateRunScript" AfterTargets="PublishTestAsSelfContained" />
 
   <Import Project="$(RepositoryEngineeringDir)testing\tests.targets" />
```
SingleAccretion: Revert this file?

src/tests/Directory.Build.targets
```diff
@@ -23,6 +23,10 @@
     <IsRuntimeTests>true</IsRuntimeTests>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetsWasm)' == 'true' and '$(RuntimeFlavor)' == 'coreclr'">
+    <BuildAllTestsAsStandalone>true</BuildAllTestsAsStandalone>
+  </PropertyGroup>
```
SingleAccretion:
```suggestion
  <PropertyGroup Condition="'$(TargetsWasm)' == 'true' and '$(RuntimeFlavor)' == 'coreclr'">
    <!-- NativeAOT-LLVM: https://github.com/dotnet/runtimelab/issues/2695. -->
    <BuildAllTestsAsStandalone>true</BuildAllTestsAsStandalone>
  </PropertyGroup>  
```

src/coreclr/inc/jithelpers.h
```diff
@@ -119,20 +119,11 @@
...
-    DYNAMICJITHELPER(CORINFO_HELP_OVERFLOW,     NULL,               METHOD__THROWHELPER__THROWOVERFLOWEXCEPTION)
-    DYNAMICJITHELPER(CORINFO_HELP_THROWDIVZERO, NULL,               METHOD__THROWHELPER__THROWDIVIDEBYZEROEXCEPTION)
-    DYNAMICJITHELPER(CORINFO_HELP_THROWNULLREF, NULL,               METHOD__THROWHELPER__THROWNULLREFEXCEPTION)
-    JITHELPER(CORINFO_HELP_THROWMISALIGN,       NULL,               METHOD__NIL)
-    DYNAMICJITHELPER(CORINFO_HELP_VERIFICATION, NULL,               METHOD__THROWHELPER__THROWVERIFICATIONEXCEPTION)
-=======
     DYNAMICJITHELPER(CORINFO_HELP_RNGCHKFAIL,   NULL,               METHOD__THROWHELPERS__THROWINDEXOUTOFRANGEEXCEPTION)
     DYNAMICJITHELPER(CORINFO_HELP_OVERFLOW,     NULL,               METHOD__THROWHELPERS__THROWOVERFLOWEXCEPTION)
     DYNAMICJITHELPER(CORINFO_HELP_THROWDIVZERO, NULL,               METHOD__THROWHELPERS__THROWDIVIDEBYZEROEXCEPTION)
     DYNAMICJITHELPER(CORINFO_HELP_THROWNULLREF, NULL,               METHOD__THROWHELPERS__THROWNULLREFEXCEPTION)
```
SingleAccretion, line 125:
```suggestion
    DYNAMICJITHELPER(CORINFO_HELP_THROWNULLREF, NULL,               METHOD__THROWHELPERS__THROWNULLREFEXCEPTION)
    JITHELPER(CORINFO_HELP_THROWMISALIGN,       NULL,               METHOD__NIL)
```

src/tests/nativeaot/nativeaot.csproj
```diff
@@ -1,8 +1,16 @@
...
+    </ItemGroup>
+
+    <!-- Wasm disabled projects are still in dirs.proj because Wasm is not using this merged test wrapper -->
+
     <ItemGroup>
-      <MergedWrapperProjectReference Include="*/**/*.??proj" Exclude="*/**/Library.csproj" />
+      <MergedWrapperProjectReference Include="*/**/*.??proj" Exclude="@(DisabledProjects)" />
     </ItemGroup>
-  
+
```
SingleAccretion:
```suggestion
  
```
Reducing diffs.

src/tests/Common/mergedrunnermobile.targets
yowl: yes, thanks.

src/tests/nativeaot/nativeaot.csproj
yowl: thanks, spaes added.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
yowl: no reason, removed.

src/libraries/System.Net.Http/src/System.Net.Http.csproj
yowl: no, removed.  Thanks.

src/libraries/System.Net.Http/src/System/Net/Http/WasiHttpHandler/WasiHttpHandler.cs
yowl: Done.

src/coreclr/jit/utils.cpp
yowl: Done, thanks.

yowl:
Latest feedback addressed,  thanks!

SingleAccretion (approved):
LGTM, thank you!

jkotas:
Thank you all!

Issue #2691, opened 2024-09-25
[Swift Binding] See about making Node reduction rules static


stephen-hawley:
The Swift5Reducer used non-static rules, mostly to get access to the demangled symbol. That can probably be passed in the reducer function. When used repeatedly, this will cut the overhead of making the rules collection.
In addition, the reduce function takes a string which is not currently used. Maybe get rid of that too.

Issue #2690, opened 2024-09-25
[Swift Binding] In node reduction put in support for inner types


stephen-hawley:
right now the demangled symbol for something like:
```swift
public struct Foo {
    public struct Bar {
    // ...
    }
    // ...
}
```
will cause an error for any symbol that references `Foo.Bar`

Issue #2664, opened 2024-08-22
[Swift Binding] Complete demangler


stephen-hawley:
The demangler as it exists returns a tree of nodes which represent the meaning of the entry point.

There exists code in Binding Tools for Swift which can accept one of these trees and match and reduce it to a wide variety of types that are meaningless in this context.

The existing match/reduce code should instead reduce to a set of `IEntryPoint`:
```
public interface IEntryPoint {
    string Symbol { get; }
}
```

We would need at least the following types for this project:
- UnknownEntryPoint
- ValueWitnessTable
- ProtocolWitnessTable
- Metadata Accessor
- Dispatch Thunk

Each of these would have members for the data that we care about:
ValueWitnessTable:
- NamedTypeSpec for which the accessor applies

ProtocolWitnessTable:
- NamedTypeSpec for the protocol represented
- NamedTypeSpec for the type implementing the protocol

Metadata Accessor:
- NamedTypeSpec for the type
- Signature of the accessor

DispatchThunk:
- NamedTypeSpec for the type that methods applies to
- Signature of the function

Challenges:
Generic members in mangled signatures are referenced by a set of coordinates (d, b) where d is the (inverse) depth of the generic in the declaration and b is the breadth. If a generic is declared in the function signature, the depth will always be 0. If the generic is declared in a class containing the function, depth will be the distance from the function to the generic declaration.
When generic symbols get imported as NamedTypeSpecs, they should probably be renamed as `T_d_b`. This will make matching easier later on.

stephen-hawley:
AFAICT this is complete

PR #2692, opened 2024-09-25
[Swift] Add first piece of rule reduction

stephen-hawley:
This PR adds demangling tree reduction to the project.
What does this mean?
When a symbol is demangled there is an intermediate state where there is a tree of Node which represent the semantic meaning of the symbol. In order for this to be useful to us, we need to reduce a tree of Node to a more directly accessible set of types.

To do this, there are 3 main pieces:
1. a way to determine if a Node (and possibly subnodes) match a given pattern. This is `MatchRule`
2. a way to uniformly apply rules (RulesRunner)
3. a way, given a matched node, to reduce it to a more easily consumable data structure

This PR partially addresses issue [2664](https://github.com/dotnet/runtimelab/issues/2664).

What happens: given a `Swift5Demangler` instance, calling `Run` with a symbol returns an `IReduction` that represents the result of the reduction. Right now, we're only ever instantiating either a `ProtocolWitnessTableReduction`, `TypeSpec` or a `ReductionError`, but I created placeholder for functions which will help enable dispatch thunks later.

Inner types will not work yet - issue opened [here](https://github.com/dotnet/runtimelab/issues/2690).

The reduction rules are non-static and the collection gets rebuilt for every symbol. This is [known](https://github.com/dotnet/runtimelab/issues/2691).

There is a name argument in the rule which is not used (yet). I'm trying to decide still if we're going to need it.

MatchRule is very flexible - it can do the simplest matching on a single `Node.Kind` but it can also match on multiples as well. It can match on child nodes too. Overall, this is a simplified version of a regular expression in a way. Nothing in the current code requires it, but we will need it for more complicated reductions later (thunks, for example).

Where this is going: we should be able to get all the protocol witness table symbols from a file with a single expression:
```
var protoWitnesses = machoFiies.AllPublicSymbols().Select(nle => new Swift5Demangler(nle.str).Run()).OfType(ProtocolWitnessTableReduction);
```
Similar expressions could be used to get all the dispatch thunks etc.

### DIFF
--- a/src/Swift.Bindings/src/Demangler/MatchRule.cs
+++ b/src/Swift.Bindings/src/Demangler/MatchRule.cs
@@ -1,0 +1,133 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace BindingsGeneration.Demangling;
+/// <summary>
+/// MatchRule represents a tree reducing rule for a demangled Swift symbol.
+/// Given a tree of nodes, a match rule contains conditions for the match,
+/// which include:
+/// - If and how the node content should be matched
+/// - If the number of children should match
+/// - Rules to run on the children
+/// - One or more NodeKinds to match
+/// If a match occurs, a reducer function can be run on the node.
+/// </summary>
+internal class MatchRule {
+    /// <summary>
+    /// The name of this rule - usefule for debugging
+    /// </summary>
+    public required string Name { get; init; }
+
+    /// <summary>
+    /// A list of node kinds to match. 
+    /// </summary>
+    public List<NodeKind> NodeKindList { get; set; } = new List<NodeKind>();
+
+    /// <summary>
+    /// A convenience accessor to match on a single NodeKind
+    /// </summary>
+    public NodeKind NodeKind {
+        get {
+            if (NodeKindList.Count != 1) {
+                throw new InvalidOperationException($"NodeKind is invalid when NodeKindList has {NodeKindList.Count} entries.");
+            }
+            return NodeKindList[0];
+        }
+        set {
+            NodeKindList = new List<NodeKind> { value };
+        }
+    }
+
+    /// <summary>
+    /// If and how to match the content of the Node
+    /// </summary>
+    public MatchNodeContentType MatchContent { get; init; } = MatchNodeContentType.None;
+
+    /// <summary>
+    /// Rules to apply to children node.
+    /// </summary>
+    public List<MatchRule> ChildRules { get; init; } = new List<MatchRule>();
+
+    /// <summary>
+    /// Whether or not the total number of children should match
+    /// </summary>
+    public bool MatchChildCount { get; init; } = false;
+
+    /// <summary>
+    /// A reducer to apply if the node matches
+    /// </summary>
+    public required Func<Node, string?, IReduction> Reducer { get; init; }
+
+    /// <summary>
+    /// Returns true if and only if the given node matches this rule
+    /// </summary>
+    /// <param name="n">a node to match on</param>
+    /// <returns></returns>
+    public bool Matches(Node n)
+    {
+        return NodeKindMatches(n) && ContentMatches(n) && ChildrenMatches(n);
+    }
+
+    /// <summary>
+    /// Returns true if and only if the NodeKind matches
+    /// </summary>
+    /// <param name="n">a node to match on</param>
+    /// <returns></returns>
+    bool NodeKindMatches(Node n)
+    {
+        return NodeKindList.Contains(n.Kind);
+    }
+
+    /// <summary>
+    /// Returns true if and only if the content of the given node matches
+    /// </summary>
+    /// <param name="n">a node to match on</param>
+    /// <returns></returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    bool ContentMatches(Node n)
+    {
+        // Only care about the content type not its value
+        switch (MatchContent)
+        {
+        case MatchNodeContentType.AlwaysMatch:
+            return true;
+        case MatchNodeContentType.Index:
+            return n.HasIndex;
+        case MatchNodeContentType.Text:
+            return n.HasText;
+        case MatchNodeContentType.None:
+            return !n.HasIndex && !n.HasText;
+        default:
+            throw new InvalidOperationException ($"Unknown match instruction {MatchContent} in match rule.");
+        }
+    }
+
+    /// <summary>
+    /// Returns true if and only if the children rules matches the given node's children
+    /// </summary>
+    /// <param name="n"></param>
+    /// <returns></returns>
+    bool ChildrenMatches (Node n)
+    {
+        // if the rule says the child count matters, apply
+        if (MatchChildCount && n.Children.Count != ChildRules.Count)
+            return false;
+
+        // match up to the minimum of each list
+        // if MatchChileCount is true, min is the size of both lists
+        int minimumChildCount = Math.Min (n.Children.Count, ChildRules.Count);
+        for (var i = 0; i < minimumChildCount; i++) {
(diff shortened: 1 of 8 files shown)


src/Swift.Bindings/tests/DemanglerTests/BasicDemanglingTests.cs
```diff
@@ -0,0 +1,71 @@
...
+        Assert.Equal("GenericTestFramework.IsItReal", protoWitnessReduction.ProtocolType.Name);
+    }
+
+    [Fact]
+    public void TestFailDemangleNonsense()
+    {
+        var symbol = "_$ThisIsJustGarbage";
+        var demangler = new Swift5Demangler (symbol);
+        var result = demangler.Run ();
+        var err = result as ReductionError;
```
matouskozak, line 56: What is expected to happen if we would instead do `result as ProtocolWitnessTableReduction`? I

src/Swift.Bindings/src/Demangler/MatchRule.cs
```diff
@@ -0,0 +1,133 @@
...
+/// which include:
+/// - If and how the node content should be matched
+/// - If the number of children should match
+/// - Rules to run on the children
+/// - One or more NodeKinds to match
+/// If a match occurs, a reducer function can be run on the node.
+/// </summary>
+internal class MatchRule {
+    /// <summary>
+    /// The name of this rule - usefule for debugging
```
matouskozak:
```suggestion
    /// The name of this rule - useful for debugging
```

src/Swift.Bindings/src/Demangler/MatchRule.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    /// <summary>
+    /// Returns true if and only if the content of the given node matches
+    /// </summary>
+    /// <param name="n">a node to match on</param>
+    /// <returns></returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    bool ContentMatches(Node n)
+    {
+        // Only care about the content type not its value
```
matouskozak, line 93: Shouldn't we rename it to `ContentTypeMatches` if we only care about the content type?

src/Swift.Bindings/src/Demangler/MatchRule.cs
```diff
@@ -0,0 +1,133 @@
...
+/// MatchRule represents a tree reducing rule for a demangled Swift symbol.
+/// Given a tree of nodes, a match rule contains conditions for the match,
+/// which include:
+/// - If and how the node content should be matched
+/// - If the number of children should match
+/// - Rules to run on the children
+/// - One or more NodeKinds to match
+/// If a match occurs, a reducer function can be run on the node.
+/// </summary>
+internal class MatchRule {
```
matouskozak, line 19: Do you think it would be beneficial to ask someone from the Roslyn team to review this (and maybe the previously introduced Parser)?

I think this is an area commonly found in the frontend of compilers and they may have some interesting insights for us if we plan to grow this further.

matouskozak (approved):
Thanks a lot for the great descriptions, it helps a lot in understanding the PR.

Good point on future work to make the `BuildMatchRules` static https://github.com/dotnet/runtimelab/issues/2691 (maybe we could include it as part of this PR already but unsure if there are some hidden obstacles that would make it too complex)

src/Swift.Bindings/src/Demangler/MatchRule.cs
```diff
@@ -0,0 +1,133 @@
...
+                return false;
+        }
+        return true;
+    }
+
+    /// <summary>
+    /// Creates a simple string representation of this rule
+    /// </summary>
+    /// <returns></returns>
+    public override string ToString() => Name;
```
mcumming, line 136: Wouold it help to have a DebuggerDisplayAttribute that surfaces this value?

src/Swift.Bindings/src/Demangler/Swift5Reducer.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        return Convert (node.Children [0]);
+    }
+
+    /// <summary>
+    /// Return a string in the form mangledName: expected xxxx but got yyyy
+    /// </summary>
+    string ExpectedButGot (string expected, string butGot)
+    {
+        return $"Demangling {mangledName}: expected {expected} but got {butGot}";
```
mcumming, line 122: Where do these strings surface? Should they be localized?

src/Swift.Bindings/src/Model/SwiftFunction.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// Returns a hashcode for the function
+    /// </summary>
+    /// <returns></returns>
+    public override int GetHashCode() => ToString ().GetHashCode ();
+
+    /// <summary>
+    /// Returns a string representation of the function
+    /// </summary>
+    /// <returns></returns>
+    public override string ToString () => $"{Provenance}.{Name}{ParameterList} -> {Return}";
```
mcumming, line 58: Again, any usefulness in having a DebuggerDisplayAttribute?

src/Swift.Bindings/tests/DemanglerTests/BasicDemanglingTests.cs
mcumming: I would think that in non-test code you would always switch on the type or check for a null result after the cast. `result as ProtocolWitnessTableReduction` should return `null` if `result` is of the `ReductionError` type

stephen-hawley: It would absolutely return null. The test is there to ensure that a symbol that is obviously not a Swift mangled symbol will correctly generate an error. Anything other cast will fail.

src/Swift.Bindings/src/Demangler/Swift5Reducer.cs
stephen-hawley: Eventually, they should be localized. I'm not concerned with that right now. However, putting the very common "Expected but got" pattern in one place makes that easier down the road.

src/Swift.Bindings/src/Model/Provenance.cs
```diff
@@ -0,0 +1,111 @@
...
+// Licensed under the MIT License.
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace BindingsGeneration;
+
+/// <summary>
+/// Represents the origin of a particular function
+/// </summary>
+public class Provenance {
```
kotlarmilos, line 11: Isn't this the same property as `public required BaseDecl? ParentDecl { get; set; }`?

src/Swift.Bindings/src/Model/SwiftFunction.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System.Diagnostics;
+
+namespace BindingsGeneration;
+
+/// <summary>
+/// Represents a Swift function signature and its provenance.
+/// </summary>
+[DebuggerDisplay("{ToString()}")]
+public class SwiftFunction {
```
kotlarmilos, line 12: Why can't we utilize MethodDecl and it's members instead?

kotlarmilos (approved):

dalexsoto (approved):

src/Swift.Bindings/src/Model/Provenance.cs
stephen-hawley: BaseDecl, IIRC is a C# representation not a Swift declaration. Further, extensions make this more complicated because you can have multiple extension on the same type from difference modules.

src/Swift.Bindings/src/Model/SwiftFunction.cs
stephen-hawley: See above. This is strictly Swift not C#.

PR #2694, opened 2024-09-27
[Swift Bindings] Added demangling support for protocol conformance descriptors and nested type names

stephen-hawley:
This PR makes protocol conformance descriptors available to the demangler.
Both protocol witness tables and protocol conformance descriptors share the same reduction rule because they are effectively the same in terms of the contents. Protocol conformance descriptors have one more child for the module.

Added a rule for the `Module` kind which will reduce to a `Provenance` reduction.

Refactored the `Type` reducer to include support for inner types.

Note on inner types in the tree: they are represented as a set of nodes that are built in a depth-first fashion. This lends itself to a recursive decoder.  For example, for the type `GenericTestFramework.Foo.Thingy`, the node structure looks like this:
```
      kind=Type
        kind=Class
          kind=Class
            kind=Module, text="GenericTestFramework"
            kind=Identifier, text="Foo"
          kind=Identifier, text="Thingy"
```

Set the default MatchContentType to `AlwaysMatch` since we don't have any nodes that require content matching (yet), so this seems like a better default and it was breaking the Module node reduction with the previous default.

### DIFF
--- a/src/Swift.Bindings/src/Demangler/IReduction.cs
+++ b/src/Swift.Bindings/src/Demangler/IReduction.cs
@@ -7,6 +7,9 @@ namespace BindingsGeneration.Demangling;
 /// Represents the result of reducing a Node or tree of nodes
 /// </summary>
 public interface IReduction {
+    /// <summary>
+    /// Returns a the mangled symbol associated with a reduction
+    /// </summary>
     string Symbol { get; init; }
 }
 
@@ -14,6 +17,9 @@ public interface IReduction {
 /// Represents an error in an attempted reduction
 /// </summary>
 public class ReductionError : IReduction {
+    /// <summary>
+    /// Returns a the mangled symbol associated with a reduction
+    /// </summary>
     public required string Symbol { get; init; }
     /// <summary>
     /// Returns an error message describing the error
@@ -25,6 +31,9 @@ public class ReductionError : IReduction {
 /// Represents a reduction that reduces to a single type
 /// </summary>
 public class TypeSpecReduction : IReduction {
+    /// <summary>
+    /// Returns a the mangled symbol associated with a reduction
+    /// </summary>
     public required string Symbol { get; init; }
     /// <summary>
     /// Returns a TypeSpec for the type that this node represents
@@ -36,6 +45,9 @@ public class TypeSpecReduction : IReduction {
 /// Represents a reduction that reduces to a Swift function
 /// </summary>
 public class FunctionReduction : IReduction {
+    /// <summary>
+    /// Returns a the mangled symbol associated with a reduction
+    /// </summary>
     public required string Symbol { get; init; }
     /// <summary>
     /// Returns a function that this type represents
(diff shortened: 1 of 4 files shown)


mandel-macaque (approved):

rolfbjarne (approved):

src/Swift.Bindings/src/Demangler/IReduction.cs
```diff
@@ -36,15 +45,69 @@ public class TypeSpecReduction : IReduction {
...
+        new ProvenanceReduction() { Symbol = symbol, Provenance = Provenance.TopLevel (moduleName) };
+    
+    /// <summary>
+    /// Factory method to construct an instance provenance reduction
+    /// </summary>
+    public static ProvenanceReduction Instance(string symbol, NamedTypeSpec instance) =>
+        new ProvenanceReduction() { Symbol = symbol, Provenance = Provenance.Instance (instance) };
+
+    /// <summary>
+    /// Factory method to construct an excetion provenance reduction
```
matouskozak:
```suggestion
    /// Factory method to construct an extension provenance reduction
```

Typo?

src/Swift.Bindings/src/Demangler/Swift5Reducer.cs
```diff
@@ -34,14 +35,17 @@ List<MatchRule> BuildMatchRules () =>
             Name = "Global", NodeKind = NodeKind.Global, Reducer = ConvertFirstChild
         },
         new MatchRule() {
-            Name = "ProtocolWitnessTable", NodeKind = NodeKind.ProtocolWitnessTable, Reducer = ConvertProtocolWitnessTable
+            Name = "ProtocolConformance", NodeKindList = new List<NodeKind>() { NodeKind.ProtocolWitnessTable, NodeKind.ProtocolConformanceDescriptor, } , Reducer = ConvertProtocolWitnessTable
```
matouskozak: Wouldn't be better to rename the `ConvertProtocolWitnessTable` to something like `ConvertProtocolConformance` since now we use it for more than just witness tables?

src/Swift.Bindings/src/Demangler/Swift5Reducer.cs
```diff
@@ -83,7 +91,21 @@ IReduction ConvertProtocolWitnessTable (Node node, string? name)
...
+        if (isDescriptor) {
+            grandchild = Convert (child.Children [2]);
+            if (grandchild is ProvenanceReduction prov) {
+                if (!prov.Provenance.IsTopLevel)
+                    return ReductionError (ExpectedButGot ("A top-level module name", prov.Provenance.ToString()));
+                return new ProtocolConformanceDescriptorReduction() { Symbol = mangledName, ImplementingType = impNamed, ProtocolType = protoNamed, Module = prov.Provenance.Module};
+            } else if (grandchild is ReductionError) {
+                return grandchild;
+            } else {
+                return ReductionError (ExpectedButGot ("ProvenanceReduction", grandchild.GetType().Name));
```
matouskozak, line 103:
```suggestion
            } else {
                return ReductionError (ExpectedButGot ("ProvenanceReduction", grandchild.GetType().Name));
```

Would it be possible to unify this into `else` branch since we returning propagating `ReductionError` anyways.

src/Swift.Bindings/src/Demangler/Swift5Reducer.cs
```diff
@@ -92,18 +114,63 @@ IReduction ConvertProtocolWitnessTable (Node node, string? name)
...
+            var typeName = FristChildIdentifierText (node);
+            sb.Append (moduleName).Append ('.').Append (typeName);
         }
-        return ReductionError(ExpectedButGot("Class/Struct/Enum/Protocol", kind.ToString()));
+    }
+
+    /// <summary>
+    /// Returns the text of the first child of a node if and only if that child is an Identifier, else throw
+    /// </summary>
+    string FristChildIdentifierText (Node node)
```
matouskozak:
```suggestion
    string FirstChildIdentifierText (Node node)
```

matouskozak (commented):
Thank you for the example for how the inner types are represented. If I understood them correctly these are the ones that Swift refers to as Nested Types https://docs.swift.org/swift-book/documentation/the-swift-programming-language/nestedtypes/, am I right?

src/Swift.Bindings/src/Demangler/Swift5Reducer.cs
stephen-hawley: You don't know what error has come back from the previous call. By collapsing it, you're losing all information about the failure. Best to pass it back unchanged.

src/Swift.Bindings/src/Demangler/Swift5Reducer.cs
```diff
@@ -83,7 +91,21 @@ IReduction ConvertProtocolWitnessTable (Node node, string? name)
         var impNamed = (NamedTypeSpec)implementingType.TypeSpec;
         var protoNamed = (NamedTypeSpec)protocolType.TypeSpec;
 
-        return new ProtocolWitnessTableReduction() { Symbol = mangledName, ImplementingType = impNamed, ProtocolType = protoNamed };
+        if (isDescriptor) {
+            grandchild = Convert (child.Children [2]);
```
kotlarmilos, line 95: Is this safe? Specifically `child.Children [2]`.

kotlarmilos (approved):

src/Swift.Bindings/src/Demangler/Swift5Reducer.cs
stephen-hawley: Guaranteed. If it's a descriptor, that has to be there.

dalexsoto (approved):

PR #2697, opened 2024-10-02
[NativeAOT-LLVM] Implement stack traces on WASI via precise virtual unwinding

SingleAccretion:
## The problem

To support stack traces on WASI, where no native API for virtual unwinding of any sort is available, we need to implement our own virtual unwinding capable of enumerating each active managed frame on the stack. This system need not be IP-level precise, since we don't need to provide line numbers.

## The solution

It may be noticed that we already have a capability for virtual unwinding through frames that contain EH constructs, in the form of the "virtual unwind frames" linked list. However, this list is stored in TLS, and we can do better, assuming all frames are going to be linked.

The core idea behind the solution implemented in this change, termed "precise virtual unwinding", is that we know the size of all shadow frames. Therefore, if we store a fragment of metadata ("precise virtual unwind info") in each method's shadow frame in a way that it can be accessed by any callee, we can unwind through frames by decoding the shadow frame size and incrementing our current shadow stack pointer to reach the next frame.
```
void ManagedMethod(void** pShadowStack) {
    // This way, the callee can access the unwind info by doing "pShadowStack[-1]".
    pShadowStack[ShadowFrameSize - 1] = pUnwindInfo;
}
```
While this basic idea is very simple, this change is not so simple. Complications:
1) We don't want to maintain two parallel virtual unwind frame lists for efficiency reasons. With precise virtual unwinding, the EH frames are integrated into the precise virtual unwind ones and TLS is not used. We re-term the EH-only scheme "sparse virtual unwinding" (since it doesn't capture all methods).
2) We don't want various compiler-generated helpers to show up in stack traces, since, unlike with native schemes, the only artifact identifying the frame we have is the precise virtual unwind info, which cannot be symbolicated. Additionally, we don't want to introduce precise virtual frames in compiler-generated helpers (like instantiating thunks) that would otherwise have a zero-sized shadow frame. The solution is to include metadata in the precise unwind info that would mark the frame as "hidden". As a further optimization, we can hide the frames by using the shadow frame size _itself_ as the unwind info, provided it is small enough.
3) To support various APIs that need to convert between stack trace IPs and function pointers, we include the associated function pointer into unwind info. This is complicated by the fact that both the IP -> FP and FP -> IP directions need to be supported, as well as the fact that this function pointer may need to be the boxed version of the entrypoint (recall that instance methods on value types have two entrypoints: boxed (unboxing stub) and unboxed (real method body)).
4) We further optimize our code generation by noticing that there is no need to unwind through frames that have no callees (e. g. empty methods, or things like `return true;`). Even so, we may still associate "dummy" unwind info with these methods in case they are delegate targets, to support the `DiagnosticMethodInfo.Create(Delegate)` API.

## The result

We have fully-featured virtual unwinding capable of implementing all stack trace APIs and features with 100% fidelity, at the cost of introducing one pointer-sized store of a constant into the prolog of most methods (about 3% overall code size increase).

## The future

It can be noticed that I have chosen to decouple the notion of precise virtual unwinding from the target OS - one can enable (via undocumented switches) the former when targeting browser. While the primary motivation for this was ease of debugging (Chrome debugger is currently better than wasmtime+lldb), another was the possibility of using precise virtual unwinding to experiment with precise GC.


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvm.h
```diff
@@ -181,6 +190,19 @@ enum class ValueInitKind
     Uninit
 };
 
+enum CallSiteFacts
+{
+    CSF_NONE = 0,
+    CSF_NO_THROW = 1,
+    CSF_NO_VIRTUAL_UNWIND = 1 << 1,
+    CSF_FUNCLET = 1 << 16,
```
yowl, line 198: What is the reason this is `16` ?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/Target_Wasm/WasmPreciseVirtualUnwindInfoNode.cs
```diff
@@ -0,0 +1,258 @@
...
+            if (shadowFrameSizeInSlots >= SmallShadowFrameSizeInSlotsLimit)
+            {
+                // Only a very small proportion (< 1%) of methods need to use this 'large' format.
+                builder.EmitUInt(shadowFrameSizeInSlots);
+            }
+            if (hasFunctionPointer)
+            {
+                // For instance methods on value types, we want the unboxing stub as that's what reflection uses.
+                // The delegate case ('2') may need to map back both entrypoints; that too requires access to
+                // the unboxing version - we can map it to the its target at runtime.
```
yowl:
```suggestion
                // the unboxing version - we can map it to its target at runtime.

File: src/coreclr/tools/aot/ILCompiler.RyuJit/ILCompiler.RyuJit.csproj [hunk 3]
```diff
@@ -103,8 +103,8 @@
   <ItemGroup>
     <Compile Include="JitInterface\CorInfoImpl.Llvm.cs" />
     <Compile Include="JitInterface\CorInfoImpl.Llvm.DebugInfo.cs" />
-    <Compile Include="..\..\Common\JitInterface\CorJitApiId.Shared.cspp">
-      <Link>JitInterface\CorJitApiId.Shared.cspp</Link>
+    <Compile Include="..\..\Common\JitInterface\JitEEApi.Shared.cspp">
```
yowl, line 106: Is this name change something that is already upstream?

yowl (commented):
Thanks, looks like a well thought out solution.

src/coreclr/jit/llvm.h
SingleAccretion: This is a bit of a "fake" fact (for skipping checking), so it is separated from the rest in this soft fashion.

src/coreclr/tools/aot/ILCompiler.RyuJit/ILCompiler.RyuJit.csproj
SingleAccretion: This is a downstream-only file (this whole section is), so there is no need to coordinate.

yowl (approved):
Nice, thanks!

Issue #2580, opened 2024-05-08
Review projections of Swift generics in C# for CryptoKit dev template


kotlarmilos:
## Description

This proposal discusses the projection of Swift generics in C#. The primary goal is to validate the design of projections and integrate CryptoKit APIs for direct invocation from the dotnet/runtime repository.  To limit the scope, this discussion specifically focuses on generics use-cases for the CryptoKit dev template. Protocols are considered as constraints within the Swift runtime, which should not affect functionality and they are not covered in this proposal.

## Generics in Swift functions

In Swift, when a generic type is used as a parameter, its memory layout might not be known until runtime. To manage this, the generic parameter is represented by an opaque pointer. For object allocation, the runtime utilizes init functions from the value witness table using a metadata pointer. This metadata pointer is included as an implicit argument at the end of the function's signature, enabling the runtime to correctly allocate and manage generic types. When returning a generic type, the runtime utilizes [indirect return result](https://github.com/dotnet/runtime/issues/100543) mechanism.

Here is a code snippet that illustrates how generics are handled in functions with the corresponding assembly code for arm64:
```swift
func returnData<T>(data: T) -> T {
    return data
}
```

As the function returns a generic type, the register `x8` is expected to contain the return buffer where the value should be stored.  The first parameter `x0` is an opaque pointer; the second parameter `x1` is the metadata pointer of the type.
```
00000000000079b0	sub	sp, sp, #0x30
00000000000079b4	stp	x29, x30, [sp, #0x20]
00000000000079b8	add	x29, sp, #0x20
00000000000079bc	str	x8, [sp]
```
This block shuffles regs for an object allocation. At the end, the `x0` is the return buffer, `x1` is the opaque pointer, and `x2` is the metadata pointer.
```
00000000000079c0	mov	x2, x0
00000000000079c4	ldr	x0, [sp]
00000000000079c8	str	x2, [sp, #0x8]
00000000000079cc	mov	x2, x1
00000000000079d0	ldr	x1, [sp, #0x8]
00000000000079d4	str	xzr, [sp, #0x10]
00000000000079d8	mov	x8, x2
00000000000079dc	stur	x8, [x29, #-0x8]
00000000000079e0	mov	x8, x1
```
[...]

PR #2704, opened 2024-10-07
Add experimental Swift bindings for CryptoKit

kotlarmilos:
## Description

This PR introduces experimental Swift bindings for CryptoKit. It validates the Swift interop and demonstrates the Swift bindings surface for end-users.

The bindings are manually created as a proof of concept. In the future, the bindings should be reduced, and they should be automatically generated by the projection tooling

Current limitations:
 - Protocol descriptor symbols required for protocol witness tables are missing in the JSON ABI file
 - https://github.com/dotnet/runtime/issues/107946

Resolves https://github.com/dotnet/runtimelab/issues/2580

### DIFF
--- a/src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
+++ b/src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
@@ -1,0 +1,274 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Reflection;
+
+namespace Test
+{
+    /// <summary>
+    /// Represents a Swift type in C#.
+    /// </summary>
+    public unsafe interface ISwiftObject
+    {
+        public static abstract void* Metadata { get; }
+    }
+
+    // <summary>
+    // Represents Swift UnsafePointer in C#.
+    // </summary>
+    public readonly unsafe struct UnsafePointer<T> where T : unmanaged
+    {
+        private readonly T* _pointee;
+        public UnsafePointer(T* pointee)
+        {
+            this._pointee = pointee;
+        }
+
+        public T* Pointee => _pointee;
+
+        public static implicit operator T*(UnsafePointer<T> pointer) => pointer.Pointee;
+
+        public static implicit operator UnsafePointer<T>(T* pointee) => new(pointee);
+    }
+
+    // <summary>
+    // Represents Swift UnsafeMutablePointer in C#.
+    // </summary>
+    public readonly unsafe struct UnsafeMutablePointer<T> where T : unmanaged
+    {
+        private readonly T* _pointee;
+        public UnsafeMutablePointer(T* pointee)
+        {
+            _pointee = pointee;
+        }
+
+        public T* Pointee => _pointee;
+
+        public static implicit operator T*(UnsafeMutablePointer<T> pointer) => pointer.Pointee;
+
+        public static implicit operator UnsafeMutablePointer<T>(T* pointee) => new(pointee);
+    }
+
+    // <summary>
+    // Represents Swift UnsafeRawPointer in C#.
+    // </summary>
+    public readonly unsafe struct UnsafeRawPointer
+    {
+        private readonly void* _pointee;
+        public UnsafeRawPointer(void* pointee)
+        {
+            _pointee = pointee;
+        }
+
+        public void* Pointee => _pointee;
+
+        public static implicit operator void*(UnsafeRawPointer pointer) => pointer.Pointee;
+
+        public static implicit operator UnsafeRawPointer(void* pointee) => new(pointee);
+    }
+
+    // <summary>
+    // Represents Swift UnsafeMutableRawPointer in C#.
+    // </summary>
+    public readonly unsafe struct UnsafeMutableRawPointer
+    {
+        private readonly void* _pointee;
+        public UnsafeMutableRawPointer(void* pointee)
+        {
+            _pointee = pointee;
+        }
+
+        public void* Pointee => _pointee;
+
+        public static implicit operator void*(UnsafeMutableRawPointer pointer) => pointer.Pointee;
+
+        public static implicit operator UnsafeMutableRawPointer(void* pointee) => new(pointee);
+    }
+
+    // <summary>
+    // Represents Swift UnsafeBufferPointer in C#.
+    // </summary>
+    public readonly unsafe struct UnsafeBufferPointer<T> where T : unmanaged
+    {
+        private readonly T* _baseAddress;
+        private readonly nint _count;
+        public UnsafeBufferPointer(T* baseAddress, nint count)
+        {
+            _baseAddress = baseAddress;
+            _count = count;
+        }
+
+        public T* BaseAddress => _baseAddress;
+        public nint Count => _count;
+    }
+
+    // <summary>
+    // Represents Swift UnsafeMutableBufferPointer in C#.
+    // </summary>
+    public readonly unsafe struct UnsafeMutableBufferPointer<T> where T : unmanaged
+    {
+        private readonly T* _baseAddress;
+        private readonly nint _count;
+        public UnsafeMutableBufferPointer(T* baseAddress, nint count)
+        {
+            _baseAddress = baseAddress;
+            _count = count;
+        }
+
+        public T* BaseAddress => _baseAddress;
+        public nint Count => _count;
+    }
+
+    // <summary>
+    // Represents Swift Foundation.Data in C#.
+    // </summary>
+    [StructLayout(LayoutKind.Sequential, Size = 16)]
+    [InlineArray(16)]
+    public unsafe struct Data : ISwiftObject
+    {
+        private byte _payload;
+
+        public unsafe Data(UnsafeRawPointer pointer, nint count)
+        {
+            this = Foundation.PInvoke_Data_InitWithBytes(pointer, count);
+        }
+
+        public byte Payload => _payload;
+
+        public readonly nint Count => Foundation.PInvoke_Data_GetCount(this);
+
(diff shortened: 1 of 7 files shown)


src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    // <summary>
+    // Represents Swift Foundation.Data in C#.
+    // </summary>
+    [StructLayout(LayoutKind.Sequential, Size = 16)]
+    [InlineArray(16)]
+    public unsafe struct Data : ISwiftObject
+    {
+        private byte _payload;
```
jkurdek, line 132: Until those bits flow https://github.com/dotnet/runtime/pull/108483, this will fail on mono

kotlarmilos: Currently, we test only CoreCLR.

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
```diff
@@ -0,0 +1,274 @@
...
+            public uint ExtraInhabitantCount;
+            public int Size => _Size.ToInt32();
+            public int Stride => _Stride.ToInt32();
+            public int Alignment => (int)((Flags & ValueWitnessFlags.AlignmentMask) + 1);
+            public bool IsNonPOD => Flags.HasFlag (ValueWitnessFlags.IsNonPOD);
+            public bool IsNonBitwiseTakable => Flags.HasFlag (ValueWitnessFlags.IsNonBitwiseTakable);
+            public bool HasExtraInhabitants => ExtraInhabitantCount != 0;
+        }
+
+        public static unsafe void* GetMetadata<T>(T type) where T: ISwiftObject
```
stephen-hawley, line 241: Does this need to be generic?
```csharp
public static unsafe void* GetMetadata<T>(ISwiftObject obj) {
    return obj.Metadata;
}
```
While you don't use it here, we will need a more general version of this that takes the C# `Type` as an argument and returns the type metadata from that so that we can get the type metadata for the full scope of Swift projectable types.

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
```diff
@@ -0,0 +1,274 @@
...
+            }
+            catch (Exception ex)
+            {
+                throw new InvalidOperationException($"Failed to get conformance descriptor for symbol: {symbol}", ex);
+            }
+            finally
+            {
+                if (handle != IntPtr.Zero)
+                {
+                    NativeLibrary.Free(handle);
```
stephen-hawley, line 269: Two things:
1 - will this free an already loaded library?
2 - you don't need an IntPtr.Zero check, this is documented as doing nothing on IntPtr.Zero

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
```diff
@@ -0,0 +1,274 @@
...
+            void* valueWitnessTable = (void*)Marshal.ReadIntPtr((IntPtr)metadata, -IntPtr.Size);
+            return valueWitnessTable;
+        }
+
+        public static unsafe void* GetConformanceDescriptor(string symbol)
+        {
+            IntPtr handle = IntPtr.Zero;
+            try
+            {
+                handle = NativeLibrary.Load(Foundation.Path);
```
stephen-hawley, line 257: This will only work for conformance descriptors that are in Foundation - there will be other modules that have conformance descriptors. We should open an issue on this so it doesn't get lost.

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
```diff
@@ -0,0 +1,274 @@
...
+using System.Reflection;
+
+namespace Test
+{
+    /// <summary>
+    /// Represents a Swift type in C#.
+    /// </summary>
+    public unsafe interface ISwiftObject
+    {
+        public static abstract void* Metadata { get; }
```
stephen-hawley, line 16: For the future - I'd really like to see Metadata be its own type - a struct containing either an IntPtr, a NativeHandle, or a void*. There will be sufficient funcs that will have Metadata as arguments that we are likely to have signature conflicts. Let's open up an issue for this.

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
```diff
@@ -0,0 +1,274 @@
...
+
+        public static implicit operator T*(UnsafePointer<T> pointer) => pointer.Pointee;
+
+        public static implicit operator UnsafePointer<T>(T* pointee) => new(pointee);
+    }
+
+    // <summary>
+    // Represents Swift UnsafeMutablePointer in C#.
+    // </summary>
+    public readonly unsafe struct UnsafeMutablePointer<T> where T : unmanaged
```
stephen-hawley, line 40: We don't need to correct this now, but this code will cause memory leaks without it honoring the Swift memory model. Please open an issue for this.

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
```diff
@@ -0,0 +1,274 @@
...
+        public T* BaseAddress => _baseAddress;
+        public nint Count => _count;
+    }
+
+    // <summary>
+    // Represents Swift Foundation.Data in C#.
+    // </summary>
+    [StructLayout(LayoutKind.Sequential, Size = 16)]
+    [InlineArray(16)]
+    public unsafe struct Data : ISwiftObject
```
stephen-hawley, line 130: Not important right now, but in the future we will also want Data to implement friendlier C# interfaces (IEnumerable<byte> or ICollection<byte> etc). Let's open an issue for that.

stephen-hawley (changes requested):

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
kotlarmilos: It is convenient for retrieval. Alternatively, it can be retrieved using reflection:
```csharp
        public static unsafe void* GetMetadata(Type t)
        {
            if (typeof(ISwiftObject).IsAssignableFrom(t))
            {
                var metadataProperty = t.GetProperty("Metadata", System.Reflection.BindingFlags.Static | System.Reflection.BindingFlags.Public);
                if (metadataProperty != null)
                {
                    var metadataValue = metadataProperty.GetValue(null);
                    if (metadataValue is IntPtr ptr)
                    {
                        return ptr.ToPointer();
                    }
                }
                throw new InvalidOperationException($"Type {t.Name} does not have a 'Metadata' property.");
            }
            throw new ArgumentException($"Type {t.Name} does not implement ISwiftObject.");
        }
```

I believe we should consider all use cases before making such changes, but I'm also open to refactoring it now. What do you prefer?

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
kotlarmilos: The intention is to free the loaded library only if `NativeLibrary.GetExport` fails, which is why `handle != IntPtr.Zero` is used.

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
stephen-hawley: This is going to take some thought, but I think the main entry point should be something like:
```
public static bool TryGetMetadata (Type t, out Metadata metadata)
{
    // one catch - t must not be an unbound generic.
    if (typeof(ISwiftObject).IsAssignableFrom(t)) {
        metadata = GetMetadataFromISwiftObject(t);
        return true;
    }
    // over time we add the cases we need for tuples, nint, nuint, nfloat, etc.
    // I'd like to see this ultimately be a ladder of `else if` constructs for each major type ordered by
    // most common and/or cheapest predicates. For example, identifying certain scalars should be
    // a matter of looking at `t.GetTypeCode()`
    return false;
}

static Metadata GetMetadataFromISwiftObject(Type t) // t is guaranteed to be ISwiftObject compatible 
{
    // your reflection code - no try/get pattern needed, this is a private API
}

kotlarmilos: Yes, let's talk more about this. Created a tracking issue: https://github.com/dotnet/runtimelab/issues/2705

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
kotlarmilos: Created a tracking issue: https://github.com/dotnet/runtimelab/issues/2705

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
kotlarmilos: Created a tracking issue: https://github.com/dotnet/runtimelab/issues/2705

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
kotlarmilos: Added as a task `Review memory management of the projected pointer types` in https://github.com/dotnet/runtime/issues/95633. Plan to collect more details and open a tracking issue.

src/Swift.Bindings/tests/CryptoKit/Foundation.Source.cs
kotlarmilos: Added as a task `Implement common C# interfaces (IEnumerable, ICollection, etc)` in https://github.com/dotnet/runtime/issues/95633. Plan to collect more details and open a tracking issue.

stephen-hawley (approved):

PR #2707, opened 2024-10-09
[SwiftBindings] Documented binding PATs and alternative to vtables

stephen-hawley:
This PR adds documentation for how to shape protocols with associated types and how to call C# directly from Swift.

Updated other documents to refer to calling C# directly.

### DIFF
--- a/src/docs/binding-classes.md
+++ b/src/docs/binding-classes.md
@@ -396,3 +396,5 @@ public class ValBool : SwiftNativeObject {
     }    
 }
 ```
+
+It should be noted that there is a mechanism that can be used to call the C# receivers directly from Swift. This is detailed in [this document about an alternative to vtables](vtable-alternative.md).
--- a/src/docs/binding-pats.md
+++ b/src/docs/binding-pats.md
@@ -2,7 +2,7 @@
 
 Swift PATs are an entity that do not have a direct analog in C# and as such can feel very foreign to C# programmers. The main issue with PATs is that they are not actual types, at least not in the sense of the other types in Swift and are not simply generic interfaces.
 
-[This is a great video](https://www.youtube.com/watch?v=XWoNjiSPqI8) about PATs. Even though it's based on an earlier version of Swift syntax, it still applies. When looking at the video, you'll see that he uses the keyword `typealias`, but in current Swift they use `associatedtype` instead. There is a paper which is referenced in this video [here](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/C97D5964ECC2E651EEF9A70BC50600A6/S0956796806006198a.pdf/an_extended_comparative_study_of_language_support_for_generic_programming.pdf).
+[This is a great video](https://www.youtube.com/watch?v=XWoNjiSPqI8) about PATs. Even though it's based on an earlier version of Swift syntax, it still applies. When looking at the video, you'll see that the presenter uses the keyword `typealias`, but in current Swift they use `associatedtype` instead. There is a paper which is referenced in this video [here](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/C97D5964ECC2E651EEF9A70BC50600A6/S0956796806006198a.pdf/an_extended_comparative_study_of_language_support_for_generic_programming.pdf).
 
 The easiest way to grasp onto how PATs work is to look at very simple examples. One such example is checking for equality. Consider this PAT:
 
@@ -106,6 +106,8 @@ Obviously, C# doesn't have PATs and the closest we can get is to use generic int
 
 # ABI Differences
 
+Other than the typical Swift function call ABI differences common to all types that have functions, there are no ABI differences unique to PATs.
+
 # Runtime Differences
 
 Because PATs inherently require a generic declaration, it means that the implementation details of the type comes along separately.
@@ -116,7 +118,7 @@ public func areTangent<T: TangentialProto>(a: T, b: T) -> Bool { // TangentialPr
    // ...
 }
 ```
-there are two extra implicit arguments added to the function: the type metadata for `T` and the protocol witness table for `T` with respect to `TangentialProto`. This is also the case with PATs except that the protocol witness table is not necessarily known at compile time because. Swift therefore has a data structure called a Protocol Conformance Descriptor which describes the shape of the protocol witness table and given the type metadata for the associated type(s) can generate a protocol witness table for that specialization. There is some discussion from Apple on that matter in [this forum post](https://forums.swift.org/t/need-help-understanding-protocols-and-generics/37564/35).
+there are two extra implicit arguments added to the function: the type metadata for `T` and the protocol witness table for `T` with respect to `TangentialProto`. This is also the case with PATs except that the protocol witness table is not necessarily known at compile time because the associated types effectively creates an unbound generic type. Swift therefore has a data structure called a Protocol Conformance Descriptor which describes the shape of the protocol witness table and given the type metadata for the associated type(s) can generate a protocol witness table for that specialization. There is some discussion from Apple on that matter in [this forum post](https://forums.swift.org/t/need-help-understanding-protocols-and-generics/37564/35).
 
 
 # Accessibility
@@ -127,6 +129,7 @@ To start to understand the binding of PATs (protocols with associated types), it
 
 ```swift
 public protocol IteratorProtocol {
+    associatedtype Element
     mutating func next() -> Element?
 }
 ```
--- a/src/docs/binding-pats.md
+++ b/src/docs/binding-pats.md
@@ -1,0 +1,408 @@
+# Binding Protocols With Associated Types (PATs)
+
+Swift PATs are an entity that do not have a direct analog in C# and as such can feel very foreign to C# programmers. The main issue with PATs is that they are not actual types, at least not in the sense of the other types in Swift and are not simply generic interfaces.
+
+[This is a great video](https://www.youtube.com/watch?v=XWoNjiSPqI8) about PATs. Even though it's based on an earlier version of Swift syntax, it still applies. When looking at the video, you'll see that he uses the keyword `typealias`, but in current Swift they use `associatedtype` instead. There is a paper which is referenced in this video [here](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/C97D5964ECC2E651EEF9A70BC50600A6/S0956796806006198a.pdf/an_extended_comparative_study_of_language_support_for_generic_programming.pdf).
+
+The easiest way to grasp onto how PATs work is to look at very simple examples. One such example is checking for equality. Consider this PAT:
+
+```swift
+public protocol Sameness {
+    func isSame(as: Self) -> Bool
+}
+```
+
+In this case `Self` is a special placeholder in Swift for "the type which has implemented this protocol". While not labeled as such, it *is* an associated type and in Swift. Apple's documentation on `Self` is [here](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/types#Self-Type).
+
+In adopting this protocol in Swift, you could do something like this:
+
+```swift
+public struct Imaginary : Sameness {
+    private var r, i: Double
+    public init (r: Double, i: Double) {
+        self.r = r
+        self.i = i
+    }
+    public func isSame(as: Self) {
+        return r == as.r && i == as.i
+    }
+}
+
+public func printSame<T:Sameness>(a: T, b: T) {
+    let same = a.isSame(as: b)
+    if same {
+        print("a and b are the same.")
+    } else {
+        print("a and b are different.")
+    }
+}
+```
+
+Note that `printSame` declares neither `a` nor `b` as `Sameness` but instead uses a generic type constrained to `Sameness`. This is because `Sameness` is **not** an actual type. One way to look at a PAT is not so much as a type but rather a type-shaped hole.
+
+And while it looks like this could be represented in C# as:
+
+```csharp
+public interface ISameness<T> {
+    bool IsSame(T @as);
+}
+```
+It's really closer to this:
+```csharp
+public interface ISameness<TSelf> where TSelf: ISameness<TSelf> {
+    bool IsSame(TSelf @as);
+}
+```
+
+This only gets more complicated as associated types are added and as those associated types have constraints on them. This is because the generic relationships and constraints are encapsulted in the PAT whereas in other languages the generic relationships and constraints need to be part of the declaration which uses it.
+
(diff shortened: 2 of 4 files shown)


src/docs/vtable-alternative.md
```diff
@@ -0,0 +1,92 @@
...
+```
++- Module
+  |
+  +- Headers
+  | |
+  | +- csharpdeclarations.h
+  |
+  +- module.map
+```
+The swift compiler can interpret the decalrations.
```
matouskozak:
```suggestion
The swift compiler can interpret the declarations.
```

src/docs/vtable-alternative.md
```diff
@@ -0,0 +1,92 @@
...
+
+Here's a module.map that will wotk in this example:
```
matouskozak:
```suggestion
Here's a module.map that will work in this example:
```

src/docs/vtable-alternative.md
```diff
@@ -0,0 +1,92 @@
...
+
+#ifdef __cplusplus
+}
+#endif
+```
+And in a C# file, we have:
+```csharp
+public static class ThisIsNotImportant
+{
+    [UnmanagedCallersOnly (EntryPoint = "csGetAgeImpl)]
```
matouskozak:
```suggestion
    [UnmanagedCallersOnly (EntryPoint = "csGetAgeImpl")]
```

src/docs/vtable-alternative.md
```diff
@@ -0,0 +1,92 @@
...
+2. a C# vtable to hold delegates for the Receivers that will be marshalled as unmanaged functions pointers. This definition lives in a separate static class since Receivers can't live inside of a (possibly) generic type
+3. a C# vtable to hold managed delegates to the actual implementation which lives in the bound type.
+
+As you can imagine, it can be painful to keep these types in sync.
+
+C# must call into Swift in order to initialize the vtable. In addition, if the type being bound is generic, then there needs to be a vtable for each unique specialization of the type.
+
+Because the population of the vtable is done at runtime, the calling Swift code is required to test the function pointer for null, even though it should never be null.
+
+Most of these issues can be solved if Swift can call C# exported functions directly. This can be done if the binding library is compiled with NativeAOT, generating a dylib.
```
matouskozak, line 16: Does this mean that we have to constrain the code projection tooling generates for `[UnmanagedCallersOnly]` to be NativeAOT friendly (e.g., no use of reflection)?

src/docs/binding-classes.md
```diff
@@ -395,4 +395,6 @@ public class ValBool : SwiftNativeObject {
         }
     }    
 }
-```
\ No newline at end of file
+```
+
+It should be noted that there is a mechanism that can be used to call the C# receivers directly from Swift. This is detailed in [this document about an alternative to vtables](vtable-alternative.md).
```
matouskozak, line 400: Nit. sometimes the "receivers" are written with first "r" uppercase and sometimes not. I think lowercase is fine overall.

src/docs/binding-pats.md
```diff
@@ -0,0 +1,408 @@
+# Binding Protocols With Associated Types (PATs)
+
+Swift PATs are an entity that do not have a direct analog in C# and as such can feel very foreign to C# programmers. The main issue with PATs is that they are not actual types, at least not in the sense of the other types in Swift and are not simply generic interfaces.
+
+[This is a great video](https://www.youtube.com/watch?v=XWoNjiSPqI8) about PATs. Even though it's based on an earlier version of Swift syntax, it still applies. When looking at the video, you'll see that he uses the keyword `typealias`, but in current Swift they use `associatedtype` instead. There is a paper which is referenced in this video [here](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/C97D5964ECC2E651EEF9A70BC50600A6/S0956796806006198a.pdf/an_extended_comparative_study_of_language_support_for_generic_programming.pdf).
```
matouskozak:
```suggestion
[This is a great video](https://www.youtube.com/watch?v=XWoNjiSPqI8) about PATs. Even though it's based on an earlier version of Swift syntax, it still applies. When looking at the video, you'll see that the presenter uses the keyword `typealias`, but in current Swift they use `associatedtype` instead. There is a paper which is referenced in this video [here](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/C97D5964ECC2E651EEF9A70BC50600A6/S0956796806006198a.pdf/an_extended_comparative_study_of_language_support_for_generic_programming.pdf).
```

src/docs/binding-pats.md
```diff
@@ -0,0 +1,408 @@
...
+        animal.Feed(food);
+    }
+}
+```
+
+# Language Parity Mismatches
+
+Obviously, C# doesn't have PATs and the closest we can get is to use generic interfaces. In addition, there is a language feature in Swift which is the keyword `some` which is used for return types that are PATs without having to specify the generic in the declaration. Think of it this was: in C# if you want to return a generic type from a method, it is the caller that determines what the expected return type will be. In Swift, `some` allows the *callee* not the caller to determine the specialization of the PAT. There is a way to model this in C#, and I'll explain that later.
+
+# ABI Differences
```
matouskozak, line 107: looks like it is missing content for `#ABI Differences` section, not sure if intentional?

src/docs/binding-pats.md
```diff
@@ -0,0 +1,408 @@
...
+
+Because PATs inherently require a generic declaration, it means that the implementation details of the type comes along separately.
+
+Recall that if you have this function in Swift:
+```swift
+public func areTangent<T: TangentialProto>(a: T, b: T) -> Bool { // TangentialProto is NOT a PAT
+   // ...
+}
+```
+there are two extra implicit arguments added to the function: the type metadata for `T` and the protocol witness table for `T` with respect to `TangentialProto`. This is also the case with PATs except that the protocol witness table is not necessarily known at compile time because. Swift therefore has a data structure called a Protocol Conformance Descriptor which describes the shape of the protocol witness table and given the type metadata for the associated type(s) can generate a protocol witness table for that specialization. There is some discussion from Apple on that matter in [this forum post](https://forums.swift.org/t/need-help-understanding-protocols-and-generics/37564/35).
```
matouskozak: *This is also the case with PATs except that the protocol witness table is not necessarily known at compile time because [something].* 

I think something is missing here :)

src/docs/binding-pats.md
```diff
@@ -0,0 +1,408 @@
...
+
+We accomplish this with proxy types. A proxy type is a type that can stand in the place of another type. This is an example of the [Gang of Four](https://en.wikipedia.org/wiki/Design_Patterns) pattern "[Adapter](https://en.wikipedia.org/wiki/Adapter_pattern)".
+
+Every proxy in C# inherits from `BaseAssociatedTypeProxy`. This type inherits from `SwiftNativeObject`, but it is designed to be initialized from either a heap allocated Swift object or a Swift value type. This type manages the lifespan of the underlying object.
+
+The proxies we create for PATs inherit from this type and are **dual purpose**. This means that it can act as a proxy for the C# type to Swift or the proxy for the Swift type to C#.
+
+Let's consider the proxying of a Swift object first.
+
+C# gets presented a generic Swift instance that implements `IteratorProtocol`. To work with it, we construct an instance of `SwiftIteratorProtocolProtocol<ATElement>` which implements the interface `ISwiftIteratorProtocol<ATElement>`.
```
matouskozak, line 169: We should try to think of simpler naming scheme for the projections in the future. The `SwiftIteratorProtocolProtocol` and the following xamarin wrappers e.g. `xamarin_XamWrappingFxamarin_static_wrapper_ProtocolTests_IteratorProtocol_next00000000` look a bit scary to me :)

src/docs/binding-pats.md
```diff
@@ -0,0 +1,408 @@
...
+
+This will have been constructed with some type defined in Swift.
+To call the `next` method, we call this wrapper:
+
+```swift
+public func xamarin_static_wrapper_IteratorProtocol_next<T0, T1>(this: inout T0) ->
+    Optional<T1> where T0 : IteratorProtocol, T1 == T0.Element
+{
+    return this.next();
+}
```
matouskozak, line 179: I see that the way BTfS worked with PATs it created a lot of wrappers/proxies in Swift. Could you please share what are the limitations of C# or why is it better to use Swift for the wrappers/proxies? 

I suppose our goal is to have as much of the bindings in C#, so eventually, we could try to add more runtime/language support if it would simplify the projections.

matouskozak (commented):
Thanks a lot Steve for the extended PAT description with examples, I'm learning about Swift with every PR 👍

src/docs/binding-pats.md
```diff
@@ -0,0 +1,408 @@
...
+
+# Accessibility
+
+This is an excerpt from a document I wrote about how PATs are bound in Binding Tools for Swift. It should be noted that with NativeAOT we don't need the vtable, as outlined [here](vtable-alternative.md).
+
+To start to understand the binding of PATs (protocols with associated types), it's easiest to start with a very simple PAT, `IteratorProtocol`:
+
+```swift
+public protocol IteratorProtocol {
+    mutating func next() -> Element?
```
kotlarmilos, line 133: Should it include `associatedtype Element` definition?

src/docs/binding-pats.md
```diff
@@ -0,0 +1,408 @@
...
+
+Next is the `SwiftProtocolTypeAttribute` which, given an interface, tells us what the name of the C# proxy type is, the name of the library where the protocol descriptor is defined and the symbol name of that descriptor and whether or not this is a PAT.
+
+### Wrapping And Proxies
+
+There are two cases that we need to manage:
+- a C# type unknown to Swift implements `ISwiftIteratorProtocol`
+- a Swift type unknown to C# implements `IteratorProtocol`
+  
+It should be noted that if the Swift type is known to C# and was bound knowing that it implements `IteratorProtocol` then the bound C# type will already implement `ISwiftIteratorProtocol`. This is effectively a trivial case. There is also the possibility that we have a type known to C# but at the time of binding we did not know that it adopts `IteratorProtocol` This can happen when protocol adoption happens through retroactive modeling (extension). This is not within the scope of this document.
```
kotlarmilos, line 157: Shouldn't the extension in this case be present in the input ABI file as `isFromExtension`? Do you have any real-world use cases where a type is not known during the binding phase?

src/docs/binding-pats.md
```diff
@@ -0,0 +1,408 @@
...
+{
+        retval.initialize(to: xamarin_static_wrapper_IteratorProtocol_next(this: &this));
+}
+```
+This is done so that the caller will allocate space for the optional as the return value and pass the object by reference.
+The pinvoke to call that wrapper looks like this:
+
+```csharp
+[DllImport (SwiftCore.kXamGlue, EntryPoint = "...")]
+internal static extern void PImethod_SwiftIteratorProtocolProtocolXamarin_SwiftIteratorProtocolProtocolDnext00000001 (IntPtr retval, IntPtr this0);
```
kotlarmilos, line 197: Where does this come from? Is it known during the binding phase?

src/docs/binding-pats.md
stephen-hawley: I don't have real world examples, but you can imagine a 3rd party library that adds, say, IteratorProtocol to integers to give you the bits that are turned on in an integer. In that case we have the type (nint) and we have the protocol, but someone else connected them together via an extension.

src/docs/binding-pats.md
stephen-hawley: This was a wrapper written by BTfS created during binding.

src/docs/vtable-alternative.md
stephen-hawley: Isn't that only the case if you have `IlcDisableReflection` set?

src/docs/binding-pats.md
kotlarmilos: Can we assume that the ABI input file for the 3rd-party library will be available, and that the conformance information will be known? I am evaluating the implementation cost compared to the usability of proxy objects. I'm glad we already have a solution in place, and I’m trying to understand when we may encounter this.

src/docs/binding-pats.md
kotlarmilos: Why is it not possible to project this type in C# since it is known during the binding phase?

src/docs/binding-pats.md
stephen-hawley: Oh absolutely. I used an ad-hoc name-mangling in BTfS to aggressively avoid naming conflicts. It gets...wordy.

src/docs/binding-pats.md
stephen-hawley: The limitations in this case was mostly due to not being able to call the implementation directly due to ABI differences. That changes with the work you've done and I'm looking forward to taking advantage of that.

src/docs/vtable-alternative.md
jkotas: `IlcDisableReflection` is an unsupported experimental option. You do not need to worry about it. It is likely going to be deleted at some point.

However, I do not see what you get from using native AOT. Native AOT compiled dylibs are fully self-contained and each native AOT compiler dylib has its own copy of the runtime, GC, etc. The only way to call code outside of them is through unmanaged interop.

src/docs/binding-pats.md
stephen-hawley: It is - it's part of the proxy system. I couldn't call the protocol implementation methods directly (before your work), so I wrote wrappers.

src/docs/vtable-alternative.md
stephen-hawley: Sure - using NativeAOT allows Swift code to call C# code directly instead of C# providing function pointers (ie, a vtable) and because of the way that Swift is, it is a requirement to have null check on each function pointer before calling it and it is a requirement to have an alternative path for the code for the case of the null check failing (which it never will, but there is no way to tell Swift that's the case). Being able to let the linker do the work cuts out the memory overhead of the vtables, which is decidedly small but will grow in generics which require a vtable for each specialization.
Another approach suggested by Rolf today was to write C code that knows about the runtime and calls into it and let Swift link to that.

jkotas: I do not think you want to have a native AOT compiled .dylib in the process in addition the regular .NET runtime. Each native AOT compiled .dylib has its own GC heap, etc. It is a lot of overhead.

If you need to build some glue code, writing it in C sounds like a better option to me.

kotlarmilos (approved):
Overall looks good! Let's structure the docs to contain design of bindings, Swift runtime, and tooling directories.

dalexsoto (approved):

PR #2709, opened 2024-10-09
Add readme and some fixes.

AaronRobinsonMSFT:
Add readme.
Fix a GC thread mode issue.
Add Numeric intrinsic support for x64, VL, and V512.


AaronRobinsonMSFT:
@janvorli @cshung The next assert you will hit with "hello, world" will be:

https://github.com/dotnet/runtimelab/blob/a357a46c9f7c57227c4976d7e8953ccbcd995177/src/coreclr/vm/threads.cpp#L5708

PR #2713, opened 2024-10-11
Fix unwinding for StubLink

cshung:
I have been looking into stack unwinding related issue, and I think I have found a culprit.
 
**Problem**: If we set a breakpoint on `Interpreter::ExecuteMethod`, we will fail to unwind the stack across the generate stub.
 
**Symptom**: The unwinder found the correct return address of the callee of the stub, and the unwinder fails to continue the unwinding because it can't find the module responsible for the address [here](https://github.com/dotnet/runtime/blob/eed9c19ee6125f58340f8f3d6530c797c7d53cc7/src/coreclr/unwinder/amd64/unwinder.cpp#L242)

**Cause**: It looks like the work of registering the unwind info is skipped because of `s_publishingActive` is false [here](https://github.com/dotnet/runtime/blob/cfc7c204c2317b5b4471c841ce4304fd74610822/src/coreclr/vm/codeman.cpp#L241) while we are trying to `StubLinker::EmitUnwindInfo` by `UnwindInfoTable::AddToUnwindInfoTable` [here](https://github.com/dotnet/runtime/blob/cfc7c204c2317b5b4471c841ce4304fd74610822/src/coreclr/vm/stublink.cpp#L1818)

### DIFF
--- a/src/coreclr/vm/stublink.cpp
+++ b/src/coreclr/vm/stublink.cpp
@@ -1815,6 +1815,7 @@ bool StubLinker::EmitUnwindInfo(Stub* pStubRX, Stub* pStubRW, int globalsize, Lo
 
 #ifdef TARGET_AMD64
     // Publish Unwind info to ETW stack crawler
+    UnwindInfoTable::PublishUnwindInfo(false);
     UnwindInfoTable::AddToUnwindInfoTable(
         &pStubHeapSegment->pUnwindInfoTable, pCurFunction,
         (TADDR) pStubHeapSegment->pbBaseAddress,
--- a/src/coreclr/vm/stublink.cpp
+++ b/src/coreclr/vm/stublink.cpp
@@ -1815,7 +1815,9 @@ bool StubLinker::EmitUnwindInfo(Stub* pStubRX, Stub* pStubRW, int globalsize, Lo
 
 #ifdef TARGET_AMD64
     // Publish Unwind info to ETW stack crawler
+#ifdef FEATURE_INTERPRETER
     UnwindInfoTable::PublishUnwindInfo(false);
+#endif // FEATURE_INTERPRETER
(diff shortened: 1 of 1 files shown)


src/coreclr/vm/stublink.cpp
```diff
@@ -1815,6 +1815,7 @@ bool StubLinker::EmitUnwindInfo(Stub* pStubRX, Stub* pStubRW, int globalsize, Lo
 
 #ifdef TARGET_AMD64
     // Publish Unwind info to ETW stack crawler
+    UnwindInfoTable::PublishUnwindInfo(false);
```
AaronRobinsonMSFT, line 1819: Should we limit this to the interpreter scenario or is this general goodness for the runtime?

jkotas: Unwinding through manually emitted stubs has always been a bug farm. It turns out we have not had any manually emitted stubs that require unwinding for a while. I am deleting the complicated buggy infrastructure in https://github.com/dotnet/runtime/pull/108684. (I was looking at some obviously buggy code related to the StubLinker unwinding when reviewing https://github.com/dotnet/runtime/pull/107282 only to find out that it is not actually used and that's why we never seen the bug... .)

This interpreter stub is the last outlier. I think that this fix is fine to help the experiment to move along. Is your goal still "proof of concept and plan for getting rid of all dynamically generated code with interpreter"? It means that you will need to figure out how to get rid of the interpreter stub and replace it with a non-dynamic solution.

AaronRobinsonMSFT, line 1819:
```suggestion
#ifdef FEATURE_INTERPRETER
    UnwindInfoTable::PublishUnwindInfo(false);
#endif // FEATURE_INTERPRETER
```

PR #2715, opened 2024-10-14
Update SystemTextJsonToolsetVersion (#108704)

cshung:
Port https://github.com/dotnet/runtime/pull/108704 to fix the build.

### DIFF
--- a/src/coreclr/inc/stack.h
+++ b/src/coreclr/inc/stack.h
@@ -17,12 +17,11 @@ class Stack
     unsigned m_elemsSize;
     unsigned m_elemsCount;
 
-    static const unsigned InitSize = 8;
-
     void GrowForPush()
     {
         if (m_elemsCount == m_elemsSize)
         {
+            static const unsigned InitSize = 8;
             m_elemsSize = max(InitSize, 2*m_elemsSize);
             T* newElems = new T[m_elemsSize];
             if (m_elemsCount != 0)
--- a/src/coreclr/inc/stack.h
+++ b/src/coreclr/inc/stack.h
@@ -21,7 +21,7 @@ class Stack
     {
         if (m_elemsCount == m_elemsSize)
         {
-            static const unsigned InitSize = 8;
+            const unsigned InitSize = 8;
             m_elemsSize = max(InitSize, 2*m_elemsSize);
             T* newElems = new T[m_elemsSize];
             if (m_elemsCount != 0)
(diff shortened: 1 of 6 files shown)


src/coreclr/inc/stack.h
```diff
@@ -17,12 +17,11 @@ class Stack
     unsigned m_elemsSize;
     unsigned m_elemsCount;
 
-    static const unsigned InitSize = 8;
-
     void GrowForPush()
     {
         if (m_elemsCount == m_elemsSize)
         {
+            static const unsigned InitSize = 8;
```
AaronRobinsonMSFT:
```suggestion
            const unsigned InitSize = 8;
```

src/coreclr/vm/interpreter.cpp
```diff
@@ -12315,13 +12317,13 @@ void Interpreter::PrintValue(InterpreterType it, BYTE* valAddr)
 
     case CORINFO_TYPE_NATIVEINT:
         {
-            INT64 val = static_cast<INT64>(*reinterpret_cast<NativeInt*>(valAddr));
+            long long val = static_cast<long long>(*reinterpret_cast<NativeInt*>(valAddr));
```
AaronRobinsonMSFT: This isn't really correct. A `long long` is a minimum of 64-bit, but it can be more. This should be `int64_t` and then update the `ll` prefix following the "Format macro constants" table found at https://en.cppreference.com/w/cpp/types/integer.

src/coreclr/vm/interpreter.cpp
```diff
@@ -12333,11 +12336,11 @@ void Interpreter::PrintValue(InterpreterType it, BYTE* valAddr)
     case CORINFO_TYPE_LONG:
         {
             INT64 val = *reinterpret_cast<INT64*>(valAddr);
-            fprintf(GetLogFile(), "%lld (= 0x%llx)", val, val);
+            fprintf(GetLogFile(), "%" PRId64 " (= 0x%" PRIx64 ")", val, val);
         }
         break;
     case CORINFO_TYPE_ULONG:
-        fprintf(GetLogFile(), "%lld", *reinterpret_cast<UINT64*>(valAddr));
+        fprintf(GetLogFile(), "%lld", *reinterpret_cast<unsigned long long*>(valAddr));
```
AaronRobinsonMSFT:
```suggestion
        fprintf(GetLogFile(), "%" PRIu64 "", *reinterpret_cast<UINT64*>(valAddr));
```

AaronRobinsonMSFT (approved):

PR #2717, opened 2024-10-15
Update async2 experiment from upstream

jakobbotsch:
It builds, but need to validate tests, which will require a working async2 Roslyn first (this change switches to use the standard Roslyn package without async2 support).


jakobbotsch:
cc @agocke @davidwrighton @VSadov

VSadov:
I have pushed some commits to switch to the rebased Roslyn compiler. 

The whole thing builds now and can run tests (as in `src\tests\run.cmd checked`). There are 6 failures, but they do not seem to be directly related to async2 - nopiatestil, vararg_TargetWindows, readytorun_coreroot_determinism, more_tailcalls,...

agocke:
Is the plan to just disable those 6 tests? (I'm fine with that)

VSadov:
> Is the plan to just disable those 6 tests? (I'm fine with that)

I think Jakob may want to take a look at the failures. The breaks could be sideeffects of async2 changes or something happened while merging runtime/JIT changes. I do not recall if the prototype was 100% passing tests. Maybe these are expected at this point.

Disabling would be fine, once we have an idea why things break.

jakobbotsch:
> The whole thing builds now and can run tests (as in `src\tests\run.cmd checked`). There are 6 failures, but they do not seem to be directly related to async2 - nopiatestil, vararg_TargetWindows, readytorun_coreroot_determinism, more_tailcalls,...

I guess we do not run async2 tests as part of the normal test run -- there were a bunch of failures when running them via src/tests/Loader/async/asynctests.csproj. However, all the async2 tests (except the shared generics one which is expected) are passing now after some fixes.

I also looked at `more_tailcalls` and `vararg_TargetWindows` and indeed there were async2 related bugs there, which I fixed too.

PR #2718, opened 2024-10-15
[SwiftBindings] added basic function and tuple demangling

stephen-hawley:
This adds a couple of things to the demangler:

1. A default for the reducer that returns a `ReductionError` by default. This is used as a placeholder for Child rules. It should never get called because Child reductions don't get called. I went back and forth as to whether I should remove the `required` keyword. Ultimately, I think it's more important to be told you didn't set the property instead of forgetting to put in a reducer.
2. Added match rules for tuples and tuple elements
3. Added match rule for function
4. Added match rule for function types
5. Support for variadics

In the big picture, all of this is necessary in order to get detangling for dispatch thunks.
Why? Because a dispatch thunk is a special case of functions.

In terms of what you're seeing here, there are a few weird things.
A function consists of:
- Provenance
- Name
- Labels
- Type of the function

The weirdness is that the argument labels are kept in a separate list from the type of the function - essentially this isn't needed because tuples can already have labels in them, but for functions they get put into a separate list.

The test code tests for a function with this signature: 
```swift
unitHelpFrawework.ReturnsInt(arg: (a: Swift.UInt, b: Swift.Int), c: Swift.Int) -> Swift.UInt
```
When demangled, we get a tree of nodes that looks like this:
```
kind=Global
  kind=Function
    kind=Module, text="unitHelpFrawework"
    kind=Identifier, text="ReturnsInt"
    kind=LabelList
      kind=Identifier, text="arg"
      kind=Identifier, text="c"
    kind=Type
      kind=FunctionType
        kind=ArgumentTuple
          kind=Type
            kind=Tuple
              kind=TupleElement
                kind=Type
                  kind=Tuple
                    kind=TupleElement
                      kind=TupleElementName, text="a"
                      kind=Type
                        kind=Structure
                          kind=Module, text="Swift"
                          kind=Identifier, text="UInt"
                    kind=TupleElement
                      kind=TupleElementName, text="b"
                      kind=Type
                        kind=Structure
                          kind=Module, text="Swift"
                          kind=Identifier, text="Int"
              kind=TupleElement
                kind=Type
                  kind=Structure
                    kind=Module, text="Swift"
                    kind=Identifier, text="Int"
        kind=ReturnType
          kind=Type
            kind=Structure
              kind=Module, text="Swift"
              kind=Identifier, text="UInt"
```
Variadics are part of the tuple declarations, so I went ahead and did those. I automatically desugar these into `Swift.Array<T>` and mark them as variadic. This will make the distinction between `func f(a: Int ...) -> Int` and `func f(a: [Int]) -> Int` which are effectively different function types.

### DIFF
--- a/src/Swift.Bindings/tests/DemanglerTests/BasicDemanglingTests.cs
+++ b/src/Swift.Bindings/tests/DemanglerTests/BasicDemanglingTests.cs
@@ -93,4 +93,29 @@ public class BasicDemanglingTests : IClassFixture<BasicDemanglingTests.TestFixtu
         Assert.Equal("someclient.Ageable", conf.ProtocolType.Name);
         Assert.Equal("someclient", conf.Module);
     }
+
+    [Fact]
+    public void TestFuncWithTuple()
+    {
+        var symbol = "_$s17unitHelpFrawework10ReturnsInt3arg1cS2u1a_Si1bt_SitF";
+        var demangler = new Swift5Demangler (symbol);
+        var result = demangler.Run ();
+        var func = result as FunctionReduction;
+        Assert.NotNull (func);
+        Assert.Equal ("ReturnsInt", func.Function.Name);
+        var returnType = func.Function.Return as NamedTypeSpec;
+        Assert.NotNull (returnType);
+        Assert.Equal ("Swift.UInt", returnType.Name);
+        var args = func.Function.ParameterList;
+        Assert.Equal (2, args.Elements.Count);
+        var firstarg = args.Elements [0] as TupleTypeSpec;
+        Assert.NotNull (firstarg);
+        Assert.Equal ("arg", firstarg.TypeLabel);
+        Assert.Equal (2, firstarg.Elements.Count);
+        Assert.Equal ("arg: (a: Swift.UInt, b: Swift.Int)", firstarg.ToString());
+        var secondarg = args.Elements [1] as NamedTypeSpec;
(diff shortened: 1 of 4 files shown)


src/Swift.Bindings/tests/DemanglerTests/BasicDemanglingTests.cs
```diff
@@ -93,4 +93,29 @@ public void TestOtherProtocolConformanceDescriptor()
...
+        Assert.Equal (2, args.Elements.Count);
+        var firstarg = args.Elements [0] as TupleTypeSpec;
+        Assert.NotNull (firstarg);
+        Assert.Equal ("arg", firstarg.TypeLabel);
+        Assert.Equal (2, firstarg.Elements.Count);
+        Assert.Equal ("arg: (a: Swift.UInt, b: Swift.Int)", firstarg.ToString());
+        var secondarg = args.Elements [1] as NamedTypeSpec;
+        Assert.NotNull (secondarg);
+        Assert.Equal ("c", secondarg.TypeLabel);
+        Assert.Equal ("Swift.Int", secondarg.Name);
```
rolfbjarne, line 119: Not a big deal, but I typically find it useful to add test cases for error scenarios as well, at the very least to verify error messages (and see how easy they are to read and understand).

rolfbjarne (approved):

kotlarmilos (approved):

PR #2720, opened 2024-10-16
[NativeAOT-LLVM] Merge to Oct 24

yowl:
Disable `IDynamicInterfaceCastable` tests as they now use a thunk pool (and `RhGetCommonStubAddress` which we haven't implemented) https://github.com/dotnet/runtime/pull/108235


yowl:
@jkotas Do we need a different agent/pool now that we target .net 10 ?

jkotas:
> @jkotas Do we need a different agent/pool now that we target .net 10 ?

I do not think so.

Are you asking because of the build breaks?
```
/__w/1/s/.dotnet/sdk/9.0.100-rc.1.24452.12/Microsoft.Common.CurrentVersion.targets(2413,5): warning MSB3106: Assembly strong name "/home/cloudtest_azpcontainer/.nuget/packages/microsoft.build.framework/17.8.3/ref/net8.0/Microsoft.Build.Framework.dll" is either a path which could not be found or it is a full assembly name which is badly formed. If it is a full assembly name it may contain characters that need to be escaped with backslash(\). Those characters are Equals(=), Comma(,), Quote("), Apostrophe('), Backslash(\). [/__w/1/s/src/tools/illink/src/ILLink.Tasks/ILLink.Tasks.csproj::TargetFramework=net9.0] [/__w/1/s/src/tests/build.proj]
/__w/1/s/.dotnet/sdk/9.0.100-rc.1.24452.12/Microsoft.Common.CurrentVersion.targets(2413,5): warning MSB3106: Assembly strong name "/home/cloudtest_azpcontainer/.nuget/packages/microsoft.build.utilities.core/17.8.3/ref/net8.0/Microsoft.Build.Utilities.Core.dll" is either a path which could not be found or it is a full assembly name which is badly formed. If it is a full assembly name it may contain characters that need to be escaped with backslash(\). Those characters are Equals(=), Comma(,), Quote("), Apostrophe('), Backslash(\). [/__w/1/s/src/tools/illink/src/ILLink.Tasks/ILLink.Tasks.csproj::TargetFramework=net9.0] [/__w/1/s/src/tests/build.proj]
```
I think these build breaks are caused by `Microsoft.Build.Framework` package not being restored for some reason.

yowl:
Yes thanks, I was.  Will restart in case it was temporary.

yowl:
Abandoning, and will try a shorter period.

PR #2721, opened 2024-10-17
Fix `Delegate` dispatch.

AaronRobinsonMSFT:
/cc @janvorli @cshung

### DIFF
--- a/src/coreclr/vm/comdelegate.cpp
+++ b/src/coreclr/vm/comdelegate.cpp
@@ -1892,12 +1892,6 @@ Stub* COMDelegate::GetInvokeMethodStub(EEImplMethodDesc* pMD)
 
         ILCodeStream *pCode = sl.NewCodeStream(ILStubLinker::kDispatch);
 
-        // This stub is only used for rare indirect cases, for example
-        // when Delegate.Invoke method is wrapped into another delegate.
-        // Direct invocation of delegate is expanded by JIT.
-        // Emit a recursive call here to let JIT handle complex cases like
-        // virtual dispatch and GC safety.
-
         // Load the delegate object
         pCode->EmitLoadThis();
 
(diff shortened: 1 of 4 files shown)


src/coreclr/vm/comdelegate.cpp
```diff
@@ -1901,16 +1905,21 @@ Stub* COMDelegate::GetInvokeMethodStub(EEImplMethodDesc* pMD)
...
             pCode->EmitLDARG(paramCount);
 
-        // recursively call the delegate itself
+#ifdef FEATURE_INTERPRETER
+        // Call the underlying method pointer.
+        pCode->EmitLoadThis();
+        pCode->EmitLDFLD(FIELD__DELEGATE__METHOD_PTR);
+
+        mdToken sigTok = pCode->GetSigToken(pSig, cbSig);
+        pCode->EmitCALLI(sigTok, sig.NumFixedArgs(), fReturnVal);
```
huoyaoyuan, line 1908: Is there any concern about GC safety of collectible methods? (dotnet/runtime#105082)

I'm not familiar about how interpreter interacts with GC.

AaronRobinsonMSFT: It really shouldn't impact the interpreter scenario, but it costs us nothing to make it correct. This is throw away code right now, but I will make it correct. Thanks for pointing this out.

huoyaoyuan: I originally used the same mechanism in dotnet/runtime#104731 . You may keep the mentioned points in mind when bringing the interpreter for production quality. /cc @jkotas

jkotas: This is throw-away code. We do not plan to bring this implementation of the interpreter to quality. The change looks good as a temporary measure. It has multiple corner-case problems and that's fine. 

The only 100% correct fix would be to treat `Invoke` as an intrinsic in the interpreter (ie keep this as is and change the code in interpreter.cpp instead).

jkotas (approved):

PR #2724, opened 2024-10-21
documented swift closures

stephen-hawley:
Added docs that spell out how Swift closures are represented in memory and the calling conventions.

### DIFF
--- a/src/docs/binding-closures.md
+++ b/src/docs/binding-closures.md
@@ -128,7 +128,7 @@ In running the other direction, we would need a way to convert a C# closure into
 
 # Idiomatic Differences
 
-The main idiomatic difference has to do with the escaping/non-escaping varieties of closure. Obviously, C# doesn't make this distinction. As such, if a C# method gets passed a delegate from Swift that is non-escaping, it incumbent upon the user to never store it. We can make this somewhat better by putting an attribute on such delegates that flags it as a non-escaping and create a Roslyn analyzer that looks for usage that would violate that.
+The main idiomatic difference has to do with the escaping/non-escaping varieties of closure. Obviously, C# doesn't make this distinction. As such, if a C# method gets passed a delegate from Swift that is non-escaping, it is incumbent upon the user to never store it. We can make this somewhat better by putting an attribute on such delegates that flags it as a non-escaping and create a Roslyn analyzer that looks for usage that would violate that.
 
 # Accessibility
 
--- a/src/docs/binding-closures.md
+++ b/src/docs/binding-closures.md
@@ -1,6 +1,6 @@
 # Binding Closures
 
-Closures in Swift come in two general forms: escaping and non-escaping. A closure is considered escaping if it consinues to exist outside of the context that created it. For example, a closure paramater to a function that gets stored in a member variable in a class *must* be marked `@escaping` or the compiler will flag it as an error.
+Closures in Swift come in two general forms: escaping and non-escaping. A closure is considered escaping if it continues to exist outside of the context that created it. For example, a closure parameter to a function that gets stored in a member variable in a class *must* be marked `@escaping` or the compiler will flag it as an error.
 
 In either case, Swift closures, unlike some other languages, are full-fledged closures that can have both free and bound variables within the body of the closure.
 
@@ -10,7 +10,7 @@ Beyond the broad classification of escaping/non-escaping, closures can also be a
 
 Internally, closures are represented as 2 machine words:
 - Pointer to a function entry point for the closure (see below for details)
-- Pointer to the data context object for captured free variables or 0/null if there is no context object
+- Pointer to the data context object for captured free variables or 0/null if there is no context object. When the closure is called, this will be in the context/self register.
 
 ## Language Parity Mismatches
 
@@ -22,7 +22,7 @@ If you have a function like this in Swift:
 public func sorter<T>(arr: [T], by: (T, T) -> Int) { /* implementation not important */ }
 
 ```
-And `sorter` can by called by any of the following means:
+And `sorter` can be called by any of the following means:
 
 ```swift
 private func sort0 (a: Int, b: Int) -> Int {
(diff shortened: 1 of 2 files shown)


src/docs/binding-closures.md
```diff
@@ -0,0 +1,135 @@
...
+
+In running the other direction, we would need a way to convert a C# closure into something that is callable from Swift. The approach in BTfS is heavy handed because of the lack of runtime support. Given a C# closure, we create a handle to it, then call into a Swift routine which generates a swift closure that calls back into C# with a pointer to argument and return as before, but now with the handle and a `@convention (c)` function pointer to goes back into to a C# routine that unpacks the arguments, uses the handle to get the original C# closure and calls it. We should be able to do better than this.
+
+# Idiomatic Differences
+
+The main idiomatic difference has to do with the escaping/non-escaping varieties of closure. Obviously, C# doesn't make this distinction. As such, if a C# method gets passed a delegate from Swift that is non-escaping, it incumbent upon the user to never store it. We can make this somewhat better by putting an attribute on such delegates that flags it as a non-escaping and create a Roslyn analyzer that looks for usage that would violate that.
```
rolfbjarne: Maybe non-escaping closures can be modeled using a ref struct? Ref structs ["are allocated on the stack and can't escape to the managed heap."](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/ref-struct)

Calling the closure might end up a little bit weird in C# though, since I think you'd have to call a method on the struct instead of using delegate syntax.

```cs
ref struct MyNonEscapingClosure {
    public void Invoke (...)
    {
		// call the closure here
    }
}
```

rolfbjarne (approved):

dalexsoto (approved):

src/docs/binding-closures.md
```diff
@@ -0,0 +1,135 @@
...
+
+In either case, Swift closures, unlike some other languages, are full-fledged closures that can have both free and bound variables within the body of the closure.
+
+The distinction between the two is that escaping closures that capture free variables use a dynamically allocated object to contain any captured variables. This object is reference counted. Non-escaping closures use stack allocated memory to contain any captured variables.
+
+Beyond the broad classification of escaping/non-escaping, closures can also be async and they can throw.
+
+Internally, closures are represented as 2 machine words:
+- Pointer to a function entry point for the closure (see below for details)
+- Pointer to the data context object for captured free variables or 0/null if there is no context object
```
kotlarmilos: Is this passed through context register?

src/docs/binding-closures.md
```diff
@@ -0,0 +1,135 @@
...
+
+None in particular. From a language standpoint, they both align in capability although the implementatioa are different.
+
+Swift has a delightful bit of syntax sugar that allows you to supply a closure at the call site in a way that it looks like inline code.
+If you have a function like this in Swift:
+```swift
+public func sorter<T>(arr: [T], by: (T, T) -> Int) { /* implementation not important */ }
+
+```
+And `sorter` can by called by any of the following means:
```
kotlarmilos:
```suggestion
And `sorter` can be called by any of the following means:
```

src/docs/binding-closures.md
```diff
@@ -0,0 +1,135 @@
...
+
+## ABI Differences
+
+Closures follow the same calling conventions as [functions](binding-functions.md).
+
+With support for Swift calling conventions in the runtime, we should be insulated from issues in the ABI differences.
+
+We would need to be aware reference counting of the context object. Closures are more or like like this:
+```swift
+public struct EscapingClosure {
```
kotlarmilos, line 49: Why not `class` since it is ref-counted?

src/docs/binding-closures.md
stephen-hawley: Yes.

src/docs/binding-closures.md
stephen-hawley: This is a Swift representation, not a C# representation. The closure itself isn't reference counted AFAIK only the `AnyObject` field

src/docs/binding-closures.md
```diff
@@ -0,0 +1,135 @@
+# Binding Closures
+
+Closures in Swift come in two general forms: escaping and non-escaping. A closure is considered escaping if it consinues to exist outside of the context that created it. For example, a closure paramater to a function that gets stored in a member variable in a class *must* be marked `@escaping` or the compiler will flag it as an error.
```
matouskozak:
```suggestion
Closures in Swift come in two general forms: escaping and non-escaping. A closure is considered escaping if it continues to exist outside of the context that created it. For example, a closure parameter to a function that gets stored in a member variable in a class *must* be marked `@escaping` or the compiler will flag it as an error.
```

src/docs/binding-closures.md
matouskozak: The data context object is implicitly added and not displayed in the Swift signature but can be seen in LLVM IR, right? I remember we were hitting issues with callbacks where an implicit pointer like argument was being added after the function pointer. Here are the runtime tests for callbacks https://github.com/dotnet/runtime/blob/main/src/tests/Interop/Swift/SwiftCallbackAbiStress/SwiftCallbackAbiStress.cs

src/docs/binding-closures.md
```diff
@@ -0,0 +1,135 @@
...
+        let (i, ss, b) = args.pointee
+        ret.initialize(to: a(i, ss, b))
+    }
+    callToCSharpMethod (a_adapter)
+}
+```
+
+In this case, the original closure gets captured by `a_adapter`. Before calling the adapter, C# obviously has to create the argument tuple, space for the return value (allocated but not initialized) and can call it because effectively the delegate signature has become `delegate void csAdapter (IntPtr ret, IntPtr args)`. A similar process is used to get C# closures into Swift.
+
+One obvious problem here is that this only works with escaping closures. If rewritten with a non escaping closure, the act of capturing the original closure in the new one is flagged by the Swift compiler as an error. Fortunately, Swift has a workaround to this via the function [`withoutActuallyEscaping`](https://developer.apple.com/documentation/swift/withoutactuallyescaping(_:do:)/), for which the previous code can be rewritten as this:
```
matouskozak:
```suggestion
One obvious problem here is that this only works with escaping closures. If rewritten with a non-escaping closure, the act of capturing the original closure in the new one is flagged by the Swift compiler as an error. Fortunately, Swift has a workaround to this via the function [`withoutActuallyEscaping`](https://developer.apple.com/documentation/swift/withoutactuallyescaping(_:do:)/), for which the previous code can be rewritten as this:
```

matouskozak (approved):

src/docs/binding-closures.md
stephen-hawley: @matouskozak The data context object always shows up in the self register. From a consistency standpoint, it makes sense as you can look at a closure with captured data as being equivalent to an instance method on an anonymous type (class if it's escaping, value if it's non-escaping). From this point of view, the context is 100% equivalent to `self`.

src/docs/binding-overview.md
```diff
@@ -98,7 +98,7 @@ To keep things organized, each of the separate language features are broken out
 - [tuples](binding-tuples.md)
 - [protocols](binding-protocols.md)
 - protocols with associated types
```
jkurdek:
```suggestion
- [protocols with associated types] (binding-pats.md)
```

src/docs/binding-closures.md
```diff
@@ -0,0 +1,135 @@
...
+
+In running the other direction, we would need a way to convert a C# closure into something that is callable from Swift. The approach in BTfS is heavy handed because of the lack of runtime support. Given a C# closure, we create a handle to it, then call into a Swift routine which generates a swift closure that calls back into C# with a pointer to argument and return as before, but now with the handle and a `@convention (c)` function pointer to goes back into to a C# routine that unpacks the arguments, uses the handle to get the original C# closure and calls it. We should be able to do better than this.
+
+# Idiomatic Differences
+
+The main idiomatic difference has to do with the escaping/non-escaping varieties of closure. Obviously, C# doesn't make this distinction. As such, if a C# method gets passed a delegate from Swift that is non-escaping, it incumbent upon the user to never store it. We can make this somewhat better by putting an attribute on such delegates that flags it as a non-escaping and create a Roslyn analyzer that looks for usage that would violate that.
```
jkurdek:
```suggestion
The main idiomatic difference has to do with the escaping/non-escaping varieties of closure. Obviously, C# doesn't make this distinction. As such, if a C# method gets passed a delegate from Swift that is non-escaping, it is incumbent upon the user to never store it. We can make this somewhat better by putting an attribute on such delegates that flags it as a non-escaping and create a Roslyn analyzer that looks for usage that would violate that.
```

jkurdek (approved):

PR #2738, opened 2024-10-28
Deleting parts of suspension-async flavor.

VSadov:
Deleting runtime-async related parts. They served their purpose well!! But no need to keep that code around.


VSadov:
I see some test failures (as in `src/tests/run bat checked`), but I see the same failures without the change. 
I assume the failures are either preexisting or machine-specific. We can look at that later.

jakobbotsch (approved):

VSadov:
Thanks!!

jakobbotsch:
Do we want to keep the notes about the suspension-async in the design doc, or should it be removed too?

VSadov:
> Do we want to keep the notes about the suspension-async in the design doc, or should it be removed too?

I think we should keep the notes with details of both implementations and what worked/not - as a part of learning from the experiment. 

Perhaps could move the notes to a different place - like a learning doc.

PR #2746, opened 2024-11-01
[SwiftBindings] Runtime metdata

stephen-hawley:
Added documentation for TypeMetadata as well as a placeholder for the nominal type descriptor.


src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,255 @@
...
+    MetadataKindIsNonHeap = 0x200,
+    MetadataKindIsRuntimePRivate = 0x100,
+}
+
+[Flags]
+public enum TypeMetadataRequest {
+    Complete = 0,
+    NonTransitiveComplete = 1,
+    LayoutComplete = 0x3f,
+    Abstract = 0xff;
```
jkurdek:
```suggestion
    Abstract = 0xff,
```

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,255 @@
...
+
+Swift type metadata is always a singleton for every unique type in Swift. The Swift runtime enforces this through accessors for getting at the type metadata and it will cache the values. In usage, the type metadata is always passed by reference but they should be considered read-only.
+
+There are two ways to get the type metadata for a Swift type:
+1 - by accessing the type metadata directly by symbolic name (only applies to non-generic nominal types)
+2 - using an accessor
+
+Within accessors, there are two classes of accessor:
+1 - compiler-generated accessors (for all nominal types)
+2 - runtime-provided accessors (for non-nominal types - tuples, closures, etc.)
```
kotlarmilos, line 29: Does this also include types with generics?

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,255 @@
...
+
+Generally speaking, it's probably best for us to use the accessors and if we choose, to cache those values as static fields in bound types.
+
+The signature of a type metadata accessor is a request followed by n TypeMetadata objects, one for each generic type:
+```csharp
+public static TypeMetadata Accessor(TypeMetadataRequest request [, TypeMetadata specialization0, TypeMetadata specialization1, ...])
```
kotlarmilos, line 35: Do arguments include protocol conformances as well?

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,255 @@
...
+    ErrorObject = 1 | MetadataFlags.MetadataKindIsNonType | MetadataFlags.MetadataKindIsRuntimePrivate,
+    Class = 0x800 // not really, but it's reasonable under the circumstances
+                // This number looks arbitrary, but the swift source says that they're never going to
+                // go over 0x7ff for predefined/enumerated metadata kind values
+}
+
+public record struct TypeMetadata
+{
+    const nint kMaxDiscriminator = 0x7ff;
+    NativeHandle handle;
```
kotlarmilos, line 87: Why do we need `NativeHandle` instead of `IntPtr`?

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,255 @@
...
+```csharp
+public bool TryGetNominalTypeDescriptor ([NotNullWhen (true)] out NominalTypeDescriptor? result) { }
+public string TypeName { get; } // uses the nominal type descriptor
+// NB - while Swift's Optional<T> is declarad as a generic enum, its implementation is a special case
+// since it is very common.
+public bool IsGeneric { get; } // uses nominal type descriptor
+public int GenericArgumentCount { get; }
+```
+
+In addition, we are likely to need **internal** accessors for:
```
kotlarmilos, line 137: Do we want to extend the `TypeMetadata` and have type-specific declarations?

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,255 @@
...
+In addition, while we are not likely to need but is present:
+- Protocol information (flags, number of witness tables, protocol descriptors for each, whether a protocol is ObjC)
+
+## Mapping C# Type to TypeMetadata and Vice Versa
+
+There should be a threadsafe 1 to 1 bi-map of Type <-> TypeMetadata to serve as cache.
+
+For the Type -> TypeMetadata half, there should be a strategy mechanism that handles the following cases:
+- primitive value types such as int, uint, short, ushort etc. NB: IntPtr is a synonym for nint and UIntPtr is a synonym for nuint
+- bound Swift types should each have a static method for getting the type metadata using the metadata accessor which should be inspectable via reflection or interface compliance
```
kotlarmilos, line 157: Is the reflection supported on static abstract interfaces?

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,255 @@
...
+- for tuples, there is a Swift library routine, `swift_getTupleTypeMetadata`
+- for closures, there is a Swift library routine, `swift_getFunctionTypeMetadata`
+- for existential containers, there is a Swift library routine, `swift_getExistentialTypeMetadata`
```
kotlarmilos, line 160:
```suggestion
- for existential containers, there is a Swift library routine, `swift_getExistentialTypeMetadata`
 - for types with generics, there is a Swift library routine, `swift_getGenericMetadata` and `swift_getOpaqueTypeMetadata`
```

kotlarmilos (approved):

src/docs/runtime-metadata.md
stephen-hawley: Yes.

src/docs/runtime-metadata.md
stephen-hawley: Protocol conformances are not part of the metadata accessor, but the resulting type metadata with likely (I don't recall off the top of my head) include information about protocol conformance.

src/docs/runtime-metadata.md
stephen-hawley: I mentioned that earlier in the docs. `IntPtr` is a synonym for `nint` and I want to avoid using it even though it technically could be used here.

src/docs/runtime-metadata.md
stephen-hawley: Yes. I did a prototype of it - the method name is unpronounceable in C#, but completely pronounceable in reflection.
```csharp
public interface IMeta {
    static abstract TypeMetadata GetTypeMetadata ();
}

// ...
    public static nint MetadataOf<T>(T t)
    {
        var tt = typeof(T);
        if (typeof(IF).IsAssignableFrom (tt)) {
            var mi = tt.GetMethod ("IFace.IMeta.GetTypeMetadata", BindingFlags.NonPublic | BindingFlags.Static);
            return (nint)mi.Invoke (null, null);
        }
        throw new NotImplementedException ();
    }
```

src/docs/runtime-metadata.md
kotlarmilos: It seems to be required for generics with protocols. Here is a prototype of `class UIHostingController<Content> where Content : View`:
```
$s7SwiftUI19UIHostingControllerCMa (0, $s7SwiftUI4TextVN, $s7SwiftUI4TextVAA4ViewAAWP)
--
$s7SwiftUI19UIHostingControllerCMa ---> type metadata accessor for SwiftUI.UIHostingController
$s7SwiftUI4TextVN ---> type metadata for SwiftUI.Text
$s7SwiftUI4TextVAA4ViewAAWP ---> protocol witness table for SwiftUI.Text : SwiftUI.View in SwiftUI

```

src/docs/runtime-metadata.md
kotlarmilos: i.e. ClassMetadata, StructMetadata, etc

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,256 @@
...
+    HeapGenericLocalVariable = 0 | MetadataFlags.MetadataKindIsNonType | MetadataFlags.MetadataKindIsRuntimePrivate,
+    ErrorObject = 1 | MetadataFlags.MetadataKindIsNonType | MetadataFlags.MetadataKindIsRuntimePrivate,
+    Class = 0x800 // not really, but it's reasonable under the circumstances
+                // This number looks arbitrary, but the swift source says that they're never going to
+                // go over 0x7ff for predefined/enumerated metadata kind values
+}
+
+public record struct TypeMetadata
+{
+    const nint kMaxDiscriminator = 0x7ff;
```
jkurdek, line 86: What is kMaxDiscriminator?

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,256 @@
...
+- for tuples, there is a Swift library routine, `swift_getTupleTypeMetadata`
+- for closures, there is a Swift library routine, `swift_getFunctionTypeMetadata`
+- for existential containers, there is a Swift library routine, `swift_getExistentialTypeMetadata`
+- for types with generics, there is a Swift library routine, `swift_getGenericMetadata` and `swift_getOpaqueTypeMetadata`
+
+For the TypeMetadata -> Type half, this is a little more problematic. For nominal types, we can using the type name for looking up the matching C# type in a runtime type database, but it is entirely possible to encounter types that have no exposure in C# through binding. Under those circumstances, we may just be stuck with an approximation. This means that if we're presented with a class type that we've not seen, we would have to report it as either a proxy type if it's presented as implementing a particular protocol or as a broad placeholder with no real bindings to the original type. Under these circumstances, since the type mapping is no longer 1 to 1, these fallback cases should never go in the main cache, but could/should live in a secondary cache.
```
jkurdek:
```suggestion
For the TypeMetadata -> Type half, this is a little more problematic. For nominal types, we can use the type name for looking up the matching C# type in a runtime type database, but it is entirely possible to encounter types that have no exposure in C# through binding. Under those circumstances, we may just be stuck with an approximation. This means that if we're presented with a class type that we've not seen, we would have to report it as either a proxy type if it's presented as implementing a particular protocol or as a broad placeholder with no real bindings to the original type. Under these circumstances, since the type mapping is no longer 1 to 1, these fallback cases should never go in the main cache, but could/should live in a secondary cache.
```

dalexsoto (approved):

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,256 @@
...
+
+// ...
+
+
+// getter
+internal static bool TryGetSwiftTypedMetadata (Type type, [NotNullWhen (true)] out TypeMetadata? result)
+{
+    if (typeof (ISwiftTyped).IsAssignableFrom (type)) {
+        // assuming explicit interface implementation
+        var mi = tt.GetMethod ("Swift.Runtime.ISwiftTyped.GetTypeMetadata", BindingFlags.NonPublic | BindingFlags.Static);
```
jkurdek, line 185:
```suggestion
        var mi = type.GetMethod ("Swift.Runtime.ISwiftTyped.GetTypeMetadata", BindingFlags.NonPublic | BindingFlags.Static);
```

src/docs/runtime-metadata.md
```diff
@@ -0,0 +1,256 @@
...
+
+In terms of typical implementations of the `ISwiftTyped` interface, we would have something like this:
+
+```csharp
+public class SomeBoundType : ISwiftTyped {
+    static TypeMetadata? _metadata;
+    static TypeMetadata ISwiftTyped.GetMetadata()
+    {
+        if (_metadata is null) {
+            _metadata = SomeBoundTypePinvokes.MetadataAccessor (MetadataRequest.Complete);
```
jkurdek:
```suggestion
            _metadata = SomeBoundTypePinvokes.MetadataAccessor (TypeMetadataRequest.Complete);
```

Applies to examples below as well. Also in pinvoke signatures

jkurdek (approved):

src/docs/runtime-metadata.md
stephen-hawley: This is a limit that Apple has the underlying implementation of type metadata in Swift is a discriminated union, but it's a complicated one. Any discriminator that is from 0-0x7ff has specific meanings, but a Swift class is not among them, so they use any value over 0x7ff for that.

jkurdek:
How would the `ISwiftTyped` interface proposal handle getting metadata for primitive types e.g. ints? Will we have to create new types on C# side e.g. `SwiftInt : ISwiftTyped` ?

PR #2748, opened 2024-11-05
[SwiftBindings] Nominal type descriptor

stephen-hawley:


src/docs/runtime-nominal-type-descriptor.md
```diff
@@ -1,3 +1,65 @@
...
+    }
+}
+
+public func getAThing() -> SomeProtocol
+{
+    return Foo<Int> (with: 17)
+}
+```
+
+In C# when we call `getAThing` we will get an existential container representing the protocol. We would normally build a proxy for the protocol which encapsulates the existential container and forwards the methods. We can and should be able to do better than this by looking at the type metadata inside the existential container. We will find the type metadata for `Foo<T>`, but we need the nominal type descriptor to be able to get the bound types. From that information we can instead runtime bind this to the C# binding for `Foo<T>` with the correct bound generic type and then in C# we could do this:
```
jkurdek, line 35: So we get an existential container. The existential container has a pointer to TypeMetadata. And TypeMetadata then has a pointer to Nominal Type Descriptor, right?

Also will we need a projection of the existential container?

src/docs/runtime-nominal-type-descriptor.md
```diff
@@ -1,3 +1,65 @@
 # Swift Nominal Type Descriptor
 
-TBD
\ No newline at end of file
+In Swift, descriptors are implicit data types that are separate from type metadata to provide more information about the type.
+Nominal type descriptors in Swift provide information about elements that are common to most nominal types: struct, class and enum. Protocols are **not** included, even though they are technically nominal types. Protocols with associated types are also **not** included as [has been discussed before](binding-pats.md) PATs are not actual types.
+
+Nominal type descriptors contain the following information that we will need at runtime:
```
jkurdek, line 6: Is there any Apple source which describes the layout of the Nominal Type Descriptor?

jkurdek (approved):

dalexsoto (approved):

src/docs/runtime-nominal-type-descriptor.md
stephen-hawley: Correct.
Yes, we will need a projection of the existential container.

src/docs/runtime-nominal-type-descriptor.md
stephen-hawley: Yes, [but it is marked as horribly out of date](https://github.com/swiftlang/swift/blob/c0a55e11d4309a34ef2612c91ca21453ccd5e721/docs/ABI/TypeMetadata.rst#nominal-type-descriptor).
The information I use is from the source.

src/docs/runtime-nominal-type-descriptor.md
```diff
@@ -1,3 +1,65 @@
...
+```
+which enables us to correctly cast the return value to its actual type, something we can't do otherwise.
+
+In terms of representation, we can use something like this:
+```csharp
+public struct NominalTypeDescriptor {
+    NativeHandle handle;
+
+    // this should only get called by TypeMetadata
+    internal NominalTypeDescriptor(NativeHandle handle)
```
rolfbjarne, line 53: Does the handle need any memory management - is it ever going to point to freed memory?

src/docs/runtime-nominal-type-descriptor.md
```diff
@@ -1,3 +1,65 @@
...
+    fooInst.printValue()
+} else {
+    // ...
+}
+```
+which enables us to correctly cast the return value to its actual type, something we can't do otherwise.
+
+In terms of representation, we can use something like this:
+```csharp
+public struct NominalTypeDescriptor {
```
kotlarmilos, line 49: Why it doesn't match fields from https://github.com/swiftlang/swift/blob/72f49bdf1b14bdb8253f8fb723c935f86bf90cdc/include/swift/ABI/Metadata.h#L3884?

src/docs/runtime-nominal-type-descriptor.md
stephen-hawley: Great question - not that I'm aware of. Nominal type descriptors come paired with type metadata and are accessible through a field (not a function) in type metadata for nominal types. Since type metadata instances are singletons and cached, I can't imagine that they would ever get freed.

src/docs/runtime-nominal-type-descriptor.md
stephen-hawley: Because we're just going to use the handle to the descriptor not mirror the descriptor itself.

kotlarmilos: Don't we need field information for `frozen` types?

rolfbjarne (approved):

src/docs/runtime-nominal-type-descriptor.md
stephen-hawley: I don't think we need that at run time and we can't count on there being nominal type descriptors at binding time. Generic types make the type metadata and nominal type descriptor at runtime typically, although in some cases the compiler will make those at compile time, but we can't count on it.
For the most part, any runtime type from Apple with the name "descriptor" in it, is likely to be used by the debugger and in theory for reflection if they ever implement it.

jkurdek, line 49: runtime-metadata.md mentions
>5. Getting access to [nominal type descriptors](https://github.com/dotnet/runtimelab/blob/feature/swift-bindings/src/docs/runtime-nominal-type-descriptor.md) to determine size, stride, etc.

How this representation will allow us to obtain this data?

stephen-hawley: That's actually incorrect. Wonder what I was thinking. That info is in the value witness table

PR #2751, opened 2024-11-06
[SwiftBindings] docs for existential containers

stephen-hawley:


src/docs/runtime-existential-containers.md
```diff
@@ -0,0 +1,60 @@
...
+
+In Swift when a protocol *without associated types* or a composition of protocols is passed as an argument into or returned from a function, the value gets passed in an Existential Container. This is a value type which serves as a box for any type that adopts the protocol or protocols needed. It is a [variable length struct](https://github.com/swiftlang/swift/blob/main/docs/ABI/TypeLayout.rst#existential-container-layout) of the form:
+
+| Index | Contents       | Size in Machine Words |
+|-------|----------------|-----------------------|
+|   0   | Payload        | 3                     |
+|   3   | Metadata       | 1                     |
+|   4   | Witness Tables | variable: 0 or more   |
+
+The start of the box includes 3 machine words which serve as a payload for the existential container. If the payload is 3 machine words or less, it will be copied into the existential container. If the payload is larger than 3 machine words, a copy will be allocated from the heap and the pointer to that copy will live in the payload.
```
kotlarmilos, line 11: > If the payload is 3 machine words or less, it will be copied into the existential container.

Is this true for `frozen` types only?

src/docs/runtime-existential-containers.md
```diff
@@ -0,0 +1,60 @@
...
+```csharp
+public interface IExistentialContainer {
+    IntPtr Data0 { get; set; }
+    IntPtr Data1 { get; set; }
+    IntPtr Data2 { get; set; }
+    SwiftTypeMetadata ObjectMetadata { get; set; }
+    NativeHandle this [int index] { get; set; } // this[] and Count could just as easily be IEnumerable<NativeHandle>
+    int Count { get; }
+    int Sizeof { get; }
+    unsafe IntPtr CopyTo (IntPtr memory);
```
rolfbjarne: This signature doesn't need `unsafe`

```suggestion
    IntPtr CopyTo (IntPtr memory);
```

src/docs/runtime-existential-containers.md
```diff
@@ -0,0 +1,60 @@
...
+This is noted in [Apple's documentation](https://github.com/swiftlang/swift/blob/main/docs/ABI/TypeLayout.rst#class-existential-containers). I have never encountered this in Apple's code nor in the wild. This is mostly due to Apple encouraging Swift to be a value type based language.
+
+Although this would not be a priority for us, the type could be represented like this:
+```csharp
+public interface IClassExistentialContainer {
+    NativeHandle Object { get; set; }
+    NativeHandle this [int index] { get; set; } // this and count could just as easily by IEnumerable<NativeHandle>
+    int Count { get; }
+    int Sizeof { get; }
+    unsafe IntPtr CopyTo (IntPtr memory);
```
rolfbjarne:
```suggestion
   IntPtr CopyTo (IntPtr memory);
```

src/docs/runtime-existential-containers.md
stephen-hawley: no because all information you need on how to manipulate the payload will be in the full type metadata (ie, the value witness table). And unless I'm thinking about it incorrectly, ETs are runtime types not compile time types so they are insulated from changes in the types that get boxed.

src/docs/runtime-existential-containers.md
```diff
@@ -0,0 +1,60 @@
...
+Finally, there is an inline list of pointers to protocol witness tables. This list can be 0 elements in length as in the case of the special `Any` protocol or it can be n elements in length as in the case of a protocol composition type (`P1 & P2 & ... Pn`). In the case of an existential container with more than 1 protocol witness table, they are ordered alphabetically by the protocol name (note, this is from the source code - Apple has not documented this).
+
+Because of the memory issues, existential containers need to be handled with attention paid to reference counting, but this can be managed by avoiding exposing the existential containers to the public API. This is straight-forward since in order to present protocols back and forth to each language will require dual proxies. This is detailed [here](binding-protocols.md).
+
+To start, we want to be able to operate on existential containers in the abstract. Therefore all existential containers should implement a common interface:
+```csharp
+public interface IExistentialContainer {
+    IntPtr Data0 { get; set; }
+    IntPtr Data1 { get; set; }
+    IntPtr Data2 { get; set; }
```
rolfbjarne: I'm assuming this is the payload from above, any reason this is named `Data` and not `Payload` (or something with "Payload")? Having the same thing referred to by different names makes it harder to reason about.

src/docs/runtime-existential-containers.md
```diff
@@ -0,0 +1,60 @@
...
+    SwiftTypeMetadata ObjectMetadata { get; set; }
+    NativeHandle this [int index] { get; set; } // this[] and Count could just as easily be IEnumerable<NativeHandle>
+    int Count { get; }
```
rolfbjarne, line 33: Is `Count` the number of witness tables?

src/docs/runtime-existential-containers.md
```diff
@@ -0,0 +1,60 @@
...
+    int Sizeof { get; }
```
rolfbjarne: Some xml documentation explaining each member might be nice, it took me a while to realize this is probably the size of the native existential container.

rolfbjarne (approved):

src/docs/runtime-existential-containers.md
stephen-hawley: Heh. Probably because I named it Data early and decided late that it was a payload.

src/docs/runtime-existential-containers.md
stephen-hawley: Yes - effectively, it's the number of elements in the indexer.

rolfbjarne (approved):

src/docs/runtime-existential-containers.md
```diff
@@ -0,0 +1,69 @@
...
+    // This is the type metadata of the object in the payload. The type metadata will determine
+    // whether or not the payload contains the actual object or a heap-allocated copy. This is done by
+    // looking at the value witness table, if it's a value type, which has a field for the type's size
+    TypeMetadata ObjectMetadata { get; set; }
+    // an indexer into each of the value witness tables
+    NativeHandle this [int index] { get; set; } // this[] and Count could just as easily be IEnumerable<NativeHandle>
+    // the number of value witness tables
+    int Count { get; }
+    // the size of this in MACHINE WORDS not bytes
+    int SizeOf { get; }
```
jkurdek, line 35: this = existential container? Also why we need the machine words size?

jkurdek (approved):

src/docs/runtime-existential-containers.md
stephen-hawley: misread my own code is why

PR #2760, opened 2024-11-08
[Runtime Async] Addressing test failures.

VSadov:
With these changes I see only one failure when running runtime tests (related to shared generics) and that is being fixed in 
https://github.com/dotnet/runtimelab/pull/2755

In this change:
* there is one small type system fix that makes R2R determinism tests to pass.
* Bunch of benchmark-like tests were either made build-only or have reduced number of iterations (so that we still had coverage, without timing out)


jakobbotsch (approved):
LGTM, to the best of my knowledge.

VSadov:
Thanks!!

PR #2761, opened 2024-11-08
[SwiftBindings] added full demangling results with easy accessors and cleaned up the code to require fewer allocations

stephen-hawley:
There are 3 pieces to this PR to keep in mind.

The first is that I refactored the demangler itself so that `Run ()` now takes the symbol to demangle. This is good in that we don't need to throw away the demangler after every call to Run, which is once per symbol in the file. The is less good in that we need to keep in mind that the demangler is not thread safe so instances of the demangler should not be shared across threads.

The second is that I added a general class for holding demangling results along with factory methods for generating them. Since this process is potentially slow (let's say a few milliseconds per symbol and several hundred symbols to several thousand symbols per file), I added async flavors so this could be run in parallel to another early task (for example, parsing the abi.json file).

The third is that the first test that I wrote was to parse the entire test file turned up a few Nodes that we needed but had no reductions for.  The test file is by no means exhaustive in terms of nodes we will encounter in the wild, but it's more than the piecemeal tests that we had before.
As we move forward, we are going to encounter nodes that we need to handle. That's fine - fully expected. The thing to do under these circumstances is:
Write code to check for High severity errors in the DemanglingResults, take the symbols out of the errors and open issues with the symbols in the issue. That makes it easy to write unit tests around them and fix the issues faster.

One of the new Node types I encountered is a pointer to a generic parameter. In Swift these are presented as a depth and an index. The depth is how far away the parameter is from its call site. For example, a generic parameter declared in a function will always have depth 0 since it is at its declaration site. If you have something like this:
```swift
public struct Foo<T> {
    public func doSomething<U, V> (a: T, b: U, c: V) { }
}
```
Then in `doSomething`, U and V have a depth of 0 and T has a depth of 1.
Index is the position of the generic in the generic declaration. U has an index of 0 and V has an index of 1.

For the time being, I'm making the generic name be in the form `T_depth_index`, but I could see the value of making generic parameters be unpronounceable to make it easier to distinguish them in code. For example, they could be named in the form `#depth#index` or use some other character which is generally hostile to the language as along as its easy to identify and easy to parse. The goal would be to make it easy to write the following predicate:
```csharp
bool IsUnboundGeneric (NamedTypeSpec ns) { }
```

### DIFF
--- a/src/Swift.Bindings/src/Demangler/DemanglingResults.cs
+++ b/src/Swift.Bindings/src/Demangler/DemanglingResults.cs
@@ -1,0 +1,107 @@
+using BindingsGeneration.Demangling;
+using Xamarin;
+
+/// <summary>
+/// A class to contain results from demangling the set of symbols in a MachO file.
+/// </summary>
+public class DemanglingResults
+{
+    /// <summary>
+    /// Constructs a DemanglingResults object with the desired reductions separated out
+    /// </summary>
+    /// <param name="reductions">An array of reductions</param>
+    DemanglingResults (IReduction [] reductions)
+    {
+        Errors = ArrayOf<ReductionError> (reductions);
+        MetadataAccessors = ArrayOf<MetadataAccessorReduction> (reductions);
+        DispatchThunks = ArrayOf<DispatchThunkFunctionReduction> (reductions);
+        ProtocolWitnessTables = ArrayOf<ProtocolWitnessTableReduction> (reductions);
+        ProtocolConformanceDescriptors = ArrayOf<ProtocolConformanceDescriptorReduction> (reductions);
+    }
+
+    /// <summary>
+    /// A utility routine to filter out a specific type of reduction from a set of general reductions
+    /// </summary>
+    /// <typeparam name="T">The type of the desired aggregation of reductions</typeparam>
+    /// <param name="reductions">an array of general reductions to filter</param>
+    /// <returns>An array of the requested filtered reductions</returns>
+    static T[] ArrayOf<T> (IReduction [] reductions) where T : IReduction
+    {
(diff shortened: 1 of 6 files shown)


kotlarmilos:
> we need to keep in mind that the demangler is not thread safe so instances of the demangler should not be shared across threads.

Please document this.

src/Swift.Bindings/src/Demangler/DemanglingResults.cs
```diff
@@ -0,0 +1,107 @@
...
+/// </summary>
+public class DemanglingResults
+{
+    /// <summary>
+    /// Constructs a DemanglingResults object with the desired reductions separated out
+    /// </summary>
+    /// <param name="reductions">An array of reductions</param>
+    DemanglingResults (IReduction [] reductions)
+    {
+        Errors = ArrayOf<ReductionError> (reductions);
```
kotlarmilos, line 15: Are the results accessed directly by the parser/emitter, or are they retrieved via the type registrar class?

kotlarmilos (approved):

src/Swift.Bindings/src/Demangler/DemanglingResults.cs
stephen-hawley: It depends on where it's needed. Probably at the beginning of binding, you would run the demangler in parallel with the code that reads in the .json.abi file.

stephen-hawley:
> > we need to keep in mind that the demangler is not thread safe so instances of the demangler should not be shared across threads.
> 
> Please document this.

I just looked back at the code and I explicitly made `Run` thread-safe. TFW you think of an issue, address it, and promptly forget about it.

PR #2762, opened 2024-11-10
[NativeAOT-LLVM] Merge oct 08

yowl:
This PR merges to Oct 8.

Bump wasmtime to avoid illegal instruction (probably one of the fixes listed for 21.0.2 https://github.com/bytecodealliance/wasmtime/releases)

Dont use `ResolveRuntimeFilesFromLocalBuild ` as for Wasm `RuntimeFiles` is empty and the target errors.

Mark `CORINFO_HELP_GETREFANY` as not pure for Wasm

cc @dotnet/nativeaot-llvm


src/coreclr/jit/gentree.h
```diff
@@ -4848,6 +4848,11 @@ class CallArg
         return m_lateNode == nullptr ? m_earlyNode : m_lateNode;
     }
 
+    GenTree*& GetNodeRef()
```
SingleAccretion: If you rename this to `NodeRef`, it will avoid a conflict with https://github.com/dotnet/runtime/pull/109224/files.

yowl: Thanks, done.

SingleAccretion (approved):
LGTM. Thank you!

Issue #2764, opened 2024-11-11
[RuntimeAsync] Validate and fix support for arm64, x86 and arm32


jakobbotsch:
We have not been running these targets, so we should ensure that things are building and that the tests run on them as well.

jakobbotsch:
Fixed by #2767 #2769 #2770 #2773

PR #2767, opened 2024-11-12
[RuntimeAsync] Add x86 support

jakobbotsch:
This PR makes all our async2 tests pass on win-x86. The main thing needed is that the JIT32 GC encoder has a hard limit on 4 epilogs, so the JIT needs to share suspension epilogs, and also needs to ensure we have at most 3 other normal epilogs when we get to the async2 transformation.

Contributes to #2764

### DIFF
--- a/src/coreclr/jit/async.h
+++ b/src/coreclr/jit/async.h
@@ -32,6 +32,7 @@ class Async2Transformation
     unsigned                      m_exceptionVar            = BAD_VAR_NUM;
     BasicBlock*                   m_lastSuspensionBB        = nullptr;
     BasicBlock*                   m_lastResumptionBB        = nullptr;
+    BasicBlock* m_sharedReturnBB                            = nullptr;
 
     void LiftLIREdges(BasicBlock*                    block,
                       GenTree*                       beyond,
--- a/src/coreclr/jit/fgdiagnostic.cpp
+++ b/src/coreclr/jit/fgdiagnostic.cpp
@@ -2726,6 +2726,9 @@ bool BBPredsChecker::CheckEhTryDsc(BasicBlock* block, BasicBlock* blockPred, EHb
         return true;
     }
 
+    // Async resumptions are allowed to jump into try blocks at any point. They
+    // are introduced late enough that the invariant of single entry is no
+    // longer necessary.
     if (blockPred->HasFlag(BBF_ASYNC_RESUMPTION))
     {
         return true;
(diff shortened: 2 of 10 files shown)


src/coreclr/jit/async.h
```diff
@@ -32,6 +32,7 @@ class Async2Transformation
     unsigned                      m_exceptionVar            = BAD_VAR_NUM;
     BasicBlock*                   m_lastSuspensionBB        = nullptr;
     BasicBlock*                   m_lastResumptionBB        = nullptr;
+    BasicBlock* m_sharedReturnBB                            = nullptr;
```
jakobbotsch, line 35: Formatting is a bit off here, but the JIT formatter makes a bunch of changes in various code, so I want to submit those formatting changes in a separate PR.

jakobbotsch:
PTAL @VSadov

src/coreclr/jit/fgdiagnostic.cpp
```diff
@@ -2726,6 +2726,9 @@ bool BBPredsChecker::CheckEhTryDsc(BasicBlock* block, BasicBlock* blockPred, EHb
         return true;
     }
 
+    // Async resumptions are allowed to jump into try blocks at any point. They
+    // are introduced late enough that the invariant of single entry is no
+    // longer necessary.
```
VSadov, line 2731: Roslyn rewrites awaits in `try` blocks as a switch at the single entry that dispatches to appropriate resume location. Nested cases get dispatched multiple times. It is nice that runtime async can resume directly into `try` blocks, so we can leave awaits in trys as-is.

Awaits in handlers might be harder to allow, I guess.

VSadov (approved):
LGTM. Thanks!!

src/coreclr/jit/fgdiagnostic.cpp
jakobbotsch: > Roslyn rewrites awaits in `try` blocks as a switch at the single entry that dispatches to appropriate resume location. Nested cases get dispatched multiple times. It is nice that runtime async can resume directly into `try` blocks, so we can leave awaits in trys as-is.

Yeah, this definitely simplifies things. For OSR we have to do the similar transformation as Roslyn, creating the "step" blocks at the beginning of try's.

> Awaits in handlers might be harder to allow, I guess.

Indeed, seems like it would be very hard, so I'm glad we disallow that.

PR #2768, opened 2024-11-13
[RuntimeAsync] Clean up and fix async2 tests

jakobbotsch:
- Remove the BuildOnly benchmark tests from the test wrapper (this would result in failures when running the test wrapper in some cases)
- Remove all `RequiresProcessIsolation`, since none of these tests should require process isolation (they can be built standalone if desired)
- Mark all the tests as optimized, since we get more coverage that way (tier0/tier1 testing will cover with and without JIT optimizations)
- Rename all tests to remove the async2 prefix and use a consistent style


jakobbotsch:
PTAL @VSadov

VSadov:
I also think the "Loader" is an unintuitive place for the tests. `async` is not strictly a Loader feature. 
Perhaps the tests could live right under `src/Tests`, but the move can happen at any time when convenient.

VSadov (approved):
Thanks!!

jakobbotsch:
> I also think the "Loader" is an unintuitive place for the tests. `async` is not strictly a Loader feature.
> Perhaps the tests could live right under `src/Tests`, but the move can happen at any time when convenient.

Good point -- I pushed a commit that moves it under src/tests.

PR #2769, opened 2024-11-13
[RuntimeAsync] Add support for arm64

jakobbotsch:
Mainly just fixing some missing pieces on the JIT side, and adding `x2` as the register used to return the async continuation.

Contributes to #2764


jakobbotsch:
This passes all async2 tests (once #2768 is incorporated) on my Ubuntu arm64 box. 

PTAL @VSadov

src/coreclr/vm/arm64/stubs.cpp
```diff
@@ -789,6 +789,8 @@ void HijackFrame::UpdateRegDisplay(const PREGDISPLAY pRD, bool updateFloats)
      pRD->pCurrentContext->Sp = PTR_TO_TADDR(m_Args) + s ;
 
      pRD->pCurrentContext->X0 = m_Args->X0;
+     pRD->pCurrentContext->X1 = m_Args->X1;
```
VSadov, line 792: Not sure if this is needed. 
What happens if `pRD->pCurrentContext->X0 = m_Args->X0;` is removed? Probably nothing. 
We should not be interested in calee-trash registers here. 

The leaf frame is handled in `HijackFrame::GcScanRoots`. Only that cares about returns.

VSadov: I think we can have this either way - just add x1 and x2 for consistency or remove x0. 

The whole `HijackFrame` deal is not strictly necessary now (except maybe on x86), since call sites are naturally interruptible. 

I'll look into removing `HijackFrame`. (in the main repo) - so that we do not need to make these kind of changes on every architecture and on NativeAOT as well.

VSadov (approved):
LGTM. Thanks!!

src/coreclr/vm/arm64/stubs.cpp
jakobbotsch: Yeah, I added them here for consistency, but indeed it seems like they are unnecessary.

PR #2770, opened 2024-11-13
[RuntimeAsync] Fix linux-x64 support

jakobbotsch:
- Fix Linux version of `OnHijackTripThread`
- Fix some code lost in a bad merge

Contributes to #2764


jakobbotsch:
PTAL @VSadov

src/coreclr/jit/codegenxarch.cpp
```diff
@@ -2208,6 +2208,7 @@ void CodeGen::genCodeForTreeNode(GenTree* treeNode)
             break;
 
         case GT_ASYNC_CONTINUATION:
+            genCodeForAsyncContinuation(treeNode);
```
VSadov, line 2211: Did win-x64 work without this?

VSadov: > Fix some code lost in a bad merge

I guess that refers to this change

VSadov (approved):
Thanks!!

src/coreclr/jit/codegenxarch.cpp
jakobbotsch: It worked in our tests, but only because LSRA usually ends up allocating registers such that the codegen for this node is a no-op. it seems likely it would break in rare cases or under stress.

jakobbotsch: Linux-x64 likely also worked without this change, but I noticed it while working on this and figured I would fold in the fix here.

PR #2771, opened 2024-11-13
[SwiftBindings] Document generic methods 

jkurdek:
This PR proposes C# bindings for Swift generic methods.

### DIFF
--- a/src/docs/binding-generics.md
+++ b/src/docs/binding-generics.md
@@ -5,11 +5,11 @@ This document explores projections of Swift generics into C#.
 1. [Generic functions](#generic-functions)
 2. Generic types (TBD)
 3. Generics with protocol constraints (TBD)
-4. Generic with PAT constraints (TBD)
+4. Generics with PAT constraints (TBD)
 
 ## Generic functions
 
-When a generic type is used in Swift, its memory layout isn't known at compile time. To manage this, the method parameter whose type is genericis represented by an opaque pointer. To handle this:
+When a generic type is used in Swift, its memory layout isn't known at compile time. To manage this, the method parameter whose type is generic represented by an opaque pointer. To handle this:
 
 - **Opaque Pointers**: Swift represents generic parameters as opaque pointers (`%swift.opaque`), abstracting the actual memory layout.
 - **Type Metadata**: Swift includes a type metadata pointer (`%T`) as an implicit argument at the end of a function's signature. This metadata provides the runtime information needed to manage the generic type.
@@ -118,7 +118,7 @@ static unsafe void* GetNativeHandle<T>(ref T type) where T : ISwiftObject
 unsafe interface ISwiftObject
 {
     // Other members
-    
+
     public unsafe void* GetNativeHandle();
 }
 
--- a/src/docs/binding-generics.md
+++ b/src/docs/binding-generics.md
@@ -1,0 +1,189 @@
+# Binding generics
+
+This document explores projections of Swift generics into C#.
+
+1. [Generic functions](#generic-functions)
+2. Generic types (TBD)
+3. Generics with protocol constraints (TBD)
+4. Generic with PAT constraints (TBD)
+
+## Generic functions
+
+When a generic type is used in Swift, its memory layout isn't known at compile time. To manage this, the method parameter whose type is genericis represented by an opaque pointer. To handle this:
+
+- **Opaque Pointers**: Swift represents generic parameters as opaque pointers (`%swift.opaque`), abstracting the actual memory layout.
+- **Type Metadata**: Swift includes a type metadata pointer (`%T`) as an implicit argument at the end of a function's signature. This metadata provides the runtime information needed to manage the generic type.
+- **Indirect Return**: For returning generic types, Swift uses an [indirect return result](https://github.com/dotnet/runtime/issues/100543), where the caller provides a memory location for the return value.
+
+Consider the following simple swift function:
+
+```swift
+func returnData<T>(data: T) -> T {
+    return data
+}
+```
+
+When compiled into LLVM-IR the above function has the following LLVM-IR signature:
+
+```llvm
+@"output.returnData<A>(data: A) -> A"(ptr noalias nocapture sret(%swift.opaque) %0, ptr noalias nocapture %1, ptr %T)
+```
+
+- `%0`: Indicates an **indirect return result**, where the caller allocates memory for the return value.
+- `%1`: The `data` parameter, passed as a pointer.
+- `%T`: The **type metadata** for `T`.
+
+### Projecting into C\#
+
+A sample projection into C# assuming that the type argument is projected as struct could look like this:
+
+```csharp
+[DllImport(Path, EntryPoint = "...")]
+[UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+public static extern void PInvoke_ReturnData(SwiftIndirectResult result, void* data, void* metadata);
+
+public static unsafe T ReturnData<T>(T data) where T : unmanaged, ISwiftObject
+{
+    var metadata = Runtime.GetMetadata(data);
+    nuint payloadSize = /* Extract size from metadata */;
+    
+    
+    var payload = NativeMemory.Alloc(payloadSize);
+    try 
+    {
+        var result = new SwiftIndirectResult(payload);
+        PInvoke_ReturnData(result, &data, metadata);
+        return *(T*)payload;
+    }
+    finally
+    {
+        NativeMemory.Free(payload);
+    }
+}
+```
+
+However, the projection has to support type arguments agnostic of whether they are projected as structs (frozen structs) or classes (non-frozen structs). We cannot then use the `unmanaged` constraint on `T` and will need to introduce abstractions for getting a native handle of an arbitrary Swift type and constructing a Swift type from such a handle.
+
+Those could look something like
+
+```csharp
+public static unsafe void* GetNativeHandle<T>(T type) where T : ISwiftObject
+public static unsafe T ConstructFromNativeHandle<T>(void* payload) where T : ISwiftObject
+```
+
+Using these abstractions, our projection becomes:
+
+```csharp
+[DllImport(Path, EntryPoint = "...")]
+[UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+public static extern void PInvoke_ReturnData(SwiftIndirectResult result, void* data, void* metadata);
+
+public static unsafe T ReturnData<T>(T data) where T : ISwiftObject
+{
+    var metadata = Runtime.GetMetadata(data);
+    nuint payloadSize = /* Extract size from metadata */;
+    var payload = NativeMemory.Alloc(payloadSize);
+    try 
+    {
+        SwiftIndirectResult result = new SwiftIndirectResult(payload);
+        
+        void* dataPointer = Runtime.GetNativeHandle<T>(ref data)
+        PInvoke_ReturnData(result, dataPointer, metadata);
+        return Runtime.ConstructFromNativeHandle<T>(payload);
+    }
+    finally
+    {
+        NativeMemory.Free(payload);
+    }
+}
+```
+
+**Memory Management Considerations:**
+
+- **Structs**: The CLR copies the data from unmanaged to managed memory. We must free the unmanaged memory to avoid leaks.
+- **Classes**: In this example we expect that calling `ConstructFromNativeHandle` will copy the necessary data so that the function can free all the resources it allocated.
+
+#### New API implementation
+
+##### `GetNativeHandle` Implementation
+
+We can implement `GetNativeHandle` by extending the `ISwiftObject` interface:
+
+```csharp
+static unsafe void* GetNativeHandle<T>(ref T type) where T : ISwiftObject
+{
+    return type.GetNativeHandle();
+}
+
+unsafe interface ISwiftObject
+{
+    // Other members
+    
+    public unsafe void* GetNativeHandle();
+}
+
+unsafe class NonFrozenStruct : ISwiftObject
+{
+    private void* _payload;
+
+    // Other members
+
+    public unsafe void* GetNativeHandle() => _payload;
+}
+
+unsafe struct FrozenStruct : ISwiftObject
+{
+    // Other members
+
+    public unsafe void* GetNativeHandle() => Unsafe.AsPointer(ref this);
+}
+```
+
+##### `FromNativeHandle` Implementation
+
+We can make `ISwiftObject` generic and include a factory method:
+
+```csharp
+static unsafe T FromPayload<T>(void* payload) where T : ISwiftObject<T>
+{
+    return T.FromNativeHandle(payload);
+}
+
+public unsafe interface ISwiftObject<T> where T : ISwiftObject<T>
+{
+    // Other members
+
+    public static abstract unsafe T FromNativeHandle(void* payload);
+}
+
+unsafe class NonFrozenStruct : ISwiftObject<NonFrozenStruct>
+{
+    private static nuint PayloadSize = /* Extract size from metadata */;
+    private void* _payload;
+    
+    private NonFrozenStruct(void* payload) { 
+        _payload = NativeMemory.Alloc(PayloadSize); 
+        NativeMemory.Copy(payload, _payload, PayloadSize);
+    }
+    
+    // Other members
+    
+    public static unsafe NonFrozenStruct FromNativeHandle(void* payload)
+    {
+        return new NonFrozenStruct(payload);
+    }
+}
+
+unsafe struct FrozenStruct: ISwiftObject<FrozenStruct>
+{
+    // Other members
+
+    public static unsafe FrozenStruct FromNativeHandle(void* payload) => *(FrozenStruct*)payload;
+}
+```
+
+An alternative to extending the `ISwiftObject` interface with new methods could be using reflection.
+
+**TODO:**
+
+- Figure out how to handle builtin types e.g. Int
--- a/src/docs/binding-generics.md
+++ b/src/docs/binding-generics.md
@@ -6,6 +6,7 @@ This document explores projections of Swift generics into C#.
 2. Generic types (TBD)
 3. Generics with protocol constraints (TBD)
 4. Generics with PAT constraints (TBD)
+5. Existential container for a protocol as a generic type (TBD)
 
 ## Generic functions
 
@@ -40,15 +41,15 @@ A sample projection into C# assuming that the type argument is projected as stru
 ```csharp
 [DllImport(Path, EntryPoint = "...")]
 [UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
-public static extern void PInvoke_ReturnData(SwiftIndirectResult result, void* data, void* metadata);
+public static extern void PInvoke_ReturnData(SwiftIndirectResult result, NativeHandle data, TypeMetadata metadata);
 
 public static unsafe T ReturnData<T>(T data) where T : unmanaged, ISwiftObject
 {
-    var metadata = Runtime.GetMetadata(data);
+    TypeMetadata.TryGetTypeMetadata<T>(out var metadata);
     nuint payloadSize = /* Extract size from metadata */;
-    
-    
+
     var payload = NativeMemory.Alloc(payloadSize);
+
     try 
     {
         var result = new SwiftIndirectResult(payload);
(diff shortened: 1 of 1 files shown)


stephen-hawley (commented):
Here's what I found worked best in BTfS (and there's room for improvement).
There's a method called `MetatypeOf` which looks at a C# type argument and returns a swift type metadata object for it. There is a private version of it that does most of the work [here](https://github.com/xamarin/binding-tools-for-swift/blob/404106de00dbcc47ed3f17692e5b5ec6667f3918/SwiftRuntimeLibrary/SwiftMarshal/StructMarshal.cs#L130).

Today I would write that as a `TryGet` pattern.
Essentially, it looks at the type that is being marshaled and tries to identify what bin the type falls into and either retrieves a type metadata object from that.  For native types (int, uint, etc), I keep a cache of the type metadata for those which get pulled from libswiftCore by symbolic name.

Function and tuples and created on the fly using a Swift API to do it.

In general, things fall into 3 categories:
types we own and can dictate the shape of
types we don't own but exist in swift
types that we can generate the type metadata for (tuples, functions, bound generic types)

The 3rd category is handled by calling the swift type's metadata accessor function which will look like this:
```
public static extern TypeMetadata GetMetadata (MetadataRequest type, TypeMetadata boundType0, ...)
```

src/docs/binding-generics.md
```diff
@@ -0,0 +1,189 @@
...
+}
+```
+
+However, the projection has to support type arguments agnostic of whether they are projected as structs (frozen structs) or classes (non-frozen structs). We cannot then use the `unmanaged` constraint on `T` and will need to introduce abstractions for getting a native handle of an arbitrary Swift type and constructing a Swift type from such a handle.
+
+Those could look something like
+
+```csharp
+public static unsafe void* GetNativeHandle<T>(T type) where T : ISwiftObject
+public static unsafe T ConstructFromNativeHandle<T>(void* payload) where T : ISwiftObject
```
stephen-hawley: For this, there is a similar piece of code to `MetatypeOf` which goes in reverse [here](https://github.com/xamarin/binding-tools-for-swift/blob/404106de00dbcc47ed3f17692e5b5ec6667f3918/SwiftRuntimeLibrary/SwiftMarshal/StructMarshal.cs#L718). It follows a similar pattern where it looks at the type and based on broad varieties of types and makes the matching C# type. In many cases, this can be done by blitting the memory from the pointer to the C# type. For non-blitable types (non-frozen value types or value types that contain non-bitable types), we should use the value witness table's function InitializeWithCopy. In order to do this properly, every non-blittable value type should have a constructor which is not accessible from Swift so as not to cause unwanted side-effects to the value type. In BTfS I created a unique enum in C# that wouldn't be seen in Swift, making it easy to get the ctor via reflection. We could also impose a static interface method to make a default/uninitialized version of the type.

src/docs/binding-generics.md
```diff
@@ -0,0 +1,189 @@
...
+- `%T`: The **type metadata** for `T`.
+
+### Projecting into C\#
+
+A sample projection into C# assuming that the type argument is projected as struct could look like this:
+
+```csharp
+[DllImport(Path, EntryPoint = "...")]
+[UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+public static extern void PInvoke_ReturnData(SwiftIndirectResult result, void* data, void* metadata);
```
stephen-hawley: I understand that the pinvoke signature is largely inconsequential, but it will help us to use strong types where we can. In this case, use `TypeMetadata` for the metadata argument.

src/docs/binding-generics.md
stephen-hawley: For heap allocated types (classes and actors), the pointer is not a buffer but is is and actual instance pointer. For getting that into C#, we need to have a global registry that has a map of NativeHandle -> GCHandle where the GCHandle is taken from the actual C# instance of the type. For a type that doesn't exist in the map, we use the C# type (if we have it) or use the TypeMetadata to identify the C# type and run a flavor of constructor that takes a NativeHandle to initialize the type. As part of the memory management, we take both a strong and weak reference to the Swift handle. If the C# type is disposed, we inform the registry and it removes it from the cache as well as dropping the references. In BTfS this is in the class [SwiftObjectRegistry](https://github.com/xamarin/binding-tools-for-swift/blob/main/SwiftRuntimeLibrary/SwiftObjectRegistry.cs). The strong reference is taken in the object and the weak is taken in the registry.

stephen-hawley (changes requested):

src/docs/binding-generics.md
stephen-hawley: With regards to getting type metadata, I think that the shape of the accessor should look like this:
```csharp
public interface INominalType {
    static TypeMetadata TypeMetadata { get; } // or TypeMetadata GetTypeMetadata ();
}

public class SomeBoundGenericType<T> : INominalType {
     static TypeMetadata _tmd PinvokesForSomeBoundGenericType.MetadataAccessor(MetadataRequest.Complete, Runtime.GetMetadataOrThrow(typeof (T)));
    TypeMetadata INominalType.TypeMetadata => _tmd;
}
internal static class PInvokesForSomeBoundGenericType {
    [DllImport(/* ... */)]
    public static extern TypeMetadata MetadataAccessor(MetadataRequest rq, TypeMetadata t0);
```
Runtime.GetMetadata can have a number of forms:
```
public static TypeMetadata GetMetadata(Type t) { } // what happens on fail?
// and/or
public static bool TryGetTypeMetadata(Type t, [NotNullWhen(true)]out TypeMetadata? md) { }
// and/or
public static TypeMetadata GetMetadataOrThrow(Type t)
{
    if (TryGetTypeMetadata(t, out var md)) {
        return md;
    } else {
        throw new RuntimeException($"Unable to get metadata for {t.Name}");
   }
}
```
I'd like to see the implementation of `TryGetTypeMetadata` to include a case along the lines of:
```csharp
    if (typeof(INominalType).IsAssignableFrom (t)) {
        var pi = t.GetProperty("INominalType.TypeMetadata", BindingFlags.Static);
        md = (TypeMetadata)pi.GetValue(null)!;
        return true;
    }
```

vitek-karas: Ideally we would avoid reflection as much as possible. I know that for generics this is a bit tricky as they're "unknown" by design, but maybe we can rely on C# generics in that case. 

Theoretically speaking we should not need reflection anywhere - when we generate projections we know of all possible types which will occur since we have to generate projections for them. So we may be able to rely on a generated lookup table instead of on reflection.

Reflection will get tricky due to AOT/Trim compatibility. The example above is not trim compatible and AFAIK there's no way to make it really compatible since we can't put annotation on the `t` as it's completely unknown. Also I would be careful about calling static interface methods through reflection - we need to try it, there were runtime bugs around this even on CoreCLR.

vitek-karas: Actually the trimability of this approach is something I'm not sure about. The trimming rules for static interface methods are rather complicated - and I don't know how it interacts with reflection. We will need to ask...
In any case I would expect the above to produce a Trim warning.

jkurdek: Just a wild idea. Maybe we could replace reflection with some usage of source generators. For example for primitive types like `Int` we could define a `struct Swift.IntWrapper : ISwiftObject<Swift.IntWrapper>`. The interface would provide us methods to get metadata etc. Then in order to allow the users to call the generics using regular C# `Int` we would have a source generator which would take their function call and generate a function which wraps the primitive types and calls the correct internal generic function. This way the user would get compile-time errors when trying to call a generic with unsupported argument.

src/docs/binding-generics.md
jkurdek: Thanks for the comments!

I don't really understand the `InitializeWithCopy` use case. In this scenario a new non-blittable struct is created on swift side and it being written to a buffer we provide from the C# side. So I assumed that just assigning this buffer to the C# projection should work? I cant fit where we would use `InitializeWithCopy` in this scenario. Could you please share some more context?

I would like to scope out heap allocated types from this from now and focus on structs as those represent a big chunk of our current needs. Then we would get back when we go and project classes / actors. However, I do think that once we start this description will be very relevant so I will go and include this comment it in the doc. What do you think?

src/docs/binding-generics.md
```diff
@@ -0,0 +1,189 @@
+# Binding generics
+
+This document explores projections of Swift generics into C#.
+
+1. [Generic functions](#generic-functions)
+2. Generic types (TBD)
+3. Generics with protocol constraints (TBD)
+4. Generics with PAT constraints (TBD)
```
kotlarmilos, line 8: 5. Existential container for a protocol as a generic type (TBD)

kotlarmilos (approved):
Good job Jeremi! LGTM!

src/docs/binding-generics.md
vitek-karas: I thought about this some more... and I have to admit small amount of reflection is probably better here. We could also use reflection to create a `Helper<T>` over the `t` with a constraint and inside just call the static interface method - not sure which one is better, both are probably currently problematic for AOT/trim.

That said, this is basically an equivalent code to calling an instance interface method on an `object` - and that is something trimming does support (and should) even if there are no annotations anywhere. The static method is problematic because there's no C# syntax for the instance equivalence:
```
void MYTest(object o)
{
    if (o is ISwiftObject so)
    {
        so.GetMetadata();
    }
}
```

C# can only do this through generics, but it can't do it on `Type`. So it might mean we need a new trimming feature or some kind of interesting workaround, but it should be possible without annotating `t`. Trimmer (and AOT) already works by the rule that "If the interface method is used and the type which implements it is used, then the type's implementation of that method must be kept" - static interface methods should in general work in a similar way (and if not we would have to fix the trimming).

Source generators in this case are a bit problematic - they only work well within one assembly. So for example, we can't rely on the source generator to run on a type in some random assembly - we would have to tied the usage of that somewhere. Another problem is that you might end up with 10 different generated implementations of `IntWrapper` - for value types this is probably OK (just wasteful), but if we need object identity it might become a problem. There are also versioning considerations.

And replacing reflection with a big lookup table has its own issues (how do we initialize the table if we need to merge two generated tables from two different assemblies -> type collisions -> problems - see JsonContext problems in JSON and ASP.NET DI). That said - we will have the problem of type identity across modules anyway, so that's another thing we need to potentially resolve.

kotlarmilos: Why do we need reflection here? Aren't the metadata accessors known at bind time?

stephen-hawley: I understand the issue of trimming. In order to do a fully correct implementation based on static analysis alone requires solving a problem akin to the halting problem. I would prefer to have a 100% correct solution in hand at the outset rather than requiring essentially a post processing step that will only be correct in the cases that are determinable at compile time.

Here's an example. Suppose we have the following in Swift:
```swift
public protocol Named {
    func name() -> String
}
public class Foo<T> : Named {
    private var val: T
    private var name: String
    public init(name: String, val: T) {
        self.val = val
    }
    public func getName() -> String {
        return "\(name)\(val)"
    }
}

public func getNamed () -> Named {
     return Foo<Int> ("Alex", 17)
}
```
When bound into C# we will get an Existential container with a concrete instance of Foo in it. We could represent this will a proxy, but in this particular case we can inspect the type metadata in it and find that it is actually a Foo and we can inspect the generic parts and find that it's an nint. We should therefore present this in C# as a `Foo<T>` rather than as a proxy. In this case we would need to make a C# `Foo<nint>` which would have eluded static analysis.
Similarly, I can write pathological cases in C# that will also elude static analysis.

stephen-hawley: So before I sound too negative about this, I think that static analysis will work to a degree and it can be put in place. So assuming we look at the broad shape of the T we need a Swift type metadata for and have a case that looks in a table that contains a mapping of C# bound generic types that are mapped to a cache of Swift type metadata along with a closure that will generate the type metadata the first time it's accessed.
The important thing to note is that on cache miss, you can fall back to reflection - so we can start with reflection based code and add static analysis as a later optimization with no loss of generality.

vitek-karas: > Similarly, I can write pathological cases in C# that will also elude static analysis.

If you manage to do that, I would be very interested and it's likely a bug :-)

But you have a correct point in the above. I honestly don't know how we're going to solve that. I suspect we might need to be able to fallback to a wrapper. I assume we will need a wrapper for pretty much every protocol anyway for the cases where the underlying type is not projected to C#, so if the projection is for a type which can't be safely constructed (like Foo<int>), then we would fallback to the general wrapper. This is an interesting problem. I think we may need to discuss this in a wider audience. For non-AOT cases, you're absolutely right that using reflection in this case helps a lot. But for AOT this is problematic.

And given that on iOS AOT is basically a must have.... I'd rather we have an AOT compatible solution which potentially doesn't behave ideally (like creating a wrapper in the above case).

I'd also add that we want this behavior to be predictable and stable - it should not be magic. As a developer I need to experience the exact same behavior between debug and runtime and it remains consistent whether I add some more code to my app or not - and ideally, I should be able to predict the behavior by just looking at the projected code. There will be rough edges for sure, but we should try to minimize those. Maybe we could return a generic wrapper which implements `IDynamicCastableInterface` that could fix this for interfaces. We don't have a way to cast to a different class/struct type though, so it's not a solution for `Foo<int>`.

src/docs/binding-generics.md
stephen-hawley: The InitializeWithCopy should be used if we need to return a type to a swift caller, for example a closure, virtual method - anything done from a reverse invoke.

Absolutely include this in the doc. I understand the focus on value types since that's most of what Apple prefers in Swift, but since the generic programming can take any type, we need to ensure that we build with an eye towards that.

src/docs/binding-generics.md
MichalStrehovsky: > Here's an example. Suppose we have the following in Swift:

How does the problematic code on the C# side look like? I'm trying to understand why we wouldn't know about `Foo<nint>` at compile time. I would think that if someone calls `getNamed` and then proceeds to handle the returned `Named` as not just `Named` but as a `Foo<nint>`, we would see `Foo<nint>` _somewhere_ at AOT compile time, it wouldn't just suddenly dynamically appear at runtime.

vitek-karas: I think the question if what should happen if C# calls `getNamed()` and doesn't cast it. It would mean we don't see `Foo<nint>` anywhere, so it doesn't exist and thus we would have to fallback to a protocol wrapper. The thing I'm nervous about is that this effectively changes runtime behavior based on type usage in the app and it's observable as I can then call `GetType().ToString()` on the return value and see if it was indeed `Foo<nint>` or the wrapper. It would also change behavior between Debug/Release.

Maybe we'll have to live with this, I'm not sure...

MichalStrehovsky: Forgetting the generics for a moment - what's the trimming story for this? If `getNamed` returned a `Named` that was also `SomeFoo`. Would we potentially trim `SomeFoo` if nobody was using it? Or are all Swift types rooted in this design?

vitek-karas: Great question - and it's basically what I was getting at originally... 
I think the answer right now is "We don't know" and somebody will have to figure it out.

MichalStrehovsky: If we know about `Foo<nint>` at compile time (in the same way that we know about the non-generic `SomeFoo`) this at least leaves a foot in the door to allow making this AOT-safe. I don't know the shape of a solution and I don't know if we'd need new AOT compiler features or it can be handled by a generator, but it would be a start.

src/docs/binding-generics.md
jkurdek: Updated API proposal. It would not work for types not know at compile time, but it should be NativeAOT compatible - thanks @MichalStrehovsky for guidance.

```csharp
public static class Helper<T> where T : ISwiftObject
{
    public static TypeMetadata GetTypeMetadata()
    {
        return T.Metadata;
    }

    public static unsafe T FromPayload(void* pointer)
    {
        return (T)T.FromPayload(pointer);
    }
}

public record struct TypeMetadata
{
...
  public static bool TryGetTypeMetadata<T>([NotNullWhen(true)] out TypeMetadata? result)
  {
      {
          var helperType = typeof(Helper<>).MakeGenericType(typeof(T));
          result = (TypeMetadata)helperType.GetMethod("GetTypeMetadata")!.Invoke(null, null);
          return true;
      }
      else
      {
          result = null;
          return false;
      }
  }
  ...
}

public unsafe interface ISwiftObject
{
    public static abstract TypeMetadata Metadata { get; }

    public void* GetPayload();

    public static abstract object FromPayload(void* pointer);

}

public static class Runtime
{
    public static unsafe void* GetPayload<T>(ref T type)
    {
        {
            return ((ISwiftObject)type).GetPayload();
        }
        else
        {
            ...
        }
    }

    public static unsafe T FromPayload<T>(void* payload)
    {
        {
            var helperType = typeof(Helper<>).MakeGenericType(typeof(T));
            return (T)helperType.GetMethod("FromPayload")!.Invoke(null, new object[] { (IntPtr)payload })!;
        }
        else
        {
            ...
        }
    }
}
```

stephen-hawley: Remember that we are preferring to use `NativeHandle` instead of `void *` and `IntPtr`. The advantage being that `NativeHandle` won't require `unsafe` and won't conflict with methods that that `nint` as an argument (which is super common in Swift since the `Int` type is a platform size-dependent integer and is thus `nint`)

jkurdek: Sure, will update before adding to the docs. We will have to revisit  `SwiftIndirectResult` one day as it currently accepts only `void*`

src/docs/binding-generics.md
stephen-hawley: In looking at this issue, we have a couple of different problems here:
1. how to optimally trim the output assemblies
2. how to keep to the principal of least astonishment (code should behave as you would expect to the greatest degree)

So in this particular case, we have a function, `getNamed` that returns a protocol, or more precisely an existential container that contains a `Foo<Int>` in the payload, the type metadata for `Foo<Int>` and the protocol witness table for `Named` with respect to `Foo<Int>`.

The simplest solution is to create a C# proxy that implements `INamed` and encapsulates the existential container. This allows `Foo<Int>` to be treated as `INamed` but as nothing else beyond the proxy type. This allows trimming but violates the principal of least astonishment because in Swift I could do this:
```swift
let x = getNamed ()
if let y = x as? Foo<Int> {
    print("It's a Foo<Int>")
}
```
Whereas in C# if I did:
```csharp
var x = TopLevelFunctions.GetNamed ();
if (x is Foo<Int> y) {
    Console.WriteLine ("this will never get reached");
}
```
And in theory we can handle this because we have everything we need in the existential container to project the existential container in C# as `Foo<Int>`: a handle to the Swift instance, and its type metadata, but in order to make this case work we need the ability to look up the type metadata in a runtime database to get `Foo<_>` and crack open the type metadata to extract the generic specialization and look it up in the runtime type database to get `nint` which is enough for us to call a factory on `Foo<nint>` to make one from the handle. This, of course, requires reflection. The one piece that this doesn't address is when the protocol is an extension protocol and the bound type `Foo<_>` won't show that it implements the protocol.

As to whether or not this case will be common, my instincts say yes since Apple is very much in the camp of aspect oriented design rather than OO.

I should also point out that this will also be an issue for any Swift function that returns `some` of a PAT and using `some` is the entire basis of SwiftUI.

The final question for us to answer is which of the two goals are more important: size of the output binary or referential transparency.

We might be able to get the best of both worlds in this case by using the a predicate to determine whether or not a type should be trimmed based on if the type is a Swift binding and if it implements one or more protocols. While that sounds promising, because of the heavy use of aspect oriented programming, everything implements protocols.

stephen-hawley (approved):

src/docs/binding-generics.md
vitek-karas: This is probably for another discussion, but things which I found interesting in your reply:

Extension protocols - if we project them as the implementation class, that class will not have the interface, so we can't really do that. So we have to project them as some kind of wrapper - we could still make the wrapper compare equally to the underlying instance of the projected class, but that doesn't solve the problem with type checks. The corresponding `is Foo` will simply not work in C#. I think this is effectively a language limitation, since .NET/C# doesn't have the notion of extension interfaces.

> which of the two goals are more important: size of the output binary or referential transparency

I don't think we have much of an option on the binary size - we need to make it small. Let's assume that some day in the future we will be able to project all of the foundation libraries. At that time, if I write an app which makes a single use of type from Swift I would get all of those types from foundation libraries... and my app would be big. That doesn't sound like a good solution.

jkurdek: I will be merging this PR, I opened an issue which we can use for discussion: https://github.com/dotnet/runtimelab/issues/2794

PR #2773, opened 2024-11-14
[RuntimeAsync] Add arm32 support

jakobbotsch:
- Use `r2` for the async continuation return register
- Skip a prolog/epilog optimization that could cause us to overwrite this register in the epilog

Contributes to #2764


jakobbotsch:
PTAL @VSadov

VSadov (approved):
LGTM. Thanks!

PR #2774, opened 2024-11-14
Interpreter design docs

BrzVlad:


docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
...
... variable, including argument registers reside on this space. When an interpreter method starts executing, it expects the arguments to be present one after the other at the location pointed to by the stack pointer. This means that for a call exiting the interpreter, with a certain signature, we need to call a thunk that receives at least the location of the parameters and the target address and that moves each argument from the interpreter stack to the corresponding reg/native stack location followed by a native call to the target address. Once the call returns it should move the return value from the regs/stack to the interpeter stack. If we need to pass to native/compiled code a code pointer that can be used to entry the interpreter, we need to create a thunk that moves all arguments from the native regs/stack to a separate memory location. This thunk should also have embedded a pointer that identifies the interpreter method that we need to execute. The thunk should over pass the memory location where the arguments have been copied together with the interpreter method to execute, so the interpreter entry code can set up a new interpreter frame and begin executing the method code.
```
AaronRobinsonMSFT:
```suggestion
The interpreter operates on a separate stack space that it maintains. Every local variable, including argument registers reside on this space. When an interpreter method starts executing, it expects the arguments to be present one after the other at the location pointed to by the stack pointer. This means that for a call exiting the interpreter, with a certain signature, we need to call a thunk that receives at least the location of the parameters and the target address and that moves each argument from the interpreter stack to the corresponding reg/native stack location followed by a native call to the target address. Once the call returns it should move the return value from the regs/stack to the interpreter stack. If we need to pass to native/compiled code a code pointer that can be used to entry the interpreter, we need to create a thunk that moves all arguments from the native regs/stack to a separate memory location. This thunk should also have embedded a pointer that identifies the interpreter method that we need to execute. The thunk should over pass the memory location where the arguments have been copied together with the interpreter method to execute, so the interpreter entry code can set up a new interpreter frame and begin executing the method code.
```

AaronRobinsonMSFT: > we need to call a thunk that receives at least the location of the parameters and the target address and that moves each argument from the interpreter stack to the corresponding reg/native stack location

> Once the call returns it should move the return value from the regs/stack to the interpeter stack.

This movement will also need to involve create a GCInfo block so the GC can update both the interpreter locations _and_ be aware of the the "transitioned arguments".

docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
...
+
+  1. Specialized path
+  
+     A specialized path means that we will have a thunk specialized per signature that can be used to call any method with that signature. The basic set of signatures that are needed by an application are easy to compute. We will require a separate thunk for every pinvoke signature as well as for every signature of a method that is aot-compiled, because each aot-compiled method can end up being called from the interpreter. While these thunks could be emitted directly in assembly code, it makes more sense to compile them as IL wrappers and emit them together with the rest of the managed code when aot compiling an assembly. Each one of these wrappers will receive as arguments the native pointer of the method to call and the address on the interpreter stack where the arguments reside. The wrapper will be able to compute the base address of every single argument and load the argument value. It will then proceed with executing the native call, followed by writting back the result of the call to the interpreter stack in place of the arguments.
```
AaronRobinsonMSFT:
```suggestion
     A specialized path means that we will have a thunk specialized per signature that can be used to call any method with that signature. The basic set of signatures that are needed by an application are easy to compute. We will require a separate thunk for every pinvoke signature as well as for every signature of a method that is aot-compiled, because each aot-compiled method can end up being called from the interpreter. While these thunks could be emitted directly in assembly code, it makes more sense to compile them as IL wrappers and emit them together with the rest of the managed code when aot compiling an assembly. Each one of these wrappers will receive as arguments the native pointer of the method to call and the address on the interpreter stack where the arguments reside. The wrapper will be able to compute the base address of every single argument and load the argument value. It will then proceed with executing the native call, followed by writing back the result of the call to the interpreter stack in place of the arguments.
```

docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
...
+
+  1. Generic Path
+  
+     The purpose of the generic path is to be able to implement interp exit calls with a signature that it is not known at compile time. This is probably not a common scenario, but I assume there are cases where one could generate IL code at runtime that does a PInvoke call through a pointer with a dynamically generated signature that is not known at compile time. In this situation we need a unique thunk, or at least a limited number of thunks that are able to handle any type of signature.
```
AaronRobinsonMSFT:
```suggestion
     The purpose of the generic path is to be able to implement interp exit calls with a signature that is not known at compile time. This is probably not a common scenario, but I assume there are cases where one could generate IL code at runtime that does a PInvoke call through a pointer with a dynamically generated signature that is not known at compile time. In this situation we need a unique thunk, or at least a limited number of thunks that are able to handle any type of signature.
```

docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
...
+
+##### Interpreter Entry
+
+In scenarios where we need to pass a function pointer to a pinvoke from the interpreter (an `UnmanagedCallersOnly` method) or in cases where compiled code needs to invoke a method that wasn't AOT compiled, the runtime will have to dynamically generate a thunk that can enter execution into the interpreter. Given this function pointer needs to be callable from compiled code with the same signature as the original method, the thunk will have to embed at least the interpreter method pointer that identifies the method to be interpreted and an additional function pointer that will be called to do the actual argument translation and call into the interpreter. The runtime could maintain pairs of pages for the thunks, one for code and one for data, that can be remapped as many times as necessary whenever generation of a new thunk is required. Similarly with the interpreter exit case, we could have two approaches for the argument translation, a specialized and a generic one.
```
AaronRobinsonMSFT, line 42: > The runtime could maintain pairs of pages for the thunks, one for code and one for data, that can be remapped as many times as necessary whenever generation of a new thunk is required.

From what I've been told this will work on iOS, but not in a WASM environment.

docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
...
+     
+  1. Generic path
+
+     The generic path is very important for the interpreter entry scenario because this entry transition is required for methods that are interpreted, which by definition are methods that are not really known/handled at AOT compile time. This means that it would be common for us to not know the signature for the method entry, meaning we can't aot compile the necessary wrapper in advance. We could follow a similar approach to the interpreter exit generic path.
```
AaronRobinsonMSFT, line 50: > The generic path is very important for the interpreter entry scenario because this entry transition is required for methods that are interpreted

Is this stating that an interpreted method calling another interpreted method would use this entry transition?

docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
...
+
+Native code has no knowledge of fat pointers, so we will need to explicitly generate small thunks for every single `UnmanagedCallersOnly` method. On mono, during app compilation, a build task scans the assembly for all `UnmanagedCallersOnly` methods and it dynamically generates a separate C method for each one of them. This method will have its own data that is later initialized with the function pointer for the cconv translation wrapper together with the interpreter method pointer argument. Rather than have this logic in special build tasks that dynamically generate C code, it might make more sense to  simply generate a special direct call wrapper for `UnmanagedCallersOnly` methods, that are not being aot compiled to the wasm image (this might represent a scenario that is just good to have, for interp-only, but not really mandatory since we can choose to always aot compile these methods).
```
AaronRobinsonMSFT: > this might represent a scenario that is just good to have, for interp-only, but not really mandatory since we can choose to always aot compile these methods

Right. I think we should not consider this a need. It is good to have an idea on how to address it, but I think we should avoid any "nice to have" scenarios for the interpreter.

docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
...
+
+Given on wasm we might encounter situations where we don't have a necessary wrapper for a certain signature we would be in need for a fallback approach. While on browser we could rely on dynamic generation of wasm code, on wasi we have no alternative. For the rare cases where users would run into such scenarios, I think a simple approach would be for the runtime to report the missing signature when crashing and instruct the user to specify these signatures into a separate file, that can then be consumed by the wasm application build, so the additional wrappers are compiled.
```
AaronRobinsonMSFT: > While on browser we could rely on dynamic generation of wasm code, on wasi we have no alternative.

I would deprioritize any WASI-specific features. Again, great to have a way forward, but should not be prioritized at present.

docs/design/interpreter/precise-gc.md
```diff
@@ -0,0 +1,23 @@
+### Synchronisation with GC suspension
+
+Interpreter will run in GC unsafe mode. This means that suspension can happen only at certain points in the interpreter code. certain.
```
AaronRobinsonMSFT:
```suggestion
Interpreter will run in GC unsafe mode. This means that suspension can happen only at certain points in the interpreter code.
```

docs/design/interpreter/precise-gc.md
```diff
@@ -0,0 +1,23 @@
...
+* Locations where we end up calling into the runtime. For example, as part of executing an opcode, we might need to call into the runtime to ensure that the class is initialized and the cctor is being run. The same rules as for explicit safepoints should apply, by saving any information needed by the stack walker.
+
+Interpreter instruction typically read values from the interpreter stack, then proceed with the operation and in the end it might end some value back to the interpreter stack. For more complex opcodes that call into the runtime, a GC could happen in the middle of the opcode execution. All these locations where GC interruption can happen should be well defined, because we will have to enforce the rule that every object reference is refetched from the interpreter stack following a potential suspention point. Otherwise we could end up with native registers/stack containing an old ref for a moved objects. References that are being used by the opcode could also be stored in a separate memory location associated to the `GCFrame` of the native interpreter execution loop method.
```
AaronRobinsonMSFT:
```suggestion
Interpreter instruction typically read values from the interpreter stack, then proceed with the operation and in the end it might end some value back to the interpreter stack. For more complex opcodes that call into the runtime, a GC could happen in the middle of the opcode execution. All these locations where GC interruption can happen should be well defined, because we will have to enforce the rule that every object reference is refetched from the interpreter stack following a potential suspension point. Otherwise we could end up with native registers/stack containing an old ref for a moved objects. References that are being used by the opcode could also be stored in a separate memory location associated to the `GCFrame` of the native interpreter execution loop method.
```

docs/design/interpreter/compiled-code-interop.md
BrzVlad: Currently it is not obvious to me whether we would actually need this. These transitions are very fast and we can decide to never suspend while executing them, in which case there should be no need to update anything. I wrote a bit about this on the other GC document.

docs/design/interpreter/compiled-code-interop.md
BrzVlad: Yes, this section is for normal architectures. I mentioned a solution for this problem in the wasm section.

docs/design/interpreter/compiled-code-interop.md
BrzVlad: No. The idea is that a method that is not aot-compiled, therefore interpreted at runtime, could in theory happen to be called from compiled code. Therefore we would need to have a transition wrapper for its signature. But if an interpreter method will call it, it will just call it directly, without any wrappers.

docs/design/interpreter/compiled-code-interop.md
BrzVlad: This is best to have for the purpose of incrementally developing the interpreter by using an interp-only configuration, in case certain parts of the compiled code interop lag behind, since they might be more challenging. CoreCLR with aot wasm doesn't even exist at the moment and I have no idea how easy it is to reuse stuff from nativeaot. For wasm seems very likely to me that we would have an interp-only first.

docs/design/interpreter/compiled-code-interop.md
BrzVlad: I know that jitting on wasm is dubious, but I feel the purpose of the interpreter on wasm would be more geared towards wasi, where there are dynamic code gen restrictions. On browser, in theory, RyuJIT could have a wasm backend and there would be no need for interpreter. So if we are to prototype something on wasm at some point, I think it would best to go with the solution that works everywhere, aka wasi. Anyhow I don't see wasm to be relevant for the prototype, just to have an ideea on how we can move forward there once we figure stuff out on arm64.

docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
+### Compiled code interop
+
+The interpreter is expected to run in combination with aot compiled code so we will need an efficient mechanism for entering/exiting interpreter. Even in an interp only mode, we would still need to do these transitions for pinvokes and reverse pinvokes. These transitions also pose a few challenges since on ios we won't be able to dynamically generate thunks, while on WASM the call signature must be embedded into the emitted wasm code, meaning we can't reuse a generic thunk for calling methods with different signatures.
```
jkotas:
```suggestion
The interpreter is expected to run in combination with aot compiled code so we will need an efficient mechanism for entering/exiting interpreter. Even in an interp only mode, we would still need to do these transitions for pinvokes and reverse pinvokes. These transitions also pose a few challenges since on iOS we won't be able to dynamically generate thunks, while on Wasm the call signature must be embedded into the emitted Wasm code, meaning we can't reuse a generic thunk for calling methods with different signatures.
```

docs/design/interpreter/compiled-code-interop.md
yowl: As subsequently the suggestion is to use IL wrappers, I presume we are talking about IL signatures, rather than native signatures (which would be more shareable, but have other disadvantages I suppose) ?

docs/design/interpreter/compiled-code-interop.md
```diff
@@ -0,0 +1,69 @@
...
...erate thunks that can be invoked, in order to generate an interpreter entry point we could reuse the functionality of fat pointers that is already used with native aot. Fat pointers can point to additional data, if a bit is set, rather than being actual function pointers. Instead of calling the pointer directly, the calling code will check for the most significant bit. If it is set, it will instead dereference the pointer and obtain the real function pointer together with the additional argument that is passed to the call. For the purpose of entering the interpreter, we would generate a fat pointer that has the target destination as a compiled IL wrapper for the signature in question together with the interpreter method pointer that is passed. The wrapper will obtain the pointer to the interpreter stack, will move all arguments there and call into the C++ interpreter path, passing the method and the address on the stack where the arguments have been written. When aot compiling an assembly, we would need to consider every single call as a potential entry to the interpreter and, if we deem it possible, we would generate an interpreter entry wrapper for the call signature in question.
```
yowl, line 65: I believe fat pointers on Wasm use the MSB, but on non-Wsm, the LSB is used (beacuse Wsam function pointers are indicies)

docs/design/interpreter/precise-gc.md
```diff
@@ -0,0 +1,23 @@
...
+
+When the GC scans the roots from all threads, it will rely on the stack walker to unwind each frame on the thread and for each frame it will callback into interpreter via a `GcScanRoots` method. The interpreter will have to precisely identify which slots from the interpreter stack contain refs.
+
+For the purpose of precisely reporting roots from interpreter frames, as a simplification, we might want to separate the ref space from the no-ref space. The mono interpreter has a var offset allocator, which allocates offsets on the interpreter stack for each variable in the code. As vars die, stack space is reused for new vars. The offset allocator will need to always maintain two separate spaces, one for refs and one for no-refs and allocate offsets for vars accordingly. We will need to make sure we zero the space reserved for ref vars, at each method entry, making this approach potentially expensive during code execution.
```
janvorli, line 17: Based on the next paragraph, it seems that this simplification would actually complicate stuff - arguments would need special treatment. I feel like it would be actually cleaner to use something like or directly the GC info that coreclr JIT / crossgen generates. That means describing ranges of IR opcode addresses in which specific variable / argument slot contains a GC reference (it also has a concept of slots that are considered to contain reference for the whole body of the method). It seems this approach would work universally for all cases and it should not be difficult to implement. And the fact that it would be unified with the GC info for precompiled code also sounds like a benefit.

BrzVlad: My assumption was that it might be somewhat expensive to just do this tracking for every single var, so trying to group them together might make it more efficient but I'm starting to see how this is not really as important as I had thought. Anyhow, I didn't explicitly state in the document, but I think the best approach for the prototype is to simply use conservative stack scan and have some general idea on how to later move it to precise, when we have engineering time to spare. I see precise stack scanning to be a feature more relevant to a fully matured interpreter, aka a few years down the road.

janvorli: Precise GC is a requirement for coreclr, collectible code (unloadability) would not work properly with conservative GC and there are cases when we would crash if a bogus conservative root would artificially make a dead unreachable reference live again.

PR #2775, opened 2024-11-14
[NativeAOT-LLVM] only require EMSDK when building for browser

dicej:
The WASI build no longer requires EMSDK.

### DIFF
--- a/src/coreclr/build-runtime.cmd
+++ b/src/coreclr/build-runtime.cmd
@@ -349,9 +349,11 @@ for /f "delims=" %%a in ("-%__RequestedBuildComponents%-") do (
         set __CMakeTarget=!__CMakeTarget! nativeaot
 
         if "%__TargetArch%"=="wasm" (
-            if not defined EMSDK (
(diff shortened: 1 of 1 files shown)


src/coreclr/build-runtime.cmd
```diff
@@ -349,9 +349,11 @@ for /f "delims=" %%a in ("-%__RequestedBuildComponents%-") do (
...
 
         if "%__TargetArch%"=="wasm" (
-            if not defined EMSDK (
-                echo %__ErrMsgPrefix%%__MsgPrefix%Error: The EMSDK environment variable pointing to emsdk root must be set.
-                goto ExitWithError
+            if "%__TargetOS%"=="browser" (
+                if not defined EMSDK (
+                    echo %__ErrMsgPrefix%%__MsgPrefix%Error: The EMSDK environment variable pointing to emsdk root must be set.
+                    goto ExitWithError
+                )
```
SingleAccretion: I think this whole check can actually be deleted, since we call `gen-buildsys.cmd` below that includes the same check.

SingleAccretion (approved):

PR #2778, opened 2024-11-14
[NativeAOT-LLVM] update `SharedLibrary` smoke test for WASIp2

dicej:
This updates the `SharedLibrary` test for WASIp2 and adds some cases to cover `wasi:http`.


src/tests/nativeaot/SmokeTests/SharedLibrary/Library.cs
```diff
@@ -0,0 +1,176 @@
+// Generated by `wit-bindgen` 0.32.0. DO NOT EDIT!
+// <auto-generated />
```
SingleAccretion: Could you trim this to the most minimum set [of exports] we need to satisfy in order for the test to build?

I think for this test it would be useful to have the lowest-level view of the system, i. e. do without the auto-generated code.

src/tests/nativeaot/SmokeTests/SharedLibrary/wit/world.wit
```diff
@@ -0,0 +1,11 @@
+package local:local;
```
SingleAccretion, line 1: Looks unused.

dicej: It's used by `generate-shared-library-bindings.sh` (see the `wit-bindgen c-sharp -w library -r native-aot wit` command, which refers to the `wit` directory).

SingleAccretion: Regardless of my other suggestion about generated code, this file is an exact duplicate of `LibraryWorld_component_type.wit`, so, one of these should be deleted.

src/tests/nativeaot/SmokeTests/SharedLibrary/Library.cs
dicej: I could write the same code by hand, but it would look pretty much the same, minus the `None` and `Result` definitions (which we should teach `wit-bindgen` not to emit when they're not needed).  I guess I'm not seeing an advantage to doing that given that the generated code is pretty readable already and doesn't hide any of the details.  And if/when we add additional entrypoints in the future, it will be convenient to regenerate the code from the WIT.

Happy to do whatever is needed to get this PR merged, but I don't quite understand the motivation for this suggestion.  As a compromise, perhaps I could update `wit-bindgen` to omit unneeded declarations and regenerate these bindings with that change?

src/tests/nativeaot/SmokeTests/SharedLibrary/wit/world.wit
dicej: It's true that it's a duplicate.  The reason is that `world.wit` is an _input_ file to `wit-bindgen`, whereas `LibraryWorld_component_type.wit` is an _output_ file from `wit-bindgen`, intended to be passed to `wasm-component-ld` during linking.  The contents are identical in the case since the world has no dependencies, but if that wasn't the case then they would be different.

Anyway, yes, I can delete the generated one and use `world.wit` for both purposes.  If we ever make this test more complicated such that the world has dependencies, we'll need to be careful to add those inline to `world.wit` since that's what `wasm-component-ld` will expect.

src/tests/nativeaot/SmokeTests/SharedLibrary/Library.cs
SingleAccretion: > but I don't quite understand the motivation for this suggestion

The main motivation is simplifying this test. Take the current call chain for `ReturnsPrimitiveInt` for example. Right now it is:
```
LibraryWorldImpl::ReturnsPrimitiveInt [implementation]
[There is a static abstract interface ILibraryWorld that I don't think has any significant purpose]
LibraryWorld::wasmExportReturnsPrimitiveInt [Actual UCO export]
```
These are "not strictly necessary" steps and types.

The more broader point, though, is that I would consider "handrolling" here a feature, not a bug (including when adding new entrypoints).

dicej: Yeah, I can see how a "only what's strictly necessary" approach can simplify debugging; I don't necessarily agree we need to take it that far, but I'm not going to be stubborn about it.  I'll eliminate the generated code.

dicej:
This should address https://github.com/dotnet/runtimelab/issues/2648, I believe.

SingleAccretion (approved):
LGTM, thank you!

jkotas:
cc @pavelsavara

pavelsavara (approved):

PR #2779, opened 2024-11-15
[SwiftBindings] Binding process doc

stephen-hawley:
This is a document to describe the process of binding Swift entities in C# and propose an architecture for handling it cleanly.

### DIFF
--- a/src/docs/process-binding.md
+++ b/src/docs/process-binding.md
@@ -1,0 +1,87 @@
+# The Process of Binding
+
+In theory, the process of binding a Swift binary into C# should be as simple as:
+
+1. Generate and consume the abi.json file
+2. Extract and demangle the symbols from the binary
+3. Iterate over every type and function and gnerate C# and/or supporting Swift code
+4. Generate type datatbase entries
+5. Optionally compile generated code
+
+Unfortunately, it is not so simple. There are complications that need to be considered because of special cases in the Swift language as well as special cases or limitations in C#.
+
+For example
+- `open` Swift classes need to be handled differently from `public` swift classes.
+- Structs and enums that are `@frozen` are passed differently from structs that are not. 
+- Structs and enums that contain or may contain non-blitable members need to be defined differently.
+- Enums come in different forms that lend themselves to different representations in C#.
+- C# doesn't have methods or properties on enums so those need to be put into extension methods.
+- Any protocol implementation in C# will need proxy types in both C# and Swift
+- P/invokes and `[UnmanagedCallersOnly]` entrypoints (reverse p/invokes) need to be defined in a separate class from the bound type they support if that bound type is generic.
+- Mutating methods on value types need to be handled differently from non-mutating methods.
+- C# does not support methods that are homonyms.
+
+All of this makes the process of generating code challenging.
+
+The complexities fall into several broad categories:
+- Type and member naming
+- Multiple types being defined in multiple languages concurrently
+- Marhshaling handled differently based on the type of the parameter and the type of the function
+- Implicit arguments
+- Versioning based on the `@available` attribute.
+
+For this reason I strongly recommend using code-generation tools that can work in a non-linear fashion. There are several ways to achieve this, but I would strongly recommend using the Dynamo framework from Binding Tools for Swift as it can handle both C# and Swift and generates non-linearly. In addition, it shouldn't be a stretch to generate code in parallel on type boundaries.
+
+Because of this, I think we should adopt a strategy and factory pattern for handlers at various levels.
+
+The general pattern would work like this:
+1. Start with a Swift language entity
+2. Aggregate information about that entity
+3. Select a factory to create a handler for that entity
+4. The handler will generate a context object for handling that entity
+5. Execute a series of steps through the handler that will do work apropriate for each step.
+6. Aggregate the result and generate as needed.
+
+Handlers will contain factories for handling sub steps, if needed.
+
+For example, given the following Swift code:
+```Swift
+public func generateAClass(String name) -> SomeClass { }
+```
+The process would look at this and identify this as a top-level function and will select a handler factory for it.
+The handler will create a context for the object which would include a class for the top-level object to live in (C# doesn't have top-level functions) and a class to hold top-level pinvokes and a function generation context which would include a place to place function argument declarations, generic declarations, function argument pre-marshaling code, pinvoke argument declarations, pinvoke argument expressions, post-marshaling code, return type declaration, and a return expression. 
+
+The handler will execute a step to name the function and the associated pinvoke, including the entry point and library.
+Then for each argument, it will gather information about each argument and from the function handler get a factory to build an argument handler for type `String`. This will in turn name the argument, generate the C# type and add it to the C# argument declaration. It will define the argument type for the pinvoke and add it to the C# pinvoke argument list. If needed, it will generate premarshal code and add it to the premarshal list and post marshal code, and finally an expression for calling the pinvoke.
+
+A similar process will be done for handling the return type and value. In this case, the pinvoke return type will be a `NativeHandle` and it will be used in conjunction with a registry to either retrieve an already existing C# object that is bound to that handle or it will build one through a factory.
+
+After all this is done, the function handler will finish up by aggregating all the information, writing the C# method and writing the C# pinvoke.
+
+By doing this, all the special cases get isolated in their own code and state related to the code, if any, can be kept in the handler's context object.
+
+For example, if there is a pre-marshal set up and a post marshal teardown, that code can be written in the same step.
+
+## Handler Factory Selection
+
+A handler factory for a given entity will be given a blob of information about the prospective entity. It will have a try-get predicate to answer whether or not it can handle that entity and a method to build a handler.
+
+A generalized handler might look like this:
+
+```csharp
+public interface Handler<EntityInformation, ParentContext> {
+    void Begin (ParentContext parentContext, EntityInformation info);
+    void Process ();
+    void End ();
+}
+```
+
+A factory might look like this:
+```csharp
+public interface Factory<EntityInformation, ParentContext> {
+    bool TryGetHandler (EntityInformation info, [NotNullWhen(true)] out Handler<EntityInformation, ParentContext>? handler);
+}
+```
+
+Finally, since Swift allows inner types, all of these steps should be naturally recursive.
+It should also be noted that all entities at top level can be processed in parallel trivially. Inner types can also be processed in parallel at the scope at which they're declared, but some care needs to be taken to ensure that two inner types at the same scope don't step on each other in writing to the non-linear representation or have issues when creating p/invokes. This could be done with appropriate locks or using `Concurrent{Dictionary/List}` for the representational types.
--- a/src/docs/process-binding.md
+++ b/src/docs/process-binding.md
@@ -2,11 +2,12 @@
 
 In theory, the process of binding a Swift binary into C# should be as simple as:
 
-1. Generate and consume the abi.json file
-2. Extract and demangle the symbols from the binary
-3. Iterate over every type and function and gnerate C# and/or supporting Swift code
-4. Generate type datatbase entries
-5. Optionally compile generated code
+1. Read the type database from shipped resources as well as from any user-supplied files
+2. Generate and consume the abi.json file, populating the type database
+3. Extract and demangle the symbols from the binary
+4. Iterate over every type and function and generate C# and/or supporting Swift code
+5. Write the a type database file for every new entity
+6. Optionally compile generated code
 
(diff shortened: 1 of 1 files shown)


src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
+# The Process of Binding
+
+In theory, the process of binding a Swift binary into C# should be as simple as:
+
+1. Generate and consume the abi.json file
+2. Extract and demangle the symbols from the binary
+3. Iterate over every type and function and gnerate C# and/or supporting Swift code
```
matouskozak:
```suggestion
3. Iterate over every type and function and generate C# and/or supporting Swift code
```

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+4. Generate type datatbase entries
```
matouskozak:
```suggestion
4. Generate type database entries
```

matouskozak: How do you envision the type database structure and usage? In the context of this PR, are the handlers supposed to populate the database as they process the entity?

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+- P/invokes and `[UnmanagedCallersOnly]` entrypoints (reverse p/invokes) need to be defined in a separate class from the bound type they support if that bound type is generic.
+- Mutating methods on value types need to be handled differently from non-mutating methods.
+- C# does not support methods that are homonyms.
+
+All of this makes the process of generating code challenging.
+
+The complexities fall into several broad categories:
+- Type and member naming
+- Multiple types being defined in multiple languages concurrently
+- Marhshaling handled differently based on the type of the parameter and the type of the function
```
matouskozak:
```suggestion
- Marshaling handled differently based on the type of the parameter and the type of the function
```

matouskozak (approved):
Nice high-level description!

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
+# The Process of Binding
+
+In theory, the process of binding a Swift binary into C# should be as simple as:
+
+1. Generate and consume the abi.json file
```
kotlarmilos: I would like to see next level of details here, similar to https://github.com/dotnet/runtimelab/tree/feature/swift-bindings/docs#functional-outline. It is not clear how dependencies are resolved.

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+2. Extract and demangle the symbols from the binary
```
kotlarmilos: Do we want to consider this step as optional? For simpler bindings, this may not be necessary. Also, users might not know how to retrieve it from a framework.

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+- Enums come in different forms that lend themselves to different representations in C#.
+- C# doesn't have methods or properties on enums so those need to be put into extension methods.
+- Any protocol implementation in C# will need proxy types in both C# and Swift
+- P/invokes and `[UnmanagedCallersOnly]` entrypoints (reverse p/invokes) need to be defined in a separate class from the bound type they support if that bound type is generic.
+- Mutating methods on value types need to be handled differently from non-mutating methods.
+- C# does not support methods that are homonyms.
+
+All of this makes the process of generating code challenging.
+
+The complexities fall into several broad categories:
```
kotlarmilos, line 27: I think it's important to reflect on the components of our tooling. While having everything in a single project might seem easier, splitting them into multiple projects would force cleaner integration and make testing more granular.

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+- Type and member naming
+- Multiple types being defined in multiple languages concurrently
+- Marhshaling handled differently based on the type of the parameter and the type of the function
+- Implicit arguments
+- Versioning based on the `@available` attribute.
+
+For this reason I strongly recommend using code-generation tools that can work in a non-linear fashion. There are several ways to achieve this, but I would strongly recommend using the Dynamo framework from Binding Tools for Swift as it can handle both C# and Swift and generates non-linearly. In addition, it shouldn't be a stretch to generate code in parallel on type boundaries.
```
kotlarmilos: I would start by identifying limitations with the string-based emitter. Based on those, we can add a thin model layer if needed. I think that the emitter itself shouldn't handle marshalling; that should be done before emitting.

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+
+Because of this, I think we should adopt a strategy and factory pattern for handlers at various levels.
+
+The general pattern would work like this:
```
kotlarmilos, line 36: I like this approach with custom handlers :)

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+6. Aggregate the result and generate as needed.
+
+Handlers will contain factories for handling sub steps, if needed.
+
+For example, given the following Swift code:
+```Swift
+public func generateAClass(String name) -> SomeClass { }
+```
+The process would look at this and identify this as a top-level function and will select a handler factory for it.
+The handler will create a context for the object which would include a class for the top-level object to live in (C# doesn't have top-level functions) and a class to hold top-level pinvokes and a function generation context which would include a place to place function argument declarations, generic declarations, function argument pre-marshaling code, pinvoke argument declarations, pinvoke argument expressions, post-marshaling code, return type declaration, and a return expression.
```
kotlarmilos, line 89: We should be able to hold the most of information in declarations and the registrar.

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+The handler will create a context for the object which would include a class for the top-level object to live in (C# doesn't have top-level functions) and a class to hold top-level pinvokes and a function generation context which would include a place to place function argument declarations, generic declarations, function argument pre-marshaling code, pinvoke argument declarations, pinvoke argument expressions, post-marshaling code, return type declaration, and a return expression. 
+
+The handler will execute a step to name the function and the associated pinvoke, including the entry point and library.
+Then for each argument, it will gather information about each argument and from the function handler get a factory to build an argument handler for type `String`. This will in turn name the argument, generate the C# type and add it to the C# argument declaration. It will define the argument type for the pinvoke and add it to the C# pinvoke argument list. If needed, it will generate premarshal code and add it to the premarshal list and post marshal code, and finally an expression for calling the pinvoke.
```
jkurdek:
```suggestion
Then for each argument, it will gather the necessary information and from the function handler get a factory to build an argument handler for type `String`. This will in turn name the argument, generate the C# type and add it to the C# argument declaration. It will define the argument type for the pinvoke and add it to the C# pinvoke argument list. If needed, it will generate premarshal code and add it to the premarshal list and post marshal code, and finally an expression for calling the pinvoke.
```

src/docs/process-binding.md
jkurdek: Im not sure about the UX of making it optional. The user will run the code without this step and then get a failure?

src/docs/process-binding.md
jkurdek, line 27: Do those complexities impose additional steps over the five you described, or they just make the steps more complicated?

src/docs/process-binding.md
jkurdek: I am all for reusing code which already works. However, I think that two things would be useful first:
- Describing the limitations as Milos noted
- Describing what Dynamo is, and how it will solve those limitations

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+For this reason I strongly recommend using code-generation tools that can work in a non-linear fashion. There are several ways to achieve this, but I would strongly recommend using the Dynamo framework from Binding Tools for Swift as it can handle both C# and Swift and generates non-linearly. In addition, it shouldn't be a stretch to generate code in parallel on type boundaries.
+
+Because of this, I think we should adopt a strategy and factory pattern for handlers at various levels.
+
+The general pattern would work like this:
+1. Start with a Swift language entity
+2. Aggregate information about that entity
+3. Select a factory to create a handler for that entity
+4. The handler will generate a context object for handling that entity
+5. Execute a series of steps through the handler that will do work apropriate for each step.
```
jkurdek: NIT, same on the line below :) 
```suggestion
5. Execute a series of steps through the handler that will do work apropriate for each step
```

src/docs/process-binding.md
jkurdek: I got a bit lost I think. So the `handler` doing all the job will be the `Top Level functions handler`? And then the `Top Level function handler` will call `function handler` to get a `factory` which will build an `argument handler`?

src/docs/process-binding.md
```diff
@@ -0,0 +1,87 @@
...
+
+The handler will execute a step to name the function and the associated pinvoke, including the entry point and library.
+Then for each argument, it will gather information about each argument and from the function handler get a factory to build an argument handler for type `String`. This will in turn name the argument, generate the C# type and add it to the C# argument declaration. It will define the argument type for the pinvoke and add it to the C# pinvoke argument list. If needed, it will generate premarshal code and add it to the premarshal list and post marshal code, and finally an expression for calling the pinvoke.
+
+A similar process will be done for handling the return type and value. In this case, the pinvoke return type will be a `NativeHandle` and it will be used in conjunction with a registry to either retrieve an already existing C# object that is bound to that handle or it will build one through a factory.
+
+After all this is done, the function handler will finish up by aggregating all the information, writing the C# method and writing the C# pinvoke.
```
jkurdek, line 96: In this example we do not need to generate Swift code, but what about cases where we will need to do this. Will it happen here as well?

src/docs/process-binding.md
stephen-hawley: I figured that since we need to know the path to the dylib to get the abi.json, we're already there. Since we will need metadata accessors, we need the demangling there.

src/docs/process-binding.md
stephen-hawley: Will do.

src/docs/process-binding.md
stephen-hawley: It makes sense to do so. The reason being that in any cases where marshaling is not 1:1 with C# capabilities, we will need to change the way that parameters are handled.  I'll expound on this more in the docs because it will make it clear why non-linear code writing will make tasks much easier.

src/docs/process-binding.md
stephen-hawley: I envision there being likely two type databases. The first is a bind-time database which would need maximal information about the types: Swift module, Swift name, C# namespace, C# name, entity type, blitability. But this is something that will be in its own document. The database would be populated with new entries after the ability.json file has been parsed but before binding. Older entries can be read in before the abi.json file is read. The second database is run-time. Anything we can do to minimize the size and start-up overhead. For the most part, the run-time database is to support the generic programming model, but again, this would be it's own document.

dalexsoto (approved):

vitek-karas:
A general concern I have is that we haven't really described when/how are we going to generate Swift and how are we going to use/ship it. On top of that, trimming is still a major concern for me. Especially the fact that we don't have a good way to trim generated swift code.

Other interops solve this problem by generating the native code (in this case Swift) as basically the last step of building an app after it's known what interop is needed by the app. This means that the swift generation would happen at app-build time and not during the projection tool runtime. There are probably other ways to solve it, but we haven't really discussed these yet.

I think we should solve these:
- What is our position towards trimmability and the expectations of the behavior of that aspect of the product. Basically how are we solving the scaling problem of pregenerated interop (the extreme case of this is Win32 API which has ~100K methods and the whole thing simply doesn't scale if pregenerated since it's way too big).
- How we generate Swift and how do we ship it. For example, if we decide to pregenerate swift upfront, how are we going to ship that in a NuGet package so that the user can consume a prepared binding library easily. How is that going to be integrated with the final app's build? Or do we postpone Swift generation to the app's build so that we don't have to distribute it with the binding library.
- I would be interested in the general desired UX for consuming bindings. I expected this to be the same as referencing a NuGet package, but maybe we need something else?

stephen-hawley:
@vitek-karas - I understand your concern about trimming and packaging. I think both of those things are beyond the scope of this particular document.
For the sake of addressing these things, I would worry about binding on demand as it would adversely affect build/deploy/launch times. One of the main takeaways from this past team week was customer concerns about this.
There are several ways to approach trimming swift:
1. Trim on dylib boundaries - easy, but only does so much, especially with larger frameworks
2. Deliver .o files or static libraries and link to them with NativeAOT which should trim out all dead Swift code
3. Write every Swift entity in a way that has conditional compilation switches on them and based on type usage in C#, recompile the already done binding such that it will exclude unused types. This is obviously sub-optimal in terms of build times.

jkurdek (approved):

PR #2780, opened 2024-11-15
Status and incomplete fat `Precode` pointer.

AaronRobinsonMSFT:
Update status document.
Incomplete fat pointer implementation using `Precode`.

### DIFF
--- a/experiment_status.md
+++ b/experiment_status.md
@@ -6,11 +6,13 @@
   - The `MethodDesc::DoPrestub()` entry point should be referenced for many of the APIs that can produce `PCODE`. (for example, `GetStubForInteropMethod()`).
 - The ILStubResolver deletes the associated IL post JIT - `ILStubResolver::ClearCompileTimeState`.
   - This means that when an IL stub will be the interpreted, clean-up of generated IL needs to be avoided.
+    - If a future CoreCLR interpreter uses a converted byte code approach, then getting rid of the IL may not be an issue.
   - See `CEEInfo::getMethodInfo()` for how IL stubs are impacted, see `MethodDesc::IsWrapperStub()`.
 - Created a new `Precode` type to represent the interpreter entry point.
   - See `MethodDescCallSite` for the common ways CoreCLR calls from C++ into managed code. Specifically, `CallTargetWorker()` is of note.
   - The `Precode` approach seems to be a valid way to loop into the execution engine without too much disruption.
 - Passing around stub context to P/Invokes needs to be considered in the interpreter case.
+  - Disabling P/Invoke stub sharing would avoid this need. ReadyToRun is prior art for disabling stub sharing.
 
 # Status 01-NOV-2024
 
--- a/experiment_status.md
+++ b/experiment_status.md
@@ -1,3 +1,17 @@
+# Status 15-NOV-2024
+
+## Aaron's CoreCLR learnings
+
+- `PCODE` manifests itself from a myriad of places.
+  - The `MethodDesc::DoPrestub()` entry point should be referenced for many of the APIs that can produce `PCODE`. (for example, `GetStubForInteropMethod()`).
+- The ILStubResolver deletes the associated IL post JIT - `ILStubResolver::ClearCompileTimeState`.
+  - This means that when an IL stub will be the interpreted, clean-up of generated IL needs to be avoided.
+  - See `CEEInfo::getMethodInfo()` for how IL stubs are impacted, see `MethodDesc::IsWrapperStub()`.
+- Created a new `Precode` type to represent the interpreter entry point.
+  - See `MethodDescCallSite` for the common ways CoreCLR calls from C++ into managed code. Specifically, `CallTargetWorker()` is of note.
+  - The `Precode` approach seems to be a valid way to loop into the execution engine without too much disruption.
+- Passing around stub context to P/Invokes needs to be considered in the interpreter case.
+
 # Status 01-NOV-2024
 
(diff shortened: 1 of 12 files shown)


experiment_status.md
```diff
@@ -1,3 +1,17 @@
+# Status 15-NOV-2024
+
+## Aaron's CoreCLR learnings
+
+- `PCODE` manifests itself from a myriad of places.
+  - The `MethodDesc::DoPrestub()` entry point should be referenced for many of the APIs that can produce `PCODE`. (for example, `GetStubForInteropMethod()`).
+- The ILStubResolver deletes the associated IL post JIT - `ILStubResolver::ClearCompileTimeState`.
+  - This means that when an IL stub will be the interpreted, clean-up of generated IL needs to be avoided.
```
jkotas, line 8: This is correct for the current CoreCLR interpreter that interprets IL directly. It is not correct for the current Mono interpreter that "JITs" the IL into its own bytecode.

experiment_status.md
```diff
@@ -1,3 +1,17 @@
...
+  - See `CEEInfo::getMethodInfo()` for how IL stubs are impacted, see `MethodDesc::IsWrapperStub()`.
+- Created a new `Precode` type to represent the interpreter entry point.
+  - See `MethodDescCallSite` for the common ways CoreCLR calls from C++ into managed code. Specifically, `CallTargetWorker()` is of note.
+  - The `Precode` approach seems to be a valid way to loop into the execution engine without too much disruption.
+- Passing around stub context to P/Invokes needs to be considered in the interpreter case.
```
jkotas, line 14: Alternatively, we can disable P/Invoke stub sharing with the interpreter to keep things simpler. (We have not bothered with P/Invoke stub sharing in R2R images for the same reason.)

src/coreclr/vm/amd64/CallDescrWorkerAMD64.asm
```diff
@@ -42,6 +42,11 @@ StackCopyLoop:                          ; copy the arguments to stack top-down t
 
         mov     rax, [rbx + CallDescrData__dwRegTypeMap] ; save the reg (arg) type map
 
+        ; Store stub context
+        cmp     QWORD PTR [rbx + CallDescrData__pStubContextMD], 0
+        jz      NoStubContext
```
janvorli, line 47: I wonder if it would be a problem to load the r10 unconditionally here since r10 is volatile.

src/coreclr/vm/jitinterface.cpp
```diff
@@ -7697,7 +7697,7 @@ CEEInfo::getMethodInfo(
         getMethodInfoHelper(cxt, methInfo, context);
         result = true;
     }
-    else if (!ftn->IsWrapperStub() && ftn->HasILHeader())
+    else if (ftn->HasILHeader()) // !ftn->IsWrapperStub() &&
```
janvorli, line 7700: I am just curious - why was this change necessary?

src/coreclr/vm/callhelpers.cpp
```diff
@@ -234,18 +238,22 @@ void * DispatchCallSimple(
...
-            &callDescrData,
-            dwDispatchCallSimpleFlags & DispatchCallSimple_CriticalCall);
-    }
-    else
-    {
-        CallDescrWorkerWithHandler(&callDescrData, dwDispatchCallSimpleFlags & DispatchCallSimple_CriticalCall);
-    }
-
-    return *(void **)(&callDescrData.returnValue);
+    ARG_SLOT result;
```
janvorli, line 241: A note for future - this will not be sufficient e.g. for x64 Unix ABI that can return result in two general purpose registers or two xmm registers.

src/coreclr/vm/amd64/CallDescrWorkerAMD64.asm
AaronRobinsonMSFT: How do we know that `r10` is volatile in this case? I'm hoping we can simply avoid stub sharing in interpreter scenarios.

src/coreclr/vm/jitinterface.cpp
AaronRobinsonMSFT: Yep, quite surprising. The issue here is the instantiating and boxing stubs through away all the IL during the JIT process and for some reason don't come down here during JIT scenarios. I didn't dig in any more, but this is one of the many ways in which the current interpreter relies on not doing full interpretation. There are so many places where it relies on JITted code and that makes the current design quite useless. Integrating this in CoreCLR is going to require a reimagining of the contract.

src/coreclr/vm/callhelpers.cpp
AaronRobinsonMSFT: Agree. This was merely for working on x64 Windows.

src/coreclr/vm/amd64/CallDescrWorkerAMD64.asm
janvorli: The r10 is volatile in both the windows and unix x64 calling convention. And this function is only called from native runtime code, so there is no case when a special calling convention extension that we use for some stubs could expect r10 to be preserved.

AaronRobinsonMSFT: Oh! You mean volatile in the calling convention sense, not `volatile` in the memory sense. Yes, I think we can unconditionally play with `r10`. I added the check just in case it was needed by another scenario that I didn't know about.

PR #2781, opened 2024-11-16
awaits of Task-returning methods from async2 are done as thunk calls.

VSadov:
Two changes here:

* bump the Roslyn reference to a version that emits awaits to Task-returning methods as async2 thunk calls when the caller is an async2 method
I.E - regardless if the calee is an actual async2 method or just something that returns Task/ValueTask, compiler emits a call to an async2 method (actual or a thunk).  
This way the calee can swap back and forth between async1 and async2 without binary-breaking the call.

awaits of custom task-like things, which are not Task/ValueTask are still emitted as GetAwaiter+etc pattern.

* tried to fix the a2->a1 thunks at least enough to pass tests.  (incomplete)

### DIFF
--- a/buildroslynnugets.cmd
+++ b/buildroslynnugets.cmd
@@ -1,7 +1,7 @@
 setlocal ENABLEEXTENSIONS
 pushd %~dp0
-set ASYNC_ROSLYN_COMMIT=8bb20e76b6d7e041a2cd0a3152d95fdff6578050
-set ASYNC_SUFFIX=async-9
+set ASYNC_ROSLYN_COMMIT=fa251ef06afcc9d1c79761a112b89c6341c0ccfa
+set ASYNC_SUFFIX=async-10
 set ASYNC_ROSLYN_BRANCH=demos/async2-experiment1
 
 cd ..
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -43,9 +43,9 @@
       Any tools that contribute to the design-time experience should use the MicrosoftCodeAnalysisVersion_LatestVS property above to ensure
       they do not break the local dev experience.
     -->
-    <MicrosoftCodeAnalysisCSharpVersion>4.13.0-async-9</MicrosoftCodeAnalysisCSharpVersion>
-    <MicrosoftCodeAnalysisVersion>4.13.0-async-9</MicrosoftCodeAnalysisVersion>
-    <MicrosoftNetCompilersToolsetVersion>4.13.0-async-9</MicrosoftNetCompilersToolsetVersion>
+    <MicrosoftCodeAnalysisCSharpVersion>4.13.0-async-10</MicrosoftCodeAnalysisCSharpVersion>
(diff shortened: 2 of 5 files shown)


VSadov:
As we did not test scenarios involving a2->a1 thunks, there are obvious issues. (no support for nongeneric Task, no support for ValueTask, shared generics seems to have problems,...)

I tried to fix what is obvious, but did not get the tests to pass. 
Shared generics test still fails with:

```
System.AggregateException: One or more errors occurred. (Assert.Equal() Failure: Values differ
Expected: typeof(int)
Actual:   null)
 ---> Xunit.Sdk.EqualException: Assert.Equal() Failure: Values differ
Expected: typeof(int)
Actual:   null
   at Xunit.Assert.Equal[T](T expected, T actual, IEqualityComparer`1 comparer) in /_/src/Microsoft.DotNet.XUnitAssert/src/EqualityAsserts.cs:line 174
   at GenericClass`1.StaticMethodAsync1(Type t) in E:\dotnetA2\runtimelab\src\tests\async\shared-generic.cs:line 83
   at GenericClass`1.StaticMethodAsync1(Type t)
   at Async2SharedGeneric.Async2EntryPoint[T](Type t, T value) in E:\dotnetA2\runtimelab\src\tests\async\shared-generic.cs:line 39
   at System.Runtime.CompilerServices.RuntimeHelpers.DispatchContinuations(Continuation continuation)
   at System.Runtime.CompilerServices.RuntimeHelpers.FinalizeTaskReturningThunk(Continuation continuation)
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at Async2SharedGeneric.TestEntryPoint() in E:\dotnetA2\runtimelab\src\tests\async\shared-generic.cs:line 16
   at Program.<<Main>$>g__TestExecutor9|0_10(StreamWriter tempLogSw, StreamWriter statsCsvSw, <>c__DisplayClass0_0&) in E:\dotnetA2\runtimelab\artifacts\tests\coreclr\obj\windows.x64.checked\Managed\async\asynctests\generated\XUnitWrapperGenerator\XUnitWrapperGenerator.XUnitWrapperGenerator\FullRunner.g.cs:line 288
```

We should be passing a type object to the call, but somehow we see `null` on the other side. Not sure how that happens.

jakobbotsch (approved):

jakobbotsch:
I took a quick look and it looks like the VM is not telling the JIT that the async2 to async1 thunk is actually a runtime-async function, so the JIT does not generate a state machine for it. I can push a fix to this PR.

jakobbotsch:
There were some more missing signatures for the helper methods used by the AsyncToTask thunks that I pushed implementations for. Now this PR passes all tests again.

VSadov:
Thank you!!

PR #2786, opened 2024-11-18
[NativeAOT-LLVM] Add WasmExternalDwarf property to create an external_debug_info section

yowl:
This PR adds a `WasmExternalDwarf ` build property, `true` by default for `browser` that will add a `external_debug_info` section and place the DWARF info a separate `.debug.wasm` file.  `wasmtime` does not support this, at present, so only applies for the browser build.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
@@ -107,8 +107,8 @@
     <Delete Files="$(PublishDir)\$(TargetName).wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
     <Copy SourceFiles="$(NativeOutputPath)$(TargetName).wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
 
-    <Delete Files="$(PublishDir)\$(TargetName).wasm.debug.wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
-    <Copy SourceFiles="$(NativeOutputPath)$(TargetName).wasm.debug.wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
+    <Delete Files="$(PublishDir)\$(TargetName).debug.wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
+    <Copy SourceFiles="$(NativeOutputPath)$(TargetName).debug.wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
 
     <ItemGroup>
       <_FilesToCopyToNative Include="$(NativeOutputPath)\dotnet.native.js" />
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
@@ -106,7 +106,10 @@
 
     <Delete Files="$(PublishDir)\$(TargetName).wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
     <Copy SourceFiles="$(NativeOutputPath)$(TargetName).wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
-    
+
+    <Delete Files="$(PublishDir)\$(TargetName).wasm.debug.wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
+    <Copy SourceFiles="$(NativeOutputPath)$(TargetName).wasm.debug.wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
+
     <ItemGroup>
       <_FilesToCopyToNative Include="$(NativeOutputPath)\dotnet.native.js" />
       <_FilesToCopyToNative Include="$(NativeOutputPath)\dotnet.native.wasm" />
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
@@ -107,8 +107,8 @@
     <Delete Files="$(PublishDir)\$(TargetName).wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
     <Copy SourceFiles="$(NativeOutputPath)$(TargetName).wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
 
-    <Delete Files="$(PublishDir)\$(TargetName).debug.wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
-    <Copy SourceFiles="$(NativeOutputPath)$(TargetName).debug.wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
+    <Delete Files="$(PublishDir)\$(TargetName).debug.wasm" Condition="'$(_targetOS)' == 'browser'"/>
+    <Copy SourceFiles="$(WasmSymbolFile)" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and Exists('$(WasmSymbolFile)')"/>
 
     <ItemGroup>
       <_FilesToCopyToNative Include="$(NativeOutputPath)\dotnet.native.js" />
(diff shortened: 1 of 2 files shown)


yowl:
Cc @dotnet/nativeaot-llvm

pavelsavara:
Do we need to do something to add the `.debug.wasm` file to the static assets ?
Do we want to do the same in upstream too ?

SingleAccretion:
> Do we want to do the same in upstream too ?

It is a bit less pressing in upstream's case because the managed `.pdb`s are separate already and native debug info is off by default, but I think it would be a good feature to both enable that native debug info by default and use this option to have it not affect size.

However, I foresee this may be tricky in the upstream's case because of the names section. The names section is used by the engines to print out stack traces on traps (unhandled exceptions, asserts, ...). It would have to be tested whether the browsers out there are eager to download this sidecar file (that will contain the names section stripped from the main file) to print these stack traces. I suspect not.

In the NAOT's case this is not a big problem because we store the stack trace metadata directly in the executable, and are not super interested in native runtime stacks (since this 'native' part is small).

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -35,6 +35,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <IlcTreatWarningsAsErrors Condition="'$(IlcTreatWarningsAsErrors)' == ''">$(TreatWarningsAsErrors)</IlcTreatWarningsAsErrors>
     <_IsiOSLikePlatform Condition="'$(_targetOS)' == 'maccatalyst' or $(_targetOS.StartsWith('ios')) or $(_targetOS.StartsWith('tvos'))">true</_IsiOSLikePlatform>
     <_IsApplePlatform Condition="'$(_targetOS)' == 'osx' or '$(_IsiOSLikePlatform)' == 'true'">true</_IsApplePlatform>
+    <WasmExternalDwarf Condition="'$(WasmExternalDwarf)' == ''and '$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'">true</WasmExternalDwarf>
```
SingleAccretion: There is already a documented property of this in the NAOT toolchain: `StripSymbols`. We can use it.

Also, why disable for `DotNetJsApi`? I think it makes sense for `DotNetJsApi` as well.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -575,6 +581,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <!-- see https://github.com/emscripten-core/emscripten/issues/16836 for the issue that means we have to force the linker to include these symbols before LTO -->
       <!--<CustomLinkerArg Condition="'$(Optimize)' != 'true'" Include="$(WasmOptimizationSetting) -flto" /> -->
       <CustomLinkerArg Condition="'$(IlcLlvmTarget)' != ''" Include="-target $(IlcLlvmTarget)" />
+
+	    <CustomLinkerArg Condition="'$(WasmExternalDwarf)' == 'true'" Include="-gseparate-dwarf" />
```
```suggestion
	    <CustomLinkerArg Condition="'$(WasmExternalDwarf)' == 'true'" Include="-gseparate-dwarf" />
```
I think we should pass something like `GetFilenameWithoutExtension($(NativeBinary)).debug.wasm` here to avoid the odd `.wasm.debug.wasm` thing.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -439,12 +443,14 @@ The .NET Foundation licenses this file to you under the MIT license.
       <WasmCompilerPath Condition="'$(EMSDK)' == ''">&quot;$(EmscriptenSdkToolsPath)bin/clang++&quot;</WasmCompilerPath>
       <WasmLinkerPath Condition="'$(EMSDK)' != ''">&quot;$(EMSDK)/upstream/emscripten/emcc$(EmccScriptExt)&quot;</WasmLinkerPath>
       <WasmLinkerPath Condition="'$(EMSDK)' == ''">&quot;$(EmscriptenSdkToolsPath)emscripten/emcc$(EmccScriptExt)&quot;</WasmLinkerPath>
+      <WasmDwarfLinkerPath Condition="'$(EMSDK)' != ''">&quot;$(EMSDK)/upstream/bin/llvm-dwp$(ExeExt)&quot;</WasmDwarfLinkerPath>
+      <WasmDwarfLinkerPath Condition="'$(EMSDK)' == ''">&quot;$(EmscriptenSdkToolsPath)/bin/llvm-dwp$(ExeExt)&quot;</WasmDwarfLinkerPath>
```
SingleAccretion: Looks like leftovers from DWP (here and above with `DwarfObjects`).

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: Probably, I didn't really know, and just copied what was in the publish file for the `.wasm` file.  Will remove it.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: thanks

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -100,6 +100,7 @@ The .NET Foundation licenses this file to you under the MIT license.
 
     <NativeObject>$(NativeIntermediateOutputPath)$(TargetName)$(NativeObjectExt)</NativeObject>
     <NativeBinary Condition="'$(NativeBinary)' == ''">$(NativeOutputPath)$(TargetName)$(NativeBinaryExt)</NativeBinary>
+    <WasmBinary Condition="'$(_targetArchitecture)' == 'wasm'">$(NativeOutputPath)$(TargetName).wasm</WasmBinary>
```
```suggestion
```
Unused.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -390,9 +391,11 @@ The .NET Foundation licenses this file to you under the MIT license.
     <ItemGroup>
       <LlvmObjects Include="@(_IlcProducedFiles)" Condition="'%(Extension)' == '$(LlvmObjectExt)'">
         <NativeObject>%(RelativeDir)%(Filename)$(NativeObjectExt)</NativeObject>
+        <DwarfObject>&quot;%(RelativeDir)%(Filename).dwo&quot;</DwarfObject>
       </LlvmObjects>
       <NativeObjects Include="@(_IlcProducedFiles)" Condition="'%(Extension)' == '$(NativeObjectExt)'" />
       <NativeObjects Include="@(LlvmObjects->'%(NativeObject)')" />
+      <DwarfObjects Include="@(LlvmObjects->'%(DwarfObject)')" />
```
```suggestion
      </LlvmObjects>
      <NativeObjects Include="@(_IlcProducedFiles)" Condition="'%(Extension)' == '$(NativeObjectExt)'" />
      <NativeObjects Include="@(LlvmObjects->'%(NativeObject)')" />
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
```diff
@@ -106,7 +106,10 @@
 
     <Delete Files="$(PublishDir)\$(TargetName).wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
     <Copy SourceFiles="$(NativeOutputPath)$(TargetName).wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
-    
+
+    <Delete Files="$(PublishDir)\$(TargetName).debug.wasm" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
+    <Copy SourceFiles="$(NativeOutputPath)$(TargetName).debug.wasm" DestinationFolder="$(PublishDir)" Condition="'$(_targetOS)' == 'browser' and '$(DotNetJsApi)' != 'true'"/>
```
SingleAccretion: This should happen for `DotNetJsApi` as well and check whether the `.debug.wasm` file is present to copy, ideally using `$(WasmSymbolFile)`.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -587,6 +591,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Include="-s TOTAL_STACK=$(IlcWasmStackSize)" />
       <CustomLinkerArg Condition="'$(WasmEnableJSBigIntIntegration)' == 'true'" Include="-s WASM_BIGINT=1" />
       <CustomLinkerArg Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'" Include="-s DISABLE_EXCEPTION_CATCHING=0" />
+	    <CustomLinkerArg Condition="'$(StripSymbols)' == 'true'" Include="-gseparate-dwarf=&quot;$(WasmSymbolFile)&quot;" />
```
SingleAccretion, line 591: ~~Hmm, does this embeds an absolute path, doesn't it. I think it would better to embed a relative path, i. e. just the file name. Otherwise I don't see it working even in our case where we copy from `/native` to `/publish`.~~

SingleAccretion: Actually, testing what this property does, this in fact the correct way to use it. Using an absolute path results in the relative path, and using a relative path result in something that is relative to the current working directory. Tricky but workable.

yowl: Not sure I follow the outcome of the comment here.  It is ok as is?  I think it was working when I tested it.

SingleAccretion: >  It is ok as is?

Yes.

SingleAccretion (approved):
LGTM, exciting to see this! Could you check that our `DotNetJs` test is still debuggable, just in case?

PR #2790, opened 2024-11-19
[RuntimeAsync] Add support for async struct instance methods (potential implementation 1)

jakobbotsch:
This adds support for async2 struct instance calls, with the restriction that the "this" argument passed to these calls in IL must be syntactically either `ldloca <some local>` or `ldarg 0`, with the latter only being allowed for self-instance calls. Furthermore, async2 struct instance methods are not allowed to use the `ldarga 0` or `starg 0` instructions.

Conceptually the VM and JIT rewrite these calls from "await x.Foo()" into something corresponding to `x = await Foo(x)`. Technically, the way this works is that on suspension, a struct-instance call will box its own instance such that it can restore it again on resumption. Additionally, during resumption we propagate the struct instance back to the caller continuation as if it was a return result: for self-instance calls resumption is done simply by overriding the "this" pointer from the callee, while the local calls resumption is done by overwriting the local on the stack frame.

The end result is that mutations are propagated and visible to the caller, even when the callee suspends (something that was not possible to accomplish in async1). This avoids having to introduce copies on the synchronous path, although for backwards compatibility it is expected that Roslyn will have to introduce these copies manually in many cases.


jakobbotsch:
cc @333fred @agocke @VSadov -- this includes some tests if you're interested in looking at the semantics (but the semantics just match normal struct methods with this implementation)

src/tests/async/struct.cs
```diff
@@ -0,0 +1,61 @@
...
+        public async2 Task Test()
+        {
+            S copy = this; // should be done by Roslyn for backwards compat
+            Assert.Equal(100, copy.Value);
+            copy.Value++;
+            await copy.InstanceCall();
+            Assert.Equal(101, copy.Value);
+
+            await copy.TaskButNotAsync();
+            Assert.Equal(102, copy.Value);
```
jakobbotsch, line 41: This copy is what I expect that Roslyn will have to insert for backwards compat. It can presumably be skipped for readonly functions.

jakobbotsch:
There are some other hard cases with this approach, e.g.
```csharp
private static async Task Foo()
{
    await Bar().Foo(); // cannot be emitted as an async2 await call
}

private static S s_s;
private static ref S Bar() => ref s_s;

private struct S
{
    public int X;

    public Task Foo()
    {
        X++;
        return Task.CompletedTask;
    }
}
```
Where Roslyn would be unable to emit the call in `Foo` as an async2 call.

VSadov:
Another complicated case:

```cs
    internal class Program
    {
        {
            Foo().Wait();
        }

        private static async Task Foo()
        {
            // prints 0
            Console.WriteLine(inst.Val);

            // we do not know at call site if we call an async method
            // or whether containing type is a struct, so who will make a copy?
            await inst.Foo();

            // prints 1
            Console.WriteLine(inst.Val);

            // we do not know at call site if we call an async method
            // or whether containing type is a struct, so who will make a copy?
            await inst.Bar();

            // still prints 1
            Console.WriteLine(inst.Val);
        }

        interface I1
        {
            int Val { get;}

            Task Foo();
            Task Bar();
        }

        private static I1 inst = new S();

        private struct S : I1
        {
            private int x;

            public int Val => x;

            public Task Foo()
            {
                x++;
                return Task.CompletedTask;
            }

            public async Task Bar()
            {
                x++;
            }
        }

    }
```

jakobbotsch:
> Another complicated case:
> 

The copies Roslyn would be inserting would be in the callee, not in the caller, so these cases do not look that complicated to me. See my example test case in this PR, it manually makes the copies I would expect Roslyn to insert.

jakobbotsch:
I looked at the IL produced for constrained calls and I think that would work fine with the restrictions outlined here, too.

I think the real nail in the coffin for this approach is in the cases I pointed at above. It is going to require Roslyn to fall back to async1 in a bunch of cases.
Initially I thought this would only be in rare cases where the LHS cannot be hoisted, and in other cases Roslyn could do the normal hoisting trick and write the struct instance back after a normal async2 call.
However, this hoisting would have observable behavior changes with current async1 behavior since the callee can observe that its mutations are not being done immediately to the expected instance. Think of something like
```csharp
async Task Foo(S[] s)
{
  await s[0].Bar(s); // This call site cannot be encoded as an async2 call with the current design -- copying to a local first is not equivalent
}

struct S
{
  public int X;
  public Task Bar(S[] s)
  {
    X++;
    Debug.Assert(X == s[0].X);
    return Task.CompletedTask;
  }
}
```
It's similar to my example above, but it shows that _any_ non-local and non-self call would have to be done as async1 calls given this design.

PR #2791, opened 2024-11-19
[NativeAOT-LLVM] Use `DebugType=none` instead of `NativeDebugSymbols=false`

SingleAccretion:
`DebugType` is a documented SDK property.

Also, remove the mention of stack traces not working with this since they actually work fine now.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (changes requested):
Thanks!

PR #2793, opened 2024-11-20
[SwiftBindings] SwiftHandle

stephen-hawley:
Added SwiftHandle to Swift.Runtime.

This class is taken from macios `NativeHandle` and renamed to `SwiftHandle`.

This type should be used any place a p/invoke into Swift would take a `void *` or `IntPtr` or for C# constructors for class bindings from an already existing Swift type. I would like to avoid `unsafe` code if we don't need it, especially in public APIs.

The reason being that if you have this type in Swift
```swift
public class IntBox {
    public var value: Int
    public init (value: Int) {
        self.value = value
    }
    public func same(other: IntBox) -> Bool {
        return other.value == self.value
    }
    public func same(other: Int) -> Bool {
        return other == self.value
    }
}
```
The natural C# binding might look like this would be (some details left out):
```csharp
public class IntBox {
    IntPtr handle;
    public IntBox (nint value) {
         handle = // pinvoke to init
    }
    public IntBox (IntPtr handle) { // construct from existing handle
        this.handle = handle;
    }
    public bool Same(IntBox other) {
        return PISame(handle, other) & 1 != 0
    }
    public bool Same(nint other) {
        return PISame(handle, other) & 1 != 0
    }
    [DllImport()]
    extern byte PISame (IntPtr handle, IntPtr other);
    [DllImport()]
    extern byte PISame (IntPtr handle, nint other);
}
```

Unfortunately, `nint` and `IntPtr` are indistinguishable so the two constructors and the two pinvokes are identical so the C# code won't compile.

### DIFF
--- a/src/Swift.Runtime/src/Library/SwiftHandle.cs
+++ b/src/Swift.Runtime/src/Library/SwiftHandle.cs
@@ -1,0 +1,143 @@
+using System;
+using System.Collections.Generic;
+
+#nullable enable
+
+namespace Swift.Runtime;
+
+/// <summary>
+/// Represents an opaque handle to a Swift object
+/// </summary>
+public readonly struct SwiftHandle : IEquatable<SwiftHandle> {
+    readonly IntPtr handle;
+
+    public IntPtr Handle => handle;
+
+    /// <summary>
+    /// Returns an SwiftHandle with a zero value
+    /// </summary>
+    public readonly static SwiftHandle Zero = default (SwiftHandle);
+
+    /// <summary>
+    /// Constructs a SwiftHandle from the given IntPtr
+    /// </summary>
+    public SwiftHandle (IntPtr handle)
+    {
+        this.handle = handle;
+    }
+
+    /// <summary>
+    /// Returns true if the SwiftHandle and the IntPtr are the same
+    /// </summary>
+   public static bool operator == (SwiftHandle left, IntPtr right)
+    {
+        return left.handle == right;
+    }
+
+    /// <summary>
+    /// Returns true if both SwiftHandles are the same
+    /// </summary>
+    public static bool operator == (SwiftHandle left, SwiftHandle right)
+    {
+        return left.handle == right.handle;
+    }
+
+    /// <summary>
+    /// Returns true if the IntPtr and the SwiftHandle are the same
+    /// </summary>
+    public static bool operator == (IntPtr left, SwiftHandle right)
+    {
+        return left == right.Handle;
+    }
+
+    /// <summary>
+    /// Returns true if the SwiftHandle and the IntPtr are different
+    /// </summary>
+    public static bool operator != (SwiftHandle left, IntPtr right)
+    {
+        return left.handle != right;
+    }
+
+    /// <summary>
+    /// Returns true if the IntPtr and the SwiftHandle are different
+    /// </summary>
+    public static bool operator != (IntPtr left, SwiftHandle right)
+    {
+        return left != right.Handle;
+    }
+
+    /// <summary>
+    /// Returns true if the SwiftHandles are different
+    /// </summary>
+    public static bool operator != (SwiftHandle left, SwiftHandle right)
+    {
+        return left.handle != right.Handle;
+    }
+
+
+    /// <summary>
+    /// Implicit conversion from SwiftHandle to IntPtr
+    /// </summary>
+    public static implicit operator IntPtr (SwiftHandle value)
+    {
+        return value.Handle;
+    }
+
+    /// <summary>
+    /// Implicit conversion from IntPtr to SwiftHandle
+    /// </summary>
+    public static implicit operator SwiftHandle (IntPtr value)
+    {
+        return new SwiftHandle (value);
+    }
+
+    /// <summary>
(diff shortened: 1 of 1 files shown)


jkotas:
> This type should be used any place a p/invoke into Swift would take a void * or IntPtr or for C# constructors for class bindings from an already existing Swift type.

This assumes that the low-level calling convention handling of `void*` is always the same as low-level calling convention handling of a struct type that wraps `void*`. It may be the case for architectures targeted by Swift today, but it is not a safe assumption to make in general.

stephen-hawley:
> > This type should be used any place a p/invoke into Swift would take a void * or IntPtr or for C# constructors for class bindings from an already existing Swift type.
> 
> This assumes that the low-level calling convention handling of `void*` is always the same as low-level calling convention handling of a struct type that wraps `void*`. It may be the case for architectures targeted by Swift today, but it is not a safe assumption to make in general.

Can you explain to me under what circumstances that would happen? It strikes me that that would be a fundamental change in Swift calling conventions and Swift structure packing, something that is already [defined by Apple](https://github.com/swiftlang/swift/blob/main/docs/ABI/TypeLayout.rst#fragile-struct-and-tuple-layout). If that changed, wouldn't it change pretty much all aspects of calling conventions and we would have to deal with that?

rolfbjarne (approved):

mcumming (approved):

kotlarmilos:
> public IntBox (IntPtr handle) { // construct from existing handle
        this.handle = handle;
    }

I would like to see a strong case for adding such layers, as they affect performance and increase maintenance costs. What is the practical use of this? Is it a constructor for bindings used internally?

stephen-hawley:
> > public IntBox (IntPtr handle) { // construct from existing handle
> > this.handle = handle;
> > }
> 
> I would like to see a strong case for adding such layers, as they affect performance and increase maintenance costs. What is the practical use of this? Is it a constructor for bindings used internally?

This is learning that was taken from a similar change in macios by @rolfbjarne to avoid similar problems in terms of signature conflicts. I don't see how this would introduce a performance issue as it is a fixed-sized struct and should just become a machine word when passed. I don't see how it increases maintenance costs - quite the opposite - it should reduce them by eliminating signature conflicts from the start.

kotlarmilos:
> This is learning that was taken from a similar change in macios by @rolfbjarne to avoid similar problems in terms of signature conflicts. 

Regarding the Objective-C example, how frequently was it used externally? Do we have any ballpark estimates? From my perspective, it seems to be mostly used internally.

> I don't see how this would introduce a performance issue as it is a fixed-sized struct and should just become a machine word when passed. 

I’m referring to the amount of emitted code. If we rely on JIT optimizations, that’s true, but the behavior might not always be deterministic.

> I don't see how it increases maintenance costs - quite the opposite - it should reduce them by eliminating signature conflicts from the start.

Are these conflicts significant enough for the new abstraction?  I’m concerned that adding complexity without clear use cases could make it harder to track changes/debug across multiple levels of abstraction.

rolfbjarne:
> > This is learning that was taken from a similar change in macios by @rolfbjarne to avoid similar problems in terms of signature conflicts.
> 
> Regarding the Objective-C example, how frequently was it used externally? Do we have any ballpark estimates? From my perspective, it seems to be mostly used internally.

It makes it possible for us to create better bindings for Objective-C, because we have more options for method signatures, so it's used/visible externally.

Example: `NSMutableArray` has an [init method that takes a `capacity` parameter](https://developer.apple.com/documentation/foundation/nsmutablearray/1415811-initwithcapacity?language=objc), specifying the initial size of the array.

The obvious way to bind this in C# is as an `NSMutableArray (nint)` constructor, but if we hadn't created `ObjCRuntime.NativeHandle`, that constructor would already exist as `NSMutableArray (IntPtr)`, because we need it to create a managed `NSMutableArray` instance from an existing native pointer.

Note that Apple uses native-sized integers pretty much exclusively for numbers in the API (as opposed to fixed 32-bit or 64-bit sized integers), so `nint` is used a lot to represent numbers.

> > I don't see how it increases maintenance costs - quite the opposite - it should reduce them by eliminating signature conflicts from the start.
> 
> Are these conflicts significant enough for the new abstraction? I’m concerned that adding complexity without clear use cases could make it harder to track changes/debug across multiple levels of abstraction.

After three years of having `ObjCRuntime.NativeHandle`, I'm happy we added it. While it was possible to work around this in the past, the workarounds felt clunky (and also added to the maintenance burden).

kotlarmilos (approved):
LGTM!

jkotas:
> Can you explain to me under what circumstances that would happen? It strikes me that that would be a fundamental change in Swift calling conventions and Swift structure packing, something that is already [defined by Apple](https://github.com/swiftlang/swift/blob/main/docs/ABI/TypeLayout.rst#fragile-struct-and-tuple-layout). If that changed, wouldn't it change pretty much all aspects of calling conventions and we would have to deal with that?

Calling conventions have this problem in general. Here is an x86 example that shows that int is passed differently than a struct that wraps int: https://godbolt.org/z/Yqfve917d

Looks like the Swift calling convention has built-in assumption that the item is passed the same way as a struct that wraps an item, so we should be ok.

PR #2795, opened 2024-11-20
[NativeAOT-LLVM] Remove PS download on Linux

SingleAccretion:
It appears PS 7 is already present in the docker image we use for Linux.


yowl (approved):
Thanks!

PR #2796, opened 2024-11-21
[SwiftBindings] Describe projecting generic structs

jkurdek:
Follow up to #2771. Proposes projections of generic structs.

### DIFF
--- a/src/docs/binding-generics.md
+++ b/src/docs/binding-generics.md
@@ -45,7 +45,7 @@ public static extern void PInvoke_ReturnData(SwiftIndirectResult result, NativeH
 
 public static unsafe T ReturnData<T>(T data) where T : unmanaged, ISwiftObject
 {
-    TypeMetadata.TryGetTypeMetadata<T>(out var metadata);
+    var metadata = GetTypeMetadataOrThrow<T>();
     nuint payloadSize = /* Extract size from metadata */;
 
     var payload = NativeMemory.Alloc(payloadSize);
@@ -81,7 +81,7 @@ public static extern void PInvoke_ReturnData(SwiftIndirectResult result, NativeH
 
 public static unsafe T ReturnData<T>(T data)
 {
-    TypeMetadata.TryGetTypeMetadata<T>(out var metadata);
+    var metadata = GetTypeMetadataOrThrow<T>();
     nuint payloadSize = /* Extract size from metadata */;
     NativeHandle payload = new NativeHandle(NativeMemory.Alloc(payloadSize));
 
@@ -91,9 +91,9 @@ public static unsafe T ReturnData<T>(T data)
         
         NativeHandle dataPayload = Runtime.GetPayload(ref data);
         PInvoke_ReturnData(result, dataPayload, metadata!.Value);
-        return Runtime.FromPayload<T>(payload);
+        return Runtime.FromPayload<T>(payload); // Transfers ownership of the payload
     }
-    finally
+    catch
     {
         NativeMemory.Free(payload);
     }
@@ -134,6 +134,16 @@ public static bool TryGetTypeMetadata<T>([NotNullWhen(true)] out TypeMetadata? r
         result = null;
         return false;
     }
+}
+
+public static TypeMetadata GetTypeMetadataOrThrow<T>()
+{
+    if (TryGetTypeMetadata<T>(out var result))
+    {
+        return result!;
+    }
+
+    throw new InvalidOperationException($"Could not obtain TypeMetadata for {typeof(T)}.");
 }
 
 public static NativeHandle GetPayload<T>(ref T type)
@@ -203,8 +213,7 @@ class NonFrozenStruct : ISwiftObject
 
     private unsafe NonFrozenStruct(NativeHandle payload) 
     { 
-        _payload = new NativeHandle(NativeMemory.Alloc(PayloadSize));
-        NativeMemory.Copy(payload, _payload, PayloadSize);
+        _payload = payload;
     }
 }
 
@@ -222,7 +231,9 @@ struct FrozenStruct : ISwiftObject
 
     public static unsafe object FromPayload(NativeHandle payload)
     {
-        return *(FrozenStruct*)payload;
+        var struct =  *(FrozenStruct*)payload;
+        NativeMemory.Free(payload);
+        return struct;
     }
 }
 ```
@@ -282,8 +293,8 @@ class Pair<T, U> : ISwiftObject
 
         SwiftIndirectResult swiftIndirectResult = new SwiftIndirectResult(_payload);
 
-        TypeMetadata.TryGetTypeMetadata<T>(out var firstMetadata);
-        TypeMetadata.TryGetTypeMetadata<U>(out var secondMetadata);
+        var firstMetadata = GetTypeMetadataOrThrow<T>();
+        var secondMetadata = GetTypeMetadataOrThrow<U>();
 
         var nativeHandleFirst = Runtime.GetPayload(ref first);
         var nativeHandleSecond = Runtime.GetPayload(ref second);
@@ -296,8 +307,8 @@ class Pair<T, U> : ISwiftObject
         /* This should be cached */
         get
         {
-            TypeMetadata.TryGetTypeMetadata<T>(out var firstMetadata);
-            TypeMetadata.TryGetTypeMetadata<U>(out var secondMetadata);
+            var firstMetadata = GetTypeMetadataOrThrow<T>();
+            var secondMetadata = GetTypeMetadataOrThrow<U>();
 
             return PairPInvokes.PInvokeMetadata(TypeMetadataRequest.Complete, firstMetadata!.Value, secondMetadata!.Value);
         }
@@ -318,8 +329,7 @@ class Pair<T, U> : ISwiftObject
 
     private unsafe Pair(NativeHandle payload)
     {
-        _payload = new NativeHandle(NativeMemory.Alloc(PayloadSize));
-        NativeMemory.Copy(payload, _payload, PayloadSize);
+        _payload = payload;
     }
 }
 ```
--- a/src/docs/binding-generics.md
+++ b/src/docs/binding-generics.md
@@ -3,7 +3,7 @@
 This document explores projections of Swift generics into C#.
 
 1. [Generic functions](#generic-functions)
-2. Generic types (TBD)
+2. [Generic types](#generic-types)
 3. Generics with protocol constraints (TBD)
 4. Generics with PAT constraints (TBD)
 5. Existential container for a protocol as a generic type (TBD)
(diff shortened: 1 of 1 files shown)


src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+
+        var nativeHandleFirst = Runtime.GetPayload(ref first);
+        var nativeHandleSecond = Runtime.GetPayload(ref second);
+
+        PairPInvokes.Pair(swiftIndirectResult, nativeHandleFirst, nativeHandleSecond, firstMetadata!.Value, secondMetadata!.Value);
+    }
+
+    public static TypeMetadata Metadata
+    {
+        /* This should be cached */
```
stephen-hawley, line 307: Agreed. Let's open an issue to design the caching mechanism.
Starting point:
```csharp
public static class TypeMetadataCache {
    TypeMetadata MetadataOf(Type t, Func<Type, TypeMetadata> generator) {
          // make thread safe:
          if (cache.TryGet(t, out var result)) return result;
          var md = generator(t);
          cache.Add(t, md);
          return md;
}
```
Then you can rewrite this as:
```
static TypeMetadata typeMetadata = TypeMetadataCache.MetadataOf(typeof(Pair<T, U>), (t) => {
    var typeMetadataT = TypeMetadataCache.MetadataOf (typeof (T), (tT) => TypeMetadata.GetTypeMetadataOrThrow(tT));
    var typeMetadataU = TypeMetadataCache.MetadataOf (typeof(U), (tU) => TypeMetadata.GetTypeMetadataOrThrow(tU));
    return PairPInvokes.PInvokeMetadata (typeMetadataRequest.Complete, tT, TU);
});
public static TypeMetadata Metadata => typeMetadata;
```

Again, this is a start. The notion is that there is a cache which has a dictionary of Type onto TypeMetadata and takes a type to look up and if it's not present, takes a generator to call to retrieve that type.

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+        get
+        {
+            TypeMetadata.TryGetTypeMetadata<T>(out var firstMetadata);
+            TypeMetadata.TryGetTypeMetadata<U>(out var secondMetadata);
```
stephen-hawley: Seeing that you didn't check the return tells me that we should absolutely have a GetOrThrow accessor that will throw an exception on fail.

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+    public unsafe void ChangeFirst(T newFirst)
+    {
+        SwiftSelf self = new SwiftSelf(_payload);
+        PairPInvokes.ChangeFirst(Runtime.GetPayload(ref newFirst), Metadata, self);
+    }
+
+    private unsafe Pair(NativeHandle payload)
+    {
+        _payload = new NativeHandle(NativeMemory.Alloc(PayloadSize));
+        NativeMemory.Copy(payload, _payload, PayloadSize);
```
stephen-hawley: This should use the `InitializeWithCopy` from the value witness table or else you aren't abiding by swift's memory model.

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+#### Frozen and non-frozen generic structs
+
+The frozen keyword can be applied to a generic struct just like it can be applied to a regular struct. However because the generic parameters type (and hence size and layout) is unknown at the compile time, so is the size and the layout of the generic struct. Specifically, the generic parameter can be a frozen or a non-frozen struct.
+
+If we consider the four cases:
+
+1. Non-frozen struct and at least one non-frozen generic parameter
+2. Non-frozen struct and (recursively) frozen generic parameters
+3. Frozen struct and at least one non-frozen generic parameter
+4. Frozen struct and (recursively) frozen generic parameters
```
stephen-hawley, line 346: Like the way you're breaking this out into the possible cases.

stephen-hawley (approved):
Minor adjustments to make and notes

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+    {
+        _payload = new NativeHandle(NativeMemory.Alloc(PayloadSize));
+        NativeMemory.Copy(payload, _payload, PayloadSize);
+    }
+}
+```
+
+#### Frozen and non-frozen generic structs
+
+The frozen keyword can be applied to a generic struct just like it can be applied to a regular struct. However because the generic parameters type (and hence size and layout) is unknown at the compile time, so is the size and the layout of the generic struct. Specifically, the generic parameter can be a frozen or a non-frozen struct.
```
kotlarmilos, line 339: Does this mean that the number of fields is fixed and won't change?

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+```
+
+Which would then ideally call the following PInvoke.
+
+```csharp
+    [UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+    public static extern Pair<FrozenStruct, FrozenStruct> WithTwoFrozenParameters(Pair<FrozenStruct, FrozenStruct> pair);
+```
+
+This however would not work as `Pair<FrozenStruct, FrozenStruct>` is a class. Runtime would try to pass the value by reference (it does not even know the layout to begin with). We could solve this problem on the projection layer.
```
kotlarmilos, line 403: Let's treat representation and runtime lowering separately for a moment.

If a frozen struct doesn't require a copy constructor when moved in memory, I would still consider it a value type (struct). Additionally, as we discussed offline, the projection will flip if a generic used becomes frozen. From the user’s perspective, this can occur without their control, which is my primary concern when introducing this rule.

I understand that it doesn't match the lowering on the ABI level. However, I see this as implementation detail that belongs to the runtime. @amanasifkhalid I would like to hear your perspective.

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+Accounting for those two the projection could look something like:
+
+```csharp
+class Pair<T, U> : ISwiftObject
+{
+
+    private static nuint PayloadSize =  /* Extract size */;
+    private NativeHandle _payload;
+
+    public unsafe Pair(T first, U second)
```
matouskozak, line 290: My lack of knowledge probably but what is making this `unsafe`?

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+Which would then ideally call the following PInvoke.
+
+```csharp
+    [UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+    public static extern Pair<FrozenStruct, FrozenStruct> WithTwoFrozenParameters(Pair<FrozenStruct, FrozenStruct> pair);
+```
+
+This however would not work as `Pair<FrozenStruct, FrozenStruct>` is a class. Runtime would try to pass the value by reference (it does not even know the layout to begin with). We could solve this problem on the projection layer.
+
+We could generate a helper struct per instance of frozen closed generic type. This could look something like:
```
kotlarmilos, line 405: What is closed generic type? Is it constrained by a protocol?

matouskozak, line 405: What are the `frozen closed generic type`s?

matouskozak (approved):
Good job Jeremi!

src/docs/binding-generics.md
matouskozak: I think the problem is that user-facing C# `Pair` is a class. In the specific case of `Pair<FrozenStruct, FrozenStruct>`, it could be a struct (and that's the wrapper introduced later) but we would need to have 2x C# `Pair` (class and a struct) to handle the case of `NonFrozenStruct` as well. @jkurdek please correct me if I'm wrong here.

jkurdek: >If a frozen struct doesn't require a copy constructor when moved in memory, I would still consider it a value type (struct). 

So for generics strictly. I think we cannot tell at compile time whether a generic will have or wont have to use a copy constructor as we have no idea what the generic parameter will be. Some instances of the generic, `Pair<SomethingThatRequiresCopyConstructor, int>` will require it. Some like `Pair< FrozenStruct, FrozenStruct>` will not. 

@matouskozak I think you got it right. `Pair<FrozenStruct, FrozenStruct>` would ideally be represented as struct. 

>Additionally, as we discussed offline, the projection will flip if a generic used becomes frozen. From the user’s perspective, this can occur without their control, which is my primary concern when introducing this rule.

 Making something non-frozen frozen is an ABI breaking change. I do not think we can do much about it. One way of solving this would be representing every struct as a class on C# side and abstracting the implementation details

src/docs/binding-generics.md
jkurdek: I believe so. Both number of fields and their order

src/docs/binding-generics.md
jkurdek: Assuming we have generic `Foo<T>`, by closed generic type I meant `Foo<ConcreteType>` what would be a better way to phrase?

jkurdek: https://github.com/dotnet/runtimelab/pull/2796#discussion_r1852545797 with this difference that both the generic and the concreteType are marked with `@frozen`

src/docs/binding-generics.md
jkurdek: `NativeMemory.Alloc`

src/docs/binding-generics.md
matouskozak: Is the `@frozen` generic struct with non-frozen parameters a supported (or recommended) scenario in Swift? To me it seems like `@frozen` on the generic struct wouldn't help much with ABI stability then...

jkurdek: I cannot speak about whether it is recommended. It is definitely supported. There are some frozen generics in `SwiftUI`. I do not think it is possible to enforce using frozen parameters (maybe I am wrong), so it only up to the user of such generic struct to be kind enough to instantiate it with frozen types.

rolfbjarne (approved):

jkurdek:
While working on those projections I have encountered interesting combinations of frozen and non-frozen structs. I think they might be relevant to this discussion https://github.com/dotnet/runtimelab/issues/2797

src/docs/binding-generics.md
matouskozak: You're right, more info here https://github.com/swiftlang/swift-evolution/blob/main/proposals/0260-library-evolution.md

> It does not necessarily have a known size or alignment. A generic frozen struct's layout might depend on the generic argument provided at run time.

> Even concrete instantiations may not have a known size or alignment. A frozen struct with a field that's a *non-*frozen, has a size that may not be known until run time.

src/docs/binding-generics.md
kotlarmilos: > One way of solving this would be representing every struct as a class on C# side and abstracting the implementation details

Do you mean every generic struct? In this case, the size and alignment of an object of opaque layout, as well as whether it is trivial or bitwise movable, is determined by querying its value witness table. If contains frozen structs, the helper can be used.

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+
+```csharp
+[StructLayout(LayoutKind.Sequential, Size = 16)]
+public struct Pair_FrozenStruct_FrozenStruct
```
kotlarmilos, line 409: Why this is not part of the class?

jkurdek: I think it can be part of the class, probably should.

src/docs/binding-generics.md
```diff
@@ -236,3 +236,194 @@ The proposed API will only work with types known at compile time. In some cases
...
+    [UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+    public static extern Pair<FrozenStruct, FrozenStruct> WithTwoFrozenParameters(Pair<FrozenStruct, FrozenStruct> pair);
+```
+
+This however would not work as `Pair<FrozenStruct, FrozenStruct>` is a class. Runtime would try to pass the value by reference (it does not even know the layout to begin with). We could solve this problem on the projection layer.
+
+We could generate a helper struct per instance of frozen closed generic type. This could look something like:
+
+```csharp
+[StructLayout(LayoutKind.Sequential, Size = 16)]
```
jkurdek, line 408: I'm not sure whether the explicit size can be obtained at compile time for a generic instance, or it can be only done at runtime. We should be able to calculate it though in the worst scenario. 

I also wonder whether we need to even specify the explicit size at all (cc @amanasifkhalid @kotlarmilos )

src/docs/binding-generics.md
jkurdek: I meant all structs. Its more of stretch idea thing than actual proposal. I feel that projecting some as structs and some as classes creates a weird UX for developers who are not Swift professionals. I would prefer to have uniformity - e.g. all structs to be classes. I do understand however that this would require a lot custom runtime work and would bring its own set of problems. Anyway it is a different discussion.

src/docs/binding-generics.md
matouskozak: Thanks.

> We could generate a helper struct per instance of frozen closed generic type

Would that be a source generator or some other mechanism to generate the helpers during runtime? Or do we know all the `closed generic types` during compile time (compile time of bindings vs full app, this circles back to https://github.com/dotnet/runtimelab/pull/2779#issuecomment-2491470820)?

jkurdek: We should know all `closed generic types` which are part of the public API at compile time. They occur in e.g. `.swiftinterface`. That said we will need to gather that information as part of projection tooling.

src/docs/binding-generics.md
kotlarmilos: > I meant all structs. Its more of stretch idea thing than actual proposal. I feel that projecting some as structs and some as classes creates a weird UX for developers who are not Swift professionals.

It seems we need this layer of abstraction for structs with opaque layouts (those not known until runtime). In other cases, this abstraction isn’t necessary. We should prioritize simplicity, ensuring that users reading Swift docs know what to expect. If a struct can be projected directly, we should do so.

> I do understand however that this would require a lot custom runtime work and would bring its own set of problems. Anyway it is a different discussion.

Let's do cost estimate for this.

/cc: @jkoritzinsky

src/docs/binding-generics.md
kotlarmilos: @matouskozak Do we use the size during lowering?

matouskozak: Essentialy the lowering algorithm can be done just with the knowledge about the fields. 

Potentially, we could hit some issues where we assume the size is specified exactly but I don't think it would cause too much problems and we should be able to work around it. I would have to try it to verify.

kotlarmilos: > Potentially, we could hit some issues where we assume the size is specified exactly but I don't think it would cause too much problems and we should be able to work around it. I would have to try it to verify.

Do you have any example of that?

jkurdek: My understanding is that we would need to specify size if the Swift type had some extra padding which is not visible on the c# side, dont know whether this is something we should worry about

src/docs/binding-generics.md
jkurdek: Lets move this to the https://github.com/dotnet/runtimelab/issues/2797 issue

src/docs/binding-generics.md
matouskozak: I think particularly nested struct would cause a problem because of added padding to the inner struct on C# side. For example,

```
    [StructLayout(LayoutKind.Sequential, Size = 12)]
    struct F3_S0_S0
    {
        public nint F0;
        public uint F1;

        public F3_S0_S0(nint f0, uint f1)
        {
            F0 = f0;
            F1 = f1;
        }
    }

    [StructLayout(LayoutKind.Sequential, Size = 24)]
    struct F3_S0
    {
        public sbyte F0; // Offset 0
        public F3_S0_S0 F1; // Offset 8
        public uint F2; // Offset 20

        public F3_S0(sbyte f0, F3_S0_S0 f1, uint f2)
        {
            F0 = f0;
            F1 = f1;
            F2 = f2;
        }
    }
```

The `F3_S0` should get lowered as: i8, i64, i64.

Without specifying the size of `F3_S0_S0`, the `public uint F2` would be at offset 24 (because `F3_S0_S0` size would be padded to 16), causing the current implementation of lowering algorithm to create a different lowered sequence: i8, i64, i32, i32 (because of the padding between the `F1` field from `F3_S0_S0`  and `F2` from `F3_S0`, preventing merging the two `uint` fields into single i64). 

Note: I described the process based on Mono implementation, CoreCLR might differ, but I suspect the outcome to be similar.

While we should be able to work around this by using exact sizes of individual fields, that would still require us to have knowledge about (ALL) fields that are going to be present in the struct. I wonder if we would be able to define some kind of meta lowering for the generic struct that would get adjusted on the go based on different specializations.

matouskozak: However, I forgot that this discussion is only about `frozen closed generic type` (which are known at compile time https://github.com/dotnet/runtimelab/pull/2796#discussion_r1852712009), we should be able to get the sizes ahead of time, no? Or does the `closed generic type` refer only to the main struct and not recursively?

jkurdek: Frozen generic struct might contain other frozen generic structs as members. 
```swift
@frozen
public struct BigStruct<T>
{
    let x : T
    let y : SmallStruct<T>
    let z : SmallStruct<Float>
}

@frozen 
public struct SmallStruct<U>
{
    public let x: U
}
```
Also closed generic type might be used to close another generic like `BigStruct<SmallStruct<Int>>`

We are able to get the size of the non-generic frozen members ahead of time, we should also have access to the layout, so we should be able to recursively calculate the size ahead of time. But I do not think we can just extract the size of the the closed / bound frozen generic from the abi.json or dylib. Do you have any ideas @stephen-hawley?

src/docs/binding-generics.md
jkurdek: https://github.com/dotnet/runtimelab/issues/2807

src/docs/binding-generics.md
jkurdek: Thanks for pointing this out. I am wondering what approach should we take here:

1. We could make this constructor take an ownership over the NativeHandle - this should then eliminate the need to call the `InitializeWithCopy`. The caveat is that the caller would need to make sure that it does not use the NativeHandle after the constructor is called. Which should be ok as I presume we will only need the caller to forward a buffer from Swift for object creation.

2. Keep the current model and implement `InitializeWithCopy`. But then the caller would have to make sure to call deallocate on the no longer necessary NativeHandle.

I think I am leaning to the first solution. Let me know what you think

kotlarmilos: Does `InitializeWithCopy` perform the same operation as `NativeMemory.Copy`?

jkurdek: I think that `NativeMemory.Copy` might be problematic for `non-bitwise-movable` structs
https://github.com/dotnet/designs/blob/main/proposed/swift-interop.md#structsvalue-types

src/docs/binding-generics.md
kotlarmilos: Since we moved frozen type with frozen generic parameters to https://github.com/dotnet/runtimelab/issues/2797, let's resolve this comment. Other cases should be covered with the helper marshalling struct.

jkurdek: Let's make sure we see the same picture before resolving. How I view it is that the marshalling struct is brought specifically to solve the frozen type with frozen generic problem. non-frozen generic or non-frozen parameters will work out of the box. So in my eyes the #2797 issue doesn't really concern generics but rather struct / class projection of non-generic structs.

src/docs/binding-generics.md
kotlarmilos: > We are able to get the size of the non-generic frozen members ahead of time, we should also have access to the layout, so we should be able to recursively calculate the size ahead of time. But I do not think we can just extract the size of the the closed / bound frozen generic from the abi.json or dylib.

This information is available in type's metadata but we may not be able to call into it during binding time. Here are the ABI information for the example:
```json
{
"kind": "TypeDecl",
"name": "BigStruct",
"printedName": "BigStruct",
"children": [
  {
    "kind": "Var",
    "name": "x",
    "printedName": "x",
    "children": [
      {
        "kind": "TypeNominal",
        "name": "GenericTypeParam",
        "printedName": "τ_0_0"
      }
    ],
    "declKind": "Var",
    "usr": "s:14SwiftUIWrapper9BigStructV1xxvp",
    "mangledName": "$s14SwiftUIWrapper9BigStructV1xxvp",
    "moduleName": "SwiftUIWrapper",
    "isInternal": true,
    "declAttributes": [
      "HasStorage"
    ],
    "fixedbinaryorder": 0,
    "isLet": true,
    "hasStorage": true
  },
  {
    "kind": "Var",
    "name": "y",
    "printedName": "y",
    "children": [
      {
        "kind": "TypeNominal",
        "name": "SmallStruct",
        "printedName": "SwiftUIWrapper.SmallStruct<τ_0_0>",
        "children": [
          {
            "kind": "TypeNominal",
            "name": "GenericTypeParam",
            "printedName": "τ_0_0"
          }
        ],
        "usr": "s:14SwiftUIWrapper11SmallStructV"
      }
    ],
    "declKind": "Var",
    "usr": "s:14SwiftUIWrapper9BigStructV1yAA05SmallD0VyxGvp",
    "mangledName": "$s14SwiftUIWrapper9BigStructV1yAA05SmallD0VyxGvp",
    "moduleName": "SwiftUIWrapper",
    "isInternal": true,
    "declAttributes": [
      "HasStorage"
    ],
    "fixedbinaryorder": 1,
    "isLet": true,
    "hasStorage": true
  },
  {
    "kind": "Var",
    "name": "z",
    "printedName": "z",
    "children": [
      {
        "kind": "TypeNominal",
        "name": "SmallStruct",
        "printedName": "SwiftUIWrapper.SmallStruct<Swift.Float>",
        "children": [
          {
            "kind": "TypeNominal",
            "name": "Float",
            "printedName": "Swift.Float",
            "usr": "s:Sf"
          }
        ],
        "usr": "s:14SwiftUIWrapper11SmallStructV"
      }
    ],
    "declKind": "Var",
    "usr": "s:14SwiftUIWrapper9BigStructV1zAA05SmallD0VySfGvp",
    "mangledName": "$s14SwiftUIWrapper9BigStructV1zAA05SmallD0VySfGvp",
    "moduleName": "SwiftUIWrapper",
    "isInternal": true,
    "declAttributes": [
      "HasStorage"
    ],
    "fixedbinaryorder": 2,
    "isLet": true,
    "hasStorage": true
  }
],
"declKind": "Struct",
"usr": "s:14SwiftUIWrapper9BigStructV",
"mangledName": "$s14SwiftUIWrapper9BigStructV",
"moduleName": "SwiftUIWrapper",
"genericSig": "<τ_0_0>",
"sugared_genericSig": "<T>",
"declAttributes": [
  "AccessControl",
  "Frozen"
],
"conformances": [
  {
    "kind": "Conformance",
    "name": "Copyable",
    "printedName": "Copyable",
    "usr": "s:s8CopyableP",
    "mangledName": "$ss8CopyableP"
  },
  {
    "kind": "Conformance",
    "name": "Escapable",
    "printedName": "Escapable",
    "usr": "s:s9EscapableP",
    "mangledName": "$ss9EscapableP"
  }
]
},
{
"kind": "TypeDecl",
"name": "SmallStruct",
"printedName": "SmallStruct",
"children": [
  {
    "kind": "Var",
    "name": "x",
    "printedName": "x",
    "children": [
      {
        "kind": "TypeNominal",
        "name": "GenericTypeParam",
        "printedName": "τ_0_0"
      }
    ],
    "declKind": "Var",
    "usr": "s:14SwiftUIWrapper11SmallStructV1xxvp",
    "mangledName": "$s14SwiftUIWrapper11SmallStructV1xxvp",
    "moduleName": "SwiftUIWrapper",
    "declAttributes": [
      "HasStorage",
      "AccessControl"
    ],
    "fixedbinaryorder": 0,
    "isLet": true,
    "hasStorage": true,
    "accessors": [
      {
        "kind": "Accessor",
        "name": "Get",
        "printedName": "Get()",
        "children": [
          {
            "kind": "TypeNominal",
            "name": "GenericTypeParam",
            "printedName": "τ_0_0"
          }
        ],
        "declKind": "Accessor",
        "usr": "s:14SwiftUIWrapper11SmallStructV1xxvg",
        "mangledName": "$s14SwiftUIWrapper11SmallStructV1xxvg",
        "moduleName": "SwiftUIWrapper",
        "genericSig": "<τ_0_0>",
        "sugared_genericSig": "<U>",
        "implicit": true,
        "declAttributes": [
          "Transparent"
        ],
        "accessorKind": "get"
      }
    ]
  }
],
"declKind": "Struct",
"usr": "s:14SwiftUIWrapper11SmallStructV",
"mangledName": "$s14SwiftUIWrapper11SmallStructV",
"moduleName": "SwiftUIWrapper",
"genericSig": "<τ_0_0>",
"sugared_genericSig": "<U>",
"declAttributes": [
  "AccessControl",
  "Frozen"
],
"conformances": [
  {
    "kind": "Conformance",
    "name": "Copyable",
    "printedName": "Copyable",
    "usr": "s:s8CopyableP",
    "mangledName": "$ss8CopyableP"
  },
  {
    "kind": "Conformance",
    "name": "Escapable",
    "printedName": "Escapable",
    "usr": "s:s9EscapableP",
    "mangledName": "$ss9EscapableP"
  }
]
}
```

Do we have all necessary information about the bound generics?

jkurdek: I think so. We know the layout of the generic from this.  For the `SmallStruct<U>`

```
"name": "SmallStruct",
"printedName": "SmallStruct",
"children": [
  {
    "kind": "Var",
    "name": "x",
    "printedName": "x",
    "children": [
      {
        "kind": "TypeNominal",
        "name": "GenericTypeParam",
        "printedName": "τ_0_0"
      }
    ],
...
```

We know that I has one member of type τ_0_0. So when we will see occurrence of e.g. `SmallStruct<int>` we could just substitute that τ_0_0 with int and get the correct size. (If we had other generic parameters they would have different "types" τ_0_0, τ_0_1 ...)

stephen-hawley: > Frozen generic struct might contain other frozen generic structs as members.
> 
> ```swift
> @frozen
> public struct BigStruct<T>
> {
>     let x : T
>     let y : SmallStruct<T>
>     let z : SmallStruct<Float>
> }
> 
> @frozen 
> public struct SmallStruct<U>
> {
>     public let x: U
> }
> ```
> 
> Also closed generic type might be used to close another generic like `BigStruct<SmallStruct<Int>>`
> 
> We are able to get the size of the non-generic frozen members ahead of time, we should also have access to the layout, so we should be able to recursively calculate the size ahead of time. But I do not think we can just extract the size of the the closed / bound frozen generic from the abi.json or dylib. Do you have any ideas @stephen-hawley?

I think that the way to approach this is to have a binding time type database that is as complete as possible. Given a database that has been well seeded with datatypes from libswiftCore.dylib, we should be able to calculate size/stride/alignment without having to dip into the dylib being bound.

kotlarmilos (approved):

PR #2798, opened 2024-11-21
[NativeAOT-LLVM] `install-llvm.ps1` improvements for Linux

SingleAccretion:
1\) Gate cross-compilation on the presence of ROOTFS_DIR
2\) Pass `--parallel` to cmake build to speed it up: this cuts the LLVM build time up to 3 times in CI
3\) A bit of refactoring for readability

I've also verified locally that setting up LLVM on Linux by simply invoking the script works as intended.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks!

Issue #2803, opened 2024-11-21
[Swift runtime features] Remove SwiftTypeInfo and related types


stephen-hawley:
Refactoring to use TypeMetadata at runtime.

We might need SwiftTypeInfo's capabilities in a compile-time safe form, so keeping it around for the time being.

Issue #2802, opened 2024-11-21
[SwiftBindings] TODO - address various SwiftTypeInfo issues in parser and emitter


stephen-hawley:
This happens with the introduction of TypeMetadata and the removal of SwiftTypeInfo.

For the most part, trying to use the type metadata from the dylib at build time is probably not safe and definitely won't work for generics.

Issue #2801, opened 2024-11-21
[SwiftBindings] Update TypeMetadataKind to include current extended values


stephen-hawley:
See [here](https://github.com/swiftlang/swift/blob/main/include/swift/ABI/MetadataKind.def) for current values.

Issue #2800, opened 2024-11-21
[SwiftBindings] Add value witness table support


stephen-hawley:
Create type for ValueWitnessTable
Add accessors to TypeMetadata to include:
A ValueWitnessTable accessor
Size
Stride
Alignment

stephen-hawley:
Fixed [here](https://github.com/dotnet/runtimelab/pull/2813).

PR #2804, opened 2024-11-21
[SwiftBindings] Added TypeMetadata, removed SwiftTypeInfo

stephen-hawley:
This is a very simple implementation of TypeMetadata to get @jkurdek unblocked.

It includes access to the metadata kind and not a lot else for now.

The goal of this is to be used as a runtime type not a compile time type.

For the time being I removed SwiftTypeInfo which uses info pulled directly from the data structures in the dylib. We should discuss this in more detail in terms of how we want to get this information.  As background, it's not a great idea to fully load a dylib at compile time since it can execute arbitrary code (see `__attribute__((constructor))`) just by being loaded and may have dependencies that we don't know about or may load dependencies that conflict with the binding tool itself.

TypeMetadata is designed to be used in any place that a Swift call needs a type metadata object, such as implicit arguments in a generic function or in an existential type. To get one, you should use this as the return type for a Metadata Accessor function or the appropriate Swift runtime call to generate/get them from a cache for tuples, functions etc.

Added issues to address:
https://github.com/dotnet/runtimelab/issues/2800
https://github.com/dotnet/runtimelab/issues/2801
https://github.com/dotnet/runtimelab/issues/2802
https://github.com/dotnet/runtimelab/issues/2803

### DIFF
--- a/src/Swift.Bindings/src/Parser/SwiftABIParser.cs
+++ b/src/Swift.Bindings/src/Parser/SwiftABIParser.cs
@@ -274,17 +274,17 @@ namespace BindingsGeneration
 
             TypeDecl? decl = null;
             TypeRecord typeRecord = _typeDatabase.Registrar.RegisterType(node.ModuleName, node.Name);
-            IntPtr metadataPtr;
 
             switch (node.DeclKind)
             {
                 case "Struct":
                 case "Enum":
-                    metadataPtr = DynamicLibraryLoader.invoke(_dylibPath, GetMetadataAccessor(node));
-                    var swiftTypeInfo = new SwiftTypeInfo { MetadataPtr = metadataPtr };
+                    // metadataPtr = DynamicLibraryLoader.invoke(_dylibPath, GetMetadataAccessor(node));
+                    // var swiftTypeInfo = new SwiftTypeInfo { MetadataPtr = metadataPtr };
+                    // TODO: fix this code to not use static metadata objects if possible
 
-                    if (node.DeclAttributes != null && Array.IndexOf(node.DeclAttributes, "Frozen") != -1 && 
-                        (!swiftTypeInfo.ValueWitnessTable->IsNonPOD || !swiftTypeInfo.ValueWitnessTable->IsNonBitwiseTakable))
+                    if (node.DeclAttributes != null && Array.IndexOf(node.DeclAttributes, "Frozen") != -1) // && 
+                        // (!swiftTypeInfo.ValueWitnessTable->IsNonPOD || !swiftTypeInfo.ValueWitnessTable->IsNonBitwiseTakable))
                     {
                         decl = CreateStructDecl(node, parentDecl, moduleDecl);
                     }
@@ -292,7 +292,7 @@ namespace BindingsGeneration
                     {
                         decl = CreateClassDecl(node, parentDecl, moduleDecl);
                     }
-                    typeRecord.SwiftTypeInfo = swiftTypeInfo;
+//                    typeRecord.SwiftTypeInfo = swiftTypeInfo;
                     break;
 
                 case "Class":
--- a/src/Swift.Bindings/src/Parser/SwiftABIParser.cs
+++ b/src/Swift.Bindings/src/Parser/SwiftABIParser.cs
@@ -274,17 +274,18 @@ namespace BindingsGeneration
 
             TypeDecl? decl = null;
             TypeRecord typeRecord = _typeDatabase.Registrar.RegisterType(node.ModuleName, node.Name);
+            IntPtr metadataPtr;
 
             switch (node.DeclKind)
             {
                 case "Struct":
                 case "Enum":
-                    // metadataPtr = DynamicLibraryLoader.invoke(_dylibPath, GetMetadataAccessor(node));
-                    // var swiftTypeInfo = new SwiftTypeInfo { MetadataPtr = metadataPtr };
                     // TODO: fix this code to not use static metadata objects if possible
+                    metadataPtr = DynamicLibraryLoader.invoke(_dylibPath, GetMetadataAccessor(node));
+                    var swiftTypeInfo = new SwiftTypeInfo { MetadataPtr = metadataPtr };
 
-                    if (node.DeclAttributes != null && Array.IndexOf(node.DeclAttributes, "Frozen") != -1) // && 
-                        // (!swiftTypeInfo.ValueWitnessTable->IsNonPOD || !swiftTypeInfo.ValueWitnessTable->IsNonBitwiseTakable))
+                    if (node.DeclAttributes != null && Array.IndexOf(node.DeclAttributes, "Frozen") != -1 && 
+                        (!swiftTypeInfo.ValueWitnessTable->IsNonPOD || !swiftTypeInfo.ValueWitnessTable->IsNonBitwiseTakable))
                     {
                         decl = CreateStructDecl(node, parentDecl, moduleDecl);
                     }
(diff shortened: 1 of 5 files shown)


src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -292,7 +293,7 @@ private List<BaseDecl> CollectDeclarations(IEnumerable<Node> nodes, BaseDecl par
                     {
                         decl = CreateClassDecl(node, parentDecl, moduleDecl);
                     }
-                    typeRecord.SwiftTypeInfo = swiftTypeInfo;
+                   typeRecord.SwiftTypeInfo = swiftTypeInfo;
```
rolfbjarne: This doesn't look intentional? Unindent by one space?

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -0,0 +1,199 @@
...
+using System.Collections.Generic;
+
+namespace Swift.Runtime;
+
+/// <summary>
+/// Flags used to describe types
+/// </summary>
+[Flags]
+public enum TypeMetadataFlags {
+    /// <summary>
```
rolfbjarne, line 14: Would it make sense to add a `None = 0` value, since this enum represents flags?

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// </summary>
+    MetadataKindIsNonType = 0x400,
+    /// <summary>
+    /// The metadata doesn't live on the heap
+    /// </summary>
+    MetadataKindIsNonHeap = 0x200,
+    /// <summary>
+    /// The type is private to the runtime
+    /// </summary>
+    MetadataKindIsRuntimePrivate = 0x100,
```
rolfbjarne: Is `MetadataKind` a common prefix for all values, or will there be more flags that aren't `MetadataKind` flags? If the former, the prefix should be removed:

```suggestion
    IsNonType = 0x400,
    /// <summary>
    /// The metadata doesn't live on the heap
    /// </summary>
    IsNonHeap = 0x200,
    /// <summary>
    /// The type is private to the runtime
    /// </summary>
    IsRuntimePrivate = 0x100,
```

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -0,0 +1,199 @@
...
+    const long kMaxDiscriminator = 0x7ff;
+
+    /// <summary>
+    /// Returns the kind of this TypeMetadata
+    /// </summary>
+    public TypeMetadataKind Kind {
+        get {
+            ThrowOnInvalid ();
+			long val = ReadPointerSizedInt (handle);
+			if (val == 0)
```
rolfbjarne: Looks like there's a mix of spaces and tabs here.

rolfbjarne (approved):

rolfbjarne:
> For the time being I removed SwiftTypeInfo which uses info pulled directly from the data structures in the dylib. We should discuss this in more detail in terms of how we want to get this information. As background, it's not a great idea to fully load a dylib at compile time since it can execute arbitrary code (see `__attribute__((constructor))`) just by being loaded and may have dependencies that we don't know about or may load dependencies that conflict with the binding tool itself.

Also you can't load dylibs for other than the current platform (so you won't be able to load dylibs for iOS on macOS for instance).

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// The metadata represents a class
+    /// </summary>
+    Class = 0x800 
+}
+
+
+/// <summary>
+/// Represents the type metadata for a Swift type
+/// </summary>
+public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
```
kotlarmilos, line 104: How this correlates with `SwiftMetadata`? Do we plan to use metadata for type semantics on the C# side or only as an unmanaged pointer for callconv?

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -0,0 +1,199 @@
...
+    readonly IntPtr handle;
```
jkurdek, line 105: Will this become `SwiftHandle` in the future?

jkurdek (commented):
Thanks for bringing that in! My only concern at the moment is that we have now both `TypeMetadata` and `SwiftMetadata` which duplicate some code and as far I can understand duplicate some functionality. I think we should discuss the boundaries of responsibility / merge those two

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
stephen-hawley: SwiftMetadata is based off the same type, but the difference here is that I'm not trying to define it as a ref struct but an struct with an opaque pointer so that it's easy to manipulate, compare, hash, and pass to and from p/invokes (the most common operations by far) and because it is going to escape the stack.
SwiftMetadata as is feels like a binding-time type not a runtime type.
I learned early on that we can't rely on it at binding time because of generics. It appears the only thing it gets used for at binding time is putting in a `Size` value in the `StructLayout` attribute.
For this I have several questions: 1 - do we need that? 2 - if we do, can't we get it in some other way (for example, follow the steps in the doc'n [here](https://github.com/swiftlang/swift/blob/main/docs/ABI/TypeLayout.rst#fragile-struct-and-tuple-layout).
In the larger picture, we need a binding-time type database, which is it's own thing, but certainly one of the things we can put into in optional properties for Size, Stride, and Alignment so that we can calculate a new Struct's size.  But like I said, that's a larger discussion.

jkurdek: > 1 - do we need that?

I think this might be relevant https://github.com/dotnet/runtimelab/pull/2796#discussion_r1852658458. tldr: we need to check

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
stephen-hawley: That's a great question - since this is a private member and we should never (1) need to access this outside of the TypeMetadata (and I don't think we ever would) and (2) get this type from any other place that as an implicit argument passed to us from Swift or as a return from an explicit call, it's fine to leave it as an IntPtr.

stephen-hawley:
> Thanks for bringing that in! My only concern at the moment is that we have now both `TypeMetadata` and `SwiftMetadata` which duplicate some code and as far I can understand duplicate some functionality. I think we should discuss the boundaries of responsibility / merge those two

My inclination is that the binding time functionality of `SwiftMetadata` should exist in a different form and that `TypeMetadata` should be kept as bare-bones as possible.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
stephen-hawley: It's not really relevant since we can't know unbound generic type sizes at compile time ever. Having the Size at compile time is only beneficial for frozen structs and frozen enums.

jkurdek: Did not mean generics, but having an explicit size value in the `StructLayout` attribute for frozen structs. @matouskozak shared a great example https://github.com/dotnet/runtimelab/pull/2796#discussion_r1854188388

PR #2805, opened 2024-11-22
[RuntimeAsync] Add support for async struct instance methods (potential implementation 2)

jakobbotsch:
This adds support for async2 struct instance calls with the semantics that any such method makes a copy of "this" and operates on it instead of operating through the this-pointer.


jakobbotsch:
cc @333fred @agocke @VSadov -- in this one runtime-async struct methods make an implicit copy of this whenever they are entered, and thus mutations do not reflect back to caller. This should match the current async1 semantics of Roslyn without Roslyn introducing any explicit copies.

We still need some restrictions in the spec around mutating the "this byref" -- i.e. we probably just spec that `starg 0` and `ldarga 0` are illegal in async2 struct methods. But there are no restrictions on how struct async2 instance calls can be made.

src/coreclr/jit/importer.cpp
```diff
@@ -6750,6 +6750,11 @@ void Compiler::impImportBlockCode(BasicBlock* block)
 
                 if (compIsForInlining())
                 {
+                    if ((lclNum == 0) && compIsStructMethodThatOperatesOnCopy())
+                    {
+                        BADCODE("Illegal starg 0 in function");
```
VSadov, line 6755: Right. `this` arg in a struct method, which is a reference to the instance, should be treated as readonly in async methods.

VSadov: It is a bit surprising that arg0 it is not readonly in a struct instance method in general. Languages like C# would not need that.

However one possible use scenario that comes to mind is implementing similar remapping of `this` as we do here, but as an IL rewrite.

Smth like:

// main entry point
```
   ldloca <temp>   // get addr of the local copy
   dup                     
   ldarg.0                // 
   cpobj                  // <temp> = this
   starg.0                // &this = &<temp>
. . . existing code
```

// resume entry points
```
   ldloca <temp>   // get addr of the local copy
   starg.0                // &this = &<temp>
. . . existing code
```

jakobbotsch: I believe F# also makes use of this capability when turning tailcalls into loops.

VSadov (approved):
LGTM. Thanks!

jakobbotsch:
Let's merge it in this way to at least have the support. We can consider whether we'll want to make some changes or have some other semantics separately.

PR #2811, opened 2024-11-26
[SwiftBindings] Describe protocol constraints

jkurdek:
This PR proposes projections for generic types with protocol constraints.

### DIFF
--- a/src/docs/binding-generics.md
+++ b/src/docs/binding-generics.md
@@ -511,8 +511,7 @@ struct FrozenStruct : ISwiftObject, IPrintableProtocol
 Adding constraints on a type's generic parameters is also straightforward, with small adjustments necessary to the initializer and metadata accessor functions. Let's revisit the `Pair` example:
 
 ```swift
-@frozen 
-public struct PairNonFrozen<T: Printable, U: Printable> { 
+public struct Pair<T: Printable, U: Printable> { 
     public var first: T 
     public var second: U
  
--- a/src/docs/binding-generics.md
+++ b/src/docs/binding-generics.md
@@ -4,7 +4,7 @@ This document explores projections of Swift generics into C#.
 
 1. [Generic functions](#generic-functions)
 2. [Generic types](#generic-types)
-3. Generics with protocol constraints (TBD)
+3. [Generics with protocol constraints](#protocol-constraints)
 4. Generics with PAT constraints (TBD)
 5. Existential container for a protocol as a generic type (TBD)
 
@@ -90,7 +90,7 @@ public static unsafe T ReturnData<T>(T data)
         SwiftIndirectResult result = new SwiftIndirectResult(payload);
         
         NativeHandle dataPayload = Runtime.GetPayload(ref data);
-        PInvoke_ReturnData(result, dataPayload, metadata!.Value);
+        PInvoke_ReturnData(result, dataPayload, metadata);
         return Runtime.FromPayload<T>(payload); // Transfers ownership of the payload
     }
     catch
@@ -299,7 +299,7 @@ class Pair<T, U> : ISwiftObject
         var nativeHandleFirst = Runtime.GetPayload(ref first);
         var nativeHandleSecond = Runtime.GetPayload(ref second);
 
-        PairPInvokes.Pair(swiftIndirectResult, nativeHandleFirst, nativeHandleSecond, firstMetadata!.Value, secondMetadata!.Value);
+        PairPInvokes.Pair(swiftIndirectResult, nativeHandleFirst, nativeHandleSecond, firstMetadata, secondMetadata);
     }
 
     public static TypeMetadata Metadata
@@ -310,7 +310,7 @@ class Pair<T, U> : ISwiftObject
             var firstMetadata = GetTypeMetadataOrThrow<T>();
             var secondMetadata = GetTypeMetadataOrThrow<U>();
 
-            return PairPInvokes.PInvokeMetadata(TypeMetadataRequest.Complete, firstMetadata!.Value, secondMetadata!.Value);
+            return PairPInvokes.PInvokeMetadata(TypeMetadataRequest.Complete, firstMetadata, secondMetadata);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+ func printMe()
+}
+
+public func printPrintable<T: Printable>(data: T) {
+ data.printMe()
+}
+
+// Equivalent to: public func printPrintable<T>(data: T) where T : Printable
+```
+
```
stephen-hawley, line 462: I would add this:

When a function is declared with a generic argument, an extra implicit argument is added to the end representing the type metadata of the generic argument (see above). If that generic argument is constrained by one or more protocols, the type metadata is followed by a pointer to the protocol witness table for each protocol. In the case of multiple protocols, e.g. `T: P3 & P1 & P2`, the protocol witness tables are ordered lexically by fully qualified name of the protocol.

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+This can be projected into C\# as follows:
+
+```csharp
+public interface ISwiftPrintableProtocol
+{
+    public void PrintMe();
+    public abstract static NativeHandle GetISwiftPrintableProtocolWitnessTable();
```
stephen-hawley: To avoid conflict with macios, we've named this `SwiftHandle` instead of `NativeHandle`.

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+    var protocolWitnessTable = T.GetISwiftPrintableProtocolWitnessTable();
+    PInvokePrintPrintable(data.GetPayload(), metadata, protocolWitnessTable);
+}
+
+[DllImport(Path, EntryPoint = "...")]
+[UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+private static extern void PInvokePrintPrintable(NativeHandle data, TypeMetadata metadata, NativeHandle protocolWitnessTable);
+```
+
+The only ABI difference is that some Swift function will expect to be passed Protocol Witness Table for the generic types at the end of the argument list.
```
stephen-hawley: See above - it's slightly more complicated. IIRC, if you have this:
```swift
public func foo<T: P1, U:P2>(a: T, b:U) { }
```
Then then the arguments in order are:
1. pointer to a
2. pointer to b
3. type metadata for T
4. protocol witness table for P1
5. type metadata for U
6. protocol witness table for P2

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+ data.printMe()
+}
+
+// Equivalent to: public func printPrintable<T>(data: T) where T : Printable
+```
+
+This can be projected into C\# as follows:
+
+```csharp
+public interface ISwiftPrintableProtocol
```
stephen-hawley: Naming nit: if the protocol is named `Printable` in Swift, we can named it `IPrintable` in C#. We'd like the resulting binding to line up as closely to the original as possible while still feeling like a real C# type.

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+{
+    // Other members
+
+    public static NativeHandle GetISwiftPrintableProtocolWitnessTable()
+    {
+        // This method extracts the witness table symbol from the dynamic library and should be cached.
+        return GetWitnessValueTable("...");
+    }
+
+    public unsafe void PrintMe()
```
stephen-hawley: This doesn't need to be `unsafe`

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+}
+
+[DllImport(Path, EntryPoint = "...")]
+[UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+private static extern void PInvokeAcceptHashable(NativeHandle data, TypeMetadata metadata, NativeHandle protocolWitnessTable);
+```
+
+We would be able to call `AcceptHashable<T>` using a type that implements both `ISwiftObject` and `ISwiftHashable`, but we would not be able to call it using, for example, `System.Int64`.
+
+One way to solve this problem is by creating a wrapper struct for each of the types, which would implement the necessary interfaces:
```
stephen-hawley, line 658: I like this approach. It's a good way to fake extension protocols.

stephen-hawley (changes requested):

src/docs/binding-generics.md
jkurdek: I think we currently use `NativeHandle` on many docs in this repo. I will be happy to open a PR which changes them all to `SwiftHandle` as a follow up.

src/docs/binding-generics.md
jkurdek: I think the order is :
1. pointer to a
2. pointer to b
3. type metadata for T
4. type metadata for U

Start of the pwt in lexicographical order:

5. protocol witness table for P1
6. protocol witness table for P2

This should align with the https://github.com/dotnet/runtimelab/pull/2810

stephen-hawley: Yes, this is correct - I just double checked it.

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+```csharp
+[StructLayout(LayoutKind.Sequential, Size = /* Extract size */)]
+struct FrozenStruct : ISwiftObject, IPrintableProtocol
+{
+    // Other members
+
+    public static NativeHandle GetIPrintableProtocolWitnessTable()
+    {
+        // This method extracts the witness table symbol from the dynamic library and should be cached.
+        return GetWitnessValueTable("...");
```
kotlarmilos: To retrieve protocol witness table, protocol conformance descriptor is needed. Look at https://github.com/swiftlang/swift/blob/6221b29c6835442706fbb44b67b755d370a87d96/include/swift/Runtime/Metadata.h#L360-L370.

jkurdek: If we want to use this function then yes. However I don't really understand the benefit of it as we have to load the protocol conformance descriptor from dylib anyway, right?

For example for a struct `Foo` conforming to a protocol `Printable` I can see the following two symbols in dylib:
```
0000000000007ad0 S _$s8generics3FooVAA9PrintableAAMc
0000000000008088 S _$s8generics3FooVAA9PrintableAAWP
```

The top one is conformance descriptor the bottom is the protocol witness table. I assumed we are just going to load the `s8generics3FooVAA9PrintableAAWP` from dylib straight away.

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+    public abstract static NativeHandle GetIPrintableProtocolWitnessTable();
+
+}
+
+public static void PrintPrintable<T>(T data) where T : ISwiftObject, IPrintableProtocol
+{
+    var metadata = TypeMetadata.GetTypeMetadataOrThrow<T>();
+    var protocolWitnessTable = T.GetIPrintableProtocolWitnessTable();
+    PInvokePrintPrintable(data.GetPayload(), metadata, protocolWitnessTable);
+}
```
vitek-karas, line 480: Sort of related question - what type are we going to generate this method into? In theory it could go onto the interface itself, but then it's polluting the C# consumer view. If it goes to a different type then we need to be able to deduce that type from the interface. I know it's a sort of implementation detail, but I'm curious if we're thinking about sharing that type across multiple bindings as well.
For example let's say `IPrintableProtocol` is defined in a PrintBase framework. We project that to PrintBase.dll (and the nuget around it). Then when I'm in my app and I'm going to define a `PrintableDocument` struct/class and it implemented `IPrintableProtocol` in its implementation it will need access to this helper - would we call that helper from the `PrintBase.dll` or would we generate it again into the app? (And sorry, I'm stealing your design discussion again - for feel free to kick me out into a separate issue ;-))

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+```
+
+### Protocol constraints on types
+
+Adding constraints on a type's generic parameters is also straightforward, with small adjustments necessary to the initializer and metadata accessor functions. Let's revisit the `Pair` example:
+
+```swift
+@frozen 
+public struct PairNonFrozen<T: Printable, U: Printable> {
```
vitek-karas: For the sake of readability - could you add a comment why this is non-frozen. I know we've discussed that `@frozen` doesn't actually mean value-type. So it's possible a better name to use `PairRefType` or something like that, not sure if that's a thing in Swift world as a term though. In any case, I don't understand what makes this a non-frozen thing, is the fact that it has a field of a generic type (so that could be a value type or ref type)?
Would that also be the case if it didn't have fields of those types (and just used the generic parameter in code).

jkurdek: Sorry - this was a bad naming leaking from my experimentation code. Now it should be correct.

src/docs/binding-generics.md
kotlarmilos: Let's check if it is always available. It should be cached:
```llvm-ir
; Function Attrs: noinline nounwind memory(none)
define linkonce_odr hidden ptr @"$s7SwiftUI13_TaskModifierVAcA04ViewD0AAWl"() #8 {
entry:
  %0 = load ptr, ptr @"$s7SwiftUI13_TaskModifierVAcA04ViewD0AAWL", align 8
  %1 = icmp eq ptr %0, null
  br i1 %1, label %cacheIsNull, label %cont

cacheIsNull:                                      ; preds = %entry
  %2 = call swiftcc %swift.metadata_response @"$s7SwiftUI13_TaskModifierVMa"(i64 255) #19
  %3 = extractvalue %swift.metadata_response %2, 0
  %4 = extractvalue %swift.metadata_response %2, 1
  %5 = call ptr @swift_getWitnessTable(ptr @"$s7SwiftUI13_TaskModifierVAA04ViewD0AAMc", ptr %3, ptr undef) #9
  store atomic ptr %5, ptr @"$s7SwiftUI13_TaskModifierVAcA04ViewD0AAWL" release, align 8
  br label %cont

cont:                                             ; preds = %cacheIsNull, %entry
  %6 = phi ptr [ %0, %entry ], [ %5, %cacheIsNull ]
  ret ptr %6
}
```

src/docs/binding-generics.md
```diff
@@ -437,3 +437,191 @@ We could also consider solving this on the runtime level.
...
+[UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+private static extern void PInvokeAcceptHashable(NativeHandle data, TypeMetadata metadata, NativeHandle protocolWitnessTable);
+```
+
+We would be able to call `AcceptHashable<T>` using a type that implements both `ISwiftObject` and `IHashable`, but we would not be able to call it using, for example, `System.Int64`.
+
+One way to solve this problem is by creating a wrapper struct for each of the types, which would implement the necessary interfaces:
+
+```csharp
+public struct SwiftIntWrapper : ISwiftObject, IHashable, ...
```
vitek-karas, line 661: I assume this would return Swift metadata for `Int`, right? So technically we could have multiple C# types which all look like `Int` from Swift's point of view (there's nothing bad about it... just trying to understand).

src/docs/binding-generics.md
vitek-karas: I also like this explicit approach, less magic :-)

src/docs/binding-generics.md
jkurdek: > what type are we going to generate this method into? In theory it could go onto the interface itself, but then it's polluting the C# consumer view. If it goes to a different type then we need to be able to deduce that type from the interface. 

In this design I put the method on the interface itself which is the easiest option but I agree that it will pollute the interface. ~I am not sure how big of an issue is this pollution (any swift type we are going to project will have a public `Metadata` method coming from `ISwiftObject` anyway.~

If we wanted to avoid pollution though, we could revert to stripping this information from the type itself and reverting to some sort of mapping.

>For example let's say IPrintableProtocol is defined in a PrintBase framework. We project that to PrintBase.dll (and the nuget around it). Then when I'm in my app and I'm going to define a PrintableDocument struct/class and it implemented IPrintableProtocol in its implementation it will need access to this helper - would we call that helper from the PrintBase.dll or would we generate it again into the app?

~First, protocol witness table (PWT) is generated for each pair (type, protocol), so if you defined a new type which extends this protocol on the csharp side only, we would need to go and generate the appropriate `PrintableDocument` on the Swift side so that a new PWT is correctly generated. And then, as far as I undersatand, we will be forced to either add a new `Get<ProtocolName>ProtocolWitnessTable` on the user defined type or add a new entry (TypeName, ProtocolName) -> PWT to some map. (Also the same thing has to happen with getting the metadata)~

src/docs/binding-generics.md
jkurdek: Yes, you are getting this right.  As long as we abide to the calling convention and pass the metadata for Swift's `Int`, Swift should treat it as an int.

vitek-karas: Technically this makes another assumption - that the binary representation of value of a frozen struct with a single `Int` field is the same as of `Int` itself. Again, nothing wrong about it, but we should probably note it somewhere to make it explicit.

jkurdek: >Technically this makes another assumption - that the binary representation of value of a frozen struct with a single Int field is the same as of Int itself.

Yes, thats true. Will add this assumption to the doc.

src/docs/binding-generics.md
jkurdek: This is some interesting type you shared. It doesn't even seem to be public. Could you share more about it?

kotlarmilos: Let me try to collect use cases and I will ping you offline.

src/docs/binding-generics.md
jkurdek: I gave it more thought. Just to correct myself - we can allow user to implement the interface on the c# side only without a need to to generate the swift side, I think Steve has a proposal in `binding-protocols` doc for that. So when dealing with this scenario we wont need / have a PWT for such combination (there will be no Swift type). We will have some proxy which will decided whether a type is a Swift type and has a PWT (and has to call swift) or is a C# side only type and needs to be handled on C# side exclusively. 

With that in mind I think we should should **not** put `GetProtocolWitnessTable` on the interface itself as some of types implementing it will not have anything meaningful to return there. We could have a static function which does map lookup.

src/docs/binding-generics.md
jkurdek: >Technically this makes another assumption - that the binary representation of value of a frozen struct with a single Int field is the same as of Int itself.

So I thought about it some more and I think this relies on the swift lowering algorithm (Int is a frozen struct on the Swift side), and we already baked in that assumption in the runtime.

src/docs/binding-generics.md
vitek-karas: Not sure I fully understand - but I would prefer we avoid lookups unless necessary. If we know what needs to happen at build time (projection generation time), we should resolve it there.

As an aside, I'm curious what is the scenario where I declare an interface in C#, want it projected to Swift, but there's no such protocol in Swift. It doesn't feel like something we need right now, so maybe we can live without it...

src/docs/binding-generics.md
vitek-karas: > and we already baked in that assumption in the runtime.

Agreed - I don't think it's a problem, a comment would just help with readability.

On a related note: What if we need to do this for a ref type? Something which is a first class citizen in C# BCL, so we can't add interfaces to it, but it's a ref type. (Can't think of anything right now, so maybe it's not a problem).

src/docs/binding-generics.md
jkurdek: So it seems that the PWT symbol is not synthesised directly when dealing with PATs and when protocol is defined in a different module then the conformance.

kotlarmilos: Does this mean if PAT is declared in a framework, the bindings need to retrieve protocol witness table during runtime?

jkurdek: Not only PAT. Any protocol. I think we will have to use the conformance descriptor after all.

kotlarmilos: Library evolution mode: https://www.swift.org/blog/library-evolution/
> Finally, to cope with adding new protocol requirements, protocol witness tables require runtime instantiation. Instead of emitting a witness table in the client code directly, the compiler emits a symbolic description of the conformance. The instantiation process places the protocol requirements in the correct order and fills missing entries to point to their default implementation, to produce a well-formed witness table which can be passed off to a dispatch thunk.

src/docs/binding-generics.md
```diff
@@ -437,3 +437,190 @@ We could also consider solving this on the runtime level.
...
+
+    // Other members
+}
+```
+
+### Supporting generic constraints on types present in both C\# and Swift
+
+When projecting a Swift type into C\#, we can project all of the protocols it conforms to as interfaces and make the C\# equivalent implement those interfaces. This allows it to work with the generic constraints mechanism described above.
+
+However, types that exist in both languages require special handling—for example, primitive number types. In Swift, primitive number types conform to multiple protocols that their C\# counterparts do not.
```
matouskozak, line 627: Do I understand it correctly that plain Swift primitive types, e.g., Swift [Int](https://developer.apple.com/documentation/swift/int) conform, by default, to multiple protocols?

matouskozak (approved):
Very nice and easy-to-understand description.

src/docs/binding-generics.md
stephen-hawley: This is absolutely correct. Two things to get used to with Swift as a language is (1) they've done their best to hide that there even are primitive types. For the most part, all the integral types, floating point types, etc are just value types and you can do anything to them that you can do to any other value type. (2) extensions allow any type to implement nearly* any other protocol.

*nearly - in that some PATs can have constraints on them to make them incompatible with value types

stephen-hawley (approved):

src/docs/binding-generics.md
jkurdek: Summary of offline discussion:
- We should move the GetProtocolWitnessTable method to ISwiftType to avoid interface pollution
- supporting scenario in which the user defines a c# side only implementation of a protocol would require either a Swift side proxy (whose PWT would be passed around)  or synthesising a PWT for c# type and swift interface

I did think some time about the Swift proxy, but I don't have a clear picture of it yet - I think this is beyond the scope of this PR. I opened a tracking [issue](https://github.com/dotnet/runtimelab/issues/2856). I believe that currently we can make an assumption that generic arguments will be either swift types or some proxies (both of which should implement ISwiftType)

kotlarmilos (approved):
LGTM!

PR #2812, opened 2024-11-26
[RuntimeAsync] Keep `LoaderAllocator` objects alive from continuations

### LINKED ISSUE #2759: [RuntimeAsync] Fix GC holes for suspended runtime-async methods + collectible assemblies
There are currently two GC holes in the face of collectible assemblies:
* If a runtime-async method has its code placed in a collectible assembly then nothing keeps that assembly alive while it is suspended
* If a runtime-async method has a shared generic context then nothing reports its loader allocator after it is suspended

More context: https://github.com/dotnet/runtimelab/pull/2755#discussion_r1832882179

jakobbotsch:
Add logic to make sure we keep the exposed loader allocator object alive when a runtime async function suspends. This adds two new JIT helpers `CORINFO_HELP_ALLOC_CONTINUATION_CLASS` and
`CORINFO_HELP_ALLOC_CONTINUATION_METHOD` that store the `LoaderAllocator` associated with a class or method into the allocated continuation before returning it back to the JIT.

There are a few cases where the JIT has to call the new helpers:
1. Any function loaded into a collectible ALC must keep the `LoaderAllocator` associated with the method being compiled alive
2. A shared generic function taking a `MethodDesc*` as the generic context must keep the `LoaderAllocator` assocated with that `MethodDesc*` alive
3. A shared generic function taking a `MethodTable*` as the generic context must similarly keep that one alive.

Fix #2759


src/coreclr/jit/async.cpp
```diff
@@ -1100,6 +1098,53 @@ void Async2Transformation::Transform(
...
+    GenTree* classHandleArg = nullptr;
+    if (((m_comp->info.compMethodInfo->options & CORINFO_GENERICS_CTXT_FROM_METHODDESC) != 0) && life.IsLive(m_comp->info.compTypeCtxtArg))
+    {
+        methodHandleArg = m_comp->gtNewLclvNode(m_comp->info.compTypeCtxtArg, TYP_I_IMPL);
+    }
+    else if (((m_comp->info.compMethodInfo->options & CORINFO_GENERICS_CTXT_FROM_METHODTABLE) != 0) && life.IsLive(m_comp->info.compTypeCtxtArg))
+    {
+        // TODO: Do we also need to keep the method handle's loader allocator
+        // alive here in case the shared generic method is inside a collectible
+        // ALC?
```
jakobbotsch: @davidwrighton @VSadov Any idea? Is it sufficient to keep the `LoaderAllocator` associated with the `MethodTable*` generic context alive here? I see that `MethodDesc::GetLoaderAllocator` gets it via `GetLoaderModule()->GetLoaderAllocator()`, but `MethodDesc::GetLoaderModule()` seemingly may not always match `GetMethodTable()->GetLoaderModule()` based on this code:

https://github.com/dotnet/runtime/blob/0170d78578a04206542f32e4db3eaa88a5384973/src/coreclr/vm/method.cpp#L1479-L1492

jakobbotsch:
PTAL @VSadov @davidwrighton

src/coreclr/jit/async.cpp
jkotas: > Is it sufficient to keep the LoaderAllocator associated with the MethodTable* generic context alive here?

I do not think so. You can have `void MyMethod<T>()` instantiated as `void MyMethod<UnloadableType>`. The MethodTable won't know anything about the UnloadableType.

jkotas: Ignore my comment. I have missed that this path is for the MethodTable* generic context case.

It is sufficient to keep the LoaderAllocator associated with the MethodTable* generic context alive here.

src/coreclr/jit/async.cpp
```diff
@@ -1100,6 +1098,50 @@ void Async2Transformation::Transform(
     m_resumptionBBs.push_back(resumeBB);
 }
 
+GenTreeCall* Async2Transformation::CreateAllocContinuationCall(AsyncLiveness& life, GenTree* prevContinuation, unsigned gcRefsCount, unsigned dataSize)
+{
+    GenTree* gcRefsCountNode = m_comp->gtNewIconNode((ssize_t)gcRefsCount, TYP_I_IMPL);
+    GenTree* dataSizeNode = m_comp->gtNewIconNode((ssize_t)dataSize, TYP_I_IMPL);
+    // If VM requests that we report the method handle, or if we have a shared generic context method handle
+    // that is live here, then we need to call a different helper.
+    GenTree* methodHandleArg = nullptr;
```
VSadov, line 1107: If I read this correctly we will use an alternative continuation allocators unconditionally, if we have generic contexts and allocator will sort it out if there is an object to hold on to.  

I guess we cannot know statically if we will have loader allocator or not. Not a big deal, I think.

VSadov:
Should we have a test, at least for the `continuationsNeedMethodHandle` scenario? 

We are probably not ready to stress-test this for GC holes, but maybe for code coverage - to make sure the alternative code paths are functional.

VSadov (approved):
LGTM! Thanks!
If not too involved, perhaps add a test.

src/coreclr/jit/async.cpp
jakobbotsch: Yes, I think all shared generics need to go through these new  helpers. We cannot know if the generic context was instantiated with an unloadable type during JIT.

For the `continuationsNeedMethodHandle` case I assume the loader allocator will always be non-null (since it's only set when the ALC is collectible).

jakobbotsch:
> Should we have a test, at least for the `continuationsNeedMethodHandle` scenario?

The shared-generic test does exercise the two new helpers. I can add one that uses collectible ALCs to also exercise the collectible ALC one.

jakobbotsch:
I pushed a test and verified that it fails without the change (with AV from trying to resume in unloaded code) and succeeds with it.

PR #2813, opened 2024-11-26
[SwiftBindings] Added support for size, stride, alignment, and the value witness table

stephen-hawley:
Added an accessor for ValueWitnessTable and added Size, Stride and Alignment properties.
The definition of Size and Stride are unsigned machine word-sized ints, so I changed them to be `nuint`

This addresses issue [2800](https://github.com/dotnet/runtimelab/issues/2800).

### DIFF
--- a/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
+++ b/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
@@ -179,17 +179,17 @@ public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
     /// <summary>
     /// Returns the size of the Swift type in bytes
     /// </summary>
-    public unsafe nuint Size => (*this.ValueWitnessTable).Size;
+    public unsafe nuint Size => this.ValueWitnessTable->Size;
 
     /// <summary>
     /// Returns the stride of the Swift type in bytes
     /// </summary>
-    public unsafe nuint Stride => (*this.ValueWitnessTable).Stride;
-    
+    public unsafe nuint Stride => this.ValueWitnessTable->Stride;
+
     /// <summary>
     /// Returns the alignment of the Swift type
     /// </summary>
-    public unsafe int Alignment => (*this.ValueWitnessTable).Alignment;
+    public unsafe int Alignment => this.ValueWitnessTable->Alignment;
 
     /// <summary>
     /// Reads a pointer sized integer from the location supplied
--- a/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
+++ b/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
@@ -172,6 +172,26 @@ public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
     }
 
     /// <summary>
+    /// Returns a pointer to the value witness table for the given type
+    /// </summary>
+    public unsafe ValueWitnessTable *ValueWitnessTable => (ValueWitnessTable*)(*((IntPtr*)handle - 1));
+
+    /// <summary>
+    /// Returns the size of the Swift type in bytes
+    /// </summary>
+    public unsafe nuint Size => (*this.ValueWitnessTable).Size;
+
+    /// <summary>
+    /// Returns the stride of the Swift type in bytes
+    /// </summary>
+    public unsafe nuint Stride => (*this.ValueWitnessTable).Stride;
+    
+    /// <summary>
+    /// Returns the alignment of the Swift type
+    /// </summary>
+    public unsafe int Alignment => (*this.ValueWitnessTable).Alignment;
+
+    /// <summary>
     /// Reads a pointer sized integer from the location supplied
     /// </summary>
     /// <param name="p">a pointer to memory</param>
--- a/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
+++ b/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
@@ -174,7 +174,7 @@ public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
     /// <summary>
     /// Returns a pointer to the value witness table for the given type
     /// </summary>
-    public unsafe ValueWitnessTable *ValueWitnessTable => (ValueWitnessTable*)(*((IntPtr*)handle - 1));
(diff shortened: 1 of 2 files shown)


dalexsoto (approved):

jkurdek (approved):

haritha-mohan (approved):

kotlarmilos (commented):
Thanks Steve! Two questions:
 - Do you plan to remove the existing metadata and value witness table if you introduce the new code?
 - Do you plan to include the following functions? https://github.com/dotnet/runtimelab/blob/90d20269bba2ac2ed0f0e42f36ac290565f52c55/src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs#L25-L48 ?

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -171,6 +171,26 @@ public TypeMetadataKind Kind {
...
 
+    /// <summary>
+    /// Returns a pointer to the value witness table for the given type
+    /// </summary>
+    public unsafe ValueWitnessTable *ValueWitnessTable => (ValueWitnessTable*)(*((IntPtr*)handle - 1));
+
+    /// <summary>
+    /// Returns the size of the Swift type in bytes
+    /// </summary>
+    public unsafe nuint Size => (*this.ValueWitnessTable).Size;
```
jkotas:
```suggestion
    public unsafe nuint Size => this.ValueWitnessTable->Size;
```
`*this.ValueWitnessTable` will materialize a copy of the whole struct. The JIT may be able to optimize out the extra copy, but it is better to write the code without the extra copy so that it is performant by construction.

src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs
```diff
@@ -36,12 +36,10 @@ public ref struct ValueWitnessTable
...
-		private IntPtr _Size;
-		private IntPtr _Stride;
+		public nuint Size;
+		public nuint Stride;
 		public ValueWitnessFlags Flags;
 		public uint ExtraInhabitantCount;
-        public int Size => _Size.ToInt32();
-        public int Stride => _Stride.ToInt32();
 		public int Alignment => (int)((Flags & ValueWitnessFlags.AlignmentMask) + 1);
 		public bool IsNonPOD => Flags.HasFlag (ValueWitnessFlags.IsNonPOD);
```
jkotas:
```suggestion
		public bool IsNonPOD => (Flags & ValueWitnessFlags.IsNonPOD) != 0;
```
It is better to use simple bit test rather than `HasFlag`. `HasFlag` will get optimized most of the time to a bit test, but not all the time. For example, `HasFlag` won't get optimized in Tier0 that is bad for startup time.

src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs
```diff
@@ -36,12 +36,10 @@ public ref struct ValueWitnessTable
...
 		public bool IsNonBitwiseTakable => Flags.HasFlag (ValueWitnessFlags.IsNonBitwiseTakable);
```
jkotas:
```suggestion
		public bool IsNonBitwiseTakable => (Flags & ValueWitnessFlags.IsNonBitwiseTakable) != 0;
```

stephen-hawley:
> Thanks Steve! Two questions:
> 
> * Do you plan to remove the existing metadata and value witness table if you introduce the new code?
> * Do you plan to include the following functions? https://github.com/dotnet/runtimelab/blob/90d20269bba2ac2ed0f0e42f36ac290565f52c55/src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs#L25-L48
>    ?

We're keeping your value witness table - that's what I started with.
With regards to the functions, how would you like to see them surfaced? In BTfS I used `Marshal.GetDelegateForFunctionPointer` to do this, but this brings up the issue of allocation and memory ownership. We could make a shadow managed structure of the delegates of the VWT as a singleton that it is and cache them. What do you think?

jkotas:
> We could make a shadow managed structure of the delegates of the VWT as a singleton that it is and cache them. What do you think?

Can it just use function pointers? Or does it need to be wrapped in a delegate for some reason?

stephen-hawley:
> Can it just use function pointers? Or does it need to be wrapped in a delegate for some reason?

Sure - we could make those `delegate *unmanaged<...>` - they're well defined that way.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -171,6 +171,26 @@ public TypeMetadataKind Kind {
         }
     }
 
+    /// <summary>
+    /// Returns a pointer to the value witness table for the given type
+    /// </summary>
+    public unsafe ValueWitnessTable *ValueWitnessTable => (ValueWitnessTable*)(*((IntPtr*)handle - 1));
```
rolfbjarne: Maybe this should throw on `handle == IntPtr.Zero`?

The main annoyance I see is that if you inspect/expand `TypeMetadata.Zero` in a debugger, the process will most likely crash, because it's accessing invalid memory.

Throwing a managed exception instead would be much better, which the debugger knows how to handle.

src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs
```diff
@@ -26,25 +26,107 @@ public enum ValueWitnessFlags
...
+		/// * pointers to C types
+		/// * C function pointers
+		/// See https://github.com/swiftlang/swift/blob/6221b29c6835442706fbb44b67b755d370a87d96/docs/proposals/AttrC.rst#type-bridging
+		/// </summary>
+		public bool IsNonPOD => (Flags & ValueWitnessFlags.IsNonPOD) != 0;
+
+		/// <summary>
+		/// Returns true if the value can NOT be copied bitwise.
+		/// </summary>
+		public bool IsNonBitwiseTakable => (Flags &ValueWitnessFlags.IsNonBitwiseTakable) != 0;
```
rolfbjarne:
```suggestion
		public bool IsNonBitwiseTakable => (Flags & ValueWitnessFlags.IsNonBitwiseTakable) != 0;
```

PR #2816, opened 2024-11-29
[Swift bindings] Add Swift code generation doc

kotlarmilos:
## Description

This document:
 - Justifies the need for Swift wrappers and presents the tradeoffs
 - Identifies specific parts of the ABI that require wrappers and explains the rationale
 - Defines the priorities based on selected frameworks

Swift code shipping and trimming will be discussed separately.

### DIFF
--- a/src/docs/swift-code-generation.md
+++ b/src/docs/swift-code-generation.md
@@ -17,7 +17,7 @@ Discriminated unions are not supported in C#, and their memory layout in Swift i
 ## Async
 Priority: High (required for StoreKit2 and SwiftUI)
 
-Async context in Swift is unstable and require thin wrappers for synhronous invocation. The wrapper will convert async functions into synchronous calls by using a Task as an async worker and a DispatchSemaphore to block until the operation completes. This approach encapsulates the unstable Swift async context, exposing only the stable wrapper interface to .NET.
+Async context in Swift is unstable and require thin wrappers for synchronous invocation. The wrapper will convert async functions into synchronous calls by using a Task as an async worker and a DispatchSemaphore to block until the operation completes. This approach encapsulates the unstable Swift async context, exposing only the stable wrapper interface to .NET.
 
 Here is an example:
 ```swift
--- a/src/docs/swift-code-generation.md
+++ b/src/docs/swift-code-generation.md
@@ -38,7 +38,7 @@ public func loadProductsSync(using storeManager: StoreManager) -> [Product]? {
 ## Dynamic dispatch
 Priority: Low
 
-Vtable requirements can be reordered or expanded, making their layout unstable across resilience boundaries. To manage this, the binary framework includes a dispatch thunk which abstracts vtable offsets. The dispatch thunk is part of the framework itself, enabling direct access to the correct offsets while ensuring that the vtable can be extended without breaking compatibility.
+Vtable requirements can be reordered or expanded, making their layout unstable across versioning resilience boundaries. To manage this, the binary framework includes a dispatch thunk which abstracts vtable offsets. The dispatch thunk is part of the framework itself, enabling direct access to the correct offsets while ensuring that the vtable can be extended without breaking compatibility.
 
 ### Virtual classes
 
--- a/src/docs/swift-code-generation.md
+++ b/src/docs/swift-code-generation.md
@@ -10,7 +10,7 @@ Below, we outline the specific scenarios where Swift wrappers are required.
 
 Priority: High (required for StoreKit2 and SwiftUI)
 
-Discriminated unions are not supported in C#, and their memory layout in Swift is opaque, fragile, and unstable. To interact with Swift enums, a wrapper is necessary for payoload access and initialization. 
+Discriminated unions are not supported in C#. To interact with Swift enums, a wrapper is necessary for payload access and initialization. 
 
 [Implementation details](binding-enums.md)
 
--- a/src/docs/swift-code-generation.md
+++ b/src/docs/swift-code-generation.md
@@ -1,0 +1,70 @@
+# Swift code generation
+
+Certain parts of the Swift ABI are unsable, such as metadata, async context, dynamic dispatch thunks, and cannot be directly projected into .NET without Swift wrappers. These wrappers provide the flexibility to encapsulate unstable ABI parts into a stable calling convention.
+
+The main tradeoff is between added flexibility for supporting more unsable ABI scenarios and the increased cost of shipping and trimming. To minimize complexity, we propose generating Swift wrappers only when absolutely necessary, keeping them as thin as possible.
+
+Below, we outline the specific scenarios where Swift wrappers are required.
+
+## Enums
+
+Priority: High (required for StoreKit2 and SwiftUI)
+
+Discriminated unions are not supported in C#, and their memory layout in Swift is opaque, fragile, and unstable. To interact with Swift enums, a wrapper is necessary for payoload access and initialization. 
+
+[Implementation details](binding-enums.md)
+
+## Async
+Priority: High (required for StoreKit2 and SwiftUI)
+
+Async context in Swift is unstable and require thin wrappers for synhronous invocation. The wrapper will convert async functions into synchronous calls by using a Task as an async worker and a DispatchSemaphore to block until the operation completes. This approach encapsulates the unstable Swift async context, exposing only the stable wrapper interface to .NET.
+
+Here is an example:
+```swift
+public func loadProductsSync(using storeManager: StoreManager) -> [Product]? {
+    let semaphore = DispatchSemaphore(value: 0)
+    var result: [Product]?
+    
+    Task {
+        result = await storeManager.loadProducts()
+        semaphore.signal()
+    }
+    
+    semaphore.wait()
+    return result
+}
+```
+
+## Dynamic dispatch
+Priority: Low
+
+Vtable requirements can be reordered or expanded, making their layout unstable across resilience boundaries. To manage this, the binary framework includes a dispatch thunk which abstracts vtable offsets. The dispatch thunk is part of the framework itself, enabling direct access to the correct offsets while ensuring that the vtable can be extended without breaking compatibility.
+
+### Virtual classes
+
+The goal is to project C# subclass of an open Swift class in Swift. Virtual classes have dynamic dispatch. If they are within the same module, ordered vtable is used for dispatch. Otherwise, a dispatch thunk is generated and invoked to allow for changes.
+
+We want to support:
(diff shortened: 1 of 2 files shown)


src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
+# Swift code generation
+
+Certain parts of the Swift ABI are unsable, such as metadata, async context, dynamic dispatch thunks, and cannot be directly projected into .NET without Swift wrappers. These wrappers provide the flexibility to encapsulate unstable ABI parts into a stable calling convention.
```
jkotas:
```suggestion
Certain parts of the Swift ABI are unstable, such as metadata, async context, dynamic dispatch thunks, and cannot be directly projected into .NET without Swift wrappers. These wrappers provide the flexibility to encapsulate unstable ABI parts into a stable calling convention.
```

jkotas: How can Apple use these in Apple OS APIs when they are unstable?

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
...
+Priority: High (required for StoreKit2 and SwiftUI)
+
+Discriminated unions are not supported in C#, and their memory layout in Swift is opaque, fragile, and unstable. To interact with Swift enums, a wrapper is necessary for payoload access and initialization. 
+
+[Implementation details](binding-enums.md)
+
+## Async
+Priority: High (required for StoreKit2 and SwiftUI)
+
+Async context in Swift is unstable and require thin wrappers for synhronous invocation. The wrapper will convert async functions into synchronous calls by using a Task as an async worker and a DispatchSemaphore to block until the operation completes. This approach encapsulates the unstable Swift async context, exposing only the stable wrapper interface to .NET.
```
jkotas:
```suggestion
Async context in Swift is unstable and require thin wrappers for synchronous invocation. The wrapper will convert async functions into synchronous calls by using a Task as an async worker and a DispatchSemaphore to block until the operation completes. This approach encapsulates the unstable Swift async context, exposing only the stable wrapper interface to .NET.
```

jkotas: I would expect that async Swift APIs are going to be exposed as async methods in C#.

"Sync-over-async" is a well-known antipattern that results in poor performance. (It is possible that it is not a real problem for the set of Swift APIs that we are looking at the moment.)

src/docs/swift-code-generation.md
kotlarmilos: Referring to runtime-generated features, such as vtable in the case of generics. Additionally, this would require reverse-engineering async and closure contexts.

It's likely we'll encounter cases that are hard to represent directly in C#. Even if we manage to do so, the maintenance cost would be high. Swift wrappers can address this issue by leveraging the Swift compiler/runtime to bridge the gap.

src/docs/swift-code-generation.md
kotlarmilos: We want to offer a solution for async in StoreKit2: https://developer.apple.com/documentation/storekit/product/3851116-products

> "Sync-over-async" is a well-known antipattern that results in poor performance.

We would be able to support callconv requirements for the Swift async. I am concerned about the context the layout is unknown/unstable. To resolve this, we could use a think Swift wrapper to represent the context, but project it as `IAsyncStateMachine` in C#.

I will explore options and update this proposal accordingly.

vitek-karas: We just need to be able to create a C# `Task` on top of the underlying swift. There are many ways to do that on the C# side - we just need to find what we can get from Swift (for example, being able to register a callback which is called when the async operation is finished would be enough to build a `Task` out of it on the C# side).

src/docs/swift-code-generation.md
jkotas: I do not think "unstable ABI" is the right word to describe this. Unstable ABI is expected to break from version to version. OS APIs that apps depend on cannot have unstable ABI - it holds independent on .NET. 

I would call this "ABI details that are too complex to project to .NET". Does this suggestion describe the problem well?

jkotas: Alternative: "poorly documented ABI details that are too complex to project to .NET".

kotlarmilos: Yes, sounds good.

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
+# Swift code generation
+
+Certain parts of the Swift ABI are poorly documented and too complex to project directly to .NET, such as metadata, async contexts, and dynamic dispatch thunks. In these cases, Swift wrappers offer the flexibility to encapsulate these poorly documented ABI components within Swift code, enabling interop through the stable ABI.
```
jkotas:
```suggestion
Certain parts of the Swift ABI are poorly documented and too complex to project directly to .NET, such as metadata, async contexts, and dynamic dispatch thunks. In these cases, Swift wrappers offer the flexibility to encapsulate these poorly documented ABI components within Swift code, enabling more maintainable C#/Swift interop boundary.
```
avoid "stable ABI" term

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
...
+
+The main tradeoff is between added flexibility for supporting more unsable ABI scenarios and the increased cost of shipping and trimming. To minimize complexity, we propose generating Swift wrappers only when absolutely necessary, keeping them as thin as possible.
```
jkotas:
```suggestion
The main tradeoff is between more maintainable C#/Swift interop boundary and the increased cost of shipping and trimming. To minimize complexity, we propose generating Swift wrappers only when absolutely necessary, keeping them as thin as possible.
```
I do not think that this gives your flexibility. The solution without Swift wrappers would be as much as flexible, but it would be much more complicated and harder to maintain.

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
...
+
+Below, we outline the specific scenarios where Swift wrappers are required.
+
+## Enums
+
+Priority: High (required for StoreKit2 and SwiftUI)
+
+Discriminated unions are not supported in C#, and their memory layout in Swift is opaque, fragile, and unstable. To interact with Swift enums, a wrapper is necessary for payoload access and initialization.
```
jkotas:
```suggestion
Discriminated unions are not supported in C#. To interact with Swift enums, a wrapper is necessary for payload access and initialization. 
```

> fragile, and unstable

I do not think that this is valid statement. (Also, it is not clear what you mean by fragile.)

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
...
+Priority: High (required for StoreKit2 and SwiftUI)
+
+Discriminated unions are not supported in C#, and their memory layout in Swift is opaque, fragile, and unstable. To interact with Swift enums, a wrapper is necessary for payoload access and initialization. 
+
+[Implementation details](binding-enums.md)
+
+## Async
+Priority: High (required for StoreKit2 and SwiftUI)
+
+Async context in Swift is unknown and require thin wrappers for synchronous invocation. The wrapper will convert async functions into synchronous calls by using a Task as an async worker and a DispatchSemaphore to block until the operation completes. This approach encapsulates the unknown Swift async context, exposing only the stable wrapper interface to .NET.
```
jkotas, line 20: > Async context in Swift is unknown

I am not sure what you mean by "unknown". This should use a better word.

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
...
+    
+    semaphore.wait()
+    return result
+}
+```
+
+## Dynamic dispatch
+Priority: Low
+
+Vtable requirements can be reordered or expanded, making their layout unstable across resilience boundaries. To manage this, the binary framework includes a dispatch thunk which abstracts vtable offsets. The dispatch thunk is part of the framework itself, enabling direct access to the correct offsets while ensuring that the vtable can be extended without breaking compatibility.
```
jkotas:
```suggestion
Vtable requirements can be reordered or expanded, making their layout unstable across versioning resilience boundaries. To manage this, the binary framework includes a dispatch thunk which abstracts vtable offsets. The dispatch thunk is part of the framework itself, enabling direct access to the correct offsets while ensuring that the vtable can be extended without breaking compatibility.
```
?

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
...
+-	C# class of an open Swift class in Swift
+-	C# class of an open Swift class in C#
+-	C# subclass of an open Swift class in Swift
+-	C# subclass of an open Swift class in C#
+When a C# class is instantiated from an open Swift class, it should contain a vtable generated on the Swift side. This vtable is sufficient for performing dynamic dispatch. For a subclass of an open Swift class, a Swift wrapper is generated. The wrapper overrides all virtual methods and delegates them to C#.
+
+When called from C#, the wrapper will invoke the superclass implementation in Swift. When called from Swift, the wrapper will use a simulated vtable. This simulated vtable calls into C#, retrieves a GCHandle for the actual C# class, and invokes its implementation, which in turn calls the Swift superclass implementation.
+
+For types with generics, a vtable is generated for each specialization of the type. Additionally, due to limitations in how C# handles unmanaged callbacks, it is not possible to have an unmanaged callback within a generic class. As a result, separate unmanaged and managed vtables are required.
+
```
stephen-hawley, line 57: It's worth noting that there are very, very few uses of classes in idiomatic Swift (as opposed to Swift written to be compatible with ObjC) and as a result, binding classes can be considered, necessary but a lower priority.

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,70 @@
...
+[Implementation details](binding-classes.md)
+
+### Protocols
+
+The goal is to project C# class in Swift that implements a protocol. Otherwise, a dispatch thunk is generated and invoked to allow for changes.
```
jkurdek, line 62: I'm not sure I understand this section's wording. I presume we are talking only about the situations when we project the protocol into c#, and then we create a new type on the c# side which implements the said protocol. And then we:

1. Allow the type to be passed into Swift whenever protocol argument is expected. Swift should then route all the calls back into C# using the simulated vtable.
2. Allow C# to treat the type as if Swift did not exists (C# call to the the interface method should correctly go to C# implementation of this method)

Do I get this right?

kotlarmilos: Yes, the text is missing here. Let me update this section.

kotlarmilos:
Closing this PR in favor of https://github.com/dotnet/runtimelab/pull/2816.

PR #2820, opened 2024-12-02
[SwiftBindings] Add Type Database documentation

stephen-hawley:
Added documentation for the binding-time type database.

### DIFF
--- a/src/docs/binding-typedatabase.md
+++ b/src/docs/binding-typedatabase.md
@@ -1,0 +1,58 @@
+# Binding Type Database
+
+In binding a Swift module, there will be nominal type references presented to the binder that have been bound previously. The binder will need to perform a few important operations on these types:
+
+1. Determine the C# type that should be used in the public-facing API and what namespaces need to be references
+2. Determine the size, stride, and alignment of value types that contain the type reference, if known at binding time.
+
+After binding the module, the binder will have new type database entries to write.
+
+Therefore, it makes sense to aggregate a collection of type database files for each module. The binding tool can either read them all in at the start of the world or read them in lazily for each module reference.
+
+The file format for the type database files is not vitally important beyond the fact that it should be easily serializable. Either XML or JSON would be fine.
+
+Here are the elements that are likely to be needed in a type database entry:
+
+- Swift module name
+- Swift type name (may be a type path for inner types)
+- Size, stride, and alignment (if knowable)
+- C# namespace
+- C# type name (may be a type path for inner types)
+- Swift entity type (stuct, enum, class, actor, protocol)
+- Whether or not the type is blitable
+
+(idea - maybe also include an optional C# pinvoke type)
+
+At binding time, a two-stage process will be necessary for new types introduced by the module being bound. This is because types may reference other types in the module before they have been defined. For example, this is valid Swift:
+
+```swift
+public struct B {
+    public var a: A // not yet defined
+    public init (theA: A) {
+        a = theA
+    }
+}
+
+public struct A {
+    public var b: Int = 7
+}
+```
+
+The first stage is type introduction. During this stage, the type database will be populated with a "light" version of the information available. The information that is available at this point is limited to the type names and the entity types.
+
+The second stage is type binding, wherein the actual C# type is generated. At this stage, it will be necessary to have the full information, in addition to other information about the types. As you see in the example above, it will be to either build a dependency graph to inform the binding order or to make the binding process recursive. The downside to makeing the binding process recursive is that it makes it more challenging to bind in parallel.
+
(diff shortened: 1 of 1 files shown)


dalexsoto (approved):

rolfbjarne (approved):

vitek-karas:
If I'm an external developer, how do I get a hold of these files, so that when my Swift library references a type from Foundation, the tooling will know to not project it again, but instead depend on the projection provided in one of our NuGet packages?

src/docs/binding-typedatabase.md
```diff
@@ -0,0 +1,58 @@
...
+}
+
+public struct A {
+    public var b: Int = 7
+}
+```
+
+The first stage is type introduction. During this stage, the type database will be populated with a "light" version of the information available. The information that is available at this point is limited to the type names and the entity types.
+
+The second stage is type binding, wherein the actual C# type is generated. At this stage, it will be necessary to have the full information, in addition to other information about the types. As you see in the example above, it will be to either build a dependency graph to inform the binding order or to make the binding process recursive. The downside to makeing the binding process recursive is that it makes it more challenging to bind in parallel.
```
jkurdek, line 43:
```suggestion
The second stage is type binding, wherein the actual C# type is generated. At this stage, it will be necessary to have the full information, in addition to other information about the types. As you see in the example above, it will be to either build a dependency graph to inform the binding order or to make the binding process recursive. The downside to making the binding process recursive is that it makes it more challenging to bind in parallel.
```

Also, I would like to further understand the need of binding in parallel. I understand that it will affect performance. Do we expect that binding will take a considerable time and that this will be visible by the user (users will need to generate bindings themselves?)

src/docs/binding-typedatabase.md
```diff
@@ -0,0 +1,58 @@
+# Binding Type Database
+
+In binding a Swift module, there will be nominal type references presented to the binder that have been bound previously. The binder will need to perform a few important operations on these types:
+
+1. Determine the C# type that should be used in the public-facing API and what namespaces need to be references
```
jkurdek, line 5:
```suggestion
1. Determine the C# type that should be used in the public-facing API and what namespaces need to be referenced.
```

src/docs/binding-typedatabase.md
```diff
@@ -0,0 +1,58 @@
...
+The file format for the type database files is not vitally important beyond the fact that it should be easily serializable. Either XML or JSON would be fine.
+
+Here are the elements that are likely to be needed in a type database entry:
+
+- Swift module name
+- Swift type name (may be a type path for inner types)
+- Size, stride, and alignment (if knowable)
+- C# namespace
+- C# type name (may be a type path for inner types)
+- Swift entity type (stuct, enum, class, actor, protocol)
```
jkurdek, line 21: I think we will need information whether the type is frozen or not, as this will change the way we bind references to this type (e.g. when binding a function whose signature contains a non-frozen struct we will need to replace the struct argument with pointer argument)

jkurdek (approved):

src/docs/binding-typedatabase.md
stephen-hawley: Premature optimization is never warranted, however if the code is designed with an eye to parallelism from the start, then there are some obvious gains. Since the binding process involves writing source code, it is likely to be I/O bound and parallelism will be an obvious here. It will be less visible to the user unless we choose to do binding-on-demand, and then it will matter quite a bit. It will matter to us for testing. BTfS had nearly 2K end-to-end tests (IIRC) and the total time to run was close to an hour. I'd like to do better than that.

stephen-hawley:
> If I'm an external developer, how do I get a hold of these files, so that when my Swift library references a type from Foundation, the tooling will know to not project it again, but instead depend on the projection provided in one of our NuGet packages?

Likely it will be packaged alongside the bindings, although because this is a packaging matter, it is beyond the scope of this document.

src/docs/binding-typedatabase.md
vitek-karas: Just an observation - the code flow is typically not that hard to refactor (Recursion to multiple-pass to something else - we could use the DependencyFramework from ilc/crossgen for example, after all it's built for compiler like scenarios and producing bindings is very similar to compilers).

What makes parallelism possible is data structures - so for example if we said that the input data structures would be immutable. And in that case I do agree that it also helps with testing, not necessarily for perf, but for robustness. Immutable data means that it's highly unlikely tests will show different behavior depending on the order in which they're executed (as an example).

vitek-karas:
> Likely it will be packaged alongside the bindings, although because this is a packaging matter, it is beyond the scope of this document.

Thanks - just a suggestion - in that case we might want to consider encoding this data into the generated C# (for example as attributes). Experience with other tools says that IL is really good for these things, because we have well maintained reader/writer implementations.

PR #2847, opened 2024-12-03
[RuntimeAsync] Fix support for pinvokes

jakobbotsch:
- Make sure async transformation leaves a scratch BB around for the pinvoke prolog to be inserted into
- Make sure we properly insert pinvoke epilogs for `GT_RETURN_SUSPEND` nodes (needed on x86)
- Ensure the pinvoke frame related locals are not captured by the async transformation

### DIFF
--- a/src/coreclr/jit/async.cpp
+++ b/src/coreclr/jit/async.cpp
@@ -60,6 +60,16 @@ public:
             return true;
         }
 
+        if (lclNum == m_comp->info.compLvFrameListRoot)
+        {
+            return true;
+        }
+
+        if (lclNum == m_comp->lvaInlinedPInvokeFrameVar)
+        {
+            return true;
+        }
+
 #ifdef FEATURE_EH_WINDOWS_X86
         if (lclNum == m_comp->lvaShadowSPslotsVar)
         {
@@ -1296,11 +1306,21 @@ void Async2Transformation::CreateResumptionSwitch()
 {
     BasicBlock* newEntryBB = BasicBlock::New(m_comp, BBJ_ALWAYS);
 
+    bool hadScratchBB = m_comp->fgFirstBBisScratch();
+
     if (m_comp->fgFirstBB->hasProfileWeight())
     {
         newEntryBB->inheritWeight(m_comp->fgFirstBB);
     }
     m_comp->fgFirstBB->bbRefs--;
+
+    if (hadScratchBB)
+    {
+        // If previous first BB was a scratch BB then we have to recreate it
+        // after since later phases may be relying on it.
+        // TODO-Cleanup: The later phases should be creating it if they need it.
+        m_comp->fgFirstBBScratch = nullptr;
+    }
 
     FlowEdge* toPrevEntryEdge = m_comp->fgAddRefPred(m_comp->fgFirstBB, newEntryBB);
     toPrevEntryEdge->setLikelihood(1);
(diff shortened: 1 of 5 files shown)


jakobbotsch:
PTAL @VSadov

src/tests/async/pinvoke.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            Assert.Equal(5, GetFPtr()());
+        }
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static unsafe delegate* unmanaged<int> GetFPtr() => &GetValue;
+
+    [UnmanagedCallersOnly]
+    private static int GetValue() => 5;
```
VSadov, line 43: A thought: what should happen if `UnmanagedCallersOnly` is applied to runtime async method. 
Does it work today with ordinary async? (I guess, why not, they are just Task-returning methods)

src/coreclr/jit/async.cpp
```diff
@@ -1296,12 +1306,22 @@ void Async2Transformation::CreateResumptionSwitch()
...
     {
         newEntryBB->inheritWeight(m_comp->fgFirstBB);
     }
     m_comp->fgFirstBB->bbRefs--;
 
+    if (hadScratchBB)
+    {
+        // If previous first BB was a scratch BB then we have to recreate it
+        // after since later phases may be relying on it.
+        // TODO-Cleanup: The later phases should be creating it if they need it.
```
VSadov, line 1321: Do you plan to address this TODO or this is just a "good to have" comment?

VSadov: I do agree that relying on the presence of scratch is an arbitrary dependency that just happen to hold. 
If creating on demand is not too hard, it would be better.

src/tests/async/pinvoke.cs
jakobbotsch: I haven't checked, but I imagine `UnmanagedCallersOnly` on any function returning an object reference should result in `IlegalProgramException`. So I would hope that was the case for async1, and we should similarly disallow it for async2.

src/coreclr/jit/async.cpp
jakobbotsch: > Do you plan to address this TODO or this is just a "good to have" comment?

I don't plan to address it here, this should rather be cleaned up upstream.

> I do agree that relying on the presence of scratch is an arbitrary dependency that just happen to hold.
If creating on demand is not too hard, it would be better.

The frontend does ensure it:

https://github.com/dotnet/runtimelab/blob/4daa761347cfec8b8a0cd1330e4f05b3f1d5a3c1/src/coreclr/jit/flowgraph.cpp#L2315-L2322

It's simply that this is a very long-term invariant that is error prone and which feels unnecessary to me.

src/tests/async/pinvoke.cs
VSadov: > I haven't checked, but I imagine UnmanagedCallersOnly on any function returning an object reference should result in IlegalProgramException. So I would hope that was the case for async1, and we should similarly disallow it for async2.

I checked. 
Even though the documentation only mentions that parameters must be blittable, the runtime requires that the return type is blittable as well. That rules out runtime async signatures. Even nongeneric `ValueType`, although a struct, is not blittable.

VSadov (approved):
LGTM. Thanks!

src/coreclr/jit/async.cpp
VSadov: Exactly. Such dependencies have negative value by default. They must be justified by advantages or scenarios that they enable. If arranging a scratch block was difficult to do later, it could be a possible justification, but, as I understand, it is not the case.

PR #2854, opened 2024-12-04
[RuntimeAsync] Fix a couple of cases where x86 would overwrite ecx on return

jakobbotsch:
shared-generics was not working on x86 because the returned continuation was being overwritten in one particular case. This fixes the issue.

### DIFF
--- a/src/coreclr/jit/codegenxarch.cpp
+++ b/src/coreclr/jit/codegenxarch.cpp
@@ -10000,10 +10000,10 @@ void CodeGen::genFnEpilog(BasicBlock* block)
         if (frameSize > 0)
         {
 #ifdef TARGET_X86
-            /* Add 'compiler->compLclFrameSize' to ESP */
-            /* Use pop ECX to increment ESP by 4, unless compiler->compJmpOpUsed is true */
+            // Add 'compiler->compLclFrameSize' to ESP. Use "pop ECX" for that, except in cases
+            // where ECX may contain some state.
 
-            if ((frameSize == TARGET_POINTER_SIZE) && !compiler->compJmpOpUsed)
+            if ((frameSize == TARGET_POINTER_SIZE) && !compiler->compJmpOpUsed && !compiler->compIsAsync2())
             {
                 inst_RV(INS_pop, REG_ECX, TYP_I_IMPL);
                 regSet.verifyRegUsed(REG_ECX);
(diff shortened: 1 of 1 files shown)


jakobbotsch:
PTAL @VSadov

src/coreclr/jit/codegenxarch.cpp
```diff
@@ -10000,19 +10000,19 @@ void CodeGen::genFnEpilog(BasicBlock* block)
...
 #ifdef TARGET_X86
-            /* Add 'compiler->compLclFrameSize' to ESP */
-            /* Use pop ECX to increment ESP by 4, unless compiler->compJmpOpUsed is true */
+            // Add 'compiler->compLclFrameSize' to ESP. Use "pop ECX" for that, except in cases
+            // where ECX may contain some state.
 
-            if ((frameSize == TARGET_POINTER_SIZE) && !compiler->compJmpOpUsed)
+            if ((frameSize == TARGET_POINTER_SIZE) && !compiler->compJmpOpUsed && !compiler->compIsAsync2())
             {
                 inst_RV(INS_pop, REG_ECX, TYP_I_IMPL);
```
VSadov, line 10008: If ECX is just a "some volatile register that we can trash", can we pop EDX instead?

jakobbotsch: `edx` is used to return longs on x86, so we can't easily switch to pop into that either.

VSadov: and if the goal is to have a smaller opcode - can x64 do "pop edx" ?  (just curious)

VSadov: I guess using pop on x64 may not fit some expected epilogue pattern or smth like that.

VSadov: Anyways, here we just want to fix runtime async issue on x86 - for that the fix looks good.

VSadov (approved):
LGTM!!

src/coreclr/jit/codegenxarch.cpp
jakobbotsch: It does seem like we could optimize some epilog sequences in this way.
With Intel APX we are getting push2/pop2 instructions, and we are planning to look into prolog/epilog optimizations with those, so we'll probably look into it there.

PR #2860, opened 2024-12-05
[SwiftBindings] Added pinvokes for handling ARC

stephen-hawley:
These p/invokes will be needed for handling any heap-allocated Swift objects.

This should include anything that has a typical Swift object header.


src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,167 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Swift.Runtime;
+
+/// <summary>
+/// ARC is a class containing p/invokes for Swift reference counting and memory management.
+/// </summary>
+public static class ARC {
```
jkurdek: I am wondering about the naming. I do get that AutomaticReferenceCounting is quite lengthy. Could we at least put explanation that ARC means Automatic Reference Counting in the summary string?

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,167 @@
...
+    const string swiftCore = "libswiftCore.dylib";
+
+    /// <summary>
+    /// Retain a heap-allocated Swift object
+    /// </summary>
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    [DllImport (swiftCore)]
+    static extern void swift_retain (IntPtr p);
```
jkurdek: NIT: I think it would be beneficial to establish code style guidelines. Do we reuse and apply https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md ? Or we follow a different set of rules? For example, I think a space before parenthesis here is non-standard. 

I do not have any strong opinions about the rules, I would just like them to be consistent.

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,167 @@
...
+using System;
+using System.Runtime.InteropServices;
+
+namespace Swift.Runtime;
+
+/// <summary>
+/// ARC is a class containing p/invokes for Swift reference counting and memory management.
+/// </summary>
+public static class ARC {
+    const string swiftCore = "libswiftCore.dylib";
```
kotlarmilos: Is the library always available at the relative path?

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,167 @@
...
+    [return: MarshalAs (UnmanagedType.I1)]
+    static extern bool swift_isDeallocating (IntPtr p);
+
+    /// <summary>
+    /// Check to see if a pointer is in the process of being deallocated.
+    /// </summary>
+    /// <param name="p">A non-null pointer to an unmanaged Swift object</param>
+    /// <returns>True if and only if the pointer in the process of being deallocated.</returns>
+    /// <exception cref="ArgumentOutOfRangeException">Throws if p is null</exception>
+    public static bool IsDeallocating (IntPtr p)
```
kotlarmilos: Is this call thread-safe?

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,167 @@
...
+        swift_release (p);
+        return p;
+    }
+
+    /// <summary>
+    /// Retains an 'unowned' heap-allocated Swift object.
+    /// </summary>
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    [DllImport (swiftCore)]
+    static extern void swift_unownedRetain (IntPtr p);
```
kotlarmilos: What are unowned cases?

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,167 @@
...
+        return p;
+    }
+
+    /// <summary>
+    /// Returns the retain count for the heap-allocated Swift object.
+    /// </summary>
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    /// <returns>The retain count</returns>
+    [DllImport (swiftCore)]
+    static extern nint swift_retainCount (IntPtr p);
```
kotlarmilos: Is this thread-safe?

src/Swift.Runtime/src/Library/ARC.cs
jkotas: If we go with the abbreviation, the capitalization should be `Arc` to conform with [.NET Design Guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers).

jkotas: > I do get that AutomaticReferenceCounting is quite length

This is not an API that will be frequently used by user code, so the long name is not a big deal. You can always do `using` for files where it is used very frequently.

src/Swift.Runtime/src/Library/ARC.cs
stephen-hawley: There are 3 reference counts in every ARC object, strong (default), unowned, and weak. We care about strong and unowned.

src/Swift.Runtime/src/Library/ARC.cs
stephen-hawley: If you want. I'm following the Xamarin style as it's what the macios group uses. If/when this gets moved there, there is a bot to enforce C# formatting.

src/Swift.Runtime/src/Library/ARC.cs
stephen-hawley: It looks like it. All the memory calls that reference the bits are protected with `std::atomic`

src/Swift.Runtime/src/Library/ARC.cs
stephen-hawley: Yes.

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,168 @@
...
+    /// <exception cref="ArgumentOutOfRangeException">Throws if p is null</exception>
+    /// <exception cref="Exception">Throws if p points to an object that has been deinitialized</exception>
+    public static IntPtr Release(IntPtr p)
+    {
+        if (p == IntPtr.Zero)
+            throw new ArgumentOutOfRangeException(nameof(p));
+        if (swift_isDeallocating(p))
+        {
+            var format = $"X{IntPtr.Size * 2}";
+            throw new Exception($"Attempt to release a Swift object that has been deinitialized {p.ToString(format)}");
```
rolfbjarne: Not using interpolated strings makes adding translations easier later on:

```suggestion
            throw new Exception("Attempt to release a Swift object that has been deinitialized {0}", p.ToString(format));
```

rolfbjarne (approved):

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,168 @@
+// Copyright (c) Microsoft Corporation.
```
vitek-karas: Nit: Please rename the file to match the capitalization of the class.

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,168 @@
...
+/// ARC is a class containing p/invokes for Swift Automatic Reference Counting and memory management.
+/// </summary>
+public static class Arc {
+    const string swiftCore = "/usr/lib/swift/libswiftCore.dylib";
+
+    /// <summary>
+    /// Retain a heap-allocated Swift object
+    /// </summary>
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    [DllImport(swiftCore)]
```
vitek-karas: Especially in the swift related code I think it would be good to be explicit about calling conventions. To me it's surprising this uses the C calling convention.

src/Swift.Runtime/src/Library/ARC.cs
```diff
@@ -0,0 +1,168 @@
...
+    /// </summary>
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    /// <returns>The pointer passed in</returns>
+    /// <exception cref="ArgumentOutOfRangeException">Throws if p is null</exception>
+    /// <exception cref="Exception">Throws if p points to an object that has been deinitialized</exception>
+    public static IntPtr Release(IntPtr p)
+    {
+        if (p == IntPtr.Zero)
+            throw new ArgumentOutOfRangeException(nameof(p));
+        if (swift_isDeallocating(p))
```
vitek-karas: Why do we have to check this? I would expect Swift to do this - is it because we don't have a way to catch that error?

jkurdek (approved):

src/Swift.Runtime/src/Library/ARC.cs
stephen-hawley: Swift does not check for null in some cases. In the cases that it checks, there is no error reported. In the cases that it doesn't check, the code will crash hard if we're lucky. In the case of checking for isDeallocating, this is something that doesn't happen in Swift by contract because the compiler generates the calls to that code. If this gets triggered, we've done something horribly wrong.
Should I put this in a `#if DEBUG`?

dalexsoto (approved):

src/Swift.Runtime/src/Library/Arc.cs
```diff
@@ -0,0 +1,167 @@
...
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    [DllImport(swiftCore, CallingConvention = CallingConvention.Cdecl)]
+    static extern void swift_retain(IntPtr p);
+
+    /// <summary>
+    /// Retain a heap-allocated Swift object
+    /// </summary>
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    /// <returns>The pointer passed in.</returns>
+    /// <exception cref="ArgumentOutOfRangeException">Throws if p is null</exception>
```
mcumming: Why not `NullReferenceException`, `ArgumentOutOfRangeException` while accurate, seems odd.

src/Swift.Runtime/src/Library/Arc.cs
```diff
@@ -0,0 +1,167 @@
...
+    public static IntPtr Retain(IntPtr p)
+    {
+        if (p == IntPtr.Zero)
+            throw new ArgumentOutOfRangeException(nameof(p));
```
mcumming: This might just be me but:
If any of these argument checks fails and an exception is thrown, the message that the developer will see would be like:
>System.ArgumentOutOfRangeException: 'Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'p')'.

Since there is little maintenance involved here, can the parameter name be a little more descriptive?

mcumming (approved):

src/Swift.Runtime/src/Library/Arc.cs
jkotas: Why not `ArgumentNullException`? 

There is a lot of prior API for `ArgumentNullException` for null IntPtr that is actually a pointer. One example from many: `System.Runtime.InteropServices.Marshal.AddRef`.

src/Swift.Runtime/src/Library/Arc.cs
```diff
@@ -0,0 +1,167 @@
...
+    [DllImport(swiftCore, CallingConvention = CallingConvention.Cdecl)]
+    static extern void swift_retain(IntPtr p);
+
+    /// <summary>
+    /// Retain a heap-allocated Swift object
+    /// </summary>
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    /// <returns>The pointer passed in.</returns>
+    /// <exception cref="ArgumentOutOfRangeException">Throws if p is null</exception>
+    public static IntPtr Retain(IntPtr p)
```
jkotas, line 28: The current API design guideline is to use unmanaged pointer type (e.g. `void*`) for unmanaged pointers. Would it be appropriate here?

src/Swift.Runtime/src/Library/ARC.cs
vitek-karas: I think it's OK to keep it. The perf difference it probably unobservable, and if it does hit, we will at least have some more meaningful info about the failure.

src/Swift.Runtime/src/Library/Arc.cs
vitek-karas: @jkotas - I'm curious about the guideline here. The pointer is opaque to us on the managed side (in fact if we ever look inside it would be a problem). It's more like a handle for the managed side. `void *` implies `unsafe` which doesn't seem to fit these cases. There are actually quite a few places where we need to store the Swift's reference to the unmanaged object - if all of those would need to be `unsafe` it would mean effectively everything we do would need to be `unsafe`.

I would prefer that for places where we treat it as a handle, we keep it IntPtr (or maybe even some our struct which wraps it). And the very few places where we might need to treat it as a pointer, those would need to be `unsafe` since we would need to start using unmanaged pointers.

jkotas: If it is more like a handle, IntPtr is fine.

src/Swift.Runtime/src/Library/ARC.cs
matouskozak: @stephen-hawley is Xamarin following the [mono guidelines](https://www.mono-project.com/community/contributing/coding-guidelines/) or is there something special from Xamarin?

src/Swift.Runtime/src/Library/Arc.cs
```diff
@@ -0,0 +1,167 @@
...
+    /// <summary>
+    /// Retain a heap-allocated Swift object
+    /// </summary>
+    /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
+    /// <returns>The pointer passed in.</returns>
+    /// <exception cref="ArgumentNullException">Throws if p is null</exception>
+    public static IntPtr Retain(IntPtr p)
+    {
+        if (p == IntPtr.Zero)
+            throw new ArgumentNullException(nameof(p));
```
jkotas, line 31:
```suggestion
        ArgumentNullException.ThrowIfNull(p);
```
(dtto for other places)

stephen-hawley: This is illegal without an unsafe cast to `void *`.

jkotas: The right overload is internal unfortunately: https://github.com/dotnet/runtime/blob/d6be63071b06354d581060e633cdb64da307d37c/src/libraries/System.Private.CoreLib/src/System/ArgumentNullException.cs#L87

PR #2864, opened 2024-12-07
[NativeAOT-LLVM] Rename FEATURE_WASM_THREADS to FEATURE_WASM_MANAGED_THREADS

yowl:
...to match upstream

### DIFF
--- a/src/coreclr/gc/gcpriv.h
+++ b/src/coreclr/gc/gcpriv.h
@@ -188,7 +188,7 @@ inline void FATAL_GC_ERROR()
 #define FEATURE_PREMORTEM_FINALIZATION
 #define GC_HISTORY
 
-#if !defined(TARGET_WASM) || defined(FEATURE_WASM_THREADS)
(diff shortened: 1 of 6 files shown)


yowl:
Cc @dotnet/nativeaot-llvm

SingleAccretion (approved):

pavelsavara:
I think `_MANAGED_` in the `FEATURE_WASM_MANAGED_THREADS` name stems from earlier phase of Mono on browser, where EventPipe used emscripter/pthread but managed code was still only ST running on UI thread. That combination doesn't work anymore.

SingleAccretion:
> I think _MANAGED_ in the FEATURE_WASM_MANAGED_THREADS name stems from earlier phase of Mono on browser, where EventPipe used emscripter/pthread but managed code was still only ST running on UI thread.

Yeah... I think it's still beneficial to use the same name everywhere, even if it's not really "the best".

PR #2867, opened 2024-12-09
added TypeMetadataCache

stephen-hawley:
Added caching for type metadata.
The cache will be used by most bindings in their static GetTypeMetadata() accessor.

Typical usage will be:
```csharp
public struct SomeGeneratedBinding : ISwiftObject // this will come later, but it defines the GetTypeMetadata() method
    public static TypeMetadata GetTypeMetadata() {
        return TypeMetadata.Cache.GetOrAdd(typeof(SomeGeneratedBinding), (t) => PInvokesForSGB._MetadataAccessor());
    }
}
```
It will also be used later on in the set of static methods that will be used to get the type metadata for a given type.  For example, if the bound type is generic, it will look like this:

```
public struct SomeGeneric<T> : ISwiftObject {
    public static TypeMetadataGetTypeMetadata() {
        return TypeMetadata.Cache.GetOrAdd(typeof(SomeGeneric<T>), (t) => PInvokesForSG._MetadataAccessor (TypeMetadata.GetMetadataOrThrow(typeof(T))));
    }
}
```
The notion being that you call `GetOrAdd` only when you have something to put in (hence the factory), but you call GetMetadataOrThrow() when it's possible to fail. This is the case since we don't know what T will contain at compile time.

The implementation of GetMetadataOrThrow() will look in the cache via TryGet before trying other approaches.

### DIFF
--- a/src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
+++ b/src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
@@ -12,27 +12,20 @@ namespace Swift.Runtime;
 public interface ITypeMetadataCache
 {
     /// <summary>
-    /// Returns true if and only if the cache contains an entry for t
-    /// </summary>
-    /// <param name="type">The type to look up in the cache</param>
-    /// <returns>true if and only if the cache contains t, false otherwise</returns>
-    bool Contains(Type type);
-
-    /// <summary>
-    /// Returns true if the cache contains an entry for t and sets metadata to the resulting value,
+    /// Returns true if the cache contains an entry for type and sets metadata to the resulting value,
     /// otherwise it returns false and metadata will be null.
     /// </summary>
-    /// <param name="type"></param>
-    /// <param name="metadata"></param>
+    /// <param name="type">The type to look up in the cache</param>
+    /// <param name="metadata">The resulting metadata if found</param>
     /// <returns>true if the lookup was successful, false otherwise</returns>
     bool TryGet(Type type, [NotNullWhen(true)]out TypeMetadata? metadata);
 
     /// <summary>
-    /// Gets the TypeMetadata for the given Type t or if it is not present,
+    /// Gets the TypeMetadata for the given Type type or if it is not present,
     /// adds it to the cache using the given factory to generate the value.
     /// </summary>
-    /// <param name="type"></param>
-    /// <param name="metadataFactory"></param>
-    /// <returns>The TypeMetadata associated with the give Type t</returns>
+    /// <param name="type">The type to look up in the cache</param>
+    /// <param name="metadataFactory">a factory to generate the TypeMetadata if not present</param>
+    /// <returns>The TypeMetadata associated with the give Type type</returns>
     TypeMetadata GetOrAdd(Type type, Func<Type, TypeMetadata> metadataFactory);
 }
--- a/src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
+++ b/src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace Swift.Runtime;
+
+/// <summary>
+/// Represents an interface that defines how a cache for TypeMetadata should operate.
+/// The implementation of this interface needs to be thread safe.
+/// </summary>
+public interface ITypeMetadataCache
+{
+    /// <summary>
+    /// Returns true if and only if the cache contains an entry for t
+    /// </summary>
+    /// <param name="t">The type to look up in the cache</param>
+    /// <returns>true if and only if the cache contains t, false otherwise</returns>
+    bool Contains(Type t);
+
+    /// <summary>
+    /// Returns true if the cache contains an entry for t and sets metadata to the resulting value,
+    /// otherwise it returns false and metadata will be null.
+    /// </summary>
+    /// <param name="t"></param>
+    /// <param name="metadata"></param>
+    /// <returns>true if the lookup was successful, false otherwise</returns>
+    bool TryGet(Type t, [NotNullWhen(true)]out TypeMetadata? metadata);
+
+    /// <summary>
+    /// Gets the TypeMetadata for the given Type t or if it is not present,
+    /// adds it to the cache using the given factory to generate the value.
+    /// </summary>
+    /// <param name="t"></param>
+    /// <param name="metadataFactory"></param>
+    /// <returns>The TypeMetadata associated with the give Type t</returns>
+    TypeMetadata GetOrAdd(Type t, Func<Type, TypeMetadata> metadataFactory);
+}
--- a/src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
+++ b/src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
@@ -14,25 +14,25 @@ public interface ITypeMetadataCache
     /// <summary>
     /// Returns true if and only if the cache contains an entry for t
     /// </summary>
-    /// <param name="t">The type to look up in the cache</param>
+    /// <param name="type">The type to look up in the cache</param>
     /// <returns>true if and only if the cache contains t, false otherwise</returns>
-    bool Contains(Type t);
+    bool Contains(Type type);
 
     /// <summary>
     /// Returns true if the cache contains an entry for t and sets metadata to the resulting value,
     /// otherwise it returns false and metadata will be null.
     /// </summary>
-    /// <param name="t"></param>
+    /// <param name="type"></param>
     /// <param name="metadata"></param>
     /// <returns>true if the lookup was successful, false otherwise</returns>
-    bool TryGet(Type t, [NotNullWhen(true)]out TypeMetadata? metadata);
+    bool TryGet(Type type, [NotNullWhen(true)]out TypeMetadata? metadata);
 
     /// <summary>
     /// Gets the TypeMetadata for the given Type t or if it is not present,
     /// adds it to the cache using the given factory to generate the value.
     /// </summary>
-    /// <param name="t"></param>
+    /// <param name="type"></param>
     /// <param name="metadataFactory"></param>
     /// <returns>The TypeMetadata associated with the give Type t</returns>
-    TypeMetadata GetOrAdd(Type t, Func<Type, TypeMetadata> metadataFactory);
+    TypeMetadata GetOrAdd(Type type, Func<Type, TypeMetadata> metadataFactory);
 }
--- a/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
+++ b/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
@@ -244,7 +244,7 @@ public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
         return handle.GetHashCode ();
     }
 
-    static TypeMetadataCache cache;
+    static readonly TypeMetadataCache cache;
     /// <summary>
     /// Gets the type metadata cache for the runtime.
     /// </summary>
--- a/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
+++ b/src/Swift.Runtime/src/Metadata/TypeMetadata.cs
@@ -124,6 +124,12 @@ public enum TypeMetadataKind {
 public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
     readonly IntPtr handle;
 
+    static TypeMetadata ()
+    {
+        // TODO - add metadata for common built-in types like scalars and strings
+        cache = new TypeMetadataCache();
+    }
+
     /// <summary>
     /// An empty/invalid TypeMetadata object
     /// </summary>
@@ -237,4 +243,10 @@ public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
     {
         return handle.GetHashCode ();
     }
+
+    static TypeMetadataCache cache;
+    /// <summary>
+    /// Gets the type metadata cache for the runtime.
+    /// </summary>
+    public static ITypeMetadataCache Cache => cache;
 }
--- a/src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
+++ b/src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
@@ -29,7 +29,7 @@ internal class TypeMetadataCache : ITypeMetadataCache
     public TypeMetadataCache(IEnumerable<(Type, TypeMetadata)> initialValues)
     {
         var dictCache = (IDictionary<Type, TypeMetadata>)cache;
-        foreach (var (key, value) in initalValues)
+        foreach (var (key, value) in initialValues)
         {
             dictCache.Add(key, value);
         }
--- a/src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
+++ b/src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
@@ -1,0 +1,81 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Swift.Runtime;
+
+/// <summary>
+/// Internal implementation of ITypeMetadataCache
+/// </summary>
+internal class TypeMetadataCache : ITypeMetadataCache
+{
+    ConcurrentDictionary<Type, TypeMetadata> cache = new();
+
+    /// <summary>
+    /// Constructs an empty cache.
+    /// </summary>
+    public TypeMetadataCache()
+    {
+
+    }
+
+    /// <summary>
+    /// Constructs a cache with the supplied initial values.
+    /// </summary>
+    /// <param name="initalValues">An enumeration of tuples of Type and TypeMetadata to initialize the cache</param>
+    public TypeMetadataCache(IEnumerable<(Type, TypeMetadata)> initalValues)
+    {
+        var dictCache = (IDictionary<Type, TypeMetadata>)cache;
+        foreach (var (key, value) in initalValues)
+        {
+            dictCache.Add(key, value);
+        }
+    }
+
+
+    /// <summary>
+    /// Returns true if and only if the cache contains an entry for t.
+    /// </summary>
+    /// <param name="t">The type to look up in the cache</param>
+    /// <returns>true if and only if the cache contains t, false otherwise</returns>
+    public bool Contains(Type t)
+    {
+        return cache.ContainsKey(t);
+    }
+
+    /// <summary>
+    /// Returns true if the cache contains an entry for t and sets metadata to the resulting value,
+    /// otherwise it returns false and metadata will be null.
+    /// </summary>
+    /// <param name="t"></param>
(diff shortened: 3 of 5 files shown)


dalexsoto (approved):

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+/// The implementation of this interface needs to be thread safe.
+/// </summary>
+public interface ITypeMetadataCache
+{
+    /// <summary>
+    /// Returns true if and only if the cache contains an entry for t
+    /// </summary>
+    /// <param name="t">The type to look up in the cache</param>
+    /// <returns>true if and only if the cache contains t, false otherwise</returns>
+    bool Contains(Type t);
```
rolfbjarne: I think the guidelines says to not abbreviate (xml docs would need to be fixed accordingly as well):

```suggestion
    bool Contains(Type type);
```

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Returns true if the cache contains an entry for t and sets metadata to the resulting value,
+    /// otherwise it returns false and metadata will be null.
+    /// </summary>
+    /// <param name="t"></param>
+    /// <param name="metadata"></param>
+    /// <returns>true if the lookup was successful, false otherwise</returns>
+    bool TryGet(Type t, [NotNullWhen(true)]out TypeMetadata? metadata);
```
rolfbjarne:
```suggestion
    bool TryGet(Type type, [NotNullWhen(true)]out TypeMetadata? metadata);
```

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Gets the TypeMetadata for the given Type t or if it is not present,
+    /// adds it to the cache using the given factory to generate the value.
+    /// </summary>
+    /// <param name="t"></param>
+    /// <param name="metadataFactory"></param>
+    /// <returns>The TypeMetadata associated with the give Type t</returns>
+    TypeMetadata GetOrAdd(Type t, Func<Type, TypeMetadata> metadataFactory);
```
rolfbjarne:
```suggestion
    TypeMetadata GetOrAdd(Type type, Func<Type, TypeMetadata> metadataFactory);
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+
+    /// <summary>
+    /// Returns true if and only if the cache contains an entry for t.
+    /// </summary>
+    /// <param name="t">The type to look up in the cache</param>
+    /// <returns>true if and only if the cache contains t, false otherwise</returns>
+    public bool Contains(Type t)
```
rolfbjarne:
```suggestion
    public bool Contains(Type type)
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// <summary>
+    /// Returns true if the cache contains an entry for t and sets metadata to the resulting value,
+    /// otherwise it returns false and metadata will be null.
+    /// </summary>
+    /// <param name="t"></param>
+    /// <param name="metadata"></param>
+    /// <returns>true if the lookup was successful, false otherwise</returns>
+    public bool TryGet(Type t, [NotNullWhen(true)] out TypeMetadata? metadata)
```
rolfbjarne:
```suggestion
    public bool TryGet(Type type, [NotNullWhen(true)] out TypeMetadata? metadata)
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+        if (cache.TryGetValue(t, out var md))
+        {
+            metadata = md;
+            return true;
+        }
+        else
+        {
+            metadata = null;
+            return false;
+        }
```
rolfbjarne: This doesn't work?

```suggestion
        return cache.TryGetValue(t, out metadata);
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// <summary>
+    /// Gets the TypeMetadata for the given Type t or if it is not present,
+    /// adds it to the cache using the given factory to generate the value.
+    /// </summary>
+    /// <param name="t"></param>
+    /// <param name="metadataFactory"></param>
+    /// <returns>The TypeMetadata associated with the give Type t</returns>
+    public TypeMetadata GetOrAdd(Type t, Func<Type, TypeMetadata> metadataFactory)
```
rolfbjarne:
```suggestion
    public TypeMetadata GetOrAdd(Type type, Func<Type, TypeMetadata> metadataFactory)
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+    public TypeMetadataCache()
+    {
+
+    }
+
+    /// <summary>
+    /// Constructs a cache with the supplied initial values.
+    /// </summary>
+    /// <param name="initalValues">An enumeration of tuples of Type and TypeMetadata to initialize the cache</param>
+    public TypeMetadataCache(IEnumerable<(Type, TypeMetadata)> initalValues)
```
rolfbjarne:
```suggestion
    public TypeMetadataCache(IEnumerable<(Type, TypeMetadata)> initialValues)
```

rolfbjarne (changes requested):

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -237,4 +243,10 @@ public override int GetHashCode ()
     {
         return handle.GetHashCode ();
     }
+
+    static TypeMetadataCache cache;
```
jkurdek:
```suggestion
    private static TypeMetadataCache _cache;
```

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -124,6 +124,12 @@ public enum TypeMetadataKind {
 public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
     readonly IntPtr handle;
 
+    static TypeMetadata ()
```
jkurdek, line 127: This is private right? What is the intended usage?

jkurdek (approved):

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
mcumming: I believe this could be `readonly` as well.

mcumming (approved):

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
stephen-hawley: Oddly enough, no. The signature for ConcurrentDictionary doesn't match. Specifically, it's `public bool TryGetValue (TKey key, out TValue value);` so doing that gives you this error:
```
cannot convert from 'out Swift.Runtime.TypeMetadata?' to 'out Swift.Runtime.TypeMetadata'
```

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
stephen-hawley: FYI, adding private is redundant. The default visibility for types in C# is private.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
stephen-hawley: static constructor for initializing the cache.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
jkurdek: I know, i think the runtime code guideline recommends explicitly specifying access modifiers

stephen-hawley: Right - the mono style guideline recommends leaving out private entirely.  IIRC, `private` was added to C# for parity with Java not because it was needed.

jkurdek:
Both TypeMetadata and ValueWitnessTable types currently reside in src/Metadata directory which is not copied to the output during code generation (unlike e.g. SwiftHandle which resides in library directory). I think the reasoning was that the `Metadata` types were not intended to be used during runtime (please confirm @kotlarmilos ). But as we will definitely need those in runtime now, it would be good to figure out how they will make their way to output code.

stephen-hawley:
Probably Swift.Runtime should be an assembly instead of copying the source files. Opened an issue for that: https://github.com/dotnet/runtimelab/issues/2868

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+/// Represents an interface that defines how a cache for TypeMetadata should operate.
+/// The implementation of this interface needs to be thread safe.
+/// </summary>
+public interface ITypeMetadataCache
+{
+    /// <summary>
+    /// Returns true if and only if the cache contains an entry for t
+    /// </summary>
+    /// <param name="type">The type to look up in the cache</param>
+    /// <returns>true if and only if the cache contains t, false otherwise</returns>
```
rolfbjarne:
```suggestion
    /// <returns>true if and only if the cache contains type, false otherwise</returns>
```

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+    bool Contains(Type type);
+
+    /// <summary>
+    /// Returns true if the cache contains an entry for t and sets metadata to the resulting value,
```
rolfbjarne:
```suggestion
    /// Returns true if the cache contains an entry for type and sets metadata to the resulting value,
```

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// otherwise it returns false and metadata will be null.
+    /// </summary>
+    /// <param name="type"></param>
+    /// <param name="metadata"></param>
+    /// <returns>true if the lookup was successful, false otherwise</returns>
+    bool TryGet(Type type, [NotNullWhen(true)]out TypeMetadata? metadata);
+
+    /// <summary>
+    /// Gets the TypeMetadata for the given Type t or if it is not present,
```
rolfbjarne:
```suggestion
    /// Gets the TypeMetadata for the given Type type or if it is not present,
```

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// adds it to the cache using the given factory to generate the value.
+    /// </summary>
+    /// <param name="type"></param>
+    /// <param name="metadataFactory"></param>
+    /// <returns>The TypeMetadata associated with the give Type t</returns>
```
rolfbjarne:
```suggestion
    /// <returns>The TypeMetadata associated with the given Type type</returns>
```

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Swift.Runtime;
+
+/// <summary>
+/// Represents an interface that defines how a cache for TypeMetadata should operate.
+/// The implementation of this interface needs to be thread safe.
+/// </summary>
+public interface ITypeMetadataCache
+{
+    /// <summary>
+    /// Returns true if and only if the cache contains an entry for t
```
rolfbjarne:
```suggestion
    /// Returns true if and only if the cache contains an entry for type />
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
rolfbjarne: Ah I see, it's because `TypeMetadata` is a struct, and a nullable struct is not the same type as just a plain struct.

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+        var dictCache = (IDictionary<Type, TypeMetadata>)cache;
+        foreach (var (key, value) in initialValues)
+        {
+            dictCache.Add(key, value);
+        }
+    }
+
+
+    /// <summary>
+    /// Returns true if and only if the cache contains an entry for t.
```
rolfbjarne:
```suggestion
    /// Returns true if and only if the cache contains an entry for type.
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// </summary>
+    /// <param name="type">The type to look up in the cache</param>
+    /// <returns>true if and only if the cache contains t, false otherwise</returns>
```
rolfbjarne:
```suggestion
    /// <returns>true if and only if the cache contains type, false otherwise</returns>
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+    public bool Contains(Type type)
+    {
+        return cache.ContainsKey(type);
+    }
+
+    /// <summary>
+    /// Returns true if the cache contains an entry for t and sets metadata to the resulting value,
```
rolfbjarne:
```suggestion
    /// Returns true if the cache contains an entry for type and sets metadata to the resulting value,
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+        else
+        {
+            metadata = null;
+            return false;
+        }
+    }
+
+    /// <summary>
+    /// Gets the TypeMetadata for the given Type t or if it is not present,
```
rolfbjarne:
```suggestion
    /// Gets the TypeMetadata for the given Type type or if it is not present,
```

src/Swift.Runtime/src/Metadata/TypeMetadataCache.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// adds it to the cache using the given factory to generate the value.
+    /// </summary>
+    /// <param name="type"></param>
+    /// <param name="metadataFactory"></param>
+    /// <returns>The TypeMetadata associated with the give Type t</returns>
```
rolfbjarne:
```suggestion
    /// <returns>The TypeMetadata associated with the give Type type</returns>
```

rolfbjarne (approved):

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    /// <param name="type">The type to look up in the cache</param>
+    /// <returns>true if and only if the cache contains t, false otherwise</returns>
+    bool Contains(Type type);
```
vitek-karas: What is the intended usage of this? Give that we're introducing a public API here, we should use a pretty high bar on its necessity and quality.

vitek-karas:
Some ideas (please feel free to disregard):

Do we need the `ITypeMetadataCache` - I know that it's a good design for testability, but if we expect that this is going to be on a hot path, it effectively forces an interface call (the most expensive type of call outside of GVMs) - while in reality we will ever have only one implementation in the product.

Since the cache is global, it needs to be synchronized - adding perf cost. Given that we always go from C# type to Swift metadata (one direction only), and given that in most cases the metadata is created/known by the C# type, can we simply ask the C# type instead? It could have a static field which is lazy inited to store the metadata. (Also, if the Swift metadata token/pointer is static itself for any given type, the init can be done without initialization as well and read access can be completely lock free). The main motivation is that one of the most common cases for the cache usage will be:
```C#
class MyProjectedType : ISwiftObject
{
    static TypeMetadata ISwiftObject.GetTypeMetadata()
    {
        return TypeCache.GetOrAdd(typeof(MyProjectedType), ...); // So the code is asking about itself, while it knows everything
    }
}
```

Maybe I'm a victim of premature optimization strategy :-)

stephen-hawley:
> Some ideas (please feel free to disregard):
...

I had considered that and made the decision that it was more valuable to hide the implementation of the cache instead of making it public.

With regards to `GetTypeMetadata` I was thinking about what is going to be either the first or second most common usage. We will need to write a static function which given a C# Type returns the associated Swift TypeMetadata struct. This method is going to look like this:
```csharp
public static bool TryGetTypeMetadataOf(Type type, [NotNullWhen(true)]out TypeMetadata? result) {
    if (Cache.TryGet(type, out var cachedMD)) return md;
    // now and only now do we have to start looking at the type to figure out if it's `ISwiftObject` or something else.
}
```
So yes, we could use a static in each type but it's going to look like this:
```csharp
    static TypeMetadata _metadata = default(TypeMetadata);
    static TypeMetadata ISwiftObject.GetTypeMetadata()
    {
        if (!_metadata.IsValid) {
            _metadata = PInvokes.MetadataAccessor ();
            return TypeMetadata.Cache.GetOrAdd(typeof(MyProjectedType), () => _metadata);
        }
        return _metadata;
    }
```
But now we're storing the metadata twice: once in the cache and once in the object.
You could obviously adopt the rule that `ISwiftObject` based types don't go in the cache to avoid double storage, but now you lose the biggest gain from having the cache in the first place.

So the choice that I made was "minimize storage and maximize the benefit of having the cache".

src/Swift.Runtime/src/Metadata/ITypeMetadataCache.cs
stephen-hawley: I'll think on this. I don't know yet if it will fall into the category of YAGNI (you ain't gonna need it) or not. `GetOrAdd` combines Contains, Get, and Add into one method and `TryGet` combines `Contains` and `Get` into one method, so it might not be needed. I suppose I can take it out and we can always add it later.

vitek-karas:
> You could obviously adopt the rule that ISwiftObject based types don't go in the cache to avoid double storage, but now you lose the biggest gain from having the cache in the first place.

I don't fully understand how there's a loss here, but I trust we've thought through this.

I'm a bit worried about the synchronization requirement of the cache. But I guess we could replace the internal implementation with a lock free lookup - for example https://github.com/dotnet/runtime/blob/main/src/coreclr/tools/Common/TypeSystem/Common/Utilities/LockFreeReaderHashtable.cs - it also uses a factory mechanism, we might need to tweak it to get good memory characteristics (allocating a delegate every time is a bit expensive), but that's not a problem to solve right away.

Issue #2822, opened 2024-12-03
[Swift language features] Implement Swift structs in C#


kotlarmilos:
## Frozen structs
- [x] Project basic frozen structs
- [x] Project struct constructors
  - [x]  #2888
- [x] Project struct instance methods
- [x] Project struct static methods
- [x] Project methods accepting struct as parameter
- [x] Project non-init functions returning new instance of frozen struct

## Non-frozen structs
- [x] Project basic non-frozen structs
- https://github.com/dotnet/runtimelab/pull/2870
- [x] Project struct constructors
  - [x]  #2888
- [x] Project struct instance methods
- [x] Project struct static methods
- [x] Project methods accepting struct as parameter
- [x] #2885
  - [x] Provide MarshalTo, MarshalFrom API #2896

PR #2870, opened 2024-12-11
[Swift language features] Add support for non-frozen structs

jkurdek:
Contributes to #2822

This PR brings some improvements to projecting structs
1. Adds basic projections of non frozen structs
2. Adds usage of runtimes SwiftSelf
3. Adds usage of runtimes SwiftIndirectResult

TODO:
Simplify testing


src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -284,15 +284,11 @@ private List<BaseDecl> CollectDeclarations(IEnumerable<Node> nodes, BaseDecl par
...
-                        (!swiftTypeInfo.ValueWitnessTable->IsNonPOD || !swiftTypeInfo.ValueWitnessTable->IsNonBitwiseTakable))
-                    {
-                        decl = CreateStructDecl(node, parentDecl, moduleDecl);
-                    }
-                    else
-                    {
-                        decl = CreateClassDecl(node, parentDecl, moduleDecl);
-                    }
+                    // TODO: Find better place for this
+                    typeRecord.IsBlittable = !swiftTypeInfo.ValueWitnessTable->IsNonPOD || !swiftTypeInfo.ValueWitnessTable->IsNonBitwiseTakable; //TODO: This is not the full picture.
```
jkurdek, line 288: Deciding on whether a struct should be projected as a class or struct on c# side is more complex than that. We will have to implement a correct algorithm for that. I guess for now the question is where this should live. ABIParsing?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -6,6 +6,98 @@
...
+        }
+
+        /// <summary>
+        /// Emits the code for the specified environment.
+        /// </summary>
+        /// <param name="writer">The IndentedTextWriter instance.</param>
+        /// <param name="env">The environment.</param>
+        /// <param name="conductor">The conductor instance.</param>
+        /// <param name="typeDatabase">The type database instance.</param>
+        public void Emit(IndentedTextWriter writer, IEnvironment env, Conductor conductor, TypeDatabase typeDatabase)
```
jkurdek: TypeDatabase will be most likely used in majority of the handlers. Maybe it would make sense to inject it into every handler during their construction

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
stephen-hawley: Since it's not part of the actual parsing, my inclination is no.
Measuring the size, stride, and alignment and determining blittability should be a post processing step and should end up in the type database entry for the type.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
stephen-hawley: It should live in the environment, I suspect.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -6,6 +6,98 @@
...
+            var methodDecl = (MethodDecl)methodEnv.MethodDecl;
+            var parentDecl = methodDecl.ParentDecl ?? throw new ArgumentNullException(nameof(methodDecl.ParentDecl));
+            writer.WriteLine($"public {parentDecl.Name}({SignatureHandler.GetCSWrapperSignature(methodDecl)})");
+
+            writer.WriteLine("{");
+            writer.Indent++;
+
+            string PInvokeName = $"{methodEnv.PInvokePrefix}{methodDecl.Name}";
+
+            var pInvokeSignature = SignatureHandler.GetPinvokeSignature(methodDecl, typeDatabase);
```
stephen-hawley: As an FYI, this will generate a C# compile error if the type is generic.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -6,6 +6,98 @@
...
+        private static void EmitWrapper(IndentedTextWriter writer, MethodEnvironment methodEnv, TypeDatabase typeDatabase)
+        {
+            var methodDecl = (MethodDecl)methodEnv.MethodDecl;
+            var parentDecl = methodDecl.ParentDecl ?? throw new ArgumentNullException(nameof(methodDecl.ParentDecl));
+            writer.WriteLine($"public {parentDecl.Name}({SignatureHandler.GetCSWrapperSignature(methodDecl)})");
+
+            writer.WriteLine("{");
+            writer.Indent++;
+
+            string PInvokeName = $"{methodEnv.PInvokePrefix}{methodDecl.Name}";
```
stephen-hawley: Naming methods is going to be very common. I'd like to see naming centralized.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -6,6 +6,98 @@
...
+        /// Emits the wrapper method declaration.
+        /// </summary>
+        /// <param name="writer">The IndentedTextWriter instance.</param>
+        /// <param name="methodEnv">The method environment.</param>
+        /// <param name="typeDatabase">The type database instance.</param>
+        private static void EmitWrapper(IndentedTextWriter writer, MethodEnvironment methodEnv, TypeDatabase typeDatabase)
+        {
+            var methodDecl = (MethodDecl)methodEnv.MethodDecl;
+            var parentDecl = methodDecl.ParentDecl ?? throw new ArgumentNullException(nameof(methodDecl.ParentDecl));
+            writer.WriteLine($"public {parentDecl.Name}({SignatureHandler.GetCSWrapperSignature(methodDecl)})");
```
stephen-hawley: What will happen on a failable constructor?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -58,167 +150,243 @@ public IEnvironment Marshal(BaseDecl methodDecl)
...
-        /// <param name="methodDecl">The method declaration.</param>
-        /// <returns>The list of method parameters.</returns>
-        private List<ArgumentDecl> GetMethodParams(MethodDecl methodDecl)
+        public void HandleArguments()
         {
-            var parentDecl = methodDecl.ParentDecl ?? throw new ArgumentNullException(nameof(methodDecl.ParentDecl));
-            List<ArgumentDecl> tempDecl = new(methodDecl.CSSignature);
+            foreach (var argument in MethodDecl.CSSignature.Skip(1))
+            {
+                var typeRecord = TypeDatabase.Registrar.GetType(argument);
```
stephen-hawley: This should be a TryGet pattern since the type may not exist. In addition, if the type database entry comes from an external module, you will either need to use the full type name of inject a using statement.

stephen-hawley (changes requested):
This is a great start. I will come back and look at this more closely later. I wanted to provide immediate feedback so you could start thinking about changes you might need.

.gitignore
```diff
@@ -34,6 +34,7 @@ msbuild.wrn
 .libs/
 *.lo
 *.o
+build/
```
kotlarmilos, line 37: When `build` gets generated?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: Is `TypeDatabase` a singleton?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: @jkurdek Could you add this in the tracking issue for structs?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: I agree, especially with PInvoke methods (i.e. multiple constructors)

src/Swift.Bindings/src/Marshaler/MarshallingExtensions.cs
```diff
@@ -0,0 +1,46 @@
+
```
kotlarmilos:
```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.
```

src/Swift.Bindings/src/Marshaler/MarshallingExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using Swift.Runtime;
+
+namespace BindingsGeneration
+{
+    public static class MarshallingExtensions // TODO: Find better place for those
```
kotlarmilos: Declarations contain Swift ABI information, and shouldn't include these properties. As we don't have object model for C# (yet), could we use them as static methods until we implement more language features and identify a good unit for them?

src/Swift.Bindings/src/Marshaler/MarshallingExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            return true;
+        }
+
+        public static bool IsMarshalledAsStruct(this StructDecl decl)
+        {
+            return decl is StructDecl structDecl && structDecl.IsFrozen && structDecl.IsBlittable;
+        }
+
+        public static TypeRecord GetType(this TypeRegistrar typeRegistrar, ArgumentDecl argumentDecl)
```
kotlarmilos: Ideally, this should be part of type database. I think we should create a clear type database interface as it will be used across all components of the tooling.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
```diff
@@ -126,6 +126,146 @@ private unsafe bool VerifyFieldRecord(SwiftTypeInfo swiftTypeInfo, int fieldInde
...
+            writer.WriteLine();
+            writer.WriteLine("private static nuint _payloadSize = Metadata.Size;");
+            writer.WriteLine("private SwiftHandle _payload;");
+            writer.WriteLine("private bool _disposed = false;");
+        }
+
+        /// <summary>
+        /// Writes the Dispose method for the class.
+        /// </summary>
+        private static void WriteDisposeMethod(IndentedTextWriter writer, string className)
```
kotlarmilos, line 228: Could we simplify this with a single string constant?

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -284,15 +284,11 @@ private List<BaseDecl> CollectDeclarations(IEnumerable<Node> nodes, BaseDecl par
                     metadataPtr = DynamicLibraryLoader.invoke(_dylibPath, GetMetadataAccessor(node));
```
kotlarmilos, line 284: @stephen-hawley Is it safe to call into dylib at binding time?

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
kotlarmilos: I agree that metadata info should live in the type database. I suggest to retrieve it using MachO reader, but I think it is important to have it at parsing time. Do we want to read .dylib before the ABI parsing to consume information for projections from type database?

src/Swift.Runtime/src/TypeDatabase.xml
```diff
@@ -2,46 +2,46 @@
 <swifttypedatabase version="1.0">
```
kotlarmilos, line 2: Do we plan to use this format for type database snapshots? Do we need more information here?

I would prefer not to create a specific format if all the information can be included in the generated managed bindings.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
vitek-karas: We should always use fully qualified names - to avoid unintentional collisions. In fact we should use the `global:` prefix as well to make sure we don't pick up some other random namespace item. Just like Roslyn source generator.

src/Swift.Bindings/tests/NonFrozenStructs/NonFrozenStructsTests.cs
```diff
@@ -0,0 +1,169 @@
...
+                InitializeResources();
+            }
+
+            private static void InitializeResources()
+            {
+                BindingsGenerator.GenerateBindings("NonFrozenStructs/NonFrozenStructsTests.abi.json", "NonFrozenStructs/");
+                _assemblyPath = TestsHelper.Compile(
+                new string[] { "NonFrozenStructs/*.cs" },
+                new string[] { },
+                new string[] { });
```
vitek-karas: Nit: Indentation

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
jkurdek: I agree that this information should end up in `TypeDatabase`. But I think it also needs to be present on a `TypeDecl`, otherwise  handling each `TypeDecl`  would require reading `TypeDatabase` entry.

>Do we want to read .dylib before the ABI parsing to consume information for projections from type database?

I am not sure I understand

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: I don't see any reason to not have `TypeDatabase` as singleton. Ideally it should be immutable.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: Yeah. I think this should be fine for now. I would rather solve that in a follow up. I will open a tracking issue. I think that the `GetPinvokeSignature` methods should be very easy to extend for generics in the future.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: Thanks for pointing this out! I will create an issue and solve it in a smaller PR in the future - dont want this to grow larger than necessary.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: @stephen-hawley, I'm thinking about the behaviour on TryGet method failure.  Should we just skip this methods projection? Or we should risk getting it wrong?

src/Swift.Runtime/src/TypeDatabase.xml
jkurdek: I think we need to encode the information about blittability and frozeness in the TypeDatabase snapshot. We need to know whether a type coming from a different module is frozen or not.

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
kotlarmilos: Populate type database using MachO reader before ABI parsing, so metadata information can be consumed without calling into dylib from `HandleTypeDecl`.

src/Swift.Runtime/src/TypeDatabase.xml
kotlarmilos: We need to define format of the snapshot and let's explore if we can emit all information needed in the generated assembly. I will add it to the tracking issue, out of scope for this PR.

src/Swift.Bindings/src/Marshaler/MarshallingExtensions.cs
jkurdek: I thought about it some more. I do get your argument, but on the other decision how we project something will depend strictly on Swift ABI information present in Decls. 

Maybe those should operate on environment? I could imagine a functions set operating on e.g. "MethodEnvironment" and answering questions about the underlying method.

src/Swift.Bindings/src/Marshaler/MarshallingExtensions.cs
jkurdek: I agree. Currently type database accepts queries for types using module name and type name. I believe a lot of the queries will be driven by looking at some typeSpec, so maybe we can include that in the interface

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
stephen-hawley: I was looking into how to do this and it's a little more complicated than you think. The problem is that the value witness table for any particular type might not live in the dylib that you're reading. For example, if I define a swift type like this:
```swift
public struct NoInThisDylib {
    public let contents: Int
    public init(with: Int) {
        contents = with
    }
}
```
Then the value witness table for this type is an external reference to the value witness table for Int64 from libswiftCore.dylib. IIRC all heap allocated types (classes) share the same value witness table.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
stephen-hawley: In BTfS for this class of failure it was treated as a warning and a skip for the method.

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
stephen-hawley: No it is not. Loading a library can execute code and it may fail because we're binding a different platform.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: I think there is (at least) two types of naming we need to handle:
1. Naming things projected from Swift - some things in Swift are unpronouncable (e.g. emoji names) in csharp. We need to do the translation.
2. Generating names for extra things we need - pinvokes, proxies etc.

When talking about centralizing naming do we want to have one thing doing 1. and 2. ? 

Edit: Number one could probably be done at parsing stage and encoded in the `TypeDatabase`. So this should leave only number 2

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: I think in order to properly skip the method we need to check all the conditions before we write any code. (this is particularly important if code generation is non-linear - If I encounter problem writing the pinvoke I want to cancel writing the wrapper as well)

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: > Naming things projected from Swift - some things in Swift are unpronouncable (e.g. emoji names) in csharp. We need to do the translation.

I suggest creating a tracking issue and deferring this to a later stage. Do we want to support operator overloading?

stephen-hawley: There's a third case of naming, which is local variables used in marshaling.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
jkurdek: So this is non-trivial as by default IndentedTextWriter indents only the first line of string literal. So if we want to not mess up the indentation in generated code we need to be careful. We will probably need some helper methods for emition. Lets address this in a follow up

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: Added `TypeDatabase` to env

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: https://github.com/dotnet/runtimelab/issues/2889

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: https://github.com/dotnet/runtimelab/issues/2888

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: https://github.com/dotnet/runtimelab/issues/2887

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
vitek-karas: We could do this by creating something like "transactions" - when we're about to start projecting a method, we "start a transaction" which will reset all output buffers. Then we project/write and everything goes into output buffers, only when we're successfully done, we "commit" and appends the buffers to the actual output.
This would be more robust than trying to check for every error condition upfront - it's easy to miss something, and that doesn't count actual unexpected exceptions during the writing process (our bugs and so on). It would be really nice if the tool was resilient to its own bugs to some degree.

jkurdek: > if the type database entry comes from an external module, you will either need to use the full type name of inject a using statement.

I agree. I think we should add some mechanism which allows us to resolve a fully qualified c# name from any `Decl` and any TypeDatabase entry. This should probably be bundled with the translation of unpronouncable names is csharp. So lets track it together. Out of scope of this PR. https://github.com/dotnet/runtimelab/issues/2889

>This should be a TryGet pattern since the type may not exist. 

I like the "transactional" idea. Again I think it will require some separate discussion. Lets do it separately. Tracking issue: https://github.com/dotnet/runtimelab/issues/2890.

.gitignore
jkurdek: It is on me. My `CMake` extension picks up `CMakeLists.txt` and builds it as a separate project. Can revert this

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
jkurdek: https://github.com/dotnet/runtimelab/issues/2891

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
jkurdek: I created the following issues to track the outcomes of the discussion:
https://github.com/dotnet/runtimelab/issues/2892
https://github.com/dotnet/runtimelab/issues/2886

src/Swift.Bindings/src/Marshaler/MarshallingExtensions.cs
stephen-hawley: Here's something to put a pin in: Swift is simple in that there is only one namespace: the module name. C# not so much. In either case, it might be handy to have something like this:
```
public class ClassName { // implement IEquatable<> etc.
   public ClassName(string namespace, string name)
   {
        Namespace = namespace;
        Name = name;
        FullName = Namespace + "." + Name;
   }
   public string Namespace { get; init; }
   public string Name { get; init; } // may include inner classes as a path
   public string FullName { get; init }
}
```
And use this wherever you want to refer to either a C# class or a Swift class.
When writing code non-linearly, we can do `namespaces.AddIfNotPresent(clname.Namespace)` or `imports.AddIfNotPresent(clname.Namespace)`.

stephen-hawley (approved):

kotlarmilos (approved):
LGTM! Thanks!

Issue #2876, opened 2024-12-12
[SwiftBindings] Add a Swift runtime exception


stephen-hawley:
Our SwiftRuntime is going to need to throw exceptions that happen when failures are detected. For example, if the user tries to marshal a type into a generic Swift function and that type doesn't have an associated Swift TypeMetadata object, this is a failure.

PR #2877, opened 2024-12-12
[Swift runtime features] Added accessors for TypeMetadata

stephen-hawley:
As we've discussed in previous meetings, we need accessors for TypeMetadata. To do this, I've first created the interface `ISwiftObject` which contains an abstract method getting the `TypeMetadata` for the object. 

Then I created public accessors that are either in the form `TryGet` or `Get..OrThrow`. Each of these accessors first check the cache for the entry. Upon failure, the code defers to an uncached version which checks to see if the object implements `ISwiftObject` and if so, retrieves the `TypeMetadata` from the object.

I then added type metadata entries for common scalar types.

Finally I added a suite of unit tests for the main entry points and for each of the scalars.

Note that the code has some TODOs with issues linked to them.
Note also that when I wasn't sure what to throw, I threw Exception as a placeholder and opened an issue for that https://github.com/dotnet/runtimelab/issues/2876

Some tests were failing because they didn't see TypeMetadata, so I adjusted TestsHelper for that.

Since I needed access to libswiftCore.dylib from TypeMetadata, rather than define it as a const string twice, I put made an internal KnownLibraries class for that and refactored the code to use it. I don't know if we'll ever have more than libswiftCore in there, but if we do, at least we have a place for it.


src/Swift.Runtime/src/Library/ISwiftObject.cs
```diff
@@ -0,0 +1,16 @@
...
+/// <summary>
+/// Represents a type that can be marshaled into Swift
+/// </summary>
+public interface ISwiftObject
+{
+    /// <summary>
+    /// Returns the TypeMetadata for this object
+    /// </summary>
+    /// <returns>A type metadata object for the type.</returns>
+    static abstract TypeMetadata GetTypeMetadata();
```
rolfbjarne: This is in an interface, so contrary to everywhere else, visibility defaults to `public`, which is a bit confusing (especially if a default implementation is provided). Thus maybe make it explicit:

```suggestion
    public static abstract TypeMetadata GetTypeMetadata();
```

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -249,4 +253,125 @@ public override int GetHashCode ()
...
+    /// <param name="obj">An object to look up</param>
+    /// <param name="result">The result of looked up type metadata</param>
+    /// <returns>true on success false otherwise</returns>
+    public static bool TryGetTypeMetadata<T>(T obj, [NotNullWhen(true)] out TypeMetadata? result)
+    {
+        return TryGetTypeMetadata(typeof(T), out result);
+    }
+
+    /// <summary>
+    /// Attemp to get the Swift type metadata for the given type
```
rolfbjarne:
```suggestion
    /// Attempt to get the Swift type metadata for the given type
```

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -249,4 +253,125 @@ public override int GetHashCode ()
...
\ No newline at end of file
+
+    /// <summary>
+    /// Attempt to get the Swift type metadata for the given object instance
+    /// </summary>
+    /// <typeparam name="T">The type of the object</typeparam>
+    /// <param name="obj">An object to look up</param>
+    /// <param name="result">The result of looked up type metadata</param>
+    /// <returns>true on success false otherwise</returns>
+    public static bool TryGetTypeMetadata<T>(T obj, [NotNullWhen(true)] out TypeMetadata? result)
```
rolfbjarne: `obj` is never used?

```suggestion
    public static bool TryGetTypeMetadata<T>([NotNullWhen(true)] out TypeMetadata? result)
```

rolfbjarne (approved):

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -249,4 +253,125 @@ public override int GetHashCode ()
...
+    }
+
+    /// <summary>
+    /// Attempt to get the Swift type metadata but without accessing the cache
+    /// </summary>
+    /// <param name="type">A type to look up</param>
+    /// <param name="result">The result of looked up type metadata</param>
+    /// <returns>true on success false otherwise</returns>
+    /// <exception cref="NotImplementedException">Throws if unable to look up the ISwiftObject.GetTypeMetadata method.</exception>
+    static bool TryGetTypeMetadataUncached(Type type, [NotNullWhen(true)] out TypeMetadata? result)
```
jkurdek: I think this is not trimming / NativeAOT friendly and brings us back to the discussion we had in https://github.com/dotnet/runtimelab/pull/2771. I would recommend going with the pattern described in binding-generics. 

cc: @MichalStrehovsky, @vitek-karas

jkurdek (changes requested):

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
MichalStrehovsky: We should add `<IsAotCompatible>true</IsAotCompatible>` to the project file (that will also enable IsTrimmable, etc.) so that problematic code gets surfaced as warnings.

src/Swift.Runtime/src/Library/KnownLibraries.cs
```diff
@@ -0,0 +1,7 @@
+
+namespace Swift.Runtime;
+
+internal static class KnownLibraries {
+    public const string SwiftCore = "/usr/lib/swift/libswiftCore.dylib";
```
kotlarmilos, line 5: Is this the default path? For example, I have this library at `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-5.0/macosx/libswiftCore.dylib`.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -249,4 +253,125 @@ public override int GetHashCode ()
     /// Gets the type metadata cache for the runtime.
     /// </summary>
     public static ITypeMetadataCache Cache => cache;
```
kotlarmilos, line 255: This is more of a lookup than a cache, as it doesn't invalidate. Consider changing the name to avoid misleading interpretation.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
kotlarmilos: https://github.com/dotnet/runtimelab/pull/2880

stephen-hawley: I need some help here - what do you want me to do with this?
I can remove the reflection code and make a static initializer for the type that forces it into the cache.

src/Swift.Runtime/src/Library/KnownLibraries.cs
stephen-hawley: That is the path that should be consistent across all devices. The one in Xcode is there for developers to use as a preview in case their OS has a different version installed. It used to be that libswiftCore was included in every app, but around about swift 4 or 5 Apple made the guarantee that it would be present everywhere (yay for enable-library-evolution)

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
stephen-hawley: It's technically Memoization which is a form of caching, but is it really that confusing?

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
jkurdek: Well the simplest "fix" would be to make this method generic - that means that `TryGetTypeMetadata` will need to be generic as well and use typeof(T) to populate cache.

Then we would need something like
```csharp
public static class Helper<T> where T : ISwiftObject 
{
    public static TypeMetadata GetTypeMetadata()
    {
        return T. GetTypeMetadata();
    }
}
```

and in the body of the method you'd rewrite the reflection call to:
```csharp
{
    var helperType = typeof(Helper<>).MakeGenericType(typeof(T));
    result = (TypeMetadata)helperType.GetMethod("GetTypeMetadata")!.Invoke(null, null)!;
    return true;
}
```

This would make the trimming errors disappear.

I agree that this is not the prettiest solution, it is probably quite slow as well - but we dont have anything better now. Maybe in the future we could push the csharp lang team to give as `type is someOtherType` pattern in c#. 

Alternative would be to make the MetadataAccessors non-static and use `x is ISwiftObject` pattern. But this will cause a lot of weirdness when we need to get metadata in order to create objects instance.

stephen-hawley: Yeah, unfortunately, one of the most common cases of getting a type's metadata is when we have no instance.

jkurdek (approved):

AaronRobinsonMSFT:
If this is going to be a public API we will eventually need an API review. I think we're good for now as we experiment, but any API that set to be exposed in a public and stable environment must go through [API review](https://github.com/dotnet/runtime/blob/main/docs/project/api-review-process.md).

src/Swift.Runtime/src/Library/Arc.cs
```diff
@@ -10,13 +10,11 @@ namespace Swift.Runtime;
...
 /// </summary>
 public static class Arc {
-    const string swiftCore = "/usr/lib/swift/libswiftCore.dylib";
-
     /// <summary>
     /// Retain a heap-allocated Swift object
     /// </summary>
     /// <param name="p">Pointer to an unmanaged Swift object, must be non-null.</param>
-    [DllImport(swiftCore, CallingConvention = CallingConvention.Cdecl)]
+    [DllImport(KnownLibraries.SwiftCore, CallingConvention = CallingConvention.Cdecl)]
```
AaronRobinsonMSFT, line 17: Nit: These should all be `LibraryImport`.

src/Swift.Runtime/src/Library/KnownLibraries.cs
```diff
@@ -0,0 +1,7 @@
+
```
AaronRobinsonMSFT, line 1: License banner.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -249,4 +253,113 @@ public override int GetHashCode ()
...
+        return TryGetTypeMetadataUncached<T>(out result);
+    }
+
+    /// <summary>
+    /// Attempt to get the Swift type metadata for the given type
+    /// </summary>
+    /// <typeparam name="T">The type of the object</typeparam>
+    /// <returns>The result of the looked up type metadata on success</returns>
+    /// <exception cref="Exception">Throws when lookup fails</exception>
+    public static TypeMetadata GetTypeMetadataOrThrow<T>()
```
AaronRobinsonMSFT, line 276: I would drop the `OrThrow`. The default semantic for functions in .NET is that they can throw. The `Try` semantics above indicate the non-throwing nature.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -249,4 +253,113 @@ public override int GetHashCode ()
...
+    /// Returns an enumeration of known Type and TypeMetadata objects
+    /// </summary>
+    /// <returns>An enumeration of known Type and TypeMetadata objects</returns>
+    /// <exception cref="Exception">Throws if unable to load a library containing known types</exception>
+    static IEnumerable<(Type, TypeMetadata)> KnownMetadata()
+    {
+        var libraryHandle = NativeLibrary.Load(KnownLibraries.SwiftCore);
+        if (libraryHandle == IntPtr.Zero)
+            throw new Exception(string.Format("Unable to load library {0}", KnownLibraries.SwiftCore));
+        // types from libSwiftCore
```
AaronRobinsonMSFT, line 330: All these mangled names should likely be consolidated into a single place and documented on how to retrieve them from `libSwiftCore`.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
```diff
@@ -249,4 +253,113 @@ public override int GetHashCode ()
...
+    /// Loads type metadata from a NativeLibrary using the supplied symbol
+    /// </summary>
+    /// <param name="handle">handle to a library loaded by NativeLibrary</param>
+    /// <param name="symbolName">Swift symbol for a type metadata object</param>
+    /// <param name="libraryName">The library to load from. Defaults to libswiftCore.dylib</param>
+    /// <returns>A type metadata object for the symbol</returns>
+    /// <exception cref="Exception">Throws on failure to load symbol</exception>
+    static TypeMetadata MetadataFromNativeLibrary(IntPtr handle, string symbolName, string libraryName = KnownLibraries.SwiftCore)
+    {
+        var entryPoint = NativeLibrary.GetExport(handle, symbolName);
```
AaronRobinsonMSFT, line 360: [`NativeLibrary.GetExport()`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.nativelibrary.getexport#system-runtime-interopservices-nativelibrary-getexport(system-intptr-system-string)) throws if the export can't be found. Instead `TryGetExport()` should be used if a specific exception is desired.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
kotlarmilos: Shouldn't these be retrieved from type database snapshot, like https://github.com/dotnet/runtimelab/blob/feature/swift-bindings/src/Swift.Runtime/src/TypeDatabase.xml?

src/Swift.Bindings/tests/TestsHelper.cs
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) Microsoft Corporation.
```
kotlarmilos, line 1: Please remove this file as it is not used anymore.

kotlarmilos (approved):
LGTM!

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
AaronRobinsonMSFT: @stephen-hawley Please address this comment. The `NativeLibrary.GetExport()` is being used incorrectly.

AaronRobinsonMSFT:
@stephen-hawley It doesn't appear any of my feedback was addressed. Was there a specific reason for this?

stephen-hawley:
@AaronRobinsonMSFT - apologies.

src/Swift.Runtime/src/Metadata/TypeMetadata.cs
stephen-hawley: I understand. There is a `TryGet` variant that can be used, but this code pattern is going to get used a lot:
```
if (TryGetTypeMetadata<T>(out var metadata)) {
    // do something
} else {
     throw new SomeException("");
}
```
There are two issues that I'm addressing with this:
1. The `else` block has been identical in all cases and this is a local factoring
2. More importantly, we have had several instances of code that I have flagged in PRs that were missing the else case entirely, which is wrong. This provides a tool to make it harder to make that mistake and it also streamlines the code, removing the if/else entirely.

PR #2879, opened 2024-12-13
[Swift bindings] Upgrade dotnet runtime version to 9.0.0

kotlarmilos:
## Description

This PR upgrades the runtime to version 9.0.0, along with additional changes from arcade.

The `./build.sh` script installs the new version of dotnet locally in the `.dotnet` directory. When running `dotnet test`, it doesn't automatically detect the local version. To resolve this, add the directory to the path: `export PATH="./.dotnet:$PATH"`.

### DIFF
--- a/eng/common/native/install-dependencies.sh
+++ b/eng/common/native/install-dependencies.sh
@@ -44,15 +44,11 @@ case "$os" in
         export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
         # Skip brew update for now, see https://github.com/actions/setup-python/issues/577
         # brew update --preinstall
-
-        # Temporarily uninstall pkg-config@0.29.2 to work around https://github.com/actions/runner-images/issues/10984
-        brew uninstall --ignore-dependencies --force pkg-config@0.29.2
-
         brew bundle --no-upgrade --no-lock --file=- <<EOF
 brew "cmake"
 brew "icu4c"
 brew "openssl@3"
-brew "pkg-config"
+brew "pkgconf"
 brew "python3"
 brew "pigz"
 EOF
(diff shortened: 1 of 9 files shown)


eng/common/native/install-dependencies.sh
```diff
@@ -0,0 +1,61 @@
...
+
+    osx|maccatalyst|ios|iossimulator|tvos|tvossimulator)
+        echo "Installed xcode version: $(xcode-select -p)"
+
+        export HOMEBREW_NO_INSTALL_CLEANUP=1
+        export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+        # Skip brew update for now, see https://github.com/actions/setup-python/issues/577
+        # brew update --preinstall
+        brew bundle --no-upgrade --no-lock --file=- <<EOF
+brew "cmake"
```
stephen-hawley, line 48: You might want to talk to @mandel-macaque about this. I don't know the current state of things, but my understanding is that there are some issues around using `brew`

stephen-hawley (approved):

eng/common/native/install-dependencies.sh
kotlarmilos: This is an arcade change and I don't think our CI relies on it.

jkurdek:
Why did this bump the .NET SDK version to .NET 10? was this necessary?

kotlarmilos:
It is not necessary, but it is nice to have. Transitioning between different .NET versions in BTfS was a non-trivial task. Now, we can avoid this by performing periodic upgrades. The arcade changes are not automated on this branch, so we manually sync it with our base branch ([standalone-template](https://github.com/dotnet/runtimelab/blob/standalone-template/global.json)).

jkurdek:
I understand that this is nice to have, however I'm not sure whether it is worth it. For example bumping SDK version to .NET 10 broke my vscode dev kit, making local development a bit of a pain. Have you tested this scenario before the the bump?

kotlarmilos:
I haven't tested it with vscode devkit. Is that a known issue in .NET 10? Our CI pipeline includes tests, but it does not include the devkit.

kotlarmilos:
This was a vscode issue similar to https://github.com/microsoft/vscode-dotnettools/issues/808.

Going forward, we will test such changes on multiple contributors' machines before merging them.

PR #2878, opened 2024-12-13
[Swift bindings] Add bindings experimental NuGet to the dotnet-experimental feed

kotlarmilos:
## Description

This PR adds binding generation step to the distribution pipeline. To enable testing of the StoreKit framework, we want to provide an experimental NuGet that can be used in real-world applications such as HikingApp. This approach will also allow us to track binding coverage and identify areas for improvement. 

### Changes

This change introduces a script for generating bindings that runs after the build in the distribution pipeline. The script extracts the ABI for selected frameworks, runs the tooling, and generates NuGet package from source code. The package is then distributed using `darc` in the `dotnet-experimental` feed.  

### Verification

The pipeline was invoked manually. The NuGet package will become available once this change is merged into the `feature/swift-bindings` branch, which is configured for feed publishing.

### Follow-up

Once the package becomes available, we will start introducing framework tests as MacCatalyst apps and track coverage of supported Swift language features from included frameworks.

PR #2880, opened 2024-12-13
[Swift bindings] Make projection tooling AOT compatible

kotlarmilos:
## Description

This PR sets the `IsAotCompatible` property and updates the code to be trim-compatible. This ensures that future changes will implement trim-compatible code.


src/Swift.Bindings/src/Marshaler/Conductor.cs
```diff
@@ -17,22 +17,24 @@ namespace BindingsGeneration
     /// It initializes the handler factories and provides methods to fetch handlers for given declarations.
     /// </summary>
     public class Conductor {
-        private readonly List<IModuleHandlerFactory> _moduleHandlerFactories = [];
-        private readonly List<ITypeHandlerFactory> _typeHandlerFactories = [];
+        private readonly List<IModuleHandlerFactory> _moduleHandlerFactories = [
```
jkurdek, line 20: Im not sure those were necessary. Do we also require the tooling to be trim-friendly / native-aot friendly? I thought that the NativeAOT support requirement is only specified on the output bindings / Swift.Runtime code.

But on the other hand I think it simplifies the design and takes some magic away.

kotlarmilos: Not necessary, but nice to have. It is easy to set requirements now as the codebase is still small. @stephen-hawley What do you think?

kotlarmilos: Added AOT compatible property to tests (https://github.com/dotnet/runtimelab/pull/2871/commits/dcf3fdb7c75cf7b44f67463dd28dacb0bc448265) and experimental NuGet (https://github.com/dotnet/runtimelab/pull/2878/commits/ae322434816a84a4121071603c0ced47d47e9604).

stephen-hawley: I don't see it as a requirement for the tooling, no.

jkotas: I agree that it is not a strict requirement, but it is very nice to have. Native AOT makes the tools like compilers a lot faster. We compile number of shipping tools (e.g. the AOT compiler itself) using native AOT for this reason.

jkurdek (approved):

stephen-hawley (approved):

PR #2883, opened 2024-12-16
[NativeAOT-LLVM] Emit type debug info into a single module

### LINKED ISSUE #2728: [NativeAOT-LLVM] Debug info generation does not define types properly
The reproduction for this problem is to run the `WasmDebugging.csrpoj -c Debug` test under wasmtime+LLDB:
```
> (lldb) b Main
> r
> type lookup WasmDebugging_Program
```
The output will be:
```
struct WasmDebugging_Program {
    static int Main();
}
```
This is NOT expected, since we have a number of other static methods hanging off of this type. And indeed, if we try to step into them, it will work, but things that rely on `.debug_info` DIEs more directly, like expression evaluation, will not work. I suspect this problem is also why we often see "debugging information not available" "red" frames in the Chrome debugger.

The root cause is that the way we can duplicate type definitions across bitcode files is not supported by the system. In this case in particular, `Main` happened to have been compiled into `HelloWasm.X.bc`, while most other methods - into some other module. This can be observed directly via `wasm-tools unbundle WasmDebugging.wasm > $null && llvm-dwarfdump -F unbundled-module0.wasm | sls 'DW_AT_name [DW_FORM_strp]	("WasmDebugging_Program")' -s`.

The end result is that LLDB caches the lookup results from the `HelloWasm.X` compile unit, where `Main` was recorded, and doesn't realize there are other parts present.

The fix here should be relatively straightforward.

Edit: further investigation with C++ classes revealed that a similar pattern is not handled by the wasmtime DWARF translator either, so we may have two bugs at hand.

SingleAccretion:
So far, we have been emitting type definitions into each compile unit
where they were 'used'. This proved problematic (wrong) as the debugger
does not "unify" types across compile units - they are expected to be
the same, due to ODR.

To solve this fundamental issue, this change introduces a more elaborate emission scheme, where during compilation only type forwards are emitted, which will later get satisfied by types defined in one big DI module.

While the idea is sound, the implementation is constrained by the lack of direct support for DW_FORM_ref_addr references in LLVM. Ideally, we would use those to refer to types in the debug module directly, however, in the absence of this, forward declarations have to do.

One problem with forward declarations that had to be worked around was wasmtime's DI GC, which only resolves DWARF's 'reference's. Thus, our types need manual 'rooting', which is accomplished in this change by emitting a dummy retained subroutine that references all of them.

With these changes, as well as some fixes in wasmtime (especially https://github.com/bytecodealliance/wasmtime/issues/9828), the expanded LLDB (finally) test passes.

Fixes #2728.

### DIFF
--- a/eng/pipelines/runtimelab/install-wasmtime.ps1
+++ b/eng/pipelines/runtimelab/install-wasmtime.ps1
@@ -9,7 +9,9 @@ $ProgressPreference = "SilentlyContinue"
 
 Set-Location $InstallDir
 
-$WasmtimeVersion = "v26.0.1"
+# Currently we need to use the daily releases for the debugging test to work.
+$UsePreRelease = $true
+$WasmtimeVersion = $UsePreRelease ? "dev" : "v26.0.1"
 
 if ($IsWindows)
 {
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -841,26 +841,7 @@ void Llvm::GetJitTestInfo(CorInfoLlvmJitTestKind kind, CORINFO_LLVM_JIT_TEST_INF
     CallEEApi<EEAI_GetJitTestInfo, CORINFO_GENERIC_HANDLE>(m_pEECorInfo, kind, pInfo);
 }
 
-extern "C" DLLEXPORT int registerLlvmCallbacks(void** jitImports, void** jitExports)
-{
-    assert((jitImports != nullptr) && (jitImports[EEAI_Count] == (void*)0x1234));
-    assert(jitExports != nullptr);
-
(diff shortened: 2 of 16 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks, looks all good to me.  Will be interesting to try.  One question, in a release build does NAOT in general give the same experience as JIT ?

SingleAccretion:
>  One question, in a release build does NAOT in general give the same experience as JIT ?

The Jit is the same so theoretically the coverage of variables / stepping behavior is the same, yes. Everything else is of course different (managed vs native debugging).

In our case, unfortunately, LLVM is really bad about preserving variables, so in release only stepping somewhat works.

jkotas:
Thank you!

Issue #2881, opened 2024-12-13
[SwiftBindings] Extra enum-specific entries in ValueWitnessTable


stephen-hawley:
ValueWitnessFlags needs to be based on uint32 explicitly to ensure that it lines up with the Swift.

We need to add (in order):

/// Given a valid object of this enum type, extracts the tag value indicating
/// which case of the enum is inhabited. Returned values are in the range
/// [0..NumElements-1].
uint getEnumTag(void *enumData, TypeMetadata md);

/// Given a valid object of this enum type, destructively extracts the
/// associated payload.
void destructiveProjectEnumData(void *enumData, TypeMetadata md);

/// Given an enum case tag and a valid object of case's payload type,
/// destructively inserts the tag into the payload. The given tag value
/// must be in the range [-ElementsWithPayload..ElementsWithNoPayload-1].
void destructiveInjectEnumTag(void *enumData, uint tag, TypeMetadata md)

How to use these -
getEnumTag returns the tag for enum.
destructiveProject clears the tag from the enum payload, so if you want to extract a payload without worrying about the tag, you can do this:
```csharp
public SomeEnumProjection {
    SomeType Case0Value {
        get {
            unsafe {
                var metadata = GetTypeMetadata();
                var witnessTable = metadata.ValueWitnessTable;
                if (witnessTable->GetEnumTag() != 0) throw new Exception();
                var copy = stackalloc byte[_payload.Length];
                Array.Copy(_payload, 0, copy, 0, _payload.Length);
                fixed (void *copyPtr = copy) {
                    witnessTable->destructiveProjectEnumData(copyPtr, metadata);
                    return SwiftRuntime.SwiftMarshal.FromSwift<SomeType>((IntPtr)copy);
                }
        }
    }
}
```
Destructive project enum data removes the tag information, letting us pull the data directly from the payload, but because it's destructive, we need to work with a copy. In theory we would reinject the tag, but that creates a potential thread safety issue if there are any mutating methods on the type.

Similarly, this lets us implement factory methods like this:
```csharp
public SomeEnumProjection {
    static SomeEnumProjection CreateCase0(SomeType payload) {
        var someEnum = default(SomeEnumProjection);
        var metadata = GetTypeMetadata();
        var witnessTable = metadata.ValueWitnessTable;
        unsafe {
             fixed (void *payloadPtr = &_payload) {
                 SwiftRuntime.SwiftMarshal.ToSwift(payload, (IntPtr)payloadPtr);
[...]

PR #2893, opened 2024-12-17
[SwiftBindings] Extra enum-specific entries in ValueWitnessTable

stephen-hawley:
This addresses issue https://github.com/dotnet/runtimelab/issues/2881 and enables us to create factory methods for each of the cases and payload accessors without needing to write wrapper code.

This will enable us to bind `Swift.Optional`.

### DIFF
--- a/src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs
+++ b/src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs
@@ -102,6 +102,24 @@ namespace Swift.Runtime
 		public uint ExtraInhabitantCount;
 
 		/// <summary>
+		/// Returns a tag for the given enumeration. The tags start at 0 with the first
+		/// case with a payload and continue in declaration order. Then they are followed by
+		/// cases without payloads in declaration order.
+		/// </summary>
+		public delegate* unmanaged<void *, TypeMetadata, nuint> GetEnumTag;
+
+		/// <summary>
+		/// Strips out the tag information from a given enumeration leaving the value behind.
+		/// This is useful when pulling the payload out of the enum.
+		/// </summary>
+		public delegate* unmanaged<void *, TypeMetadata, void> DestructiveProjectEnumData;
+
+		/// <summary>
+		/// Injects a tag into an enumeration. This is useful when creating an enum.
+		/// </summary>
+		public delegate* unmanaged<void *, nuint, TypeMetadata, void> DestructiveInjectEnumTag;
+
+		/// <summary>
 		/// Returns the alignment of the type in bytes.
 		/// </summary>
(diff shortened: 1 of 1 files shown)


haritha-mohan (approved):

rolfbjarne (approved):

src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs
```diff
@@ -101,6 +101,24 @@ public unsafe ref struct ValueWitnessTable
 		/// </summary>
 		public uint ExtraInhabitantCount;
 
+		/// <summary>
+		/// Returns a tag for the given enumeration. The tags start at 0 with the first
+		/// case with a payload and continue in declaration order. Then they are followed by
+		/// cases without payloads in declaration order.
+		/// </summary>
+		public delegate* unmanaged<void *, TypeMetadata, nuint> GetEnumTag;
```
matouskozak, line 109: Nit. I would move these definitions a few lines above after https://github.com/dotnet/runtimelab/blob/64b5b7f6e9ddee65a74cbc00f3139b5a757f7ccd/src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs#L82 to keep it at one place.

matouskozak (approved):

kotlarmilos (approved):

jkurdek (approved):

src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs
```diff
@@ -101,6 +101,24 @@ public unsafe ref struct ValueWitnessTable
...
+
+		/// <summary>
+		/// Strips out the tag information from a given enumeration leaving the value behind.
+		/// This is useful when pulling the payload out of the enum.
+		/// </summary>
+		public delegate* unmanaged<void *, TypeMetadata, void> DestructiveProjectEnumData;
```
kotlarmilos, line 115: It is a good time to review when we use:
 - void*
 - IntPtr
 - SwiftHandle

dalexsoto (approved):

src/Swift.Runtime/src/Metadata/ValueWitnessTable.cs
stephen-hawley: I can't do that, though. The order is significant since we are taking these directly from memory.

matouskozak: Ah, good point, I haven't realized that. I think we should explicitly mention that we need to keep the order in sync with Swift def file.

Another question... what mechanism will we use to distinguish what parts should be used for which types. Since these 3 added APIs are for Enum only, we need to make sure we are not trying to read them for other types as well.

PR #2898, opened 2024-12-19
[Swift bindings] Enable test signing for experimental NuGet binding package

kotlarmilos:
## Description

Currently, there is a signing issue with NuGet packages in the official build pipeline. This PR enables test signing to fix the pipeline until the issue is fixed, as the experimental NuGet package isn't officially supported.

## Validation

The internal build pipeline completed successfully.

This change will be reverted: https://github.com/dotnet/runtimelab/issues/2897


akoeplinger (approved):
The signing issue should be fixed now btw

PR #2900, opened 2024-12-19
[Swift Bindings] Ensure projection tooling generate bindings that compile for selected frameworks

kotlarmilos:
## Description

This PR:
1. Ensures that the generated bindings for frameworks compile successfully
2. Handles unknown types gracefully
3. Simplifies the projection tooling UX
4. Adds projection tooling requirements

Generated bindings that compile for StoreKit2: https://gist.github.com/kotlarmilos/907cde8ce6af5f2730f8f33e432a503e

## Changes

This PR handles unknown types gracefully by introducing `AnyType` as a Swift placeholder type. Additionally, it simplifying the projection tooling UX by removing unnecessary args and unifies the `Swift` and `Swift.Runtime` namespaces into a single directory 

This PR does not change the layout of projections.

## Validation

Both official and PR pipelines now include a bindings compilation check for frameworks to ensure that generated code can compile.


rolfbjarne (approved):
I think it would be a good idea to zero out `_payload` after freeing it, which would avoid any accidental access of freed memory:

https://gist.github.com/kotlarmilos/907cde8ce6af5f2730f8f33e432a503e#file-swift-storekit-cs-L105

https://gist.github.com/kotlarmilos/907cde8ce6af5f2730f8f33e432a503e#file-swift-storekit-cs-L112

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -325,7 +347,14 @@ public void HandleReturnType()
         {
             if (!MarshallingHelpers.MethodRequiresIndirectResult(MethodDecl, ParentDecl, TypeDatabase))
             {
-                var returnType = MethodDecl.CSSignature.First().CSTypeIdentifier.Name;
+                var argument = MethodDecl.CSSignature.First();
+                var returnType = argument.CSTypeIdentifier.Name;
+                var typeRecord = MarshallingHelpers.GetType(argument, TypeDatabase.Registrar);
+                if (typeRecord == null || !typeRecord.IsProcessed)
```
jkurdek: I dont think this path can execute. `MethodRequiresIndirectResult` will call throw first when typeRecord is not found with the changes to ArgumentIsMarshalledAsCSStruct

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -278,7 +293,14 @@ public Signature Build()
         /// <param name="returnType">The return type.</param>
         private void SetReturnType(string returnType)
         {
-            _returnType = returnType;
+            if (returnType == "()")
```
jkurdek: I think we should document void handling. This can be done separately.

jkurdek (commented):
I think that gracefully handling unknown types will be tricky. The complexity is in the fact that given an argument we need to make an ABI decision (indirect result, opaque pointer, etc). We wont be able to make those decision if we dont find the type in the `TypeDatabase`.

If compilation is our goal for now maybe we can do the following approach. Do `AnyType` subtitution in wrappers signatures - this should work as this does not touch ABI. And then just not generate pinvokes when argument is not known.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -325,7 +347,14 @@ public void HandleReturnType()
         {
             if (!MarshallingHelpers.MethodRequiresIndirectResult(MethodDecl, ParentDecl, TypeDatabase))
             {
-                var returnType = MethodDecl.CSSignature.First().CSTypeIdentifier.Name;
+                var argument = MethodDecl.CSSignature.First();
+                var returnType = argument.CSTypeIdentifier.Name;
+                var typeRecord = MarshallingHelpers.GetType(argument, TypeDatabase.Registrar);
```
jkurdek: I feel that we now operate on two abstraction layers in this call. We have high-level `MethodRequiresIndirectResult` which calls type database and low level `MarshallingHelpers.GetType` which can lead to confusing behaviour as discussed in previous comment.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: There was a bug in type registration, this special handling is not needed.

kotlarmilos:
> If compilation is our goal for now maybe we can do the following approach. Do AnyType subtitution in wrappers signatures - this should work as this does not touch ABI. And then just not generate pinvokes when argument is not known.

I think this is a good start -- it is safe to assume that `AnyType` signatures will not be used. Additionally, the `AnyType` placeholder highlights missing language features that we may want to focus on.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: All types should be registered during ABI parsing, at least for lazy-loading. If a type is processed, it means the tooling will generate a projection type; otherwise, it will emit an "AnyType" placeholder. If a type can't be found in the database, it indicates that it wasn't encountered during ABI parsing in any form, and throwing an exception is fine as it is bug.

jkurdek: But then the 'typeRecord == null' part becomes unnecessary (error will be thrown before reaching it). I think we should not make GetType return nullable if null result means critical failure.

kotlarmilos: I agree.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: I feel that some marshalling should be handled before emitting. However, I would prefer implementing more support before structuring it.

When replacing unsupported types with `AnyType` placeholders, method conflicts might occur (same name and signature). To address this, I completely skipped generating wrappers to ensure compilation succeeds. The next step would be to reduce these methods during marshalling and emit them with a `NotImplementedException` body.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: Reverted. This is how I discovered a type registration issue :)

jkurdek (approved):

kotlarmilos:
Thanks for prompt review. If there is any additional feedback later, we will address it as a follow-up.

PR #2909, opened 2025-01-06
[SwiftBindings] Made AnyType compatible with ISwiftObject

stephen-hawley:
- Moved `SwiftOptional<T>` to the `Swift` namespace.

- Updated the `AnyType` class to be compatible with the `ISwiftObject` interface.

- This change adds the methods `NewFromPayload` and `MarshalToSwift` to `AnyObject`.

Since the default implementation of GetTypeMetadata returns an invalid metadata, the `MarshalToSwift` method throws an exception.

Since the default implementation has a payload of `SwiftHandle.Zero`, MarshalToSwift throws an exception.


src/Swift.Runtime/src/Swift/AnyType.cs
```diff
@@ -21,4 +21,35 @@ public AnyType(SwiftHandle payload)
...
+    /// Marshals this object to a Swift destination
+    /// </summary>
+    /// <param name="swiftDest"></param>
+    /// <returns></returns>
+    IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
+    {
+        var metadata = SwiftObjectHelper<AnyType>.GetTypeMetadata();
+        if (!metadata.IsValid)
+        {
+            throw new InvalidOperationException("Cannot marshal AnyType to Swift without metadata");
```
mcumming, line 43: Should these strings be localized?

mcumming (approved):

src/Swift.Runtime/src/Swift/AnyType.cs
stephen-hawley: Maybe. I put the exceptions in as an afterthought since it looks like this is not an actual valid type because of the lack of a valid TypeMetadata. @kotlarmilos - is `AnyType` a placeholder for the future since it isn't valid?

rolfbjarne (approved):

src/Swift.Runtime/src/Swift/AnyType.cs
kotlarmilos: Yes -- it is safe to assume the type is not valid.

kotlarmilos (approved):

kotlarmilos:
This change breaks the compilation: The type or namespace name 'MetadataRequest' could not be found.

jkurdek (approved):
Lets just get this to compile before merging

Issue #2886, opened 2024-12-17
[Projection tooling] Implement algorithm deciding whether a struct should be projected as class or struct


jkurdek:
Struct should be projected as a struct if:
- Struct needs to be marked as frozen, every member of this struct has to be either trivial or marked as frozen (recursively)
- Structs needs to be POD/Trivial / Bitwise Movable

jkurdek:
I did look into swift code - tried to find a better way to determine whether to project something as a struct or as class. There is a field on `ValueWitnessTable` called `IsNotInline` which seems to be true whenever the struct will be passes as a pointer. [Link](https://github.com/swiftlang/swift/blob/33bff48e927f2ed54f26dacbafad9e1c2f6a5a6e/include/swift/ABI/ValueWitness.def#L220). However this value will also be true for frozen structs which exceed the lowering threshold (4 machine words). Have we looked into that before? 

An alternative is to recursively traverse struct children - this however means that all of struct children have to processed before a projection involving a struct can be made - which means building up the type database of multiple modules before a single projection can be made. 

cc: @jkoritzinsky @stephen-hawley @kotlarmilos @matouskozak

stephen-hawley:
Having a complete type database seems more reliable to me.

kotlarmilos:
> Having a complete type database seems more reliable to me.

+1. We can assume that projection tooling will process modules in a correct order.

We might not be able to read `ValueWitnessTable` if target is not macOS. I suggest that the projection tooling emits a `Debug.Assert` for `IsNotInline` flag to ensures that the projections match the expected lowering, but the ABI file should be treated as the source of truth.

jkurdek:
Great! I then propose the following design:

Assumption: Tooling will process modules in correct order. Processing each module will result in a type database file.

When processing a new module the tooling will first parse the ABI and build a graph of dependencies for each type. Dependencies will go into two categories (same module / other modules).

Then in a post-processing step the tooling will process the graph and compute the necessary information. Doing this in correct order will ensure that when processing a node either its dependencies reside in one of the type databases for previous modules or have been processed earlier. After the processing is finished the tooling will lock the type database and save it.

PR #2912, opened 2025-01-08
[SwiftBindings] Added post processing step to compute struct frozenness

jkurdek:
Implements #2886 

Initialises a post ABI parsing step which calculates information about types and populates type database.

Follow-ups:
- Add tooling sorting modules in topological order for processing
- Remove type database usage in ABI parsing
- Make type database immutable after post-processing step


src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// </summary>
+        /// <param name="namedTypeSpec">Swift type specification for the struct.</param>
+        /// <param name="structDecl">The struct declaration.</param>
+        private void ProcessStruct(NamedTypeSpec namedTypeSpec, StructDecl structDecl)
+        {
+            // Ensure all fields are processed or known in the database.
+            ProcessStructFields(structDecl);
+
+            // TODO: Remove loading dylib
+            IntPtr metadataPtr = DynamicLibraryLoader.invoke(_dylibPath, $"{structDecl.MangledName}Ma");
```
stephen-hawley, line 98: As we discussed, we shouldn't do this.

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        /// <summary>
+        /// Determines whether the struct can be treated as blittable under .NET's rules.
+        /// For now, uses Swift's ValueWitnessTable flags to detect non-POD or non-bitwise-takable.
+        /// </summary>
+        /// <param name="swiftTypeInfo">Struct's Swift type info containing the ValueWitnessTable pointer.</param>
+        /// <returns>True if blittable, false otherwise.</returns>
+        private unsafe bool EvaluateBlittability(SwiftTypeInfo swiftTypeInfo)
+        {
+            // TODO: Replace with a manual approach.
```
stephen-hawley, line 185: 👍  Is there an issue for this?

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// <param name="structDecl">The struct declaration node.</param>
+        /// <param name="swiftTypeInfo">Pointer plus ValueWitnessTable for the struct.</param>
+        /// <param name="isFrozen">Whether the struct is effectively frozen.</param>
+        /// <param name="isBlittable">Whether the struct is effectively blittable.</param>
+        private void RegisterStructType(NamedTypeSpec namedTypeSpec, StructDecl structDecl,
+            SwiftTypeInfo swiftTypeInfo, bool isFrozen, bool isBlittable)
+        {
+            _typeDatabase.Registrar.RegisterType(namedTypeSpec.Module,
+                                                 namedTypeSpec.NameWithoutModule,
+                                                 new TypeRecord
```
stephen-hawley, line 204: This is a separate issue, but we will need the size, stride, and alignment as well.

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        /// <summary>
+        /// Processes a Swift struct declaration to determine final properties such as blittability
+        /// and frozenness, then registers it into the type database.
+        /// </summary>
+        /// <param name="namedTypeSpec">Swift type specification for the struct.</param>
+        /// <param name="structDecl">The struct declaration.</param>
+        private void ProcessStruct(NamedTypeSpec namedTypeSpec, StructDecl structDecl)
+        {
+            // Ensure all fields are processed or known in the database.
```
stephen-hawley, line 94: All nominal types can have inner types.  Because fields in the structure may depend on them, we should process any inner types before we process fields.

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// <param name="isFrozen">Whether the struct is effectively frozen.</param>
+        /// <param name="isBlittable">Whether the struct is effectively blittable.</param>
+        private void RegisterStructType(NamedTypeSpec namedTypeSpec, StructDecl structDecl,
+            SwiftTypeInfo swiftTypeInfo, bool isFrozen, bool isBlittable)
+        {
+            _typeDatabase.Registrar.RegisterType(namedTypeSpec.Module,
+                                                 namedTypeSpec.NameWithoutModule,
+                                                 new TypeRecord
+                                                 {
+                                                     Namespace = namedTypeSpec.Module,
```
stephen-hawley, line 206: I'd like to see a separate type for holding a type name.  Something like:
```
public class /* or record or struct */ TypeName {
     public string Namespace { get; init; }
     public string Name { get; init; }
     public string FullName { get; } => Namespace + "." + Name;
     // if not a record, make ToString, Equals, etc
}
```
This will make it easier to set up `import` statements in Swift and `using` statements in C#.

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -146,7 +146,7 @@ public ModuleDecl GetModuleDecl()
 
             moduleDecl.Fields = decls.OfType<FieldDecl>().ToList();
             moduleDecl.Methods = decls.OfType<MethodDecl>().ToList();
-            moduleDecl.Declarations = decls.Where(d => !(d is MethodDecl) && !(d is FieldDecl)).ToList();
+            moduleDecl.Types = decls.OfType<TypeDecl>().ToDictionary(d => new NamedTypeSpec(d.FullyQualifiedName), d => d);
```
stephen-hawley: One thing that we will run into right here is that the `*Decl` types are already C# and that type may not be the same as the Swift from which it came. For example, if we get a Swift type that is unpronounceable in C# (eg, contains emoji) then we've just created an incorrect NamedTypesSpec.

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -278,9 +268,7 @@ private List<BaseDecl> CollectDeclarations(IEnumerable<Node> nodes, BaseDecl par
                     return null;
             }
 
-            typeRecord.IsProcessed = true;
-
-            if (node.Children != null && decl != null)
+            if (decl != null)
```
stephen-hawley: favor `is not null` over `!= null` to avoid possible issues that arise from operator implementations.

stephen-hawley (changes requested):
Love to see the tests in place. This is a good start, but there are some things that need to be addressed. If it's not practical to resolve them, open issues and keep going.

jkurdek:
CI failures unrelated - #2896 broke compilation. should be fixed once #2909 gets in

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -0,0 +1,223 @@
+using Swift.Runtime;
```
kotlarmilos, line 4:
```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

using Swift.Runtime;
```

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
jkurdek: I agree. This will be the next step. Here is a tracking issue https://github.com/dotnet/runtimelab/issues/2914. Should also cover the comments about blittability and size, stride, alignment.

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -0,0 +1,223 @@
...
+
+namespace BindingsGeneration
+{
+    /// <summary>
+    /// Performs post-processing of types collected from the Swift ABI before generating bindings.
+    /// Calculates properties on types which are not directly available from the ABI.
+    /// Generates type database entries for structs, enums, and classes.
+    /// </summary>
+    internal class ModuleProcessor
```
kotlarmilos, line 13: I like the idea of having a processor/marshaller that based on ABI and dylib inputs, registers types in the database. For that, we need to wire up a Mach-O reader and a demangler.

src/Swift.Bindings/src/Program.cs
```diff
@@ -92,18 +88,22 @@ public static void GenerateBindings(string swiftAbiPath, string dylibPath, strin
...
+
                 // Register the module and set the filter
                 var moduleRecord = typeDatabase.Registrar.RegisterModule(moduleName);
                 moduleRecord.Path = dylibPath;
 
                 // Parse the Swift ABI file and generate declarations
                 var decl = swiftParser.GetModuleDecl();
 
+                var moduleProcessor = new ModuleProcessor(moduleName, dylibPath, decl.Types, typeDatabase);
+                moduleProcessor.FinalizeTypeProcessing();
```
kotlarmilos, line 105: If `ModuleProcessor` is exclusively creating a type registrar, should it be returned here?

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -0,0 +1,223 @@
...
+                case EnumDecl enumDecl:
+                    ProcessEnum(namedTypeSpec, enumDecl);
+                    break;
+
+                case ClassDecl classDecl:
+                    ProcessClass(namedTypeSpec, classDecl);
+                    break;
+
+                default:
+                    throw new NotSupportedException($"Unknown or unsupported type declaration '{typeDecl.GetType().Name}'.");
```
kotlarmilos: It should gracefully handle unknown types to generate partial bindings.

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
jkurdek: They are processed during ABI parsing so we have full information. Added test for that and patched some logic.

stephen-hawley (approved):

src/Swift.Bindings/src/Program.cs
jkurdek: This will be addressed in a follow-up. I need to remove logic using it from ABI parsing

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
jkurdek: Lets address this in a follow up: https://github.com/dotnet/runtimelab/issues/2918

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
jkurdek: I have a change ready, however it will affect multiple areas so Ill bring it in another PR

kotlarmilos (approved):

PR #2916, opened 2025-01-09
Added SwiftRuntimeException

stephen-hawley:
Added SwiftRuntimeException to the Swift.Runtime namespace.

As discussed in the status meeting, this is to be used for issues that arise in typical runtime operations. This will include, but is not limited to things like:
- types encountered that are not marshalable
- type metadata that doesn't map to a known C# type
- entry points that can't be found

In the process of doing this, I address @AaronRobinsonMSFT [comment](https://github.com/dotnet/runtimelab/pull/2877#discussion_r1889263918). I apologize for the oversight. That's on me.

### DIFF
--- a/src/Swift.Runtime/src/Swift/Runtime/TypeMetadata.cs
+++ b/src/Swift.Runtime/src/Swift/Runtime/TypeMetadata.cs
@@ -166,7 +166,7 @@ public readonly struct TypeMetadata : IEquatable<TypeMetadata> {
     public bool IsValid => handle != IntPtr.Zero;
 
     /// <summary>
-    /// Throws a NotSupportedException if the TypeMetadata is invalid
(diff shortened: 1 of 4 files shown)


src/Swift.Runtime/src/Swift/Runtime/TypeMetadata.cs
```diff
@@ -168,11 +168,11 @@ static TypeMetadata ()
     /// <summary>
     /// Throws a NotSupportedException if the TypeMetadata is invalid
```
mcumming: Update the documention

mcumming (approved):

AaronRobinsonMSFT (approved):

jkurdek (approved):

PR #2919, opened 2025-01-10
[SwiftBindings] Refactored TypeDatabase API

jkurdek:
Follow up to #2912. This PR does several things to improve the `TypeDatabase` API:

1. Splits TypeDatabase into module sized chunks which can be serialised independently. This should help caching in the future. 
2. TypeDatabase becomes immutable. One can only add new modules. During each module processing a new ModuleDatabase is created, after creation it becomes part of TypeDatabase 
3. Provides a set of convienience extension methods to query the type database
4. Makes (swift -> c#) name lookups at emit time go through type database

In order to satisfy tests I added some basic handling of closed generic types to module processor. This is more of a draft than a solution. Will be improved in next PRs.


src/Swift.Runtime/src/Swift/TypeDatabase.xml
```diff
@@ -43,28 +40,28 @@
         <entity managedNameSpace="System" managedTypeName="bool">
             <typedeclaration kind="struct" name="Bool" module="Swift" mangledName="$sSb" frozen="true" blittable ="true" />
         </entity>
-        <entity managedNameSpace="Swift.Runtime" managedTypeName="UnsafePointer">
-            <typedeclaration kind="struct" name="UnsafePointer`1" module="Swift" mangledName="sSP" frozen="true" blittable ="true" />
+        <entity managedNameSpace="Swift" managedTypeName="UnsafePointer">
```
kotlarmilos, line 43: `UnsafePointer` is a frozen struct: https://developer.apple.com/documentation/swift/unsafepointer

src/Swift.Runtime/src/Swift/TypeDatabase.xml
```diff
@@ -43,28 +40,28 @@
...
+        <entity managedNameSpace="Swift" managedTypeName="UnsafeRawPointer">
             <typedeclaration kind="struct" name="UnsafeRawPointer" module="Swift" mangledName="sSV" frozen="true" blittable ="true"/>
         </entity>
-        <entity managedNameSpace="Swift.Runtime" managedTypeName="UnsafeMutableRawPointer">
+        <entity managedNameSpace="Swift" managedTypeName="UnsafeMutableRawPointer">
             <typedeclaration kind="struct" name="UnsafeMutableRawPointer" module="Swift" mangledName="sSv" frozen="true" blittable ="true"/>
         </entity>
-        <entity managedNameSpace="Swift.Runtime" managedTypeName="UnsafeBufferPointer">
-            <typedeclaration kind="struct" name="UnsafeBufferPointer`1" module="Swift" mangledName="sSR" frozen="true" blittable ="true"/>
+        <entity managedNameSpace="Swift" managedTypeName="UnsafeBufferPointer">
```
kotlarmilos, line 55: `UnsafeBufferPointer` is a frozen struct: https://developer.apple.com/documentation/swift/unsafebufferpointer

src/Swift.Runtime/src/Swift/TypeDatabase.xml
```diff
@@ -1,9 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<swifttypedatabase version="1.0">
```
kotlarmilos, line 2: Does this file represent a partial type database for Foundation framework? If yes, we should reflect it in the file name.

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
```diff
@@ -1,39 +1,73 @@
...
-        public readonly TypeRegistrar Registrar = new();
+        private readonly ConcurrentDictionary<string, IModuleDatabase> _modules = new();
+
+        // This store is intended for types which are encountered in one module but should belong to another.
+        // This is true for closed generics, where a generic definition is in one module and instantiation is in another.
+        private readonly ConcurrentDictionary<string, TypeRecord> _outOfModuleTypes = new();
+
+        public TypeDatabase()
+        {
+            var voidModuleRecord = new ModuleDatabase("", "");
```
kotlarmilos: It could happen that there are more types without module. It would be good to find if void is defined as _type_ in any module and define a type database for those types.

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
```diff
@@ -1,39 +1,73 @@
...
     {
-        /// <summary>
-        /// The module and type records associated with the database.
-        /// </summary>
-        public readonly TypeRegistrar Registrar = new();
+        private readonly ConcurrentDictionary<string, IModuleDatabase> _modules = new();
+
+        // This store is intended for types which are encountered in one module but should belong to another.
+        // This is true for closed generics, where a generic definition is in one module and instantiation is in another.
+        private readonly ConcurrentDictionary<string, TypeRecord> _outOfModuleTypes = new();
```
kotlarmilos, line 21: Nit: It is confusing to have `_modules` and `_outOfModuleTypes`, let's try to find better naming. UnkownTypes/UnknownModules?

Additionally, I would keep them in the same dictionary with a known/unknown flag.

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
```diff
@@ -1,39 +1,73 @@
...
+    public class TypeDatabase : ITypeDatabase
     {
-        /// <summary>
-        /// The module and type records associated with the database.
-        /// </summary>
-        public readonly TypeRegistrar Registrar = new();
+        private readonly ConcurrentDictionary<string, IModuleDatabase> _modules = new();
+
+        // This store is intended for types which are encountered in one module but should belong to another.
+        // This is true for closed generics, where a generic definition is in one module and instantiation is in another.
```
kotlarmilos, line 19: This should be true when a module has dependencies on other modules, not necessarily related to generics, right?

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
```diff
@@ -1,39 +1,73 @@
...
 using System.Xml;
 
 namespace BindingsGeneration
 {
+
     /// <summary>
     /// Manages a mapping database between Swift types and C# types.
     /// </summary>
-    public unsafe class TypeDatabase
+    public class TypeDatabase : ITypeDatabase
```
kotlarmilos, line 14: ModuleDatabase and TypeDatabase naming might be confusing. My first insight was that a module contain types, but type database contains modules.

kotlarmilos, line 14: Does TypeDatabase always contain all modules? If yes, I would suggest:

```suggestion
    public static class TypeDatabase : ITypeDatabase
```

src/Swift.Bindings/src/Program.cs
```diff
@@ -80,29 +80,25 @@ public static void Main(string[] args)
         /// <param name="verbose">Verbosity level.</param>
         public static void GenerateBindings(string swiftAbiPath, string dylibPath, string outputDirectory, int verbose = 2)
         {
-            TypeDatabase typeDatabase = new(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Swift", "TypeDatabase.xml"));
+            var typeDatabase = new TypeDatabase();
+            typeDatabase.LoadModuleDatabaseFromFile(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "Swift", "TypeDatabase.xml"));
```
kotlarmilos: Is this Swift.Foundation database?

src/Swift.Bindings/src/Model/ModuleParsingResult.cs
```diff
@@ -1,7 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
 namespace BindingsGeneration
 {
     public sealed record ModuleParsingResult(
```
kotlarmilos: Can this be moved to the parser?

src/Swift.Runtime/src/Swift/TypeDatabase.xml
jkurdek: This database entry is about open generic. We dont have a way to determine frozenness of this entity yet so I put it to false. This will be addressed once we handle generics. Can revert if you want.

src/Swift.Runtime/src/Swift/TypeDatabase.xml
jkurdek: same as above

src/Swift.Runtime/src/Swift/TypeDatabase.xml
kotlarmilos: Ok, just add TODO/tracking issue and proceed :)

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
jkurdek: We discussed this offline, summary:

When processing a module we can encounter several type of things:
- Dependencies on the currently processed module
- Dependencies on other module
- Things which look like they exist in other modules but they exist in this module (e.g. closed generics)

This store is meant only to keep track of the things from the third point. Imo those should be kept separate from the things from first two points.

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
jkurdek: Sure, I am open to suggestions :)

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
jkurdek: I think this is something we can handle once we encounter such a type

src/Swift.Bindings/src/Model/ModuleParsingResult.cs
```diff
@@ -1,7 +1,11 @@
...
         ModuleDecl ModuleDecl,
-        Dictionary<NamedTypeSpec, TypeDecl> TypeDecls
+        Dictionary<NamedTypeSpec, TypeDecl> TypeDecls,
+        List<NamedTypeSpec> ClosedGenericTypes
```
stephen-hawley: What does "Closed" mean in this context?

src/Swift.Bindings/src/TypeDatabase/ITypeDatabase.cs
```diff
@@ -0,0 +1,52 @@
...
+/// <summary>
+/// Type database. Holds module databases and type records for Swift types.
+/// </summary>
+public interface ITypeDatabase
+{
+    /// <summary>
+    /// Loads a module database from a specified file.
+    /// </summary>
+    /// <param name="file">The file path of the module database to load.</param>
+    public void LoadModuleDatabaseFromFile(string file);
```
stephen-hawley: Since I/O is slow, let's make this async.
Do we need a stream variant?

src/Swift.Bindings/src/TypeDatabase/ITypeDatabase.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <returns><c>true</c> if the type record was found; otherwise, <c>false</c>.</returns>
+    public bool TryGetTypeRecord(string moduleName, string typeIdentifier, [NotNullWhen(returnValue: true)] out TypeRecord? record);
+
+    /// <summary>
+    /// Retrieves the library path for the specified module.
+    /// </summary>
+    /// <param name="moduleName">The name of the module.</param>
+    /// <returns>The file path of the library associated with the module.</returns>
+    /// <exception cref="Exception">Thrown if the library path does not exist for the specified module.</exception>
+    public string GetLibraryPath(string moduleName);
```
stephen-hawley, line 38: Is there any reason why this shouldn't be a property?

src/Swift.Bindings/src/TypeDatabase/ModuleDatabase.cs
```diff
@@ -0,0 +1,87 @@
...
+            return false;
+        }
+
+        /// <summary>
+        /// Retrieves all type records registered in the module.
+        /// </summary>
+        /// <returns>A list of all type records in the module.</returns>
+        public List<TypeRecord> GetTypes()
+        {
+            return _typeRecords.Values.ToList();
```
stephen-hawley: questions:
1 - how often will this get called?
2 - between calls, is it likely to get mutated?
3 - is it likely to be mutated while the return is being processed?

If this is likely to get called more than 3 or 4 times and the dictionary won't change, we should cache the list instead of regenerating it.

If it won't be mutated while being processed and the call site doesn't need random access, maybe we should make it `IEnumerable<TypeRecord>` instead.

stephen-hawley (approved):
Nice work!

A couple of things that I have questions about and one change I'd like to see.

src/Swift.Bindings/src/Model/ModuleParsingResult.cs
jkurdek: I meant generic types which have all of their type arguments specified. We need to calculate size, stride, alignment and frozenness for those

src/Swift.Bindings/src/TypeDatabase/ModuleDatabase.cs
jkurdek: Thanks for spotting this one. Now I am thinking that we might actually not need that at all - at least I cant really find a use case for now. I would be leaning towards removing this method.

src/Swift.Bindings/src/TypeDatabase/ITypeDatabase.cs
jkurdek: I'm not sure I understand. This is on `TypeDatabase`, not `ModuleDatabase`. The idea I had when writing this was that during marshalling / code emition we should not care how `TypeDatabase` looks like on the inside (how modules are structured etc.), we just interact with `ITypeDatabase`

src/Swift.Bindings/src/TypeDatabase/ITypeDatabase.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// </summary>
+    /// <param name="file">The file path of the module database to load.</param>
+    public void LoadModuleDatabaseFromFile(string file);
+
+    /// <summary>
+    /// Adds a module database to the type database.
+    /// </summary>
+    /// <param name="moduleDatabase">The module database to add.</param>
+    /// <exception cref="Exception">Thrown if a module with the same name already exists in the database.</exception>
+    public void AddModuleDatabase(IModuleDatabase moduleDatabase);
```
jkurdek: I thought about it for some time. Maybe we should remove the `IModuleDatabase` interface, and move the methods which add stuff to TypeDatabase out of this interface. This would make the `ITypeDatabase` fully readonly. Methods adding stuff to it could reside on concrete implementation and become implementation detail.

Curious to learn your thoughts.

src/Swift.Bindings/src/Model/ModuleParsingResult.cs
stephen-hawley: Thanks - can I instead suggest the term "bound" which I've seen used more often?

src/Swift.Bindings/src/TypeDatabase/ModuleDatabase.cs
stephen-hawley: Good idea. It can always be added back later.

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
kotlarmilos: Offline sync - _outOfModuleTypes refers to instantiations that are not stored in the type database and exist only in memory at binding time.

src/Swift.Bindings/src/TypeDatabase/ITypeDatabase.cs
jkurdek: Made it async, parsing xml is still synchronous so there is only partial gain. I think we can revisit this one in the future.

src/Swift.Bindings/src/TypeDatabase/TypeDatabase.cs
jkurdek: I would say it depends on how we implement module by module processing algorithm, I would defer this until then

kotlarmilos (approved):

Issue #2891, opened 2024-12-17
[SwiftBindings] Unify code emition


jkurdek:
We should decide on how do we want to handle emitting code in this project (e.g. IndentedTextWriter) and provide reusable tooling for that method.

kotlarmilos:
Closing as completed. Will be presented in the new design proposal.

PR #2927, opened 2025-01-14
[SwiftBindings] Added returning non-frozen structs from methods

### LINKED ISSUE #2885: [Swift language features] Project non-init functions returning new instance of non-frozen struct
Currently only constructors support returns using `IndirectResult`. This should be extended to cover all the methods.

jkurdek:
Fixes #2885. 

This PR introduces the following changes:
- Makes projected structs implement `ISwiftObject` interface. 
- Adds correct handling of `IndirectResult` on non-init methods
- Makes `IntPtr` to `SwiftHandle` conversion explicit to prevent name clashes (constructor accepting IntPtr vs constructor accepting SwiftHandle)
- Makes `MarshalFromSwift` use `SwiftHandle` instead of `IntPtr`

### DIFF
--- a/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
@@ -189,7 +189,13 @@ namespace BindingsGeneration
             var pInvokeName = NameProvider.GetPInvokeName(methodDecl);
             var staticKeyword = methodDecl.MethodType == MethodType.Static || parentDecl is ModuleDecl ? "static " : "";
 
-            writer.WriteLine($"public {staticKeyword}{env.SignatureHandler.GetWrapperSignature().ReturnType} {methodDecl.Name}({env.SignatureHandler.GetWrapperSignature().ParametersString()})");
+            var wrapperSignature = env.SignatureHandler.GetWrapperSignature();
+            var pInvokeSignature = env.SignatureHandler.GetPInvokeSignature();
+
+            var requiresIndirectResult = MarshallingHelpers.MethodRequiresIndirectResult(methodDecl, parentDecl, env.TypeDatabase);
+            var unsafeKeyword = requiresIndirectResult ? "unsafe " : "";
+
+            writer.WriteLine($"public {staticKeyword}{unsafeKeyword} {wrapperSignature.ReturnType} {methodDecl.Name}({wrapperSignature.ParametersString()})");
             writer.WriteLine("{");
             writer.Indent++;
 
@@ -201,14 +207,21 @@ namespace BindingsGeneration
                     writer.WriteLine($"var self = new SwiftSelf((void*)_payload);");
             }
 
-            // TODO: Add Indirect result marshalling to methods other than constructors
+            if (requiresIndirectResult)
+            {
+                writer.WriteLine($"var payload = (SwiftHandle)NativeMemory.Alloc({wrapperSignature.ReturnType}.PayloadSize);");
+                writer.WriteLine("var swiftIndirectResult = new SwiftIndirectResult((void*)payload);");
+                writer.WriteLine($"{pInvokeName}({pInvokeSignature.CallArgumentsString()});");
+                writer.WriteLine($"return SwiftMarshal.MarshalFromSwift<{wrapperSignature.ReturnType}>((SwiftHandle)swiftIndirectResult.Value);");
+            }
+            else
+            {
+                var returnPrefix = methodDecl.CSSignature.First().SwiftTypeSpec.IsEmptyTuple ? "" : "return ";
+                var invokeArguments = env.SignatureHandler.GetPInvokeSignature().CallArgumentsString();
 
-            var returnPrefix = methodDecl.CSSignature.First().SwiftTypeSpec is TupleTypeSpec tupleTypeSpec && tupleTypeSpec.IsEmptyTuple ? "" : "return ";
-            var invokeArguments = env.SignatureHandler.GetPInvokeSignature().CallArgumentsString();
-
-            // Call the PInvoke method
-            writer.WriteLine($"{returnPrefix}{pInvokeName}({invokeArguments});");
-
+                // Call the PInvoke method
+                writer.WriteLine($"{returnPrefix}{pInvokeName}({invokeArguments});");
+            }
             writer.Indent--;
             writer.WriteLine("}");
         }
--- a/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
@@ -206,7 +206,7 @@ namespace BindingsGeneration
 
             WritePrivateFields(writer, structDecl);
             WriteDisposeMethod(writer);
-            WriteFinalizer(writer, structDecl.Name);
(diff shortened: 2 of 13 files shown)


src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
```diff
@@ -348,4 +353,87 @@ public void Emit(IndentedTextWriter writer, IEnvironment env, Conductor conducto
...
+        public void WriteFrozenStructImplementation()
+        {
+            WriteGetTypeMetadata();
+            WriteNewFromPayloadFrozenStruct();
+            WriteMarshalToSwift();
+        }
+
+        private void WriteGetTypeMetadata()
+        {
+            _writer.WriteLine("public static TypeMetadata GetTypeMetadata() => PInvoke_getMetadata();");
```
stephen-hawley: In `SwiftOptional`, I made a conscious choice to implement the `ISwiftObject` explicitly who avoids polluting the API of each object. Let's do that here too.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
```diff
@@ -348,4 +353,87 @@ public void Emit(IndentedTextWriter writer, IEnvironment env, Conductor conducto
...
+            _writer.WriteLine("}");
+            _writer.WriteLine();
+
+            EmitPrivateConstructor();
+        }
+
+
+        private void EmitPrivateConstructor()
+        {
+            _writer.WriteLine($"private unsafe {_structDecl.Name}(SwiftHandle handle)");
```
stephen-hawley: I tend to avoid `private` since it's the C# default. This is very much a macois style thing.

stephen-hawley (changes requested):
I like that `SwiftHandle` changes.
Biggest issue for me is not having an explicit implementation of `ISwiftObject`

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
jkurdek: I am wondering about this one. Making a static interface method explicitly implemented makes in inaccessible within the type implementing the method, right?

However we might need to use `TypeMetadata` within the type. For example to get the size of non-frozen struct at runtime. The following will not work anymore:

```csharp
static TypeMetadata ISwiftObject.GetTypeMetadata() => ...

static nint PayloadSize => GetTypeMetadata().Size()
```

I guess we could introduce some private methods, is this an approach you'd prefer?

stephen-hawley: To get the type metadata, I used `var metadata = SwiftObjectHelper<SwiftOptional<T>>.GetTypeMetadata();`

src/Swift.Runtime/src/Swift/SwiftOptional.cs
```diff
@@ -43,7 +43,7 @@ static TypeMetadata ISwiftObject.GetTypeMetadata()
     /// <summary>
     /// Creates a new SwiftOptional from a Swift payload
     /// </summary>
-    static ISwiftObject ISwiftObject.NewFromPayload(IntPtr payload)
+    static ISwiftObject ISwiftObject.NewFromPayload(SwiftHandle payload)
```
kotlarmilos, line 46: Is there a functional difference when using SwiftHandle instead of IntPtr, or not yet? I am trying to understand the benefits of the SwiftHandle abstraction.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -189,7 +189,13 @@ private void EmitWrapperMethod(IndentedTextWriter writer, MethodEnvironment env)
             var pInvokeName = NameProvider.GetPInvokeName(methodDecl);
             var staticKeyword = methodDecl.MethodType == MethodType.Static || parentDecl is ModuleDecl ? "static " : "";
```
kotlarmilos, line 190: Should we introduce a function for the modifiers? I see `static` and `unsafe` for now, we will have `async` soon, and potentially more in the future.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -201,14 +207,21 @@ private void EmitWrapperMethod(IndentedTextWriter writer, MethodEnvironment env)
...
-            // TODO: Add Indirect result marshalling to methods other than constructors
-
-            var returnPrefix = methodDecl.CSSignature.First().SwiftTypeSpec is TupleTypeSpec tupleTypeSpec && tupleTypeSpec.IsEmptyTuple ? "" : "return ";
-            var invokeArguments = env.SignatureHandler.GetPInvokeSignature().CallArgumentsString();
-
-            // Call the PInvoke method
-            writer.WriteLine($"{returnPrefix}{pInvokeName}({invokeArguments});");
+            if (requiresIndirectResult)
+            {
+                writer.WriteLine($"var payload = (SwiftHandle)NativeMemory.Alloc({wrapperSignature.ReturnType}.PayloadSize);");
```
kotlarmilos, line 212: Nit: A single write with multiple lines would improve readability.

kotlarmilos (approved):
LGTM!

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: I agree. We discussed that on one of the previous PRs that single write will mess up the indent. I have a task #2891 on my backlog to improve that

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: I think this can be handled as a follow up, so far the use is very local

src/Swift.Runtime/src/Swift/SwiftOptional.cs
jkurdek: The difference boils down mainly to handling conflicts in generated code. Lets assume we have a struct `Foo` which on swift side has a single `Int` field. We represent `Int` as `IntPtr` on C# side. 

```csharp
struct Foo
{
    Foo(IntPtr value) {} // Constructor accepting single integer of sized determined by target platform
    Foo(SwiftHandle handle) {} // Constructor accepting a pointer
}
```

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
jkurdek: Changed to explicit interface implementation

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
jkurdek: Removed `private` from generated code

jkurdek:
Applied feedback from @stephen-hawley, merging

PR #2937, opened 2025-01-15
[SwiftBindings] Added basic C# file support and file writing code

stephen-hawley:
This adds the basic elements needed to support a simple (empty) C# file with the modern namespace declaration.
This adds a simple CodeWriter for writing elements out.
This adds the elements for C#:
- CSNamespace - a namespace declaration
- CSUsing - a using statement
- CSUsingPackages - a collection of using statements
- CSComment - a single line comment
- CSCommentBlock - a collection of single line comments
- CSTopLevelDeclarations - a collection of top-level declaration objects

Note that several of these types have the fluent method "And" which is useful for chaining things together.
for example:
```csharp
var comments = new CSCommentBlock().And("Copyright (c) Microsoft Corporation.").And("Licensed under the MIT License.");
```


src/Dynamo/src/Dynamo.CSLang/CSComment.cs
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) Microsoft Corporation.
```
kotlarmilos, line 1: Is this needed?

src/Dynamo/src/Dynamo.CSLang/CSUsing.cs
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) Microsoft Corporation.
```
kotlarmilos, line 1: Isn't this part of the emitter's environment? This should be known by parsing the ABI.

src/Dynamo/src/Dynamo.CSLang/CSNamespace.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
```
kotlarmilos, line 1: Isn't this part of the emitter's environment? This should be known by parsing the ABI.

kotlarmilos (commented):
Remove `src/Swift.Runtime/src/Metadata/TypeMetadata.cs`, I guess it was added by mistake.

kotlarmilos:
> Note that several of these types have the fluent method "And" which is useful for chaining things together.
for example:

> var comments = new CSCommentBlock().And("Copyright (c) Microsoft Corporation.").And("Licensed under the MIT License.");

We need this in the string-based emitter as well. I suggest extracting it from the `Dynamo`.

jkotas:
> `var comments = new CSCommentBlock().And("Copyright (c) Microsoft Corporation.").And("Licensed under the MIT License.");`

This is a fine example. We should not be emitting license headers like this into the generated bindings. The binding generator is a compiler. The default licensing for compiler outputs is that compiler outputs are redistributable without additional restrictions. The only restrictions are based on the license of the compiler inputs used to produce the output.

stephen-hawley:
> > `var comments = new CSCommentBlock().And("Copyright (c) Microsoft Corporation.").And("Licensed under the MIT License.");`
> 
> This is a fine example. We should not be emitting license headers like this into the generated bindings. The binding generator is a compiler. The default licensing for compiler outputs is that compiler outputs are redistributable without additional restrictions. The only restrictions are based on the license of the compiler inputs used to produce the output.

That's a great point and one which we discussed at the weekly meeting. This is meant to be an example of the fluent interface - nothing more.

stephen-hawley:
> > Note that several of these types have the fluent method "And" which is useful for chaining things together.
> > for example:
> 
> > var comments = new CSCommentBlock().And("Copyright (c) Microsoft Corporation.").And("Licensed under the MIT License.");
> 
> We need this in the string-based emitter as well. I suggest extracting it from the `Dynamo`.

If you use `StringBuilder` you get that for free - all the Add methods are fluent.

src/Dynamo/src/Dynamo.CSLang/CSNamespace.cs
stephen-hawley: It will be known by the parsing ABI, but Dynamo is a structured output generator and knows nothing about the parser. What will happen is at the beginning of working on a type, the environment for a dynamo emitter will have a `CSFile` member that the parsing ABI will populate with the namespace. As for the usings, they are still needed and can either be done just after CSFile creation or during the process of binding. Every time a C# type is referenced, a simple AddIfNotPresent gets called, ensuring that (1) you never miss a using and (2) you never duplicate a using. This also holds true for types that are not directly referenced by the ABI, for example if we need `System.Runtime` or `System.Runtime.InteropServices` etc.

src/Dynamo/README.md
kotlarmilos, line 1: Nit: Typo in title `Dyanmo` -> `Dynamo`

kotlarmilos (approved):

src/Dynamo/src/Dynamo.CSLang/CSNamespace.cs
kotlarmilos: Currently, the Dynamo contains emitter and context for marshalling. I would like to see it decoupled, and I am interested to see how it evolves in the future.

jkurdek (approved):

jkurdek:
Thanks for bringing in the recipes!

PR #2939, opened 2025-01-16
[Swift Bindings] Add basic support for throws method

### LINKED ISSUE #2888: [Swift language features] Add support for methods marked with `throws`
We should capture the `SwiftError` when method is marked with `throws` keyword.

jkurdek:
Fixes #2888 

This PR introduces basic support for handling `throws` methods. It reads value from `SwiftError` and raises a simple `SwiftRuntimeError` when the value is not null. This should be followed up with more complex error support in future milestones.

In order to reduce code duplication I extracted wrapper emission into separate class.

### DIFF
--- a/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
@@ -227,7 +227,6 @@ namespace BindingsGeneration
         /// </summary>
         private static void WritePrivateFields(IndentedTextWriter writer, StructDecl structDecl)
         {
-            writer.WriteLine();
             writer.WriteLine($"static nuint _payloadSize = SwiftObjectHelper<{structDecl.Name}>.GetTypeMetadata().Size;");
             writer.WriteLine("SwiftHandle _payload = SwiftHandle.Zero;");
             writer.WriteLine("bool _disposed = false;");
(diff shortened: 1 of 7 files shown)


src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Structs/StructsTests.swift
```diff
@@ -167,6 +167,43 @@ public struct StructBuilder
...
+{
+	public var x: Int
+	public var y: Int
+
+	public init(x: Int, y: Int) throws
+	{
+		self.x = x
+		self.y = y
+
+        throw NSError()
```
kotlarmilos, line 181: Nit: ident

kotlarmilos (approved):
LGTM!

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/TypeHandler.cs
```diff
@@ -239,20 +238,20 @@ private static void WritePrivateFields(IndentedTextWriter writer, StructDecl str
...
-            writer.Indent++;
-            writer.WriteLine("NativeMemory.Free((void*)_payload);");
-            writer.WriteLine("_payload = SwiftHandle.Zero;");
-            writer.WriteLine("_disposed = true;");
-            writer.WriteLine("GC.SuppressFinalize(this);");
-            writer.Indent--;
-            writer.WriteLine("}");
-            writer.Indent--;
-            writer.WriteLine("}");
+            var text = $$"""
```
kotlarmilos, line 241: Much better now 👍

PR #2941, opened 2025-01-17
Prototyping `T RuntimeHelpers.Await<T>(Task<T>)`

VSadov:
Not for merging, just taking a look how `RuntimeHelpers.Await` could be implemented.

The code actually runs. The implementation is suboptimal because we always suspend. 
We do not have to, but it would require implementing `ReturnAsync2` intrinsic, the value-returning counterpart of `SuspendAsync2`. 
It seems doable, but needs a bit more JIT expertise.

Even with always supending the perf is not too bad - roughly 30% worse on the `varying-yield` benchmark. 
Although this benchmark suspends a lot, thus is not impacted as much by unconditional suspend.

For the real implementation we would definitely need `ReturnAsync2`


src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
```diff
@@ -208,6 +208,75 @@ public static void UnsafeAwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter
             SuspendAsync2(sentinelContinuation);
             return;
         }
+
```
VSadov, line 211: The interesting part. The rest of code changes are mechanical - to make the helper known as special method.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
```diff
@@ -208,6 +208,75 @@ public static void UnsafeAwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter
...
+                // unreachable
+                return task.ResultOnSuccess;
+            //}
+            //else
+            //{
+            //    //   RETURN {task.Result, null}
+            //    //
+            //    T result = task.Result;
+            //    ReturnAsync2(Unsafe.AsPointer(ref result));
```
VSadov: Just guessing how `ReturnAsync2` could look. 

If generic intrisic is ok, it could also be just `ReturnAsync2<T>(result)`

src/tests/async/varying-yields.cs
```diff
@@ -91,9 +93,19 @@ async2 Task<long>
             double liveState3 = _yieldProbability;
 
             if (depth == 0)
+#if AWAIT
+                return RuntimeHelpers.Await(Loop());
+#else
                 return await Loop();
+#endif
+
```
VSadov, line 101: Using `Await` in a benchmark/test.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
```diff
@@ -208,6 +208,75 @@ public static void UnsafeAwaitAwaiterFromRuntimeAsync<TAwaiter>(TAwaiter awaiter
...
+            //
+            //    // unreachable
+            //    return result;
+            //}
+        }
+
+        internal static class AwaitHelper<T>
+        {
```
VSadov: Resuming the continuation simply pushes the result into the caller continuation. (or throws, if faulted)

VSadov: I wonder if `UnsafeAwaitAwaiterFromRuntimeAsync` needs to call `GetResult()` upon resuming as well? (in case it faulted)

VSadov: Also `GetAwaiter().GetResult()` seems more appropriate than `.Result`. This is just a rough prototype anyways.

jakobbotsch: > I wonder if `UnsafeAwaitAwaiterFromRuntimeAsync` needs to call `GetResult()` upon resuming as well? (in case it faulted)

The caller is responsible for calling `awaiter.GetResult()` (`UnsafeAwaitAwaiterFromRuntimeAsync` has no way to bind to the awaiter anyway).

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
jakobbotsch: I don't see a good reason to not write this function in terms of `UnsafeAwaitAwaiterFromRuntimeAsync`. This looks more complicated than necessary.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
jakobbotsch: I don't think we need an intrinsic, this can just be a normal return.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
VSadov: > The caller is responsible for calling awaiter.GetResult()

Ah, right, in this case we have code in the caller.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
VSadov: Would a regular return guarantee that the continuation return is null?

jakobbotsch: It would need the VM to tell the JIT this function has runtime-async calling convention. That's probably something we'd want anyway. It might also be the only thing we need to allow using `UnsafeAwaitAwaiterFromRuntimeAsync` in its implementation.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
VSadov: > I don't see a good reason to not write this function in terms of UnsafeAwaitAwaiterFromRuntimeAsync. This looks more complicated than necessary.

Possibly. I started on that path, but was running into asserts (something about conditional BB not ending with conditional jump,...).  

I was not sure if that was something that I did wrong or issues with `SuspendAsync2` in unusual context. 

This way worked though.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
VSadov: I will try. We do not do that for `UnsafeAwaitAwaiterFromRuntimeAsync`, but that may work just because it is a simpler method.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
jakobbotsch: I don't think we will need an explicit `SuspendAsync2` once we switch to use `UnsafeAwaitAwaiter...` with everything hooked up. The JIT should create the state machine itself at that point.

VSadov:
switched to state machine version of Await. 

It was a bit more involved than I expected. The main issue is that for the "magic" methods you can't figure their "async-ness" by only looking at the signature, which is kind of ByDesign. I think we may have to keep the special methods as Intrinsics. 

Explicit async is also a bit viral - once you declare something as async machine you'd need to call with async call convention and vice versa, lest you get into obscure crashes from dereferencing continuations that are not actually passed by the caller or calee and depend on what junk there was in the corresponding register. 
(thus `UnsafeAwaitAwaiterFromRuntimeAsync` had to be moved onto new plan as well)

I like the new approach more, though. It seems more generally useful/scalable in case we need more "magic" helpers.

src/coreclr/vm/threadsuspend.cpp
```diff
@@ -4915,7 +4915,9 @@ bool IsSpecialCaseAsyncRet(MethodDesc* pMD)
     // causing loading to happen? Also, can we just mark them as async2 in SPC,
     // or force them to be fully interruptible?
     LPCUTF8 name = pMD->GetName();
-    return strcmp(name, "UnsafeAwaitAwaiterFromRuntimeAsync") == 0 || strcmp(name, "AwaitAwaiterFromRuntimeAsync") == 0;
+    return strcmp(name, "UnsafeAwaitAwaiterFromRuntimeAsync") == 0 ||
+        strcmp(name, "AwaitAwaiterFromRuntimeAsync") == 0 ||
+        strcmp(name, "Await") == 0;
```
jakobbotsch, line 4920: This can probably be removed now that these are marked as runtime-async via MethodImpl

VSadov: Yes, nothing special about the helpers names now.

I think the whole special-casing of async function in hijacking will be unnecessary as we stop caring about return kinds, except on x86

VSadov:
closing, since we have https://github.com/dotnet/runtimelab/pull/2951

PR #2946, opened 2025-01-20
[SwiftBindings] Added support for generic methods without constraints

jkurdek:
Fixes https://github.com/dotnet/runtimelab/issues/2942

This PR:
- Brings basic support for generic methods. 
- Adds models for generic params and constraints
- Simplifies some code in the Emitter, removes unnecessary parameter passing, deletes redundant classes

### DIFF
--- a/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
@@ -5,75 +5,6 @@ using System.CodeDom.Compiler;
 
 namespace BindingsGeneration
 {
-    /// <summary>
-    /// Factory class for creating instances of ConstructorHandler.
-    /// </summary>
-    public class ConstructorHandlerFactory : IFactory<BaseDecl, IMethodHandler>
-    {
-        /// <summary>
-        /// Determines if the factory handles the specified declaration.
-        /// </summary>
-        /// <param name="decl">The base declaration.</param>
-        public bool Handles(BaseDecl decl)
-        {
-            return decl is MethodDecl methodDecl && methodDecl.IsConstructor;
-        }
-
-        /// <summary>
-        /// Constructs a new instance of ConstructorHandler.
-        /// </summary>
-        public IMethodHandler Construct()
-        {
-            return new ConstructorHandler();
-        }
-    }
-
-    /// <summary>
-    /// Handler class for constructor declarations.
-    /// </summary>
-    public class ConstructorHandler : BaseHandler, IMethodHandler
-    {
-        public ConstructorHandler()
-        {
-        }
-
-        /// <summary>
-        /// Marshals the specified constructor.
-        /// </summary>
-        /// <param name="methodDecl">The method declaration.</param>
-        /// <param name="typeDatabase">The type database instance.</param>
-        public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
-        {
-            if (decl is not MethodDecl methodDecl)
-            {
-                throw new ArgumentException("The provided decl must be a MethodDecl.", nameof(decl));
-            }
-            return new MethodEnvironment(methodDecl, typeDatabase);
-        }
-
-        /// <summary>
-        /// Emits the code for the specified environment.
-        /// </summary>
-        /// <param name="writer">The IndentedTextWriter instance.</param>
-        /// <param name="env">The environment.</param>
-        /// <param name="conductor">The conductor instance.</param>
-        public void Emit(IndentedTextWriter writer, IEnvironment env, Conductor conductor)
-        {
-            var methodEnv = (MethodEnvironment)env;
-            var signatureHandler = new SignatureHandler(methodEnv);
-
-            if (signatureHandler.GetWrapperSignature().ContainsPlaceholder)
-            {
-                Console.WriteLine($"Method {methodEnv.MethodDecl.Name} has unsupported signature: ({signatureHandler.GetWrapperSignature().ParametersString()}) -> {signatureHandler.GetWrapperSignature().ReturnType}");
-                return;
-            }
-
-            var wrapperEmitter = new WrapperEmitter(methodEnv, signatureHandler);
-            wrapperEmitter.EmitConstructor(writer);
-            PInvokeEmitter.EmitPInvoke(writer, methodEnv, signatureHandler);
-            writer.WriteLine();
-        }
-    }
 
     /// <summary>
     /// Represents a method handler factory.
@@ -87,7 +18,7 @@ namespace BindingsGeneration
         /// <returns></returns>
         public bool Handles(BaseDecl decl)
         {
-            return decl is MethodDecl methodDecl && !methodDecl.IsConstructor;
+            return decl is MethodDecl;
         }
 
         /// <summary>
@@ -139,7 +70,14 @@ namespace BindingsGeneration
             }
 
             var wrapperEmitter = new WrapperEmitter(methodEnv, signatureHandler);
-            wrapperEmitter.EmitMethod(writer);
+            if (methodEnv.MethodDecl.IsConstructor)
+            {
+                wrapperEmitter.EmitConstructor(writer);
+            }
+            else
+            {
+                wrapperEmitter.EmitMethod(writer);
+            }
             PInvokeEmitter.EmitPInvoke(writer, methodEnv, signatureHandler);
             writer.WriteLine();
         }
--- a/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
@@ -136,8 +136,8 @@ namespace BindingsGeneration
             var argument = _env.MethodDecl.CSSignature.First();
             if (argument.IsGeneric)
             {
-                var CSName = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()];
-                SetReturnType(CSName.TypeName);
+                var placeholderName = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()].PlaceholderName;
+                SetReturnType(placeholderName);
             }
             else
             {
@@ -155,8 +155,8 @@ namespace BindingsGeneration
             {
                 if (argument.IsGeneric)
                 {
-                    var CSName = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()];
-                    AddParameter(CSName.TypeName, argument.Name);
+                    var placeholderName = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()].PlaceholderName;
+                    AddParameter(placeholderName, argument.Name);
                 }
                 else
                 {
@@ -242,8 +242,8 @@ namespace BindingsGeneration
             {
                 if (argument.IsGeneric)
                 {
-                    var CSName = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()];
-                    AddParameter("IntPtr", CSName.PayloadName);
+                    var payloadName = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()].PayloadName;
+                    AddParameter("IntPtr", payloadName);
                 }
                 else if (MarshallingHelpers.ArgumentIsMarshalledAsCSStruct(argument, _env.TypeDatabase))
                 {
@@ -264,8 +264,8 @@ namespace BindingsGeneration
         {
             foreach (var genericParameter in _env.MethodDecl.GenericParameters)
             {
-                var CSName = _env.GenericTypeMapping[genericParameter.TypeName];
-                AddParameter("TypeMetadata", CSName.MetadataName);
+                var metadataName = _env.GenericTypeMapping[genericParameter.TypeName].MetadataName;
+                AddParameter("TypeMetadata", metadataName);
             }
         }
 
@@ -671,7 +671,7 @@ namespace BindingsGeneration
         {
             var genericParams = _env.MethodDecl.IsGeneric switch
             {
-                true => $"<{string.Join(", ", _env.MethodDecl.GenericParameters.Select(p => _env.GenericTypeMapping[p.TypeName].TypeName))}>",
+                true => $"<{string.Join(", ", _env.MethodDecl.GenericParameters.Select(p => _env.GenericTypeMapping[p.TypeName].PlaceholderName))}>",
                 false => ""
             };
             writer.WriteLine($"public {_env.ParentDecl.Name}{genericParams}({_wrapperSignature.ParametersString()})");
@@ -685,7 +685,7 @@ namespace BindingsGeneration
         {
             var genericParams = _env.MethodDecl.IsGeneric switch
             {
-                true => $"<{string.Join(", ", _env.MethodDecl.GenericParameters.Select(p => _env.GenericTypeMapping[p.TypeName].TypeName))}>",
+                true => $"<{string.Join(", ", _env.MethodDecl.GenericParameters.Select(p => _env.GenericTypeMapping[p.TypeName].PlaceholderName))}>",
                 false => ""
             };
 
--- a/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
@@ -502,6 +502,7 @@ namespace BindingsGeneration
             EmitBodyStart(writer);
             EmitSwiftSelf(writer);
             EmitIndirectResultConstructor(writer);
+            EmitGenericArguments(writer);
             EmitPInvokeCall(writer);
             EmitSwiftError(writer);
             EmitReturnConstructor(writer);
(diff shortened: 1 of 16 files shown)


src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Generics/GenericTests.swift
```diff
@@ -0,0 +1,47 @@
...
+    public var y: Int
+
+    public init(x: Int, y: Int) {
+        self.x = x
+        self.y = y
+    }
+}
+
+public func AcceptsGenericParametersAndThrows<T, U> (a: T, b: U) throws -> Int {
+    throw NSError()
```
matouskozak, line 30: Why do we throw `NSError` instead of swift error? Is it simpler to catch at the current state of the tooling?

src/Swift.Bindings/src/Model/TypeDecl/BaseDecl.cs
```diff
@@ -11,7 +11,7 @@ public record BaseDecl
         /// <summary>
         /// Name of the declaration.
         /// </summary>
-        public required string Name { get; set; }
+        public required string Name { get; set; } //TODO: Hide or remove this property. This might not contain a correct name.
```
matouskozak, line 14: Could you provide more context to this comment?

matouskozak (commented):
Looking good!

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Generics/GenericTests.swift
jkurdek: I think it is easiest to throw without defining a custom enum. For now it doesnt really matter what error we throw the behaviour is going to be the same on the projections side. I think we can revisit this once we have more complex error support

src/Swift.Bindings/src/Model/TypeDecl/BaseDecl.cs
jkurdek: Each node in `abi.json` has a name field. However this name field is not necessary the thing we expect. For example for generic arguments name is always `GenericTypeParam`. As there is no direct mapping between `node.Name` and `BaseDecl.Name` I think this might get a bit confusing long run.

Also, different decls will have different requirements about the name. For example for types we might need to get to the fully qualified path, the fully qualified path without module or just type name depending on context. We don't need this for e.g. `ArgumentDecl`.

I think we should drop `BaseDecl.Name` and replace it with more specialised name types on the relevant `Decls`.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -171,6 +111,7 @@ public static string GetCallArgumentString(Parameter parameter)
             return parameter switch
             {
                 { Type: "SwiftHandle" } => $"{parameter.Name}.Payload",
+                { Type: "IntPtr" } => parameter.Name,
```
kotlarmilos: How is this different from the default `_ => parameter.Name`?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -181,36 +122,48 @@ public class WrapperSignatureBuilder
...
         /// </summary>
         public void HandleReturnType()
         {
-            var argument = MethodDecl.CSSignature.First();
-            var typeRecord = TypeDatabase.GetTypeRecordOrAnyType(argument.SwiftTypeSpec);
-            SetReturnType(typeRecord.CSTypeIdentifier);
+            var argument = _env.MethodDecl.CSSignature.First();
+            if (argument.IsGeneric)
+            {
+                var CSName = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()];
```
kotlarmilos: Should `GenericTypeMapping` be part of the type database?

src/Swift.Bindings/src/Model/TypeDecl/MethodDecl.cs
```diff
@@ -29,10 +28,20 @@ public sealed record MethodDecl : BaseDecl
...
-		public required bool Throws { get; set; }
+        /// <summary>
+        /// Indicates if method can throw an exception.
+        /// </summary>
+        public required bool Throws { get; set; }
+
+        /// <summary>
+        /// Generic parameters of the method.
+        /// </summary>
+        public required List<string> GenericParameters { get; set; }
```
kotlarmilos: Is `GenericParameters` redundant with `CSSignature`?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -565,20 +515,36 @@ private void EmitIndirectResultMethod(IndentedTextWriter writer)
                 return;
             }
 
-            writer.WriteLine($"var payload = (SwiftHandle)NativeMemory.Alloc({_wrapperSignature.ReturnType}.PayloadSize);");
+            writer.WriteLine($"var returnMetadata = TypeMetadata.GetTypeMetadataOrThrow<{_wrapperSignature.ReturnType}>();");
+            writer.WriteLine($"var payload = (SwiftHandle)NativeMemory.Alloc(returnMetadata.Size);");
```
kotlarmilos: Is this memory deallocated?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: I dont think so, this is just a per MethodEnvironment mapping, outside word should not see it.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: Added deallocation

src/Swift.Bindings/src/Model/TypeDecl/MethodDecl.cs
jkurdek: `GenericParameters` is the list of generic type params in the order they appear in the function declaration, `Signature` contains a list of arguments (e.g. one generic type can appear multiple times in function signature)

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -130,15 +62,23 @@ public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
         public void Emit(IndentedTextWriter writer, IEnvironment env, Conductor conductor)
         {
             var methodEnv = (MethodEnvironment)env;
-            if (methodEnv.SignatureHandler.GetWrapperSignature().ContainsPlaceholder)
+            var signatureHandler = new SignatureHandler(methodEnv);
+
+            if (methodEnv.MethodDecl.GenericParameters.Any(x => x.Constraints.Count > 0))
```
kotlarmilos, line 67: We need to support these scenarios for the StoreKit2 purchase APIs.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -130,15 +62,23 @@ public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
         public void Emit(IndentedTextWriter writer, IEnvironment env, Conductor conductor)
         {
             var methodEnv = (MethodEnvironment)env;
-            if (methodEnv.SignatureHandler.GetWrapperSignature().ContainsPlaceholder)
+            var signatureHandler = new SignatureHandler(methodEnv);
```
kotlarmilos, line 65: What is the relation between env and the signature handler? I think there is a circular dependency.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -130,15 +62,23 @@ public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
...
             {
-                Console.WriteLine($"Method {methodEnv.MethodDecl.Name} has unsupported signature: ({methodEnv.SignatureHandler.GetWrapperSignature().ParametersString()}) -> {methodEnv.SignatureHandler.GetWrapperSignature().ReturnType}");
+                Console.WriteLine($"Method {methodEnv.MethodDecl.Name} has unsupported signature: ({signatureHandler.GetWrapperSignature().ParametersString()}) -> {signatureHandler.GetWrapperSignature().ReturnType}");
                 return;
             }
 
-            var wrapperEmitter = new WrapperEmitter(methodEnv);
-            wrapperEmitter.EmitMethod(writer);
-            PInvokeEmitter.EmitPInvoke(writer, methodEnv);
+            var wrapperEmitter = new WrapperEmitter(methodEnv, signatureHandler);
```
kotlarmilos, line 79: Should we keep all relevant information in the env and reduce the number of arguments in handlers?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: Is this just a placeholder name, like T?

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -181,36 +121,48 @@ public class WrapperSignatureBuilder
...
         public void HandleArguments()
         {
-            foreach (var argument in MethodDecl.CSSignature.Skip(1))
+            foreach (var argument in _env.MethodDecl.CSSignature.Skip(1))
             {
-                var typeRecord = TypeDatabase.GetTypeRecordOrAnyType(argument.SwiftTypeSpec);
-                AddParameter(typeRecord.CSTypeIdentifier, argument.Name);
+                if (argument.IsGeneric)
+                {
+                    var CSName = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()];
```
kotlarmilos: Does this support open generics only?

src/Swift.Bindings/src/Model/TypeDecl/ArgumentDecl.cs
```diff
@@ -22,5 +22,10 @@ public record ArgumentDecl : BaseDecl
         /// Indicates the inout annotation of the argument.
         /// </summary>
         public required bool IsInOut { get; set; }
+
+        /// <summary>
+        /// Indicates if the argument is generic.
+        /// </summary>
+        public required bool IsGeneric { get; set; }
```
kotlarmilos, line 29: Is this redundant with `SwiftTypeSpec`? The same is for `IsInOut`.

src/Swift.Bindings/src/Parser/GenericSignatureParser.cs
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) Microsoft Corporation.
```
kotlarmilos, line 1: I think there should be a matching rule for generics in the `Swift5Reducer`.  There is `GenericParameters` property in the `ConvertFunction`.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: Yes, I would like to introduce support for constraints in a separate PR. This line was introduced so that we do not wrongfully generate projections which are not working.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: I removed the signature handler from the env. Now it takes the env instead.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: If we reintroduce SignatureHandler to the environment it will need to take database, methodDecl, parentDecl etc separately to avoid circular dependency. I think this way the amount of params is minimized

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: Yes this is placeholder like `T`. And yes it supports only open generics. Closed generics currently behave like regular types.

src/Swift.Bindings/src/Model/TypeDecl/ArgumentDecl.cs
jkurdek: I think that `SwiftTypeSpec` is not able to understand whether it contains a generic param or not at the moment it will just hold the appropriate string. I believe it needs additional context, this is what this property is intended to do. I am referring to creating a new `TypeSpec` out of node's printed name - `TypeSpecParser.Parse(node.PrintedName);` where for generics `PrintedName` is equal to e.g. `τ_0_0`. 

I'm not sure about `IsInOut`

cc: @stephen-hawley

src/Swift.Bindings/src/Parser/GenericSignatureParser.cs
jkurdek: I would happy to drop this one if we already have a similar functionality. 

@stephen-hawley, should we grab the mangled name and run it through the demangler to parse the generic signature?

src/Swift.Bindings/src/Model/TypeDecl/ArgumentDecl.cs
stephen-hawley: TypeSpec should have generic parameter information in it. Be aware that there is no way to tell if a generic parameter is bound or unbound without more context to it. Swift generic parameters are labeled using a coordinate system of (depth, index) where depth is the distance from the current site to the declaration of the generic type. A depth of 0 means that the type is declared in the same site. > 0 means that the generic is declared in the enclosing site (or it's enclosing site...). The index is it's position in the actual declaration.

Example:
```swift
public class Foo<T>  {
    public var x: T // coordinate 0, 0 - same site as Foo, index 0

    public func doSomething<U>(a: U, b: T) { } // U is 0,0 T is 1, 0

    public struct Bar<X, Y> {
        public func other<A, Z, X>(a: A, b: Z, c: X, d: T) // A is 0, 0 Z is 0,1, X is 1,0 T 2, 0
    }
}
```
We can get away with using a naming convention that matches the coordinates system (`τ_depth_index`), but in order to correctly answer the question "is this a generic parameter" we also need the entire context.
}

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
jkurdek: this are three things:
- valid placeholder name like `T`
- variable name for holding metadata for `T`
- variable name for holding payload of value of type`T`

all of those are internal to the method

src/Swift.Bindings/src/Parser/GenericSignatureParser.cs
jkurdek: I took a look on doing the same using `Swift5Reducer` the resulting `FunctionReduction` has indeed generic parameters but they are simple `TypeSpecs` ignorant of the protocol constraints. I think the protocol constraints should not live on a `TypeSpec` but on a `MethodDecl` or `TypeDecl`.

I can imagine adding some additional information about protocol constraints to `FunctionReduction` if we decide that those should live there. I think we would lose the information about `sugared` type placeholders which might be nice to have from the UX perspective though.

src/Swift.Bindings/src/Model/TypeDecl/ArgumentDecl.cs
kotlarmilos: The `TypeSpec` contains `ContainsGenericParameters` bool and list `GenericParameters` of `TypeSpec`. Is this enough to contain information for projections?

src/Swift.Bindings/src/Parser/GenericSignatureParser.cs
kotlarmilos: Ok. Let's review this in Milestone 3.

src/Swift.Bindings/src/Model/TypeDecl/ArgumentDecl.cs
jkurdek: The `TypeSpec` you are referring to I believe is the one on `ArgumentDecl`, this one will have an empty `ContainsGenericParameters` even when `argDecl.IsGeneric == true`. This is because there is not enough context for the `TypeSpec` to know this.

kotlarmilos: Offline discussion -- We aim to retrieve all relevant information from the ABI file and expand decls. The TypeSpec will be reviewed after the Milestone 3.

kotlarmilos (approved):

PR #2948, opened 2025-01-21
[SwiftBindings] Add array support

stephen-hawley:
This pull request adds a binding for `Swift.Array`.

This binding demonstrates a fundamental difference between the design of core types in Swift vs. .NET.
Swift uses consistent "copy on write" semantics for pretty much everything. This means that if you write this in Swift:
```swift
var a1:[Int] = [1, 2, 3]
var a2:[Int] = a1
a2[0] = 17
print(a1[0])
print(a2[0])
```
you will see
```
1
17
```

This is clearly not what a C# developer would expect and as such this binding doesn't reflect that. Any operation that mutates an instance of the array will effectively throw away the old instance which isn't precisely ideal, but here we are.

This is another example of a generic binding but this is different from `SwiftOptional` in that there are two types of methods: non-mutating (ie, normal methods) and mutating methods. These have very different calling conventions.

For non-mutating methods, the instance gets passed in as the first argument (not using the self register) and the type metadata of the generic **parameter** gets passed in.
For mutating methods, a pointer to a copy of the instance gets passed in in the self register and the type metadata of the **array** gets passed in.

Like many Swift primitives, this type is...unusual. I think it's technically a value type (it sure is acting like one), but under the hood the actual implementation may be one of several types. As a default, it looks like it's an `NSArray` for most cases.

Question: should I include an `IDisposable` implementation here or do we want to put that into a common base class?

### DIFF
--- a/src/Swift.Runtime/src/Swift/SwiftArray.cs
+++ b/src/Swift.Runtime/src/Swift/SwiftArray.cs
@@ -1,0 +1,361 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.Swift;
+using Swift.Runtime;
+using Swift.Runtime.InteropServices;
+
+namespace Swift;
+
+/// <summary>
+/// Represents a Swift array
+/// </summary>
+/// <typeparam name="T">the element type</typeparam>
+public sealed class SwiftArray<T> : ISwiftObject, IList<T>
+{
+
+    SwiftHandle handle;
+
+    /// <summary>
+    /// Constructs a new SwiftArray from the given handle to a Swift array
+    /// </summary>
+    public SwiftArray(SwiftHandle handle)
+    {
+        this.handle = handle;
+    }
+
+
+    /// <summary>
+    /// Constructs a new empty SwiftArray
+    /// </summary>
+    public SwiftArray()
+        : this(PInvokesForSwiftArray.Init(ElementTypeMetadata))
+    {
+    }
+
+    /// <summary>
+    /// Constructs a new SwiftArray and populates it with the given list
+    /// </summary>
+    /// <param name="list">a list of items to populate the array</param>
+    public SwiftArray(IList<T> list)
+        : this()
+    {
+        AddRange(list);
+    }
+
+    /// <summary>
+    /// Constructs a new SwiftArray and populates it with the enumeration
+    /// </summary>
+    /// <param name="collection">an enumeration of items to populate the array</param>
+    public SwiftArray(IEnumerable<T> collection)
+        : this()
+    {
+        AddRange(collection);
+    }
+
+    /// <summary>
+    /// Constructs a new SwiftArray and populates it with the given array
+    /// </summary>
+    /// <param name="items">an array of items to populate the array</param>
+    public SwiftArray(params T[] items)
+        : this()
+    {
+        AddRange(items);
+    }
+
+    /// <summary>
+    /// Returns the type metadata for the element
+    /// </summary>
+    static TypeMetadata ElementTypeMetadata
+    {
+        get => TypeMetadata.GetTypeMetadataOrThrow<T>();
+    }
+
+    /// <summary>
+    /// Returns the size of the element in bytes
+    /// </summary>
+    static unsafe nuint ElementSize
+    {
+        get => ElementTypeMetadata.ValueWitnessTable->Size;
+    }
+
+    /// <inheritdoc/>
+    static TypeMetadata ISwiftObject.GetTypeMetadata()
+    {
+        return TypeMetadata.Cache.GetOrAdd(typeof(SwiftArray<T>), _ =>
+                PInvokesForSwiftArray._MetadataAccessor(TypeMetadataRequest.Complete, ElementTypeMetadata));
+    }
+
+    /// <inheritdoc/>
+    static ISwiftObject ISwiftObject.NewFromPayload(SwiftHandle payload)
+    {
+        return new SwiftArray<T>(payload);
+    }
+
+    /// <inheritdoc/>
+    IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
+    {
+        var metadata = SwiftObjectHelper<SwiftArray<T>>.GetTypeMetadata();
+        unsafe
+        {
+            metadata.ValueWitnessTable->InitializeWithCopy((void*)swiftDest, (void*)handle, metadata);
+        }
+        return swiftDest;
+    }
+
+    /// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


matouskozak (commented):
Based on https://developer.apple.com/documentation/swift/array, array is a `@frozen` generic struct. Do we plan to remove this projection code (`SwiftArray.cs`) when the tooling is able to successfully project any `@frozen` generic struct or is the plan that selected Swift types have special projections like this?

I'm just curious about the long-term plan. I think it makes sense for these "often-used" types to have special projections if they need to be handled with caution to satisfy both Swift constraints and .NET developers.

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,361 @@
...
+using Swift.Runtime;
+using Swift.Runtime.InteropServices;
+
+namespace Swift;
+
+/// <summary>
+/// Represents a Swift array
+/// </summary>
+/// <typeparam name="T">the element type</typeparam>
+public sealed class SwiftArray<T> : ISwiftObject, IList<T>
```
kotlarmilos, line 19: How memory is handled? Who owns the memory?

kotlarmilos:
> Question: should I include an IDisposable implementation here or do we want to put that into a common base class?

I wouldn't include it unless it's actually used. Let's review the UX at some point (added to the plan).

kotlarmilos:
Could you add the manual bindings to the type database xml?

src/Swift.Runtime/src/Swift/SwiftArray.cs
jkurdek, line 19: Array conforms to a bunch of protocols in Swift. Assuming that we would like to have protocol constraints projected as generic constraints on c# side, how are we gonna make this work?

E.g. to call https://developer.apple.com/documentation/storekit/product/products(for:) with `SwiftArray` we will need to be able to somehow model `Collection<T>` conformance

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,361 @@
...
+{
+
+    SwiftHandle handle;
+
+    /// <summary>
+    /// Constructs a new SwiftArray from the given handle to a Swift array
+    /// </summary>
+    public SwiftArray(SwiftHandle handle)
```
jkurdek, line 27: Should this be public?

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,361 @@
...
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sS2ayxGycfC")]
+    public static extern SwiftHandle Init(TypeMetadata typeMetadata);
+
+    [UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sSayxSicig")]
+    public static unsafe extern void Get(SwiftIndirectResult result, nint index, SwiftHandle handle, TypeMetadata elementMetadata);
+
+    [UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sSayxSicis")]
+    public static unsafe extern void Set(SwiftSelf self, SwiftHandle value, nint index, TypeMetadata elementMetadata);
```
jkurdek, line 341: I think a bit less confusing order would be `SwiftHandle handle, nint index, TypeMetadata elementMetadata, SwiftSelf self`. Current version will compile anyway as `self` will go into register and be removed from parameter list before the call, but I think it would be nice to stay consistent with how this looks in llvm-ir where `self` is the last parameter. (we enforce that for `SwiftSelf<T>` as well) Also applies to others.

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,361 @@
...
+{
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sSaMa")]
+    public static extern TypeMetadata _MetadataAccessor(TypeMetadataRequest request, TypeMetadata typeMetadata);
+
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sS2ayxGycfC")]
+    public static extern SwiftHandle Init(TypeMetadata typeMetadata);
+
+    [UnmanagedCallConv(CallConvs = [typeof(CallConvSwift)])]
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sSayxSicig")]
+    public static unsafe extern void Get(SwiftIndirectResult result, nint index, SwiftHandle handle, TypeMetadata elementMetadata);
```
jkurdek, line 337: >For non-mutating methods, the instance gets passed in as the first argument (not using the self register) and the type metadata of the generic parameter gets passed in.

The mutating functions look as expected, but the non-mutating are truly interesting. Her a pointer to the copy is technically the second argument. This is something that generated projections would get terribly wrong. Is there any common pattern to the functions which behave like this on other types? Or each case requires manual investigation?

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: > E.g. to call https://developer.apple.com/documentation/storekit/product/products(for:) with SwiftArray we will need to be able to somehow model Collection<T> conformance

Do we need to model `Collection<T>` for this use-case? Is this requirement applied to the method where the conformance is needed? Is the conformance already available in the dynamic library?

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: There is `"funcSelfKind": "NonMutating"` property in the ABI.

src/Swift.Runtime/src/Swift/SwiftArray.cs
jkurdek: I think it depends on how we want the projection to look like. 

The pattern used by `products(for:) calls for something like this (I skip async)
```csharp
public Product products<T> (T identifiers) where T : ICollection<Swift.String>
```

I guess we could not map ICollection into c#, and have

```csharp
public Product products<T> (T identifiers)
```

but then we would have to hardcode some stuff:
- assumption that anything which goes into the function is valid
- we would still somehow need to obtain the protocol conformance descriptor and PWT

kotlarmilos: > assumption that anything which goes into the function is valid

I would assume the input is valid until focus on protocols support.

> we would still somehow need to obtain the protocol conformance descriptor and PWT

Since the protocol's mangled name and generic type are known at runtime, can we retrieve the conformance using Swift runtime API?

kotlarmilos:
Closing this PR in favour of https://github.com/dotnet/runtimelab/pull/2964.

Issue #2837, opened 2024-12-03
[Swift language features] Implement Swift async


kotlarmilos:
## Description

Swift async function is projected into a function with C# Task. For each Swift function, there is a Swift wrapper function that calls into C# callback upon completion of the async operation.

The existing approach generate a wrapper Swift function for each async function.

Follow-up work for Swift async: https://github.com/dotnet/runtimelab/issues/2950

PR #2949, opened 2025-01-22
[Swift language features] Implement Swift async method bindings

kotlarmilos:
## Description

This PR implements bindings for Swift async methods. It introduces another string writer for Swift wrappers, adds a rule to the demangler to detect whether a method is async, and implements the necessary handling.

Fixes https://github.com/dotnet/runtimelab/issues/2837

## Tasks
 - [x] Implement Swift signature helper
 - [x] Implement NameProvider methods for delegate and gchandle
 - [x] Add more tests


mcumming (approved):

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -55,10 +55,10 @@ public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
         /// <summary>
         /// Emits the code for the specified environment.
         /// </summary>
-        /// <param name="writer">The IndentedTextWriter instance.</param>
+        /// <param name="csWriter">The IndentedTextWriter instance.</param>
         /// <param name="env">The environment.</param>
         /// <param name="conductor">The conductor instance.</param>
-        public void Emit(IndentedTextWriter writer, IEnvironment env, Conductor conductor)
+        public void Emit(IndentedTextWriter csWriter, IndentedTextWriter swiftWriter, IEnvironment env, Conductor conductor)
```
jkurdek: I think it might be good idea to introduce some strong typed version of `csWriter` and `swiftWriter` they can just wrap `IndentedTextWriter` for now but it will reduce probability of passing one when another is expected

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
```diff
@@ -527,66 +546,125 @@ private void EmitSwiftSelf(IndentedTextWriter writer)
...
+
+            text = $$"""
+            extension {{_methodDecl.ParentDecl!.Name}} {
+                @_silgen_name("{{_methodDecl.Name}}_async")
+                public func {{NameProvider.GetPInvokeName(_methodDecl)}}_async({{string.Join(", ", new[] {
+                                                                                    $"callback: @escaping ({_methodDecl.CSSignature.First().SwiftTypeSpec}) -> Void"
+                                                                                }.Concat(
+                                                                                    _methodDecl.CSSignature.Skip(1)
+                                                                                        .Select(p => p.Name + ": " + p.SwiftTypeSpec)
+                                                                                ))}}) {
```
jkurdek: NIT: I think extracting this to separate variables would increase readability

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -359,20 +365,24 @@ private MethodDecl CreateMethodDecl(Node node, BaseDecl parentDecl, ModuleDecl m
...
-                MangledName = node.Kind == "Constructor" ? PatchMangledName(node.MangledName) : node.MangledName,
+                MangledName = mangledName,
                 MethodType = node.@static ?? false ? MethodType.Static : MethodType.Instance,
                 IsConstructor = node.Kind == "Constructor",
                 CSSignature = new List<ArgumentDecl>(),
                 ParentDecl = parentDecl,
                 ModuleDecl = moduleDecl,
-                Throws = node.throwing ?? false
+                Throws = node.throwing ?? false,
+                IsAsync = (reduction as FunctionReduction)?.Function?.IsAsync ?? false
```
jkurdek: I think that we should error out if reduction is not FunctionReduction in this scenario

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: Do you suggest creating a Swift signature wrapper?

jkurdek: I think it might be a good idea in the future when there is more variance to the generated signatures. But I'm not sure we need this now.  Currently I just suggest extracting computation from the string interpolation.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: Updated.

src/Swift.Bindings/src/Emitter/StringCSharpEmitter/Handler/MethodHandler.cs
kotlarmilos: Added.

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
kotlarmilos: Added a tracking issue: https://github.com/dotnet/runtimelab/issues/2954. Currently, the demangler fails when an unknown node is encountered (such as `static`) and it affects the existing test cases.

src/docs/swift-code-generation.md
```diff
@@ -0,0 +1,71 @@
...
+            {
+                _callbackHandle.Value.Free();
+                _callbackHandle = null;
+            }
+        }
+    };
+    _callbackHandle = GCHandle.Alloc(callbackDelegate);
+    var self = new SwiftSelf((void*)_payload);
+    
+    waitFor(Marshal.GetFunctionPointerForDelegate(callbackDelegate), IntPtr.Zero, seconds, self);
```
matouskozak:
```suggestion
    waitFor_async(Marshal.GetFunctionPointerForDelegate(callbackDelegate), IntPtr.Zero, seconds, self);
```

I think this should call `waitFor_async` otherwise the signatures don't match.

jkotas: Can we use UnmanagedCallersOnly and function pointers instead of marshalled delegates? Marshalled delegates have much higher overhead and they come with limitations on platforms that disallow JIT.

kotlarmilos: `waitFor_async` is a mangled name from the dynamic library we set using `@_silgen_name`. We still use the original PInvoke name from C#.

kotlarmilos: Here is an example:
```csharp
public unsafe Task<Int32> waitFor( UInt64 seconds)
{
    ...
    PInvoke_waitFor(Marshal.GetFunctionPointerForDelegate(callbackDelegate), IntPtr.Zero, seconds, self);
    ...
}

[UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
[DllImport("/Users/miloskotlar/dotnet/runtimelab-swift-net10/artifacts/bin/Swift.Bindings.Integration.Tests/Debug/net9.0/Structs/libStructsTests.dylib", EntryPoint = "waitFor_async")]
private static extern void PInvoke_waitFor(IntPtr callback, IntPtr context,  UInt64 seconds,  SwiftSelf self);
```

The mangled name is `waitFor_async` while C# PInvoke doesn't change.

matouskozak: Ah, I understand now. I got confused that the example is calling `waitFor` instead of `PInvoke_waitFor`. Thanks for the example.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -624,130 +691,167 @@ private void EmitSwiftError(IndentedTextWriter writer)
...
+            EmitBodyEnd(csWriter, swiftWriter);
         }
 
         /// <summary>
         /// Emits the body start.
         /// </summary>
-        /// <param name="writer">The IndentedTextWriter instance.</param>
-        private void EmitBodyStart(IndentedTextWriter writer)
+        /// <param name="csWriter">The IndentedTextWriter instance.</param>
+        private void EmitBodyStart(CSharpWriter csWriter, SwiftWriter swiftWriter)
```
jkurdek: Those methods don't really do anything with `swiftWriter`, maybe we should move them to `CSharpWriter` altogether

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -624,130 +691,167 @@ private void EmitSwiftError(IndentedTextWriter writer)
...
             }
         }
 
         /// <summary>
         /// Emits the return statement for the method.
         /// </summary>
-        /// <param name="writer">The IndentedTextWriter instance.</param>
-        private void EmitReturnMethod(IndentedTextWriter writer)
+        /// <param name="csWriter">The IndentedTextWriter instance.</param>
+        private void EmitReturnMethod(CSharpWriter csWriter, SwiftWriter swiftWriter)
```
jkurdek: This is becoming quite nested as nested in #2955. We should simplify it. 

Can a method be async and return indirect result?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: Yes, we should remove it when not used. Alternatively, create a single `StringWriter` type with two `IndentedTextWriter`'s.

kotlarmilos:
This is the first attempt to simplify the string-based emitter for methods when: **(1)** methods are similar (instance methods and constructors); **(2)** methods significantly differ (non-async and async methods).

Based on the discussion in https://github.com/dotnet/runtimelab/issues/2955:
 - Wrapper and PInvoke builders are parameterized to be reusable
 - Constructors are moved to a handler that extends the default method handler and overrides only the necessary building blocks
 - Async methods are moved to a handler that implements separate async flow using default method builders

These changes improve extensibility and allow for more localized changes. This is not the final proposal and will be improved in the future.

/cc: @jkurdek I might have overridden some of your changes. If needed, we can revert them.

src/Swift.Bindings/src/Emitter/IEmitter.cs
```diff
@@ -6,7 +6,7 @@ namespace BindingsGeneration
     /// <summary>
     /// Represents an interface for emitting C# source code.
```
jkurdek, line 7: I think comments need to change as well :)

src/Swift.Bindings/src/Marshaler/MarshallingHelpers.cs
```diff
@@ -5,23 +5,23 @@ namespace BindingsGeneration
 {
     public static class MarshallingHelpers // TODO: Find better place for those
     {
-        public static bool MethodRequiresIndirectResult(MethodEnvironment env)
+        public static bool MethodRequiresIndirectResult(MethodDecl decl, ITypeDatabase typeDatabase)
```
jkurdek: I think those should become extension methods on `MethodEnv`. If a method needs to accept typeDatabase and and decl it looks like it is making a decision on the based on `MethodEnv`

src/Swift.Bindings/src/Marshaler/IHandler.cs
```diff
@@ -16,15 +16,15 @@ public interface IHandler
         /// </summary>
         /// <param name="baseDecl">The base declaration.</param>
         /// <returns>The environment corresponding to the base declaration.</returns>
-        IEnvironment Marshal(BaseDecl baseDecl, ITypeDatabase typeDatabase);
+        IEnvironment? Marshal(BaseDecl baseDecl, ITypeDatabase typeDatabase);
```
jkurdek: What does it mean that an `IEnvironment` is nullable? That we failed to marshal a method? I would be very cautious about introducing new nullables, this can result in unexpected behaviour and hard to debug bugs

src/Swift.Bindings/src/Marshaler/IEnvironment.cs
```diff
@@ -84,8 +91,48 @@ public class MethodEnvironment(MethodDecl methodDecl, ITypeDatabase typeDatabase
...
+
+        /// <summary>
+        /// Gets the return type.
+        /// </summary>
+        public string ReturnType { get; set; } = string.Empty;
+
+        /// <summary>
+        /// Gets a value indicating whether the method requires an indirect result.
+        /// </summary>
+        public bool RequiresIndirectResult { get; set ; } = false;
```
jkurdek: I like the direction here, I think the environment is a good place to keep this information.

 I am a bit worried about the default values though. They are not "default", they indicate choosing a certain emission strategy. I think we should not assign the default values but make the marshalling logic set the correct values on init. This way we will only interact with valid instances of MethodEnvironment.

src/Swift.Bindings/src/Marshaler/Conductor.cs
```diff
@@ -28,6 +28,8 @@ public class Conductor
         ];
         private readonly List<IFieldHandlerFactory> _fieldHandlerFactories = [];
         private readonly List<IMethodHandlerFactory> _methodHandlerFactories = [
+            new MethodAsyncHandlerFactory(),
```
jkurdek: I am worried about this pattern. One of the points you noticed in #2955 is that the handlers had to be executed in hard-coded order. This repeats the same issue. `MethodHandlerFactory` will handle any function so if it executed before `MethodConstructorHandlerFactory` or `MethodAsyncHandlerFactory` we get a wrong projection.

src/Swift.Bindings/src/Emitter/StringEmitter/TextWriter/IndentedTextWriter.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SwiftWriter"/> class.
+        /// </summary>
+        /// <param name="writer">The writer.</param>
+        public SwiftWriter(StringWriter writer) : base(writer)
+        {
+        }
+    }
+}
```
jkurdek, line 35: NIT: missing empty line at the end of the file

src/Swift.Bindings/src/Emitter/StringEmitter/ModuleEmitter.cs
```diff
@@ -9,7 +9,7 @@ namespace BindingsGeneration
     /// <summary>
     /// Represents an string-based C# emitter.
```
jkurdek, line 10: NIT: comment

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -188,55 +188,55 @@ public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
...
+            csWriter.WriteLine($"public unsafe class {structDecl.Name} : IDisposable, {typeof(ISwiftObject).Name}");
+            csWriter.WriteLine("{");
+            csWriter.Indent++;
 
-            WritePrivateFields(writer, structDecl);
-            WriteDisposeMethod(writer);
-            WriteFinalizer(writer, structDecl);
-            WritePayloadSize(writer);
-            WritePayload(writer);
+            WritePrivateFields(csWriter, swiftWriter, structDecl);
```
jkurdek: I think that none of the functions here actually uses swiftWriter.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/Method/MethodHandler.cs
```diff
@@ -380,374 +432,186 @@ public Signature GetWrapperSignature()
...
         }
 
         /// <summary>
         /// Emits the return statement for the method.
         /// </summary>
-        /// <param name="writer">The IndentedTextWriter instance.</param>
-        private void EmitReturnMethod(IndentedTextWriter writer)
+        /// <param name="env">The method environment.</param>
+        /// <param name="returnType">The return type.</param>
+        public virtual string EmitReturnMethod(MethodEnvironment env, string returnType)
```
jkurdek: NIT: Naming this function doesnt emit anything anymore

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/Method/MethodHandler.cs
```diff
@@ -44,42 +43,99 @@ public MethodHandler()
...
+                swiftSelf = methodEnv.ParentDecl.Name;
 
-            var wrapperEmitter = new WrapperEmitter(methodEnv, signatureHandler);
-            wrapperEmitter.Emit(writer);
-            PInvokeEmitter.EmitPInvoke(writer, methodEnv, signatureHandler);
-            writer.WriteLine();
+            var voidReturn = methodEnv.MethodDecl.CSSignature.First().SwiftTypeSpec.IsEmptyTuple;
+            var hasReturn = !(methodEnv.RequiresIndirectResult || voidReturn);
+
+            csWriter.WriteLine($$"""
```
jkurdek: This changes the approach from passing the code writer and emitting code on the way to recursively building a string from subnodes. I think this might be a valid approach to discuss further.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/Method/MethodHandler.cs
```diff
@@ -380,374 +432,186 @@ public Signature GetWrapperSignature()
...
-            if (_requiresIndirectResult)
+            if (env.RequiresIndirectResult)
             {
-                writer.WriteLine($"return SwiftMarshal.MarshalFromSwift<{_wrapperSignature.ReturnType}>((SwiftHandle)swiftIndirectResult.Value);");
+                return $"return SwiftMarshal.MarshalFromSwift<{returnType}>((SwiftHandle)swiftIndirectResult.Value);";
             }
             else
             {
-                if (_env.MethodDecl.CSSignature.First().SwiftTypeSpec.IsEmptyTuple)
+                if (returnType == "void")
```
jkurdek: We should have a one correct way of checking whether the return is void.

jkurdek (changes requested):
I think you have identified problems within `MethodHandler` correctly. I agree that we should do the computation during marshalling not during emission. Code emission should just look at some API and emit code. 

I disagree with the solution though. I think that introducing an inheritance type hierarchy increases the complexity and makes the code much harder to debug. I would recommend moving the Emitter redesign changes to a separate PR and having a design discussion so not to block the async work necessary for milestone 1.

I would love to see the opinions of others @vitek-karas, @matouskozak, @mcumming

kotlarmilos:
> I disagree with the solution though. I think that introducing an inheritance type hierarchy increases the complexity and makes the code much harder to debug. I would recommend moving the Emitter redesign changes to a separate PR and having a design discussion so not to block the async work necessary for milestone 1.

> I would love to see the opinions of others @vitek-karas, @matouskozak, @mcumming

Ok, let's collect input and schedule a meeting. I will revert the last commit and make changes required for the basic async support.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: Passed as an argument only when used.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: Simplified according to https://github.com/dotnet/runtimelab/issues/2955.

> Can a method be async and return indirect result?

No, the return type is always a C# Task.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -624,130 +691,164 @@ private void EmitSwiftError(IndentedTextWriter writer)
...
         }
 
         /// <summary>
         /// Emits the method signature.
         /// </summary>
         /// <param name="writer">The IndentedTextWriter instance.</param>
-        private void EmitSignatureMethod(IndentedTextWriter writer)
+        private void EmitSignatureMethod(CSharpWriter csWriter)
         {
+            if (_requiresSwiftAsync)
```
jkurdek, line 771: is there a reason this lives here and not in a separate method. I am not sure how it is related to `Signature` emission

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -589,28 +639,45 @@ private void EmitGenericArguments(IndentedTextWriter writer)
...
+            {
+                string parameters = string.Join(", ",
+                    new[]
+                    {
+                        $"(IntPtr)s_{_env.MethodDecl.Name}Callback",
+                        "IntPtr.Zero",
+                        "GCHandle.ToIntPtr(handle)",
+                        _pInvokeSignature.CallArgumentsString()
+                    }.Where(arg => !string.IsNullOrEmpty(arg))
```
jkurdek: I think this can be moved to Signature Builder. So that `_pInvokeSignature.CallArgumentsString()` always returns a correct list.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -524,21 +543,52 @@ private void EmitSwiftSelf(IndentedTextWriter writer)
...
+            var text = $$"""
+            TaskCompletionSource{{(_env.MethodDecl.CSSignature.First().SwiftTypeSpec.IsEmptyTuple ? "" : $"<{_wrapperSignature.ReturnType}>")}} task = new TaskCompletionSource{{(_env.MethodDecl.CSSignature.First().SwiftTypeSpec.IsEmptyTuple ? "" : $"<{_wrapperSignature.ReturnType}>")}}();
+            GCHandle handle = GCHandle.Alloc(task, GCHandleType.Normal);
+            """;
+            csWriter.WriteLines(text);
+
+            string parameters = string.Join(", ", new[] {$"callback: @escaping ({(_env.MethodDecl.CSSignature.First().SwiftTypeSpec.IsEmptyTuple ? "" : $"{_env.MethodDecl.CSSignature.First().SwiftTypeSpec}, ")}Int64) -> Void", $"task: Int64"}.Concat(_env.MethodDecl.CSSignature.Skip(1).Select(p => p.Name + ": " + p.SwiftTypeSpec)));
+
+            text = $$"""
+            extension {{_env.MethodDecl.ParentDecl!.Name}} {
```
jkurdek: _env already has a non-null accessor for ParentDecl

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -524,21 +543,52 @@ private void EmitSwiftSelf(IndentedTextWriter writer)
...
+        /// Emits the Async task.
+        /// </summary>
+        /// <param name="csWriter">The IndentedTextWriter instance.</param>
+        private void EmitAsync(CSharpWriter csWriter, SwiftWriter swiftWriter)
+        {
+            if (!_requiresSwiftAsync)
+                return;
+
+            var text = $$"""
+            TaskCompletionSource{{(_env.MethodDecl.CSSignature.First().SwiftTypeSpec.IsEmptyTuple ? "" : $"<{_wrapperSignature.ReturnType}>")}} task = new TaskCompletionSource{{(_env.MethodDecl.CSSignature.First().SwiftTypeSpec.IsEmptyTuple ? "" : $"<{_wrapperSignature.ReturnType}>")}}();
```
jkurdek: This one is on me.  I think it might be good idea to have an function (probably on MethodDecl, or as an extension method) which tells us whether the return is `void`

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -524,21 +543,52 @@ private void EmitSwiftSelf(IndentedTextWriter writer)
...
+        /// <summary>
+        /// Emits the Async task.
+        /// </summary>
+        /// <param name="csWriter">The IndentedTextWriter instance.</param>
+        private void EmitAsync(CSharpWriter csWriter, SwiftWriter swiftWriter)
+        {
+            if (!_requiresSwiftAsync)
+                return;
+
+            var text = $$"""
```
jkurdek: There are multiple `text` being defined in one scope. Should they have unique names?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -385,22 +385,38 @@ internal static class PInvokeEmitter
...
-            writer.WriteLine("[UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]");
-            writer.WriteLine($"[DllImport(\"{libPath}\", EntryPoint = \"{methodDecl.MangledName}\")]");
-
             var pInvokeSignature = signatureHandler.GetPInvokeSignature();
 
-            writer.WriteLine($"private static extern {pInvokeSignature.ReturnType} {pInvokeName}({pInvokeSignature.ParametersString()});");
+            csWriter.WriteLine("[UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]");
+            if (methodDecl.IsAsync)
+            {
+                string parameters = string.Join(", ",
```
jkurdek: This can go into SignatureBuilder. EntryPoint could go into NameProvider, this way this function wouldn't even need to have an if statement

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: Moved to a separate method.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: Added specific handler in the PInvokeSignature.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: Added specific handler in the PInvokeSignature.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: I simplified this a bit. I suggest we keep it until we figure out how to refactor the emitters.

jkurdek (approved):
LGTM!

PR #2953, opened 2025-01-22
[NativeAOT-LLVM] Don't anchor functions to source-based CUs

SingleAccretion:
So far, we have been producing object files that split functions across CUs according to the source files in which those functions are defined, leading to many small CUs on each object file.

This has no practical benefit.

It has the practical downside of making debug info larger, about 5% for the .debug_info section.

Switch to the more traditional "one object file <=> one CU" scheme.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks

PR #2958, opened 2025-01-24
[SwiftBindings] Add ProtocolWitnessTable metadata

### LINKED ISSUE #2957: [Swift language features] Implement Swift Protocol Witness Table retrieval
To move forward with protocol constraints we need to add runtime mechanism for retrieval of protocol witness tables. This should be accompanied by a caching mechanism similar to the one we have for `TypeMetadata` 

In order to obtain a Protocol Witness Table we must:
1. Obtain symbol for Protocol Conformance Descriptor
2. Read Protocol Conformance Descriptor from native library
3. Call `get_witnessTable` function to correctly construct a PWT

jkurdek:
Fixes #2957 

This PR adds Swift.Runtime metadata for obtaining Protocol Conformance Descriptor and Protocol Witness Table. The implementation is similar to the one of TypeMetadata.

The process of obtaining a PWT is the following:
1. Obtain symbol for Protocol Conformance Descriptor 
2. Read Protocol Conformance Descriptor from native library
3. Call get_witnessTable function to correctly construct a PWT

Step 1 will be performed by the MachO during projection generation. Steps 2 and 3 happen during runtime. 

A new method on `ISwiftObject` is introduced - `GetProtocolConformanceDescriptor`. The intention is that the type will know the symbols for its conformance descriptors and will call a function to load them from the native library.

An example implementation of `GetProtocolConformanceDescriptor` implementation on type implementing`ISwiftObject`  might look like follows:

```csharp
static ProtocolConformanceDescriptor GetProtocolConformanceDescriptor<U>()
{
    var symbols = new Dictionary<Type, string>
    {
        { typeof(ISwiftHashable), "symbolForSwiftHashableDescriptor" } // This is known at binding time
    };

    if (!symbols.TryGetValue(typeof(U), out var symbol))
    {
        ...
    }

    return ProtocolConformanceDescriptor.LoadProtocolConformanceDescriptor(symbol, "someLib.dylib");
}
```

TODO:
- Add cache for PWT (possibly out of scope for this PR) #2962


src/Swift.Runtime/src/Swift/Runtime/ISwiftObject.cs
```diff
@@ -23,6 +23,11 @@ public interface ISwiftObject
     /// Marshals this object to a Swift destination
     /// </summary>
     unsafe IntPtr MarshalToSwift(IntPtr swiftDest);
+
+    /// <summary>
+    /// Gets the protocol conformance descriptor for the given type
+    /// </summary>
+    public static abstract ProtocolConformanceDescriptor GetProtocolConformanceDescriptor<U>();
```
vitek-karas: Just curious - in general I'm always a bitty worried by introducing lot of new generics, as they hurt AOT code gen (both mono and NativeAOT) if they're over value types. In this case, the U is the protocol, so I assume in C# it should always be an interface right? (so reference type) If that is the case, we might want to constrain U to reference type (in which case it doesn't hurt AOT almost at all).

vitek-karas: I would rename `U` to something like `TProtocol` - to make it self-documenting what is expected to be there.

src/Swift.Runtime/src/Swift/AnyType.cs
```diff
@@ -46,10 +47,22 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
+            metadata.ValueWitnessTable->InitializeWithCopy((void*)swiftDest, (void*)_payload, metadata);
         }
         return swiftDest;
     }
 
+    /// <summary>
+    /// Gets the protocol conformance descriptor for the given type
+    /// </summary>
+    /// <typeparam name="U"></typeparam>
+    /// <returns></returns>
```
vitek-karas: Just asking about the coding style - do we copy the full XML comments for overrides?

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+
+namespace Swift.Runtime;
+
+public readonly struct ProtocolConformanceDescriptor : IEquatable<ProtocolConformanceDescriptor>
```
vitek-karas: Typically, the coding style is to have one class per file...

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
```diff
@@ -0,0 +1,105 @@
...
+    public readonly static ProtocolConformanceDescriptor Zero = default;
+
+    public bool IsValid => _handle != IntPtr.Zero;
+
+    public bool Equals(ProtocolConformanceDescriptor other)
+    {
+        return _handle == other._handle;
+    }
+
+    internal static bool TryGetProtocolConformanceDescriptor<T, U>([NotNullWhen(true)] out ProtocolConformanceDescriptor? result)
```
vitek-karas: I know we discussed this already, but this is not NativeAOT compatible - it will produce warnings for sure.
It's possible the compiler is clever enough to figure this out and avoid runtime problems, but I'm not sure about it.

Just curious - we're making this generic on T so that we can call the static interface method on ISwiftObject, right?

I have to admit that this makes me nervous :-)
Could we start a separate issue/discussion about possible solutions to this? And run it by runtime people?

vitek-karas: Naming - I would use more descriptive names for the generic type parameters `TType`, `TProtocol` for example.

vitek-karas: Just curious - what is the scenario where this is going to be called on something which is not `ISwiftObject` ?

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public static ProtocolConformanceDescriptor LoadProtocolConformanceDescriptor(string libraryPath, string symbolName)
+    {
+        var libraryHandle = NativeLibrary.Load(libraryPath);
+
+        if (libraryHandle == IntPtr.Zero)
+        {
+            throw new SwiftRuntimeException($"Unable to load library: {libraryPath}");
+
```
vitek-karas: Nit: empty line.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
```diff
@@ -0,0 +1,105 @@
...
+            var helperType = typeof(ProtocolConformanceDescriptorHelper<,>).MakeGenericType(typeof(T), typeof(U));
+            result = (ProtocolConformanceDescriptor)helperType.GetMethod("GetProtocolConformanceDescriptor")!.Invoke(null, null)!;
+            return true;
+        }
+
+        result = null;
+        return false;
+    }
+
+    public static ProtocolConformanceDescriptor LoadProtocolConformanceDescriptor(string libraryPath, string symbolName)
```
vitek-karas: Where are we going to get the `libraryPath` from?
There are potential security implications when loading code from file paths... so we should be careful.
If this is supposed to be just like PInvokes, then we should use the mechanism from the other overload of `NativeLibrary.Load`.

vitek-karas: Naming nit. The callsite as is would look like:
```C#
var pcd = ProtocolConformanceDescriptor.LoadProcotolConformanceDescriptor(libraryPath, "mangled_name");
```

Typically we want to avoid repeating the name of the class in a factory method, as it's already there in the type name. So either `Load` or `LoadFromSymbol` or something like that.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
```diff
@@ -0,0 +1,105 @@
...
+    public readonly static ProtocolWitnessTable Zero = default;
+
+    public bool IsValid => _handle != IntPtr.Zero;
+
+    public bool Equals(ProtocolWitnessTable other)
+    {
+        return _handle == other._handle;
+    }
+
+    public ProtocolWitnessTable GetProtocolWitnessTable<T, U>()
```
vitek-karas: This should be static....

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            throw new SwiftRuntimeException($"Unable to get protocol conformance descriptor for {typeof(T)} and {typeof(U)}");
+        }
+
+        return GetProtocolWitnessTable(conformanceDescriptor.Value, metadata);
+    }
+
+    ProtocolWitnessTable GetProtocolWitnessTable(ProtocolConformanceDescriptor conformanceDescriptor, TypeMetadata typeMetadata)
+    {
+
```
vitek-karas, line 126: Nit: empty line

vitek-karas (commented):
Looks good overall with one complain - no tests. I would expect it to not be that difficult to write a unit test which does all of the interesting things (we can hardcode the symbol into the test by hand, we don't need the MachO reader for it now).

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
jkurdek: So the `libraryPath` will be always known at compile time (this is the path of the dylib the type is declared in). What would be the safest way to obtain it using the symbol? I used `NativeLibrary.Load` instead of `PInvoke` as the` Protocol Conformance Descriptor` is not a function within the dylib as opposed to the `Metadata Accessor`.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
jkurdek: csharp-only types which implement a projected Swift protocol. If we support this scenario we will need to somehow obtain the descriptor of the wrapper here. This is a future problem and we don't have a design yet in place.

jkurdek: I don't think it produces any NativeAOT warnings. This project also has NativeAOT warnings enabled so we should see something happening.

The key problem we are trying to solve for variety of functions in this repo is the following. We need to execute some behavior (load metadata, create an instance of a C# object from swift handle etc.). This behavior is dispatched by the type (we might not have any valid "object" to work with). We need to be able to do that in generic context. And lastly, we need to handle things which come from Swift (ISwiftObject) and things which do not (primitive types, C# only types conforming to some protocol etc.) so there is no common interface we can slap on everything. 

The pattern used in this repo is to have helpers and use `MakeGenericType` which is NativeAOT compatible to simulate the following construct which is not supported by C#

```csharp
void DoSomething<T>()
{
    if (T is ISwiftObject) return T.DoSomething()
    else ...
}
```

I will open a separate issue to track the discussion.

cc: @MichalStrehovsky

src/Swift.Runtime/src/Swift/AnyType.cs
jkurdek: I think we haven't agreed on coding style in this repo yet. It will be good to have a discussion and agree on the coding style. For this particular comment I don't have a preference. 

\cc @kotlarmilos

src/Swift.Runtime/src/Swift/Runtime/ISwiftObject.cs
jkurdek: Yes, I added the class constraint

src/Swift.Runtime/src/Swift/AnyType.cs
```diff
@@ -46,10 +47,23 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
 
+    /// <summary>
+    /// Gets the protocol conformance descriptor for the given type
+    /// </summary>
+    /// <typeparam name="TProtocol"></typeparam>
+    /// <returns></returns>
+    static ProtocolConformanceDescriptor ISwiftObject.GetProtocolConformanceDescriptor<TProtocol>()
+        where TProtocol : class
+    {
+        // TODO: Implement
```
kotlarmilos:
```suggestion
        // TODO: https://github.com/dotnet/runtimelab/issues/2957
```

Please update the tracking issue if needed.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
vitek-karas: Thanks - I guess the NativeAOT compiler is clever enough to figure this out (as it can see all instantiations) - I didn't expect it, but good :-)

I think a discussion would be good for two reasons:
- Determine the best possible workaround without going to IL/Roslyn. Specifically performance concerns.
- Provide a good reference for asks on Roslyn to finally do this and/or discuss an IL rewrite option if the workaround is not good enough.

src/Swift.Runtime/src/Swift/Runtime/ISwiftObject.cs
```diff
@@ -23,6 +23,11 @@ public interface ISwiftObject
     /// Marshals this object to a Swift destination
     /// </summary>
     unsafe IntPtr MarshalToSwift(IntPtr swiftDest);
+
+    /// <summary>
+    /// Gets the protocol conformance descriptor for the given type
+    /// </summary>
+    public static abstract ProtocolConformanceDescriptor GetProtocolConformanceDescriptor<TProtocol>() where TProtocol : class;
```
kotlarmilos, line 30: Is `class` constraint required? What happens if a type is projected as a struct?

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT License.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+
+namespace Swift.Runtime;
+
+public readonly struct ProtocolConformanceDescriptor : IEquatable<ProtocolConformanceDescriptor>
+{
+    private readonly IntPtr _handle;
```
kotlarmilos, line 14: Do we want to use `SwiftHandle` instead?

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
```diff
@@ -0,0 +1,59 @@
...
+            var helperType = typeof(ProtocolConformanceDescriptorHelper<,>).MakeGenericType(typeof(TType), typeof(TProtocol));
+            result = (ProtocolConformanceDescriptor)helperType.GetMethod("GetProtocolConformanceDescriptor")!.Invoke(null, null)!;
+            return true;
+        }
+
+        result = null;
+        return false;
+    }
+
+    public static ProtocolConformanceDescriptor LoadFromSymbol(string libraryPath, string symbolName)
```
kotlarmilos: Please add a tracking issue to replace this with MachO reader in future.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
vitek-karas: We can't rely on the path being the same between compile time and runtime. Even system libraries can be in different locations potentially (in the future) and non-system ones are in effectively random locations - for example if it ships with the app.
I think this should rely on the same resolution mechanism as PInvoke (unless I'm missing something). I understand we can't use PInvoke directly, but we should be able to use NativeLibrary.Load(name, assembly, ...) instead.

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            throw new SwiftRuntimeException($"Unable to load library: {libraryPath}");
+        }
+
+        if (NativeLibrary.TryGetExport(libraryHandle, symbolName, out var handle))
+        {
+            return new ProtocolConformanceDescriptor(handle);
+        }
+
+        throw new SwiftRuntimeException($"Unable to find symbol: {symbolName} in library: {libraryPath}");
```
kotlarmilos: If the handle is not `IntPtr.Zero`, deallocate the memory using `NativeLibrary.Free`.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
```diff
@@ -0,0 +1,52 @@
...
+        return witnessTable;
+    }
+
+    /// <summary>
+    /// Gets the protocol witness table for a given type and protocol.
+    /// </summary>
+    /// <param name="conformanceDescriptor">The protocol conformance descriptor.</param>
+    /// <param name="typeMetadata">The type metadata.</param>
+    /// <param name="instantiationArgs">The instantiation arguments used for conditional conformance.</param>
+    [DllImport(KnownLibraries.SwiftCore, CallingConvention = CallingConvention.Cdecl)]
```
kotlarmilos, line 145: Do we want to use `CallingConvention.CallConvSwift`?

src/Swift.Runtime/src/Swift/SwiftOptional.cs
```diff
@@ -71,6 +71,18 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
 
+    /// <summary>
+    /// Gets the protocol conformance descriptor for the given type
+    /// </summary>
+    /// <typeparam name="TProtocol"></typeparam>
+    /// <returns></returns>
+    static ProtocolConformanceDescriptor ISwiftObject.GetProtocolConformanceDescriptor<TProtocol>()
+        where TProtocol : class
+    {
+        // TODO: Implement
```
kotlarmilos:
```suggestion
        // TODO: https://github.com/dotnet/runtimelab/issues/2957
```

Please update the tracking issue if needed.

kotlarmilos (approved):
LGTM!

src/Swift.Runtime/src/Swift/AnyType.cs
kotlarmilos: What is the standard across .NET repositories? Can we follow https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md?

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
kotlarmilos: @jkurdek Do we plan to replace it with the MacO reader? If there is no accessor function, can we read the content of the descriptor?

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
MichalStrehovsky: Yep, this is fine and it's contractual. We document it here: https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/intrinsic-requiresdynamiccode-apis#typemakegenerictypetype-method-net-9

This was added because it's the only way to bridge constraints currently and people use these patterns to solve the same problems that this is solving.

It's not possible to express this in IL right now either (the runtime validates the constraints meet in the uninstantiated forms - if they happen to be met for a particular substitution, it's not enough). There's a proposal of an IL extension that would allow this at https://github.com/dotnet/runtime/pull/97084 but without a commitment from C# team to use this, it's not moving forward.

If perf is of concern, consider e.g. caching a delegate (use MethodInfo.CreateDelegate to get a delegate and cache it using the generic parameters as a key). Of course then one needs to think about unloadability, etc.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
jkurdek: @vitek-karas you are right, I meant library name. I will follow the pattern you outlined. 

@kotlarmilos Correct me if I'm wrong. My understanding is that during compile time we will use `MachO` to obtain the symbol for the protocol conformance descriptor. However, obtaining the protocol conformance descriptor needs to happen during runtime (as we need to use the platform version of the library). And in order to read that symbol we need to load native library right?

src/Swift.Runtime/src/Swift/Runtime/ISwiftObject.cs
jkurdek: TProtocol should always be an interface: https://github.com/dotnet/runtimelab/pull/2958#discussion_r1929879061

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
jkurdek: `ProtocolConformanceDescriptor` is "top-level" handle just as `SwiftHandle`. They both wrap around the IntPtr. Same is true for `ProtocolWitnessTable`. The idea is that now we can write pinvokes signatures using `SwiftHandle`, `ProtocolConformanceDescriptor` and `ProtocolWitnessTable` as opposed to raw `IntPtr`

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
jkurdek: As noted in a previous comment I think that `MachO` will only give us the symbol and we need to load it from the library ourselves.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
jkurdek: I think this is a regular c-like call. So `Cdecl` is more appropriate. Anyway this call just accepts a bunch of pointers.

src/Swift.Runtime/src/Swift/AnyType.cs
jkurdek: I don't think this document describes XML comments. I think we can go and use `inheritdoc` if you are alright with that.

Opened an issue on coding style: https://github.com/dotnet/runtimelab/issues/2961

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
kotlarmilos: > @kotlarmilos Correct me if I'm wrong. My understanding is that during compile time we will use MachO to obtain the symbol for the protocol conformance descriptor. However, obtaining the protocol conformance descriptor needs to happen during runtime (as we need to use the platform version of the library). And in order to read that symbol we need to load native library right?

Yes, that is right.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
kotlarmilos: The `cdecl` callconv will work here, but for consistency I suggest using Swift callconv. The libSwiftCore contains Swift standard library functions and structures.

jkurdek: We only use `cdecl` in `Swift.Runtime`. I'm fine with changing all of them to `CallConvSwift`. Also should this be a `LibraryImport`?

jkotas: This function is exported from libSwiftCore with standard C calling convention: https://github.com/swiftlang/swift/blob/c1915c8cdd7052291f4761b37348f81511113a76/include/swift/Runtime/RuntimeFunctions.def#L1084 (`C_CC` means C calling convention).

So it is more appropriate to use Cdecl here.

kotlarmilos: @jkurdek Please add a comment so this doesn't get overridden in the future.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
vitek-karas: I don't understand how the compiler knows - the callsite to MakreGenericType here can guarnatee that the second type parameter is a ref type, but the first type is unconstrained. It's possible that all current uses are ref types, but I don't think we can guarantee that constrain.

Let's also think about how to solve this some other way avoiding static virtual methods.

vitek-karas: Even though this works and is effectively supported as the official solution - for our use case this is going to be a perf problem. I'm thinking about cases like calling a swift method which takes a protocol parameter. We will run through this for every call to marshal the parameter. That feels rather expensive.

For the instance based cases we can do way better by simply having an instance method on ISwiftObject. I'll have to think about the cases where it's not instance based...

MichalStrehovsky: > I don't understand how the compiler knows - the callsite to MakreGenericType here can guarnatee that the second type parameter is a ref type, but the first type is unconstrained.

The first one is constrained to be `typeof(T)`, the second is `typeof(U)`. We just model it as if there was a call from the generic method to the method on the instantiated generic type. (The PR that added this was https://github.com/dotnet/runtime/pull/99037 if you're interested.)

> Even though this works and is effectively supported as the official solution - for our use case this is going to be a perf problem.

You could add a hashtable in front of it. This should have performance characteristics close to doing a generic virtual method call (that one is also a hashtable lookup). Maybe a bit slower but not by much. People call generic virtual methods in perf critical code all the time (often because they're unaware GVMs are the slowest form of calls in .NET, much much slower than delegates, but I digress).

```csharp
if (!s_dict.TryGetValue((typeof(T), typeof(U)), out var del))
{
    {
        var helperType = typeof(ProtocolConformanceDescriptorHelper<,>).MakeGenericType(typeof(T), typeof(U));
        del = helperType.GetMethod("GetProtocolConformanceDescriptor")!.CreateDelegate<Func<ProtocolConformanceDescriptor>>();
    }
    else
    {
        del = null;
    }
    s_dict.Add((typeof(T), typeof(U)), del);
}

if (del != null)
{
    result = del();
    return true;
}

result = null;
return false;
```

MichalStrehovsky: Actually, this can be optimized into a delegate invocation. If something like https://github.com/dotnet/runtime/pull/109679 ever lands in the JIT/VM, even the delegate invocation would go away and this would be a zero-cost abstraction. We could make this into a zero cost abstraction on the native AOT side too if necessary, but even as-is, it's just a static constructor check followed by delegate invocation.

```csharp
class DelegateCache<T, U>
{
    public static Func<ProtocolConformanceDescriptor> Func { get; } =
        typeof(ISwiftObject).IsAssignableFrom(typeof(T))
        ? typeof(ProtocolConformanceDescriptorHelper<,>).MakeGenericType(typeof(T), typeof(U)).GetMethod("GetProtocolConformanceDescriptor")!.CreateDelegate<Func<ProtocolConformanceDescriptor>>()
        : null;
}

static bool TryGetProtocolConformanceDescriptor<T, U>(out ProtocolConformanceDescriptor result)
{
    var d = DelegateCache<T, U>.Func;
    if (d == null)
    {
        result = null;
        return false;
    }

    result = d();
    return true;
}
```

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    internal static bool TryGetProtocolConformanceDescriptor<TType, TProtocol>([NotNullWhen(true)] out ProtocolConformanceDescriptor? result)
+        where TProtocol : class
+    {
+        var type = typeof(TType);
+
+        if (typeof(ISwiftObject).IsAssignableFrom(type))
+        {
+            var helperType = typeof(ProtocolConformanceDescriptorHelper<,>).MakeGenericType(typeof(TType), typeof(TProtocol));
+            result = (ProtocolConformanceDescriptor)helperType.GetMethod("GetProtocolConformanceDescriptor")!.Invoke(null, null)!;
```
matouskozak: Are we sure that `GetMethod("GetProtocolConformanceDescriptor")!` always returns non-null method?

matouskozak (approved):
LGTM! Good job Jeremi.

It would be great to add some tests around PCD and PWT as these look like key APIs for future projections.

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
jkurdek: We would have to add the same comment to every pinvoke in `Arc` as well. Seems a bit superfluous. I don't think someone should go and change calling conventions on working pinvokes :sweat_smile:

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
jkurdek: Created an issue to track this discussion and moved some of the comments there:
https://github.com/dotnet/runtimelab/issues/2962

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
jkurdek: I cannot imagine a scenario this fails. We define the `GetProtocolConformanceDescriptor` function ourselves in another file

src/Swift.Runtime/src/Swift/Runtime/ProtocolWitnessTable.cs
kotlarmilos: I think this is valuable information for future changes and for any new members joining the project. Feel free to include it in the docs wherever you think it fits.

src/Swift.Runtime/src/Swift/SwiftOptional.cs
jkurdek: https://github.com/dotnet/runtimelab/issues/2963

Issue #2833, opened 2024-12-03
[Swift language features] Implement Swift.Array in C#


kotlarmilos:
## Tasks
 - [ ] Port manual bindings from https://github.com/xamarin/binding-tools-for-swift/blob/main/SwiftRuntimeLibrary/SwiftArray.cs
 - [ ] Add functional tests

Issue #2930, opened 2025-01-14
[Swift language features] Implement Swift generics support for structs


kotlarmilos:
## Tasks

 - [x] Generic methods
 - [ ] Generic types
 - [ ] Protocol constraints
 - [ ] PAT constraints

kotlarmilos:
Moved to https://github.com/dotnet/runtimelab/issues/2824.

Issue #2851, opened 2024-12-03
[Swift ABI compatibility] Implement correct memory handling in C# bindings


kotlarmilos:
## Tasks
 - [x] https://github.com/dotnet/runtimelab/pull/3008
 - [x] https://github.com/dotnet/runtimelab/pull/3009
 - [x] #3020
 - [x] #3035
 - [x] #3036
 - [x] #2978
 - [x] https://github.com/dotnet/runtimelab/issues/2975
 - [x] https://github.com/dotnet/runtimelab/issues/2974
 - [x] https://github.com/dotnet/runtimelab/issues/3018

PR #2964, opened 2025-01-28
[Swift language feature] Implement Swift.Array support

kotlarmilos:
## Description

This PR is a follow-up to https://github.com/dotnet/runtimelab/pull/2948. It introduces support for `Swift.Array`.

In Swift, Array is defined as `@frozen struct Array<Element>`. It is projected into a SwiftArray class with a helper struct `ArrayBuffer`, which is used at the ABI level to ensure correct lowering and PInvoke signatures.

Fixes https://github.com/dotnet/runtimelab/issues/2833


src/Swift.Runtime/tests/LibraryTests/SwiftArrayTests.cs
```diff
@@ -0,0 +1,190 @@
...
+    [Fact]
+    public unsafe void ArrayDispose()
+    {
+        var array = new SwiftArray<int>();
+        var payload = *(IntPtr*)array.buffer.storage;
+        var count = Arc.RetainCount(payload);
+
+        array.Dispose();
+
+        Assert.Equal(count - 1, Arc.RetainCount(payload));
```
jkotas, line 136: What guarantees that `payload` is still alive at this point?

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,294 @@
...
+///    |   struct ArrayBuffer                                                 |
+///    |   +--------------------------------------------------------------+   |
+///    |   |   struct BridgeStorage                                       |   |
+///    |   |   +------------------------------------------------------+   |   |
+///    |   |   | var rawValue: IntPtr                                 |   |   |
+///    |   |   +------------------------------------------------------+   |   |
+///    |   +--------------------------------------------------------------+   |
+///    +-----------------------------------------------------------------------+
+/// </summary>
+[StructLayout(LayoutKind.Sequential, Size = 8)]
```
jkotas: What's the purpose of `Size = 8`?

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,294 @@
...
+public struct ArrayBuffer
+{
+    public IntPtr storage;
+}
+
+/// <summary>
+/// Represents a Swift array.
+/// </summary>
+/// <typeparam name="Element">The element type contained in the array.</typeparam>
+public unsafe class SwiftArray<Element> : IDisposable, ISwiftObject
```
jkotas: Is see that you have `unsafe` in the individual methods. Does the whole type really need to be marked as unsafe too?

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,294 @@
...
+{
+    static nuint _payloadSize = SwiftObjectHelper<SwiftArray<Element>>.GetTypeMetadata().Size;
+    static nuint _elementSize = ElementTypeMetadata.Size;
+
+    // Swift array is a value type and doesn't contain an IntPtr payload
+    public ArrayBuffer buffer;
```
jkotas: Should this field be private?

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,294 @@
...
+    bool _disposed = false;
+    public void Dispose()
+    {
+        if (!_disposed)
+        {
+            Arc.Release(*(IntPtr*)buffer.storage);
```
jkotas: Should we set buffer.storage to null after it was released? We may be able to use `null` to mark the object as disposed instead of having separate `_disposed` field.

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,294 @@
...
+/// <summary>
+/// Represents a Swift array.
+/// </summary>
+/// <typeparam name="Element">The element type contained in the array.</typeparam>
+public unsafe class SwiftArray<Element> : IDisposable, ISwiftObject
+{
+    static nuint _payloadSize = SwiftObjectHelper<SwiftArray<Element>>.GetTypeMetadata().Size;
+    static nuint _elementSize = ElementTypeMetadata.Size;
+
+    // Swift array is a value type and doesn't contain an IntPtr payload
```
vitek-karas: This comment is weird - here it says it doesn't contain IntPtr payload, but then ArrayBuffer contains IntPtr payload... so which one is it?

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+    
+    ~SwiftArray()
+    {
+        Arc.Release(*(IntPtr*)buffer.storage);
+    }
+    public static nuint PayloadSize => _payloadSize;
+    public static nuint ElementSize => _elementSize;
+    static TypeMetadata ISwiftObject.GetTypeMetadata()
```
vitek-karas: Nit - formatting - empty lines in between

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /// <summary>
+    /// Appends the given element to the array.
+    /// </summary>
+    public void Append(Element item)
+    {
+        unsafe
+        {
+            IntPtr T0Payload = IntPtr.Zero;
```
vitek-karas: This is a weird name for a local variable... does it come from some Swift terminology?

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,294 @@
...
+            var metadata = SwiftObjectHelper<SwiftArray<Element>>.GetTypeMetadata();
+            try
+            {
+                T0Payload = (IntPtr)NativeMemory.Alloc(ElementSize);
+                SwiftMarshal.MarshalToSwift(item, (IntPtr)T0Payload);
```
vitek-karas: The cast to IntPtr seems unnecessary

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,285 @@
...
+public struct ArrayBuffer
+{
+    public IntPtr storage;
+}
+
+/// <summary>
+/// Represents a Swift array.
+/// </summary>
+/// <typeparam name="Element">The element type contained in the array.</typeparam>
+public class SwiftArray<Element> : IDisposable, ISwiftObject
```
jkurdek, line 46: This is making the SwiftArray follow reference semantics on c# side. On Swift side SwiftArray is a value type, for perf purposes it utilises copy on write mechanism. I think we need to be intentional about making this either a reference or value type, as this will set an example for projecting other collections from swift.

Personally I see value in making the projections thin and keeping them as close to Swift as possible. First it will make the onboarding for our users simpler - they can just follow Swift documentation and expect the types to work as described there. Second, this will make the translation of use case scenario from Swift to C# much simpler.

src/Swift.Runtime/tests/LibraryTests/SwiftArrayTests.cs
kotlarmilos: Good point. The test is updated to retain the payload before the dispose. Please let me know if there are better ways to test this.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: This should be a hint for the runtime lowering. Currently, the projections contain the attribute, but it shouldn't be necessary in the most of the cases.

/cc: @matouskozak

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: Yes, this is manual binding and we can do that.

My idea was to keep this aligned with the generated code.

src/Swift.Runtime/src/Swift/SwiftArray.cs
jkurdek: Something to keep in mind - array is interesting, it is a frozen generic type, but it seems that even when the generic parameter is frozen it doesn't go through lowering when passed to functions. That means that every function which accepts `array<SomethingConcrete>` will expect an opaque pointer. 

Edit: this comment is wrong

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,285 @@
...
+{
+    static nuint _payloadSize = SwiftObjectHelper<SwiftArray<Element>>.GetTypeMetadata().Size;
+    static nuint _elementSize = ElementTypeMetadata.Size;
+
+    // Swift array is a value type and doesn't contain an IntPtr payload
+    private ArrayBuffer buffer;
```
jkurdek: NIT: missing newline

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,285 @@
...
+            buffer.storage = IntPtr.Zero;
+            GC.SuppressFinalize(this);
+        }
+    }
+
+    unsafe ~SwiftArray()
+    {
+        Arc.Release(*(IntPtr*)buffer.storage);
+        buffer.storage = IntPtr.Zero;
+    }
```
jkurdek, line 78: NIT: missing newline

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,285 @@
...
+    {
+        return ProtocolConformanceDescriptor.Zero;
+    }
+
+    /// <summary>
+    /// Constructs a new SwiftArray from the given handle.
+    /// </summary>
+    unsafe SwiftArray(SwiftHandle handle)
+    {
+        this.buffer = *(ArrayBuffer*)(handle);
```
jkurdek: Should this just copy the memory? What about arrays of reference types?

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Inserts the given element at the given index.
+    /// </summary>
+    public unsafe void Insert(int index, Element item)
+    {
+        if (index < 0 || index >= Count)
+            throw new ArgumentOutOfRangeException(nameof(index));
```
vitek-karas: Just curious about our general approach: Technically there's nothing in this projection which requires the index to be in the bounds of the array. And for other projections we won't be able to generate argument validation this way. Should we leave this up to the Swift's implementation fully? (For this case, I could imagine that Insert after end means - expand the array and set the element for example).

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,285 @@
...
+[StructLayout(LayoutKind.Sequential, Size = 8)]
+public struct ArrayBuffer
+{
+    public IntPtr storage;
+}
+
+/// <summary>
+/// Represents a Swift array.
+/// </summary>
+/// <typeparam name="Element">The element type contained in the array.</typeparam>
```
vitek-karas, line 45: Do we already know where we're going to use this projection? (in real code somewhere)
There are two ways we can approach this:
- "the expected behavior" - It's an array, so it should be array on the other side as well
- "the precise behavior" - it's a swift struct with special behavior, so we project it as a special thing

We could also use both - have the precise projection as the base and have a layer on top which makes it behave more like an array in C#.

For example - if I have a Swift method which returns an array of numbers, when projected I would expect to be able to `foreach` on the return value to enumerate the numbers. This projection will not make that work. That's not necessarily wrong, but I think we should be pretty clear what is the goal of this projection.

Side note: Swift arrays have a value semantics, that makes it pretty "Simple" to marshal them by "copy" - so in theory we could create a new C# array and marshal all of the elements one by one. Again, maybe we'll do that as a layer on top of this projection.

This is somewhat related to the discussion on marshalling strings. We might create a projection for Swift string as well, or we might chose to just always marshal it as some C# native type (like System.String).

src/Swift.Runtime/src/Swift/SwiftArray.cs
jkotas:
```suggestion
    private ArrayBuffer _buffer;
```
Follow the coding style?

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: I think it is lowered, but contains a single pointer as a payload. Here is an example:
```llvm
// Function that accepts array (frozen struct)
define protected swiftcc void @"output.dummyCallArray(x: [output.FrozenStruct]) -> ()"(ptr %0) #0 !dbg !169 {
entry:
  %x.debug = alloca ptr, align 8
  call void @llvm.dbg.declare(metadata ptr %x.debug, metadata !174, metadata !DIExpression()), !dbg !176
  call void @llvm.memset.p0.i64(ptr align 8 %x.debug, i8 0, i64 8, i1 false)
  store ptr %0, ptr %x.debug, align 8, !dbg !177
  ret void, !dbg !178
}

// Function that accepts a frozen struct (with the same layout as an array)
define protected swiftcc void @"output.dummyCallFrozen(x: output.FrozenStruct) -> ()"(ptr %0) #0 !dbg !162 {
entry:
  %x.debug = alloca ptr, align 8
  call void @llvm.dbg.declare(metadata ptr %x.debug, metadata !164, metadata !DIExpression()), !dbg !166
  call void @llvm.memset.p0.i64(ptr align 8 %x.debug, i8 0, i64 8, i1 false)
  store ptr %0, ptr %x.debug, align 8, !dbg !167
  ret void, !dbg !168
}

// Function that accepts array (frozen struct)
define protected swiftcc void @"output.dummyCallNonFrozen(x: output.NonFrozenStruct) -> ()"(ptr noalias nocapture dereferenceable(8) %0) #0 !dbg !184 {
entry:
  %x.debug = alloca ptr, align 8
  call void @llvm.dbg.declare(metadata ptr %x.debug, metadata !186, metadata !DIExpression()), !dbg !187
  call void @llvm.memset.p0.i64(ptr align 8 %x.debug, i8 0, i64 8, i1 false)
  %.x = getelementptr inbounds %T6output15NonFrozenStructV, ptr %0, i32 0, i32 0, !dbg !188
  %.x._rawValue = getelementptr inbounds %TSV, ptr %.x, i32 0, i32 0, !dbg !188
  %1 = load ptr, ptr %.x._rawValue, align 8, !dbg !188
  store ptr %1, ptr %x.debug, align 8, !dbg !190
  ret void, !dbg !191
}
```

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: Ideally, this should be projected as a C# struct to maintain "copy-on-write" semantics. However, its implementation consists of a single payload pointer, and when a struct is copied, the pointer value remains unchanged. The main reason for projecting it as a class is the need for a destructor.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: I would remove this constructor as it goes against the expected behavior on Swift side. Even if payload value is copied, it would point to the same instance.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: We need to confirm whether it is lowered or not: https://github.com/dotnet/runtimelab/pull/2964#discussion_r1934048643

It might be passed by reference; otherwise lowered with a a single-pointer payload.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: Changed. No, it comes from the projection tooling.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: > We could also use both - have the precise projection as the base and have a layer on top which makes it behave more like an array in C#.

This sounds like a reasonable approach. This is a precise projection as the base and doesn't implement C# array.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: Removed, not necessary.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: Can't be removed, it is used in the `ISwiftObject.NewFromPayload`. It should perform a shallow copy of the `ArrayBuffer` struct.

Anyway, it doesn't have the same semantics as in Swift.

src/Swift.Runtime/src/Swift/SwiftArray.cs
matouskozak: > This should be a hint for the runtime lowering. Currently, the projections contain the attribute, but it shouldn't be necessary in the most of the cases.

Yes, I believe that the `Size` is only important for nested structs so that we avoid any unwanted padding inside the struct. That said, I think we can keep it around until we figure out something better (e.g., https://github.com/dotnet/runtime/issues/100896).

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,286 @@
...
+
+    /// <summary>
+    /// Gets the protocol conformance descriptor for the given type.
+    /// </summary>
+    /// <typeparam name="TProtocol"></typeparam>
+    /// <returns></returns>
+    static ProtocolConformanceDescriptor ISwiftObject.GetProtocolConformanceDescriptor<TProtocol>()
+        where TProtocol : class
+    {
+        return ProtocolConformanceDescriptor.Zero;
```
jkurdek: This will have to be able to obtain the descriptor for `array : Collection`

src/Swift.Runtime/src/Swift/SwiftArray.cs
jkurdek: @kotlarmilos you are right. It will be always lowered to the single pointer. No matter what `T` ends up being. Still I'm not sure we need the explicit size, can this have any extra padding?

src/Swift.Runtime/src/Swift/SwiftArray.cs
jkurdek: Ok, maybe this is a future problem once we start supporting reference types. We might need to call `InitializeWithCopy` here

src/Swift.Runtime/src/Swift/SwiftArray.cs
jkurdek: it seems to be lowered to a single pointer. It doesn't use `IndirectResult` thats for sure.

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,286 @@
...
+
+    static nuint _elementSize = ElementTypeMetadata.Size;
+
+    private ArrayBuffer _buffer;
+
+    public unsafe void Dispose()
+    {
+        if (_buffer.storage != IntPtr.Zero)
+        {
+            Arc.Release(*(IntPtr*)_buffer.storage);
```
jkotas, line 68: Did you have any discussions about thread-safety and security of the Swift projections?

For example, with this implementation - user code calling Dispose method on multiple threads can lead to releasing the `_buffer.storage` twice (double-free security bug). .NET core libraries typically try to avoid turning race conditions in user code into memory safety violations. Are we ok with Swift projections not providing this hardening?

src/Swift.Runtime/tests/LibraryTests/SwiftArrayTests.cs
```diff
@@ -0,0 +1,183 @@
...
+        private static void InitializeResources()
+        {
+        }
+    }
+
+    [Fact]
+    static void SmokeTest()
+    {
+        var metadata = TypeMetadata.GetTypeMetadataOrThrow<SwiftArray<int>>();
+        Assert.Equal((nuint)8, metadata.Size);
```
matouskozak, line 37: I don't understand how is `metadata.Size` correlated with `SwifArray` of `int`. What I mean is, how is the size of `metadata` affected by `SwiftArray` and its specializations?

matouskozak (commented):
Good job Milos!

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: > maybe this is a future problem once we start supporting reference types

Could you provide an example?

> We might need to call InitializeWithCopy here

Added to https://github.com/dotnet/runtimelab/issues/2930.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: Not needed for this scenario.

src/Swift.Runtime/tests/LibraryTests/SwiftArrayTests.cs
kotlarmilos: This should be the payload size (`IntPtr`). I don't think it is affected by the `Element`.

matouskozak: Thank you. For clarity that this is not related to the actual array, I would put a comment here or used `sizeof(IntPtr)` instead of 8.

jkurdek: Isn't it the size of ArrayBuffer? We retrieve the value witness table and we query it for the size

src/Swift.Runtime/src/Swift/SwiftArray.cs
jkurdek: I don't have anything specific in my mind yet. I was think about having an array of reference types. If we just make a shallow copy of the array the ref count of those references will not be updated.

src/Swift.Runtime/src/Swift/SwiftArray.cs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
+    {
+        var metadata = SwiftObjectHelper<SwiftArray<Element>>.GetTypeMetadata();
+        unsafe
+        {
+            fixed (void* _payloadPtr = &_buffer)
+            {
+                metadata.ValueWitnessTable->InitializeWithCopy((void*)swiftDest, (void*)_payloadPtr, metadata);
```
vitek-karas, line 108: General comment about current design of MarshalToSwift - this feels very prone to buffer overruns.
It's a virtual call, so the caller technically doesn't know what the callee will be precisely. But at the same time, the caller MUST now the size of the buffer to allocate for swiftDest. That feels wrong.

What is the scenario where we have a preexisting buffer and need to marshal to it? So far the callers I've seen all need to allocate before calling this method. And they have to get the size right, otherwise we'll get buffer overruns at runtime.

I think the minimum should be that we don't pass a raw pointer, but some representation of pointer + size. Or better yet, this method would be responsible for allocating the buffer and returning it.

I also find it weird that the method takes a destination buffer, but it may choose to return another buffer. What is the memory ownership? We should at least precisely document this in comments, but ideally redesign the API to make it "secure by default".

vitek-karas: Maybe we could use `MemoryManager<byte>` (our own implementation) which would handle the memory ownership. But it's also possible that having a simple struct of our own would be good enough.

jkurdek: I agree that we should review the `MarshalToSwift` design. I would not tackle this in this PR though.

jkurdek (approved):

src/Swift.Runtime/tests/LibraryTests/SwiftArrayTests.cs
kotlarmilos: Yes, it is. The payload in the `SwiftArray` class is an `ArrayBuffer`, which ends up containing the size of its payload, which is `IntPtr`.

`SwiftArray` -> `ArrayBuffer` -> `IntPtr`

src/Swift.Runtime/src/Swift/SwiftArray.cs
vitek-karas: Absolutely - created https://github.com/dotnet/runtimelab/issues/2974 to track that.

src/Swift.Runtime/src/Swift/SwiftArray.cs
vitek-karas: Great point - created https://github.com/dotnet/runtimelab/issues/2975 to track this to revisit all projections and fix it there (if we have any other really). In any case which is something we might want to have a helper for.

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: Added to https://github.com/dotnet/runtimelab/issues/2851

src/Swift.Runtime/src/Swift/SwiftArray.cs
kotlarmilos: Thanks @vitek-karas for creating the tracking issues. We need to revisit this.

On the same note, we've already encountered a multi-threading scenario — in-app purchase request can't be performed on the main thread obviously.

Issue #2834, opened 2024-12-03
[Swift language features] Implement Swift.Set in C#


kotlarmilos:
## Tasks

 - [ ] Port manual bindings from https://github.com/xamarin/binding-tools-for-swift/blob/main/SwiftRuntimeLibrary/SwiftSet.cs
 - [ ] Add functional tests

PR #2968, opened 2025-01-29
[Swift language feature] Implement empty Swift.Set in C#

kotlarmilos:
## Description

This PR introduces support for an empty `Swift.Set` used in https://developer.apple.com/documentation/storekit/product/purchase(options:). In Swift, Set is defined as `@frozen struct Array<Element> where Element : Hashable`. The type is projected similar to [`Swift.Array`](https://github.com/dotnet/runtimelab/issues/2833).

Fixes https://github.com/dotnet/runtimelab/issues/2834


mcumming (approved):

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
```diff
@@ -141,4 +155,10 @@ public static ProtocolConformanceDescriptor LoadFromSymbol(string libraryName, s
             NativeLibrary.Free(libraryHandle);
         }
     }
+
+    private static ProtocolConformanceDescriptor GetProtocolConformanceDescriptor(TypeMetadata metadata, ProtocolDescriptor descriptor)
+        => swift_conformsToProtocol(metadata, descriptor);
+
+    [DllImport(KnownLibraries.SwiftCore, CallingConvention = CallingConvention.Cdecl)]
+    private static extern ProtocolConformanceDescriptor swift_conformsToProtocol(TypeMetadata metadata, ProtocolDescriptor descriptor);
```
jkurdek: Is this the function?https://github.com/swiftlang/swift/blob/136cdee1950a5e72b99d7b8177bb2ab5043cd974/include/swift/Runtime/RuntimeFunctions.def#L1863

I think it returns a `ProtocolWitnessTable` not a protocol conformance descriptor. Which makes me wonder how it works. But on the other hand we don't have any tests yet which verify that.

src/Swift.Runtime/src/Swift/Runtime/ISwiftObject.cs
```diff
@@ -27,10 +27,15 @@ public interface ISwiftObject
     /// <summary>
     /// Gets the protocol conformance descriptor for the given type
     /// </summary>
-    public static abstract ProtocolConformanceDescriptor GetProtocolConformanceDescriptor<TProtocol>() where TProtocol : class;
+    public static abstract ProtocolConformanceDescriptor GetProtocolConformanceDescriptor<TProtocol>() where TProtocol : ISwiftProtocol;
 }
 
-/// <summary>  
+public interface ISwiftProtocol
```
jkurdek: I am not sure about this:
```csharp
public interface IProtocol
{
    public static abstract string GetSomething();
}

public interface IPrintable : IProtocol
{
    static string IProtocol.GetSomething() => "IPrintable";
}

public interface IWritable : IProtocol
{
    static string IProtocol.GetSomething() => "IWritable";
}

public class SomeClass : IPrintable, IWritable
{
// I dont think this will compile
}
```

Also this would make any class implementing a protocol be a `ISwiftProtocol` as in `val is ISwiftProtocol` returns true

src/Swift.Runtime/src/Swift/Protocols.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License.
+
+using System;
+using Swift;
+using Swift.Runtime;
+
+/// <summary>
+/// Represents Swift Hashable protocol in C#.
+/// </summary>
+public unsafe interface IHashable : ISwiftProtocol
```
jkurdek: I would prefer `ISwiftHashable` I think that we should be explicit about this not being a `c#` interface

src/Swift.Runtime/src/Swift/Set.cs
```diff
@@ -0,0 +1,151 @@
...
+
+    unsafe ~SwiftSet()
+    {
+        Arc.Release(*(IntPtr*)variant.rawValue);
+        variant.rawValue = IntPtr.Zero;
+    }
+    public static nuint PayloadSize => _payloadSize;
+    public Variant Payload => variant;
+    public static nuint ElementSize => _elementSize;
+    public static ProtocolWitnessTable IHashableWitnessTable
```
jkurdek: I would use `ProtocolWitnessTable.GetOrThrow` here rather than assigning an invalid value

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
kotlarmilos: Yes, I missed that. Let's move it to the PWT and add tests.

src/Swift.Runtime/src/Swift/Set.cs
jkurdek: I think that if the only thing we need is an empty set we might not actually need to use `IHashableWitnessTable` at all. We should be good with just passing a nulltpr. We will need to have a valid protocol witness table when we use functions which actually try to hash something.

kotlarmilos:
Reverted protocol witness table changes in favour of https://github.com/dotnet/runtimelab/pull/2969.

src/Swift.Runtime/src/Swift/Set.cs
kotlarmilos: Yes, sounds good.

src/Swift.Runtime/src/Swift/Protocols.cs
kotlarmilos: I agree. Reverted from this PR.

src/Swift.Runtime/src/Swift/Runtime/ISwiftObject.cs
kotlarmilos: I agree. Reverted from this PR.

src/Swift.Runtime/src/Swift/Runtime/ProtocolConformanceDescriptor.cs
kotlarmilos: These changes will be introduced in a separate PR.

dalexsoto (approved):

src/Swift.Runtime/src/Swift/Set.cs
```diff
@@ -0,0 +1,150 @@
...
+            _variant.rawValue = IntPtr.Zero;
+            GC.SuppressFinalize(this);
+        }
+    }
+
+    unsafe ~SwiftSet()
+    {
+        Arc.Release(*(IntPtr*)_variant.rawValue);
+        _variant.rawValue = IntPtr.Zero;
+    }
```
vitek-karas: Is there a reason this doesn't use the canonical Dispose pattern (having a Dispose(bool disposing) method which is called from both dispose and finalizer?

src/Swift.Runtime/src/Swift/Set.cs
```diff
@@ -0,0 +1,150 @@
...
+
+    public static ProtocolWitnessTable IHashableWitnessTable => ProtocolWitnessTable.Zero;
```
jkurdek: I think this can be removed now

src/Swift.Runtime/src/Swift/Set.cs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /// <summary>
+    /// Gets the protocol conformance descriptor for the given type.
+    /// </summary>
+    /// <typeparam name="TProtocol"></typeparam>
+    /// <returns></returns>
+    static ProtocolConformanceDescriptor ISwiftObject.GetProtocolConformanceDescriptor<TProtocol>()
+    {
+        return ProtocolConformanceDescriptor.Zero;
```
jkurdek: This will have to return a valid ICollection conformance descriptor

src/Swift.Runtime/src/Swift/Set.cs
kotlarmilos: Added a tracking issue: https://github.com/dotnet/runtimelab/issues/2978. We should review the generated bindings as well.

src/Swift.Runtime/tests/LibraryTests/SwiftSetTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    [Fact]
+    static void SmokeTest()
+    {
+        var metadata = TypeMetadata.GetTypeMetadataOrThrow<SwiftSet<SwiftIntMock>>();
+        // sizeof (Variant)
+        Assert.True(metadata.Size == 8);
+
+        var set = new SwiftSet<SwiftIntMock>();
```
vitek-karas, line 39: Good additional test would be to new up an array in C# and pass it to a swift method which validate that it works as an array correctly.

src/Swift.Runtime/src/Swift/SwiftSet.cs
```diff
@@ -0,0 +1,174 @@
...
+            var witnessTable = ProtocolWitnessTable.GetOrThrow<Element, ISwiftHashable>();
+            return (int)SwiftSetPInvokes.Count(_variant, ElementTypeMetadata, witnessTable);
+        }
+    }
+}
+
+internal static class SwiftSetPInvokes
+{
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sShMa")]
+    public static extern TypeMetadata PInvoke_getMetadata(TypeMetadataRequest request, TypeMetadata typeMetadata, ProtocolWitnessTable witnessTable);
```
vitek-karas, line 168: Do these methods have Swift names? The `Init` and `Count` specifically.

src/Swift.Runtime/src/Swift/SwiftSet.cs
```diff
@@ -0,0 +1,174 @@
...
+
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sS2hyxGycfC")]
+    public static extern Variant Init(TypeMetadata typeMetadata, ProtocolWitnessTable witnessTable);
```
vitek-karas: We should probably name the first parameter `elementTypeMetadata` - to make it more descriptive what it needs to be.

src/Swift.Runtime/src/Swift/SwiftSet.cs
kotlarmilos: I am not following -- what do you mean?

src/Swift.Runtime/tests/LibraryTests/SwiftSetTests.cs
kotlarmilos: Good point, added.

src/Swift.Runtime/src/Swift/SwiftSet.cs
vitek-karas: The entry point name looks like mangled mess - is there some humanly readable name which is used in the Swift world for these methods? I doubt it's `Init` as it lacks the "array" context.

src/Swift.Runtime/tests/LibraryTests/SwiftSetTests.cs
vitek-karas: Sorry - I don't see it. I was thinking something like:
```csharp
var set = new SwiftSet<SwiftIntMock>();
var sum = SumAllElementsInSwift(set);  // This would be a swift call which does foreach over the array and sums it up
Assert.Equal(sum, 0);
```

kotlarmilos: Look at https://github.com/dotnet/runtimelab/pull/2968/files#diff-7c9e8d78468b134d5a41fcdc008666026d5335720c62f7959f1dd556a50ad8f8R361. It is in the functional tests under Structs, as it is a struct and there is infrastructure for compiling Swift code.

Extended with you case.

src/Swift.Runtime/src/Swift/SwiftSet.cs
kotlarmilos: Well-known types might be shortened. Here is an output of the demangler:
```
xcrun swift-demangle --expand "sS2hyxGycfC"                                   
Demangling for $sS2hyxGycfC
kind=Global
  kind=Allocator
    kind=Structure
      kind=Module, text="Swift"
      kind=Identifier, text="Set"
    kind=Type
      kind=FunctionType
        kind=ArgumentTuple
          kind=Type
            kind=Tuple
        kind=ReturnType
          kind=Type
            kind=BoundGenericStructure
              kind=Type
                kind=Structure
                  kind=Module, text="Swift"
                  kind=Identifier, text="Set"
              kind=TypeList
                kind=Type
                  kind=DependentGenericParamType
                    kind=Index, index=0
                    kind=Index, index=0
$sS2hyxGycfC ---> Swift.Set.init() -> Swift.Set<A>
```

jkurdek (approved):

jkurdek:
Great work Milos!

Issue #2944, opened 2025-01-17
[Swift language features] Add basic support for PATs constraints on generic methods


jkurdek:
Add ABI level support for calling generic methods where the generic type is constrained to a protocol with associated type

Issue #2918, opened 2025-01-10
[Projection tooling] Add symbols from dynamic library to type database


jkurdek:
## Description

The Swift runtime exposes metadata accessors, witness tables, and conformance descriptors through native symbols. At bindings time, we projection tooling can't invoke them directly because the target platform may differ. However, the tooling can read these symbols and store them in the type database. These symbols will be consumed by emitters and called at runtime.

## Tasks
 - [x] https://github.com/dotnet/runtimelab/issues/2665
 - [x] #2954
 - [x] Integrate TBDParser and Demangler with the rest of the code
 - [x] Read symbols during the post processing step and store them in the type database
   - [x] Metadata accessors
   - [x] Conformance descriptors
 - [x] Replace hardcoded metadata accessors with symbols from the library.
 - [x] Add unit tests

PR #2969, opened 2025-01-30
[SwiftBindings] Adds runtime support for protocol constraints

### LINKED ISSUE #2943: [Swift language features] Add basic support for protocol conformance to generic methods
Add ABI level support for calling generic methods where the generic type is constrained to a protocol.

jkurdek:
Fixes #2943 and #2944. This PR adds runtime support for protocol constraints. Those enable to call Swift generic functions which have protocol and PAT constraints on the generic argument. 

As discussed on weekly meeting we are currently not exposing Swift generic constraints to C# on type system level.

The changes include:
- Adding trivial projections of protocols
- Adding support for conformances in type declarations
- Population of `GetProtocolConformanceDescriptor` method on `ISwiftObject`
- Adding pwt retrieval and passing to pinvokes

In order to make the symbol population work I did create a mock interface for using `MachO`. I did put all of the logic into one file so that it can be easily modified / removed when https://github.com/dotnet/runtimelab/issues/2918 is implemented. 

While working on the PR I noticed several issues coming from previous PRs:
- Handling of generic paramaters in constructor calls. The logic was straight wrong. I reverted the changes removing `ConstructorHandler` so that Swift init methods with generic signatures are not handled.
- Handling of functions with multiple arguments having same generic type e.g. `foo<T>(a: T, b: T)`. Fixed the logic so that the projection is correct

### DIFF
--- a/src/Swift.Bindings/src/DemangledSymbols.cs
+++ b/src/Swift.Bindings/src/DemangledSymbols.cs
@@ -1,0 +1,63 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using BindingsGeneration;
+using Xamarin;
+
+public readonly record struct DemangledSymbols
+{
+    public Dictionary<(NamedTypeSpec, NamedTypeSpec), string> ProtocolConformanceDescriptors { get; init; }
+
+    public DemangledSymbols(Dictionary<(NamedTypeSpec, NamedTypeSpec), string> descriptors)
+    {
+        ProtocolConformanceDescriptors = descriptors;
+    }
+}
+
+public sealed class DemangledSymbolsRegister
+{
+    private static readonly Lazy<DemangledSymbolsRegister> _instance = new(() => new DemangledSymbolsRegister());
+
+    private DemangledSymbols _symbols = new(new Dictionary<(NamedTypeSpec, NamedTypeSpec), string>());
+    private bool _isLoaded = false;
+
+    public static DemangledSymbolsRegister Instance => _instance.Value;
+
+    private DemangledSymbolsRegister() { }
+
+    public DemangledSymbols GetData(string dylibPath = "")
+    {
+        if (!_isLoaded)
+        {
+            if (string.IsNullOrEmpty(dylibPath))
+            {
+                throw new ArgumentException("dylibPath cannot be null or empty.");
+            }
+
+            Load(dylibPath);
+        }
+        return _symbols;
+    }
+
+    private void Load(string dylibPath)
+    {
+        try
+        {
+            var descriptors = DemanglingResults.FromFile(dylibPath, Abi.ARM64).ProtocolConformanceDescriptors;
+            var dictionary = new Dictionary<(NamedTypeSpec, NamedTypeSpec), string>();
+
+            foreach (var descriptor in descriptors)
+            {
+                dictionary[(descriptor.ImplementingType, descriptor.ProtocolType)] = descriptor.Symbol[1..];
+            }
+
+            _symbols = new DemangledSymbols(dictionary);
+            _isLoaded = true;
+        }
+        catch (Exception e)
+        {
+            // This happens if the dylib cannot be located on disk, which will be the case for Apple shipped dylibs. (e.g. Foundation, StoreKit, etc.)
+            Console.Error.WriteLine($"Error loading demangled symbols: {e.Message}");
+        }
+    }
+}
--- a/src/Swift.Bindings/src/DemangledSymbols.cs
+++ b/src/Swift.Bindings/src/DemangledSymbols.cs
@@ -43,7 +43,8 @@ public sealed class DemangledSymbolsRegister
     {
         try
         {
-            var descriptors = DemanglingResults.FromFile(dylibPath, Abi.ARM64).ProtocolConformanceDescriptors;
+            var abis = MachO.GetArchitectures(dylibPath);
+            var descriptors = DemanglingResults.FromFile(dylibPath, abis[0]).ProtocolConformanceDescriptors;
             var dictionary = new Dictionary<(NamedTypeSpec, NamedTypeSpec), string>();
 
             foreach (var descriptor in descriptors)
--- a/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
@@ -64,9 +64,9 @@ namespace BindingsGeneration
             var methodEnv = (MethodEnvironment)env;
             var signatureHandler = new SignatureHandler(methodEnv);
 
-            if (methodEnv.MethodDecl.GenericParameters.Any(x => x.Constraints.Count > 0))
+            if (methodEnv.MethodDecl.GenericParameters.Any(x => x.Constraints.OfType<AssociatedTypeConformance>().ToList().Count > 0))
             {
-                Console.WriteLine($"Method {methodEnv.MethodDecl.Name} has unsupported generic constraints");
+                Console.WriteLine($"Method {methodEnv.MethodDecl.Name} has unsupported PAT generic constraints");
                 return;
             }
 
@@ -292,6 +292,23 @@ namespace BindingsGeneration
         }
 
         /// <summary>
+        /// Handles the protocol conformances of the generic parameters of the method.
+        /// </summary>
+        public void HandleProtocolConformance()
+        {
+            foreach (var genericParameter in _env.MethodDecl.GenericParameters)
+            {
+                var conformances = genericParameter.Constraints.Where(c => c is ProtocolConformance).OrderBy(c => c.ProtocolSpec.NameWithoutModule);
+                foreach (var conformance in conformances)
+                {
+                    var protocolConformance = (ProtocolConformance)conformance;
+                    var pwtName = $"{_env.GenericTypeMapping[genericParameter.TypeName].PlaceholderName}_{protocolConformance.ProtocolSpec.NameWithoutModule}_pwt";
+                    AddParameter("ProtocolWitnessTable", pwtName);
+                }
+            }
+        }
+
+        /// <summary>
         /// Handles the SwiftSelf parameter of the method.
         /// </summary>
         public void HandleSwiftSelf()
@@ -376,6 +393,7 @@ namespace BindingsGeneration
                 pInvokeSignature.HandleSwiftAsync();
                 pInvokeSignature.HandleArguments();
                 pInvokeSignature.HandleGenericMetadata();
+                pInvokeSignature.HandleProtocolConformance();
                 pInvokeSignature.HandleSwiftSelf();
                 pInvokeSignature.HandleSwiftError();
                 _pInvokeSignature = pInvokeSignature.Build();
@@ -485,6 +503,7 @@ namespace BindingsGeneration
             EmitSwiftSelf(csWriter);
             EmitIndirectResultConstructor(csWriter);
             EmitGenericArguments(csWriter);
+            EmitProtocolWitnessTables(csWriter);
             EmitPInvokeCall(csWriter);
             EmitSwiftError(csWriter);
             EmitReturnConstructor(csWriter);
@@ -527,6 +546,7 @@ namespace BindingsGeneration
             EmitSwiftSelf(csWriter);
             EmitIndirectResultMethod(csWriter);
             EmitGenericArguments(csWriter);
+            EmitProtocolWitnessTables(csWriter);
             EmitPInvokeCall(csWriter);
             EmitSwiftError(csWriter);
             EmitReturnMethod(csWriter);
@@ -655,12 +675,31 @@ namespace BindingsGeneration
             foreach (var argument in _env.MethodDecl.CSSignature.Skip(1).Where(a => a.IsGeneric))
             {
                 var (typeName, metadataName, payloadName) = _env.GenericTypeMapping[argument.SwiftTypeSpec.ToString()];
+
                 var text = $$"""
                 var {{metadataName}} = TypeMetadata.GetTypeMetadataOrThrow<{{typeName}}>();
                 {{payloadName}} = (IntPtr)NativeMemory.Alloc({{metadataName}}.Size);
                 SwiftMarshal.MarshalToSwift({{argument.Name}}, {{payloadName}});
                 """;
                 csWriter.WriteLines(text);
+            }
+            csWriter.WriteLine();
+        }
+
+
+        private void EmitProtocolWitnessTables(CSharpWriter csWriter)
+        {
+            foreach (var genericParameter in _env.MethodDecl.GenericParameters)
+            {
+                var (placeholderName, _, _) = _env.GenericTypeMapping[genericParameter.TypeName];
+                var conformances = genericParameter.Constraints.Where(c => c is ProtocolConformance).OrderBy(c => c.ProtocolSpec.NameWithoutModule);
+                foreach (var conformance in conformances)
+                {
+                    var protocolConformance = (ProtocolConformance)conformance;
+                    var pwtName = $"{_env.GenericTypeMapping[genericParameter.TypeName].PlaceholderName}_{protocolConformance.ProtocolSpec.NameWithoutModule}_pwt";
+                    var protocolName = NameProvider.GetInterfaceName(protocolConformance.ProtocolSpec.NameWithoutModule);
+                    csWriter.WriteLine($"var {pwtName} = ProtocolWitnessTable.GetOrThrow<{placeholderName}, {protocolName}>();");
+                }
             }
             csWriter.WriteLine();
         }
--- a/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
@@ -37,26 +37,17 @@ namespace BindingsGeneration
         {
         }
 
-        /// <summary>
-        /// Marshals the specified constructor.
-        /// </summary>
-        /// <param name="methodDecl">The method declaration.</param>
-        /// <param name="typeDatabase">The type database instance.</param>
-        public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
+        /// <inheritdoc/>
+        public IEnvironment Marshal(BaseDecl baseDecl, ITypeDatabase typeDatabase)
         {
-            if (decl is not MethodDecl methodDecl)
+            if (baseDecl is not MethodDecl methodDecl)
             {
-                throw new ArgumentException("The provided decl must be a MethodDecl.", nameof(decl));
+                throw new ArgumentException("The provided decl must be a MethodDecl.", nameof(baseDecl));
             }
             return new MethodEnvironment(methodDecl, typeDatabase);
         }
 
-        /// <summary>
-        /// Emits the method declaration.
-        /// </summary>
-        /// <param name="csWriter">The IndentedTextWriter instance.</param>
-        /// <param name="env">The environment.</param>
-        /// <param name="conductor">The conductor instance.</param>
+        /// <inheritdoc/>
         public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment env, Conductor conductor)
         {
             var methodEnv = (MethodEnvironment)env;
@@ -114,26 +105,17 @@ namespace BindingsGeneration
         {
         }
 
-        /// <summary>
-        /// Marshals the method declaration.
-        /// </summary>
-        /// <param name="methodDecl">The method declaration.</param>
-        /// <param name="typeDatabase">The type database instance.</param>
-        public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
+        /// <inheritdoc/>
+        public IEnvironment Marshal(BaseDecl baseDecl, ITypeDatabase typeDatabase)
         {
-            if (decl is not MethodDecl methodDecl)
+            if (baseDecl is not MethodDecl methodDecl)
             {
-                throw new ArgumentException("The provided decl must be a MethodDecl.", nameof(decl));
+                throw new ArgumentException("The provided decl must be a MethodDecl.", nameof(baseDecl));
             }
             return new MethodEnvironment(methodDecl, typeDatabase);
         }
 
-        /// <summary>
-        /// Emits the method declaration.
-        /// </summary>
-        /// <param name="csWriter">The IndentedTextWriter instance.</param>
-        /// <param name="env">The environment.</param>
-        /// <param name="conductor">The conductor instance.</param>
+        /// <inheritdoc/>
         public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment env, Conductor conductor)
         {
             var methodEnv = (MethodEnvironment)env;
--- a/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
@@ -5,6 +5,81 @@ using System.CodeDom.Compiler;
 
 namespace BindingsGeneration
 {
+    /// <summary>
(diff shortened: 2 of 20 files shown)


src/Swift.Bindings/src/DemangledSymbols.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    public Dictionary<(NamedTypeSpec, NamedTypeSpec), string> ProtocolConformanceDescriptors { get; init; }
+
+    public DemangledSymbols(Dictionary<(NamedTypeSpec, NamedTypeSpec), string> descriptors)
+    {
+        ProtocolConformanceDescriptors = descriptors;
+    }
+}
+
+public sealed class DemangledSymbolsRegister
```
jkurdek, line 17: fyi @matouskozak As mentioned in the description this is just a temporary solution to getting symbols until https://github.com/dotnet/runtimelab/issues/2918 is implemented. I tried to put all the logic here so it can be easily removed. I initially thought we can just synthesise the symbols for now, but it is more complicated than I thought.

This is not gonna work for Apple shipped modules which dont have easily accessible dylib (CryptoKit or Foundation) as Demangler will fail to resolve the library. The error is not critical. It will result in `GetProtocolConformance` methods not being populated.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -5,6 +5,81 @@
 
 namespace BindingsGeneration
 {
+    /// <summary>
+    /// Factory class for creating instances of ConstructorHandler.
+    /// </summary>
+    public class ConstructorHandlerFactory : IFactory<BaseDecl, IMethodHandler>
```
jkurdek, line 11: Reverted some of the  changes made in #2946

kotlarmilos, line 11: Consider moving this to a separate file, i.e. `ConstructorHandler.cs`.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -453,66 +539,39 @@ internal WrapperEmitter(MethodEnvironment methodEnv, SignatureHandler signatureH
...
         /// </summary>
         /// <param name="writer">The IndentedTextWriter instance.</param>
-        private void EmitConstructor(CSharpWriter csWriter)
+        internal void EmitConstructor(CSharpWriter csWriter)
         {
             EmitSignatureConstructor(csWriter);
             EmitBodyStart(csWriter);
 
-            EmitDeclarationsForAllocations(csWriter);
+            // EmitDeclarationsForAllocations(csWriter);
```
kotlarmilos: Remove comments.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -519,16 +519,120 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
+        }
+
+        /// <summary>
+        /// Marshals the specified protocol declaration.
+        /// </summary>
+        /// <param name="protocolDecl">The protocol declaration.</param>
+        /// <param name="typeDatabase">The type database instance.</param>
+        public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
+        {
+            if (decl is not ProtocolDecl structDecl)
```
kotlarmilos: Nit: rename `structDecl` -> `ProtocolDecl`.

src/Swift.Bindings/src/Parser/GenericSignatureParser.cs
```diff
@@ -59,25 +59,33 @@ private static List<string> ExtractGenericParams(string signature)
...
+        return [.. signature[(whereIndex + "where".Length)..].Split(',').Select(ParseConstraint).Where(c => c is not null).Cast<ProtocolConformance>()];
     }
 
     /// <summary>
     /// Parses a constraint clause into a Conformance object.
     /// </summary>
     /// <param name="clause">The constraint clause to parse.</param>
     /// <returns>A Conformance object.</returns>
-    private static Conformance ParseConstraint(string clause)
+    private static ProtocolConformance? ParseConstraint(string clause)
```
kotlarmilos, line 84: What means when `null` is returned?

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Generics/GenericTests.swift
```diff
@@ -45,3 +45,125 @@ public func AcceptsGenericParameters<T, U> (a: T, b: U) throws -> Int {
...
+    public init(value: Double) {
+        self.value = value
+    }
+
+    public func increase() -> Double {
+        return value * 2
+    }
+}
+
+public func AcceptsIntContainer<T: Container>(a: T) -> Int where T.Element == Int {
```
kotlarmilos, line 174: Could you add a test case where two parameters conform to the same protocol?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
jkurdek: I would rather keep it all here until we go through Emitters design

src/Swift.Bindings/src/Parser/GenericSignatureParser.cs
jkurdek: It means that the constraint parsed refers to the associated type. Those are not necessary until we add full support for type constraints on C# side. They dont have any impact on ABI.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -665,6 +735,23 @@ private void EmitGenericArguments(CSharpWriter csWriter)
             csWriter.WriteLine();
         }
 
+
+        private void EmitProtocolWitnessTables(CSharpWriter csWriter)
+        {
+            foreach (var genericParameter in _env.MethodDecl.GenericParameters)
+            {
+                var placeholderName = _env.GenericTypeMapping[genericParameter.TypeName].PlaceholderName;
```
vitek-karas: Naming nit: Placeholder is confusing (at least to me). In C# these are called type parameters, or generic parameters. I've never heard the term "placeholder".
Is there a reason we chose this name?
The official docs seem to use "type parameter": https://docs.swift.org/swift-book/documentation/the-swift-programming-language/genericparametersandarguments/#Generic-Parameter-Clause

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -519,16 +519,120 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
             {
-                throw new NotImplementedException();
+                var dispatch = new Dictionary<Type, string>
+                {
+                    {{GenerateGetProtocolConformanceDictionaryEntries()}}
+                };
+
+                if (!dispatch.ContainsKey(typeof(TProtocol)))
+                {
+                    throw new SwiftRuntimeException($"Attempted to retrieve protocol conformance descriptor for type {{_structDecl.Name}} and protocol {typeof(TProtocol).Name}, but no conformance was found.");
```
vitek-karas, line 500: The second string parameter needs double `{{`.

vitek-karas: Maybe eventually this should turn into a C# based message - since the mistake is on the C# side. But it's not very important now.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -519,16 +519,120 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
         private void WriteGetProtocolConformanceDescriptor()
         {
+            var libPath = _typeDatabase.GetLibraryPath(_moduleDecl.Name);
             var text = $$"""
-            static ProtocolConformanceDescriptor ISwiftObject.GetProtocolConformanceDescriptor<U>()
+            static ProtocolConformanceDescriptor ISwiftObject.GetProtocolConformanceDescriptor<TProtocol>()
+                where TProtocol : class
             {
-                throw new NotImplementedException();
+                var dispatch = new Dictionary<Type, string>
```
vitek-karas: The runtime performance of this is not great.

The dictionary is initialized on every call (I don't think Roslyn will optimize this and JIT can't). We should turn it into a static field instead (and init in static .ctor to avoid threading problems)

The lookup is done twice - instead it should be something like:
```csharp
if (!dispatch.TryGet(typeof(TProtocol), var out symbolName)) throw;
LoadFromSymbol(symbolName);
```

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
jkurdek: No big reason. It used to be `TypeName` in the past https://github.com/dotnet/runtimelab/pull/2946/commits/eec9e3363fad28bb9d5da92e534b1f660716866e#diff-0d8888e3e8f8b2faf5860214283c2e3829a7f8668153de57ecbeb1ec9b7cd1d6

I think we could change to `typeParamaterName` or even `csTypeParamName` to indicate that this is a `c#` translation of the type parameter. 

This is valid swift 😅 
```swift
struct 📦<🎁> {
    var item: 🎁
}
```

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
jkurdek: This is actually correct, the substitution of the second string parameter happens during compilation of projections

vitek-karas: Right - sorry - this is too "meta" for me :-)

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
vitek-karas: > This is valid swift

1 - ROTFL
2 - OK, but why should we care? (as in, I won't lose any sleep if we don't support this anytime in this decade)

jkurdek: This is just an extreme example to show that swift has much more freedom in naming stuff. Less extreme example would be type param in swift being keyword in c#. I do not think we need to outright guarantee support for such names in the projection tooling, but it is good to remember that we need to sanitize/translate any typename before using it in projections.

vitek-karas: I definitely doesn't question that we need to sanitize the names...

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -519,28 +519,44 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
+        {
+            var text = $$"""
+            private static Dictionary<Type, string> _protocolConformanceSymbols;
 
+            static {{_structDecl.Name}}()
+            {
+                _protocolConformanceSymbols = new Dictionary<Type, string>
+                {
+                    {{GenerateGetProtocolConformanceDictionaryEntries()}}
+                };
```
vitek-karas, line 524: Long term idea (just so that others hear it, no action right now): With the type mapping effort we will likely introduce some way to store frozen dictionaries directly in the binary with basically 0 init cost. This would be a good candidate for that functionality.

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Generics/GenericTests.swift
```diff
@@ -45,3 +45,144 @@ public func AcceptsGenericParameters<T, U> (a: T, b: U) throws -> Int {
...
+    public var value: Double
+
+    public init(value: Double) {
+        self.value = value
+    }
+
+    public func increase() -> Double {
+        return value * 2
+    }
+}
```
matouskozak: I think this one is not used anywhere.

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Protocols/ProtocolsTests.cs
```diff
@@ -0,0 +1,38 @@
...
+        public class TestFixture
+        {
+            static TestFixture()
+            {
+                InitializeResources();
+            }
+
+            private static void InitializeResources()
+            {
+                // Initialize
```
matouskozak, line 27: Should there be something or is it just a placeholder?

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Protocols/ProtocolsTests.cs
```diff
@@ -0,0 +1,38 @@
...
+            }
+        }
+
+        [Fact]
+        public void ProtocolIsProjected()
+        {
+            // This test is to ensure that the protocol is projected, it just needs to compile
+            Assert.NotNull(typeof(ISwiftPrintable));
```
matouskozak: I'm unsure how this test works. Could you explain how compiling the source code means that we verified that `ISwiftPrintable` is projected?

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -341,7 +358,31 @@ private StructDecl CreateStructDecl(Node node, BaseDecl parentDecl, ModuleDecl m
         /// <returns>The class declaration.</returns>
         private ClassDecl CreateClassDecl(Node node, BaseDecl parentDecl, ModuleDecl moduleDecl)
         {
+            var fullyQualifiedName = ExtractFullyQualifiedName(parentDecl.FullyQualifiedName, node.Name);
             return new ClassDecl
+            {
+                Name = ExtractUniqueName(node.Name),
+                FullyQualifiedName = ExtractFullyQualifiedName(parentDecl.FullyQualifiedName, node.Name),
```
matouskozak: Why do we need to `ExtractFullyQualifiedName` again (line 361 does the same)?

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -317,14 +332,16 @@ private List<BaseDecl> CollectDeclarations(IEnumerable<Node> nodes, BaseDecl par
         /// <returns>The struct declaration.</returns>
         private StructDecl CreateStructDecl(Node node, BaseDecl parentDecl, ModuleDecl moduleDecl, bool hasFrozenAttribute)
         {
+            var fullyQualifiedName = ExtractFullyQualifiedName(parentDecl.FullyQualifiedName, node.Name);
```
matouskozak, line 335: What is the motivation behind moving the `ExtractFullyQualifiedName` to a separate variable assignment instead of doing it the previous way (inside the `new StructDecl`)?

src/Swift.Bindings/src/Parser/GenericSignatureParser.cs
```diff
@@ -59,25 +59,33 @@ private static List<string> ExtractGenericParams(string signature)
...
+    private static List<ProtocolConformance> ExtractConstraints(string signature)
     {
         var whereIndex = signature.IndexOf("where", StringComparison.OrdinalIgnoreCase);
         if (whereIndex == -1)
-            return new List<Conformance>();
+            return new List<ProtocolConformance>();
 
-        return signature[(whereIndex + "where".Length)..].Split(',').Select(ParseConstraint).ToList();
+        return [.. signature[(whereIndex + "where".Length)..].Split(',').Select(ParseConstraint).Where(c => c is not null).Cast<ProtocolConformance>()];
     }
```
matouskozak: Nit. I wonder how long line we will eventually get :D

src/Swift.Bindings/src/Marshaler/NameProvider.cs
```diff
@@ -7,10 +7,10 @@ namespace BindingsGeneration;
 /// <summary>
 /// Represents a generic parameter name mapping.
 /// </summary>
-/// <param name="PlaceholderName">The name of the generic type parameter e.g. T0.</param>
+/// <param name="TypeParameter">The name of the generic type parameter e.g. T0.</param>
 /// <param name="MetadataName">The name of the metadata type parameter.</param>
 /// <param name="PayloadName">The name of the payload type parameter. </param>
-public record struct GenericParameterCSName(string PlaceholderName, string MetadataName, string PayloadName);
+public record struct GenericParameterCSName(string TypeParameter);
```
matouskozak: Should these two get removed since the signature only accepts `string TypeParameter` now?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -519,16 +519,136 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
+            static ProtocolConformanceDescriptor ISwiftObject.GetProtocolConformanceDescriptor<TProtocol>()
+                where TProtocol : class
             {
-                throw new NotImplementedException();
+                if (!_protocolConformanceSymbols.TryGetValue(typeof(TProtocol), out var symbolName))
+                {
+                    throw new SwiftRuntimeException($"Attempted to retrieve protocol conformance descriptor for type {{_structDecl.Name}} and protocol {typeof(TProtocol).Name}, but no conformance was found.");
+                }
+
+                return ProtocolConformanceDescriptor.LoadFromSymbol("{{libPath}}", symbolName);
```
matouskozak, line 503: Do I understand it correctly that we're planning to read symbols from dylibs during runtime? I'm wondering from the security standpoint what constrains we should put on the read dylibs to unsure that we're handling correctly untrusted inputs.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -519,16 +519,136 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
+        {
+        }
+
+        /// <summary>
+        /// Marshals the specified protocol declaration.
+        /// </summary>
+        /// <param name="protocolDecl">The protocol declaration.</param>
+        /// <param name="typeDatabase">The type database instance.</param>
+        public IEnvironment Marshal(BaseDecl decl, ITypeDatabase typeDatabase)
+        {
```
matouskozak: The param names in the comment don't match the signature. Is this intentional?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -519,16 +519,136 @@ IntPtr ISwiftObject.MarshalToSwift(IntPtr swiftDest)
...
+
+        /// <summary>
+        /// Emits the code for the specified environment.
+        /// </summary>
+        /// <param name="writer">The IndentedTextWriter instance.</param>
+        /// <param name="env">The environment.</param>
+        /// <param name="conductor">The conductor instance.</param>
+        /// <param name="typeDatabase">The type database instance.</param>
+        public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment env, Conductor conductor)
+        {
```
matouskozak: The documentation comment doesn't match the signature. Similar case is in `MethodHandler.cs` file.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -5,6 +5,81 @@
...
+        /// <param name="env">The environment.</param>
+        /// <param name="conductor">The conductor instance.</param>
+        public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment env, Conductor conductor)
+        {
+            var methodEnv = (MethodEnvironment)env;
+            var signatureHandler = new SignatureHandler(methodEnv);
+
+            if (methodEnv.MethodDecl.IsGeneric)
+            {
+                Console.WriteLine($"Constructor {methodEnv.MethodDecl.Name} has unsupported generic parameters");
```
matouskozak, line 59: Not related to this PR but maybe we should introduce a more comprehensive logging mechanism instead of using `Console.WriteLine`.

matouskozak (commented):
Massive piece of work Jeremi, good job!

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Protocols/ProtocolsTests.cs
jkurdek: I think all our test files follow this pattern. It is a placeholder. We might go and delete those everywhere if they are not necessary.

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
jkurdek: It is used in two places inside of new StructDecl so I hoisted it to avoid double computation

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
jkurdek: My understanding is that we can safely use `MachO` to retrieve the symbols during projection generation - as this doesn't "load' the dylib. However, we cannot load the values behind those symbols until runtime due to potential OS difference. 

I'm not sure how this differs from pinvokes where we do the same thing (compute symbol during projection tooling and call into dylib during runtime)

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
jkurdek: I agree,  https://github.com/dotnet/runtimelab/issues/2973

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Protocols/ProtocolsTests.cs
jkurdek: Should be a bit more readable now. The idea is that we currently project protocols as empty interfaces so there is no much to test except the fact that it exists and it is an interface.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
jkurdek: Fixed

kotlarmilos (approved):

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
vitek-karas: Related to that - we discussed having "transaction writers" for the cases where we fail when writing something. Until then, can we mark the places in the code where that should happen somehow? Or maybe we should throw for now?

jkurdek: I would not throw yet. As this will basically make us completely unable to run the tooling on some frameworks. But I agree that this is a prime candidate for the future "revert" functionality. 

An alternative would be to have a two layered approach (we have a set of rules which need to be satisfied before projecting - if they fail we do not project), and then during writing if we see something unexpected we throw. So in this case we could have a rule that we do not project types with generic init methods. I do realise this might be too idealistic as we would have to do heavy lifting on marshalling layer.

For now I can put a comment, and we can revisit it in milestone 2.

vitek-karas: (My future 2 cents: two code paths which have to agree on set of rules == trouble, we shortly tried that in the trimmer and it didn't go very well)

Issue #2924, opened 2025-01-13
[Projection tooling] Provide mapping between TypeDecls, TypeSpecs and TypeDatabase entries


jkurdek:
`TypeDatabase` entries are currently queried using swift name - module + type name. We need a uniform way of queuing TypeDatabase when we encounter a TypeDecl or NamedTypeSpec during marshalling.

PR #2980, opened 2025-02-04
[SwiftBindings] Introduce SwiftTypeName model for nominal swift names

jkurdek:
Fixes https://github.com/dotnet/runtimelab/issues/2924

This PR introduces a new type `SwiftTypeName` whose goal is to model nominal type / protocol names coming from swift. It is designed to be used as a key when obtaining information about a type from different mappings (TypeDatabase, ProtocolConformance symbols etc.)

The motivation behind having such a type as opposed to using combination of NamedTypeSpecs and raw strings is to:
- Centralize name construction 
- Make sure we are handling a valid type name (NamedTypeSpec can be anything - not necessarily a valid type)
- Catch potential surprises early (if a type doesn't have module, it is probably an error) 

This also splits ProtocolConformance into two types:
- One modelling conformances of generic attributes (T : sth)
- Other modelling conformances of types (Foo: sth)


src/Swift.Bindings/src/Model/TypeDecl/ProtocolConformance.cs
```diff
@@ -7,9 +7,19 @@ namespace BindingsGeneration;
...
+);
+
+/// <summary>
+/// Represents a generic parameter conformance.
+/// </summary>
+/// <param name="GenericParameter">The generic parameter</param>
+/// <param name="ProtocolSpec">The protocol that the generic parameter conforms to</param>
+public record GenericParameterConformance(
+    NamedTypeSpec GenericParameter,
+    SwiftTypeName ProtocolSpec
```
vitek-karas: Why is it called `ProtocolSpec` here, but in the `ProtocolConformance` it's called just `Protocol` - maybe I just don't know the difference.

jkurdek: No reason, thanks for catching that!

src/Swift.Bindings/src/TypeDatabase/TypeDatabaseExtensions.cs
```diff
@@ -13,36 +16,52 @@ public static class TypeDatabaseExtensions
     /// <returns>True if the type has been processed; otherwise, false.</returns>
     public static bool IsTypeProcessed(this ITypeDatabase typeDatabase, TypeSpec typeSpec)
```
jkurdek, line 17: I believe TypeDatabase should only store nominal types—those with explicit declarations in the abi.json files. Therefore, we shouldn’t query it using `TypeSpec`, as `TypeSpec` can represent tuples, closures, generic parameters (like `T`), or other non-nominal constructs.

Looking ahead, I think we'll need an additional layer to handle complex `TypeSpec` cases. This layer would decompose them into their nominal components before querying the `TypeDatabase`.

I introduced the current set of `TypeSpec`-accepting methods to minimize the impact on other areas, particularly MethodHandler, which determines whether to project a method based on the presence of `AnyType` in its signature.

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeName.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// <summary>
+    /// The type name.
+    /// </summary>
+    public string Name { get; }
+
+    /// <summary>
+    /// The module-qualified type name. Includes all parent types.
+    /// </summary>
+    public string ModuleQualifiedName { get; }
```
kotlarmilos, line 24: Is this the unique type name in the type database?

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeName.cs
```diff
@@ -0,0 +1,63 @@
...
+        Name = name;
+        ModuleQualifiedName = moduleQualifiedName;
+    }
+
+    /// <summary>
+    /// Creates a new SwiftTypeName from a module-qualified name.
+    /// </summary>
+    /// <param name="moduleQualifiedName">The module-qualified name.</param>
+    /// <returns>The SwiftTypeName.</returns>
+    public static SwiftTypeName FromModuleQualifiedName(string moduleQualifiedName)
```
kotlarmilos, line 41: Does this function need to handle generics?

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeName.cs
```diff
@@ -0,0 +1,63 @@
...
+            throw new ArgumentException($"Invalid module-qualified name: {moduleQualifiedName}");
+        }
+
+        return new SwiftTypeName(parts.First(), parts.Last(), moduleQualifiedName);
+    }
+
+    /// <summary>
+    /// Swift type name for void.
+    /// </summary>
+    public static readonly SwiftTypeName VoidType = new SwiftTypeName(string.Empty, "()", "()");
```
kotlarmilos, line 57: This could be set in the type database initializer.

src/Swift.Bindings/src/Model/TypeDecl/ProtocolConformance.cs
```diff
@@ -7,9 +7,19 @@ namespace BindingsGeneration;
...
+    SwiftTypeName Protocol
+);
+
+/// <summary>
+/// Represents a generic parameter conformance.
+/// </summary>
+/// <param name="GenericParameter">The generic parameter</param>
+/// <param name="Protocol">The protocol that the generic parameter conforms to</param>
+public record GenericParameterConformance(
+    NamedTypeSpec GenericParameter,
```
kotlarmilos: Could we encapsulate all information in the `SwiftTypeName` and use `NamedTypeSpec` in the demangler only?

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeNameConverter.cs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace BindingsGeneration;
+
+/// <summary>
+/// Utility class for converting things to Swift type names.
+/// </summary>
+public static class SwiftTypeNameConverter
```
kotlarmilos, line 9: Is this layer of abstraction useful?

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -527,10 +523,13 @@ private static string ExtractUniqueName(string name)
...
 
-        private static string ExtractFullyQualifiedName(string parentName, string name)
-        {
-            return string.IsNullOrEmpty(parentName) ? ExtractUniqueName(name) : $"{parentName}.{ExtractUniqueName(name)}";
-        }
+        private static SwiftTypeName GetSwiftTypeName(BaseDecl parentDecl, string name)
+            => parentDecl switch
+            {
+                ModuleDecl moduleDecl => SwiftTypeName.FromModuleQualifiedName($"{moduleDecl.Name}.{name}"),
+                TypeDecl typeDecl => SwiftTypeName.FromModuleQualifiedName($"{typeDecl.SwiftTypeName.ModuleQualifiedName}.{name}"),
```
kotlarmilos, line 530: Does `SwiftTypeName` exist at invocation time? Could we ensure it is not called if exists?

src/Swift.Bindings/src/TypeDatabase/TypeDatabaseExtensions.cs
```diff
@@ -3,6 +3,9 @@
 
 namespace BindingsGeneration;
 
+// TODO: TypeDatabase should hold only nominal types (represented by NamedTypeSpec). Specifically tuples, closures etc. should not reside inside TypeDatabase.
```
kotlarmilos, line 6: Could you expand this? Not sure what is the intent.

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeName.cs
jkurdek: Once we have a mature generic types solution - no. The type database would contain open generic types. And when a new bound generic is encountered a special handler would dissect into a set of nominal types (open generic, type for each generic paramater etc). Then each of the nominal types could be read from the database. 

Currently we keep some generic types in the type database (as a workaround) so this will be invoked with generic params (they are not parsed though) so e.g. "Module.Foo<UInt8> "would just be hashed and used to query the type database.

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeName.cs
jkurdek: It is special cased as I dont consider `VoidType` as a nominal type form Swift perspective. It is an empty tuple so in my eyes "it doesnt belong" to the type database. I can imagine having a `TupleHandler` in the future which would handle the empty tuple gracefully.

src/Swift.Bindings/src/Model/TypeDecl/ProtocolConformance.cs
jkurdek: So here GenericParameter means something like `T`. I don't see that as `SwiftTypeName` (as it is not a "real" type but a placeholder). We could probably use raw strings here to not use `NamedTypeSpec` or introduce some thin string wrapper for type safety. 

Moving away from using TypeSpecs outside of demangler will not be trivial. Basically the problem is that we parse some method declaration and we see e.g. a bound generic type. We need some mechanism which will allow us to extract the open generic name and names of each of the type arguments. NamedTypeSpec has this functionality. I think we should discuss this further.

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeNameConverter.cs
jkurdek: I'm fine with merging this with `SwiftTypeName` if you feel that is too much. The idea was to extract code which is subject to change into separate class.

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
jkurdek: I'm not sure I understand. you mean the parent's `SwiftTypeName` ?

src/Swift.Bindings/src/TypeDatabase/TypeDatabaseExtensions.cs
jkurdek: When processing a module, suppose we encounter a tuple like `(Int, Double, Foo, Boo, Int, Double)`. The idea here is that this tuple should not be added to the `TypeDatabase`. Instead, the individual types—`Int`, `Double`, `Foo`, and `Boo`—are stored in the `TypeDatabase`, and we construct the corresponding C# projection for the tuple using these types.

This gives us the following distinction:

- Types that need to be explicitly stored in the `TypeDatabase`, such as fundamental types and named types.
- Types that can be constructed from existing types, like tuples, bound generics, and closures, which are built dynamically based on the types already present.

I think this this will give us clarity - `TypeDatabase` remains focused on core types, making it easier to understand and maintain; and consistency - when encountering a `TypeDecl` during ABI parsing, we create a new record tied to the module where the type was found, which keeps module associations clear.

If we allow constructed types (like tuples or bound generics) in the TypeDatabase, we run into problems with module ownership. For example, if `Foo` is defined in module `A` and we later encounter `Foo<Bar>` in module `B`, it's unclear which module `Foo<Bar>` belongs to.

One downside of not caching these constructed types in the `TypeDatabase` is performance. Each time we encounter the same complex tuple / closure / generic , we have to re-read the `TypeDatabase` for its components. This can be addressed with an in-memory cache for constructed types, separate from the persistent `TypeDatabase.` One such attempt was the previously removed `OutOfModuleTypeRecords`.

src/Swift.Bindings/src/Model/TypeDecl/ProtocolConformance.cs
kotlarmilos: Yes -- @matouskozak has an idea to utilize the Swift demangler instead which provides an AST.

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeNameConverter.cs
kotlarmilos: No need, let's review it at the design session.

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
kotlarmilos: Sorry -- Is this only invoked at ABI parsing time? I wonder if it could be called multiple times for a type.

kotlarmilos (approved):
LGTM, thanks!

src/Swift.Bindings/src/Model/TypeNames/SwiftTypeName.cs
jkurdek: Yes - this should represent the full module qualified path to a type

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
jkurdek: This should be invoked once for each type decl during parsing.

PR #2983, opened 2025-02-06
[Swift language feature] Implement `Swift.String` wrapper in C#

### LINKED ISSUE #2831: [Swift language features] Implement Swift.String in C#
> [!NOTE]  
> This issue tracks progress and will be expanded with more details. If specific tasks for this issue are not clear, this means that the task is not well defined yet and needs attention.

kotlarmilos:
## Description

This PR projects `Swift.String` used in [products(for:)](https://developer.apple.com/documentation/storekit/product/products(for:)).

In Swift, String is defined as `frozen struct String` with an internal payload of `Swift.Data`. The projection is a thin C# wrapper around `Swift.Data`, with methods to convert between `UTF-8` and `UTF-16` encodings.

## Out of Scope

 - In Swift, the string payload requires release of internals, which is not implemented in this PR. The plan is to review and refactor memory handling of projected types as outlined in #2851.
 - This projection does not expose the full API surface of `Swift.String`.

Fixes #2831


src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,170 @@
...
+    /// Gets the protocol conformance descriptor for the given type.
+    /// </summary>
+    /// <typeparam name="TProtocol"></typeparam>
+    /// <returns></returns>
+    static ProtocolConformanceDescriptor ISwiftObject.GetProtocolConformanceDescriptor<TProtocol>()
+        where TProtocol : class
+    {
+        if (!_protocolConformanceSymbols.TryGetValue(typeof(TProtocol), out var symbolName))
+        {
+            throw new SwiftRuntimeException($"Attempted to retrieve protocol conformance descriptor for type SwiftSet and protocol {typeof(TProtocol).Name}, but no conformance was found.");
```
vitek-karas:
```suggestion
            throw new SwiftRuntimeException($"Attempted to retrieve protocol conformance descriptor for type SwiftString and protocol {typeof(TProtocol).Name}, but no conformance was found.");
```

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,170 @@
...
+            var resultType = TypeMetadata.GetTypeMetadataOrThrow<long>();
+
+            var len = Length;
+            if (len <= 0)
+                return string.Empty;
+
+            string? str = null;
+
+            var arr = GetUtf8ContiguousArray(_payload);
+            GCHandle handle = GCHandle.Alloc(str, GCHandleType.Pinned);
```
vitek-karas: Why do we need to pin the str?
We don't use it anywhere as far as I can tell. The callback will end up calling a closure method (the IL version of this generates probably a struct which stores the str local and the lambda as an instance method, and the below call create a native callable delegate over that lambda) - so it will have access to the str local just fine (and GC will understand the whole thing). Additionally the `str` at this point is always `null` so we're creating a GCHandle for `null` which is not very useful (I'm honestly surprised this doesn't throw).

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,170 @@
...
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sSS5countSivg")]
+    public static extern long GetLength(Data str);
+
+    [UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sSS11utf8CStrings15ContiguousArrayVys4Int8VGvg")]
+    public static unsafe extern IntPtr GetUtf8ContiguousArray(Data str);
+
+    [UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$ss15ContiguousArrayV15withUnsafeBytesyqd__qd__SWKXEKlF")]
+    public static extern unsafe IntPtr WithUnsafeBytes(CallbackDelegate callback, IntPtr context, IntPtr contiguousArray, TypeMetadata elementType, TypeMetadata resultType);
```
vitek-karas: I'm curious, is there some doc which described these functions?
It would be really nice to link to that doc here.

vitek-karas:
I think we should have a discussion if we want to project string (which is what this PR does) or if we want to always marshal string. Personal preference would be to marshal string whenever possible (params, return value), but I guess we need the projection just like we will likely need boxes for other primitive types (mainly due to protocol conformance).
Either way we should have this decision discussed and written down somewhere.

src/Swift.Runtime/src/Swift/SwiftString.cs
kotlarmilos: I wasn't sure how the GC would mark str when it is used in the callback invoked from Swift.

src/Swift.Runtime/src/Swift/SwiftString.cs
kotlarmilos: Added:
 - https://developer.apple.com/documentation/swift/string/utf8cstring
 - https://developer.apple.com/documentation/swift/contiguousarray/withunsafebytes(_:)

kotlarmilos:
> I think we should have a discussion if we want to project string 

Let's collect input on this PR.

> Personal preference would be to marshal string whenever possible (params, return value)

Could you provide more details?

 > I guess we need the projection just like we will likely need boxes for other primitive types (mainly due to protocol conformance).

Yes + it is lowered as `(i64, ptr)`.

src/Swift.Runtime/src/Swift/SwiftString.cs
vitek-karas: I'm definitely no expert on this, but the sign is the fact that you don't use the `handle` for anything. Pinning an object should not be observable from managed code - pinned object should behave just like any other object.
To be honest, I don't even know why we support pinning ref types - since you should never access the memory in it anyway (memory layout for ref types is implementation detail) - but I guess there's some reason for it.

The other reason is that in this case you actually don't want to pin the string value (which is what the call does) you want to "pin" the `str` variable so that you can assign to it from the callback. As noted, pinning `null` is really meaningless.

But the callback is managed, so that doesn't matter - the pinning (if there's any) will happen on the PInvoke boundary where the native delegate is created and when that delegate gets called.

vitek-karas:
More details on the personal preference:
Let's say we have a swift code like this:
```swift
Label("Your name please:")  // Assuming Label init takes a string parameter, which it doesn't, but it's a good example
```

When projected to C# I would want to be able to write:
```csharp
new Label("Your name please:");
```

Basically I should not be forced to write things like:
```csharp
new Label(new SwiftString("Your name please:"));
```

But we might be able to achieve this with implicit casts.

src/Swift.Runtime/src/Swift/SwiftString.cs
jkotas: > To be honest, I don't even know why we support pinning ref types - since you should never access the memory in it anyway (memory layout for ref types is implementation detail) - but I guess there's some reason for it.

The layout of individual fields is well-defined. For example, pinning MyClass is useful if you want to pass pointer to `MyClass._returnValue` to an async callback in
```csharp
class MyClass
{
    int _returnValue;
}
```

kotlarmilos:
> When projected to C# I would want to be able to write:

> new Label("Your name please:");

Yes, this is a UX decision.

The preferred Swift encoding is UTF-8: https://www.swift.org/blog/utf8-string/. Currently, we access it through [ContiguousArray](https://developer.apple.com/documentation/swift/contiguousarray) and use a closure to access the array’s contiguous storage.

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,164 @@
...
+
+    public unsafe delegate IntPtr CallbackDelegate(IntPtr param);
+
+    [UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
+    [DllImport(KnownLibraries.SwiftCore, EntryPoint = "$sSSMa")]
+    public static extern TypeMetadata PInvoke_getMetadata();
+
+    [UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
+    [DllImport(KnownLibraries.SwiftCore, CharSet = CharSet.Unicode, EntryPoint = "$sSS21_builtinStringLiteral17utf8CodeUnitCount7isASCIISSBp_BwBi1_tcfC")]
+    public static unsafe extern Data Create(byte* str, long len, byte flag);
```
matouskozak: Why do we sometimes prefix the pinvoke signature with `PInvoke_` and here we don't?

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Structs/StructsTests.swift
```diff
@@ -257,3 +257,18 @@ public func sumArray(array: Array<Int32>) -> Int32
 {
     return array.reduce(0, +)
 }
+
+public func getString() -> String
```
jkurdek: I understand this is here as it requires some swift code. But I dont think the label 'struct tests' describes it very well. Maybe we should create a new test file?

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Text;
+using Swift.Runtime;
+using Swift.Runtime.InteropServices;
+
+namespace Swift;
+
+/// <summary>
+/// Represents Foundation.Data type.
+/// </summary>
+public struct Data
```
jkurdek: NIT: This is used outside of string extensively right? I think we encountered pointers using that when projecting crypto kit - maybe it deserves it own file

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/// <summary>
+/// Represents a Swift string with Foundation.Data payload.
+/// </summary>
+public struct SwiftString : ISwiftObject
+{
+    static nuint _payloadSize = SwiftObjectHelper<SwiftString>.GetTypeMetadata().Size;
+
+    private Data _payload;
```
jkurdek, line 23: NIT: code style I think the runtime guidelines suggest prefixing every member with access modifier. 

EDIT: Meant line above

jkurdek (approved):

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,164 @@
...
+
+    private static Dictionary<Type, string> _protocolConformanceSymbols;
+
+    static SwiftString()
+    {
+        _protocolConformanceSymbols = new Dictionary<Type, string> { };
+    }
+
+    public static nuint PayloadSize => _payloadSize;
```
vitek-karas, line 32: Just curious - why are we exposing these as public? I would expect the projection to "hide" the specifics of the payload and calling conv for me. And unless we have a specific need, we should hide it.

jkurdek: Those are public now because of the design of the `MarshalToSwift` API. We need to know how big something is to be able to allocate enough space before calling `MarshalToSwift`. Once we redesign this API this can become private.

src/Swift.Runtime/src/Swift/SwiftString.cs
jkotas: Prefixing PInvokes with `PInvoke_` prefix is unusual C# naming convention. I do not think that I have seen a project that does this.

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,164 @@
...
+            var resultType = TypeMetadata.GetTypeMetadataOrThrow<long>();
+
+            var len = Length;
+            if (len <= 0)
+                return string.Empty;
+
+            string? str = null;
+
+            var arr = PInvoke_GetUtf8ContiguousArray(_payload);
+            PInvoke_WithUnsafeBytes(bytes =>
```
jkotas: How much do you care about performance? 

This will allocate a marshalled delegate on every call that is quite expensive operation. It would be a lot more efficient to use function pointers and use the `context` argument to pass the local context.

kotlarmilos: Yes, this is a good point. We should implement the correct layout and memory management for the closure context. Also, we should be able to reduce the number of calls using a thin wrapper.

Created a tracking issue: https://github.com/dotnet/runtimelab/issues/2990

src/Swift.Runtime/src/Swift/SwiftString.cs
kotlarmilos: Yes, sounds good.

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Structs/StructsTests.swift
kotlarmilos: Yes, sounds good.

src/Swift.Runtime/src/Swift/SwiftString.cs
kotlarmilos: What’s another approach? We want a way to distinguish between P/Invokes and wrapper methods, without relying solely on the parameters overload.

vitek-karas: We could use `swift_functionName` - would make it very clear that we're transitioning to swift at that point.

jkurdek: I would separate this from this PR as we use the `PInvoke_` prefix extensively across the whole repository:
https://github.com/dotnet/runtimelab/issues/2991

src/Swift.Runtime/src/Swift/SwiftString.cs
kotlarmilos: Moved and created a tracking issue: https://github.com/dotnet/runtimelab/issues/2992.

kotlarmilos:
Follow-up tasks:
 - https://github.com/dotnet/runtimelab/issues/2851
 - https://github.com/dotnet/runtimelab/issues/2992

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,155 @@
...
+                        str = Encoding.UTF8.GetString((byte*)bytes, len);
+                        return IntPtr.Zero;
+                    }
+                }, IntPtr.Zero, arr, elementType, resultType);
+
+            return str!;
+        }
+    }
+
+    public unsafe delegate IntPtr CallbackDelegate(IntPtr param);
```
jkotas: Should the callback have default calling convention or Swift calling convention? (Related to my other comment about function pointers.)

src/Swift.Runtime/src/Swift/SwiftString.cs
jkotas: Doing this properly using function pointers is like 10-line edit on the new code added in this PR...

src/Swift.Runtime/src/Swift/SwiftString.cs
jkotas: Also, should the callback signature have the closure `context` as the argument?

src/Swift.Runtime/src/Swift/SwiftString.cs
kotlarmilos: Ok, let's fix that in this PR.

src/Swift.Runtime/src/Swift/SwiftString.cs
kotlarmilos: Yes.

`ToString` has been refactored to use function pointers. The `length` is passed through the closure context to the callback via the self register. A utf8 buffer is allocated, and the content of the contiguous array is copied into it. The buffer is then marshalled using `PtrToStringUTF8`, and the native buffer is deallocated.

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,157 @@
...
+    public static unsafe IntPtr WithUnsafeBytesCallback(byte* bytes, SwiftSelf context)
+    {
+        int length = *(int*)context.Value;
+        void* utf8Ptr = NativeMemory.Alloc((nuint)(length + 1));
+        Buffer.MemoryCopy(bytes, utf8Ptr, length, length);
+
+        ((byte*)utf8Ptr)[length] = 0;
+
+        return (IntPtr)utf8Ptr;
+    }
```
jkotas:
```suggestion

     private struct ToStringCallbackContext
     {
         public int _length;
         public string _returnString;
     }
```

src/Swift.Runtime/src/Swift/SwiftString.cs
```diff
@@ -0,0 +1,157 @@
...
+            return string.Empty;
+
+        var contiguousArray = PInvoke_GetUtf8ContiguousArray(_payload);
+        unsafe
+        {
+            IntPtr utf8Ptr = PInvoke_WithUnsafeBytes(&WithUnsafeBytesCallback, (IntPtr)(int*)&length, contiguousArray, elementType, resultType);
+            string result = Marshal.PtrToStringUTF8(utf8Ptr, length);
+            NativeMemory.Free((void*)utf8Ptr);
+            return result;
+        }
```
jkotas:
```suggestion
        unsafe
        {
            ToStringCallbackContext callbackContext;
            callbackContext._length = length;
            PInvoke_WithUnsafeBytes(&Callback, (IntPtr)(void*)&callbackContext, contiguousArray, elementType, resultType);
            return callbackContext._returnString;

            [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvSwift) })]
            static IntPtr Callback(byte* bytes, SwiftSelf context)
            {
                ToStringCallbackContext* pContext = (ToStringCallbackContext*)context.Value;
                pContext->_returnString = Encoding.UTF8.GetString(new ReadOnlySpan<byte>(bytes, pContext->_length));
                return default;
            }
        }
```

jkotas (commented):
This is a better way to do it without an extra copy

kotlarmilos:
> This is a better way to do it without an extra copy

Changed `ToStringCallbackContext` to a ref type to allow the callback to take an unmanaged pointer, as it contains managed fields.

vitek-karas (approved):
Thanks!

kotlarmilos:
@jkotas Thanks for help! If you have any additional feedback regarding the function pointers, please let me know and we can address them in a follow-up.

jkotas:
> Changed ToStringCallbackContext to a ref type to allow the callback to take an unmanaged pointer, as it contains managed fields.

That's just an annoying C# warning. It is fine to disable it. We have it disabled globally in dotnet/runtime. We use the pattern that I have suggested in many places. One example from many: https://github.com/dotnet/runtime/blob/c6fd0543b67aec813b16fc9458afd2ec2716178b/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureData.Nls.cs#L370-L403 

There is nothing wrong with taking an address of a struct on stack that contains object references. You just need to be careful about what you are doing - like you need to be careful everywhere else in unsafe code. GCHandle and the reference type instead of struct are just a bunch of unnecessary overhead.

Issue #2971, opened 2025-01-31
[Swift language features] Add support for property accessors


kotlarmilos:
## Description

Surface property setters and getters into C#. We should support both stored and computed properties
- [ ] #2987 
- [ ] #2995
- [ ] #2996
- [ ] https://github.com/dotnet/runtimelab/issues/3013

PR #2988, opened 2025-02-10
[SwiftBindings] Document properties bindings

jkurdek:
Adds a proposal on how to handle instance properties bindings for Swift structs. While getters are simple - they have the same calling convention as as regular instance function which accesses data on a struct, the setters are complex. They accepts the struct using the `SwiftSelf` pointer and modify it inplace.

Contributes to https://github.com/dotnet/runtimelab/issues/2971

### DIFF
--- a/src/docs/binding-properties.md
+++ b/src/docs/binding-properties.md
@@ -53,21 +53,10 @@ public struct FrozenPoint {
 
     private unsafe void Set_x(nint value)
     {
-        void* payload;
-        try 
+        fixed (FrozenPoint* p = &this)
         {
-            var metadata = ...;
-            payload = (SwiftHandle)NativeMemory.Alloc(returnMetadata.Size);
-            SwiftMarshal.MarshalToSwift(this, (IntPtr)payload);
-            var self = new SwiftSelf(payload);
-
+            var self = new SwiftSelf(p);
             PInvoke_set_x(value, self);
-
-            this = SwiftMarshal.MarshalFromSwift<FrozenPoint>(payload);
-        }
-        finally
-        {
-            NativeMemory.Free(payload);
         }
     }
 
--- a/src/docs/binding-properties.md
+++ b/src/docs/binding-properties.md
@@ -101,3 +101,7 @@ public class NonFrozenPoint : IDisposable {
 ## Static Properties
 
 TODO
+
+## Async Properties
+
+TODO
--- a/src/docs/binding-properties.md
+++ b/src/docs/binding-properties.md
@@ -1,0 +1,114 @@
+# Properties
+
+## Simple Properties on Swift Structs
+
+Properties in Swift structs have different calling conventions depending on whether the struct is frozen or not, and whether it's a getter or setter:
+
+- Getters:
+  - For frozen structs: Take the struct value lowered into registers
+  - For non-frozen structs: Take a pointer to the struct instance
+  
+- Setters:
+  - For both frozen and non-frozen structs: Always take a pointer to self and a value to set
+  - The pointer to self is required because setters need to modify the struct's memory in-place
+
+This means that when binding properties to C#, we need to handle these different cases appropriately:
+
+- For frozen struct getters, we can pass `SwiftSelf<T>`.
+- For non-frozen struct getters, we need to pass `SwiftSelf` pointer.
+- For all setters, we must always pass the new value and `SwiftSelf` pointer.
+
+Example of how this affects the binding:
+
+```swift
+@frozen public struct FrozenPoint {
+    public var x: Int
+}
+
+public struct NonFrozenPoint {
+    public var x: Int
+}
+```
+
+The generated C# binding needs to handle the different calling conventions:
+
+```csharp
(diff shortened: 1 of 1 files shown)


src/docs/binding-properties.md
```diff
@@ -0,0 +1,114 @@
...
+
+Properties in Swift structs have different calling conventions depending on whether the struct is frozen or not, and whether it's a getter or setter:
+
+- Getters:
+  - For frozen structs: Take the struct value lowered into registers
+  - For non-frozen structs: Take a pointer to the struct instance
+  
+- Setters:
+  - For both frozen and non-frozen structs: Always take a pointer to self and a value to set
+  - The pointer to self is required because setters need to modify the struct's memory in-place
```
kotlarmilos, line 13: Is this true for all mutating functions including setters?

src/docs/binding-properties.md
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private unsafe void Set_x(nint value)
+    {
+        IntPtr payload;
+        try 
+        {
+            var metadata = ...;
+            payload = (SwiftHandle)NativeMemory.Alloc(returnMetadata.Size);
+            SwiftMarshal.MarshalToSwift(instance, payload);
```
kotlarmilos: Is instance variable `this`?

src/docs/binding-properties.md
```diff
@@ -0,0 +1,114 @@
...
+        return PInvoke_get_x(self);
+    }
+
+    private unsafe void Set_x(nint value)
+    {
+        IntPtr payload;
+        try 
+        {
+            var metadata = ...;
+            payload = (SwiftHandle)NativeMemory.Alloc(returnMetadata.Size);
```
kotlarmilos: I know this this is a general approach to Swift marshalling, however It isn't intuitive to allocate a new memory and later have `this` assignment:
```
this = SwiftMarshal.MarshalFromSwift<FrozenPoint>(payload);
```

jkurdek: I think the key problem here is that we need to send the underlying memory of csharp struct into swift. And then Swift has to make some changes to that memory which later need to be visible on csharp side. Is this possible without making a copy?

src/docs/binding-properties.md
```diff
@@ -0,0 +1,114 @@
...
+
+    private unsafe nint Get_x()
+    {
+        var self = new SwiftSelf<FrozenPoint>(instance);
+        return PInvoke_get_x(self);
+    }
+
+    private unsafe void Set_x(nint value)
+    {
+        IntPtr payload;
```
jkotas:
```suggestion
        void* payload;
```
to avoid all casts to `void*`?

src/docs/binding-properties.md
```diff
@@ -0,0 +1,114 @@
...
+        try 
+        {
+            var metadata = ...;
+            payload = (SwiftHandle)NativeMemory.Alloc(returnMetadata.Size);
+            SwiftMarshal.MarshalToSwift(instance, payload);
+            var self = new SwiftSelf((void*)_payload);
```
vitek-karas: This should probably be:
```suggestion
            var self = new SwiftSelf((void*)payload);
```

src/docs/binding-properties.md
vitek-karas: @jkotas, @AaronRobinsonMSFT  do you know?

I could not find a way - using fixed doesn't work as it doesn't allow fixing "this". And using GCHandle won't work since it needs to box the value -> creates a copy of the struct.

The runtime would need to be able to do 2 different things:
- If the C# property setter is called on a struct which is on a stack (so local variable, parameter, ...) then it's just "fixed" (no pinning required as the struct can't move)
- If the C# property setter is called on a struct which is on the heap (field on a class for example) then it would have to pin the "parent" class (GCHandle and all)

jkotas: Try this:
```csharp
struct MyStruct
{
    public void M()
    {
        fixed (MyStruct* p = &this)
        {
           // do stuff with p
        }
    }
}
```

src/docs/binding-properties.md
jkurdek: Thats a great point! Yes, this will be true for all mutating functions. Setters would be just one of the use cases

src/docs/binding-properties.md
```diff
@@ -53,21 +53,21 @@ public struct FrozenPoint {
 
     private unsafe void Set_x(nint value)
     {
-        IntPtr payload;
+        void* payload;
         try 
         {
             var metadata = ...;
             payload = (SwiftHandle)NativeMemory.Alloc(returnMetadata.Size);
```
kotlarmilos: Is this still required? Sorry, not sure if you plan more changes :)

src/docs/binding-properties.md
jkurdek: I verified it works and replaced the relevant section

src/docs/binding-properties.md
jkurdek: Removed

kotlarmilos (approved):

Issue #2973, opened 2025-02-03
[Swift projection tooling] Introduce better diagnostics and error handling


jkurdek:
Design error / diagnostic system which gracefully handle the following scenarios:
- known unsupported features
- errors coming from either unknown unsupported features or logic bugs

Questions:
- Can we default to any "basic" behaviour in case of unknown scenario

Currently we use a temporary mixture of console logging and exceptions.

- [ ] Port logger from https://github.com/dotnet/runtimelab/pull/3034 to the rest of the tooling

PR #2989, opened 2025-02-10
[Swift language features] Bind property getters

### LINKED ISSUE #2986: [Swift language features] Bind property getters
To implement simple in-app purchase flow using StoreKit framework, we want to expose property accessors, such as:
 - [ ] https://developer.apple.com/documentation/storekit/product/displayname
 - [ ] https://developer.apple.com/documentation/storekit/product/description
 - [ ] https://developer.apple.com/documentation/storekit/product/displayprice
 - [ ] https://developer.apple.com/documentation/storekit/product/price

jkurdek:
Adds support for property getters. An appropriate property getter is generated on csharp side. All fields used for lowering of frozen structs are made private.

Fixes #2986

### DIFF
--- a/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
@@ -1,0 +1,86 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.CodeDom.Compiler;
+
+namespace BindingsGeneration;
+
+/// <summary>
+/// Factory class for creating instances of PropertyHandler.
+/// </summary>
+public class PropertyHandlerFactory : IFactory<BaseDecl, IPropertyHandler>
+{
+    public bool Handles(BaseDecl decl)
+    {
+        return decl is PropertyDecl propertyDecl && propertyDecl.Accessors.Any();
+    }
+
+    public IPropertyHandler Construct()
+    {
+        return new PropertyHandler();
+    }
+}
+
+/// <summary>
+/// Handler class for property declarations that generates the binding code for Swift properties.
+/// </summary> 
+public class PropertyHandler : BaseHandler, IPropertyHandler
+{
+    private readonly MethodHandler _methodHandler = new();
+
+    /// <inheritdoc/>
+    public IEnvironment Marshal(BaseDecl baseDecl, ITypeDatabase typeDatabase)
+    {
+        if (baseDecl is not PropertyDecl propertyDecl)
+        {
+            throw new ArgumentException("The provided decl must be a PropertyDecl.", nameof(baseDecl));
+        }
+        return new PropertyEnvironment(propertyDecl, typeDatabase);
+    }
+
+    /// <inheritdoc/>
+    public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment env, Conductor conductor)
+    {
+        var propertyEnv = (PropertyEnvironment)env;
+        var propertyDecl = propertyEnv.PropertyDecl;
+
+        var typeRecord = propertyEnv.TypeDatabase.GetTypeRecordOrThrow(propertyDecl.SwiftTypeSpec);
+
+        var staticModifier = propertyDecl.IsStatic ? "static " : "";
+        var csTypeName = typeRecord.CSTypeIdentifier;
+
+
+        // First emit the accessor methods using MethodHandler
+        foreach (var accessor in propertyDecl.Accessors)
+        {
+            var accessorEnv = new MethodEnvironment(accessor.Method, propertyEnv.TypeDatabase);
+            _methodHandler.Emit(csWriter, swiftWriter, accessorEnv, conductor);
+        }
+
+        // Then emit the property
+        csWriter.WriteLine($"public {staticModifier}{csTypeName} {propertyDecl.Name}");
+        csWriter.WriteLine("{");
+        csWriter.Indent++;
+
+        var getter = propertyDecl.Accessors.OfType<GetAccessorDecl>().FirstOrDefault();
+        if (getter != null)
+        {
+            EmitGetter(csWriter, getter);
+        }
+
+        csWriter.Indent--;
+        csWriter.WriteLine("}");
+        csWriter.WriteLine();
+    }
+
+    /// <summary>
+    /// Emits the getter implementation for a property.
+    /// </summary>
+    /// <param name="csWriter">The C# code writer to emit to</param>
+    /// <param name="getter">The getter accessor declaration</param>
+    private void EmitGetter(CSharpWriter csWriter, GetAccessorDecl getter)
+    {
+        csWriter.WriteLine($"get => {getter.Method.Name}();");
+    }
+}
+
--- a/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
@@ -28,6 +28,13 @@ public class PropertyHandler : BaseHandler, IPropertyHandler
 {
     private readonly MethodHandler _methodHandler = new();
 
+    // Dictionary of Swift property names that need to be renamed in C#
+    // Temporary workaround for https://github.com/dotnet/runtimelab/issues/2997 to keep StoreKit tests passing
+    private static readonly Dictionary<string, string> PropertyNameMappings = new()
+    {
+        { "isEligibleForIntroOffer", "isEligibleForIntroOfferProperty" },
+    };
+
     /// <inheritdoc/>
     public IEnvironment Marshal(BaseDecl baseDecl, ITypeDatabase typeDatabase)
     {
@@ -64,14 +71,14 @@ public class PropertyHandler : BaseHandler, IPropertyHandler
             return;
         }
 
-        if (propertyDecl.Accessors.Any(a => a.Method.IsAsync)) // TODO: https://github.com/dotnet/runtimelab/issues/2996
-        {
-            Console.WriteLine($"PropertyHandler: Async properties are not supported. Skipping property {propertyDecl.Name}.");
-            return;
-        }
-
+        // TODO Detect and skip / Handle async properties https://github.com/dotnet/runtimelab/issues/2996
         var csTypeName = typeRecord!.CSTypeIdentifier;
 
+        // Get the C# property name, using the mapping dictionary if necessary
+        // Temporary workaround for https://github.com/dotnet/runtimelab/issues/2997 to keep StoreKit tests passing
+        var propertyName = PropertyNameMappings.TryGetValue(propertyDecl.Name, out var mappedName)
+            ? mappedName
+            : propertyDecl.Name;
 
         // First emit the accessor methods using MethodHandler
         foreach (var accessor in propertyDecl.Accessors)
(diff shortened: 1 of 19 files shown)


src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+        var staticModifier = propertyDecl.IsStatic ? "static " : "";
+        var csTypeName = typeRecord.CSTypeIdentifier;
+
+
+        // First emit the accessor methods using MethodHandler
+        foreach (var accessor in propertyDecl.Accessors)
+        {
+            var accessorEnv = new MethodEnvironment(accessor.Method, propertyEnv.TypeDatabase);
+            _methodHandler.Emit(csWriter, swiftWriter, accessorEnv, conductor);
+        }
```
vitek-karas: I'm guessing this will generate public accessors on the C# class - would it make sense to make these private instead?

jkurdek: Fixed so that the generated pinvokes and wrappers are private and are available only through the property.

src/Swift.Bindings/src/Parser/ModuleProcessor.cs
```diff
@@ -156,9 +156,9 @@ private void ProcessStruct(NamedTypeSpec namedTypeSpec, StructDecl structDecl)
         /// </exception>
         private void ProcessStructFields(StructDecl structDecl)
```
matouskozak: Should the name change to `ProcessStructProperties` now?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -98,36 +97,9 @@ public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment en
...
-            if ((System.Runtime.InteropServices.Marshal.PtrToStringAnsi((IntPtr)fieldRecord->Name.Target) ?? string.Empty) != fieldDecl.Name)
-            {
-                return false;
-            }
-
-            return true;
-        }
     }
 
+
```
matouskozak:
```suggestion
```

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    /// <inheritdoc/>
+    public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment env, Conductor conductor)
+    {
+        var propertyEnv = (PropertyEnvironment)env;
+        var propertyDecl = propertyEnv.PropertyDecl;
+
+        var typeRecord = propertyEnv.TypeDatabase.GetTypeRecordOrThrow(propertyDecl.SwiftTypeSpec);
+
+        var staticModifier = propertyDecl.IsStatic ? "static " : "";
```
matouskozak: Do we support static properties? I think it is not part of the [proposal](https://github.com/dotnet/runtimelab/pull/2988) and I'm unsure what quirks there are. I think we should assert if we encounter something we don't support (or partially) yet.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+    public IEnvironment Marshal(BaseDecl baseDecl, ITypeDatabase typeDatabase)
+    {
+        if (baseDecl is not PropertyDecl propertyDecl)
+        {
+            throw new ArgumentException("The provided decl must be a PropertyDecl.", nameof(baseDecl));
+        }
+        return new PropertyEnvironment(propertyDecl, typeDatabase);
+    }
+
+    /// <inheritdoc/>
```
matouskozak, line 41: I think it would be nice to write in the comment an example what this will generate or link to the proposal document for future reference.

matouskozak (commented):
Good job Jeremi!

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+    public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment env, Conductor conductor)
```
matouskozak, line 45: I think it would be nice to write in the comment an example what this will generate or link to the proposal document for future reference.

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/Structs/StructsTests.swift
```diff
@@ -257,3 +257,36 @@ public func sumArray(array: Array<Int32>) -> Int32
...
+    public var varProperty: Int
+    private let multiplier: Int
+    
+    public init(letValue: Int, varValue: Int, multiplier: Int) {
+        self.letProperty = letValue
+        self.varProperty = varValue
+        self.multiplier = multiplier
+    }
+    
+    public var computedProperty: Int {
```
kotlarmilos: Thanks for including computed props!

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -437,27 +438,71 @@ private MethodDecl CreateMethodDecl(Node node, BaseDecl parentDecl, ModuleDecl m
...
+
+            foreach (var accessor in accessors)
+            {
+                switch (accessor.AccessorKind)
+                {
+                    case "get":
+                        result.Add(CreateGetAccessor(accessor, fieldName, parentDecl, moduleDecl));
+                        break;
+                    default:
+                        Console.WriteLine($"Unsupported accessor kind '{accessor.AccessorKind}' encountered.");
```
kotlarmilos, line 453: We should centralize this. The caller catches exceptions, and some paths throw but I am not sure if this is the best way to handle unknown scenarios. Let’s consider this during design sessions.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -74,19 +74,18 @@ public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment en
...
-                // TODO: Fix memory access violation
-                // // Verify field against Swift type information
-                // if (swiftTypeInfo.HasValue && !VerifyFieldRecord(swiftTypeInfo.Value, structDecl.Fields.IndexOf(fieldDecl), fieldDecl))
-                // {
-                //     Console.WriteLine("Field record does not match the field declaration");
-                // }
+                if (conductor.TryGetPropertyHandler(propertyDecl, out var propertyHandler))
+                {
+                    var fieldRecord = env.TypeDatabase.GetTypeRecordOrThrow(propertyDecl.SwiftTypeSpec);
+                    csWriter.WriteLine($"private {fieldRecord.CSTypeIdentifier} {propertyDecl.Name}_;");
```
vitek-karas: Can you add an explanation to why we need to emit the field to C#?
I guess it's because of the type layout for frozen structs, but it would be good to document that. Also such fields are never assigned and never accessed... which is another interesting quirck which should be documented - maybe even in the generated code (accessing the field directly will not do the right thing).

Aside - this means that I can't have a property without a backing field in swift?

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
jkurdek: I agree. Updated the https://github.com/dotnet/runtimelab/issues/2973

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
jkurdek: Good catch! There was a bug in the tests (wrong handling of private fields / properties without public accessors) which made me think there needs to be a backing fields for computed properties. 

Computed properties do not have a backing field

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+/// <summary>
+/// Handler class for property declarations that generates the binding code for Swift properties.
+/// </summary> 
+public class PropertyHandler : BaseHandler, IPropertyHandler
+{
+    private readonly MethodHandler _methodHandler = new();
+
+    // Dictionary of Swift property names that need to be renamed in C#
+    // Temporary workaround for https://github.com/dotnet/runtimelab/issues/2997 to keep StoreKit tests passing
+    private static readonly Dictionary<string, string> PropertyNameMappings = new()
```
jkurdek, line 29: This is an ugly workaround for https://github.com/dotnet/runtimelab/issues/2997
The problem was present before this PR - it was just surfaced by a computed property in StoreKit.

Fixing the name conflicts is a bigger thing - this is something to solve when we have proper c# translation.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -74,19 +74,28 @@ public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment en
...
             csWriter.Indent++;
 
-            // Emit each field in the struct
-            foreach (var fieldDecl in structDecl.Fields)
+            // For frozen structs, we need to emit fields that match the Swift struct's memory layout exactly.
+            // These backing fields are required for proper memory layout and marshalling, even though they
+            // are never directly accessed from C# code. The actual value access happens through Swift's
+            // accessor methods.
+            //
+            // Important: Direct access to these fields from C# will not provide the correct value - always
```
vitek-karas, line 83: I would write this into the generated code. One useful property of generating C# is that others can look at it/debug it. Even for projections, if you Go To Definition in VS/VS Code it will use source link on the NuGet and download the source code from github.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
jkurdek: Added a small comment. We can link proposal once the proposal is merged

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+        return new PropertyHandler();
+    }
+}
+
+/// <summary>
+/// Handler class for property declarations that generates the binding code for Swift properties.
+/// </summary> 
+public class PropertyHandler : BaseHandler, IPropertyHandler
+{
+    private readonly MethodHandler _methodHandler = new();
```
kotlarmilos: Should method handler be created in the factory and then selected based on Handles condition?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/PropertyHandler.cs
kotlarmilos: Consider using NameProvider for this where we could handle more cases until we refactor this.

kotlarmilos (approved):
LGTM!

PR #2993, opened 2025-02-12
[NativeAOT-LLVM] Merge feb25

yowl:
Merge from upstream b1ab309161278d3bb81efe2718399cf9f0dfad41 .

I think it is esasier to read the commits in reverse order because of a couple of mistakes I made along the way.

The commit history is one way to see the issues, but here is a summary

- Fix linux build by removing `extern` from `main` in test.
- Add back some Rational IR specific bits in `ssabuilder`
- Updates for removal of "init BB".  I didn't leave in the possible optimisation for excluding an empty first block from the unhandled exception handler as I hit a few issues.  For this EH, I set `bbRefs` to `1` for the filter and handler.
- Update Wasi SDK to 25
- Remove unsupported `--no-lock` from `brew`: presume this is also upstream in a later commit, but didn't check.
- Reinstate GS Stress as `RhpGcStressOnce` and `CORINFO_HELP_LLVM_STRESS_GC`
- Exclude `TestRuntime109496Regression` from the `Interfaces` test, we don't support `RhGetCommonStubAddress`.
- A little churn around `ReflectionExecution.ConvertStackTraceIpToFunctionPointer`
- Exclude `Preinitialization` test , looks like an uncovered bug in codegen for floats, but not sure.
- Don't set `TargetsMobile` for Wasm, instead add checks for `TargetsWasm`
- We now install `aotminipal`


yowl:
Thanks all on discord for the help!

yowl:
cc @dotnet/nativeaot-llvm

Directory.Build.props
```diff
@@ -35,7 +35,7 @@
     <HostOS>$(_hostOS)</HostOS>
     <TargetOS Condition="'$(TargetOS)' == '' and '$(RuntimeIdentifier)' == 'browser-wasm'">browser</TargetOS>
     <TargetOS Condition="'$(TargetOS)' == ''">$(_hostOS)</TargetOS>
-    <TargetsMobile Condition="'$(TargetOS)' == 'ios' or '$(TargetOS)' == 'iossimulator' or '$(TargetOS)' == 'maccatalyst' or '$(TargetOS)' == 'tvos' or '$(TargetOS)' == 'tvossimulator' or '$(TargetOS)' == 'android' or '$(TargetOS)' == 'browser' or '$(TargetOS)' == 'wasi'">true</TargetsMobile>
+    <TargetsMobile Condition="'$(TargetOS)' == 'ios' or '$(TargetOS)' == 'iossimulator' or '$(TargetOS)' == 'maccatalyst' or '$(TargetOS)' == 'tvos' or '$(TargetOS)' == 'tvossimulator' or '$(TargetOS)' == 'android'">true</TargetsMobile>
```
SingleAccretion: What was the problem with `TargetsMobile`? My preference would be to fix the place that it makes not work instead of turning it off again.

(Both turning it on and turning it on has merge problems, but having it on makes our changes additive, which are generally easier to reason about than "sideways" changes.)

eng/Subsets.props
```diff
@@ -88,19 +88,11 @@
...
 
-<<<<<<< HEAD
-  <PropertyGroup>
-    <RuntimeFlavor Condition="'$(TargetsMobile)' == 'true' and !$(_subset.Contains('+clr.nativeaotlibs+'))">Mono</RuntimeFlavor>
-    <RuntimeFlavor Condition="('$(TargetsMobile)' == 'true' or '$(TargetsLinuxBionic)' == 'true') and ($(_subset.Contains('+clr.')) or $(_subset.Contains('+nativeaot.')) or '$(TestNativeAot)' == 'true')">CoreCLR</RuntimeFlavor>
-    <RuntimeFlavor Condition="'$(RuntimeFlavor)' == '' and ($(_subset.Contains('+mono+')) or $(_subset.Contains('+mono.runtime+'))) and (!$(_subset.Contains('+clr+')) and !$(_subset.Contains('+clr.runtime+')))">Mono</RuntimeFlavor>
-=======
-
   <PropertyGroup Condition="'$(RuntimeFlavor)' == ''">
     <RuntimeFlavor Condition="('$(TargetsMobile)' == 'true' or '$(TargetsLinuxBionic)' == 'true') and ($(_subset.Contains('+clr.nativeaotlibs+')) or $(_subset.Contains('+clr.runtime+')))">CoreCLR</RuntimeFlavor>
```
SingleAccretion: (Related to `TargetsMobile` above, I suppose.)

I think we want to restore our changes to this condition, so that it looks like this:
```
Condition="('$(TargetsMobile)' == 'true' or '$(TargetsLinuxBionic)' == 'true') and ($(_subset.Contains('+clr.')) or $(_subset.Contains('+nativeaot.')) or '$(TestNativeAot)' == 'true')"
```

src/coreclr/build-runtime.cmd
```diff
@@ -44,12 +44,9 @@ set __TargetArchX64=0
 set __TargetArchX86=0
 set __TargetArchArm=0
 set __TargetArchArm64=0
-<<<<<<< HEAD
-set __TargetArchWasm=0
-=======
 set __TargetArchLoongArch64=0
 set __TargetArchRiscV64=0
->>>>>>> runtime/main
+set __TargetArchWasm=0
```
SingleAccretion: Just a note: I think when resolving this kind of conflict (with a long list of arches and we need to add 'wasm' at the end), it's beneficial to sneak in 'wasm' in the middle or even at the very beginning, so that this kind of conflict can be avoided.

src/coreclr/components.cmake
```diff
@@ -14,12 +14,7 @@ add_component(debug)
 # iltools and paltests should be minimal subsets, so don't add a dependency on coreclr_misc
 set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME coreclr_misc)
 add_component(coreclr_misc)
-<<<<<<< HEAD
-add_dependencies(jit coreclr_misc)
 add_dependencies(wasmjit coreclr_misc)
```
SingleAccretion: Maybe this (`add_dependencies(wasmjit coreclr_misc)`) is not necessary anymore?

src/coreclr/gc/CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
-if(FEATURE_STANDALONE_GC)
+if(FEATURE_STANDALONE_GC AND NOT CLR_CMAKE_TARGET_ARCH_WASM)
```
SingleAccretion: Do we need this? I see `FEATURE_STANDALONE_GC` is already set to `0` in `coreclr/CMakeLists.txt`.

src/coreclr/inc/jithelpers.h
```diff
@@ -124,22 +124,12 @@
...
-    DYNAMICJITHELPER(CORINFO_HELP_THROWNULLREF, NULL,               METHOD__THROWHELPERS__THROWNULLREFEXCEPTION)
-    JITHELPER(CORINFO_HELP_THROWMISALIGN,       NULL,               METHOD__NIL)
-    DYNAMICJITHELPER(CORINFO_HELP_VERIFICATION, NULL,               METHOD__THROWHELPERS__THROWVERIFICATIONEXCEPTION)
-=======
     DYNAMICJITHELPER(CORINFO_HELP_USER_BREAKPOINT,         NULL,    METHOD__DEBUGGER__USERBREAKPOINT)
     DYNAMICJITHELPER_NOINDIRECT(CORINFO_HELP_RNGCHKFAIL,   NULL,    METHOD__THROWHELPERS__THROWINDEXOUTOFRANGEEXCEPTION)
     DYNAMICJITHELPER_NOINDIRECT(CORINFO_HELP_OVERFLOW,     NULL,    METHOD__THROWHELPERS__THROWOVERFLOWEXCEPTION)
     DYNAMICJITHELPER_NOINDIRECT(CORINFO_HELP_THROWDIVZERO, NULL,    METHOD__THROWHELPERS__THROWDIVIDEBYZEROEXCEPTION)
     DYNAMICJITHELPER_NOINDIRECT(CORINFO_HELP_THROWNULLREF, NULL,    METHOD__THROWHELPERS__THROWNULLREFEXCEPTION)
     DYNAMICJITHELPER_NOINDIRECT(CORINFO_HELP_VERIFICATION, NULL,    METHOD__THROWHELPERS__THROWVERIFICATIONEXCEPTION)
```
SingleAccretion: `CORINFO_HELP_THROWMISALIGN` looks gone again. `CORINFO_HELP_LLVM_STRESS_GC` also needs to be added to the end of this file.

src/coreclr/jit/gentree.h
```diff
@@ -941,15 +941,19 @@ struct GenTree
                                                                              // and fix everyplace that reads undefined
                                                                              // values
         regNumber reg = (regNumber)_gtRegNum;
+#if !TARGET_WASM
         assert((gtRegTag == GT_REGTAG_NONE) || // TODO-Cleanup: get rid of the NONE case, and fix everyplace that reads
                                                // undefined values
                (reg >= REG_FIRST && reg <= REG_COUNT));
+#endif // !TARGET_WASM
```
SingleAccretion: Why is this needed?

src/coreclr/jit/llvmlower.cpp
```diff
@@ -201,9 +200,9 @@ void Llvm::initializeLlvmArgInfo()
 
     if (m_info->compRetBuffArg != BAD_VAR_NUM)
     {
-        // The return buffer is always pinned in our calling convetion, so that we can pass it as an LLVM argument.
+        // The return buffer is always pinned in our calling convention, so that we can pass it as an LLVM argument.
         LclVarDsc* retBufVarDsc = _compiler->lvaGetDesc(m_info->compRetBuffArg);
-        assert(retBufVarDsc->TypeGet() == TYP_BYREF);
+        assert(retBufVarDsc->TypeGet() == TYP_BYREF || retBufVarDsc->TypeGet() == TYP_I_IMPL);
         retBufVarDsc->lvType = TYP_I_IMPL;
         retBufVarDsc->lvCorInfoType = CORINFO_TYPE_PTR;
```
SingleAccretion: Our convention is now the same as upstream, so there is no need to retype. This should be just:
```
// The return buffer is always a stack pointer.
LclVarDsc* retBufVarDsc = _compiler->lvaGetDesc(m_info->compRetBuffArg);
assert(retBufVarDsc->TypeGet() == TYP_I_IMPL);
retBufVarDsc->lvCorInfoType = CORINFO_TYPE_PTR;
```

src/coreclr/jit/llvmlower.cpp
```diff
@@ -1991,13 +1979,14 @@ void Llvm::computeBlocksInFilters()
         {
             for (BasicBlock* block : _compiler->Blocks(ehDsc->ebdFilter, ehDsc->BBFilterLast()))
             {
-                if (m_blocksInFilters == BlockSetOps::UninitVal())
+                unsigned bbNumMax = _compiler->fgBBNumMax;
+                BitVecTraits bitVecTraits(_compiler->fgBBNumMax + 1, _compiler);
```
SingleAccretion: Hoist this outside all of the loops so that we don't construct the traits on each iteration?

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -92,7 +80,6 @@ void Compiler::fgResetForSsa(bool deepClean)
             }
         }
 
-<<<<<<< HEAD
 #if defined(TARGET_WASM)
         if (blk->IsLIR())
```
SingleAccretion: I think this should be moved under `deepClean` for consistency.

src/coreclr/jit/register.h
```diff
@@ -352,6 +352,22 @@ REGDEF(STK_CANDIDATE_UNCONDITIONAL, 3, 0x08, "SS_UNCONDITIONAL")
...
+#undef REG_LLVM
+#define REG_LLVM JITREG_LLVM
+#undef REG_STK_CANDIDATE_UNCONDITIONAL
+#define REG_STK_CANDIDATE_UNCONDITIONAL JITREG_STK_CANDIDATE_UNCONDITIONAL
+#undef REG_STK_CANDIDATE_TENTATIVE
+#define REG_STK_CANDIDATE_TENTATIVE JITREG_STK_CANDIDATE_TENTATIVE
+#undef REG_STK_CANDIDATE_COMMITED
+#define REG_STK_CANDIDATE_COMMITED JITREG_STK_CANDIDATE_COMMITED
+#undef REG_STK
+#define REG_STK JITREG_STK
```
SingleAccretion: Instead of this, we should move the WASM reg definitions to be under the `LA64 and RV64 don't require JITREG_ workaround for Android (see register.h)` branch in `target.h`.

src/coreclr/nativeaot/BuildIntegration/BuildIntegration.proj
```diff
@@ -42,10 +42,7 @@
     </ItemGroup>
     <Copy SourceFiles="@(AnalyzerFile)" DestinationFolder="$(RuntimeBinDir)\analyzers\dotnet\cs" />
   </Target>
-<<<<<<< HEAD
-=======
 
   <Target Name="Restore" />
```
SingleAccretion: We have a `ProjectReference` in this project (unlike upstream). No-oping restore seems like it will cause issues.

src/tasks/tasks.proj
```diff
@@ -24,6 +24,12 @@
     <ProjectReference Include="installer.tasks\installer.tasks.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetsMobile)' != 'true' and '$(TargetsWasm)' == 'true'">
```
SingleAccretion: Relevant to the `TargetsMobile` above.

src/native/libs/build-native.proj
```diff
@@ -30,7 +30,7 @@
   </ItemGroup>
 
   <PropertyGroup Condition="'$(TargetOS)' == 'wasi'">
-    <WASI_SDK_PATH Condition="'$(WASI_SDK_PATH)' == '' or !Exists('$(WASI_SDK_PATH)/WASI-SDK-VERSION-25.0')">$([MSBuild]::NormalizeDirectory($(RepoRoot), 'src', 'mono', 'wasi', 'wasi-sdk'))</WASI_SDK_PATH>
+    <WASI_SDK_PATH Condition="'$(WASI_SDK_PATH)' == '' or !Exists('$(WASI_SDK_PATH)/VERSION25')">$([MSBuild]::NormalizeDirectory($(RepoRoot), 'src', 'mono', 'wasi', 'wasi-sdk'))</WASI_SDK_PATH>
```
SingleAccretion: Why the change? It's best to include this marker file in one's local WASI SDK.

src/tests/nativeaot/SmokeTests/Preinitialization/Preinitialization.csproj
```diff
@@ -3,7 +3,8 @@
     <OutputType>Exe</OutputType>
     <CLRTestPriority>0</CLRTestPriority>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <CLRTestTargetUnsupported Condition="'$(IlcMultiModule)' == 'true'">true</CLRTestTargetUnsupported>
+    <!-- TODO-LLVM: Fails on Wasm: Assert.AreEqual(unchecked((byte)GetFloat()), FloatConversions.s_byte); -->
+    <CLRTestTargetUnsupported Condition="'$(IlcMultiModule)' == 'true' or '$(TargetsWasm)' == 'true'">true</CLRTestTargetUnsupported>
```
SingleAccretion: Could you exclude it in `dirs.proj` instead and open an issue about this?

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
```diff
@@ -32,6 +32,7 @@
     <EmccExportedFunction Include="stackAlloc" />
     <EmccExportedFunction Include="stackRestore" />
     <EmccExportedFunction Include="stackSave" />
+    <EmccExportedFunction Include="_minipal_get_non_cryptographically_secure_random_bytes" />
```
SingleAccretion: Do you know what uses it / why it needs to be exported?

SingleAccretion (commented):
Mostly minor comments; some build questions about `TargetsMobile`.

Directory.Build.props
yowl: First problem, perhaps the only one is 

https://github.com/dotnet/runtime/blob/454673e1d6da4067750646523ee67e066a388cd7/eng/Subsets.props#L30

where CoreCLR is not enabled.

eng/Subsets.props
yowl: I see, missed that.  Looks like if we out this back the first issue is not an issue.

src/coreclr/build-runtime.cmd
yowl: good point, will move this one to the beginning.

src/native/libs/build-native.proj
yowl: I couldn't see how this used to work, but this sentinel name is the same as
https://github.com/dotnet/runtimelab/blob/9d025fa5e9d07e3e93061b4457ba6279b178df98/eng/pipelines/runtimelab/install-wasi-sdk.ps1#L36

SingleAccretion: This looks to be the fallout from https://github.com/dotnet/runtime/pull/110654, so we just need to update `install-wasi-sdk.ps1` to use the new file name scheme.

pavelsavara: This is just sanity check, at some point our CI had too many versions on WASI-SDK and it was difficult to understand why it's wrong. 
The check implemented as simple file name is easier than reading the content of `VERSION` file. Because there are multiple places.

The better alternative is
https://github.com/dotnet/runtime/blob/37e4d45236e68946db9d264593aa31a9c00534bc/src/mono/wasi/build/WasiApp.targets#L103-L108

but we don't do that for in-tree builds.

Directory.Build.props
yowl: I've reverted these changes now. I'd forgotten we had chosen `TargetsMobile` as the wanted direction,

eng/Subsets.props
yowl: Restored.

src/coreclr/components.cmake
yowl: Indeed, removed.

src/coreclr/gc/CMakeLists.txt
yowl: No, removed.

src/coreclr/inc/jithelpers.h
yowl: Reinstated, thanks.

src/coreclr/jit/gentree.h
yowl: Removed, thanks.

src/coreclr/jit/llvmlower.cpp
yowl: Thanks, changed to the suggested.

src/coreclr/jit/llvmlower.cpp
yowl: Missed that, thanks.

src/coreclr/jit/ssabuilder.cpp
yowl: Done.

src/coreclr/jit/register.h
yowl: Right, moved.

src/coreclr/nativeaot/BuildIntegration/BuildIntegration.proj
yowl: Added a `Condition` for the no op., thanks,

src/tasks/tasks.proj
yowl: Reverted.

src/native/libs/build-native.proj
yowl: Thanks, updated `install-wasi-sdk.ps1` as suggested.

src/tests/nativeaot/SmokeTests/Preinitialization/Preinitialization.csproj
yowl: Yes, issue created.  https://github.com/dotnet/runtimelab/issues/3074

src/coreclr/nativeaot/BuildIntegration/DotNetJsApi.targets
yowl: Hmm, funny, its not needed so removed it.

eng/testing/tests.props
```diff
@@ -11,7 +11,7 @@
     <TestSingleFile Condition="'$(TestReadyToRun)' == 'true'">true</TestSingleFile>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(TargetsMobile)' == 'true'">
+  <PropertyGroup Condition="'$(TargetsMobile)' == 'true' or '$(TargetsWasm)' == 'true'">
```
SingleAccretion:
```suggestion
  <PropertyGroup Condition="'$(TargetsMobile)' == 'true'">
```

src/coreclr/jit/ssabuilder.cpp
```diff
@@ -121,6 +93,19 @@ void Compiler::fgResetForSsa(bool deepClean)
...
+        {
+            for (GenTree* tree : LIR::AsRange(blk))
+            {
+                if (tree->IsLocal() || tree->OperIs(GT_LCL_ADDR))
+                {
+                    tree->AsLclVarCommon()->SetSsaNum(SsaConfig::RESERVED_SSA_NUM);
+                }
+            }
+        }
+#endif
```
SingleAccretion: Should be moved under `if (deepClean)`.

src/coreclr/jit/regset.h
```diff
@@ -74,8 +74,10 @@ class RegSet
     bool rsModifiedRegsMaskInitialized; // Has rsModifiedRegsMask been initialized? Guards against illegal use.
 #endif                                  // DEBUG
 
+#if !defined(TARGET_WASM)
     regMaskTP rsAllCalleeSavedMask = RBM_CALLEE_SAVED;
     regMaskTP rsIntCalleeSavedMask = RBM_INT_CALLEE_SAVED;
+#endif
```
SingleAccretion: It would be better to add these to `targetwasm.h` with some dummy value (e. g. zero) than add ifdefs here (applies to another ifdef below as well).

src/coreclr/jit/gentree.h
```diff
@@ -949,7 +949,9 @@ struct GenTree
 
     void SetRegNum(regNumber reg)
     {
+#if !TARGET_WASM
         assert(reg >= REG_FIRST && reg <= REG_COUNT);
+#endif // !TARGET_WASM
```
SingleAccretion: Needed?

src/coreclr/nativeaot/BuildIntegration/BuildIntegration.proj
```diff
@@ -42,10 +42,7 @@
     </ItemGroup>
     <Copy SourceFiles="@(AnalyzerFile)" DestinationFolder="$(RuntimeBinDir)\analyzers\dotnet\cs" />
   </Target>
-<<<<<<< HEAD
-=======
 
-  <Target Name="Restore" />
+  <Target Name="Restore" Condition="'$(TargetsWasm)' != 'true'"/>
```
SingleAccretion: Does this work correctly? I though MSBuild would override this target anyway, it just wouldn't 'run' it under the condition (which for this no-op target would not be meaningful).

I think we should just remove this in our fork; it's needed upstream because the upstream project doesn't use the traversal SDK.

src/tasks/tasks.proj
```diff
@@ -24,6 +24,12 @@
     <ProjectReference Include="installer.tasks\installer.tasks.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetsMobile)' != 'true'">
+    <ProjectReference Include="WasmAppBuilder\WasmAppBuilder.csproj" />
+    <ProjectReference Include="WasmBuildTasks\WasmBuildTasks.csproj" />
+    <ProjectReference Include="MonoTargetsTasks\MonoTargetsTasks.csproj" />
+  </ItemGroup>
+
```
SingleAccretion:
```suggestion
```
`TargetsMobile` follow-up.

src/coreclr/components.cmake
SingleAccretion: > Indeed, removed.

Hmm, seems the change hasn't been commited?

src/coreclr/nativeaot/BuildIntegration/BuildIntegration.proj
yowl: Hmm, not sure to be honest.  Have removed it.

SingleAccretion (approved):
LGTM, thank you!

yowl:
Thanks all!

Issue #2850, opened 2024-12-03
[Swift shipping] Implement testing framework


kotlarmilos:
## Infrastructure
   - [x] https://github.com/dotnet/runtimelab/pull/2871
   - [x] https://github.com/dotnet/runtimelab/pull/2880
   - [x] https://github.com/dotnet/runtimelab/pull/2998
   - [x] https://github.com/dotnet/runtimelab/issues/2849
   - [x] https://github.com/dotnet/runtimelab/issues/3006

PR #2998, opened 2025-02-12
[Swift shipping] Setup MacCatalyst tests on the CI

kotlarmilos:
## Description
This PR introduces MacCatalyst tests using MAUI test runners and the xUnit framework. The goal is to update our infrastructure to support writing xUnit tests within MAUI projects. This will allow us to reference a NuGet package with bindings and run framework tests, such as in-app purchase.

## Changes
 - Added a MAUI project for running tests using XHarness
 - Added a new `test-job.yml` for device tests
 - Moved the HikingApp to the samples directory

Contributes to https://github.com/dotnet/runtimelab/issues/2850


eng/pipelines/common/templates/build-job.yml
```diff
@@ -52,14 +52,14 @@ jobs:
...
                 /p:TargetPlatform=${{ parameters.archType }}
-        displayName: Build and Test
+        displayName: Build and test
 
-      - script: ./verify.sh
+      - script: ./dotnet.sh format --exclude artifacts --verify-no-changes
         displayName: Verify code formatting
 
       - ${{ if and(eq(parameters.bindings.isBindingsBuild, true), ne(parameters.osGroup, 'Windows_NT')) }}:
         - script: ./generate.sh --configuration $(_BuildConfig) ${{ parameters.bindings.scriptArgs }}
```
jkurdek, line 61: We might want to pass a verbosity flag here as well.

eng/pipelines/runtimelab.yml
```diff
@@ -71,3 +71,11 @@ stages:
         bindings:
           isBindingsBuild: true
           scriptArgs: '--platform MacOSX --arch x86_64-apple-macos --framework StoreKit'
+
+    - template: /eng/pipelines/common/templates/test-job.yml
+      parameters:
+        osGroup: OSX
+        platform: maccatalyst
+        archType: x64
```
jkurdek, line 68: Why x64? Is this a HW availability limitation?

src/Swift.Bindings/tests/IntegrationTests/Swift.Bindings.Integration.Tests.csproj
```diff
@@ -23,7 +23,7 @@
 
     <!-- Compile generated bindings with tests -->
     <ItemGroup>
-      <Compile Include="$(OutputPath)/**/*.cs" Exclude="$(OutputPath)/Metadata/*.cs;$(OutputPath)/ManualBindings/*.cs" />
```
jkurdek, line 26: Why do we need that change?

rolfbjarne (approved):

global.json
```diff
@@ -1,18 +1,18 @@
...
+      "dotnet": "10.0.100-alpha.1.24573.1",
+      "runtimes": {
+        "dotnet": [
+          "$(MicrosoftNETCoreAppVersion)"
+        ]
+      }
+    },
+    "msbuild-sdks": {
+      "Microsoft.DotNet.Arcade.Sdk": "10.0.0-beta.24606.6"
     }
```
vitek-karas: Is this change intentional? It looks like indentation change only (and wrong indentation it seems)

eng/pipelines/runtimelab.yml
kotlarmilos: Yes, this is the default image. There is a macOS-arm64 image and we need to figure out how to use it.

https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml

src/Swift.Bindings/tests/IntegrationTests/Swift.Bindings.Integration.Tests.csproj
kotlarmilos: This cleanup is unrelated to the current PR; I can move it to a separate PR if confusing.

These are old directories used for handling redundant files in the testing infrastructure. The Swift.Runtime files are referenced in the project and emitted by the projection tooling. Conflicts were resolved by removing redundant Swift.Runtime files from the cmake, so this is not needed anymore.

jkurdek: ok, thanks for the explanation! It is fine for me to keep it with this PR.

global.json
kotlarmilos: No, I messed up ident.

eng/pipelines/runtimelab.yml
kotlarmilos: ~The `hostArchitecture: Arm64` was missing.~

global.json
kotlarmilos: It is still showing up in the diff

eng/pipelines/runtimelab.yml
kotlarmilos: @akoeplinger Do you know if we can run macOS-arm64 containers?

akoeplinger: not if you run tests directly on the AzDO build machine, the arm64 macOS pool is only available in internal builds, not in public.

jkurdek (approved):

eng/pipelines/runtimelab.yml
kotlarmilos: Is the arm64 job redundant? Should we use Helix instead for arm64 jobs?

akoeplinger: which arm64 job? the one in this file only _builds_ for arm64 which does work on x64.

if you can send to Helix then yeah you could send to an arm64 queue

kotlarmilos: > which arm64 job? the one in this file only builds for arm64 which does work on x64.

Yes, that was my question. So even it builds for arm64, it runs on x64.

akoeplinger: Yes

Issue #2982, opened 2025-02-06
[Swift language features] Add special handling of `Swift.Array` and `Swift.Set` bound generics


kotlarmilos:
## Description

- [x] Add special handling of Swift.Array and Swift.Set bound generics to the method handler
- [x] Add functional tests

Issue #2889, opened 2024-12-17
[Projection tooling] Unify csharp naming


jkurdek:
There are many places where we will need to name csharp entities.

- Naming things projected from Swift
- Generating names for extra things we need - pinvokes, proxies etc.
- Naming local variables using for marshalling

We should also be aware that some things coming from swift are unpronounceable in csharp and require translation:
#2923 

We should implement a unit which will be able to provide us with all the aforementioned names. We also need a common model to represent a csharp name (see comments) as well as model to represent swift name. Those should be uniform across the repo.

When projecting types we will use the information encoded in those models to correctly handle namespaces.

jkurdek:
Here's something to put a pin in: Swift is simple in that there is only one namespace: the module name. C# not so much. In either case, it might be handy to have something like this:
```
public class ClassName { // implement IEquatable<> etc.
   public ClassName(string namespace, string name)
   {
        Namespace = namespace;
        Name = name;
        FullName = Namespace + "." + Name;
   }
   public string Namespace { get; init; }
   public string Name { get; init; } // may include inner classes as a path
   public string FullName { get; init }
}
```
And use this wherever you want to refer to either a C# class or a Swift class.
When writing code non-linearly, we can do `namespaces.AddIfNotPresent(clname.Namespace)` or `imports.AddIfNotPresent(clname.Namespace)`.

_Originally posted by @stephen-hawley in https://github.com/dotnet/runtimelab/pull/2870#discussion_r1888646514_

jkurdek:
I'd like to see a separate type for holding a type name.  Something like:
```
public class /* or record or struct */ TypeName {
     public string Namespace { get; init; }
     public string Name { get; init; }
     public string FullName { get; } => Namespace + "." + Name;
     // if not a record, make ToString, Equals, etc
}
```
This will make it easier to set up `import` statements in Swift and `using` statements in C#.

_Originally posted by @stephen-hawley in https://github.com/dotnet/runtimelab/pull/2912#discussion_r1907835643_

PR #3003, opened 2025-02-14
[Swift Bindings] Initialize bound generics handling

jkurdek:
The PR initializes handling of bound generic types. It enables passing of Swift.Set and Swift.Array. Also, this makes handling of bound generics using TypeDatabase unnecessary - I removed the related logic.

Fixes https://github.com/dotnet/runtimelab/issues/2982
Fixes https://github.com/dotnet/runtimelab/issues/3007


src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+{
+    private readonly ITypeDatabase _typeDatabase;
+
+    public BoundGenericsHandler(ITypeDatabase typeDatabase)
+    {
+        _typeDatabase = typeDatabase;
+    }
+
+    // Almost all generics will be projected into C# as classes.
+    // This collection contains types that must be marshalled as structs.
```
kotlarmilos, line 18: Is this determined based on the type layout?

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+    // We might introduce a new field in the TypeRecord to indicate this
+    private static readonly HashSet<SwiftTypeName> s_structGenerics = new()
+        {
+            SwiftTypeName.FromModuleQualifiedName("Swift.UnsafeMutableBufferPointer"),
+            SwiftTypeName.FromModuleQualifiedName("Swift.UnsafeMutablePointer"),
+        };
+
+    // TODO: Add more types as needed.
+    // Mapping of Swift generic types to their corresponding buffer types.
+    private static readonly Dictionary<SwiftTypeName, string> s_bufferTypeMap = new()
```
kotlarmilos, line 28: If we don't want to maintain this list, we could add inner buffer types that follow the same naming convention (e.g., Swift.Array.Buffer, Swift.Set.Buffer).

src/Swift.Runtime/src/Swift/FoundationDatabase.xml
```diff
@@ -64,14 +64,14 @@
...
-            <typedeclaration kind="struct" name="UnsafeMutablePointer&lt;Swift.UInt8&gt;" module="Swift" mangledName="sSp" frozen="true" blittable="true"/>
-        </entity>
-        <entity managedNameSpace="Swift" managedTypeName="UnsafeMutableBufferPointer&lt;System.Byte&gt;">
-            <typedeclaration kind="struct" name="UnsafeMutableBufferPointer&lt;Swift.UInt8&gt;" module="Swift" mangledName="sSr" frozen="true" blittable="true"/>
-        </entity>
         <entity managedNameSpace="Swift" managedTypeName="SwiftString">
             <typedeclaration kind="struct" name="String" module="Swift" mangledName="sSS" frozen="true" blittable="true"/>
         </entity>
+        <entity managedNameSpace="Swift" managedTypeName="SwiftArray">
+            <typedeclaration kind="struct" name="Array" module="Swift" mangledName="sSS" frozen="true" blittable="true"/>
```
kotlarmilos:
```suggestion
            <typedeclaration kind="struct" name="Array" module="Swift" mangledName="sSa" frozen="true" blittable="true"/>
```

src/Swift.Runtime/src/Swift/FoundationDatabase.xml
```diff
@@ -64,14 +64,14 @@
...
+        </entity>
+        <entity managedNameSpace="Swift" managedTypeName="SwiftSet">
+            <typedeclaration kind="struct" name="Set" module="Swift" mangledName="sSS" frozen="true" blittable="true"/>
```
kotlarmilos:
```suggestion
            <typedeclaration kind="struct" name="Set" module="Swift" mangledName="sSh" frozen="true" blittable="true"/>
```

kotlarmilos:
Could you please check why `products` and `purchase` are not generated?

https://dev.azure.com/dnceng-public/public/_build/results?buildId=952586&view=logs&jobId=c799f0b0-a2b0-5c26-b3cc-8b8fbac55a95&j=c799f0b0-a2b0-5c26-b3cc-8b8fbac55a95&t=3b4a96ca-1086-567a-e53c-0c8bf5d6c4fd

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
jkurdek: In a sense yes. Those should be all the types whose layout is independent of `T` and who do not contain any references. So e.g. `Swift.Array` has layout which is independent of `T` but it contains references so it has to be class.

On the other hand we need to be careful when designing manual projections so that they match the above rule.

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
jkurdek: I think this might be the next step (once we generalize the solution to support other bound generics)

jkurdek:
@kotlarmilos the `products` and `purchase` are not generated because the demangler throws errors. 
https://github.com/dotnet/runtimelab/issues/3007

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -159,7 +159,7 @@ public record Parameter(string Type, string Name, string modifier = "")
     /// <param name="Parameters"></param>
     public record Signature(string ReturnType, IReadOnlyList<Parameter> Parameters)
     {
-        public bool ContainsPlaceholder => Parameters.Any(p => p.Type == "AnyType") || ReturnType == "AnyType";
+        public bool ContainsPlaceholder => Parameters.Any(p => p.Type.Contains("AnyType") || ReturnType.Contains("AnyType"));
         public string ParametersString() => string.Join(", ", Parameters.Select(p => p.SignatureString()));
```
vitek-karas: This check would return true even if the return type (for example) is `MyAnyType` - which would be wrong. If the type names are qualified, then we should use the namespace qualified name for comparison. If this is about generic instantiation, then we need to have a different mechanism do detect generics with AnyType in them.

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+            { SwiftTypeName.FromModuleQualifiedName("Swift.Set"), "Swift.Variant" }
+        };
+
+    /// <summary>
+    /// Determines whether the specified property declaration represents a bound generic type.
+    /// </summary>
+    /// <param name="propertyDecl">The property declaration.</param>
+    /// <returns><c>true</c> if the property’s Swift type contains generic parameters; otherwise, <c>false</c>.</returns>
+    public bool IsBoundGeneric(PropertyDecl propertyDecl) =>
+        propertyDecl.SwiftTypeSpec is NamedTypeSpec namedTypeSpec && namedTypeSpec.ContainsGenericParameters; // TODO: Check whether return type is not type's generic parameter
```
vitek-karas: We should start getting into habit of not adding TODOs. If there's a piece of implementation we want to postpone - create an issue and link it here. TODO alone will never get fixed :-)

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+        return !s_structGenerics.Contains(swiftTypeName);
+    }
+
+    /// <summary>
+    /// Translates the Swift generic type of the given property declaration into a C# type name.
+    /// </summary>
+    /// <param name="propertyDecl">The property declaration.</param>
+    /// <returns>The C# type name with generic parameters.</returns>
+    /// <exception cref="NotSupportedException">Thrown when the property is not bound generic.</exception>
+    public string TranslateToCSharpName(PropertyDecl propertyDecl)
```
vitek-karas: Naming nit - this sounds like it will produce the name of the property, not the name of the property's type. Maybe `TranslateToCSharpTypeName`...

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+        List<string> translatedGenericParameters = new();
+        foreach (var genericParameter in namedTypeSpec.GenericParameters)
+        {
+            if (genericParameter is not NamedTypeSpec namedGenericParameter)
+                throw new NotSupportedException(
+                    $"Generic parameter {genericParameter} is not a named type spec");
+            translatedGenericParameters.Add(TranslateToCSharpName(namedGenericParameter));
+        }
+
+        var typeReference = _typeDatabase.GetTypeRecordOrAnyType(namedTypeSpec); // TODO: consider throwing an exception instead
```
vitek-karas, line 116: Throwing an exception is always the safer, more maintainable way - so I would prefer we throw now... and relax it later if we need to.

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+        return typeReference.CSTypeIdentifier +
+               (translatedGenericParameters.Count > 0
+                    ? $"<{string.Join(", ", translatedGenericParameters)}>"
+                    : "");
```
vitek-karas, line 120: Curious - when would this happen? It's a bound generic type but it has no generic parameters? Is that the case `T` alone (so type parameter of the enclosing type?)

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /// <summary>
+    /// Gets the buffer type name used for marshalling the specified bound generic argument.
```
vitek-karas, line 124: Could you please explain what the "buffer type" really means here - if it's the type which is used for lowering or something else?
Currently the method will succeed only for 2 types - that feels weird. For hardcoded types it's OK, but what about user types...

jkurdek: So if a generic projection is a class we will need some "buffer" struct for the purpose of doing the lowering. There are multiple options of generating those buffers / helpers - if we add a runtime support for lowering generics the buffer could be a generic struct, if not we will need to generate a struct per generic instantiation e.g. `Foo_Int_Int>` for `Foo<int,int>`. The idea is described in binding-generics.md

It supports only two types as those are required for milestone one and do not require the things mentioned above. Both the array and the set have layout independent of the type parameter which can be represented as a non-generic struct for lowering purposes.

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
jkurdek: This brings us back to he conversation about transactional code emission. This step happens during code emission. If we threw here we would break emission of the whole module.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
jkurdek: This is both about generic instatiation and non generic type handling. A consequence of us not yet implementing the transactional emission model. The current solution which we have when we are not able to reason about something is to place this `AnyType` type on the signature. If the signature contains this placeholder we know that we need to skip this method.

We could keep AnyType for now, make sure it is referenced with fully qualified name. For bound generics emit AnyType if any of the generic params is unknown. Moving forward - when implementing the transactional model - we should review whether we even need `AnyType`

cc @kotlarmilos

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
vitek-karas: OK - thanks for the explanation.

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
vitek-karas: Thanks for the explanation

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: It sounds like we're ready to iterate on the design :)

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
kotlarmilos: @jkurdek Could you please write down "transactional code emission" idea internally as I've seen it around?

kotlarmilos:
> @kotlarmilos the `products` and `purchase` are not generated because the demangler throws errors.
> #3007

--- a/src/Swift.Bindings/src/Demangler/Swift5Demangler.cs
+++ b/src/Swift.Bindings/src/Demangler/Swift5Demangler.cs
@@ -1223,12 +1223,14 @@ namespace BindingsGeneration.Demangling
```
             if (funcType.Kind != NodeKind.FunctionType && funcType.Kind != NodeKind.NoEscapeFunctionType)
                 return null;
 
-            var parameterType = funcType.Children[0];
-            if (parameterType.Kind == NodeKind.ThrowsAnnotation)
-                parameterType = funcType.Children[1];
+            int parameterIndex = 0;
+            if (funcType.Children[parameterIndex].Kind == NodeKind.ThrowsAnnotation)
+                parameterIndex++;
 
-            if (parameterType.Kind == NodeKind.AsyncAnnotation)
-                parameterType = funcType.Children[1];
+            if (funcType.Children[parameterIndex].Kind == NodeKind.AsyncAnnotation)
+                parameterIndex++;
+
+            var parameterType = funcType.Children[parameterIndex];
 
             if (parameterType.Index == 0)
                 return null;
```

Let me know if you want to include it here or to create a separate PR.

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
jkurdek: This will happen at last level of recursion So If I have e.g. `Foo<Bar<Int>>` this will be true for Int

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
jkurdek: Modified the logic to check the fully qualified named instead. This should avoid problems outlined by @vitek-karas. Lets discuss the future of `AnyType` separately.

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
jkurdek: Right, https://github.com/dotnet/runtimelab/issues/3013. Added to the code.

jkurdek:
@kotlarmilos included the demangler patch. `products` and `purchase` should be correctly generated

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
vitek-karas: In that case I will add another comment which I was thinking originally about: Currently this implementes Swift->C# name translation only for generic entities. Would it make sense to make a more general version which can work on non-generic entities as well?
The "bottom of the recursion" level in generic case is the same as the non-generic case anyway.

vitek-karas (approved):
Looks good - thanks!

src/Swift.Bindings/src/Marshaler/BoundGenericsHandler.cs
jkurdek: >Would it make sense to make a more general version which can work on non-generic entities as well?

Absolutely, we have it on the backlog already. https://github.com/dotnet/runtimelab/issues/2889. Implementing that will affect a lot of code so I would like to bring it as a separate change. 

Renamed the function so it should be clear what it does

kotlarmilos (approved):

PR #3004, opened 2025-02-15
[Swift Bindings] Propagate cmake warnings and fix existing warnings

vitek-karas:
Show cmake warnings when building swift files.
Also fixes cmake warnings which were present.

### DIFF
--- a/src/Swift.Bindings/tests/IntegrationTests/Swift.Bindings.Integration.Tests.csproj
+++ b/src/Swift.Bindings/tests/IntegrationTests/Swift.Bindings.Integration.Tests.csproj
@@ -6,8 +6,9 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
-    <PreTestCommand>cd $(OutputPath)$(TargetFramework) &amp;&amp; cmake $(MSBuildThisFileDirectory) &amp;&amp; make</PreTestCommand>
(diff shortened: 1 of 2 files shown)


src/Swift.Bindings/tests/IntegrationTests/Swift.Bindings.Integration.Tests.csproj
```diff
@@ -21,6 +22,17 @@
     <!-- Generate Swift libraries and projection tooling bindings -->
     <Exec Command="$(PreTestCommand)" />
```
kotlarmilos, line 23: Alternatively, you can use `ConsoleToMsBuild` which captures the output as an item: https://learn.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2022
```
<Exec Command="$(PreTestCommand)"
      ConsoleToMSBuild="true">
  <Output TaskParameter="ConsoleOutput" ItemName="CMakeOutputLines"/>
</Exec>
```

kotlarmilos (approved):
Thanks!

src/Swift.Bindings/tests/IntegrationTests/Swift.Bindings.Integration.Tests.csproj
vitek-karas: Actually having the output in a log file helps, since the cmake warnings are multi-line and thus not easily "parsable" via regexes and such. So the detection I added is to just notice a warning - then go to the log and figure out what the problem realy is.
The full log does show up in binlog as well if we need to get it from there.

PR #3008, opened 2025-02-18
[Swift ABI compatibility] Document memory ownership and handling at the Swift–C# boundary

kotlarmilos:
## Description
This PR adds documentation outlining the native memory management approach for projected value types. The scenarios included will be extended in future.

The goal is to validate the proposed approach and provide documentation.

## Out-of-scope

This PR doesn't include the handling of projected reference types; it only includes projected value types with reference properties. Additionally, generics are not included.

### DIFF
--- a/src/docs/memory-management.md
+++ b/src/docs/memory-management.md
@@ -41,10 +41,17 @@ To define memory handling rules at the Swift and C# boundary, the following scen
 ### Code blocks
 
 When a value type is initialized within a block, it is stack-allocated. At the end of the block, all reference properties must be deallocated. For example:
+```swift
+public func TestBlock()
+{
+    var vtype = VType()
+}
+```
 ```
 %vtype = alloca %T6output5VTypeV, align 8
 ...
-%0 = call ptr @"outlined destroy of output.VType"(ptr %vtype), !dbg !156
+call swiftcc void @"output.VType.init() -> output.VType"(ptr noalias nocapture sret(%T6output5VTypeV) %vtype), !dbg !248
+%0 = call ptr @"outlined destroy of output.VType"(ptr %vtype), !dbg !250
 
 // VWT->Destroy
 define linkonce_odr hidden ptr @"outlined destroy of output.VType"(ptr %0) #8 !dbg !143 {
@@ -64,7 +71,6 @@ When a reference type or a value type containing a reference property is passed 
 var refType = RefType()
 var vtype = VType(refType: refType)
 ```
-
 ```
 %3 = call ptr @swift_retain(ptr returned %2) #4, !dbg !164
 store ptr %2, ptr %refType, align 8, !dbg !164
@@ -77,24 +83,40 @@ call void @swift_release(ptr %toDestroy) #1, !dbg !170
 
 When a reference type or a value type containing a reference property is passed as a parameter, a copy is made (reference counter is retained before the call) and released after:
 
+```swift
+public func TestParameter()
+{
+    var vtype = VType()
+    callByVal(vtype: vtype)
+}
+
+public func callByVal(vtype: VType) { }
 ```
-%1 = load ptr, ptr %vtype.refType, align 8, !dbg !154
-%2 = call ptr @swift_retain(ptr returned %1) #4, !dbg !154
-%.refType = getelementptr inbounds %T6output5VTypeV, ptr %0, i32 0, i32 0, !dbg !156
-store ptr %1, ptr %.refType, align 8, !dbg !156
-call swiftcc void @"output.PassThrough(vtype: output.VType) -> ()"(ptr noalias nocapture dereferenceable(8) %0), !dbg !157
-%3 = call ptr @"outlined destroy of output.VType"(ptr %0), !dbg !158
+```
+%1 = load ptr, ptr %vtype.refType, align 8, !dbg !258
+%2 = call ptr @swift_retain(ptr returned %1) #5, !dbg !258
+%.refType = getelementptr inbounds %T6output5VTypeV, ptr %0, i32 0, i32 0, !dbg !260
+store ptr %1, ptr %.refType, align 8, !dbg !260
+call swiftcc void @"output.callByVal(vtype: output.VType) -> ()"(ptr noalias nocapture dereferenceable(8) %0), !dbg !261
+%3 = call ptr @"outlined destroy of output.VType"(ptr %0), !dbg !262
 ```
 
 ### Parameters passed by reference (inout)
 
 When `inout` reference type or a value type containing a reference property is passed as a parameter, it is passed by reference, so no counters are updated.
+```swift
+public func TestInOutParameter()
+{
+    var vtype = VType()
+    callByRef(vtype: &vtype)
+}
+
+public func callByRef(vtype: inout VType) { }
 ```
-%vtype.debug = alloca ptr, align 8
-call void @llvm.dbg.declare(metadata ptr %vtype.debug, metadata !285, metadata !DIExpression(DW_OP_deref)), !dbg !286
-call void @llvm.memset.p0.i64(ptr align 8 %vtype.debug, i8 0, i64 8, i1 false)
-store ptr %0, ptr %vtype.debug, align 8, !dbg !287
-ret void, !dbg !288
+```
+%vtype = alloca %T6output5VTypeV, align 8
+...
+call swiftcc void @"output.callByRef(vtype: inout output.VType) -> ()"(ptr nocapture dereferenceable(8) %vtype), !dbg !280
 ```
 
 ### Return types
@@ -105,23 +127,44 @@ When a value type is returned from a function, the caller takes ownership of the
  - If an inout parameter is returned, a copy is created and returned
 
 Pass-through callee:
+```swift
+public func TestPassThrough()
+{
+    var vtype = VType()
+    var result = PassThrough(vtype: vtype)
+}
+
+public func PassThrough(vtype: VType) -> VType
+{
+    return vtype
+}
 ```
-%test.debug = alloca ptr, align 8
-call void @llvm.dbg.declare(metadata ptr %test.debug, metadata !265, metadata !DIExpression()), !dbg !266
-call void @llvm.memset.p0.i64(ptr align 8 %test.debug, i8 0, i64 8, i1 false)
-%.refType = getelementptr inbounds %T6output5VTypeV, ptr %1, i32 0, i32 0, !dbg !267
-%2 = load ptr, ptr %.refType, align 8, !dbg !267
-store ptr %2, ptr %test.debug, align 8, !dbg !269
-%3 = call ptr @swift_retain(ptr returned %2) #5, !dbg !270
-%.refType1 = getelementptr inbounds %T6output5VTypeV, ptr %0, i32 0, i32 0, !dbg !270
-store ptr %2, ptr %.refType1, align 8, !dbg !270
-ret void, !dbg !271
+```
+%vtype.debug = alloca ptr, align 8
+%.refType = getelementptr inbounds %T6output5VTypeV, ptr %1, i32 0, i32 0, !dbg !314
+%2 = load ptr, ptr %.refType, align 8, !dbg !314
+store ptr %2, ptr %vtype.debug, align 8, !dbg !316
+%3 = call ptr @swift_retain(ptr returned %2) #5, !dbg !317
+%.refType1 = getelementptr inbounds %T6output5VTypeV, ptr %0, i32 0, i32 0, !dbg !317
+store ptr %2, ptr %.refType1, align 8, !dbg !317
+ret void, !dbg !318
 ```
 
 New instance callee:
+```swift
+public func TestNewInstance()
+{
+    var result = NewInstance()
+}
+
+public func NewInstance() -> VType
+{
+    return VType()
+}
 ```
-call swiftcc void @"$s6output5VTypeVACycfC"(ptr noalias nocapture sret(%T6output5VTypeV) %0), !dbg !302
-ret void, !dbg !303
+```
+call swiftcc void @"output.VType.init() -> output.VType"(ptr noalias nocapture sret(%T6output5VTypeV) %0), !dbg !332
+ret void, !dbg !333
 ```
 
 ## Memory handling
--- a/src/docs/memory-management.md
+++ b/src/docs/memory-management.md
@@ -103,7 +103,7 @@ call swiftcc void @"output.callByVal(vtype: output.VType) -> ()"(ptr noalias noc
 
 ### Parameters passed by reference (inout)
 
-When `inout` reference type or a value type containing a reference property is passed as a parameter, it is passed by reference, so no counters are updated.
+When `inout` reference type or a value type containing a reference property is passed as a parameter, it is passed by reference, so no reference counters are updated.
 ```swift
 public func TestInOutParameter()
 {
@@ -177,5 +177,5 @@ To ensure correct memory handling:
  - When a type is marshalled to Swift as a function parameter, `InitWithCopy` should be invoked to create the copy
  - When a type is marshalled to Swift as an `inout` function parameters, an instance reference is passed
  - When a type is marshalled to Swift as a return paramter, `InitWithCopy` should be invoked to create the copy
- - When a type is marshalled from Swift as a return paramter, a created instance in a callee is consumed and counters are not updated
+ - When a type is marshalled from Swift as a return paramter, no reference counters are updated
  - When using a private "copy" constructor on the C# side for marshalling from Swift, `InitWithCopy` should be invoked
--- a/src/docs/memory-management.md
+++ b/src/docs/memory-management.md
@@ -169,7 +169,7 @@ ret void, !dbg !333
 
 ## Memory handling
 
-To handle native memory in the scenarios above, the projections should could use the value witness table to invoke `InitWithCopy` for copy operations and `Destroy` for finalization. These functions manage reference counts at any level of nesting.
+To handle native memory in the scenarios above, the projections should use the value witness table to invoke `InitWithCopy` for copy operations and `Destroy` for finalization. These functions manage reference counts at any level of nesting.
 
 To ensure correct memory handling:
  - Swift value types that contain reference properties should be projected as C# classes
--- a/src/docs/memory-management.md
+++ b/src/docs/memory-management.md
@@ -1,0 +1,103 @@
+# Memory management
+
+This section outlines the native memory management approach for projected value types, focusing on when and how to retain and release ref counters. The goal is to ensure that native memory is handled correctly across Swift and C# boundaries.
+
+## Requirements
+ - Projected types shall call the value witness table functions to retain and release owned ref types when required
+ - Projected types shall not invoke retain or release functions when implicit copies of ref types are created on the C# side
+
+## Memory ownership
+
+To identify scenarios that require explicit memory handling, consider the following Swift example:
+```swift
+public class RefType {
+    public init() { }
+
+    deinit { }
+}
+
+public struct VType {
+    public var refType: RefType
+
+    public init() {
+        refType = RefType()
+    }
+
+    public init(refType: RefType) {
+        self.refType = refType
+    }
+}
+
+To define memory handling rules at the Swift and C# boundary, the following scenarios are included:
+ - Code blocks
+ - Initializers
+ - Parameters
+ - Return types
+
+### Code blocks
+```
+When a value type is initialized within a block, it is stack-allocated. At the end of the block, all reference properties must be deallocated. For example:
+```
+%vtype = alloca %T6output5VTypeV, align 8
+...
+%0 = call ptr @"outlined destroy of output.VType"(ptr %vtype), !dbg !156
+
+// VWT->Destroy
+define linkonce_odr hidden ptr @"outlined destroy of output.VType"(ptr %0) #8 !dbg !143 {
+entry:
+  %.refType = getelementptr inbounds %T6output5VTypeV, ptr %0, i32 0, i32 0, !dbg !144
+  %toDestroy = load ptr, ptr %.refType, align 8, !dbg !144
+  call void @swift_release(ptr %toDestroy) #1, !dbg !144
+  ret ptr %0, !dbg !144
+}
+```
+
+### Initializers
+
+When a reference type or a value type containing a reference property is passed as a parameter to a constructor, its reference count is retained before the call and released after:
+
+```swift
+var refType = RefType()
+var vtype = VType(refType: refType)
+```
+```
+%3 = call ptr @swift_retain(ptr returned %2) #4, !dbg !164
+store ptr %2, ptr %refType, align 8, !dbg !164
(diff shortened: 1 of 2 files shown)


src/docs/memory-management.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Memory handling
+
+To handle native memory in the scenarios above, the projections should could use the value witness table to invoke `InitWithCopy` for copy operations and `Destroy` for finalization. These functions manage reference counts at any level of nesting.
+
+To ensure correct memory handling:
+ - Value types that contain reference properties should be projected as C# classes with destructors that call `Destroy`
+ - When types are marshalled into Swift, `InitWithCopy` should be invoked to create the copy/ `Destroy` should be called on the copy at the end of the block
+ - When types are marshalled from Swift, `Destroy` should be called at the end of the block
+ - When using the "copy" constructor on the C# side, `InitWithCopy` should be invoked
```
vitek-karas: Could you please exactly specify what is meant by "copy" constructor?
C# typically doesn't have such a term - structs can't really have it (it's implicit effectively and can't be controlled and or intercepted), classes typically don't want it (it's rare you need to create an exact copy of an existing class).

Is this referring to the `.ctor(IntPre swiftPayload)` ?

src/docs/memory-management.md
```diff
@@ -0,0 +1,103 @@
...
+```
+%1 = load ptr, ptr %vtype.refType, align 8, !dbg !154
+%2 = call ptr @swift_retain(ptr returned %1) #4, !dbg !154
+%.refType = getelementptr inbounds %T6output5VTypeV, ptr %0, i32 0, i32 0, !dbg !156
+store ptr %1, ptr %.refType, align 8, !dbg !156
+call swiftcc void @"output.PassThrough(vtype: output.VType) -> ()"(ptr noalias nocapture dereferenceable(8) %0), !dbg !157
+%3 = call ptr @"outlined destroy of output.VType"(ptr %0), !dbg !158
+```
+
+### Return types
```
vitek-karas, line 122: What is the behavior around inout parameters? Is it the same?

src/docs/memory-management.md
```diff
@@ -0,0 +1,103 @@
+# Memory management
+
+This section outlines the native memory management approach for projected value types, focusing on when and how to retain and release ref counters. The goal is to ensure that native memory is handled correctly across Swift and C# boundaries.
+
+## Requirements
+ - Projected types shall call the value witness table functions to retain and release owned ref types when required
+ - Projected types shall not invoke retain or release functions when implicit copies of ref types are created on the C# side
+
+## Memory ownership
```
vitek-karas, line 9: I would make it explicit that this describes Swift behavior - and only Swift behavior.

src/docs/memory-management.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+## Memory handling
+
+To handle native memory in the scenarios above, the projections should could use the value witness table to invoke `InitWithCopy` for copy operations and `Destroy` for finalization. These functions manage reference counts at any level of nesting.
+
+To ensure correct memory handling:
+ - Value types that contain reference properties should be projected as C# classes with destructors that call `Destroy`
+ - When types are marshalled into Swift, `InitWithCopy` should be invoked to create the copy/ `Destroy` should be called on the copy at the end of the block
+ - When types are marshalled from Swift, `Destroy` should be called at the end of the block
```
vitek-karas: So for C# code like this - it would do something like:
```csharp
ProjectedVType a;

///
// IntPtr r = SwiftPinvoke();
// a = ProjectedVType.NewFromPayload(r);
//     - this internally does a = new ProjectedVType(r)
//           - this in turn effectively does a._payload = InitWithCopy(r);
// Destroy(r);
a = SwiftCallWhichReturnsVType(); 

// Destroy(a._payload)
a.Dispose(); // Or destructor
```

I don't know enough about Swift if this is what will happen in Swift's case as well. C# doesn't have the corresponding "block variable" behavior (object lifetime is not scoped by blocks really). So the equivalent Swift code would be to assign the return value of the function to a field on for example a ref class.

src/docs/memory-management.md
```diff
@@ -0,0 +1,103 @@
...
+
+    public init() {
+        refType = RefType()
+    }
+
+    public init(refType: RefType) {
+        self.refType = refType
+    }
+}
+
```
jkurdek, line 33: missing ```?

src/docs/memory-management.md
```diff
@@ -0,0 +1,103 @@
...
+%3 = call ptr @"outlined destroy of output.VType"(ptr %0), !dbg !158
+```
+
+### Return types
+
+When a value type is returned from a function, a new instance is created, and its reference count is retained before the return. The caller is responsible for releasing the reference:
+```
+%2 = load ptr, ptr %.refType, align 8, !dbg !173
+store ptr %2, ptr %vtype.debug, align 8, !dbg !175
+%3 = call ptr @swift_retain(ptr returned %2) #4, !dbg !176
```
jkurdek: I think this retain doesnt come from the allocation but from the assignment to a field.

```swift
public func ReturnRefType(a: Int) -> RefType {
    return RefType()
}
```

```
define protected swiftcc ptr @"$s6output13ReturnRefType1aAA0cD0CSi_tF"(i64 %0) #0 !dbg !41 {
entry:
  %a.debug = alloca i64, align 8
  call void @llvm.dbg.declare(metadata ptr %a.debug, metadata !47, metadata !DIExpression()), !dbg !49
  call void @llvm.memset.p0.i64(ptr align 8 %a.debug, i8 0, i64 8, i1 false)
  store i64 %0, ptr %a.debug, align 8, !dbg !50
  %1 = call swiftcc %swift.metadata_response @"$s6output7RefTypeCMa"(i64 0) #9, !dbg !51
  %2 = extractvalue %swift.metadata_response %1, 0, !dbg !51
  %3 = call swiftcc ptr @"$s6output7RefTypeCACycfC"(ptr swiftself %2), !dbg !51
  ret ptr %3, !dbg !52
}
```

It seems that  creates a new instance with ref count set to one 
```
%3 = call swiftcc ptr @"$s6output7RefTypeCACycfC"(ptr swiftself %2), !dbg !51 
```

src/docs/memory-management.md
jkurdek: I think we don't consider structs here as anything that has a ref type (hence need `InitWithCopy`) will be projected as class

src/docs/memory-management.md
jkurdek: Reading this document I get to a conclusion that we do not need `InitWithCopy` when dealing with return values as the reference count will be updated by the callee. Is this right?

vitek-karas: That was sort of where my question is going to... if Swift can avoid the +/- on return values, maybe we can as well.
Another question related to this is - what are all the ways we can get to an instance of a swift object?
- Return value of a method call
- Create from nothing (driven by us)

The ones which are not covered by this but will become important eventually are:
- Passed in as a parameter of a method implemented in C# but called from Swift
- Accessed as a direct field access? (do we have this? can we actually directly access fields at all anywhere, or do we always go through accessors)

jkurdek: >Accessed as a direct field access? (do we have this? can we actually directly access fields at all anywhere, or do we always go through accessors)

We could technically access things directly on things projected as structs - we do not do it now though But then those would be guaranteed to not have ref types (if they had the projection would be a class)

src/docs/memory-management.md
kotlarmilos: This is a private "copy" constructor on the C# side used for marshalling from Swift. I will clarify it.

src/docs/memory-management.md
kotlarmilos: I will expand the docs with `inout` parameters.

src/docs/memory-management.md
kotlarmilos: Good catch, you are right. Let me update the docs.

src/docs/memory-management.md
kotlarmilos: Updated, please take a look.

src/docs/memory-management.md
kotlarmilos: Added, please take a look.

src/docs/memory-management.md
kotlarmilos: Update: There are two scenarios -- pass-through and new instance.

Pass-through callee:
```
%test.debug = alloca ptr, align 8
call void @llvm.dbg.declare(metadata ptr %test.debug, metadata !265, metadata !DIExpression()), !dbg !266
call void @llvm.memset.p0.i64(ptr align 8 %test.debug, i8 0, i64 8, i1 false)
%.refType = getelementptr inbounds %T6output5VTypeV, ptr %1, i32 0, i32 0, !dbg !267
%2 = load ptr, ptr %.refType, align 8, !dbg !267
store ptr %2, ptr %test.debug, align 8, !dbg !269
%3 = call ptr @swift_retain(ptr returned %2) #5, !dbg !270
%.refType1 = getelementptr inbounds %T6output5VTypeV, ptr %0, i32 0, i32 0, !dbg !270
store ptr %2, ptr %.refType1, align 8, !dbg !270
ret void, !dbg !271
```

New instance callee:
```
call swiftcc void @"$s6output5VTypeVACycfC"(ptr noalias nocapture sret(%T6output5VTypeV) %0), !dbg !302
ret void, !dbg !303
```

It seems that in both cases a new instance/copy is created and the caller has the ownership.

src/docs/memory-management.md
kotlarmilos: Refined the rules, please take a look. 

> Return value of a method call
> Create from nothing (driven by us)

Caller has the ownership and release them at the end of the block.

> Passed in as a parameter of a method implemented in C# but called from Swift

Only if the parameter is returned, a new copy is created/count is retained. Anyway, a caller has the ownership.

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+ - Code blocks
+ - Initializers
+ - Parameters passed by value
+ - Parameters passed by reference (inout)
+ - Return types
+
+### Code blocks
+
+When a value type is initialized within a block, it is stack-allocated. At the end of the block, all reference properties must be deallocated. For example:
+```
```
AaronRobinsonMSFT, line 50: Is there swift code to translate this to?

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+call swiftcc void @"output.VType.init(refType: output.RefType) -> output.VType"(ptr noalias nocapture sret(%T6output5VTypeV) %vtype, ptr %2), !dbg !168
+%toDestroy = load ptr, ptr %refType, align 8, !dbg !170
+call void @swift_release(ptr %toDestroy) #1, !dbg !170
+```
+
+### Parameters passed by value
+
+When a reference type or a value type containing a reference property is passed as a parameter, a copy is made (reference counter is retained before the call) and released after:
+
+```
```
AaronRobinsonMSFT, line 94: Swift code?

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+%.refType = getelementptr inbounds %T6output5VTypeV, ptr %0, i32 0, i32 0, !dbg !156
+store ptr %1, ptr %.refType, align 8, !dbg !156
+call swiftcc void @"output.PassThrough(vtype: output.VType) -> ()"(ptr noalias nocapture dereferenceable(8) %0), !dbg !157
+%3 = call ptr @"outlined destroy of output.VType"(ptr %0), !dbg !158
+```
+
+### Parameters passed by reference (inout)
+
+When `inout` reference type or a value type containing a reference property is passed as a parameter, it is passed by reference, so no counters are updated.
+```
```
AaronRobinsonMSFT, line 115: Swift code?

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+
+New instance callee:
+```
+call swiftcc void @"$s6output5VTypeVACycfC"(ptr noalias nocapture sret(%T6output5VTypeV) %0), !dbg !302
+ret void, !dbg !303
+```
+
+## Memory handling
+
+To handle native memory in the scenarios above, the projections should could use the value witness table to invoke `InitWithCopy` for copy operations and `Destroy` for finalization. These functions manage reference counts at any level of nesting.
```
AaronRobinsonMSFT:
```suggestion
To handle native memory in the scenarios above, the projections should use the value witness table to invoke `InitWithCopy` for copy operations and `Destroy` for finalization. These functions manage reference counts at any level of nesting.
```

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+
+To ensure correct memory handling:
+ - Swift value types that contain reference properties should be projected as C# classes
+ - When a type goes out of the block the destructor/dispose should invoke `Destroy` function
```
AaronRobinsonMSFT: > destructor

We don't have those. If the finalizer is being referred to here, it might be worth considering inheriting from `SafeHandle` if there is any possibility of resurrection. Are we requiring callers to wrap in a `using` statement? If so, it might be more practical store the pointer in a `SafeHandle` and make it a field on the projected type. This would allow us to drop any `Dispose` requirement.

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+ - When a type is marshalled to Swift as a function parameter, `InitWithCopy` should be invoked to create the copy
+ - When a type is marshalled to Swift as an `inout` function parameters, an instance reference is passed
```
AaronRobinsonMSFT: So this means there are no lifetime considerations for us. What about if the instance is moved to something that outlives the method scope (for example, added to a C# collection), do we then make a copy?

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+ - When a type is marshalled to Swift as a return paramter, `InitWithCopy` should be invoked to create the copy
```
AaronRobinsonMSFT:
```suggestion
 - When a type is marshalled to Swift as a return parameter, `InitWithCopy` should be invoked to create the copy
```

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+ - When a type is marshalled from Swift as a return paramter, a created instance in a callee is consumed and counters are not updated
```
AaronRobinsonMSFT:
```suggestion
 - When a type is marshalled from Swift as a return parameter, a created instance in a callee is consumed and counters are not updated
```

src/docs/memory-management.md
```diff
@@ -0,0 +1,138 @@
...
+ - When using a private "copy" constructor on the C# side for marshalling from Swift, `InitWithCopy` should be invoked
```
AaronRobinsonMSFT, line 181: What is a "copy" constructor in C#? Does this mean "making a copy"?

src/docs/memory-management.md
AaronRobinsonMSFT: > a created instance in a callee is consumed 

Consumed as in, returned to the caller?

"counters" == "ref count"?

src/docs/memory-management.md
kotlarmilos: Good point, added.

src/docs/memory-management.md
kotlarmilos: Simplified with "no reference counters are updated"

> "counters" == "ref count"?

Yes, thanks

src/docs/memory-management.md
kotlarmilos: Are you referring to something like this?
```csharp
internal sealed class SwiftVTypeSafeHandle : SafeHandleZeroOrMinusOneIsInvalid
{
    public SwiftVTypeSafeHandle(IntPtr handle)
        : base(ownsHandle: true)
    {
        SetHandle(handle);
    }

    protected override bool ReleaseHandle()
    {
        // VWT->Destroy
        return true;
    }
}

public class SwiftVType
{
    private SwiftVTypeSafeHandle _handle;

    public SwiftVType(IntPtr nativePointer)
    {
        _handle = new SwiftVTypeSafeHandle(nativePointer);
    }

    ~SwiftVType()
    {
        if (_handle != null && !_handle.IsClosed)
        {
            _handle.Dispose();
        }
    }
}
```
> Are we requiring callers to wrap in a using statement?

We don't have any strong requirements. Would this pollute the usage of the bindings?

src/docs/memory-management.md
kotlarmilos: In this case, we should not create a copy because it is a reference type in C#, and the destroy function should not be called more times than allowed by the reference count.

I tried to emphasize this in the requirements:
> - Projected types shall not invoke retain or release functions when implicit copies of value types with reference properties are created on the C# side

If we do not consider the bindings as first-class citizens, meaning we only use these types at the C#‑Swift boundary, we shouldn't end up in such scenarios. The question remains whether they will be usable.

src/docs/memory-management.md
kotlarmilos: Yes, it is a private constructor that has a native handle as a parameter. It is typically used in the marshalling from Swift, when a copy is needed.

/cc: @jkurdek Feel free to expand with more scenarios.

src/docs/memory-management.md
AaronRobinsonMSFT: Generally, yes. The finalizer in `SwiftVType` wouldn't be needed though. It could be the case that `SwiftVType` implemented `IDisposable`, but that would depend on UX considerations. Just having the `SwiftVTypeSafeHandle` be a field in `SwiftVType` should be sufficient.

vitek-karas: So to clarify my understanding - the proposal is to rely on the finalizer in `SafeHandle` to guarantee freeing the native resources and disconnect the projection from this responsibility. The projection type may or may not implement IDisposable to provide a .NET native way of handling the native resource dependency. Correct?

I like this - `SafeHandle` is one of the "standard" ways to manage native resources in BCL, so that fits. I also like the flexibility of choosing if we want to expose `IDisposable` or not.

_Future consideration: we may consider adding a different way to free the native resource which is not `IDisposale`. `IDisposable` is the standard way, but currently it carries a rather significant cost - there are analyzers which make it "viral". Having a different way would allow us to provide a way to free the native resource without the "viral" part._

AaronRobinsonMSFT: > The projection type may or may not implement IDisposable to provide a .NET native way of handling the native resource dependency.

Correct.

> Future consideration: we may consider adding a different way to free the native resource which is not IDisposale

Agree. I personally don't like `IDisposable` for object instance projections. Streams and memory buffers make sense to me because they are in and of themselves just sinks/sources of data. Object instances on the other have have state and functionality, which is different. The GC manages these objects directly because that is how we use them in OOP.  If we didn't live in an OOP world then `IDisposable` would seem to make sense, but we don't. I had previously suggested following the COM pattern and adding a similar API to [`Marshal.ReleaseComObject`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.marshal.releasecomobject) for Swift.

None of this means we can't use `IDisposable`, but as you said, we'd make it an explicit decisions in cases that didn't become viral and where it was intuitive. Any new API for explicit release would likely be for niche scenarios or testing for memory leaks when predictable control was needed.

vitek-karas:
One other consideration - .NET GC's inability to see the Swift instance.
We currently have an issue in ObjectiveC (and it does exist in Android as well, but to lesser degree) where the app is killed by the OS due to Out of memory situation - but GC didn't even try to run. This is caused by a managed projection of a large native object. GC doesn't see memory pressure since the managed heap is small (the projection is just a pointer), but the app consumes large amounts of memory due to the native objects which can't be freed since .NET holds a reference to them.

GC has a mechanism to tell it about the native allocated memory:
https://learn.microsoft.com/en-us/dotnet/api/system.gc.addmemorypressure?view=net-9.0

@AaronRobinsonMSFT did we do something like this in other interops you're aware of?

AaronRobinsonMSFT:
> @AaronRobinsonMSFT did we do something like this in other interops you're aware of?

This exists for UWP. It is on the `IReferenceTrackerHost` interface via [`IReferenceTrackerHost::AddMemoryPressure`](https://learn.microsoft.com/windows/win32/api/windows.ui.xaml.hosting.referencetracker/nf-windows-ui-xaml-hosting-referencetracker-ireferencetrackerhost-addmemorypressure) and [`IReferenceTrackerHost::RemoveMemoryPressure`](https://learn.microsoft.com/windows/win32/api/windows.ui.xaml.hosting.referencetracker/nf-windows-ui-xaml-hosting-referencetracker-ireferencetrackerhost-removememorypressure). There were some places the interop sub-system previously called `GC.AddMemoryPressure`, but I think most of them have been removed over time.

I don't think we should be adding any new public interop APIs for this mechanism though. I consider the `IReferenceTrackerHost` interface a legacy approach for interop. Going forward the recommendation would be that upstack interop libraries should expose/abstract any managed API that provides the needed functionality.

A simple approach is below, but one could architect any number of approaches to encapsulate the same idea.

```csharp
[UnmanagedCallersOnly]
static void AddMemoryPressure(nint memoryInBytes)
    => GC.AddMemoryPressure((long)memoryInBytes);

[UnmanagedCallersOnly]
static void RemoveMemoryPressure(nint memoryInBytes)
    => GC.RemoveMemoryPressure((long)memoryInBytes);
```

vitek-karas:
I didn't mean to add these to some native API - in fact for Swift I don't think we would need it from native code at all. I was mostly curious if we call these somewhere and if we have experience with it. In Swift I actually think we should not call these by default - mostly because it's very hard to determine even an estimate of true object size (what if I have a swift struct which itself has just a pointer to some large piece of memory - swift metadata would not know about that either). So I imagine that this would be similar to IDisposable - we would add these as an opt-in for things we know are big (like projection of a class which represents an image).

jkotas:
> GC has a mechanism to tell it about the native allocated memory:

BTW: GC.AddMemoryPressure / GC.RemoveMemoryPressure APIs are not part of the core GC. These APIs are implemented as a standalone set of counters and heuristic on top of these counters that calls GC.Collect when it seems to be profitable.

src/docs/memory-management.md
kotlarmilos: Thanks for the feedback. Created a tracking issue: https://github.com/dotnet/runtimelab/issues/3018

kotlarmilos:
Added to https://github.com/dotnet/runtimelab/issues/3019

src/docs/memory-management.md
AaronRobinsonMSFT: I agree with the above statement, but for `inout` it would seem we are getting an non-incremented reference. 

This would mean that if a Swift instance was passed into C# by reference and it was simply used within the C# callee the reference count would be fine. However, if the projected Swift instance in C# remains in C# beyond the lifetime of the callee, don't we need to ensure its lifetime is extended?

kotlarmilos: Yes, that is right, a copy should be made. Added to the docs.

src/docs/memory-management.md
jkurdek: I think we might not need this "copy" constructor. Reading this write up it seems to me that whenever we get something from swift we can assume that the references were updated correctly (including the one coming from csharp). I would say it should be safe to write the value without making a copy.

jkurdek (approved):
Great work Milos!

AaronRobinsonMSFT (approved):

PR #3009, opened 2025-02-18
[Swift ABI compatibility] Functional tests for `InitWithCopy` and `Destroy` functions in value witness table

kotlarmilos:
## Description

This PR adds functional tests for `InitWithCopy` and `Destroy` functions in value witness table. The goal is to verify that `InitWithCopy` and `Destroy` could be used to maintain ref counting for value types at arbirary level of nesting. Additionally, 

They could be used to verify refactoring effort in destructor and marshalling scenarios.

Additionally, the tests can be used to verify assumptions about the ARC mechanism and to validate future changes in destructor and marshalling scenarios.

## Out-of-scope

This PR doesn't change projections.


src/Swift.Runtime/src/Swift/Runtime/ValueWitnessTable.cs
```diff
@@ -38,7 +38,7 @@ public unsafe ref struct ValueWitnessTable
         /// void Destroy (object, witnessTable)
```
jkurdek: NIT: update the comment as well

src/Swift.Bindings/tests/IntegrationTests/FunctionalTests/MemoryTests/MemoryTests.swift
```diff
@@ -0,0 +1,130 @@
+// // Copyright (c) Microsoft Corporation.
```
jkurdek: NIT: double comment

jkurdek (approved):

Issue #2849, opened 2024-12-03
[Swift shipping] Create `Swift.Bindings.Experimental` and tooling NuGet packages in the dotnet-experimental feed


kotlarmilos:
Tasks:
 - [x] https://github.com/dotnet/runtimelab/pull/2878
 - [x] https://github.com/dotnet/runtimelab/pull/2879
 - [x] #2897
 - [x] https://github.com/dotnet/runtimelab/pull/2900
 - [x] ~Make the tooling and tests compatible with Windows dev environment~ Note: MachO reader is Apple-specific.
 - [x] Update packaging to bundle generated Swift code

> [!NOTE]  
> This issue tracks progress and will be expanded with more details. If specific tasks for this issue are not clear, this means that the task is not well defined yet and needs attention.

PR #3014, opened 2025-02-21
[Swift shipping] Add `Swift.Bindings.Experimental` nuget to `dotnet-experimental` for device tests

kotlarmilos:
## Description

This PR adds the `Swift.Bindings.Experimental` nuget package to the `dotnet-experimental` feed for device testing. The package provides both C# and Swift bindings for selected frameworks. 

It will be used on Helix for testing `StoreKit` scenarios and other frameworks in the future.

## Changes
The `generate.sh` script is updated to generate an `.xcframework` that contains platform and architecture specific bindings. The publish job is updated to upload `Swift.Bindings.Experimental` artifacts to the storage.

## Testing
Internal build for the `dotnet-experimental` feed.

## Out-of-scope
The bindings are generated with the `--experimental` flag and currently contain only the `Swift.Runtime` namespace. Once the package is produced and validated locally from the `feature/swift-bindings`, the flag will be removed and the bindings will be updated to include `StoreKit` bindings.

Fixes https://github.com/dotnet/runtimelab/issues/2849
Contributes to https://github.com/dotnet/runtimelab/issues/2850

### DIFF
--- a/generate.sh
+++ b/generate.sh
@@ -111,8 +111,8 @@ function InvokeProjectionTooling {
     fi
 }
 
-# Function to generate NuGet package
-function PackNuGet {
+# Function to generate project
+function CreateProject {
     local project_file="./Swift.Bindings.${platform}.Experimental.csproj"
 
     cat <<EOL > "$project_file"
@@ -129,7 +129,7 @@ function PackNuGet {
 </Project>
 EOL
 
-    $scriptroot/dotnet.sh pack "$project_file"
+    $scriptroot/dotnet.sh build "$project_file"
 }
 
 function Generate {
@@ -144,7 +144,7 @@ function Generate {
         fi
     done
 
-    PackNuGet
+    CreateProject
 
     echo "Process completed."
 }
--- a/generate.sh
+++ b/generate.sh
@@ -6,14 +6,13 @@ usage()
 {
   echo "Common settings:"
   echo "  --platform <value>         Platform: MacOSX, iPhoneOS, iPhoneSimulator, AppleTVOS, AppleTVSimulator"
+  echo "  --version <value>          Version of the SDK"
   echo "  --arch <value>             Architecture: arm64e-apple-macos, x86_64-apple-macos"
   echo "  --framework <value>        Framework to generate bindings for"
   echo "  --configuration <value>    Configuration: Debug, Release"
+  echo "  --experimental             Generates only Runtime.Swift namespace when bindings for frameworks are not complete"
   echo "  --help                     Print help and exit (short: -h)"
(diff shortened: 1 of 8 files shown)


rolfbjarne (approved):

generate.sh
```diff
@@ -101,19 +105,112 @@ function InvokeProjectionTooling {
...
     cat "./Swift.$framework.cs"
     echo ""
 
     if $experimental; then
         rm -rf "./Swift.$framework.cs"
+        rm -rf "./Swift.$framework.swift"
+    fi
+
+    if [ -f "Swift.${framework}.swift" ]; then
+        echo "import ${framework}" | cat - Swift.${framework}.swift > temp && mv temp Swift.${framework}.swift
```
matouskozak, line 123: This is used when we need to generate swift code to make the bindings work?

generate.sh
```diff
@@ -101,19 +105,112 @@ function InvokeProjectionTooling {
...
+
+    # x86_64
+    echo "Building ${framework} for ${platform} x86_64..."
+    swiftc -emit-library -target x86_64-apple-$(echo "$platform" | tr '[:upper:]' '[:lower:]')${version} -module-name ${framework} -o ${framework}-${platform}-x64.dylib *.swift -F /Applications/Xcode.app/Contents/Developer/Platforms/${platform}.platform/Developer/SDKs/${platform}.sdk/System/Library/Frameworks/
+
+    # arm64
+    echo "Building ${framework} for ${platform} arm64..."
+    swiftc -emit-library -target arm64-apple-$(echo "$platform" | tr '[:upper:]' '[:lower:]')${version} -module-name ${framework} -o ${framework}-${platform}-arm64.dylib *.swift -F /Applications/Xcode.app/Contents/Developer/Platforms/${platform}.platform/Developer/SDKs/${platform}.sdk/System/Library/Frameworks/
+
+    if [[ $platform == "MacOSX" ]]; then
```
matouskozak, line 144: Why do we need the extra `ios15.0-macabi` targets for `MacOSX`?

matouskozak (approved):

generate.sh
kotlarmilos: This code snippet appends framework import in the generated Swift bindings to allow for successful compilation. The projection tooling generates some Swift code but doesn't have context about frameworks.

generate.sh
kotlarmilos: When targeting the desktop, we generate both `macOS` and `MacCatalyst` bindings. The primary targets for these bindings are Apple mobile platforms, and some frameworks are mobile-only.

PR #3016, opened 2025-02-21
[SwiftBindings] Add type for storing information about csharp type name

jkurdek:
Contributes to #2889 

The PR adds a new type to represent csharp type name plus some unit tests

### DIFF
--- a/src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
+++ b/src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
@@ -1,0 +1,65 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace BindingsGeneration;
+
+/// <summary>
+/// Represents a C# type name.
+/// </summary>
+public record CSharpTypeName
+{
+    /// <summary>
+    /// The namespace name.
+    /// </summary>
+    public string Namespace { get; }
+
+    /// <summary>
+    /// The type name.
+    /// </summary>
+    public string Name { get; }
+
+    /// <summary>
+    /// The fully qualified type name, including namespace.
+    /// </summary>
+    public string FullyQualifiedName { get; }
+
+    /// <inheritdoc />
+    public override string ToString() => FullyQualifiedName;
+
+    private CSharpTypeName(string @namespace, string name, string fullyQualifiedName)
+    {
+        Namespace = @namespace;
+        Name = name;
+        FullyQualifiedName = fullyQualifiedName;
+    }
+
+    /// <summary>
+    /// Creates a new CSharpTypeName from a namespace and type name.
+    /// </summary>
+    /// <param name="namespace">The namespace.</param>
+    /// <param name="name">The type name.</param>
+    /// <returns>The CSharpTypeName.</returns>
+    public static CSharpTypeName FromNamespaceAndName(string @namespace, string name)
+    {
+        ArgumentException.ThrowIfNullOrEmpty(name, nameof(name));
+        ArgumentException.ThrowIfNullOrEmpty(@namespace, nameof(@namespace));
+
+        if (name.Contains('<'))
+        {
+            throw new ArgumentException("Cannot create a CSharpTypeName from a generic type.");
+        }
+
+        var fullyQualifiedName = $"{@namespace}.{name}";
+        return new CSharpTypeName(@namespace, name, fullyQualifiedName);
+    }
+
+    /// <summary>
+    /// C# type name for void.
+    /// </summary>
+    public static readonly CSharpTypeName VoidType = new CSharpTypeName("", "", "void");
+
+    /// <summary>
+    /// C# type name for object.
+    /// </summary>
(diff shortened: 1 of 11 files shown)


src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    /// <summary>
+    /// The type name.
+    /// </summary>
+    public string Name { get; }
+
+    /// <summary>
+    /// The fully qualified type name, including namespace.
+    /// </summary>
+    public string FullyQualifiedName { get; }
```
vitek-karas, line 24: I assume this is not assembly qualified, at least make it clear.
It also means that we can't ever rely on this being unique across C# types.

src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        var fullyQualifiedName = $"{@namespace}.{name}";
+        return new CSharpTypeName(@namespace, name, fullyQualifiedName);
+    }
+
+    /// <summary>
+    /// C# type name for void.
+    /// </summary>
+    public static readonly CSharpTypeName VoidType = new CSharpTypeName("", "", "void");
```
vitek-karas, line 59: Shouldn't this be `new CSharpTypeName("System", "Void", "System.Void");` ?
I guess it depends if we want to model this as a semantic model versus syntax.

Similarly - what will string look like? Or int, ...
Definitely worth a comment.

vitek-karas (approved):

src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        ArgumentException.ThrowIfNullOrEmpty(name, nameof(name));
+        ArgumentException.ThrowIfNullOrEmpty(@namespace, nameof(@namespace));
+
+        if (name.Contains('<'))
+        {
+            throw new ArgumentException("Cannot create a CSharpTypeName from a generic type.");
+        }
+
+        var fullyQualifiedName = $"{@namespace}.{name}";
```
matouskozak, line 52: Is it possible we might create a `fullyQualifiedName` that would coincide with a contemporary C# type and could it cause an issue down the line?

matouskozak (approved):

src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace BindingsGeneration;
+
+/// <summary>
+/// Represents a C# type name.
+/// </summary>
+public record CSharpTypeName
```
kotlarmilos, line 9: What is the corresponding type for the Swift types?

src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace BindingsGeneration;
+
+/// <summary>
+/// Represents a C# type name.
```
kotlarmilos, line 7: I would emphasis that this is a name used for lookups.

kotlarmilos (approved):

src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
jkurdek: This represents a name as it would occur in the generated code. Writing `System.Void` instead of `void` in c# is illegal when specyfying return type. For other primitives we can do e.g.`System.Int32` or `System.Boolean`. We can introduce aliases later on

src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
jkurdek: yes it is certainly possible to do this now. I will address is in a follow up when implementing a mechanism which does translation of types from swift to csharp.

src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
jkurdek: The corresponding type is `SwiftTypeName`, we already have in the repo.

src/Swift.Bindings/src/Model/TypeNames/CSharpTypeName.cs
jkurdek: This is not intended to be used for lookups (`SwiftTypeName` is). When traversing a abi.json tree we see swift type names, we use them for querying type database adn we extract c# names from the value of the record.

Issue #2960, opened 2025-01-27
[Swift language feature] Implement Swift async support for generics


kotlarmilos:
## Tasks
 - [x] Update the method handler to generate the correct bindings
 - [x] Add functional tests

Issue #3022, opened 2025-02-24
[Swift language feature] Implement generic return type in Swift async functions


kotlarmilos:
## Description

We need to implement AOT compatible C# async callback. Here is an example:
```csharp
[UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvCdecl) })]
private static void GenericCollectionConstraintOnComplete(System.IntPtr rawResult, IntPtr task)
{
    GCHandle handle = GCHandle.FromIntPtr(task);
    try
    {

        var result = SwiftMarshal.MarshalFromSwift<System.IntPtr>((SwiftHandle)new IntPtr(&rawResult));
        if (handle.Target is TaskCompletionSource<System.IntPtr> tcs)
        {
            tcs.TrySetResult(result);
        }
    }
    finally
    {
        handle.Free();
    }
}
```
The `rawResult` parameter should be generic type, which is not allowed in `UnmanagedCallersOnly` methods.

PR #3023, opened 2025-02-24
[Swift language feature] Implement generic parameters/types with generics in Swift async functions

kotlarmilos:
## Description

This PR implements generic parameters/types with generics in Swift async functions. This is needed to correctly project StoreKit examples:
- https://developer.apple.com/documentation/storekit/product/products(for:)
- https://developer.apple.com/documentation/storekit/product/purchase(options:)

## Changes
 - In addition to the generic parameters, the ABI parser collects associated type conformances. They are needed to correctly invoke Swift async functions from the Swift wrappers and aren't used in the C# bindings
 - C# callback is updated to correctly marshal results at the ABI boundary
 - Async integration tests are expanded and moved to the separate file

## Out-of-scope

Generic return types in Swift async functions are not supported yet. Tracking issue https://github.com/dotnet/runtimelab/issues/3022

Fixes https://github.com/dotnet/runtimelab/issues/2960
Fixes https://github.com/dotnet/runtimelab/issues/3020

### DIFF
--- a/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
@@ -391,7 +391,7 @@ namespace BindingsGeneration
         {
             foreach (var genericParameter in _env.MethodDecl.GenericParameters)
             {
-                var conformances = genericParameter.Constraints.OrderBy(c => c.Protocol.ModuleQualifiedName);
+                var conformances = genericParameter.GenericConformances.OrderBy(c => c.Protocol.ModuleQualifiedName);
                 foreach (var conformance in conformances)
                 {
                     var pwtName = NameProvider.GetProtocolWitnessTableName(_env.GenericTypeMapping[genericParameter.TypeName].TypeParameter, conformance.Protocol.Name);
@@ -532,7 +532,7 @@ namespace BindingsGeneration
 
             csWriter.WriteLine("[UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]");
             csWriter.WriteLine($"[DllImport(\"{libPath}\", EntryPoint = \"{NameProvider.GetMangledName(methodDecl)}\")]");
-            csWriter.WriteLine($"private static extern {pInvokeSignature.ReturnType} {pInvokeName}({pInvokeSignature.ParametersString()});");
+            csWriter.WriteLine($"private static extern {(methodDecl.IsAsync ? "void" : pInvokeSignature.ReturnType)} {pInvokeName}({pInvokeSignature.ParametersString()});");
         }
     }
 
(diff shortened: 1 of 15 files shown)


src/Swift.Bindings/src/Parser/GenericSignatureParser.cs
```diff
@@ -92,8 +92,7 @@ private static List<GenericParameterConformance> ExtractConstraints(string signa
         var target = parts[0];
         var protocol = parts[1];
 
-        if (!target.Contains(".")) return new GenericParameterConformance(target, SwiftTypeName.FromModuleQualifiedName(protocol));
-
-        return null;
+        var isAssociatedType = target.Contains(".");
+        return new GenericParameterConformance(target, SwiftTypeName.FromModuleQualifiedName(protocol), isAssociatedType);
```
jkurdek: Neat!

src/Swift.Bindings/src/Model/TypeDecl/ProtocolConformance.cs
```diff
@@ -19,7 +19,9 @@ SwiftTypeName Protocol
 /// </summary>
 /// <param name="GenericParameter">The generic parameter</param>
 /// <param name="Protocol">The protocol that the generic parameter conforms to</param>
+/// <param name="IsAssociatedType">Indicates if the conformance is on an associated type</param>
 public record GenericParameterConformance(
     string GenericParameter,
-    SwiftTypeName Protocol
+    SwiftTypeName Protocol,
```
jkurdek: I think this is a bit more complex than this. 

When dealing with a simple generic conformance to a protocol we basically have a pair (generic parameter, protocol)

With PATs you can put constraints on the associated type. So the LHS (GenericParameter) can be a path - which we will need to parse. Also the constraint can be either a protocol or a concrete type. I think we will need a separate type to model this as the difference is greater than a bool flag.

I'd suggest creating another type which models a path and a Protocol / Type conformance.

~~What is the use case of this change?~~

kotlarmilos: Updated to parse path and distinguish between concrete types and protocols. The path is a simple array of strings which could be improved in the future according to the needs.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -659,29 +665,58 @@ private void EmitAsync(CSharpWriter csWriter, SwiftWriter swiftWriter)
...
                 new[]
                 {
-                    $"callback: @escaping ({(isEmptyTuple ? "" : $"{_env.MethodDecl.CSSignature.First().SwiftTypeSpec}, ")}Int64) -> Void",
+                    $"callback: @escaping ({(isEmptyTuple ? "" : $"{(_env.MethodDecl.CSSignature.First().IsGeneric ? _env.MethodDecl.GenericParameters[0].SugaredTypeName : _env.MethodDecl.CSSignature.First().SwiftTypeSpec)}, ")}Int64) -> Void",
                     "task: Int64"
                 }.Concat(
                     _env.MethodDecl.CSSignature
                         .Skip(1)
-                        .Select(p => $"{p.Name}: {p.SwiftTypeSpec}")
+                        .Select(p => $"{p.Name}: {(p.IsGeneric ? _env.MethodDecl.GenericParameters[genericIndex++].SugaredTypeName : p.SwiftTypeSpec)}")
```
jkurdek: NIT: this becomes rather hard to understand - incrementing an indexing within a select within a concat.

kotlarmilos: Yes, it is incremented only if a generic parameter is encountered. improved it with `.Find` expression.

jkurdek (approved):

PR #3027, opened 2025-03-04
[NativeAOT-LLVM] Start running the wasi library tests

jsturtevant:
upstream stopped running the library tests for wasi with https://github.com/dotnet/runtime/pull/112716.  To ensure we still have coverage this adds them back on the branch here.

Note, I am not sure this will work as is.  Still figuring out the how the build system works. Any pointers welcome :-)

fyi @pavelsavara

### DIFF
--- a/eng/pipelines/runtimelab.yml
+++ b/eng/pipelines/runtimelab.yml
@@ -100,7 +100,7 @@ extends:
             nameSuffix: '_Smoke'
             extraBuildArgs: /p:EnableAggressiveTrimming=true /p:RunWasmSamples=true /p:AotHostArchitecture=x64 /p:AotHostOS=$(_hostedOS)
             shouldRunSmokeOnly: true
-            alwaysRun: ${{ variables.isRollingBuild }}
+            alwaysRun: true
             scenarios:
               - WasmTestOnWasmtime
 
--- a/eng/pipelines/runtimelab.yml
+++ b/eng/pipelines/runtimelab.yml
@@ -88,6 +88,22 @@ extends:
                 - template: /eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
                   parameters:
                     librariesConfiguration: Debug
+        
+        #
+        # Build and test wasi library tests
+        #
+        - template: /eng/pipelines/common/templates/wasm-library-tests.yml
+          parameters:
+            platforms:
+              - wasi_wasm
+              - wasi_wasm_win
+              - wasi_wasm_linux_x64_naot_llvm
+            nameSuffix: '_Smoke'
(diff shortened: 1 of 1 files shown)


SingleAccretion:
@jsturtevant the NAOT-LLVM libraries tests are defined here:

https://github.com/dotnet/runtimelab/blob/0bcd11183e4b1b473d3e89a4c8e44d8796fb92fc/src/libraries/tests.proj#L622-L629

It's intentionally only a very small subset of tests since they take a long time to compile.

https://github.com/dotnet/runtimelab/blob/0bcd11183e4b1b473d3e89a4c8e44d8796fb92fc/eng/pipelines/runtimelab/runtimelab-post-build-steps.yml#L69-L72

pavelsavara:
You can also have separate CI leg for Mono WASI and run tests there.

SingleAccretion:
> You can also have separate CI leg for Mono WASI and run tests there.

Well... I don't think we would have the resources (time/expertise) to keep it running.

pavelsavara:
@lewing said Azdo spend is fine. 
We just disabled the same on the main CI, which was running it many times a day. 
This is low-traffic branch in comparison. It will be cheaper here.

The interesting question is if you guys would be willing and able to keep up with the maintenance of it. I guess with James' help

SingleAccretion:
> The interesting question is if you guys would be willing and able to keep up with the maintenance of it

That is my main concern. The logistics of having upstream CI downstream do not make sense to me. Let's consider what it implies:

1) If we have some sort of break in this pipeline, it is likely (let's say 90%+) an upstream issue, since we don't have (many) diffs in the libraries portion.
2) We ingest upstream infrequently and in bulk. It means every merge will have to deal with all these breaks _at once_. Merging is already full of mystery issues.
3) Fixing upstream issues should be done upstream. It means that unless the downstream patches are accepted upstream verbatim, they will be a constant source of conflicts. This is on top of needing to coordinate changes across two repositories.

IMO, if upstream is not happy with running WASI CI on each PR, it should run it on some periodical pipeline that is monitored (we have many Jit pipelines like that - I also understand that it is not a "great" solution).

Alternatively, expanding the set of libraries tests that NAOT-LLVM runs would also be completely fine with me.

@dotnet/nativeaot-llvm

yowl:
My 2 cents.  Merging is not the most exciting task so I personally wouldn't relish having more breaks there.  There seems, as an outsider, to be a bit of a disconnect with what Microsoft want.  James (from MS) is working on the WIT side, which is great, and there has been some discussion to doing the mono work first (as that is the supported Wasm backend) which is fine and understandable, but it's not my personal/selfish interest.  If someone else wants to keep the mono side working here, good.

jsturtevant:
I haven't had an opportunity to revisit this. I am going to close it for now

PR #3030, opened 2025-03-06
[RuntimeAsync] Merge from upstream

jakobbotsch:


VSadov:
@jakobbotsch   I've merged upstream roslyn (up to `bfe9d9f9059008d919d867fe5bdfabfe8b6ed69d`, which seems what runtime upstream is using) into our Roslyn branch and pushed a commit here to switch Roslyn ref to that. 

With updated C# compiler I can build `clr+libs` as well as `src/tests`

Preexisting tests seem to be mostly running. But Async tests are failing with some JIT asserts. Hopefully something obvious...

jakobbotsch:
Now all tests are passing again. Going to merge.

src/coreclr/jit/compiler.h
```diff
@@ -10933,6 +10933,13 @@ class Compiler
 
     bool compObjectStackAllocation()
     {
+        if (compIsAsync2())
```
VSadov, line 10936: In theory we could treat async calls as "escaping" object refs that are live after a call. That way object temps used in between cc calls could still be stack allocated. 
Probably not the high priority though.

VSadov (approved):
LGTM! Thanks!!

VSadov:
> Now all tests are passing again. Going to merge.

Passing for me too. All runtime tests - async and others.

src/coreclr/jit/compiler.h
jakobbotsch: What I'm leaning towards is that for async2 methods object stack allocation should only stack allocate objects where it can see all field accesses. If all field accesses are visible then stack allocation even across suspension points is ok (they will be captured the same as any other value). But this is quite far down the list of priorities at this point.

VSadov: Allowing stack allocated objects across suspension points would imply that upon resuming the references to such object would need to be adjusted as the object could be at a different address - we may even be on a different thread/stack. 

I think it is possible to support stackallocated objects with roughly same restrictions as in non-async case, but may not be worth the extra complexity.

jakobbotsch: It would only be allowed for cases where all accesses of the object are visible to the JIT. We would not allow it for cases like the one you describe.

Issue #3029, opened 2025-03-05
[Swift language features] Add equality comparison for Swift Structs


jkurdek:
By default there is no notion of equality between swift structs. Swift structs can conform to the `Equatable` protocol to allow for custom comparison. 

Swift types that conform to `Equatable` define a custom `==` operator. When projected into C#, there is currently no direct way to invoke that operator. Consequently, two instances that are equal in Swift might not be recognized as equal in C#.

## Proposed Approach
Generate an `EqualsValue` (or similar) method on the C# side that calls into Swift’s `==` operator for any type conforming to `Equatable`. This ensures C# has access to the same equality logic defined in Swift.

Optionally, `.Equals()` or the `==` operator could be overridden in generated C# code to delegate to this method, allowing a straightforward usage pattern.

PR #3031, opened 2025-03-07
[Swift Bindings] Add support for Equality protocol

jkurdek:
Fixes https://github.com/dotnet/runtimelab/issues/3029

This pull request introduces Equatable support for Swift structs and includes corresponding tests to ensure functionality. The key changes involve adding a new method writer for Equatable implementations, updating the protocol conformance handling, and adding tests for various struct types.

Currently the equality method is surfaced as a separate `SwiftEquals` static method. We should consider whether we should override `.Equals` and `==` in c#.

### DIFF
--- a/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
@@ -59,6 +59,7 @@ namespace BindingsGeneration
             var typeRecord = env.TypeDatabase.GetTypeRecordOrThrow(structDecl.SwiftTypeName);
 
             var ISwiftObjectMethodWriter = new ISwiftObjectMethodWriter(csWriter, env.TypeDatabase, moduleDecl, structDecl);
+            var SwiftEquatableMethodWriter = new SwiftEquatableMethodWriter(csWriter, structDecl);
 
             SwiftTypeInfo? swiftTypeInfo = typeRecord?.SwiftTypeInfo;
 
@@ -106,6 +107,8 @@ namespace BindingsGeneration
             }
             csWriter.WriteLine();
 
+            // Add Equatable support if the struct conforms to Equatable
+            SwiftEquatableMethodWriter.WriteSwiftEquatableImplementation();
             ISwiftObjectMethodWriter.WriteFrozenStructImplementation();
 
             base.HandleBaseDecl(csWriter, swiftWriter, structDecl.Types, conductor, env.TypeDatabase);
@@ -167,6 +170,7 @@ namespace BindingsGeneration
             var moduleDecl = structDecl.ModuleDecl ?? throw new ArgumentNullException(nameof(structDecl.ModuleDecl));
 
             var ISwiftObjectMethodWriter = new ISwiftObjectMethodWriter(csWriter, env.TypeDatabase, moduleDecl, structDecl);
+            var SwiftEquatableMethodWriter = new SwiftEquatableMethodWriter(csWriter, structDecl);
 
             csWriter.WriteLine($"public unsafe class {structDecl.Name} : IDisposable, {typeof(ISwiftObject).Name}");
             csWriter.WriteLine("{");
@@ -189,6 +193,8 @@ namespace BindingsGeneration
             WritePayloadSize(csWriter);
             WritePayload(csWriter);
 
+            // Add Equatable support if the struct conforms to Equatable
+            SwiftEquatableMethodWriter.WriteSwiftEquatableImplementation();
             ISwiftObjectMethodWriter.WriteNonFrozenStructImplementation();
 
             csWriter.WriteLine();
@@ -530,12 +536,21 @@ namespace BindingsGeneration
 
         private string GenerateGetProtocolConformanceDictionaryEntries()
         {
+            var crossModuleSupportedProtocols = new HashSet<string> // TODO: Remove this once we process multiple modules
+            {
+                { "Swift.Equatable"},
+            };
             var libPath = _typeDatabase.GetLibraryPath(_moduleDecl.Name);
             var entries = new List<string>();
             var protocolConformanceDescriptors = DemangledSymbolsRegister.Instance.GetData(libPath).ProtocolConformanceDescriptors;
 
-            foreach (var conformance in _structDecl.Conformances.Where(c => c.Protocol.Module == _moduleDecl.Name)) // Process only protocol conformances from current module for now
+            foreach (var conformance in _structDecl.Conformances)
             {
+                if (conformance.Protocol.Module != _moduleDecl.Name && !crossModuleSupportedProtocols.Contains(conformance.Protocol.ModuleQualifiedName))
+                {
+                    continue;
+                }
+
                 var protocol = NameProvider.GetInterfaceName(conformance.Protocol.Name);
                 var typeRecord = _typeDatabase.GetTypeRecordOrThrow(_structDecl.SwiftTypeName);
                 var protocolConformanceSymbol = protocolConformanceDescriptors.GetValueOrDefault((_structDecl.SwiftTypeName, conformance.Protocol)); // TODO: Get rid of TypeSpec https://github.com/dotnet/runtimelab/issues/2889
@@ -544,6 +559,34 @@ namespace BindingsGeneration
             }
 
             return string.Join(",\n", entries);
+        }
+    }
+
+    public class SwiftEquatableMethodWriter
+    {
+        private readonly IndentedTextWriter _writer;
+        private readonly StructDecl _structDecl;
+
+        public SwiftEquatableMethodWriter(CSharpWriter csWriter, StructDecl structDecl)
+        {
+            _writer = csWriter;
+            _structDecl = structDecl;
+        }
+
+        public void WriteSwiftEquatableImplementation()
+        {
+            if (_structDecl.Conformances.Any(c => c.Protocol.Name == "Equatable"))
+            {
+                var code = $$"""
+                public static bool SwiftEquals({{_structDecl.Name}} a, {{_structDecl.Name}} b)
+                {
+                    return Swift.Runtime.SwiftEquatable.Equals(a, b);
+                }
+                """;
+
+                _writer.WriteLines(code);
+                _writer.WriteLine();
+            }
         }
     }
 
--- a/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
+++ b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
@@ -74,8 +74,16 @@ namespace BindingsGeneration
                 }
             }
 
-            string interfaces = implementsEquatable ? $", ISwiftEquatable, IEquatable<{structDecl.Name}>" : "";
-            csWriter.WriteLine($"public unsafe struct {structDecl.Name} : {typeof(ISwiftObject).Name}{interfaces} {{");
+            var interfaces = new List<string> {
+                typeof(ISwiftObject).Name,
+            };
+            if (implementsEquatable)
+            {
+                interfaces.Add($"IEquatable<{structDecl.Name}>");
+            }
+
+            csWriter.WriteLine($"public unsafe struct {structDecl.Name} : {string.Join(", ", interfaces)}");
+            csWriter.WriteLine("{");
             csWriter.Indent++;
 
             csWriter.WriteLine(@"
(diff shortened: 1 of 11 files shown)


src/Swift.Runtime/src/Swift/SwiftEquatable.cs
```diff
@@ -0,0 +1,75 @@
...
+            ProtocolWitnessTable equatableProtocolWitnessTable);
+
+        /// <summary>
+        /// Compares two objects using Swift's Equatable protocol.
+        /// </summary>
+        /// <typeparam name="T">Type that implements IEquatable&lt;T&gt;</typeparam>
+        /// <param name="lhs">Left-hand side object</param>
+        /// <param name="rhs">Right-hand side object</param>
+        /// <returns>True if the objects are equal according to Swift's equality</returns>
+        public static unsafe bool Equals<T>(T lhs, T rhs)
```
vitek-karas, line 33: Should we add a generic constrain that `T` must implment `ISwiftEquatable`?
We requrie the `T` to provide PWT to `ISwiftEquatable` so it's close to that already anyway.

vitek-karas (commented):
Two more questions:

1) Why don't have the Equals method on the `ISwiftEquatable` interface and project it there? Currently we're adding a public method, which is not on the original struct, so it "pollutes" the API on the C# side. We could explicitely implement the interface method instead.
2) I think we should override ==/Equals on structs which implement equatable - it's is the standard way to compare objects in C#. I actually don't even know what the default implementation does for structs (byte compare???), but it's probably wrong for projected structs.

jkotas:
> I think we should override ==/Equals on structs which implement equatable - it's is the standard way to compare objects in C#.

I agree that it makes sense to override Equals. Also, implement `IEquatable<T>` that's also part of the standard pattern. I am wondering whether you need `ISwiftEquatable` if you implement `IEquatable<T>`.

`Equals` should match `GetHashCode`. How does Swift deal with hash codes? If the hashcode implementation is not available, you may want to override GetHashCode to throw.

> what the default implementation does for structs (byte compare???), but it's probably wrong for projected structs.

Yes, it is wrong and very slow.

jkurdek:
>I agree that it makes sense to override Equals. Also, implement IEquatable<T> that's also part of the standard pattern.

By default swift structs cannot be compared if they do not implement `Equatable` - I guess we could just go and make the appropriate methods on csharp side throw?

> I am wondering whether you need ISwiftEquatable if you implement IEquatable<T>

In the future we might expose some of the type constraints to the user. Then we probably need to differentiate between `IEquatable<T>` and `SwiftEquatable` - we cant obtain a PWT with respect to ``IEquatable<T>`


>Equals should match GetHashCode. How does Swift deal with hash codes? If the hashcode implementation is not available, you may want to override GetHashCode to throw.

In Swift there are two protocols - `Equatable` and `Hashable`. Types can choose to implement any combination of them.

src/Swift.Runtime/src/Swift/SwiftEquatable.cs
jkurdek: I do not think we emit interface conformance as of now. That is a type conforming to SwiftEquitable will not have `: ISwiftEquitable` on its definition, making it fail the compile time check. 

Since `Swift.Int` implements `Equatable` this brings us back to the boxing discussion

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -59,6 +59,7 @@ public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment en
             var typeRecord = env.TypeDatabase.GetTypeRecordOrThrow(structDecl.SwiftTypeName);
 
             var ISwiftObjectMethodWriter = new ISwiftObjectMethodWriter(csWriter, env.TypeDatabase, moduleDecl, structDecl);
+            var SwiftEquatableMethodWriter = new SwiftEquatableMethodWriter(csWriter, structDecl);
```
matouskozak: Are we planning to have a dedicated writer for each protocol that adds method to the implementing type or is this a temporary solution before the Emitter redesign?

vitek-karas:
> By default swift structs cannot be compared if they do not implement Equatable - I guess we could just go and make the appropriate methods on csharp side throw?

For structs which don't implement `Equatable`, we would just not generate the methods. The implemented protocols are part of the public ABI, so if they change it would make sense to change the C# projection for such type.

> we cant obtain a PWT with respect to ``IEquatable`

I think the suggestion is to treat `IEquatable` as the C# projeciton of `Equatable` from swift? So there would not be any `ISwiftEquatable` - instead the PWT retrieval code would understand that asking for `IEquatable` means to get `Equatable` PWT. Doing this for classes/structs is effectively impossible, but it might be doable for interfaces/protocols.

The Swift `Equatable` is effectively generic as well - it uses the `Self` type parameter. So for our purposes, we could treat only `IEquatable<TheAffectedType>` as the projection and ignore all other instantiations. But that's next level :-)

src/Swift.Runtime/src/Swift/SwiftEquatable.cs
```diff
@@ -0,0 +1,75 @@
...
+            var metadata = TypeMetadata.GetTypeMetadataOrThrow<T>();
+            var equatablePwt = ProtocolWitnessTable.GetOrThrow<T, ISwiftEquatable>();
+
+            IntPtr lhsPayload = IntPtr.Zero;
+            IntPtr rhsPayload = IntPtr.Zero;
+
+            try
+            {
+                lhsPayload = (IntPtr)NativeMemory.Alloc(metadata.Size);
+                rhsPayload = (IntPtr)NativeMemory.Alloc(metadata.Size);
```
matouskozak, line 49: Is it possible in Swift to compare different types for equality?

jkurdek: I dont think you can achieve that without some conversion to common type first

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
jkurdek: I think that having the emission of methods on common types grouped by protocols seems reasonable. I can imagine them going into multiple files. Do you see any alternative designs?

jkurdek:
>For structs which don't implement Equatable, we would just not generate the methods. The implemented protocols are part of the public ABI, so if they change it would make sense to change the C# projection for such type.

I think the problem is that by default swift doesnt allow comparing structs. In csharp we project some of them as structs and some as classes. If the projection is a value type the `==` in csharp will compare all the fields, if the projection is a reference type it will go and compare the references - neither of those two match the swift behaviour. I think that making "throw" a default in such cases would be a good alternative. 

>I think the suggestion is to treat IEquatable as the C# projeciton of Equatable from swift? So there would not be any ISwiftEquatable - instead the PWT retrieval code would understand that asking for IEquatable means to get Equatable PWT. Doing this for classes/structs is effectively impossible, but it might be doable for interfaces/protocols.

Assuming we do surface type constraints one day:

```csharp
static bool SomeSwiftProjectedFunction<T>(T first, T second) where T : IEquatable<T>
```

This function would accept any type which conforms to IEquatable<T> - including csharp-only types which have nothing to do with Swift (and thus we cannot obtain PWT for them). 


On the other hand:

```csharp
static bool SomeSwiftProjectedFunction<T>(T first, T second) where T : SwiftEquatable<T>
```

this would force the user to pass arguments which are guaranteed to have a PWT

kotlarmilos:
> I think the problem is that by default swift doesnt allow comparing structs. In csharp we project some of them as structs and some as classes. If the projection is a value type the == in csharp will compare all the fields, if the projection is a reference type it will go and compare the references - neither of those two match the swift behaviour. I think that making "throw" a default in such cases would be a good alternative.

We have the same problem with implicit copy, where we assume users will not perform such assignments. In this case, does this mean that if a type conforms to IEquatable, we call into Swift; otherwise, we throw?

> Assuming we do surface type constraints one day:
> static bool SomeSwiftProjectedFunction<T>(T first, T second) where T : IEquatable<T>

Is it possible to enforce that the parameters also implement ISwiftObject?
```csharp
static bool SomeSwiftProjectedFunction<T>(T first, T second) where T : IEquatable<T>, ISwiftObject
```

jkurdek:
>We have the same problem with implicit copy, where we assume users will not perform such assignments. In this case, does this mean that if a type conforms to IEquatable, we call into Swift; otherwise, we throw?

For csharp-side structs (frozen structs) i would always implement IEquatable and just throw if the type does not conform to Equatable protocol on the Swift side. This way we match the Swifts equality semantics. 

For csharp-side classes (non-frozen structs) - I'm not sure. We definitely need to expose a way to call into Swift's `Equals`. But on the other hand I'm not sure whether we should remove the ability to compare two classes by reference. Maybe we need both? What I would like to avoid is non-frozen structs which do not implement Equatable to be compared by reference and the one's which do through Swift's `Equals`. This would be vey confusing.

jkurdek:
>Is it possible to enforce that the parameters also implement ISwiftObject?

I think the problem is that we cannot guarantee that every implementation of SwiftEquals protocol will exist on swift side (ISwiftObject). We can have a type which exists in csharp world only and implements `SwiftEquatable` through some base class indirection.

kotlarmilos:
>  We definitely need to expose a way to call into Swift's Equals. But on the other hand I'm not sure whether we should remove the ability to compare two classes by reference.

If a type is Swift struct projected as C# class, then it should be safe to override the operator as it is Swift struct and users expect a value comparison. If a type is Swift class, again, it should be safe to override the operator and use it as Swift type projected into C#. 

If the projected type is considered as C#-first type, then the above is not true.

Would a separate interface as you initially suggested resolve this problem?

> I think the problem is that we cannot guarantee that every implementation of SwiftEquals protocol will exist on swift side (ISwiftObject). We can have a type which exists in csharp world only and implements SwiftEquatable through some base class indirection.

Does this mean that the type is used on C# side only?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
matouskozak: I think it is good for now, we might reconsider this approach in the future when we support more protocols. 

> Do you see any alternative designs?

Possibly a dedicated component for writing Protocol methods that would encapsulate all supported protocols.

src/Swift.Runtime/src/Swift/SwiftEquatable.cs
matouskozak: Do you know if Swift compiler generates these conversions automatically or is it a user responsibility?

Actually, I just tried simple example with:

```swift
let x: Double = 4.3;
let y: Float = 2.3;

if (x == y) {
...
}
```
and it generates a compiler warning that it requires conversion to common type.

vitek-karas:
> I think the problem is that we cannot guarantee that every implementation of SwiftEquals protocol will exist on swift side

If there's a type which implement `ISwiftEquatable`, but not `ISwiftObject` we won't be able to get a PWT for it anyway - so calling `SwfitEquals` on it will fail, right?

jkurdek:
> Does this mean that the type is used on C# side only?

I meant the situation when a csharp only class / struct implements a Swift side protocol.

>If there's a type which implement ISwiftEquatable, but not ISwiftObject we won't be able to get a PWT for it anyway - so calling SwfitEquals on it will fail, right?

I think it depends on how we decide to solve this kind of scenarios. But after some though I do think that this is a very future problem.

I spend some more time thinking about it. I got to the following conclusion. We should by default override the .Equals and `==` on every projected struct and make it throw (classes can use default equality).  If and only if the swift struct implements the `Equatable` protocol we then implement the equality methods routing the calls into swift. I agree that we should then also implement `IEquatable<T>` on those types.

After the discussion on this PR I do not see any strong reason to have `SwiftEquatable` and `IEquatable<T>` separate. I think this is more of a convention thing now - same question will apply to `Collection` protocol (`IEnumerable`). There are some differences between those pairs, do they justify being separate interfaces on csharp side?  I think we can choose any option now and review our choices when we get to the point that we start adding more protocol support.

vitek-karas:
Thanks a lot Jeremi - this makes sense.
Once we do support "overriding" the operations then we will probably have to introduce the explicit swift projected interfaces, as otherwise it would be difficult to tell if the tool should project the implementation to swift or not. But that will only come once we either support PWTs for nonprojected types, or once we support inheritance.

jkurdek:
@kotlarmilos it seems that GeneratedBindings/Swift.Bindings.Experimental.csproj has different nullable settings for maccatalyst and osx?

kotlarmilos:
Try to set `<Nullable>disable</Nullable>` in the `generate.sh`. Tracking issue: https://github.com/dotnet/runtimelab/issues/3046

kotlarmilos (approved):
Thanks!

PR #3032, opened 2025-03-08
[RuntimeAsync] Some VM cleanups, TODOs, renames and better comments.

VSadov:
Mostly renames and comments.

Also settles the terminology for the pairs of methods that we create. Let's call them "variants".

### DIFF
--- a/src/coreclr/vm/clrtocomcall.cpp
+++ b/src/coreclr/vm/clrtocomcall.cpp
@@ -230,7 +230,7 @@ I4ARRAYREF SetUpWrapperInfo(MethodDesc *pMD)
 
         GCX_PREEMP();
 
-        if (pMD->IsAsyncHelperMethod())
+        if (pMD->IsAsync2Method())
         {
             ThrowHR(COR_E_NOTSUPPORTED);
         }
@@ -509,7 +509,7 @@ UINT32 CLRToCOMLateBoundWorker(
     LPCUTF8 strMemberName;
     ULONG uSemantic;
 
-    if (pItfMD->IsAsyncHelperMethod())
+    if (pItfMD->IsAsync2Method())
     {
         ThrowHR(COR_E_NOTSUPPORTED);
     }
--- a/src/coreclr/vm/clsload.cpp
+++ b/src/coreclr/vm/clsload.cpp
@@ -2801,7 +2801,7 @@ TypeHandle ClassLoader::PublishType(const TypeKey *pTypeKey, TypeHandle typeHnd)
         {
             MethodDesc * pMD = it.GetMethodDesc();
             CONSISTENCY_CHECK(pMD != NULL && pMD->GetMethodTable() == pMT);
-            if (!pMD->IsUnboxingStub() && !pMD->IsAsync2HelperMethod())
+            if (!pMD->IsUnboxingStub() && !pMD->IsAsync2VariantMethod())
             {
                 pModule->EnsuredStoreMethodDef(pMD->GetMemberDef(), pMD);
             }
--- a/src/coreclr/vm/clsload.cpp
+++ b/src/coreclr/vm/clsload.cpp
@@ -2801,7 +2801,7 @@ TypeHandle ClassLoader::PublishType(const TypeKey *pTypeKey, TypeHandle typeHnd)
         {
             MethodDesc * pMD = it.GetMethodDesc();
             CONSISTENCY_CHECK(pMD != NULL && pMD->GetMethodTable() == pMT);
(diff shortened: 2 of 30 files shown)


src/coreclr/vm/method.hpp
```diff
@@ -58,29 +58,54 @@ EXTERN_C VOID STDCALL NDirectImportThunk();
...
-    AsyncImplHelper,
-
-    // Synthetic Async2 method that forwards to a TaskReturning method
-    AsyncThunkHelper,
-
-    // Actual IL method that is explicitly declared as Async2 and thus compiled into a state machine.
-    // Such methods do not get Async thunks and can only be called from another Async2 method using Async2 call convention.
-    // This is used in a few infrastructure methods like `Await`
-    AsyncImplExplicit,
+    //=============================================================
```
VSadov, line 73: Trying to explain the naming here.

src/coreclr/vm/clrtocomcall.cpp
```diff
@@ -230,7 +230,7 @@ I4ARRAYREF SetUpWrapperInfo(MethodDesc *pMD)
 
         GCX_PREEMP();
 
-        if (pMD->IsAsyncHelperMethod())
+        if (pMD->IsAsync2Method())
```
VSadov, line 233: We use `IsAsync2Method` if we want to special case any Async2 method, most likely because the scenario cannot handle them. This would include the `Await` intrinsic, for example, if it somehow makes here.

src/coreclr/vm/clsload.cpp
```diff
@@ -2801,7 +2801,7 @@ TypeHandle ClassLoader::PublishType(const TypeKey *pTypeKey, TypeHandle typeHnd)
         {
             MethodDesc * pMD = it.GetMethodDesc();
             CONSISTENCY_CHECK(pMD != NULL && pMD->GetMethodTable() == pMT);
-            if (!pMD->IsUnboxingStub() && !pMD->IsAsyncHelperMethod())
+            if (!pMD->IsUnboxingStub() && !pMD->IsAsync2VariantMethod())
```
VSadov, line 2804: We use `IsAsync2VariantMethod` to special case Async2 variant methods that were introduced in addition to what was defined in metadata. This is for scenarios that only want to deal with "real/defined" methods. 

As an example, if `Await` makes here - it is not a special case, it is a real defined method.

src/coreclr/vm/method.hpp
VSadov, line 73: Trying to explain the naming here.

src/coreclr/vm/method.hpp
```diff
@@ -1824,38 +1849,54 @@ class MethodDesc
...
     {
         LIMITED_METHOD_DAC_CONTRACT;
         if (!HasAsyncMethodData())
             return false;
         auto asyncKind = GetAddrOfAsyncMethodData()->kind;
-        return asyncKind == AsyncMethodKind::AsyncThunkHelper ||
-            asyncKind == AsyncMethodKind::AsyncImplHelper;
+        return asyncKind == AsyncMethodKind::Async2VariantThunk ||
+            asyncKind == AsyncMethodKind::Async2VariantImpl ||
+            asyncKind == AsyncMethodKind::Async2ExplicitImpl;
```
jakobbotsch, line 1862: Why did we end up with another kind here?

jakobbotsch (approved):
LGTM

src/coreclr/vm/method.hpp
VSadov: What was `IsAsyncHelperMethod` has become `IsAsync2VariantMethod`, it still checks for two kinds.

I moved `IsAsync2Method` a bit more upfront here, so if someone reads the code sequentially, they will get introduced to `IsAsync2Method` method first, that is simpler thing to understand than `IsAsync2VariantMethod`.

As a result the diff looks like `IsAsync2Method` now checks for one more kind, but that actually did not change.

src/coreclr/vm/method.hpp
```diff
@@ -1824,38 +1849,54 @@ class MethodDesc
...
     }
 
-    inline bool IsTaskReturningMethod() const
+    // Is this an Async2-callable variant method?
+    // If yes, the method has another non-Async2 variant.
+    inline bool IsAsync2VariantMethod() const
```
VSadov, line 1867: Most uses of this, I think, are actually about "is this the variant that was not in metadata?". 
However, in the current IL API it always happens to be the Async2 one, so "is this the variant that is async2?"  is basically the same question.

VSadov:
Thanks!

Issue #2665, opened 2024-08-22
[Projection tooling] Implement Demangler query methods


stephen-hawley:
## Tasks

- [ ] Import a complete TypeSpec based signature for methods
- [ ] Given a type and protocol return the protocol witness table
- [x] Given a type and protocol return the conformance
- [ ] Given an async method return the async pointer
- [x] Given a type return the metadata accessor
- [ ] Given a func find the dispatch thunk for it

Issue #2845, opened 2024-12-03
[Projection tooling] Refactor projection tooling to be extensible and ensure it outputs only supported types


kotlarmilos:
## Tasks

 - [x] https://github.com/dotnet/runtimelab/issues/2844
 - [x] https://github.com/dotnet/runtimelab/issues/2799
 - [ ] Refactor filtering handler
 - [ ] Refactor intermediate representation
 - [ ] Refactor marshalling handlers
 - [ ] Refactor emitters
 - [ ] Code improvements
   - [ ] Improve _protocolConformanceSymbols to use a switch statement for startup improvements
   - [ ] https://github.com/dotnet/runtimelab/issues/2868
   - [ ] Generate _payloadSize only when it is used to improve startup cost
   - [ ] Prefix private static fields with s_
   - [ ] https://github.com/dotnet/runtimelab/issues/2997
   - [ ] https://github.com/dotnet/runtimelab/issues/2802
   - [ ] https://github.com/dotnet/runtimelab/issues/2925
   - [ ] https://github.com/dotnet/runtimelab/issues/2803
   - [ ] https://github.com/dotnet/runtimelab/issues/2928
   - [ ] Reduce redundant fields in the decls and type records

> [!NOTE]  
> This issue tracks progress and will be expanded with more details. If specific tasks for this issue are not clear, this means that the task is not well defined yet and needs attention.

PR #3034, opened 2025-03-11
[swift] Use TBD files to retrieve symbols missing from abi.json

matouskozak:
## Description
This PR integrates TBD file support to retrieve symbol information missing from the abi.json file. The reasoning why to use TBD files instead of dylib files is included as part of the [documentation](https://github.com/dotnet/runtimelab/pull/3034/files#diff-dc81b6dc81d0245f5e3185c444d2e0c12be54e14a2935775665d2d5ce05fcd3c). To limit the scope of this work, I only implemented the TBD parser for YAML format (TBD v.1-4). Future work will include parser of TBDs in JSON format (v.5 +).

## Key changes include
- Introducing a new TBD parser (TbdParser) for YAML format. 
- Updating demangling and ABI parsing logic to use TBD-derived metadata accessors and protocol conformance descriptors (stored inside `StructDecl` for now).
- Logger prototype (currently only used in TbdParser domain).

## Validation
- Added simple unit tests for TbdParser
- Validity of retrieved symbols from TbdParser verified through the contemporary tests for scenarios where metadata accessors or protocol conformance descriptors were previously hardcoded or retrieved from dylib file.

## Future work
- Remove MachO logic from the tooling https://github.com/dotnet/runtimelab/issues/3047
- Port the new logger to the rest of the tooling https://github.com/dotnet/runtimelab/issues/2973
- Implement a parser of TBD files in JSON format https://github.com/dotnet/runtimelab/issues/3048
- Introduce more comprehensive testing of abi.json + TbdParser + demangler workflow. https://github.com/dotnet/runtimelab/issues/3049

---

Contributes to:
- https://github.com/dotnet/runtimelab/issues/2665 
    - > Given a type and protocol return the conformance
    - > Given a type return the metadata accessor
- https://github.com/dotnet/runtimelab/issues/2918
    - Partially 
        - > Read symbols during the post processing step and store them in the type database 
    - > Replace hardcoded metadata accessors with symbols from the library. 

#### TODO:
- [x] Open issues for future work


src/Swift.Bindings/src/Demangler/TbdParser/Logging/ConsoleLogger.cs
```diff
@@ -0,0 +1,64 @@
+using System;
+
+namespace TbdParser.Logging
+{
+    /// <summary>
+    /// Simple console logger implementation
+    /// </summary>
+    public class ConsoleLogger : ILogger
```
kotlarmilos, line 11: Can we add this to be project-wide? There is -v flag but it doesn't implement any logger.

src/Swift.Bindings/src/Model/TypeDecl/StructDecl.cs
```diff
@@ -16,5 +16,10 @@ public sealed record StructDecl : TypeDecl
         /// Protocol conformances.
         /// </summary>
         public required List<TypeConformance> Conformances { get; set; }
+
+        /// <summary>
+        /// Metadata accessor.
+        /// </summary>
+        public required string MetadataAccessor { get; set; }
```
kotlarmilos, line 23: Should this be part of the type database instead?

src/docs/retrieving-symbols-outside-abi-json.md
```diff
@@ -0,0 +1,94 @@
+To generate bindings between .NET and Swift symbols such as metadata accessors, protocol conformance descriptors, or async function pointers are needed.
```
kotlarmilos, line 1: Please add a reference in the docs to this page.

src/Swift.Bindings/src/Demangler/TbdParser/Logging/ConsoleLogger.cs
matouskozak: > There is -v flag but it doesn't implement any logger.

Yes, currently we are passing `int verbosity` around and based on the value we print to the console. 

If this format would be appropriate, there shouldn't be an issue to port it for the entire codebase. I would probably make it a separate PR to keep this PR smaller as it already has ~1000 line changes.

src/Swift.Bindings/src/Model/TypeDecl/StructDecl.cs
matouskozak: I put it inside `StructDecl` as it was a convenient place for storing it to be able to pass the value from SwiftABIParser to ModuleProcessor and further. Additionally, at the moment, we are relying more on TypeDecl representations rather than TypeRecord when emitting the bindings (https://github.com/dotnet/runtimelab/blob/d612995031c74ccab34c9199567eff71e694d76b/src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs#L385) so I decided to leave it as part of StructDecl.

We duplicate the value to type database as well https://github.com/dotnet/runtimelab/blob/d612995031c74ccab34c9199567eff71e694d76b/src/Swift.Bindings/src/Parser/ModuleProcessor.cs#L263 but I didn't find a record of the use anywhere in the tooling.

I think these inconsistencies will likely get resolved after the emitter re-design where we should rely on some IR to emit the bindings as proposed by @jkurdek

kotlarmilos: Ok, added to https://github.com/dotnet/runtimelab/issues/2845

jkurdek: I think type database should only consist of things we need for projecting references to a type. If something is needed to project a type declaration only then it should live elsewhere

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
```diff
@@ -305,8 +318,21 @@ private TypeConformance HandleConformance(Node node, SwiftTypeName typeName)
...
+
+            try
+            {
+                protocolConformanceDescriptor = _demangledTbd.GetProtocolConformanceDescriptor(typeName, protocolName);
+            }
+            catch (Exception e)
+            {
+                // TODO: We currently do not support retrieving cross-module protocol conformance descriptors.
+                if (_verbose > 0)
+                    Console.WriteLine($"Error while getting protocol conformance descriptor for '{typeName}' and protocol '{protocolName}': {e.Message}");
```
jkurdek, line 333: So getting a protocol conformance descriptor for struct Foo defined in module A and `Equatable` protocol defined in `SwiftCore` will fail now?

src/Swift.Bindings/src/Demangler/DemanglingResults.cs
```diff
@@ -1,8 +1,10 @@
+using BindingsGeneration;
 using BindingsGeneration.Demangling;
+using TbdParsing;
 using Xamarin;
 
 /// <summary>
-/// A class to contain results from demangling the set of symbols in a MachO file.
+/// A class to contain results from demangling the set of symbols in a MachO or TBD file.
```
jkurdek, line 7: Since we might be retiring MachO in the future (in case of obtaining dylibs become impossible), it might be better to keep the logic in separate files so that the process is easier in the future

src/Swift.Bindings/src/Demangler/TbdParser/Logging/ConsoleLogger.cs
jkurdek: Could we add this to a tracking issue? I also think this would be great to have on the entire codebase!

rolfbjarne (approved):

src/Swift.Bindings/src/Parser/SwiftABIParser.cs
matouskozak: That's a great question. When I wrote the comment, I thought that is the case but further investigation releaved that it is more complicated than this. I'll update the comment.

I found that there are some "implicit" conformances, e.g. for the `GenericTets.FrozenStruct`:
```
Error while getting protocol conformance descriptor for 'GenericTests.FrozenStruct' and protocol 'Swift.Copyable': Protocol conformance descriptor not found for type 'GenericTests.FrozenStruct' and protocol 'Swift.Copyable'.
Error while getting protocol conformance descriptor for 'GenericTests.FrozenStruct' and protocol 'Swift.Escapable': Protocol conformance descriptor not found for type 'GenericTests.FrozenStruct' and protocol 'Swift.Escapable'.
Error while getting protocol conformance descriptor for 'GenericTests.FrozenStruct' and protocol 'Swift.Sendable': Protocol conformance descriptor not found for type 'GenericTests.FrozenStruct' and protocol 'Swift.Sendable'.
```
While the definition doesn't specify any protocol: `public struct FrozenStruct`

If you would specify `public struct FrozenStruct : Equatable`, the protocol conformance descriptor for `Equatable` will be present: `_$s12GenericTests12FrozenStructVSQAAMc => protocol conformance descriptor for GenericTests.FrozenStruct : Swift.Equatable in GenericTests`.

src/Swift.Bindings/src/Demangler/DemanglingResults.cs
matouskozak: In terms of retiring MachO, there isn't a hard dependency of `DemanglingResults` class on the MachO code. We would just remove the factory methods that use MachO https://github.com/dotnet/runtimelab/blob/4f8fb72dca5808a68442572c38e710548cf88116/src/Swift.Bindings/src/Demangler/DemanglingResults.cs#L60-L108 . The logic for accessing demagled results is invariant to the source. 

Do you suggest creating a separate layer for demangled results results coming from TBD files? I think that would be beneficial if we would be using both MachO and TBDParser in the future as we could keep the factory methods separate while maintaining the logic for accessing demagled results in one place.

kotlarmilos (approved):
Thanks, great work!

src/Swift.Bindings/src/Demangler/DemanglingResults.cs
jkurdek: Hmm, how likely are we to be using both MachO and TBD parser? Arent they serving the same purpose? Anyway, I do not have a hard opinion on that. Maybe it is beneficial to leave it as it is to keep things simple.

BTW: Just noticed that this class is not part of any namespace.

matouskozak: > Hmm, how likely are we to be using both MachO and TBD parser? Arent they serving the same purpose?

I very much hope so that we won't need that. Yes, they serve the same purpose. I'm a bit unsure regarding the implicit conformances as described in https://github.com/dotnet/runtimelab/pull/3034#discussion_r1995828187 but I haven't found them in dylib either.

> Maybe it is beneficial to leave it as it is to keep things simple. 

I can just open a follow-up to remove MachO and it should be done https://github.com/dotnet/runtimelab/issues/3047

> BTW: Just noticed that this class is not part of any namespace.

Good catch. Added.

jkurdek (approved):
Looks great! Thanks!

Issue #3018, opened 2025-02-24
[Swift ABI compatibility] Implement Swift payload as SafeHandler to decouple IDisposable from the bindings


kotlarmilos:
## Tasks
 - [x] Implement Swift payload as a SafeHandler
 - [x] Implement `Destroy` call in the Dispose method
 - [x] Add functional tests

Issue #2985, opened 2025-02-10
[Swift Language Features] Generate correct Dispose for non-frozen structs and generic types


jkurdek:
Non frozen structs and generic types are projected as classes. We need to manualy release all the resources in the dispose method. We need to learn what needs disposing and genearte an appropriate Dispose method.

> By projecting non-frozen types and generics as classes, we allow them to have destructors. The problem is that without knowing their memory layout, we don’t know what needs to be released in the destructor.
> 
> For example, consider `Swift.String`:
> ```
>   store ptr %2, ptr %str.debug._guts._object._object, align 8, !dbg !136
>   call void @swift_bridgeObjectRelease(ptr %2) #5, !dbg !137
> ```
> 
> The value witness table might actually resolve this without requiring explicit knowledge of the layout:
> https://github.com/swiftlang/swift/blob/b185bf9562f007c37acdfcdbe42eebfd96bd9973/include/swift/ABI/ValueWitness.def#L114
> 
> This is something we will investigate further. 

 _Originally posted by @kotlarmilos in [#2970](https://github.com/dotnet/runtimelab/issues/2970#issuecomment-2647914808)_

Issue #2975, opened 2025-02-03
[Swift language features] Revisit thread safety of ref counting in projections


vitek-karas:
from: https://github.com/dotnet/runtimelab/pull/2964#discussion_r1935938434

Dispose in our projections should be resilient to parallel calls from multiple threads - this means it needs to correctly handle checking for already released references.

rolfbjarne:
> Dispose in our projections should be resilient to parallel calls from multiple threads

Making only `Dispose` thread-safe doesn't change much, because then any other instance method would still have a race condition.

If thread A does:

```cs
var obj = new SwiftObject ();
Console.WriteLine (obj.GetSwiftDescription ());
obj.Dispose ();
```

and thread B the same thing:


```cs
var obj = new SwiftObject ();
Console.WriteLine (obj.GetSwiftDescription ());
obj.Dispose ();
```

then even if `Dispose` itself is thread-safe, any of the two calls to `GetSwiftDescription()` might access freed memory.

One way around this could be to use a read-write lock, where each usage of the internal handle acquires the read lock, and then Dispose acquires the write lock. This way Dispose won't run until nobody else is using the internal handle.

The downside is that every instance method would need locking, slowing down the bindings.

FWWI I can't remember hearing about any other API binding/projection that does this kind of safety guarantee / locking.

jkotas:
> I can't remember hearing about any other API binding/projection that does this kind of safety guarantee / locking.

The base class library APIs like Socket or FileStream do provide this kind of safety guarantee / locking. I agree that it comes with extra overhead. It is typically done using SafeHandle.

rolfbjarne:
> It is typically done using SafeHandle.

Ah I see, so it's basically a refcounted handle. That would work.

PR #3038, opened 2025-03-12
[Swift Bindings] Project value types with heap-allocated properties as C# classes

kotlarmilos:
## Description

This PR updates marshalling to project Swift value types that require explicit memory handling as C# classes. Heap-allocated Swift types and value types with heap allocated properties are projected as C# classes. The projected C# classes implement `ValueWitnessTable->Destroy` in their finalizer to ensure proper memory deallocation.

## Changes

- Extended `TypeRecord` with `Kind` and `Flag` enums for marshalling
- Marshalling is made explicit by replacing `StructIsMarshalledAsCSStruct` and `ArgumentIsMarshalledAsCSStruct` with `IsFrozen` and `IsHeapAllocated`
- Removed redundant `MarshalToSwift` copy for frozen bound generics in P/Invoke calls
- Replaced local allocations for generics with `stackalloc` for implicit memory management and perf improvements
- Updated manual bindings to implement correct `Dispose` handling

## Validation

Integration tests and  manual bindings tests are expanded to verify memory deallocation and parameter ref counting.

## Out of Scope

- `inout` parameters are not yet supported, so memory handling is not implemented.
- `UnmanagedCallersOnly` is not yet supported, so memory handling is not implemented
- Debug checks for memory handling are not implemented: https://github.com/dotnet/runtimelab/issues/3037
- Projections could be simplified by removing the IDisposable pattern and implementing SafeHandlers: https://github.com/dotnet/runtimelab/issues/3018

Fixes https://github.com/dotnet/runtimelab/issues/3035 https://github.com/dotnet/runtimelab/issues/3035 https://github.com/dotnet/runtimelab/issues/3036 https://github.com/dotnet/runtimelab/issues/2985


src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -114,6 +151,54 @@ public void Emit(CSharpWriter csWriter, SwiftWriter swiftWriter, IEnvironment en
...
+            var text = $$"""
+            public void Dispose()
+            {
+                Dispose(true);
+                GC.SuppressFinalize(this);
+            }
+
+            protected virtual void Dispose(bool disposing)
+            {
+                if (Interlocked.CompareExchange(ref _disposed, 1, 0) == 0)
```
rolfbjarne: Does `Dispose` really have to be thread-safe? It seems that if two threads are accessing the same instance at the same time, then they're already in trouble, since it seems unlikely both threads will _only_ be calling `Dispose` (and if any of those two thread are also calling instance members, then that's not thread-safe even if `Dispose` by itself is).

kotlarmilos: Calling instance methods doesn't change ref counts, but calling Dispose does -- that's the main difference. Also, Swift doesn't have mechanism to prevent from calling Destroy more than once. 

It is true that we haven't run into a situation where that becomes a problem.

rolfbjarne: > Calling instance methods doesn't change ref counts, but calling Dispose does -- that's the main difference.

Right, but code like this isn't safe anyways:

```cs
// thread 1
Console.WriteLine (obj.ToString ());
obj.Dispose ();

// thread 2
obj.Dispose ();
```

Because if thread 2's `Dispose` is executed when thread 1 is executing `ToString`, a crash will occur. The fact that the `Dispose` methods are thread-safe doesn't change this.

kotlarmilos: Yes, you are right.

kotlarmilos: What are the implications? I see these two:
 - Perf implications
 - Not a valid scenario

vitek-karas: There is a slight difference - if somebody calls an instance method after Dispose the payload pointer will be null which is typically a well handled error scenario - so the thing will throw but otherwise leave the state uncorrupted (most likely).
If we free the same instance twice, that could lead to memory corruption and hard crash - depends on Swift's implementation if it has protections in place for this or not.

rolfbjarne: I agree that calling Dispose twice should be handled correctly, but my point is that the extra work of making it thread-safe seems like it's not actually fixing any real-world scenarios.

src/Swift.Bindings/src/TypeDatabase/TypeRecord.cs
```diff
@@ -5,6 +5,29 @@
...
+/// <summary>
+/// Represents a type kind.
+/// </summary>
+public enum TypeRecordKind
+{
+    Struct,
+    Enum,
+    Class,
+    Protocol,
+    Tuple
```
jkurdek: Do we need to place tuples in the type database? Arent they always constructed from other types?

src/Swift.Bindings/src/Marshaler/MarshallingHelpers.cs
```diff
@@ -27,15 +28,20 @@ public static bool MethodRequiresSwiftSelf(MethodEnvironment env)
             return true;
         }
 
-        public static bool StructIsMarshalledAsCSStruct(StructDecl decl)
+        public static bool IsTypeFrozen(TypeRecord typeRecord)
```
jkurdek, line 31: We might need to have a note somewhere that "Frozen" on the tooling side is not equal to "Frozen" on swift side (e.g. frozen struct containing a reference to a reference type).

Edit: I meant that the fact the a frozen type contains a reference type or a non-frozen struct does not make it "non-frozen" on Swift side. It is still "frozen" but has layout unknown at compile time. In the tooling I made a mistake of starting using "frozen" and "non-frozen" to mean projection as struct / class, which is loosly connected to the above.

AFAIK the rule is:
if the layout is known at compile time && no memory management is necessary: struct
Else: class

Since you are making the names more descriptive (which I really like!) I think it would be good to at least encode that information somewhere

src/Swift.Bindings/src/TypeDatabase/TypeRecord.cs
```diff
@@ -5,6 +5,29 @@
...
 
+/// <summary>
+/// Represents type's flags.
+/// </summary>
+[Flags]
+public enum TypeRecordFlags
+{
+    None = 0,
+    Frozen = 1 << 0,
+    HeapAllocated = 1 << 1,
```
jkurdek: What does it mean that a type is heap allocated? That it is allocated on heap in swift?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -1007,13 +1045,6 @@ private void EmitFinally(CSharpWriter csWriter)
         {
             csWriter.WriteLine("finally");
```
jkurdek: I guess we do not need the whole "finally block" then if all the allocations are done on stack.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
```diff
@@ -318,14 +319,17 @@ public void HandleReturnType()
...
+            if (_env.MethodDecl.IsAsync && !MarshallingHelpers.IsTypeFrozen(returnTypeRecord))
             {
                 SetReturnType("IntPtr");
                 return;
             }
 
-            var returnTypeRecord = _env.TypeDatabase.GetTypeRecordOrThrow(returnType.SwiftTypeSpec);
-            SetReturnType(returnTypeRecord.CSharpTypeName.FullyQualifiedName);
+            if (MarshallingHelpers.IsTypeHeapAllocated(returnTypeRecord))
+                SetReturnType(returnTypeRecord.CSharpTypeName.FullyQualifiedName + ".Buffer");
```
jkurdek, line 330: Can csharp types contain "."?

src/Swift.Bindings/src/TypeDatabase/TypeRecord.cs
kotlarmilos: Void is the only case and it could be special-handled.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: It isn’t allowed as part of an individual type’s identifier. Here, "." is used as a delimiter for the nested buffer type.

src/Swift.Bindings/src/TypeDatabase/TypeRecord.cs
kotlarmilos: HeapAllocated flag indicates the type is heap allocated (class) or contains a heap allocated property (struct with a ref property). (Added a comment)

src/Swift.Bindings/src/Marshaler/MarshallingHelpers.cs
kotlarmilos: Thanks, added as a comment.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
jkurdek: I'm not sure I understand `SetReturnType(...)` should take a valid csharp type which will be emitted into the code, am I missing something?

src/Swift.Bindings/src/TypeDatabase/TypeRecord.cs
jkurdek: I see, thanks! Could we encode this somehow into the name of this flag? Like `requiresMemoryManagement` or something similar?

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
kotlarmilos: I've reverted the thread-safety changes. Both points are valid -- the bindings should be thread-safe, and making only Dispose thread-safe does not guarantee overall thread-safety of the bindings. 

I recommend that we revisit this issue when we encounter real-world scenarios.

rolfbjarne (approved):

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/MethodHandler.cs
kotlarmilos: Consider the following scenario where a frozen Swift struct is projected as a C# class due to memory handling requirements. In Swift, we have:
```swift
public class RefType {
}

@frozen
public struct FrozenStructHeapAllocated {
    public var a: RefType
    public var b: Int32
}
```
This struct is projected into C# as:
```csharp
public class FrozenStructHeapAllocated {
  public struct Buffer {
     IntPtr _a;
     Int _b;
  }
  
  ~FrozenStructHeapAllocated() {
    // VWT->Destroy
  }
}
```

At the ABI boundary, the struct is enregistered as `FrozenStructHeapAllocated.Buffer`.

jkurdek: I see, thanks for explanation!

src/Swift.Runtime/src/Swift/SwiftEquatable.cs
```diff
@@ -40,31 +40,21 @@ public static unsafe bool Equals<T>(T lhs, T rhs)
             var metadata = TypeMetadata.GetTypeMetadataOrThrow<T>();
             var equatablePwt = ProtocolWitnessTable.GetOrThrow<T, IEquatable<T>>();
 
-            IntPtr lhsPayload = IntPtr.Zero;
-            IntPtr rhsPayload = IntPtr.Zero;
+            var lhsPayloadSpan = stackalloc byte[(int)metadata.Size];
+            IntPtr lhsPayload = (IntPtr)Unsafe.AsPointer(ref lhsPayloadSpan[0]);
```
jkotas:
```suggestion
            byte* lhsPayload = stackalloc byte[(int)metadata.Size];
```
Since you are not using the Span anyway, you can skip the ceremony and allocate `byte*` directly.

jkotas: (I am not commenting on all places.)

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
jkotas: Related to https://github.com/dotnet/runtimelab/issues/2975 . 

> we revisit this issue when we encounter real-world scenarios.

This is about how to use the bindings securely. You may want to think it through earlier rather than later. Once you encounter real-world scenarios related to security, it is typically too late.

kotlarmilos: Thanks, lets add them here or in a follow-up.

jkurdek (approved):

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
kotlarmilos: Follow-up https://github.com/dotnet/runtimelab/pull/3052

Issue #2932, opened 2025-01-14
[Swift shipping] Add end-to-end tests for the StoreKit2 framework


kotlarmilos:
## Tasks
 - [ ] Create test App ID with StoreKit configuration in the App Store Connect
 - [ ] Add products to the StoreKit configuration file
 - [ ] Setup the CI testing using [SKTestSession](https://developer.apple.com/documentation/storekittest/sktestsession) 
 - [ ] Create SandBox accounts for local testing

PR #3040, opened 2025-03-13
[Shipping] Publish StoreKit bindings as NuGet packages and run device tests

kotlarmilos:
## Description

This PR enables publishing StoreKit bindings as NuGet packages to the `dotnet-experimental` feed. The generated packages are:
```
Swift.Bindings.MacOSX.Experimental.nupkg
Swift.Bindings.MacCatalyst.Experimental.nupkg
Swift.Bindings.iPhoneOS.Experimental.nupkg
Swift.Bindings.iPhoneSimulator.Experimental.nupkg
Swift.Bindings.AppleTVOS.Experimental.nupkg
Swift.Bindings.AppleTVSimulator.Experimental.nupkg
```

Users need to add the experimental NuGet feed to the configuration and reference the appropriate package for the target platform.

## Changes

 - Updated `generate.sh` to create platform-specific bindings
   - Projection tooling inputs platform-specific ABIs
   - Script generates arm64 native libraries for devices and simulators
   - Script generates x64 native libraries for simulators
 - Updated PR build to generate MacCatalyst bindings and run device tests
 - Updated official build to generate bindings for all supported platforms
 - Updated build to install .NET 10 MAUI workloads

Contributes to https://github.com/dotnet/runtimelab/issues/2932


generate.sh
```diff
@@ -133,23 +195,24 @@ function CreateFramework {
         return
     fi
 
-    # x86_64
+    if [[ $is_maccatalyst == true ]]; then
+        fpath="/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/iOSSupport/System/Library/Frameworks"
```
rolfbjarne: You're assuming Xcode is in /Applications/Xcode.app here, I think something like this would be better:

```suggestion
        fpath="$(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/iOSSupport/System/Library/Frameworks"
```

generate.sh
```diff
@@ -133,23 +195,24 @@ function CreateFramework {
...
+        sdk="/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"
+    else
+        fpath="/Applications/Xcode.app/Contents/Developer/Platforms/${platform}.platform/Developer/SDKs/${platform}.sdk/System/Library/Frameworks/"
```
rolfbjarne:
```suggestion
        fpath="$(xcode-select -p)/Platforms/${platform}.platform/Developer/SDKs/${platform}.sdk/System/Library/Frameworks/"
```

generate.sh
```diff
@@ -224,19 +285,45 @@ function CreateProject {
...
+    <Content Include="./SwiftBindings_x86_64.xcframework/tvos-x86_64-simulator/**">
+        <PackagePath>runtimes/tvossimulator-x64/native/</PackagePath>
+        <Pack>true</Pack>
+    </Content>
     <Content Include="./SwiftBindings_arm64.xcframework/macos-arm64/**">
         <PackagePath>runtimes/osx-arm64/native/</PackagePath>
         <Pack>true</Pack>
     </Content>
-    <Content Include="./SwiftBindings_x64.xcframework/macos-64/**">
+    <Content Include="./SwiftBindings_x86_64.xcframework/macos-x86_64/**">
```
rolfbjarne: This won't handle symlinks correctly, because nupkgs don't support symlinks (nupkgs with symlinks aren't unzipped correctly, the symlinks aren't preserved).

Zipping up `./SwiftBindings_x86_64.xcframework` into `SwiftBindings_x86_64.xcframework.zip` and adding that zip file to the nupkg is the fix.

generate.sh
```diff
@@ -5,11 +5,12 @@ set -o pipefail
 usage()
 {
   echo "Common settings:"
-  echo "  --platform <value>         Platform: MacOSX, iPhoneOS, iPhoneSimulator, AppleTVOS, AppleTVSimulator"
+  echo "  --platform <value>         Platform: iPhoneOS, iPhoneSimulator, AppleTVOS, AppleTVSimulator, MacOSX"
   echo "  --version <value>          Version of the SDK"
-  echo "  --arch <value>             Architecture: arm64e-apple-macos, x86_64-apple-macos"
   echo "  --framework <value>        Framework to generate bindings for"
   echo "  --configuration <value>    Configuration: Debug, Release"
+  echo "  --maccatalyst              Use Mac Catalyst instead of iOS"
```
matouskozak: Why do we need special argument for MacCatalyst if we have the `--platform` argument already?

I think it might be better to specify it directly as an option if we need to to distinguish between `MacOSX` and `MacCatalyst` rather than using `iPhoneOS` + `--maccatalyst`.

generate.sh
```diff
@@ -68,12 +58,78 @@ while [[ $# > 0 ]]; do
...
+fi
+
+if [[ -z $version ]]; then
+    version=$(xcrun --sdk "$(echo "$platform" | tr '[:upper:]' '[:lower:]')" --show-sdk-version)
+fi
+
+case "$platform" in
+    "iPhoneOS")
+        target="apple-ios"
+        if $is_maccatalyst; then
```
matouskozak: shouldn't `$is_maccatalyst` be treated together with `macOS` rather than `iOS`?

generate.sh
```diff
@@ -103,11 +159,17 @@ function ExtractABI {
 function InvokeProjectionTooling {
     local framework=$1
 
-    $scriptroot/dotnet.sh $scriptroot/artifacts/bin/Swift.Bindings/$configuration/net9.0/Swift.Bindings.dll -a "./$framework.abi.json" -d "/System/Library/Frameworks/$framework.framework/$framework" -t "/Applications/Xcode.app/Contents/Developer/Platforms/$platform.platform/Developer/SDKs/$platform.sdk/System/Library/Frameworks/$framework.framework/$framework.tbd" -o "./"
+    if $tool; then
+        echo "Using tool to generate bindings for framework '$framework'"
+        $scriptroot/dotnet.sh swiftbindings -a "./$framework.abi.json" -d "/System/Library/Frameworks/$framework.framework/$framework" -t "/Applications/Xcode.app/Contents/Developer/Platforms/$platform.platform/Developer/SDKs/$platform.sdk/System/Library/Frameworks/$framework.framework/$framework.tbd" -o "./"
```
matouskozak: I'm not sure this will work correctly for MacCatalyst... 

Looking at the `StoreKit.tbd`, I see that `MacOSX.platform` directory has a subdirectory `iOSSupport` which I believe is for `MacCatalyst`. `./Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/iOSSupport/System/Library/Frameworks/StoreKit.framework/StoreKit.tbd`

generate.sh
```diff
@@ -31,23 +32,16 @@ scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
...
 version=''
-arch=''
 frameworks=()
 configuration='Debug'
+is_maccatalyst=false
+tool=false
 experimental=false
-
-output_dir="./GeneratedBindings"
+dotnet_version="net9.0"
```
matouskozak, line 38: How do we decide which .NET version we're on?

matouskozak (commented):
Looking good overall, left a few comments.

generate.sh
kotlarmilos: Initially, I wrapped it with iPhoneOS as the target is `apple-ios-macabi`. Now it is a different platform.

generate.sh
kotlarmilos: Target is `apple-ios-macabi` and SDK is macOS. I made separate case for it.

generate.sh
kotlarmilos: Good catch, updated.

generate.sh
kotlarmilos: It is related to the version of the tooling:
```
local script_path="$scriptroot/artifacts/bin/Swift.Bindings/$configuration/$dotnet_version/Swift.Bindings.dll"
```

generate.sh
kotlarmilos: Updated, thanks. Just to confirm:
 - The .zip file containing multiple targets is included in the NuGet package
 - During app build, the .zip file is copied into the app bundle
- The iOS SDK unzips the xcframework and places the target-specific framework into the bundle

matouskozak (approved):
LGTM! good work setting this up.

generate.sh
rolfbjarne: Step 2 (copying the zip file) doesn't happen because it's not necessary, we unzip directly from the nuget on disk, but otherwise yes, that's what's supposed to happen.

It should be fairly easy to verify it works as well: just reference the nupkg from an iOS or MacCatalyst project, make sure to add a reference in code to the nuget and build.

rolfbjarne (approved):

Issue #3010, opened 2025-02-18
[RuntimeAsync] Problem with boxing in resume stub when return type is a struct instantiated over object type


VSadov:
Re: https://github.com/dotnet/runtimelab/pull/2951#discussion_r1957542273

A possible repro is to add a scenario in `shared-generics.cs` test that looks like:
(regardless of Await optimization)

```cs
        Async1EntryPoint<S<string>>(typeof(S<string>), new S<string> {t = "ghj" }).Wait();
. . .

struct S<T>
{
    public T t;
}
```

We end up generating a resume stub that boxes `S<__Canon>` and that ends up failing around

```c
            // There is a pathological case where invalid IL refereces __Canon type directly, but there is no dictionary availabled to store the lookup.
            if (!pContextMD->IsSharedByGenericInstantiations())
                   COMPlusThrow(kInvalidProgramException);
```

VSadov:
@jakobbotsch 

I think the resumption stub for shared generics may need to be generic. Not sure how feasible this is for IL stubs.

Alternatively, since this is just for storage and the boxed instance will never be exposed to user code, we may use a `S<object>` for boxing - like construct a type that uses `object` in places where `__Canon` is used. 
(the upconvert in `return (T?)finalResult.GCData![0];` will need to use Unsafe cast in such case)

It may also be possible to deal with this by allowing boxing of canonical subtypes in this particular case without looking up for `T`, but allowing `__Canon` in scenarios that never seen it before could be breaking too many invariants expected form correct IL.

Boxing to actual type could also be nicer if something needs to introspect continuations - like for debugging purposes.

VSadov:
Another thing to think about is how this works if T is unloadable type. If we operate with `object`, `byte[]` or `__Canon` instead, would it be possible to unload too early. 

I think we might be ok, since resume stub does not create new instances, only boxes/serializes what exists already. Since something will eventually unbox/deserialize the data back to the actual type, the type must be rooted by other means.

jakobbotsch:
Another way to fix this would be to stop boxing the return types when they contain GC refs. The problem is that it requires storing all the GC refs of the return type into the `object[]` from C# / IL code, and figuring out where all of those GC refs are from C# is non-trivial. But it could probably be done by looking at the GCDescs from C#.

> Another thing to think about is how this works if T is unloadable type. If we operate with object, byte[] or __Canon instead, would it be possible to unload too early.

[...]

PR #3042, opened 2025-03-16
[RuntimeAsync] Normalize the box type in resume stubs

VSadov:
Fixes: https://github.com/dotnet/runtimelab/issues/3010

With this PR we create box instances via direct allocation. It has effect of bypassing all the layers that treat `Nullable` and `__Canon` in special ways, thus presence of either is no longer an issue.


jkotas:
> nullables

FYI: The pre-existing solution for this problem is to create nullable box by calling alloc helper directly, without creating new generic instantiations: https://github.com/dotnet/runtime/blob/4ca1691e9861e5d0c7dd8f4356b825f27c8dc8f5/src/coreclr/System.Private.CoreLib/src/System/RuntimeHandles.cs#L1193-L1195

VSadov:
> FYI: The pre-existing solution for this problem is to create nullable box by calling alloc helper directly, without creating new generic instantiations:

I was thinking of doing something like that, but was not sure how much of a taboo it is to have an object with Nullable MT. I think only JIT/typesystem know about Nullable, but GC could not care less, so for our limited use of the box - for storage only, maybe it is ok.

VSadov:
>  I think only JIT/typesystem know about Nullable, but GC could not care less, so for our limited use of the box - for storage only, maybe it is ok.

By this line of reasoning, perhaps using `__Canon` in the type is ok as well. GC does not care about that either.

src/tests/async/shared-generic.cs
```diff
@@ -6,16 +6,51 @@
...
     {
+        // simple cases
         Async1EntryPoint<int>(typeof(int), 42).Wait();
         Async1EntryPoint<string>(typeof(string), "abc").Wait();
         Async1EntryPoint<object>(typeof(object), "def").Wait();
 
+        // struct with an obj and its nullable
+        Async1EntryPoint<S0>(typeof(S0), new S0(42)).Wait();
+        // TODO: uncomment to repro https://github.com/dotnet/runtimelab/issues/3043
+        // Async1EntryPoint<S0?>(typeof(S0?), new S0(42)).Wait();
```
VSadov, line 31: Logged an issue (https://github.com/dotnet/runtimelab/issues/3043) to track this failure. 

I do not think this is related to returning a struct. In this case we start with non-null value, but lose the value right after resuming from `await Task.Yield` and then we return `null` because that is what we have. 

Even though a repro involves `Nullable`, it may not be related to `Nullable` or boxing, or may be related only indirectly.

VSadov:
So, after switching to direct allocation of the box instance, as @jkotas [suggested](https://github.com/dotnet/runtimelab/pull/3042#issuecomment-2727155456), I ended up removing the actual normalization of the box type as no longer needed. 

Once we create box instances via direct allocation we no longer involve any layer that care about `Nullable` or `__Canon` or the actual type of the box. The box is effectively just a `plain storage with matching GC layout` for the struct data.

VSadov:
I think I've done all that I planned here. @jakobbotsch @jkotas , please, take a look.

src/coreclr/vm/jitinterface.cpp
```diff
@@ -14526,11 +14526,27 @@ CORINFO_METHOD_HANDLE CEEJitInfo::getAsyncResumptionStub()
             // Now we have the GC array. At the first index is the result.
             pCode->EmitLDC(0);
 
-            // Box the result.
-            pCode->EmitLDLOC(resultLoc);
-            pCode->EmitBOX(pCode->GetToken(resultTypeHnd));
+            if (resultTypeHnd.IsValueType())
+            {
+                // make a box and dup the ref
```
jakobbotsch, line 14538: Add some comments why this doesn't just use regular `box` instruction?

jakobbotsch: Also, how does this affect suspension/resumption performance?

jakobbotsch (approved):

src/coreclr/vm/jitinterface.cpp
jkotas: For best performance of suspension/resumption, we would want to create a fake type that has all storage inlined (no separate boxes for each variable) and that has just GC info required for GC to work (no EEClass, etc.).

jakobbotsch: I discussed that a bit with @davidwrighton, but he suggested that actually the `object[]` and `byte[]` approach is nice for start up because it does not involve any sort of type loading. Although I suppose your suggestion would only require some very lightweight form of type loading.

FWIW, not all variables get boxed -- it only happens for the return value of a resumed runtime async method if that return value is a struct containing GC pointers. Other variables are stored in the continuation with untracked data in the `byte[]` and GC pointers in the `object[]`. The JIT moves contained GC pointers in/out of the `object[]` as necessary. That would likely also be slightly faster with a JIT-time generated type.

jkotas: > Although I suppose your suggestion would only require some very lightweight form of type loading.

Right, it is really just allocate a chunk of memory on loader heap and fill it in with bytes. I would not be surprised if it is a bit cheaper than generating the code to deal with the boxes, etc.

VSadov: > Also, how does this affect suspension/resumption performance?

We can measure. I would expect this scenario to be dominated by how much memory we touch and how many bytes we allocate. Both did not change. 
There is some change in the logic. The box allocation is now a Qcall, but it is hard to say if it will have a lot of difference next to the cost of suspending/resuming. 

I'll try to construct something that can measure the impact.

VSadov: > Also, how does this affect suspension/resumption performance?

As expected, there is a penalty for allocating via QCall, but it is not large. 
On a recursive Fibonacci(30) that wraps results in structs and yields in leaf calls I see about ~6% total throughput regression. 

Measurements with 
```
DOTNET_gcServer=1
DOTNET_TieredCompilation=0
```

before the change:
```
async2, returning a struct with an object

456 ms result=1346269
465 ms result=1346269
460 ms result=1346269
456 ms result=1346269
465 ms result=1346269
462 ms result=1346269
468 ms result=1346269
459 ms result=1346269
453 ms result=1346269
443 ms result=1346269
allocated: 7216888480

async2, returning a struct with no object

398 ms result=1346269
396 ms result=1346269
398 ms result=1346269
397 ms result=1346269
398 ms result=1346269
396 ms result=1346269
397 ms result=1346269
399 ms result=1346269
398 ms result=1346269
397 ms result=1346269
allocated: 6247465000  15% less when we do not box the return values.
```

after the change
```
async2, returning a struct with an object    (best is 6% slower than baseline best)

487 ms result=1346269
484 ms result=1346269
485 ms result=1346269
477 ms result=1346269
486 ms result=1346269
478 ms result=1346269
483 ms result=1346269
481 ms result=1346269
496 ms result=1346269
471 ms result=1346269
allocated: 7216893016  
  
NB: not all allocations here are ours and "precise" alloc accounting gets different results each time. 
This is very close to the baseline though.

async2, returning a struct with no object   (roughly the same as baseline)

395 ms result=1346269
392 ms result=1346269
392 ms result=1346269
395 ms result=1346269
393 ms result=1346269
396 ms result=1346269
393 ms result=1346269
394 ms result=1346269
395 ms result=1346269
400 ms result=1346269
allocated: 6247459344
```

VSadov: The benchmark I used for the above:
```cs

// #define WITH_OBJECT

using System;
using System.Threading.Tasks;
using System.Diagnostics;
using Xunit;

public class Async2FibonacceWithYields
{
    const int iterations = 10;

    [Fact]
    public static void Test()
    {
        long allocated = GC.GetTotalAllocatedBytes(precise: true);

        AsyncEntry().GetAwaiter().GetResult();

        allocated = GC.GetTotalAllocatedBytes(precise: true) - allocated;
        System.Console.WriteLine("allocated: " + allocated);
    }

    public struct MyInt
    {
        public int i;
#if WITH_OBJECT
        public object dummy;
#else
        IntPtr dummy;
#endif
        public MyInt(int i) => this.i = i;
    }

    public static async2 Task AsyncEntry()
    {
        for (int i = 0; i < iterations; i++)
        {
            var sw = Stopwatch.StartNew();
            MyInt result = await Fib(new MyInt(30));
            sw.Stop();

            Console.WriteLine($"{sw.ElapsedMilliseconds} ms result={result.i}");
        }
    }

    static async2 Task<MyInt> Fib(MyInt n)
    {
        int i = n.i;
        if (i <= 1)
        {
            await Task.Yield();
            return new MyInt(1);
        }

        int i1 = (await Fib(new MyInt(i - 1))).i;
        int i2 = (await Fib(new MyInt(i - 2))).i;

        return new MyInt(i1 + i2);
    }
}

```

VSadov: Out of curiosity - how that works with async1

```
async1 returning a struct with an object

554 ms result=1346269
550 ms result=1346269
558 ms result=1346269
552 ms result=1346269
562 ms result=1346269
554 ms result=1346269
558 ms result=1346269
568 ms result=1346269
574 ms result=1346269
570 ms result=1346269
allocated: 3661864408

async1 returning a struct with no object

516 ms result=1346269
521 ms result=1346269
531 ms result=1346269
529 ms result=1346269
530 ms result=1346269
534 ms result=1346269
532 ms result=1346269
533 ms result=1346269
529 ms result=1346269
533 ms result=1346269
allocated: 3661872696
```

async1 is slightly slower, but allocates 50% less. 

The difference in alocations is because async1 on the first suspend allocates a display struct that is large enough for all  awaits in the method, and then reuses if any other await in the method suspends. 
async2 allocates continuations according to the need of a particular await and does not reuse. 

The benchmark happens to suspend 2 times in every method, but I would question whether this behavior matches real programs. Evidently reduction in allocations does not compensate for extra complexity. async2 is still faster.

Async2 could reuse continuations as well, if can be proven useful in real world cases, but that is somewhat doubtful, so it makes sense to start with simpler, no reuse, solution.

Perhaps if there are awaits inside a loop, it could make sense to reuse continuations. 
There will be time for this kind of tweaks, when everything works.

VSadov: And if the `Yield` is removed from the benchmark, async2 is 10X faster.

async1
```
async1 returning a struct with no object no yield

48 ms result=1346269
48 ms result=1346269
46 ms result=1346269
46 ms result=1346269
45 ms result=1346269
45 ms result=1346269
45 ms result=1346269
45 ms result=1346269
44 ms result=1346269
47 ms result=1346269
allocated: 2154035888

async1 returning a struct with no object no yield, ValueTask  
(interestingly, allocates less, and yet slower)

64 ms result=1346269
61 ms result=1346269
61 ms result=1346269
61 ms result=1346269
62 ms result=1346269
61 ms result=1346269
61 ms result=1346269
61 ms result=1346269
61 ms result=1346269
61 ms result=1346269
allocated: 5560
```

async2
```
async2 returning a struct with no object no yield  (10X faster than async1)

6 ms result=1346269
5 ms result=1346269
5 ms result=1346269
6 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
allocated: 5408

async2 returning a struct with no object no yield, ValueTask  (the same as Task!!!)

5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
allocated: 5408
```

VSadov:
>> Although I suppose your suggestion would only require some very lightweight form of type loading.

> Right, it is really just allocate a chunk of memory on loader heap and fill it in with bytes. I would not be surprised if it is a bit cheaper than generating the code to deal with the boxes, etc.

The box always contains object references, so that adds some complications with using native memory for it. 

Continuations are used to store the state of suspended methods and thus can be a mix of long-lived/short-lived things, may happen in huge quantities, form cactus-shape graphs because of WaitAll, and such. Managed memory is the best for this, but we may be able to squeeze some extra perf or size improvements from specializing storage shape. Eventually, maybe.. 

The boxes are different. They are typically very short-lived. The purpose is basically to pass a result of calee invocation to the suspended caller. The most efficient way to pass that would be on the stack. 
In theory it is possible that the caller has a different sync context and will need to run on a different thread, thus we have to pass the return value via heap. However, the most common case is when the caller will run on the same thread immediately after the callee returns (modulo dispatching machinery in between).  

I think it could be an interesting optimization in cases when the caller can run on the same stack as calee, and when we have enough stack left, to just call the caller's resume from the calee. It would be logically similar optimization to the "continuation inlining" that Task does. In such case it might be possible to place the return value on the stack and pick it up in the caller. 
The issue with GC reporting will still need to be dealt with - maybe put current thread in noGC state for the duration of transfer, introduce conservatively-scanned frame, or just make stack scanner understand how to report the refs using some map.

I think it is a bit too early to embark on this kind of "turn it to eleven" optimizations. Probably worth to make a note for the future.

VSadov:
It looks like for the purpose of this PR the change is complete (I did add comments about manual boxing).
So, to not block fixing https://github.com/dotnet/runtimelab/issues/3043 , I will merge this.

Thanks!

VSadov:
Re: Allocation helper being a QCall. - This can be improved, if there is enough need.  
We could make the helper an FCALL that tries to handle typical cases (small obj, no finalizer) via pointer move. If that does not work, which should be rare, set up a frame and call into GC Alloc.  
Allocations via a helper are quite rare though, compared to JIT allocations, so, I guess, it's perf was never a big priority.

jakobbotsch:
> The box always contains object references, so that adds some complications with using native memory for it.

I do not think @jkotas was suggesting to use native memory for this, but rather to remove the `Continuation` class entirely in favor of generating types during JIT-time when the layout of the continuation is known. The generated type would have a field for the result at a known offset that the previous continuation's resumption stub would write to.

Another potential painpoint with using "fake" type for this is probably in how we expose `Continuation` instances to diagnostics. (Current layout is of course not immune to this and needs some solution also.)

> The difference in alocations is because async1 on the first suspend allocates a display struct that is large enough for all awaits in the method, and then reuses if any other await in the method suspends.
async2 allocates continuations according to the need of a particular await and does not reuse.
> ...
> I think it is a bit too early to embark on this kind of "turn it to eleven" optimizations. Probably worth to make a note for the future.

I've also discussed some of these with David and the general opinion was to leave this up to the GC and let it do its job. Reusing continuations will come with some cost to the synchronous path as well, although this cost might be small.
In any case it might be interesting to benchmark, but it's unclear to me how to faithfully benchmark something like this.

By the way, there is a list of potential future improvements here: https://github.com/dotnet/runtimelab/blob/feature/async2-experiment/docs/design/features/runtime-handled-tasks.md#potential-future-improvements.
Feel free to expand on it.

VSadov:
> Another potential painpoint with using "fake" type for this is probably in how we expose Continuation instances to diagnostics. (Current layout is of course not immune to this and needs some solution also.)

Roslyn has quite different codegen for async when building Debug code. It captures nearly everything because users do not like locals "disappearing" in debugger after await. Storage fields are mapped to locals one-to-one in Debug with no reuse. The fields also have special name mangling, so that debugger could figure which field is backing which local. The display struct is actually a class in Debug - to support Edit-end-Continue...  
It will be interesting if any of these de-optimizations (or scenarios that they enable) are relevant to runtime async. 

Capturing all locals that can be captured in Debug may be already happening - because all locals are untracked.
For regular debugging like stepping through understanding Continuation may not be needed, as active methods will not have locals in storage. 
If there are other scenarios that require to know the values of locals in suspended frames, it would be an interesting challenge.

PR #3050, opened 2025-03-20
[RuntimeAsync] Made Fibonacci tests to actually run Fibonacci algorithm

VSadov:
Test-only change:

- Made Fibonacci tests to actually run Fibonacci algorithm. 
I do not know the history of tests. I assume Fibonacci was the intent, but the tests ended up running some simple loops, not exercising recursion.
If tests say Fibonacci, perhaps they should run Fibonacci.

- deleted couple useless benchmarks that do the same thing as others

- added a Fibonacci test that returns results wrapped in a struct 
(as in https://github.com/dotnet/runtimelab/pull/3042#discussion_r1999896743)

The overall testpass for runtime async takes roughly the same time. Many of these tests are actually build-only tests for manual testing/benchmarking/comparing and do not contribute to testpass time.


jakobbotsch (approved):

VSadov:
Thanks!!

PR #3051, opened 2025-03-21
[RuntimeAsync] Recognizing new overloads of Await related to the ConfigureAwait pattern.

VSadov:
- Adds overloads of Await as specified in  
 https://github.com/dotnet/runtime/pull/113675 and JIT logic to recognise them

The  true/false arguments to ConfigureAwait are recognised, but not passed further and have no effect yet as we do not have the actual mechanism to flow (or not) the execution context. 
We can, though, recognize and optimize the pattern that is emitted by `await SomethingAsync().ConfigureAwait(false/true)`

- Adds a test.

### DIFF
--- a/src/coreclr/jit/importer.cpp
+++ b/src/coreclr/jit/importer.cpp
@@ -9013,11 +9013,18 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                 if (opcode != CEE_CALLI)
                 {
                     bool isAwait = false;
-                    if (JitConfig.JitOptimizeAwait())
+                    // TODO: The configVal should be wired to the actual implementation
+                    //       that control the flow of sync context.
+                    //       We do not have that yet.
+                    int configVal= -1;  // -1 not congigured, 0/1 configured to false/true
+                    if (compIsAsync2() && JitConfig.JitOptimizeAwait())
                     {
                         // If we see the following code pattern in runtime async methods:
                         //
                         //    call[virt] <Method>
+                        //    [ OPTIONAL ]
+                        //       ldc.i4.0 / ldc.i4.1
+                        //       call[virt] <ConfigureAwait>
                         //    call       <Await>
                         //
                         //  we emit an eqivalent of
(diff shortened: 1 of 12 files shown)


VSadov:
To be sure that the new helpers work, if not optimized, and then get better with JIT help:

The test `fibonacci-without-yields-config-await` set to 10 iterations  demonstrates that optimization works and reduces both the run time and allocations:

```
E:\A2\runtimelab\artifacts\tests\coreclr\windows.x64.Release>set DOTNET_JitOptimizeAwait=0

E:\A2\runtimelab\artifacts\tests\coreclr\windows.x64.Release>async\fibonacci-without-yields-config-await\fibonacci-without-yields-config-await.cmd
BEGIN EXECUTION
 "E:\A2\runtimelab\artifacts\tests\coreclr\windows.x64.Release\Tests\Core_Root\corerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false" -p "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization=true"  fibonacci-without-yields-config-await.dll
45 ms result=1346269
42 ms result=1346269
41 ms result=1346269
41 ms result=1346269
34 ms result=1346269
14 ms result=1346269
14 ms result=1346269
14 ms result=1346269
14 ms result=1346269
14 ms result=1346269
allocated: 141427160
Expected: 100
Actual: 100
END EXECUTION - PASSED
PASSED
```

With default settings:

```
E:\A2\runtimelab\artifacts\tests\coreclr\windows.x64.Release>set DOTNET_JitOptimizeAwait=

E:\A2\runtimelab\artifacts\tests\coreclr\windows.x64.Release>async\fibonacci-without-yields-config-await\fibonacci-without-yields-config-await.cmd
BEGIN EXECUTION
 "E:\A2\runtimelab\artifacts\tests\coreclr\windows.x64.Release\Tests\Core_Root\corerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false" -p "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization=true"  fibonacci-without-yields-config-await.dll
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
5 ms result=1346269
allocated: 5408
Expected: 100
Actual: 100
END EXECUTION - PASSED
PASSED
```

src/coreclr/jit/importer.cpp
```diff
@@ -9031,7 +9038,36 @@ void Compiler::impImportBlockCode(BasicBlock* block)
...
+                            {
+                                goto checkForAwait;
+                            }
+
+                            configVal = nextOp == CEE_LDC_I4_0 ? 0 : 1;
+                            // skip {ldc; call; <ConfigureAwait>}
+                            nextOpcode += 1 + 1 + sizeof(mdToken);
+                        }
+
+                    checkForAwait:
```
jakobbotsch: This logic is getting complicated enough that I think we should extract it all to a function.

VSadov: Moved the Await pattern-match out as a separate method.

VSadov: Are there more concerns or ok to merge?

jakobbotsch (approved):

VSadov:
Thanks!!

Issue #3011, opened 2025-02-19
[RuntimeAsync] Implement struct async thunks.


VSadov:
Re: https://github.com/dotnet/runtimelab/pull/2951/files#r1960790542

There is an assert:

```c
       // Struct async thunks not yet implemented
       _ASSERTE(!this->GetMethodTable()->IsValueType());
```

It can be hit by `Async2Struct` testcase by changing the `void Async()` entry point to be an ordinary async method. 

```cs
      private static async Task Async()    // not async2
```

Some of the scenarios in the test may become irrelevant since it will be C# responsibility to capture by value.

However the general case of an async struct method still needs to be handled, so this needs a follow up.

VSadov:
Fixed in https://github.com/dotnet/runtimelab/pull/3054

PR #3054, opened 2025-03-23
[RuntimeAsync] Turn off the assert and enable test scenario for struct method thunks

VSadov:
Fixes:  https://github.com/dotnet/runtimelab/issues/3011

The change basically removes an assert and adds a test scenario. It looks like thunks for struct methods do not require anything special - we transparently call the async2 variant with exactly same args and then finalize the result into Task.  
(maybe they required something special in previous variation of spec for struct async methods and thus we had the assert).

### DIFF
--- a/src/coreclr/vm/prestub.cpp
+++ b/src/coreclr/vm/prestub.cpp
@@ -1843,8 +1843,6 @@ void MethodDesc::EmitJitStateMachineBasedRuntimeAsyncThunk(MethodDesc* pAsyncOth
             DWORD localArg = 0;
             if (thunkMsig.HasThis())
             {
-                // Struct async thunks not yet implemented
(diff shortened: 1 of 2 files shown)


VSadov:
Possibly I am missing something, but it does not look like from the point of async2->Task thunks that struct methods are any different. 
Maybe there was some concern with passing `this` by value, but that is gone now.

jakobbotsch (approved):
Yeah, looks like I just missed this direction in #2805.

BTW we are still using the old struct based implementation, where JIT creates copies. We should be able to remove that once this part is implemented in Roslyn.

VSadov:
> We should be able to remove that once this part is implemented in Roslyn.

Yes. I think it can be removed. Probably do not even need to wait for Roslyn.

VSadov:
Thanks for reviewing!!

PR #3057, opened 2025-03-26
[Swift projection tooling] Initial support for Logger

matouskozak:
This PR adds initial support for Logger inside the Swift projection tooling. 

### Key Changes
- Logger implementation is moved to `Utils.Logging` namespace
- `--verbose` now translates to a Logger setting which is passed around to ensure appropriate and consistent logging. The possible options are: 
```
- 0 = No logging 
- 1 = Only errors
- 2 = Errors and warnings
- 3 = Errors, warnings, and info
- 4 = Errors, warnings, info, debug
```

### Future Work
- Onboard `Conductor` + `Handler*` codebase as this requires more extensive work rather than just replacing `int verbose` with `ILLogger logger`.

---

FYI: If you were using `--verbose 2` to get all logging (including debug prints), please switch to `--verbose 4` after this PR.

Partially contributes to https://github.com/dotnet/runtimelab/issues/2973

PR #3058, opened 2025-03-26
[Swift projection tooling] New logging mechanism using Microsoft.Extensions.Logging API

matouskozak:
This is an alternative to https://github.com/dotnet/runtimelab/pull/3057 using https://learn.microsoft.com/en-us/dotnet/core/extensions/logging instead of custom logger implementation.

---

TODO:
- [ ] Determine where it is appropriate to use `LogInformation`, `LogWarning`, `LogError`, etc.

Partially contributes to https://github.com/dotnet/runtimelab/issues/2973


kotlarmilos (approved):
LGTM!

kotlarmilos:
>  Determine where it is appropriate to use LogInformation, LogWarning, LogError, etc.

I think you used it right:
 - Info - stages at the tooling
 - Warning - can't process a type but continue
 - Error - throw an exception and stop the execution

matouskozak:
## Overview
This log shows the process of generating bindings for a StoreKit ABI JSON file, with numerous warnings occurring during the operation.

## Component Summary

| Component | Levels | Key Issues |
|-----------|--------|------------|
| `BindingsGeneration.SwiftABIParser` | warn: 171 | Protocol Conformance Issues, Missing Metadata Accessors, Generic Type Not Supported |
| `BindingsGeneration.PropertyHandlerFactory` | warn: 117 | Property Processing Failures |
| `BindingsGeneration.ModuleProcessor` | warn: 68 | Type should have been processed in a previous module. |
| `BindingsGeneration.MethodHandlerFactory` | warn: 35 | Unsupported Types/Signatures |
| `BindingsGeneration.ConstructorHandlerFactory` | warn: 10 | Unsupported Types/Signatures |
| `BindingsGeneration.BindingsGenerator` | info: 2 | None |
| `TbdParsing.TbdParser` | info: 2 | None |
| `TbdParsing.Parsing.YamlLikeTbdFormatParser` | warn: 1 | None |

## Common Issues

1. **Protocol Conformance Issues** (143 occurrences):
   - "Protocol conformance descriptor not found for type X and protocol Y"
   - Particularly with Swift.Sendable, Swift.Equatable, Swift.Hashable

2. **Property Processing Failures** (117 occurrences):
   - "PropertyHandler: Couldn't process property X of type Y. Skipping."
   - Foundation types (Date, UUID, Decimal) particularly problematic

3. **Type should have been processed in a previous module.** (66 occurrences):
   - "Skipping property X of type Y from module Z"
   - "Type should have been processed in a previous module but was not found"

4. **Unsupported Types/Signatures** (45 occurrences):
   - "Method X has unsupported signature"
   - "Constructor init has unsupported signature"
   - Generic parameters not supported

5. **Missing Metadata Accessors** (7 occurrences):
   - "Metadata accessor not found for type X"
   - Often occurs with Foundation types like Date, UUID, String

6. **Generic Type Not Supported** (5 occurrences):
   - "Generic type 'Array' not supported. Skipping."
   - "Generic type 'Dictionary' not supported. Skipping."

7. **Unsupported Node Types** (2 occurrences):
   - "Unsupported node kind 'Subscript' encountered"
   - "Error while processing node: Index was outside the bounds of the array"

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
```diff
@@ -244,8 +275,15 @@ public ITypeHandler Construct()
     /// </summary>
     public class NonFrozenStructHandler : BaseHandler, ITypeHandler
     {
-        public NonFrozenStructHandler()
+        private readonly ILogger _logger;
```
vitek-karas: It looks like `BaseHandler` will need the logger as well (it has `Console.WriteLine` in it) - so it would be probably cleaner to put the logger onto the base class and "share" it.

Also - consider doing the same for the HandlerFactory - we don't have a base class yet, but we could create one which just stores the logger for now.

matouskozak: Good catch, I missed this one. I'll add it.

> so it would be probably cleaner to put the logger onto the base class and "share" it.
> we don't have a base class yet, but we could create one which just stores the logger for now.

Do you mean creating just one logger for the `BaseHandler` and sharing it among all `*Handler` subclasses?

If we would put the logger just to the base class and share it among the subclasses, we would lose the ability to distinguish from which subclass the logging came. Now we see:

```
2025-03-26T18:02:36.3724510Z warn: BindingsGeneration.PropertyHandlerFactory[0]
2025-03-26T18:02:36.3726390Z       PropertyHandler: Couldn't process property agreed of type (StoreKit.Product.SubscriptionInfo.RenewalInfo.PriceIncreaseStatus.Type) -> StoreKit.Product.SubscriptionInfo.RenewalInfo.PriceIncreaseStatus. Skipping.
2025-03-26T18:02:36.3727310Z warn: BindingsGeneration.MethodHandlerFactory[0]
2025-03-26T18:02:36.3728180Z       Method hash has unsupported signature: ( Swift.AnyType into) -> void
```

and with sharing the logger from the base class, all logging would be indistinguishable unless we would specify it in the log message. (It's my first time using these APIs, so please let me know if there is a better way.)

Now it is also not ideal as all handlers are sharing the logger from their corresponding factory. So `PropertyHandler` logging is reported under `BindingsGeneration.PropertyHandlerFactory`. We should ideally create a new logger for each created handler to get more granual logging.

I'm open to suggestions on how to make it better.

vitek-karas: Hmm - didn't realize the type in the logger is "persistent".
Based on my reading, there's no reason to store the fully typed `ILogger<OwnerType>` as it makes no difference to the behavior of the logger. This is basically "proven" by the fact that `ILoggerFactory.CreateLogger` returns the `ILogger` - it's only the extension generic method which returns `ILogger<T>`. It also says "Generally used to enable activation of a named ILogger from dependency injection." in the docs.
https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LoggerT.cs#L16

Since we don't use DI, the only place where the type parameter is important is at the creation, otherwise we could store just `ILogger` everywhere.

As for the level of granularity - I agree that it would be really nice to have something like `MethodHandler[Swift.String.init]` be the category, so that it's very clear what we're trying to do. Not sure if we have this info at the place where we would need to create the logger.

matouskozak: Actually, I was confused and 

> so it would be probably cleaner to put the logger onto the base class and "share" it.
> we don't have a base class yet, but we could create one which just stores the logger for now.

seems to work.

> Not sure if we have this info at the place where we would need to create the logger.

I think with the current design, we will have to rely on writing descriptive message in this regard.

jkurdek (approved):

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
matouskozak: @vitek-karas are the latest changes what you had on mind?

src/Swift.Bindings/src/Demangler/TbdParser/Parsing/JsonTbdFormatParser.cs
```diff
@@ -17,21 +17,21 @@ public class JsonTbdFormatParser : TbdFormatParserBase
         /// <summary>
         /// Creates a new JSON TBD format parser
         /// </summary>
-        public JsonTbdFormatParser(ILogger logger) : base(logger)
+        public JsonTbdFormatParser(ILogger<JsonTbdFormatParser> logger) : base(logger)
```
vitek-karas:
```suggestion
        public JsonTbdFormatParser(ILogger logger) : base(logger)
```

Per our dicussion, I think the "category" for the logger should be specified only when we're creating the logger - the thing using the logger should not care about it.

vitek-karas: Similarly in other places. Basically I don't see a reason to use `ILogger<T>` anywhere as an explicit type. It would make sense if we used DI, but we don't...

src/Swift.Bindings/src/Demangler/TbdParser/TbdParser.cs
```diff
@@ -22,25 +23,25 @@ public class TbdParser
         /// <summary>
         /// Initializes a new instance of the TBD parser with a default logger
         /// </summary>
-        public TbdParser() : this(new ConsoleLogger { MinimumLevel = LogLevel.Info })
+        public TbdParser() : this(NullLoggerFactory.Instance)
```
vitek-karas: Nit: I would remove this constructor - it's only used in tests and in tests it's better to be explicit about where the logging will go (if anywhere). Also this means that tests don't use the same code as the product, which is always a bit iffy.

vitek-karas (approved):
I think this is good as-is, I added some comment on possible improvements, but don't consider them blocking.

src/Swift.Bindings/src/Emitter/StringEmitter/Handler/TypeHandler.cs
vitek-karas: Yes - thanks a lot.

Issue #2843, opened 2024-12-03
[Projection tooling] Design projection tooling to be extensible and ensure it outputs only supported types


kotlarmilos:
## Tasks
 - [x] https://github.com/dotnet/runtimelab/issues/2890
 - [x] https://github.com/dotnet/runtimelab/issues/2891
 - [x] #2955
 - [ ] Create design diagram/document for filtering handler
 - [ ] Create design diagram/document for intermediate representation
 - [ ] Create design diagram/document for marshalling handlers
 - [ ] Create design diagram/document for emitters

> [!NOTE]  
> This issue tracks progress and will be expanded with more details. If specific tasks for this issue are not clear, this means that the task is not well defined yet and needs attention.

PR #3059, opened 2025-03-28
[Swift Bindings] Emitter redesign proposal

jkurdek:
This PR introduces a design doc of potential emitter redesign.

Fixes https://github.com/dotnet/runtimelab/issues/2843


src/docs/emitter-redesign-proposal.md
```diff
@@ -0,0 +1,333 @@
...
+2. Graph pruning and TypeDatabase population
+
+### Graph traversal and type annotation
+
+The input of this step is a collection of TypeDecls. Each declaration contains metadata about whether it is a ref type, whether it is frozen, etc.
+
+We can think about this input as a directed graph:
+
+- Nodes are `TypeDecls` / `TypeDatabase` entries
+- Edges are fields and stored properties (`TypeSpecs`)
```
vitek-karas, line 37: It's unclear what will be other things?
- Methods - if the method parameter is a type which can't be projected... can we project the method? But even if we can, we should populate the type into the type database. At least according to this description the type database should be effectively immutable after the type pre-processing and validation.
- Any other swift language artifacts (associated types?) which may act as type referenced and thus would require "type load".

src/docs/emitter-redesign-proposal.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### Marshalling Label Decision Rules
+
+During graph traversal, we assign each type one of the following marshalling labels:
+
+```csharp
+public enum MarshallingLabel
+{
+    Struct,               // Frozen structs with only frozen struct fields
+    ClassWithOpaquePayload, // Non-frozen structs and mixed-field frozen structs
```
kotlarmilos, line 51: With safe handlers that are responsible for memory disposal, we could project Swift structs as C# structs.

src/docs/emitter-redesign-proposal.md
```diff
@@ -0,0 +1,333 @@
...
+   }
+   ```
+
+2. **ClassWithOpaquePayload** - Used for non-frozen structs or frozen structs that contain fields that must be projected as classes with opaque payload.
+
+   Example:
+
+   ```swift
+   struct DynamicBuffer {  // Non-frozen
+       var elements: [Int]
```
kotlarmilos: I suggest using a different example, since Array has a buffer payload with a pointer, which gets lowered.

src/docs/emitter-redesign-proposal.md
```diff
@@ -0,0 +1,333 @@
...
+The input of this step is a pruned collection of TypeDecls and a TypeDatabase containing types from the current module.
+
+The simplest representation of a type is a `TypeRepresentation` which contains information about the type itself and its members.
+
+```csharp
+public record struct TypeRepresentation
+{
+    public required CSharpTypeName CSharpTypeName { get; init; }
+    public required MarshallingLabel Label { get; init; }
+    public required List<string> Methods { get; init; }
```
kotlarmilos, line 158: What is stored in `List<string> Methods`? Is it the source code of the methods which are emitted later?

src/docs/emitter-redesign-proposal.md
```diff
@@ -0,0 +1,333 @@
...
+```
+
+In the simplest scenario 1:1 mapping between MethodDecl and MethodRepresentation the `MarshallingContext` would contain the `MethodDecl` and the `MethodRepresentation`.
+
+```csharp
+public record struct MarshallingContext
+{
+    public MethodDecl MethodDecl { get; set; }
+    public ITypeRepresentation ContainingType { get; set; }
+    public MethodRepresentation MethodRepresentation { get; set; }
```
kotlarmilos, line 246: Can MethodRepresentation embed MethodDecl information? It looks like MethodDecl is an input in multiple stages?

src/docs/emitter-redesign-proposal.md
```diff
@@ -0,0 +1,333 @@
...
+   }
+   ```
+
+5. **Unknown** - Used for types that cannot be marshalled or are not supported.
+
+#### UNKNOWN Label Propagation Rules
+
+The `Unknown` label propagates according to these rules:
+
+1. If a type would be marshalled as something that requires layout (Struct, ClassWithBufferStruct) and any of its stored properties are `Unknown`, the type itself becomes `Unknown`.
```
matouskozak, line 108: > requires layout

This refers to the need of knowing exact layout needed for correct struct lowering?

src/docs/emitter-redesign-proposal.md
```diff
@@ -0,0 +1,333 @@
...
+## Type processing
+
+The purpose of this step is to process each type and its members, building a representation of how they should be marshalled.
+
+The input of this step is a pruned collection of TypeDecls and a TypeDatabase containing types from the current module.
+
+The simplest representation of a type is a `TypeRepresentation` which contains information about the type itself and its members.
+
+```csharp
+public record struct TypeRepresentation
```
matouskozak, line 154: It appears that `TypeRepresentation` has very similar structure to current `TypeDecl` type. Would it make sense to reuse `TypeDecl` inside `TypeRepresentation` and extend it with the additional code?

src/docs/emitter-redesign-proposal.md
```diff
@@ -0,0 +1,333 @@
...
+
+A property has to be marshalled as up to three different things:
+
+- A field (stored properties on frozen structs)
+- A PInvoke
+- A property
+
+A simple handler can be created which will do the necessary work and update the necessary bits of the representation.
+
+**NOTE**: Each of the handlers will need to decide whether a given method or property is supported or not. This can be done by checking the TypeDatabase for the appropriate type. This can be done by adding a separate flag to the MethodRepresentation or by having some handler validate the arguments / properties types beforehand.
```
matouskozak, line 319: > Each of the handlers will need to decide whether a given method or property is supported or not.

Since we are working with pruned graph here, we should be confident that there is always a handler for each encountered type, correct?

src/docs/emitter-redesign-proposal.md
jkurdek: this should get rid of one of the reasons we need to project structs as classes. We will still need to project non-frozen structs as classes due to the layout being unknown.

src/docs/emitter-redesign-proposal.md
jkurdek: yes, thats right

src/docs/emitter-redesign-proposal.md
jkurdek: I think we need multiple things:
- TypeDecl - this is a collection of facts about **Swift** - those are populated using input files with minimal transformations
- TypeRepresentation - this is representation of **C#** type.

Since there is not 1:1 mapping between the two above I think keeping them separate is better

src/docs/emitter-redesign-proposal.md
jkurdek: At the end of the document:

```
A matter for discussion is whether we want to serialize the representation to a string and then emit it, or whether we want to emit the code directly from the representation. Sometimes it is easier to just append a string to the the representation rather than build up a whole representation - this is particularly true when handling interface implementation methods. One way would be to support both.
```

It it up to the implementation to decide whether the TypeRepresentation should contain a list of MethodRepresentations (described below) or a list of strings and some MethodRepresentationBuilder.

src/docs/emitter-redesign-proposal.md
jkurdek: I think it doesnt make much difference. What is the key point here is that we need a reference to MethodDecl

src/docs/emitter-redesign-proposal.md
jkurdek: There are two stages of pruning necessary:
1. Pruning types - this is done before this step.
2. Pruning unknown type members - e.g there is a generic constructor on a type. This is done here.

What I mean is that at any given point of this project development up to completion there will be a set of unsupported constructs. Be it generic constructors or async properties. We need some mechanism to not emit them.

src/docs/emitter-redesign-proposal.md
matouskozak: Thank you, not sure if I missed this but having this description somewhere inside the docs would be nice.

src/docs/emitter-redesign-proposal.md
matouskozak: Thanks, I didn't realize the two stages of pruning.

Do you think we could unify the stages to a single one? I.e., in the pruning graph stage also figure out the unknown type members?

src/docs/emitter-redesign-proposal.md
kotlarmilos: This should be independent of the projection because types now provide safe handlers. If struct is non-frozen, SafeHandle is passed at the PInvoke boundary; otherwise a buffer is retrieved from the SafeHandle as
```
public unsafe PayloadBuffer<{structDecl.Name}.Buffer> PayloadBuffer => new PayloadBuffer<{structDecl.Name}.Buffer>(_payload);
```

kotlarmilos (approved):
Great work, thanks!

src/docs/emitter-redesign-proposal.md
jkurdek: >Methods - if the method parameter is a type which can't be projected... can we project the method? But even if we can, we should populate the type into the type database. At least according to this description the type database should be effectively immutable after the type pre-processing and validation.

I am not sure I understand that. We populate type database when we see a declaration of the type. Seeing a type reference in method parameter type is not enough information to generate a valid type database entry - we do not know whether it is a class or a struct etc.

If a method contains a type which is not known at the time of the projection there might be two causes:
- The type is private - I do not think this will be possible
- We failed to project the type before for some reason

I think that due to processing dependencies in topological order those are the only two options

jkurdek: >Any other swift language artifacts (associated types?) which may act as type referenced and thus would require "type load".

Associated types will not be a reference. But generic constraints are a type refence reference. So are the methods parameters types. They do not affect layout though.

src/docs/emitter-redesign-proposal.md
jkurdek: In this proposal we freeze the type database after the first pruning as it prunes types depending on whether we are able to represent their layout. The second step doesn't touch the type database - it just decided that some parts of the types should be omitted - however those pieces do not affect other constructs.

It is possible to merge those two steps into one - but then we would need to handle the types / members in topological order as we cannot rely on type database being populated

src/docs/emitter-redesign-proposal.md
vitek-karas: Maybe I misunderstood how this is supposed to work. I thought that we load types into the type db "on demand", that is if there is a type reference, it will cause the load into the db.
If we assume we preloaded everything, then please ignore my comment.

jkurdek:
Merging. In the future this should be updated with more detailed description of safe handles

PR #3060, opened 2025-03-30
[NativeAOT-LLVM] Wasm managed threads - build new nupkg

yowl:
This PR provides a base for working on multithread support, as and when we have Wasm runtimes that support threads.  With the browser this could be done now using Web Workers, the same as mono, for WASI with wasmtime we will need to wait for at least preview3 v2, possibly later.

No actual work for the dotnet runtime functionality that we will need, e.g. GC polling is included, this PR just creates CI jobs, runs the tests that passed, and creates 2 new packages for testing.


src/libraries/System.Runtime.InteropServices.JavaScript/src/System.Runtime.InteropServices.JavaScript.csproj
```diff
@@ -76,6 +76,11 @@
     <Compile Include="System\Runtime\InteropServices\JavaScript\JSWebWorker.cs" />
     <Compile Include="System\Runtime\InteropServices\JavaScript\JSSynchronizationContext.cs" />
     <Compile Include="System\Runtime\InteropServices\JavaScript\JSAsyncTaskScheduler.cs" />
+    <!-- Can't build System.Threading.Channels.csproj without these -->
```
jkotas: Why is that? This workaround does not look right.

yowl: Thanks for looking, it was as though even though these transitive dependencies were built, they weren't found for this project.  However it doesn't seem consistent and I removed this change now.

yowl:
Cc @dotnet/nativeaot-llvm

eng/pipelines/common/platform-matrix.yml
```diff
@@ -500,6 +500,44 @@ jobs:
         buildConfig: ${{ parameters.buildConfig }}
         ${{ insert }}: ${{ parameters.jobParameters }}
 
+- ${{ if containsValue(parameters.platforms, 'wasi_multithread_wasm_win') }}:
```
SingleAccretion: I don't think we need these new platforms; we should be able to use the existing ones like here: https://github.com/dotnet/runtime/blob/9e5e6aa7bc36aeb2a154709a9d1192030c30a2ef/eng/pipelines/common/templates/wasm-build-only.yml#L17-L36, utilizing the `nameSuffix` parameter to `global-build-job.yml`.

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -76,6 +76,26 @@ You should now be able to publish the project for Wasm: `dotnet publish -r brows
 
 Once you build the repo, you can use the produced binaries in one of four ways specified below ("Using built binaries", "Building packages", "Convenience Visual Studio "repro" project", "Running tests").
 
+## Building for Multithreaded packages
+
+This is a work in progress and far from functional.  Currentlty there exists just enough infrastucture to build packages for mutlithreaded runtime and libs, but they are not functional in the sense that they support multithreaded programs yet.  To build the WASI packages:
+```
+build clr.aot+libs+nativeaot.packages -c Debug -a wasm -os wasi -cmakeargs -DCLR_CMAKE_TARGET_OS_SUBGROUP=multithread '/p:WasmEnableThreads=true'
```
SingleAccretion: We should funnel `/p:WasmEnableThreads` down to cmake/clang automatically. E. g. the libs native build sets `CMAKE_USE_PTHREADS`, we can extended that pattern to the CoreCLR's `runtime.proj`.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -31,8 +31,17 @@ steps:
         displayName: Set up ILC emulation environment
 
     - ${{ if eq(parameters.archType, 'wasm') }}:
-      - script: $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) ${{ parameters.osGroup }} $(crossArg) $(_officialBuildParameter) ci tree nativeaot /p:LibrariesConfiguration=${{ parameters.librariesConfiguration }}
-        displayName: Build runtime tests
+      - ${{ if eq(parameters.platform, 'browser_multithread_wasm_win') }}:
```
SingleAccretion: It would be good to avoid this duplication. We can add a parameter to these post-build steps like `wasmEnableThreadsArg: /p:WasmEnableThreads=true` and use it here (and below to excluded the libs tests).

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -438,6 +438,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CompileWasmArgs Condition="'$(NativeDebugSymbols)' == 'true'">$(CompileWasmArgs) -g3</CompileWasmArgs>
       <CompileWasmArgs Condition="'$(WasmEnableNonTrappingFloatToIntConversions)' == 'true'">$(CompileWasmArgs) -mnontrapping-fptoint</CompileWasmArgs>
       <CompileWasmArgs Condition="'$(IlcLlvmExceptionHandlingModel)' == 'wasm'">$(CompileWasmArgs) -fwasm-exceptions</CompileWasmArgs>
+      <CompileWasmArgs Condition="'$(WasmEnableThreads)' == 'true'">$(CompileWasmArgs) -pthread -matomics -mbulk-memory</CompileWasmArgs>
```
SingleAccretion: You also need to modify the `IlcLlvmTarget` to include `-threads` I think.

src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
```diff
@@ -1151,6 +1151,8 @@ REDHAWK_PALEXPORT bool PalGetMaximumStackBounds(_Out_ void** ppStackLowOut, _Out
     {
 #if defined(HOST_WASM) && !defined(FEATURE_WASM_MANAGED_THREADS)
         PalGetMaximumStackBounds_SingleThreadedWasm(&pStackLowOut, &pStackHighOut);
+#elif defined(HOST_WASM) && defined(FEATURE_WASM_MANAGED_THREADS)
+        PalGetMaximumStackBounds_MultiThreadedWasm(&pStackLowOut, &pStackHighOut);
```
SingleAccretion: This should use the `pthread_attr_init` code from below instead of a bespoke version.

src/native/external/zlib-ng.cmake
```diff
@@ -26,8 +26,13 @@ if (CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI)
 
   # zlib-ng uses atomics, so we need to enable threads when requested for browser/wasi, otherwise the wasm target won't have thread support.
   if (CMAKE_USE_PTHREADS)
+    if (CLR_CMAKE_TARGET_BROWSER)  
       add_compile_options(-pthread)
       add_linker_flag(-pthread)
+    else()
+      add_compile_options(-D_WASI_EMULATED_PTHREAD)
+      add_linker_flag(-Wl,-lwasi-emulated-pthread)
```
SingleAccretion: Why didn't we need this before?

SingleAccretion (commented):
Just a bit of a skim so far, will continue tomorrow.

yowl:
What is it about `PackagingTests.csproj` that changes the way the target runtime package name is generated?

SingleAccretion:
> What is it about PackagingTests.csproj that changes the way the target runtime package name is generated?

The packaging test implementation project (`HelloWasm\PackagingTests\PackagingTestProject.csproj`) receives the things it needs from the parent project (`HelloWasm\PackagingTests.csproj`) via the `Test_` properties. The target package name, however, is actually not decided by these properties but by the target RID via nuget `runtime.json` magic. Since we don't have a target RID for multi-threaded WASM (a bit of a design flaw in the current scheme, IMO, but that ship has sailed long ago), we're in a bit of a pickle. Maybe disable the packaging test for MT and leave solving this problem for later?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: I didn't think there were any triples for threads, at least not yet.

eng/native/gen-buildsys.cmd
```diff
@@ -65,6 +65,11 @@ if /i "%__Arch%" == "wasm" (
         set "WASI_SDK_PATH=!WASI_SDK_PATH:\=/!"
         if not "!WASI_SDK_PATH:~-1!" == "/" set "WASI_SDK_PATH=!WASI_SDK_PATH!/"
         set __CmakeGenerator=Ninja
+        if defined __CMakeArgs (
+            if not "!__CMakeArgs:multithread=!" == "!__CMakeArgs!" (
+                set __ExtraCmakeParams=%__ExtraCmakeParams% -D_WASI_EMULATED_PTHREAD
+            )
+        )
```
SingleAccretion: Needed?

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -8,6 +8,7 @@ parameters:
   runtimeVariant: ''
   isOfficialBuild: false
   librariesConfiguration: Debug
+  wasmEnableThreadsArgg: ''
```
SingleAccretion:
```suggestion
  wasmEnableThreadsArg: ''
```

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -17,21 +18,25 @@ steps:
       displayName: Build the ILC and RyuJit cross-compilers
 
     # Build target packages (note: target libs already built).
-    - script: $(Build.SourcesDirectory)/build$(scriptExt) nativeaot.packages -os ${{ parameters.osGroup }} -a wasm -c $(buildConfigUpper) $(_officialBuildParameter) -ci
-      displayName: Build target packages
+    - ${{ if ne(parameters.nameSuffix, 'Multithreaded') }}:
```
SingleAccretion: Can use `wasmEnableThreadsArg` here.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -17,21 +18,25 @@ steps:
...
+        displayName: Build target multithread packages
 
     # Build host packages.
     - script: $(Build.SourcesDirectory)/build$(scriptExt) libs+nativeaot.packages -a $(hostedTargetArch) -c $(buildConfigUpper) -cross $(_officialBuildParameter) -ci
       displayName: Build host packages
 
   # Build coreclr native test output outside of official build
   - ${{ if ne(parameters.isOfficialBuild, true) }}:
-    - ${{ if and(eq(parameters.archType, 'wasm'), ne(parameters.nameSuffix, '')) }}:
+    - ${{ if and(eq(parameters.archType, 'wasm'), and(ne(parameters.nameSuffix, ''), ne(parameters.nameSuffix, 'Multithreaded'))) }}:
```
SingleAccretion: Better to use `wasmEnableThreadsArg != ''` here, the existing `nameSuffix` check is already a bit of a hack.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -49,7 +54,7 @@ steps:
         displayName: Run runtime tests
 
     # Don't compile/run the libraries tests with emulated ILC to save CI time/resources.
-    - ${{ if and(eq(parameters.archType, 'wasm'), eq(parameters.nameSuffix, '')) }}:
+    - ${{ if and(and(eq(parameters.archType, 'wasm'), eq(parameters.nameSuffix, '')), not(eq(parameters.nameSuffix, 'Multithreaded'))) }}:
```
SingleAccretion: Ditto with `wasmEnableThreadsArg`.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
SingleAccretion: There is `wasm32-wasi-threads`, but looks like we don't need anything for Emscripten indeed and `-pthread` is enough.

src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
```diff
@@ -1169,6 +1169,8 @@ REDHAWK_PALEXPORT bool PalGetMaximumStackBounds(_Out_ void** ppStackLowOut, _Out
         status = pthread_attr_get_np(thread, &attr);
 #elif HAVE_PTHREAD_GETATTR_NP
         status = pthread_getattr_np(thread, &attr);
+#elif defined(HOST_WASM) && defined(FEATURE_WASM_MANAGED_THREADS)
+        // We dont have a pthread_getattr_np, but so far we don't need it.
```
SingleAccretion: Do we not? Could you test whether our Emscripten version already has a working version of it, it needs `_GNU_SOURCE` defined to be available in the headers.

eng/pipelines/runtimelab.yml
```diff
@@ -71,6 +71,28 @@ extends:
...
+        #
+        - template: /eng/pipelines/common/platform-matrix.yml
+          parameters:
+            jobTemplate: /eng/pipelines/common/global-build-job.yml
+            helixQueuesTemplate: /eng/pipelines/coreclr/templates/helix-queues-setup.yml
+            buildConfig: debug
+            osSubgroup: multithread
+            platforms:
+            - browser_wasm_win
+            - wasi_wasm_win
```
SingleAccretion: How is this going to work with WASI, considering we don't have a `-threads` triple for WASIp2 yet? I would focus on making the fundamentals work on browser for now, we can enable WASI later when the libc support comes along.

src/coreclr/nativeaot/Runtime/wasm/PalRedhawkWasm.cpp
```diff
@@ -191,6 +191,186 @@ extern "C" int __cxa_thread_atexit(Dtor dtor, void* obj, void*)
...
+}
+
+//
+// Note that we return the native stack bounds here, not shadow stack ones. Currently this functionality is mainly
+// used for RuntimeHelpers.TryEnsureSufficientExecutionStack, and we do use the native stack in codegen, so this
+// is an acceptable approximation.
+//
+extern "C" unsigned char __stack_low;
+extern "C" unsigned char __stack_high;
+
```
SingleAccretion: There is a lot of code copied here and it's not clear what part of it is needed and what part is there because WASI doesn't actually have `-pthread` support yet.

src/coreclr/nativeaot/System.Private.CoreLib/src/System.Private.CoreLib.csproj
```diff
@@ -35,6 +35,11 @@
   <PropertyGroup Condition="'$(TargetsSingleThreadedWasm)' != 'true'">
     <FeaturePortableThreadPool>true</FeaturePortableThreadPool>
     <FeaturePortableTimer>true</FeaturePortableTimer>
+    <!-- LLVM-TODO: Remove FeatureWasmManagedThreads in favor of WasmEnableThreads. -->
+    <FeatureWasmManagedThreads Condition="'$(WasmEnableThreads)' == 'true'">true</FeatureWasmManagedThreads>
+    <FeatureWasmPerfTracing Condition="('$(TargetsBrowser)' == 'true' or '$(TargetsWasi)' == 'true') and ('$(WasmEnableThreads)' == 'true')">true</FeatureWasmPerfTracing>
```
SingleAccretion: Is `FeatureWasmPerfTracing` actually used for anything? We don't really have any tracing support...

src/coreclr/nativeaot/System.Private.CoreLib/src/System.Private.CoreLib.csproj
```diff
@@ -347,6 +352,8 @@
     <Compile Include="System\Diagnostics\StackFrame.NativeAot.Wasm.cs" />
     <Compile Include="System\Diagnostics\StackFrame.NativeAot.Wasi.cs" Condition="'$(TargetsWasi)' == 'true'" />
     <Compile Include="System\Diagnostics\StackFrame.NativeAot.Browser.cs" Condition="'$(TargetsBrowser)' == 'true'" />
+    <Compile Include="System\Threading\PortableThreadPool.NativeAot.Browser.cs" Condition="'$(TargetsBrowser)' == 'true' and '$(WasmEnableThreads)' == 'true'" />
+    <Compile Include="System\Threading\ThreadPool.NativeAot.Browser.cs" Condition="'$(TargetsBrowser)' == 'true' and '$(WasmEnableThreads)' == 'true'" />
```
SingleAccretion: Why are the portable thread pool files insufficient and we need these as well?

src/native/external/zlib-ng.cmake
```diff
@@ -26,8 +26,10 @@ if (CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI)
 
   # zlib-ng uses atomics, so we need to enable threads when requested for browser/wasi, otherwise the wasm target won't have thread support.
   if (CMAKE_USE_PTHREADS)
+    if (CLR_CMAKE_TARGET_BROWSER)  
       add_compile_options(-pthread)
       add_linker_flag(-pthread)
+    endif()
```
SingleAccretion: Per the above comment of not fixing WASI before it can be fixed properly, I think we can remove this condition as well.

SingleAccretion (commented):
Larger overall point on testing.

We currently don't have even the basic support for e. g. correct memory orderings, so stuff will fail due to race conditions randomly in CI. Should we disable all tests (and only do build) for now, before we know the configuration is at least not 100% known to be flaky? Alternatively, only enable like one test (e. g. HelloWasm)?

eng/pipelines/common/platform-matrix.yml
```diff
@@ -489,6 +489,7 @@ jobs:
       helixQueuesTemplate: ${{ parameters.helixQueuesTemplate }}
       variables: ${{ parameters.variables }}
       osGroup: wasi
+      osSubGroup: ${{ parameters.osSubGroup }}
```
SingleAccretion: Needed (also below)?

src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
yowl: Thanks, have updated this and removed `PalGetMaximumStackBounds_MultiThreadedWasm` with the logic move to `pthread_attr_init` and `pthread_attr_getstack`

src/native/external/zlib-ng.cmake
yowl: Seems we don't need it after all.  Removed.

yowl:
> Larger overall point on testing.
> 
> We currently don't have even the basic support for e. g. correct memory orderings, so stuff will fail due to race conditions randomly in CI. Should we disable all tests (and only do build) for now, before we know the configuration is at least not 100% known to be flaky? Alternatively, only enable like one test (e. g. HelloWasm)?

Ok, that's fine with me.  Will remove in next commit.

src/native/external/zlib-ng.cmake
yowl: cmake seems broken with this for wasi

```
  [vcvarsall.bat] Environment initialized for: 'arm64'
  Commencing build of native components

  Not searching for unused variables given on the command line.
  -- The C compiler identification is Clang 19.1.5
  -- Detecting C compiler ABI info
  -- Detecting C compiler ABI info - done
  -- Check for working C compiler: c:/github/wasi-sdk25/bin/clang.exe - skipped
  -- Detecting C compile features
  -- Detecting C compile features - done
  -- Performing Test COMPILER_SUPPORTS_W_RESERVED_IDENTIFIER
  -- Performing Test COMPILER_SUPPORTS_W_RESERVED_IDENTIFIER - Success
  -- Looking for include files sys/auxv.h, asm/hwcap.h
  -- Looking for include files sys/auxv.h, asm/hwcap.h - not found
  -- Looking for sysctlbyname
  -- Looking for sysctlbyname - not found
  -- Looking for arc4random_buf
  -- Looking for arc4random_buf - found
  -- Looking for O_CLOEXEC
  -- Looking for O_CLOEXEC - found
  -- Looking for include file windows.h
  -- Looking for include file windows.h - not found
  -- Looking for include files windows.h, bcrypt.h
  -- Looking for include files windows.h, bcrypt.h - not found
  -- Looking for clock_gettime_nsec_np
  -- Looking for clock_gettime_nsec_np - not found
  -- Using CMake version 4.0.1
  -- ZLIB_HEADER_VERSION: 1.3.1
  -- ZLIBNG_HEADER_VERSION: 2.2.1
  -- Arch detected: 'wasm32-wasip2'
  -- Basearch of 'wasm32-wasip2' has been detected as: 'wasm32'
  -- Using CMake toolchain: C:/github/wasi-sdk25/share/cmake/wasi-sdk-p2.cmake
  -- Performing Test FNO_LTO_AVAILABLE
  -- Performing Test FNO_LTO_AVAILABLE - Failed
  -- Looking for arm_acle.h
  -- Looking for arm_acle.h - not found
  -- Looking for sys/auxv.h
  -- Looking for sys/auxv.h - not found
  -- Looking for sys/sdt.h
  -- Looking for sys/sdt.h - not found
  -- Looking for unistd.h
  -- Looking for unistd.h - not found
  -- Looking for linux/auxvec.h
  -- Looking for linux/auxvec.h - not found
  -- Looking for sys/types.h
  -- Looking for sys/types.h - not found
  -- Looking for stdint.h
  -- Looking for stdint.h - not found
  -- Looking for stddef.h
  -- Looking for stddef.h - not found
  -- Check size of off64_t
  -- Check size of off64_t - failed
  -- Check size of _off64_t
  -- Check size of _off64_t - failed
  -- Check size of __off64_t
  -- Check size of __off64_t - failed
  -- Looking for fseeko
  -- Looking for fseeko - not found
  -- Looking for strerror
  -- Looking for strerror - not found
  -- Looking for posix_memalign
  -- Looking for posix_memalign - not found
  -- Performing Test HAVE_NO_INTERPOSITION
  -- Performing Test HAVE_NO_INTERPOSITION - Failed
  -- Performing Test HAVE_ATTRIBUTE_VISIBILITY_HIDDEN
  -- Performing Test HAVE_ATTRIBUTE_VISIBILITY_HIDDEN - Failed
  -- Performing Test HAVE_ATTRIBUTE_VISIBILITY_INTERNAL
  -- Performing Test HAVE_ATTRIBUTE_VISIBILITY_INTERNAL - Failed
  -- Performing Test HAVE_ATTRIBUTE_ALIGNED
  -- Performing Test HAVE_ATTRIBUTE_ALIGNED - Failed
  -- Performing Test HAVE_BUILTIN_ASSUME_ALIGNED
  -- Performing Test HAVE_BUILTIN_ASSUME_ALIGNED - Failed
  -- Performing Test HAVE_BUILTIN_CTZ
  -- Performing Test HAVE_BUILTIN_CTZ - Failed
  -- Performing Test HAVE_BUILTIN_CTZLL
  -- Performing Test HAVE_BUILTIN_CTZLL - Failed
  -- Performing Test HAVE_PTRDIFF_T
  -- Performing Test HAVE_PTRDIFF_T - Failed
  -- Check size of void *
  -- Check size of void * - failed
  -- sizeof(void *) is  bytes
  CMake Error at C:/github/runtimelab/src/native/external/zlib-ng/CMakeLists.txt:556 (message):
    sizeof(void *) is neither 32 nor 64 bit
```

SingleAccretion: I would expect this to be broken (in some way... maybe not in this exact way), this comment was packaged together with https://github.com/dotnet/runtimelab/pull/3060#discussion_r2208458502. I. e. we don't need this once we don't expect WASI to work with MT.

When we do the work to enable WASI+MT, this check will _need_ to go anyway, otherwise we'd be building zlib with atomics/locks stripped and that surely won't lead to anything good.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: Ok, I will comment it.  The problem is that cmake fails for zlib-ng with pthread as the libc is not build with threads (atomics, bulk-memory)

src/native/external/zlib-ng.cmake
yowl: I've left this in with a comment for now.  zlib-ng cannot be built with it as it would require a libc also with the same switches which doesn't exist for p2.

eng/pipelines/common/platform-matrix.yml
yowl: Have added it to some more naot wasm places.

SingleAccretion: Hmm, I meant that it is not clear it is needed... Since we already have a name suffix (the current name for the jobs has `multithread` duplicated: `Build wasimultithread-wasm windows debug Multithreaded`).

src/native/external/zlib-ng.cmake
SingleAccretion: To be clear, my proposal is to drop WASI from this change (completely), since we can't make it work right now.

eng/pipelines/common/platform-matrix.yml
yowl: I went this way (OsSubGroup) is that is what I saw from mono.   We could just use the namesuffix though, that would be fine with me,

src/native/external/zlib-ng.cmake
yowl: I see, its the only place we are running any tests though?

SingleAccretion: Well, if we can't make it work, what's the point of running any tests on it? We will have browser still for MT testing. Single-threaded WASI should of course stay as-is.

yowl: We aren't running any MT browser tests, they have different problems.  Running the test serves to check the compilation succeeds, and the result passes the test.  It might not be multithreaded or correct yet, but it runs and passes the test.

yowl: What we could do is keep this draft until we get a multithreaded wasi ?

SingleAccretion: >  Running the test serves to check the compilation succeeds, and the result passes the test. It might not be multithreaded or correct yet, but it runs and passes the test.

It is perfectly fine to have build-only validation for this work for now. It would be perfectly fine to only start enabling the tests once we do all the "significant" work like runtime suspension, fix all the missing ifdefs, etc. It is fine because we know that at the end of it we'll get to something that works.

I don't see any reason to try and maintain this "hybrid" WASI configuration that is "kind of MT but actually single-threaded" (since we can't actually create pthreads with it). It's not a useful intermediate state since it has no future.

> What we could do is keep this draft until we get a multithreaded wasi ?

I don't think we need to, unless you want to, e. g. for ease of development reasons?

yowl: I think I get it, we leave the browser build of the tests, but dont run them.  And remove the wasi MT build as far as it means removing hacks like this?

SingleAccretion: Yep.

src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
yowl: As we are not enabling a wasi build now we can remove this.

eng/pipelines/runtimelab.yml
yowl: removed.

src/coreclr/nativeaot/Runtime/wasm/PalRedhawkWasm.cpp
yowl: removed

src/coreclr/nativeaot/System.Private.CoreLib/src/System.Private.CoreLib.csproj
yowl: Removed.

src/coreclr/nativeaot/System.Private.CoreLib/src/System.Private.CoreLib.csproj
yowl: The problem is this line
https://github.com/dotnet/runtimelab/blob/79c861e8e1abb8e86b65ebc74b8723c135dd66cf/src/libraries/System.Private.CoreLib/src/System/Threading/ThreadPoolWorkQueue.cs#L1169
As `YieldFromDispatchLoop` is internal perhaps it should really be in `PortableThreadPool`, but that would be a larger question.

eng/pipelines/common/platform-matrix.yml
yowl: Removed now.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -26,13 +27,14 @@ steps:
 
   # Build coreclr native test output outside of official build
   - ${{ if ne(parameters.isOfficialBuild, true) }}:
-    - ${{ if and(eq(parameters.archType, 'wasm'), ne(parameters.nameSuffix, '')) }}:
+    - ${{ if and(eq(parameters.archType, 'wasm'), and(ne(parameters.nameSuffix, ''), eq(parameters.wasmEnableThreadsArg, ''))) }}:
```
SingleAccretion: BTW, it turns out `and` is a vararg operator, e. g. `and(a, b, c)` is meant to work.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -26,13 +27,14 @@ steps:
...
       - script: pwsh $(Build.SourcesDirectory)/eng/pipelines/runtimelab/set-ilc-emulation-environment.ps1 -Arch $(hostedTargetArch)
         displayName: Set up ILC emulation environment
 
     - ${{ if eq(parameters.archType, 'wasm') }}:
-      - script: $(Build.SourcesDirectory)/src/tests/build$(scriptExt) nativeaot $(buildConfigUpper) ${{ parameters.osGroup }} $(crossArg) $(_officialBuildParameter) ci tree nativeaot /p:LibrariesConfiguration=${{ parameters.librariesConfiguration }}
-        displayName: Build runtime tests
+      - ${{ if not(and(eq(parameters.osGroup, 'wasi'), ne(parameters.wasmEnableThreadsArg, ''))) }}:
```
SingleAccretion: Should this condition be dropped now that we don't have WASI?

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -42,14 +44,15 @@ steps:
           displayName: Build runtime tests
 
     - ${{ if contains(parameters.platform, 'win') }}:
-      - script: $(Build.SourcesDirectory)/src/tests/run$(scriptExt) runnativeaottests $(buildConfigUpper) ${{ parameters.archType }} ${{ parameters.osGroup }}
-        displayName: Run runtime tests
+      - ${{ if not(and(eq(parameters.osGroup, 'browser'), ne(parameters.wasmEnableThreadsArg, ''))) }}:
```
SingleAccretion:
```suggestion
      - ${{ if eq(parameters.wasmEnableThreadsArg, '') }}:
```

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -60,4 +63,4 @@ steps:
   - ${{ if eq(parameters.uploadIntermediateArtifacts, true) }}:
     - template: /eng/pipelines/common/upload-intermediate-artifacts-step.yml
       parameters:
-        name: ${{ parameters.platform }}${{ parameters.nameSuffix }}
+        name: ${{ parameters.platform }}${{ parameters.nameSuffix }}
```
SingleAccretion: Accidental diff?

src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp
```diff
@@ -1169,6 +1169,8 @@ REDHAWK_PALEXPORT bool PalGetMaximumStackBounds(_Out_ void** ppStackLowOut, _Out
         status = pthread_attr_get_np(thread, &attr);
 #elif HAVE_PTHREAD_GETATTR_NP
         status = pthread_getattr_np(thread, &attr);
+// #elif defined(HOST_WASM) && defined(FEATURE_WASM_MANAGED_THREADS)
+//         // We dont have a pthread_getattr_np, but so far we don't need it.
```
SingleAccretion:
```suggestion
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System.Private.CoreLib.csproj
```diff
@@ -35,6 +35,10 @@
   <PropertyGroup Condition="'$(TargetsSingleThreadedWasm)' != 'true'">
     <FeaturePortableThreadPool>true</FeaturePortableThreadPool>
     <FeaturePortableTimer>true</FeaturePortableTimer>
+    <!-- LLVM-TODO: Remove FeatureWasmManagedThreads in favor of WasmEnableThreads. -->
+    <FeatureWasmManagedThreads Condition="'$(WasmEnableThreads)' == 'true'">true</FeatureWasmManagedThreads>
+    <DefineConstants Condition="'$(WasmEnableThreads)' == 'true'">$(DefineConstants);FEATURE_WASM_MANAGED_THREADS</DefineConstants>
+    <DefineConstants Condition="'$(FeatureWasmPerfTracing)' == 'true'">$(DefineConstants);FEATURE_WASM_PERFTRACING</DefineConstants>
```
SingleAccretion:
```suggestion
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System.Private.CoreLib.csproj
SingleAccretion: I think we should follow the pattern of stealing Mono files for these (as below). <https://github.com/dotnet/runtime/blob/d817f032075f77c7852b9f653c064a3346948667/src/mono/System.Private.CoreLib/src/System/Threading/ThreadPool.Browser.Threads.Mono.cs#L10> for `YieldFromDispatchLoop`, <https://github.com/dotnet/runtime/blob/d817f032075f77c7852b9f653c064a3346948667/src/mono/System.Private.CoreLib/src/System/Threading/PortableThreadPool.Browser.Threads.Mono.cs> for the other two properties.

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.csproj
```diff
@@ -10,6 +10,7 @@
     <RequiresProcessIsolation>true</RequiresProcessIsolation>
     <ReferenceXUnitWrapperGenerator>false</ReferenceXUnitWrapperGenerator>
     <CLRTestTargetUnsupported Condition="'$(TargetsWasm)' != 'true'">true</CLRTestTargetUnsupported>
+    <DefineConstants Condition="'$(TargetsWasm)' == 'true' and '$(WasmEnableThreads)' == 'true'">$(DefineConstants);NO_EXPLICIT_FINALIZER</DefineConstants>
```
SingleAccretion:
```suggestion
```

src/tests/nativeaot/SmokeTests/HelloWasm/HelloWasm.cs
```diff
@@ -40,7 +40,9 @@ private static unsafe int Main(string[] args)
         TestMetaData();
 
         TestGC();
+#if !NO_EXPLICIT_FINALIZER
         TestFinalization();
+#endif
```
SingleAccretion: It's best to fix the finalizer in some following change so that the test doesn't hang instead (it'll need to be done eventually anyway) (or it may end up being the case that the test itseld needs fixing).
```suggestion
        TestFinalization();
```

eng/pipelines/runtimelab.yml
```diff
@@ -71,6 +71,27 @@ extends:
...
 
+        #
+        # Build and test Wasm Debug multithreaded libraries and Debug runtime
+        #
+        - template: /eng/pipelines/common/platform-matrix.yml
+          parameters:
+            jobTemplate: /eng/pipelines/common/global-build-job.yml
+            helixQueuesTemplate: /eng/pipelines/coreclr/templates/helix-queues-setup.yml
+            buildConfig: debug
+            osSubgroup: multithread
```
SingleAccretion:
```suggestion
```

eng/pipelines/common/platform-matrix.yml
SingleAccretion:
```suggestion
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -604,6 +605,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Include="-s MAXIMUM_MEMORY=$(EmccMaximumHeapSize)" Condition="'$(EmccMaximumHeapSize)' != ''" />
       <CustomLinkerArg Include="-s INITIAL_MEMORY=$(EmccInitialHeapSize)" Condition="'$(EmccInitialHeapSize)' != ''" />
       <CustomLinkerArg Include="-s ASSERTIONS=1" Condition="'$(EmccEnableAssertions)' == 'true'" />
+      <CustomLinkerArg Condition="'$(WasmEnableThreads)' == 'true'" Include="-pthread" />
```
SingleAccretion: I saw earlier Emscripten gives some warning like "JS will be slow with memory growth + threads", seems we should disable it.

src/coreclr/nativeaot/Runtime/wasm/PalRedhawkWasm.h
```diff
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#ifndef FEATURE_WASM_MANAGED_THREADS
+#ifdef FEATURE_WASM_MANAGED_THREADS
+int __cxa_thread_atexit(void (*func)(), void *obj, void *dso_symbol);
+#else
```
SingleAccretion:
```suggestion
#ifndef FEATURE_WASM_MANAGED_THREADS
```
Unused.

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -76,6 +76,19 @@ You should now be able to publish the project for Wasm: `dotnet publish -r brows
...
+## Building for Multithreaded packages
+
+This is a work in progress and far from functional.  Currentlty there exists just enough infrastucture to build packages for mutlithreaded runtime and libs, but they are not functional in the sense that they support multithreaded programs yet.
+To build the browser multithreaded packages:
+```
+build clr.aot+libs+nativeaot.packages -c Debug -a wasm -os browser '/p:WasmEnableThreads=true'
+```
+To build the runtime tests for browser
+```
+src\tests\build nativeaot Debug wasm tree nativeaot browser /p:LibrariesConfiguration=debug /p:TestWrapperTargetsWindows=true /p:WasmEnableThreads=true
```
```suggestion
src\tests\build nativeaot Debug wasm tree nativeaot browser /p:LibrariesConfiguration=debug /p:WasmEnableThreads=true
```
Or is it needed for some reason?

SingleAccretion:
The python interpreter issue looks as annoying as ever, maybe we can do a little patch in `runtimelab\src\coreclr\nativeaot\Runtime\eventpipe\CMakeLists.txt`, like:
```diff
include(FindPython)

+# Workaround from https://github.com/dotnet/runtime/pull/117513
+if (NOT "$ENV{EMSDK_PYTHON}" STREQUAL "")
+  set(Python_EXECUTABLE "$ENV{EMSDK_PYTHON}")
+endif()
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System.Private.CoreLib.csproj
jkotas: You can upstream fixes to move shareable files under libraries.

Prior art: https://github.com/dotnet/runtime/pull/114548/

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: disabled growth and `MAXIMUM MEMORY` for threads.

SingleAccretion: Hmm, I only meant to disable the warning. I don't think we can disable memory growth...

eng/pipelines/common/platform-matrix.yml
yowl: done

docs/workflow/building/coreclr/nativeaot.md
yowl: done

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
yowl: done

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: just pthread is left now, thanks.

eng/native/gen-buildsys.cmd
yowl: removed

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: lol, I was thinking we would block users from running slow code.

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
```diff
@@ -42,14 +43,15 @@ steps:
...
+      - ${{ if eq(parameters.wasmEnableThreadsArg, '') }}:
+        - script: $(Build.SourcesDirectory)/src/tests/run$(scriptExt) runnativeaottests $(buildConfigUpper) ${{ parameters.archType }} ${{ parameters.osGroup }}
+          displayName: Run runtime tests
     - ${{ else }}:
       - script: $(Build.SourcesDirectory)/src/tests/run$(scriptExt) --runnativeaottests $(buildConfigUpper) ${{ parameters.archType }} ${{ parameters.osGroup }}
         displayName: Run runtime tests
 
     # Don't compile/run the libraries tests with emulated ILC to save CI time/resources.
-    - ${{ if and(eq(parameters.archType, 'wasm'), eq(parameters.nameSuffix, '')) }}:
+    - ${{ if and(eq(parameters.archType, 'wasm'), eq(parameters.wasmEnableThreadsArg, '')) }}:
```
SingleAccretion:
```suggestion
    - ${{ if and(eq(parameters.archType, 'wasm'), eq(parameters.nameSuffix, ''), eq(parameters.wasmEnableThreadsArg, '')) }}:
```
(Restoring the check for emulated ILC.)

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -601,9 +602,11 @@ The .NET Foundation licenses this file to you under the MIT license.
       <CustomLinkerArg Include="-s ERROR_ON_UNDEFINED_SYMBOLS=0" Condition="'$(IlcTreatWarningsAsErrors)' != 'true'" />
       <CustomLinkerArg Include="-s WASM_BIGINT=1" Condition="'$(WasmEnableJSBigIntIntegration)' == 'true'" />
       <CustomLinkerArg Include="-s DISABLE_EXCEPTION_CATCHING=0" Condition="'$(IlcLlvmExceptionHandlingModel)' == 'cpp'" />
-      <CustomLinkerArg Include="-s MAXIMUM_MEMORY=$(EmccMaximumHeapSize)" Condition="'$(EmccMaximumHeapSize)' != ''" />
+      <CustomLinkerArg Include="-s MAXIMUM_MEMORY=$(EmccMaximumHeapSize)" Condition="'$(EmccMaximumHeapSize)' != ''"/>
```
```suggestion
      <CustomLinkerArg Include="-s MAXIMUM_MEMORY=$(EmccMaximumHeapSize)" Condition="'$(EmccMaximumHeapSize)' != ''" />
```

SingleAccretion (approved):
LGTM aside from the last comments. Thank you!

eng/pipelines/runtimelab/runtimelab-post-build-steps.yml
yowl: Thanks

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
yowl: sorry, missed that.  Thanks

docs/workflow/building/coreclr/nativeaot.md
```diff
@@ -76,6 +76,19 @@ You should now be able to publish the project for Wasm: `dotnet publish -r brows
 
 Once you build the repo, you can use the produced binaries in one of four ways specified below ("Using built binaries", "Building packages", "Convenience Visual Studio "repro" project", "Running tests").
 
+## Building for Multithreaded packages
+
+This is a work in progress and far from functional.  Currentlty there exists just enough infrastucture to build packages for mutlithreaded runtime and libs, but they are not functional in the sense that they support multithreaded programs yet.
```
jkotas:
```suggestion
This is a work in progress and far from functional.  Currently there exists just enough infrastructure to build packages for multithreaded runtime and libs, but they are not functional in the sense that they support multithreaded programs yet.
```

yowl:
Thanks all, I think that's it for now.

jkotas:
Thanks!

PR #3062, opened 2025-03-31
Use prototype roslyn compiler

agocke:
Everything under `async` will use the prototype compiler. All async methods will be compiled as async by default (the feature flag is enabled). The attribute definition for the tests is included, which can be used to selectively disable async2.


agocke:
Note: this currently fails due to the compiler throwing

src/tests/async/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <Import Project="../Directory.Build.props" />
+  <PropertyGroup>
+    <Features>$(Features);runtime-async=on</Features>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)RuntimeAsyncMethodGenerationAttribute.cs" />
+  </ItemGroup>
+</Project>
```
VSadov, line 9: I think switching all tests to experimental compiler is good for runtimelab, but in main repo we might need to limit this to async folder. 

I assume this will work in CI, right?

VSadov:
I looked through the changes and PR looks reasonable. 
I did not try it locally, since we have a known issue with `Task.Yield`. We'd need to fix that before merging. 

I think that can be handled by making an async1 wrapper over `Task.Yield`, but maybe it is easier to wait for a Roslyn fix.

agocke:
Closing in favor of the runtime PR: https://github.com/dotnet/runtime/pull/115531

PR #3063, opened 2025-04-01
[RuntimeAsync] Clean up and document JIT parts

jakobbotsch:
- Remove `CLASSID_SYSTEM_BYTE` that was added early as part of async2 but no longer used
- Remove unnecessary code
- Split async2 transformation into smaller documented chunks
- Add function headers for all missing functions
- Add missing license headers

No functional changes.


jakobbotsch:
cc @VSadov

src/coreclr/jit/codegencommon.cpp
```diff
@@ -7343,6 +7354,13 @@ void CodeGen::genMarkReturnGCInfo()
     }
 }
 
+//------------------------------------------------------------------------
+// genCodeForAsyncContinuation:
+//   Generate code for a GC_ASYNC_CONTINUATION node.
```
VSadov:
```suggestion
//   Generate code for a GT_ASYNC_CONTINUATION node.
```

src/coreclr/jit/async.cpp
```diff
@@ -26,158 +21,234 @@ class AsyncLiveness
...
-        }
+    LclVarDsc* dsc = m_comp->lvaGetDesc(lclNum);
 
-        LclVarDsc* dsc = m_comp->lvaGetDesc(lclNum);
+    if ((dsc->TypeGet() == TYP_BYREF) || ((dsc->TypeGet() == TYP_STRUCT) && dsc->GetLayout()->HasGCByRef()))
+    {
+        // Even if these are address exposed we expect them to be dead at
+        // suspension points. TODO: It would be good to somehow verify these
+        // aren't obviously live, if the JIT creates live ranges that span a
+        // suspension point then this makes it quite hard to diagnose that.
```
VSadov, line 156: I assume this comment reflects current implementation and will change when "zero out ref/ref-like" is implemented.

src/coreclr/jit/async.cpp
```diff
@@ -240,6 +314,9 @@ PhaseStatus Async2Transformation::Run()
 
     AsyncLiveness liveness(m_comp, m_comp->opts.OptimizationEnabled());
 
+    // Now walk the IR for all the blocks that contain async2 calls. Keep track
+    // of liveness and outstanding LIR edges as we go; the LIR edges that cross
+    // async2 calls are additional live variables that must be spilled.
```
VSadov, line 319: Can we end up spilling byrefs here?

src/coreclr/jit/async.cpp
jakobbotsch: Correct (this comment isn't new, this function was just moved)

src/coreclr/jit/async.cpp
jakobbotsch: Yes, but with the existing scheme that was basically undefined behavior. With the new scheme we'll have to substitute zero for these edges.

jakobbotsch: We should probably also have a small wording change in the spec that indicates that IL stack temporaries containing byrefs are also zeroed around async calls.

VSadov: Right. IL stack is the same as locals. Maybe that can be unified somehow in the spec.

VSadov: Here I was more concerned about JIT itself introducing byref values that cross awaits. 
I do not have a good sense of how early async rewrite happens. If it happens very early we only need to be concerned with what is in IL, and after capture/restore is introduced, it will disallow optimizations that keeping byref temps to locals across awaits.  

Although I still wonder what happens if byref temps do not point to locals - like point to a threadstatic.

src/coreclr/jit/importer.cpp
```diff
@@ -9095,7 +9095,7 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                     // TODO: The configVal should be wired to the actual implementation
                     //       that control the flow of sync context.
                     //       We do not have that yet.
-                    int configVal= -1;  // -1 not congigured, 0/1 configured to false/true
+                    int configVal = -1; // -1 not congigured, 0/1 configured to false/true
```
VSadov:
```suggestion
                    int configVal = -1; // -1 not configured, 0/1 configured to false/true
```

src/coreclr/jit/async.cpp
```diff
@@ -836,330 +1196,387 @@ void Async2Transformation::Transform(
...
-    BasicBlock* storeResultBB         = resumeBB;
+    // Copy call return value.
+    if (layout.ReturnSize > 0)
+    {
+        CopyReturnValueOnResumption(call, callDefInfo, resumeByteArrLclNum, resumeObjectArrLclNum, layout,
+                                    storeResultBB);
+    }
 
-    if (gcRefsCount > 0)
+    return resumeBB;
```
VSadov, line 1229: Perhaps worth assigning null to data array fields after deserializing them. The dispatcher will keep the current continuation alive until we return (since it needs `.Next`). That may take a while, but we no longer need the arrays, so no need to keep them alive.

src/coreclr/jit/async.cpp
```diff
@@ -836,330 +1196,387 @@ void Async2Transformation::Transform(
...
-                GenTree* objectArr = m_comp->gtNewLclvNode(resumeObjectArrLclNum, TYP_REF);
-                unsigned offset    = OFFSETOF__CORINFO_Array__data + (inf.GCDataIndex * TARGET_POINTER_SIZE);
-                GenTree* value     = LoadFromOffset(objectArr, offset, TYP_REF);
-                GenTree* store     = m_comp->gtNewStoreLclVarNode(inf.LclNum, value);
+        GenTree* store;
+        if (dsc->IsImplicitByRef())
+        {
+            GenTree* baseAddr = m_comp->gtNewLclvNode(inf.LclNum, dsc->TypeGet());
+            // TODO-CQ: Incoming data has no non-null GC refs, so this does not need write barriers.
+            // Backend does not handle GTF_IND_TGT_NOT_HEAP for STORE_BLK.
```
VSadov: Besides, this is actually `NOT_HEAP`, right?

src/coreclr/jit/async.cpp
```diff
@@ -1173,6 +1590,18 @@ GenTreeIndir* Async2Transformation::LoadFromOffset(GenTree*     base,
     return load;
 }
 
+//------------------------------------------------------------------------
+// Async2Transformation::LoadFromOffset:
```
VSadov:
```suggestion
// Async2Transformation::StoreAtOffset:
```

src/coreclr/jit/layout.cpp
```diff
@@ -536,6 +536,12 @@ ClassLayout* ClassLayout::Create(Compiler* compiler, const ClassLayoutBuilder& b
     return newLayout;
 }
 
+//------------------------------------------------------------------------
+// HasGCByRef: //   Check if this classlayout has a TYP_BYREF GC pointer in it.
+//
+// Return value:
```
VSadov, line 542: Should use `ClassLayout::IsStackOnly` instead?

src/coreclr/jit/async.cpp
VSadov: Since this is mostly refactoring, it would be ok to just add a comment or put this onto the "suggestions" pile.

src/coreclr/jit/layout.cpp
VSadov: Although this is probably not a PR for considering such changes.

VSadov (approved):
LGTM. Some parts are much more clear now!

src/coreclr/jit/async.cpp
VSadov: Importer transformations must run before this, since we rely on them, so perhaps need some care about introducing byref locals. 
It has this for example, but not sure if can be a problem:

```c
        // Exact method descriptor as passed in
        ctxTree = gtNewLclvNode(impInlineRoot()->info.compTypeCtxtArg, TYP_I_IMPL);
```

jakobbotsch: > Here I was more concerned about JIT itself introducing byref values that cross awaits.

CSE is one source of these (CSE is also utilized by loop hoisting). That's why CSE has special logic that makes byref CSEs unavailable across suspension points.

Loop opts introduced in .NET 9 probably also needs some handling now that I think about it.

Object stack allocation is also disabled for async2 because it introduces stack allocations and takes the address of it. With some work we could do object stack allocation in cases where all definitions/uses of the stack allocated object are visible.

I wouldn't be surprised if there are other cases, but I would be somewhat surprised if these cases are tree temps/LIR edges and not locals. In any case, if we do this through the LIR edge path we will hit this assert when we try to lay out the continuation:

https://github.com/dotnet/runtimelab/blob/94f8e5ab15f97aed6a1fd1e1ff1a8010ee858b64/src/coreclr/jit/async.cpp#L405

> I do not have a good sense of how early async rewrite happens. If it happens very early we only need to be concerned with what is in IL, and after capture/restore is introduced, it will disallow optimizations that keeping byref temps to locals across awaits.

The async rewrite happens very late. If this happened early, it would inhibit most optimizations to a significant degree because it introduces spurious data flow due to resumption. Currently most optimizations get to optimize the function as if there is no suspension/resumption and the fact that it is an async method is not even visible to them.
It means we need to teach the JIT in various places that it cannot introduce byrefs that live across suspension points, and it cannot take the address of a local across a suspension point.

> Although I still wonder what happens if byref temps do not point to locals - like point to a threadstatic.

We will hit the assert above.

> It has this for example, but not sure if can be a problem:
> 
> ```c
>         // Exact method descriptor as passed in
>         ctxTree = gtNewLclvNode(impInlineRoot()->info.compTypeCtxtArg, TYP_I_IMPL);
> ```

`TYP_I_IMPL` is a native pointer, so this shouldn't be a problem.

src/coreclr/jit/async.cpp
jakobbotsch: Yes, let me update this. It's a bit subtle because the local can be an implicit byref parameter that we actually type as `TYP_BYREF` today, even though they are always stack-only (this issue is https://github.com/dotnet/runtime/issues/82166).

src/coreclr/jit/layout.cpp
jakobbotsch: Yeah, once we implement the zeroing behavior this should be switched to check for whether the layout is a byref struct since that's what we settled on.

src/coreclr/jit/async.cpp
jakobbotsch: I think we can make it so that `DispatchContinuations` doesn't keep the strong reference to `continuation` beyond the resumption call.

jakobbotsch: I opened https://github.com/dotnet/runtimelab/issues/3064

src/coreclr/jit/async.cpp
VSadov: > The async rewrite happens very late. If this happened early, it would inhibit most optimizations to a significant degree because it introduces spurious data flow due to resumption.

Right, even though await formally destroys byrefs/reflikes, it has no effect on byvals, which are the majority. 
It makes sense to do the transform late so that value-based optimizations continue working, but need somehow prevent anything that would rely on byrefs being stable across await.

PR #3067, opened 2025-04-02
[RuntimeAsync] Fix handling of implicit byref liveness

### LINKED ISSUE #3043: [RuntimeAsync] Some issues with preserving value of a nullable struct with mix of ref/value fields across a suspension point.
I am not completely sure if this is related to Nullable. Nullable could be special if there was boxing, but this is just a local variable losing its value after Yield. No boxing is involved in this case.
Possibly it is a more general issue with liveness/storing/restoring of mixed value/ref fields of a struct, just a repro happened to involve nullable.

== To repro:
 In `shared-generics.cs` uncomment 2 scenarios marked with 
`// TODO: uncomment to repro https://github.com/dotnet/runtimelab/issues/3043` and run the test.

The test will fail. It looks like we lose the value here:
```cs
    public static async2 Task<T> StaticReturnClassType(T value)
    {
// value != null 
        await Task.Yield();
// value == null
        return value;
    }
```

Also, I think, while modifying the test with `Console.WriteLine(value)` in various places to see where exactly the value is lost, I think I have seen the scenario to pass, but lost track to what shape the test was and can't make it pass again. There is a thought that we may be dealing with uninitialized data, maybe...

jakobbotsch:
This scenario was handled in the first version of the prototype, so the rest of the support is already there to handle these properly.
It looks like I broke the liveness part of the support in 34897d7c9401e412857a2c98c4de2cb7c0fa7a19 when I introduced the byref check.

Fix #3043

### DIFF
--- a/src/coreclr/jit/async.cpp
+++ b/src/coreclr/jit/async.cpp
@@ -148,7 +148,8 @@ bool AsyncLiveness::IsLive(unsigned lclNum)
 
     LclVarDsc* dsc = m_comp->lvaGetDesc(lclNum);
 
-    if ((dsc->TypeGet() == TYP_BYREF) || ((dsc->TypeGet() == TYP_STRUCT) && dsc->GetLayout()->HasGCByRef()))
+    if (((dsc->TypeGet() == TYP_BYREF) && !dsc->IsImplicitByRef()) ||
+        ((dsc->TypeGet() == TYP_STRUCT) && dsc->GetLayout()->HasGCByRef()))
     {
         // Even if these are address exposed we expect them to be dead at
         // suspension points. TODO: It would be good to somehow verify these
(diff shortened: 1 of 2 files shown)


jakobbotsch:
cc @VSadov

VSadov:
We have couple test scenario commented out because of the issue. I think we should uncomment them now. (I tested, they are passing with the fix).

VSadov:
I've just pushed the test changes

VSadov (approved):
LGTM. Thanks!!

jakobbotsch:
> We have couple test scenario commented out because of the issue. I think we should uncomment them now. (I tested, they are passing with the fix).

Ah yes, I forgot to commit this change. Thanks.

PR #3072, opened 2025-04-03
[RuntimeAsync] Use explicit store types

jakobbotsch:
The type should not be determined from the incoming node: for small types this does not make sense.

Diff of the suspension code from the test:
```diff
@@ -1,19 +1,19 @@
 G_M5761_IG05:  ;; offset=0x00DB
        xor      edx, edx
        mov      r8d, 1
        call     [CORINFO_HELP_ALLOC_CONTINUATION]
        mov      rcx, 0x7FFAA7E20720      ; code for (dynamicClass):IL_STUB_AsyncResume_SmallType_Optimized(System.Object):System.Object
        mov      qword ptr [rax+0x20], rcx
        xor      ecx, ecx
        mov      qword ptr [rax+0x28], rcx
        mov      rcx, gword ptr [rax+0x10]
-       mov      dword ptr [rcx+0x10], ebx
+       mov      byte  ptr [rcx+0x10], bl
        mov      rcx, rax
```
The base technically writes beyond the end of the memory that was allocated, though in practice for coreclr that doesn't cause any buffer overflow.

### DIFF
--- a/src/coreclr/jit/async.cpp
+++ b/src/coreclr/jit/async.cpp
@@ -828,14 +828,14 @@ BasicBlock* Async2Transformation::CreateSuspension(BasicBlock*               blo
     GenTree* newContinuation = m_comp->gtNewLclvNode(m_newContinuationVar, TYP_REF);
     unsigned resumeOffset    = m_comp->info.compCompHnd->getFieldOffset(m_async2Info.continuationResumeFldHnd);
     GenTree* resumeStubAddr  = CreateResumptionStubAddrTree();
-    GenTree* storeResume     = StoreAtOffset(newContinuation, resumeOffset, resumeStubAddr);
+    GenTree* storeResume     = StoreAtOffset(newContinuation, resumeOffset, resumeStubAddr, TYP_I_IMPL);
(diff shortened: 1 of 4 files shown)


jakobbotsch:
cc @VSadov

src/coreclr/jit/async.cpp
```diff
@@ -1093,7 +1093,7 @@ void Async2Transformation::FillInDataOnSuspension(const jitstd::vector<LiveLocal
         }
         else
         {
-            store = StoreAtOffset(byteArr, offset, value);
+            store = StoreAtOffset(byteArr, offset, value, dsc->TypeGet());
```
jakobbotsch, line 1096: The actual fix.

VSadov (approved):
LGTM. Thanks!

VSadov:
> The base technically writes beyond the end of the memory that was allocated, though in practice for coreclr that doesn't cause any buffer overflow.

Is that because allocated memory would be aligned on pointer size?

jakobbotsch:
> > The base technically writes beyond the end of the memory that was allocated, though in practice for coreclr that doesn't cause any buffer overflow.
> 
> Is that because allocated memory would be aligned on pointer size?

I believe so, at least.

PR #3079, opened 2025-04-09
[RuntimeAsync] Merge from upstream

jakobbotsch:
Fixes:
- `BuildResumptionStubCalliSignature` was building a signature with `HASTHIS | EXPLICITTHIS` calling convention always, even for signatures without any 'this'. With https://github.com/dotnet/runtime/issues/113666 this broke as we started reordering the first argument with ret buffers.
- Update JIT interface `getEHinfo` to handle getting EH clause information when trying to inline methods with transient IL, like async thunks
- Update for new refactored jitinterface classes
- Update for removed `g_isNewExceptionHandlingEnabled`

cc @vsadov


VSadov:
x64 builds and passes runtime tests.

For x86 I see build errors. 

```
  E:\A2\runtimelab\src\coreclr\vm\jithelpers.cpp(1693): error C2039: 'InsureInit': is not a member of 'HelperMethodFrame'
  E:\A2\runtimelab\src\coreclr\vm\frames.h(1204): note: see declaration of 'HelperMethodFrame'
  E:\A2\runtimelab\src\coreclr\inc\jithelpers.h(326): error C2440: 'type cast': cannot convert from 'overloaded-function' to 'LPVOID'
  E:\A2\runtimelab\src\coreclr\inc\jithelpers.h(326): note: Context does not allow for disambiguation of overloaded function
  E:\A2\runtimelab\src\coreclr\vm\jithelpers.cpp(3461): error C2737: 'hlpFuncTable': const object must be initialized
```

I think I have fixes for this in my other PR. 
I'll cherry-pick the fixes and push here.

VSadov:
x86 builds, but a few tests fail - for example `JIT\Directed\arglist\vararg_TargetWindows`

I think I have a fix for that too

VSadov:
now only R2RDump fails on x86  (`readytorun\r2rdump\FrameworkTests\R2RDumpTests\R2RDumpTests.cmd`)

I think it is ok to merge and take a look at that later.

VSadov (approved):
LGTM. Thanks!

jakobbotsch:
Thanks for fixing that!

PR #3080, opened 2025-04-10
[RuntimeAsync] Remove new OSR transition helper

jakobbotsch:
With https://github.com/dotnet/runtime/pull/114381 we can just use one of the existing OSR resumption helpers from the resume code.


jakobbotsch:
cc @VSadov

VSadov (approved):
LGTM!

PR #3082, opened 2025-04-12
[RuntimeAsync] Do not rely on signature parsing to figure if a call is Async

VSadov:
I think parsing the signature to figure Async-ness is now mostly a vestige from times where we could have such signatures in metadata. And that was a requirement for the non-JIT based prototype. None of that is relevant anymore.

Parsing the signature is also less reliable than `IsAsyncMethod()` as we now have Async helpers that do not have the signature that parses as Async, so we need to work around that by matching helpers by name to force the right call convention. But since they are `MethodImpl.Async`, we already classify these helpers as async. Let’s use `IsAsyncMethod()` for helpers and everything else.

With this change only await helpers that participate in pattern-matching optimizations need to be `[Intrinsic]`. Methods like `..AwaitAwaiter...` do not need to be intrinsics.

Calling `IsAsyncMethod()` should also be faster than signature parsing, although not sure if it would matter.

The signature parsing routine got a bit simpler too.

### DIFF
--- a/src/coreclr/jit/importercalls.cpp
+++ b/src/coreclr/jit/importercalls.cpp
@@ -523,18 +523,6 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
         {
             call->AsCall()->gtCallMoreFlags |= GTF_CALL_M_SPECIAL_INTRINSIC;
         }
-
-        // Temporary hack since these functions have to be recognized as async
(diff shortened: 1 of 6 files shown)


jakobbotsch (approved):

VSadov:
Thanks!

Issue #2472, opened 2023-12-26
[NativeAOT-LLVM] Post-merge tasks


SingleAccretion:
This issue is intended to be a continuously updated list of things that should be done after the next upstream merge or in the merge PR itself (if necessary for a "clean" merge).

Current list:
- Switch `NativeAotSupported` to `true`. This shouldn't have the fallout it did last time with https://github.com/dotnet/runtime/pull/103273. (?)
- Remove our cmake code to use C++17 in the LLVM Jit: https://github.com/dotnet/runtime/pull/118858.
- Remove `WASM_BIGINT` and `-mnontrapping-fptoint` conditions from our targets on the next Emscripten (> 4.0.0) bump.
  - Latest bump ref: https://github.com/dotnet/runtime/pull/119409.
- Revert changes from https://github.com/dotnet/runtime/pull/125629.

yowl:
Will we be able to remove `Microsoft.DotNet.ILCompiler.SingleEntry.targets`,  https://github.com/dotnet/runtimelab/pull/2248/files#diff-29f00866a1dc9b2d5c3acb94480513bacbc05775e9ac3dea25215287a20d979aR3 at some point?

SingleAccretion:
> Will we be able to remove `Microsoft.DotNet.ILCompiler.SingleEntry.targets` at some point?

Not sure. It seems a fine way to share upstream code while pulling runtime pieces from NuGets for now.

PR #3083, opened 2025-04-13
[NativeAOT-LLVM] Do some pending post-merge tasks

SingleAccretion:
Minor changes that need to/can be done after the recent merge. See also https://github.com/dotnet/runtimelab/issues/2472.

### DIFF
--- a/src/coreclr/jit/llvmlssa.cpp
+++ b/src/coreclr/jit/llvmlssa.cpp
@@ -360,27 +360,6 @@ private:
                     continue;
                 }
 
-                // Handle a special case: calls with return buffer pointers need them pinned.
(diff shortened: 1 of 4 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvmlssa.cpp
```diff
@@ -360,27 +360,6 @@ class ShadowStackAllocator
                     continue;
                 }
 
-                // Handle a special case: calls with return buffer pointers need them pinned.
```
yowl, line 363: Why don't we need this anymore?

yowl (approved):
Looks good, thanks

Issue #3074, opened 2025-04-04
[NativeAOT-LLVM] Preinitialization test fails


yowl:
<CLRTestTargetUnsupported Condition="'$(IlcMultiModule)' == 'true' or '$(TargetsWasm)' == 'true'">true</CLRTestTargetUnsupported>

SingleAccretion:
The issue is that our codegen for casts saturates based on the target small type, while (I assume) the correct semantic is to saturate through `int`. Will confirm this and post a fix as appropriate.

PR #3084, opened 2025-04-13
[NativeAOT-LLVM] Fix saturating FP casts to small types

SingleAccretion:
Match upstream behavior (we differ because there is an ifdef in morph around the code that inserts the intermediate cast).

Fixes https://github.com/dotnet/runtimelab/issues/3074.


SingleAccretion:
@dotnet/nativeaot-llvm

pavelsavara (approved):

PR #3086, opened 2025-04-14
[RuntimeAsync] Possible way to optimize GetParallelMethodDesc for async (if needed)

VSadov:
This is a possible way to speed up the GetParallelMethodDesc - in most cases switching from linear to constant time.  
(where N is the number of methoddescs in a type)

It is a space-for-speed trade that we typically do only for something we need to look up at run time. At JIT time may be not enough reason. 
We should know at some point if this is necessary

I am not going to merge this. Just to show that it is possible and what a potential optimization could entail.
(per item ` (VM) Optimize MethodTable::GetParallelMethodDesc` in https://github.com/dotnet/runtime/pull/113976)https://github.com/dotnet/runtime/pull/113976

### DIFF
--- a/src/coreclr/vm/methodtablebuilder.cpp
+++ b/src/coreclr/vm/methodtablebuilder.cpp
@@ -5631,7 +5631,8 @@ MethodTableBuilder::PlaceNonVirtualMethods()
 #endif // _DEBUG
 
         if (!fCanHaveNonVtableSlots ||
-            it->GetMethodType() == mcInstantiated)
+            it->GetMethodType() == mcInstantiated ||
+            it->GetAsyncMethodKind() != AsyncMethodKind::NotAsync)
         {
             // We use slot during remoting and to map methods between generic instantiations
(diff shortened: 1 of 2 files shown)


src/coreclr/vm/methodtablebuilder.cpp
```diff
@@ -5631,7 +5631,8 @@ MethodTableBuilder::PlaceNonVirtualMethods()
 #endif // _DEBUG
 
         if (!fCanHaveNonVtableSlots ||
-            it->GetMethodType() == mcInstantiated)
+            it->GetMethodType() == mcInstantiated ||
+            it->GetAsyncMethodKind() != AsyncMethodKind::NotAsync)
```
VSadov, line 5635: This would force async/task-returning methods to have real slots. This is the "size"  tradeoff.

VSadov:
@jakobbotsch - just in case you have questions, thoughts other ideas.

PR #3089, opened 2025-04-15
[RuntimeAsync] Mark continuation GC return for arm64/arm32/LA64/RISCV64

jakobbotsch:
This was missed when support was added outside xarch.

cc @VSadov


VSadov (approved):
LGTM!

PR #3090, opened 2025-04-15
[RuntimeAsync] Model async continuation definition via delay-free

jakobbotsch:
Instead of introducing a new mechanism to mark the async continuation register as busy until a kill we can use the delay-free mechanism, provided we take a bit care about where we insert the ref positions and about ensuring the async continuation ends up in the right spot.

### DIFF
--- a/src/coreclr/jit/lower.cpp
+++ b/src/coreclr/jit/lower.cpp
@@ -733,8 +733,7 @@ GenTree* Lowering::LowerNode(GenTree* node)
             break;
 
         case GT_ASYNC_CONTINUATION:
-            LowerAsyncContinuation(node);
-            break;
+            return LowerAsyncContinuation(node);
 
         case GT_RETURN_SUSPEND:
             LowerReturnSuspend(node);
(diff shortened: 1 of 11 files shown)


jakobbotsch:
cc @VSadov

src/coreclr/jit/lower.cpp
```diff
@@ -5447,9 +5453,13 @@ void Lowering::LowerAsyncContinuation(GenTree* asyncCont)
                 node->AsCall()->gtIsAsyncCall = true;
             }
 
+            BlockRange().Remove(asyncCont);
+            BlockRange().InsertAfter(node, asyncCont);
```
VSadov, line 5457: If we go backwards from continuation to find the call, why do we need to force the order? Would it be "after" anyways?"

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -998,6 +992,11 @@ int LinearScan::BuildCall(GenTreeCall* call)
     buildInternalRegisterUses();
 
     // Now generate defs and kills.
+    if (call->IsAsync() && compiler->compIsAsync() && !call->IsFastTailCall())
+    {
+        MarkAsyncContinuationBusyForCall(call);
+    }
+
```
VSadov, line 999: Should another `MarkAsyncContinuationBusyForCall` below be deleted?

src/coreclr/jit/lower.cpp
VSadov: I see - to make sure there is nothing in between.

VSadov (approved):
LGTM!

src/coreclr/jit/lsrariscv64.cpp
jakobbotsch: Yes, good catch

VSadov:
After fetching from the async-experiment am getting an assert on x86.
Could it be related to the change?

```
 "E:\A2\runtimelab\artifacts\tests\coreclr\windows.x86.Checked\Tests\Core_Root\corerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false" -p "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization=true"  asynctests.dll
14:34:01.026 Running test: async\awaitingnotasync\awaitingnotasync.dll

Assert failure(PID 9104 [0x00002390], Thread: 47108 [0xb804]): Assertion failed '!"Missing BBF_PROF_WEIGHT flag"' in 'AwaitNotAsync:get_sProp():int' during 'Transform async' (IL size 37; hash 0x49da8969; FullOpts)

    File: E:\A2\runtimelab\src\coreclr\jit\fgprofile.cpp:4756
    Image: E:\A2\runtimelab\artifacts\tests\coreclr\windows.x86.Checked\Tests\Core_Root\corerun.exe
```

jakobbotsch:
No, changes from this PR wouldn't have any effect during the async transformation. This is likely a latent bug that is now showing up after merging because of recent JIT flowgraph changes (e.g. https://github.com/dotnet/runtime/pull/114016). I will take a look when time permits.

VSadov:
I think I have run x86 since the merge, but maybe I didn't or scenario needs some more conditions that were not there.

PR #3094, opened 2025-04-16
[RuntimeAsync] Propagate profile weight flag for shared return BB on x86

jakobbotsch:
Fixes https://github.com/dotnet/runtimelab/pull/3090#issuecomment-2810874361

cc @VSadov


VSadov (approved):
LGTM! It does fix the x86 assert. Thanks!

PR #3095, opened 2025-04-18
[RuntimeAsync] Feedback from the merging PR in the runtime repo.

VSadov:
This is to keep track of the changes done to address PR feedback in https://github.com/dotnet/runtime/pull/114675


VSadov:
I think it is time to merge this as we need to update to the latest API shape   (i.e. `AsyncHelpers`)

src/coreclr/vm/jithelpers.cpp
```diff
@@ -1685,25 +1685,35 @@ HCIMPL1(void, IL_ThrowExact, Object* obj)
...
-
     OBJECTREF oref = ObjectToOBJECTREF(obj);
+    GetThread()->GetExceptionState()->SetRaisingForeignException();
+
+#ifdef FEATURE_EH_FUNCLETS
+    Thread *pThread = GetThread();
+    
+    SoftwareExceptionFrame exceptionFrame;
+#ifdef TARGET_X86
+    _ASSERTE(!"Introduce `IL_ThrowExact_x86` following the pattern of other Throw helpers.");
```
jakobbotsch, line 1696: Does it mean the EH test fails on x86 now? Can we disable it somehow?

jakobbotsch (approved):
LGTM

src/coreclr/vm/jithelpers.cpp
VSadov: x86 does not do `FEATURE_EH_FUNCLETS` right now so it will work. Thus we can postpone introducing `IL_ThrowExact_x86`.

When x86 is moved onto `FEATURE_EH_FUNCLETS` plan, which I understand will happen at some point, we can remove `HELPER_METHOD_FRAME_BEGIN_ATTRIB_NOPOLL` path, which is desirable, but will also need to implement `IL_ThrowExact_x86`. It takes an extra parameter, I have only vague idea why.

VSadov: All runtime tests should be passing right now, including async and including EH - that is on win-x64 and win-x86, where it is easy to test once in a while.

VSadov:
Thanks!

src/coreclr/vm/jithelpers.cpp
jkotas: > When x86 is moved onto FEATURE_EH_FUNCLETS plan, which I understand will happen at some point, 

It is expected to happen very soon. The changes are in https://github.com/dotnet/runtime/pull/113576. The last remaining item before this gets enabled is private diagnostic tests run. If you would like to get ahead of this, you can apply the 10 line change locally and do the required fixes to make things work.

am11: Context about last testing item: https://github.com/dotnet/runtime/issues/113985#issuecomment-2798207726.

filipnavara: > It takes an extra parameter, I have only vague idea why.

win-x86 doesn't have unwinding information for native code. Thus we need to capture the precise register context in assembly code before giving control to the C code since we cannot run VirtualUnwind to get it.

This should be fairly mechanical change so I can make a PR for the code.

filipnavara: https://github.com/dotnet/runtimelab/pull/3100

PR #3097, opened 2025-04-21
[RuntimeAsync] Replace async marker with a flag

jakobbotsch:
This was a standalone bool because I got tired of conflicts. Now that we are about to merge in it should be changed back.

Also add a high level description of the transformation that happens in async.cpp.


VSadov (approved):
LGTM!

PR #3099, opened 2025-04-23
[RuntimeAsync] Use approved value for MethodImplAttributes.Async

VSadov:
I missed this change in the PR that adjusted the feature to use the approved API. 

`MethodImplAttributes.Async` now has a different value, to make sure there is no unnecessary confusion with formerly used values.

This requires updating Roslyn ref.
Otherwise it is just a cherrypick of corresponding change in [the "Moving to main" PR.](https://github.com/dotnet/runtime/pull/114675)


jakobbotsch (approved):

VSadov:
Thanks!

PR #3101, opened 2025-04-25
[RuntimeAsync] Put back BypassReadyToRun untill we are sure about R2R support

VSadov:
This fixes async tests when run standalone - regardless x64 or x86.

Once the AOT typesystem understands `MethodImpl.Async` and its impact on calling convention, we should be able to remove `[BypassReadyToRun]` from the helpers, but not just yet.


VSadov:
@jakobbotsch I am going to merge this, since it is a partial revert to what we had before, and because it fixes tests, which is somewhat blocking.

PR #3104, opened 2025-04-28
[NativeAOT-LLVM] Migrate type debug info emission in code modules to the push model

SingleAccretion:
This unifies the debug type info emission code to use the "push" model for both code and the final "debug" module.

This is the continuation of https://github.com/dotnet/runtimelab/pull/2883, where the push model was introduced, and undoes the duplication that resulted from having two ways to emit LLVM DI metadata.

No debug info diffs.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.DebugInfo.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.DebugInfo.cs
@@ -33,183 +33,13 @@ namespace Internal.JitInterface
     //    object writing, which is single-threaded anyway, 'phase 2' is essentially free TP-wise.
     //
     // Notably, the current implementation has a lot of room for TP improvement: we don't need all the information UDTD provides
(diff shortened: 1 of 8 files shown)


src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.DebugInfo.cs
```diff
@@ -366,8 +186,96 @@ IEnumerable<DebugVarInfoMetadata> GetResults()
 
             return GetResults();
         }
+    }
+
+    internal abstract unsafe class LlvmDebugInfoEmit : ITypesDebugInfoWriter
```
SingleAccretion, line 191: This is the key change: `LlvmDebugInfoEmit` is the base class of `LlvmCodeDebugInfoEmit`, which produces type forwards for structs/classes/arrays/enums and is used when emitting code, and `LlvmTypesDebugInfoEmit`, which produces type definitions and is used to write out the "debug" module.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
I'm not that familiar with the issue resolved here, but from the comments in the code this looks like a nice change.

jkotas:
Thank you!

PR #3113, opened 2025-05-29
[NativeAOT-LLVM] Fix `GetDebugInfoForCurrentMethod`

SingleAccretion:
We can get here more than once when the Jit retries a compilation (e. g. after a failed `noway_assert`).

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.DebugInfo.cs
+++ b/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.DebugInfo.cs
@@ -44,8 +44,13 @@ namespace Internal.JitInterface
         private void GetDebugInfoForCurrentMethod(CORINFO_LLVM_METHOD_DEBUG_INFO* pInfo)
         {
             *pInfo = default;
-            MethodDesc method = _methodCodeNode.Method;
+            if (_debugInfo == null)
+            {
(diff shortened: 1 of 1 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.Llvm.DebugInfo.cs
```diff
@@ -44,8 +44,13 @@ internal sealed unsafe partial class CorInfoImpl
         private void GetDebugInfoForCurrentMethod(CORINFO_LLVM_METHOD_DEBUG_INFO* pInfo)
         {
             *pInfo = default;
-            MethodDesc method = _methodCodeNode.Method;
+            if (_debugInfo == null)
```
yowl, line 47: I see that this guards against a NRE in the `foreach` below when it has already been determined that there is no debug info, but what does `_debugInfo` contain the first time in?

SingleAccretion: This field is set in `CompileMethodInternal` calling `SetDebugInformation`, so the first time around we get a valid debug info object (which may not contain any _useful_ info, however, aka the null object pattern).

Edit: the reason for the nulling out of the field is to know at final DI emission time whether the method had native DI attached or not without duplicating the logic here.

yowl (approved):
Thanks!

PR #3115, opened 2025-05-30
[NativeAOT-LLVM] Add linux-arm64 build and testing

SingleAccretion:
We do not have linux-arm64-hosted build machines, only linux-x64 crosscompiling to arm64, so while the build process is fairly straightforward (tunneling the "host target arch" though the `.yml`s), testing is a bit tricky.

The solution I've settled on is to just use qemu user mode emulation, the ILC targets only needed a very minimal change to enable this. So while we do produce real arm64 packages and all, the only part that's "arm64" in this new build is this emulated ILC. Fortunately, it is also the only part that "needs" to be arm64.

It is an imperfect approach, since e. g. it doesn't catch arm64-specific issues in downstream components like EMSDK/WASI-SDK, but it is better than the current state of no arm64 at all. Even the "proper" alternative of using Helix would require us to move the whole build there to exercise native arm64 EMSDK/WASI-SDK.

Another downside to this is the Linux-only nature of qemu, so, unfortunately, Windows ARM64 is not opened up by this change.

Important: I am not adding anything to the official build yet. It will be a separate change, so that we can revert it by itself if needed.

### DIFF
--- a/eng/liveBuilds.targets
+++ b/eng/liveBuilds.targets
@@ -32,7 +32,8 @@
          does not scale for local development because it requires you to build the same host tools for each target configuration.
          E. g. you need to build the same Debug compiler 3 times if you want to compile code in Debug, Checked and Release. -->
     <IlcConfig Condition="'$(IlcConfig)' == ''">$(CoreCLRConfiguration)</IlcConfig>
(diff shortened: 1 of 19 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm The linux-arm64 change.

The CI failure is unrelated (https://github.com/dotnet/runtime/issues/116746).

eng/pipelines/runtimelab/runtimelab-pre-build-steps.yml
yowl, line 1: What is this for?

SingleAccretion: Nothing; a loose file I forgot to remove from some earlier change. Deleted.

SingleAccretion:
More dotnet/runtime/issues/116746-like CI failures...

yowl (approved):
LGTM, thanks!

jkotas:
Thanks

PR #3118, opened 2025-06-17
[NativeAOT-LLVM] Enhance unresolved symbol handling

SingleAccretion:
Stop silently ignoring all unresolved symbols on WASI, warn instead, consistent with Emscripten. Also, respect `TreatWarningsAsErrors`.

### DIFF
--- a/eng/pipelines/runtimelab/install-emscripten.ps1
+++ b/eng/pipelines/runtimelab/install-emscripten.ps1
@@ -18,10 +18,16 @@ git checkout ca7b40ae222a2d8763b6ac845388744b0e57cfb7
 ./emsdk install 3.1.56
 ./emsdk activate 3.1.56
 
-
 if ($CI)
 {
     Write-Host "Setting EMSDK to '$InstallDir/emsdk'"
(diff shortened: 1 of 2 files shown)


eng/pipelines/runtimelab/install-emscripten.ps1
```diff
@@ -18,10 +18,16 @@ git checkout ca7b40ae222a2d8763b6ac845388744b0e57cfb7
...
 
-
 if ($CI)
 {
     Write-Host "Setting EMSDK to '$InstallDir/emsdk'"
     Write-Output "##vso[task.setvariable variable=EMSDK]$InstallDir/emsdk"
+
+    if ($env:EMSDK_PYTHON)
+    {
+        # Workaround for https://github.com/dotnet/runtime/issues/116746
```
SingleAccretion, line 28: (I don't actually know if this actually works, but it _seems_ to have helped. The idea here was that the CI build is picking up system python instead of the emsdk bundled python, and that has some issues.)

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks!

jkotas (approved):
Thanks

PR #3120, opened 2025-06-24
[NativeAOT-LLVM] Replace extern method accessors with cells

SingleAccretion:
Before:
```
  <push args>
  call get.method
  call_indirect
```

After:
```
  <push args>
  i32.const 0
  i32.load 2 <cell.method>
  call_indirect
```

This is a marginally larger sequence for a marginally smaller payload of the cell itself (4 bytes vs 8 for an accessor). It has worse "best-case" characteristics (WASM compilers won't be able to understand the indirect call target and potentially inline the call), but better "average case" ones (e. g. wasmtime currently does not do any WASM inlining at all).

However, the main motivation for this change is better compatibility with the -fPIC case, where the static table index relocation makes it impossible to use the ILC-produced code as-is. While we don't support -fPIC, we do have some users who use ILC directly and then do their own bitcode compilation and such.

### DIFF
--- a/src/coreclr/jit/llvm.cpp
+++ b/src/coreclr/jit/llvm.cpp
@@ -777,11 +777,10 @@ const char* Llvm::GetAlternativeFunctionName()
     return CallEEApi<EEAI_GetAlternativeFunctionName, const char*>(m_pEECorInfo);
 }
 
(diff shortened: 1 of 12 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Looks good, thanks!

PR #3122, opened 2025-06-25
[NativeAOT-LLVM] Stop passing the signatures around for external methods

SingleAccretion:
No longer needed now that we have settled on having ABI handling in ILC anyway. This is a no-diff change.

I've also added a convenience `clr.ilc` subset that only builds ILC, but does not publish it, intended for inner loop development (e. g. `./build clr.wasmjit+clr.ilc`).

Depends on #3120.

### DIFF
--- a/eng/Subsets.props
+++ b/eng/Subsets.props
@@ -179,7 +179,7 @@
 
   <ItemGroup>
     <!-- NativeAOT-LLVM subsets. -->
-    <SubsetName Include="NativeAot.Packages" Description="The CoreCLR native AOT compiler." />
+    <SubsetName Include="NativeAot.Packages" Description="The CoreCLR native AOT packages." />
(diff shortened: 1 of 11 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks, nice to remove some code.

eng/Subsets.props
```diff
@@ -178,9 +178,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- CoreClr Native AOT -->
+    <!-- NativeAOT-LLVM subsets. -->
     <SubsetName Include="NativeAot.Packages" Description="The CoreCLR native AOT compiler." />
```
jkotas:
```suggestion
    <SubsetName Include="NativeAot.Packages" Description="The CoreCLR native AOT packages." />
```

PR #3136, opened 2025-07-18
[NativeAOT-LLVM] Add detailed logging for not-spilling reasons to LSSA

SingleAccretion:
Workings towards fixing #3135.

Examples:
```scala
 -- Processing a safe point:
N010 ( 43, 21) [000046] -ACXG+-----                         *  CALL      int    System.Text.Unicode.Utf8Utility:TranscodeToUtf16(uint,int,uint,int,byref,byref):int $242
V42/2: GC_EXPOSED_UNKNOWN -> GC_EXPOSED_YES
V42/2 is live, inserted a spill in BB02:
               [000575] -----------                  t575 =    LCL_VAR   byref  V42 tmp27        u:2 LLVM REG LLVM
                                                            /--*  t575   byref
               [000576] DA---------                         *  STORE_LCL_VAR byref  V42 tmp27
V42/2: GC_EXPOSED_YES -> GC_EXPOSED_SPILL
BlockShadowSlotsState[BB02]::Defs: adding V42
V44/2: GC_EXPOSED_UNKNOWN -> GC_EXPOSED_YES
V44/2 is live, inserted a spill in BB02:
               [000577] -----------                  t577 =    LCL_VAR   byref  V44 tmp29        u:2 LLVM REG LLVM
                                                            /--*  t577   byref
               [000578] DA---------                         *  STORE_LCL_VAR byref  V44 tmp29
V44/2: GC_EXPOSED_YES -> GC_EXPOSED_SPILL
BlockShadowSlotsState[BB02]::Defs: adding V44
V02/1: GC_EXPOSED_UNKNOWN -> GC_EXPOSED_YES
V02/1 is live, inserted a spill in BB01:
               [000579] -----------                  t579 =    LCL_VAR   byref  V02 arg2         u:1 LLVM REG LLVM
                                                            /--*  t579   byref
               [000580] DA---------                         *  STORE_LCL_VAR byref  V02 arg2
V02/1: GC_EXPOSED_YES -> GC_EXPOSED_SPILL
BlockShadowSlotsState[BB01]::Defs: adding V02
V03/1: GC_EXPOSED_UNKNOWN -> GC_EXPOSED_YES
V03/1 is live, inserted a spill in BB01:
               [000581] -----------                  t581 =    LCL_VAR   byref  V03 arg3         u:1 LLVM REG LLVM
                                                            /--*  t581   byref
               [000582] DA---------                         *  STORE_LCL_VAR byref  V03 arg3
[...]

PR #3138, opened 2025-07-21
[NativeAOT-LLVM] Stop caching `GC_EXPOSED_NO` results in LSSA

### LINKED ISSUE #3135: [NativeAOT-LLVM] LSSA issues with caching the GC exposed status
The LSSA algorithm assumes that once a def has been spilled, it can be relied upon to root downstream (derived) defs. This is only true however if the def's spilled value doesn't change (on the shadow stack), which is currently only verified for the root def. This bug can cause GC holes.

Example of a suspect chain could be something like below:
```cs
V45/3 def <= V29/2 def <= ADD [000251] <= V26/2 def <= ADD [000206] <= V44/2 def <= spilled due to [000046]
```
Here we've cached the GC exposed status for `V26/2` as `GC_EXPOSED_NO`, but in reality it is dependent on `V44/2`'s spilled value staying active at the point when `V45/3` crosses a safe point.

SingleAccretion:
Since they can't actually be cached.

The diffs for this (perhaps somewhat surprisingly) are rather modest (HelloWasm):
```
Summary of Code Size diffs:
(Lower is better)

Total bytes of base: 3031558
Total bytes of diff: 3031581
Total bytes of delta: 23 (0.00% % of base)
Average relative delta: 3.08%
    diff is a regression
    average relative diff is a regression

Top method regressions (percentages):
           9 ( 7.20% of base) : 1002.dasm - HelloWasm_System_Runtime_JitTesting_FunctionalLSSATests__SpilledCandidateSlotInvalidated
           7 ( 1.24% of base) : 1000.dasm - S_P_CoreLib_System_Text_Unicode_Utf8__ToUtf16
           7 ( 0.81% of base) : 1001.dasm - S_P_CoreLib_System_Reflection_Runtime_BindingFlagSupport_QueriedMemberList_1<System___Canon>__Create

3 total methods with Code Size differences (0 improved, 3 regressed)
```
Fixes #3135.
Depends on #3136.

### DIFF
--- a/src/coreclr/jit/llvmlssa.cpp
+++ b/src/coreclr/jit/llvmlssa.cpp
@@ -615,6 +615,16 @@ private:
 
         void SpillSdsuValue(LIR::Range& blockRange, GenTree* defNode, unsigned* pSpillLclNum)
         {
+            // Like with candidates, the GC status of an SDSU can change during its lifetime, so we need to check it
+            // at every safepoint, in the general case. We could tighten this by filtering out more SDSUs early (like
+            // 'null' and such), but live-across-a-safepoint SDSUs are not that common, so we currently don't.
+            INDEBUG(NotExposedReason reason(this));
+            if (!IsGcExposedSdsuValue(defNode, DefStatus::Active DEBUGARG(&reason)))
+            {
+                JITDUMPEXEC(reason.Print("[%06u] is live, but not exposed: ", "\n", Compiler::dspTreeID(defNode)));
+                return;
+            }
+
             if (*pSpillLclNum != BAD_VAR_NUM)
             {
                 // We may have already spilled this def live across multiple safe points.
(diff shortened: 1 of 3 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks, do you think a Wasm/shadow stack focused LSSA algorithm would be beneficial?

SingleAccretion:
> Thanks, do you think a Wasm/shadow stack focused LSSA algorithm would be beneficial?

Hmm, I am not sure what you have in mind with that.

My current thinking on the "next generation" LSSA algorithm is building a graph of SSA definitions with collapsed strongly-connected parts and then doing local optimizations on that representation (e. g. if B = A, C = A, both B and C are exposed, we move the spill from B and C to A instead, so that it can root both).

PR #3141, opened 2025-07-23
[NativeAOT-LLVM] Separate out trivial includes and types from llvm.h

SingleAccretion:
These can be moved out to places where they're actually used without many/any other changes.

This helps compile times and makes this more local.

Also move some old Jit/EE enums to the shared file.

Ultimately I want to make it so that `llvm.h` doesn't include _any_ LLVM headers, but that requires factoring out codegen into its own class (or compromising by hiding `_builder` behind an indirection). Perhaps that would be a good change on its own...


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks, LGTM!

PR #3146, opened 2025-08-17
[NativeAOT-LLVM] Do not specify the compilation directory for LLVM modules

SingleAccretion:
Fixes a bug reported on Discord.

In DWARF, the final source code path as determined by the debugger
consists of three parts, recorded in .debug_line:
1) The compilation directory (for a compile unit).
2) The "include directory".
3) The file name.

If the "include directory" is relative (i. e. does not start with 'C:'
or '/'), the final path is going to be the concatenation of all three.

We don't control the "include directory" as it comes directly from
the managed PDBs, and we can't even know if it's truly relative or
an absolute path that just looks like a relative one (e. g. a URL).

Concatenating such an absolute path with any compilation directory
we might have is going to be wrong, so this change stops emitting it.

### DIFF
--- a/src/coreclr/jit/llvmdebuginfo.cpp
+++ b/src/coreclr/jit/llvmdebuginfo.cpp
@@ -268,21 +268,18 @@ private:
     DICompileUnit* InitializeCompileUnit()
     {
         StringRef path = m_context->Module.getName();
-        StringRef dir;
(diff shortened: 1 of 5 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

PR #3151, opened 2025-08-20
[NativeAOT-LLVM] Clean up physical memory computation for WASM

SingleAccretion:
This doesn't actually fix much (the GC mostly ignores these values currently), but it does make the code more correct, in particular on WASI where `sysconf(SYSCONF_PAGES)` isn't implemented.

### DIFF
--- a/src/coreclr/gc/unix/gcenv.unix.cpp
+++ b/src/coreclr/gc/unix/gcenv.unix.cpp
@@ -184,6 +184,8 @@ static uint8_t* g_helperPage = 0;
 // Mutex to make the FlushProcessWriteBuffersMutex thread safe
 static pthread_mutex_t g_flushProcessWriteBuffersMutex;
 
+uint64_t GetTotalPhysicalMemory_Wasm();
+
 size_t GetRestrictedPhysicalMemoryLimit();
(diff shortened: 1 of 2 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/gc/unix/gcenv.unix.cpp
```diff
@@ -184,6 +184,8 @@ static uint8_t* g_helperPage = 0;
 // Mutex to make the FlushProcessWriteBuffersMutex thread safe
 static pthread_mutex_t g_flushProcessWriteBuffersMutex;
 
+uint64_t GetTotalPhysicalMemory_Wasm();
```
yowl, line 187: Does this need to be in a `#ifdef TARGET_WASM` ?

yowl (commented):
Just one comment

src/coreclr/gc/unix/gcenv.unix.cpp
SingleAccretion: It doesn't - unused declarations are ok.

yowl (approved):
Looks good. thanks!

PR #3154, opened 2025-08-22
[NativeAOT-LLVM] Fix null elements in DI

SingleAccretion:
A relatively recent addition to LLVM made this a verification failure (https://github.com/llvm/llvm-project/commit/2dc5682dacab2dbb52a771746fdede0e938fc6e9).

### DIFF
--- a/src/coreclr/jit/llvmdebuginfo.cpp
+++ b/src/coreclr/jit/llvmdebuginfo.cpp
@@ -328,8 +328,7 @@ private:
         // Forward-declare our structure to handle inheritance.
         llvm::TempDIType declType = llvm::TempDIType(
             m_diBuilder.createReplaceableCompositeType(DW_TAG_structure_type, "", nullptr, nullptr, 0));
-        unsigned debugElementsCount = (pInfo->BaseClass != NO_DEBUG_TYPE) + pInfo->InstanceFieldCount;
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/llvmdebuginfo.cpp
```diff
@@ -355,12 +354,12 @@ class TypeDebugInfoModule
 
     DIType* EmitEnumType(CORINFO_LLVM_ENUM_TYPE_DEBUG_INFO* pInfo)
     {
-        llvm::SmallVector<Metadata*, 24> elements(static_cast<size_t>(pInfo->ElementCount));
+        llvm::SmallVector<Metadata*, 32> elements(static_cast<size_t>(pInfo->ElementCount));
```
SingleAccretion, line 357: The `SmallVector<T>` constructor, unlike e. g. `List<T>::.ctor`, accepts size and not capacity. So we were constructing an array like `{null, null, null, elem0, elem1, elem2}` instead of `{elem0, elem1, elem2}`.

SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks!

yowl:
@jkotas would you be able to initiate the dnc-eng build again please?  It has failed on a known problem, a race condition perhaps.

jkotas:
It succeeded on rerun.

PR #3162, opened 2025-09-05
[NativeAOT-LLVM] Add support for WASI SDK 26.0/27.0

SingleAccretion:
This should unblock the wit-bindgen CI. This is expected to be temporary, until we move the new SDK upstream (and then downstream).


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas (approved):
Thanks!

PR #3166, opened 2025-09-11
[NativeAOT-LLVM] Reduce host package sizes

SingleAccretion:
Only pack ILC (without its 60MB PDB) into host packages. Only pack the target libraries into target packages.

Reduces the size of the host package from ~64MB to ~13MB.

This is a ported combination of https://github.com/dotnet/runtime/pull/116882 and https://github.com/dotnet/runtime/pull/102803.


SingleAccretion:
@dotnet/nativeaot-llvm

Issue #3152, opened 2025-08-22
[NativeAOT-LLVM] Make `IlcGenerateMstatFile` work


SingleAccretion:
The LLVM-compiled methods are not `ObjectNode`s since they represent 'externally' compiled code, therefore we need to do something special for them.

My initial idea was to:
1) Introduce a concept of an "external object node" to the node dumper, where it would emit relocations for the 'size' of those nodes, thus producing a "relocatable" mstat file (format TBD).
2) Write an MSBuild build task / some tool we can invoke from MSBuild that'd "relocate" this mstat by parsing the LLVM-produced object files.

Alternatively, we can simplify the setup by moving the invocation of `clang` from MSBuild to inside ILC. The downside is we'll lose the pretty timers that now run with `CompileWasmObject` and will have to manage process invoke ourselves. Perhaps that's ok since the longer-term goal is to move this part of the build into ILC anyway.

PR #3171, opened 2025-09-16
[NativeAOT-LLVM] Add Mstat support

SingleAccretion:
Since we only know the size of the methods once we compile them, introduce a "relocatable object node" concept:
1) We record LLVM methods at object emission time with 0 sizes, recording the location in the IL stream where the size is set.
2) We request the linker to produce a link map.
3) We "relocate" the mstat in an MSBuild task by parsing the aforementioned section together with the link map.

Fixes https://github.com/dotnet/runtimelab/issues/3152.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/MstatObjectDumper.Llvm.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/MstatObjectDumper.Llvm.cs
@@ -1,0 +1,76 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Ecma335;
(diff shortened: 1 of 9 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/MstatObjectDumper.Llvm.cs
```diff
@@ -0,0 +1,76 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
yowl: Duplicate license?

yowl (commented):
Thanks, one nit.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/MstatObjectDumper.Llvm.cs
SingleAccretion: Oops, fixed.

jkotas (approved):

Issue #3163, opened 2025-09-06
[NativeAOT-LLVM] Optimize GC transitions


SingleAccretion:
1) PI:
    - Make `TransitionFrame*` equal to the shadow stack top, s. t. the transition is just one store.
2) RPI:
    - Fold the shadow stack initialization into `RhpReversePInvoke` (return it from the call).
    - Think about how to improve unhandled exception handling - it currently also requires an on-entry and on-exit call to a runtime helper (for all RPI methods).

PR #3173, opened 2025-09-19
[NativeAOT-LLVM] Set up the shadow stack in LSSA

SingleAccretion:
This paves the way for introducing some RPI/PI optimizations and also moves us a bit closer to the model where all that's possible to do in IR is done in IR.

Setting up the shadow stack in IR is a tricky problem since we don't want to spill it to an alloca even in debug code. We therefore modify how we refer to the shadow stack:
1) For the main function, we use an untracked SSA local. Note how we are introducing a new concept here - usually all SSA locals are tracked.
2) For funclets, we repurpose the pre-existing PHYSREG node to allow us to refer to the LLVM argument directly.

However, the IR representation is just one part of the problem. Another is the fact codegen needs to refer to the shadow stack directly, for debug info and helper call generation purposes. Since some of this needs to happen "in the prolog", before any IR is generated, we also introduce a concept of "late" and "early" prologs. They are sequenced as follows:

 - "Early"  prolog - codegen (LLVM IR)
 - "Middle" prolog - LSSA (IR)
 - "Late"   prolog - codegen (LLVM IR)

This is almost a zero-diff change; the few diffs are due to the different ordering of native stack init and shadow stack init for RPI methods.

Part of https://github.com/dotnet/runtimelab/issues/3163.

### DIFF
--- a/src/coreclr/jit/llvmcodegen.cpp
+++ b/src/coreclr/jit/llvmcodegen.cpp
@@ -3238,8 +3238,7 @@ Value* Llvm::getShadowStack()
 {
     if (getCurrentLlvmFunctionIndex() == ROOT_FUNC_IDX)
     {
-        Value* value = _localsMap[{m_shadowStackLclNum, m_shadowStackSsaNum}];
-        return value;
+        return _localsMap[{m_shadowStackLclNum, m_shadowStackSsaNum}];
     }
 
     // Note that funclets also have the shadow stack arg in the 0th position.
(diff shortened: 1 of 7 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -3255,11 +3238,11 @@ Value* Llvm::getShadowStack()
 {
     if (getCurrentLlvmFunctionIndex() == ROOT_FUNC_IDX)
     {
-        assert(m_rootFunctionShadowStackValue != nullptr);
-        return m_rootFunctionShadowStackValue;
+        Value* value = _localsMap[{m_shadowStackLclNum, m_shadowStackSsaNum}];
+        return value;
```
yowl: nit, is `value` useful for debugging?

yowl (approved):
I only half understand this, and it looks fine to me.  Thanks.

src/coreclr/jit/llvmcodegen.cpp
SingleAccretion: We can remove this one. I do like this pattern for debugging purposes, but no need to do it proactively.

PR #3179, opened 2025-10-05
[NativeAOT-LLVM] Smoke test non-x64 ILC on Helix instead of using emulation

SingleAccretion:
The underlying goal here is to enable win-arm64 testing.

The main obstacle to this is the fact we can't test it with out existing approaches. There are no win-arm64 build machines, and there is no user mode qemu for Windows either.

The only remaining option is Helix (see https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.Helix/Sdk#readme). Thus, this change moves our testing of linux-arm64 from qemu to Helix. This is done by sending our samples and WASI SDK over, along with the published packages, and compiling those samples. It is very light validation, but doing more is both more complicated infrastructure-wise, and not absolutely necessary since the property we want to test is effectively that we compiled and linked the Jit correctly. We don't have, and don't expect to add, host-specific code to the Jit.

Along the way, I also had to change the platform from which we publish "host" packages from browser to WASI since sending the EMSDK over is rather more complicated than sending WASI SDK, and it's not as easy to procure one for ARM64 platforms. And we only need one ARM64 job per host OS.


SingleAccretion:
This will wait until https://github.com/dotnet/runtimelab/pull/3206 since there is a lot of YAML churn which has a high chance of breaking official build.

PR #3184, opened 2025-10-08
Pipe signing conditions, and disable signing in runtime lab

hoyosjs:
- **Pipe signing properties all the way down**
- **Disable signing in runtimelab**


hoyosjs:
Actually signing required the reverted commit. Its better not to sign in this repo - but it's better to add that change separately as the arcade change brings a lot more changes since the SDK does a lot more.

hoyosjs:
A better thing would be to take an update from main if possible.

jkotas (approved):
Thanks

SingleAccretion (approved):
Thank you!

> A better thing would be to take an update from main if possible.

To be clear, do you mean that once we merge upstream, this change _should_ be reverted? Or it's better to still leave the signing turned off?

jkotas:
> To be clear, do you mean that once we merge upstream, this change should be reverted? Or it's better to still leave the signing turned off?

It may be better to leave the signing turned off. Signing tends to require periodic maintenance.

PR #3195, opened 2025-10-28
[NativeAOT-LLVM] Add support for `<NativeLib>Static</NativeLib>`

SingleAccretion:
Also sink more WASM specifics into the WASM-specific targets.

Initially I also wanted to combine all runtime implementation libraries into one big archive with `llvm-ar qL`, but after thinking about it a bit more, it doesn't seem like it would move the needle that much in terms of presenting a clean public contract, since you still need to be very careful with both the linking order, and the arguments. So "copy `NativeLib=Shared` link command line and be prepared to alter it once in a while" it still is.


SingleAccretion:
@dotnet/nativeaot-llvm

heartacker:
wanderful

yowl (approved):
LGTM, thanks!

PR #3196, opened 2025-10-29
Add RequiresUnsafeAttribute and start annotating corelib

agocke:


src/libraries/System.Private.CoreLib/src/System/Delegate.cs
```diff
@@ -128,9 +128,15 @@ public TDelegate Current
             public bool MoveNext()
             {
                 int index = _index + 1;
-                if ((_current = Unsafe.As<TDelegate>(_delegate?.TryGetAt(index))) == null)
+                unsafe
                 {
-                    return false;
+                    unsafe
```
stephentoub, line 133: This one is really, really unsafe I guess :)

agocke: lol not sure what happened there

src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/RequiresUnsafeAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// Indicates that the specified method requires dynamic access to code that is not referenced
+    /// statically, for example through <see cref="Reflection"/>.
+    /// </summary>
+    /// <remarks>
+    /// This allows tools to understand which methods are unsafe to call when removing unreferenced
+    /// code from an application.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Constructor, Inherited = false)]
+    internal sealed class RequiresUnsafeAttribute : Attribute
```
tannergooding, line 15: Since we know some version of this attribute is going to be required, should we go ahead and take it to API review so we can land on the name (even if the exact constructors it contains may change over time)?

There's been a few names that have been floated around such as `RequiresUnsafe` (closest prior is `RequiresUnreferencedCode`) or `UnsafeCallersOnly` (closest prior is `UnmanagedCallersOnly`). There may be other ideas/opinions that more correctly conveys "this member is unsafe" as well.

src/libraries/System.Private.CoreLib/src/System/Buffers/SharedArrayPool.cs
```diff
@@ -60,7 +60,10 @@ public override T[] Rent(int minimumLength)
             SharedArrayPoolThreadLocalArray[]? tlsBuckets = t_tlsBuckets;
             if (tlsBuckets is not null && (uint)bucketIndex < (uint)tlsBuckets.Length)
             {
-                buffer = Unsafe.As<T[]>(tlsBuckets[bucketIndex].Array);
+                unsafe
+                {
+                    buffer = Unsafe.As<T[]>(tlsBuckets[bucketIndex].Array);
```
tannergooding, line 65: We could remove the unsafe here and just cast to `T[]` here, right?

The only reason this isn't already `T[]` and that it's using `System.Array` instead is to avoid the generic code size increase for the `SharedArrayPoolThreadLocalArray` struct, but maybe that's something we could track addressing as well if the cast is too expensive.

jkotas, line 65: This is a good case-study for the annotation guidelines. This is safe because we manually ensure that we only assign `T[]` into Array field. Should the Array field and all places that assign to it be marked unsafe as well?

jkotas: Right, this has perf fine-tuned using unsafe code. Number of other cases in this PR are in the same category.

Issue #3200, opened 2025-11-02
[NativeAOT-LLVM] Fix official build (originally 're-enable signing')


SingleAccretion:
Otherwise the default experience has just too many NuGet warnings:
```
    C:\Users\Accretion\AppData\Local\Temp\wasm\console\console.csproj : warning NU3018: Package 'Microsoft.DotNet.ILCompiler.LLVM 10.0.0-preview.2.25509.1' from source 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json': The author primary signature's signing certificate is not trusted by the trust provider.
    C:\Users\Accretion\AppData\Local\Temp\wasm\console\console.csproj : warning NU3027: Package 'Microsoft.DotNet.ILCompiler.LLVM 10.0.0-preview.2.25509.1' from source 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json': The signature should be timestamped to enable long-term signature validity after the certificate has expired.
    C:\Users\Accretion\AppData\Local\Temp\wasm\console\console.csproj : warning NU3018: Package 'runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM 10.0.0-preview.2.25509.1' from source 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json': The author primary signature's signing certificate is not trusted by the trust provider.
    C:\Users\Accretion\AppData\Local\Temp\wasm\console\console.csproj : warning NU3027: Package 'runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM 10.0.0-preview.2.25509.1' from source 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json': The signature should be timestamped to enable long-term signature validity after the certificate has expired.
    C:\Users\Accretion\AppData\Local\Temp\wasm\console\console.csproj : warning NU3018: Package 'runtime.browser-wasm.Microsoft.DotNet.ILCompiler.LLVM 10.0.0-preview.2.25509.1' from source 'https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-experimental/nuget/v3/index.json': The author primary signature's signing certificate is not trusted by the trust provider.
[...]

PR #3217, opened 2025-12-06
[NativeAOT-LLVM] Restore some removed upstream publishing infrastructure

SingleAccretion:
This is just a straight copy from https://github.com/dotnet/runtime/pull/108990.

Should in theory address the error in https://github.com/dotnet/runtimelab/issues/3200#issuecomment-3619512855. It is likely there are more missing pieces.

It is a big liability that we're staying on this old infrastructure. But it would require someone with access to the official build and knowledge of how this is now supposed to work to change that.

### DIFF
--- a/eng/Publishing.props
+++ b/eng/Publishing.props
@@ -5,6 +5,118 @@
     <AutoGenerateSymbolPackages>false</AutoGenerateSymbolPackages>
   </PropertyGroup>
 
+  <!-- NativeAOT-LLVM: using the old publishing method from circa https://github.com/dotnet/runtime/pull/108990 -->
+  <ItemGroup>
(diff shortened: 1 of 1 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas:
@hoyosjs PTLA

hoyosjs (approved):
not sure whether this will work until it goes through the build. That being said, if we aren't signing, an easier step might be to add `-publish` to the build and let it generate assets per leg long term.

SingleAccretion:
> That being said, if we aren't signing, an easier step might be to add -publish to the build and let it generate assets per leg long term.

I'll try to do that after we fix the current build and get at least one (working) package out, since there the old packages are rather broken currently.

PR #3219, opened 2025-12-11
[NativeAOT-LLVM] Bump wasi-sdk to v29

yowl:
Gets us the latest async component support in wasm-component-ld.

### DIFF
--- a/eng/native/configurecompiler.cmake
+++ b/eng/native/configurecompiler.cmake
@@ -24,9 +24,13 @@ include(${CMAKE_CURRENT_LIST_DIR}/configureoptimization.cmake)
 # Initialize Cmake compiler flags and other variables
 #-----------------------------------------------------
 
+message("Configuring compiler for ${CLR_CMAKE_HOST_UNIX} ${CLR_CMAKE_HOST_WASI} ${CMAKE_CXX_COMPILER_ID}")
 if (CLR_CMAKE_HOST_UNIX OR CLR_CMAKE_HOST_WASI)
+    message("UNIX or WASI")
     add_compile_options(-Wall)
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        message("Clang")
+
         add_compile_options(-Wno-null-conversion)
         add_compile_options(-Wno-jump-misses-init)
         add_compile_options(-Wno-c++-keyword)
--- a/eng/native/configurecompiler.cmake
+++ b/eng/native/configurecompiler.cmake
@@ -24,13 +24,9 @@ include(${CMAKE_CURRENT_LIST_DIR}/configureoptimization.cmake)
 # Initialize Cmake compiler flags and other variables
 #-----------------------------------------------------
 
-message("Configuring compiler for ${CLR_CMAKE_HOST_UNIX} ${CLR_CMAKE_HOST_WASI} ${CMAKE_CXX_COMPILER_ID}")
 if (CLR_CMAKE_HOST_UNIX OR CLR_CMAKE_HOST_WASI)
-    message("UNIX or WASI")
     add_compile_options(-Wall)
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-        message("Clang")
-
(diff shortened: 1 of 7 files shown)


yowl:
Cc @dotnet/nativeaot-llvm

eng/native/configurecompiler.cmake
```diff
@@ -24,9 +24,10 @@ include(${CMAKE_CURRENT_LIST_DIR}/configureoptimization.cmake)
 # Initialize Cmake compiler flags and other variables
 #-----------------------------------------------------
 
-if (CLR_CMAKE_HOST_UNIX)
+  if (CLR_CMAKE_HOST_UNIX OR CLR_CMAKE_HOST_WASI)
     add_compile_options(-Wall)
-    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    # CMake is not detecting Clang for Wasi and CMAKE_CXX_COMPILER_ID is empty 
+    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CLR_CMAKE_HOST_WASI)
```
SingleAccretion: Do you have a log for this? Is it a new problem with the new SDK or it has always been like this?

yowl: Not needed, must have confused myself, or some cache thing.

yowl:
I've removed the debug stuff, and this is ok for another review now, thanks.

src/native/libs/CMakeLists.txt
```diff
@@ -2,7 +2,8 @@ cmake_minimum_required(VERSION 3.20)
 
 include(CheckCCompilerFlag)
 
-project(LibsNative C)
+# NativeAOT-LLVM, adding CXX to configure compiler for CLang
+project(LibsNative C CXX)
```
SingleAccretion: I took another look at this and it turns out [these](https://github.com/dotnet/runtimelab/blob/0af42e640c4964ebfcc3ca7a9bc2a2f6d83011cc/eng/native/configurecompiler.cmake#L31-L34) suppressions are actually downstream-only.

So I think it would be better to move them from there to someplace else instead (e. g. https://github.com/dotnet/runtimelab/blob/29dfa22e8a676dfdf534c5364a9d3b4427a860a5/src/native/libs/CMakeLists.txt#L134-L136).

src/native/libs/CMakeLists.txt
```diff
@@ -2,8 +2,6 @@ cmake_minimum_required(VERSION 3.20)
 
 include(CheckCCompilerFlag)
 
-project(LibsNative C)
```
SingleAccretion, line 5: (Accidental wholesale removal?)

yowl: oops

yowl:
missing pdb file which I think is known, try again.

src/native/libs/System.Native/pal_networking.c
```diff
@@ -3400,7 +3400,7 @@ int32_t SystemNative_GetWasiSocketDescriptor(intptr_t socket, void** entry)
     }
 
     int fd = ToFileDescriptor(socket);
-    if(!descriptor_table_get_ref(fd, entry))
+    if(descriptor_table_get_ref(fd) == NULL)
```
SingleAccretion: Oh, sorry for not noticing earlier... This should initialize `*entry` shouldn't it, otherwise we're not doing the same thing anymore.

yowl: From https://github.com/alexcrichton/wasi-libc/blob/436494a6048624ab07a84696c2ceaff16d2b0b67/libc-bottom-half/sources/descriptor_table.c#L138 I will only set `entry` if `descriptor_table_get_ref` returns non null, i..e We won't NULL the pointer on failure.

SingleAccretion (approved):

yowl:
Thanks!, it was meant to be a small quick PR.

jkotas:
Thanks!

PR #3231, opened 2026-01-30
[NativeAOT-LLVM] More official build fixing (3)

SingleAccretion:
I've tried to find out how this worked upstream (when it was working), but haven't really come across anything illuminating, only some old issues on arcade like https://github.com/dotnet/arcade/issues/3744.

Should fix https://github.com/dotnet/runtimelab/issues/3200#issuecomment-3672636096.

### DIFF
--- a/eng/common/darc-init.ps1
+++ b/eng/common/darc-init.ps1
@@ -13,6 +13,7 @@ function InstallDarcCli ($darcVersion, $toolpath) {
   $dotnetRoot = InitializeDotNetCli -install:$true
   $dotnet = "$dotnetRoot\dotnet.exe"
   $toolList = & "$dotnet" tool list -g
+  $env:DOTNET_ROOT = $dotnetRoot # NativeAOT-LLVM context: https://github.com/dotnet/runtimelab/issues/3200#issuecomment-3672636096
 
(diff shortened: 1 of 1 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

jkotas (approved):

hoyosjs (commented):
Generally - you don't want to modify anything in eng/common since that's seen as an arcade responsibility.

hoyosjs:
Looks like this is missing https://github.com/dotnet/arcade/commit/9b521bdb03239f91830fb39a6ee2bf23bba07989

SingleAccretion:
> Looks like this is missing https://github.com/dotnet/arcade/commit/9b521bdb03239f91830fb39a6ee2bf23bba07989

Wow, thank you for finding that :). Maybe we can pick just that change up in #3233 ?

hoyosjs:
In general we tend to upgrade eng/common all in one step. I can undo some of the package upgrades. The official build passed with that PR but I see it has other issues.

SingleAccretion:
Superceded by https://github.com/dotnet/runtimelab/pull/3233.

PR #3233, opened 2026-02-04
Update eng/common files to BAR ID 299554

hoyosjs:
Upgrade only minimal parts of the shared infra to BAR ID 299554. The full dependency upgrade has issues with some of the TFMs of test projects being old (i.e. net8.0), so that can be done in a larger update of the branch.


hoyosjs:
@SingleAccretion - this updates the workflow. That being said, it's not great to be in this torn state.

SingleAccretion (approved):
Thank you!

> That being said, it's not great to be in this torn state.

Hopefully this problem solves itself on next merge without us having to think about it.

jkotas (approved):
Thanks!

PR #3241, opened 2026-02-15
[NativeAOT-LLVM] Support `LinkFlavor=lld`

SingleAccretion:
This enables building core modules that use WASIp2 interfaces.

Example:
```cs
[UnmanagedCallersOnly(EntryPoint = "add")]
private static int Add(int x, int y) => x + y;
```
```sh
> dotnet publish -r wasi-wasm /p:LinkFlavor=lld /p:NativeLib=shared
> wasmtime -W unknown-imports-default=y --invoke add example.wasm 2 40
42
```

By itself, this is not too interesting a capability, since you end up with a lot of unresolved WASIp2 imports that no runtime knows how to deal with. However, combining this with a static stub for those WASI interfaces, this can be used as a foundation to build things with custom external interfaces based on core modules.


SingleAccretion:
@dotnet/nativeaot-llvm

yowl (approved):
Thanks

Issue #3257, opened 2026-04-12
[NativeAOT-LLVM] Incorrect GC-aware struct copy when using [StructLayout] in nested structs


thomasfn:
## Summary

When used a struct with either Sequential or Explicit struct layout nested in a value type along with managed references, ILC appears to emit incorrect code for copying the struct, leading to corruption of data and in some cases outright heap corruption (writing outside of the stack frame or out-of-bounds into heap etc).

## Environment

- **Microsoft.DotNet.ILCompiler.LLVM**: `10.0.0-rc.1.26117.1`
- **runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM**: `10.0.0-rc.1.26117.1`
- **.NET SDK**: 10.0
- **RID**: `wasi-wasm`

## Reproduction

The following example demonstrates a managed struct with a nested unmanaged struct with three nested unmanaged structs, one with explicit, one with sequential and one without. The presence of the managed object in the outer struct is required to reproduce this (presumably without that it doesn't do a GC-aware copy and uses a more simple memcpy based copy). Only one of the sequential or explicit structs are needed, the issue repros with either. The issue does not repro if you don't use struct layout at all. There may be other factors at play but I've had trouble reducing it any more than this - it's already a heavily reduced example from my actual codebase which ran into this issue in a live environment.

Needless to say the issue does not repro in a regular C# JIT environment, it only happens in WASM/NativeAOT.

```CSharp
private struct TestNested(object obj, int i1, int i2, int i3, int i4, int i5, int i6)
{
    public object Object = obj;
    public TestSixInts SixInts = new(i1, i2, i3, i4, i5, i6);

    public override string ToString() => $"[{Object}, {SixInts}]";
}

[StructLayout(LayoutKind.Sequential, Size = 24)]
private struct TestSixInts(int i0, int i1, int i2, int i3, int i4, int i5)
{
    public TestTwoInts Normal = new(i0, i1);
    public TestTwoInts_Sequential Sequential = new(i2, i3);
    public TestTwoInts_Explicit Explicit = new(i4, i5);

    public override string ToString() => $"[{Normal}, {Sequential}, {Explicit}]";
}

[StructLayout(LayoutKind.Explicit, Size = 8)]
private struct TestTwoInts_Explicit(int i0, int i1)
{
    [FieldOffset(0)]
    public int I0 = i0;
    [FieldOffset(4)]
    public int I1 = i1;

[...]

PR #3259, opened 2026-04-22
[NativeAOT-LLVM] Fix copies with significant padding and GC refs

SingleAccretion:
The memcpy we were using was copying bytes from the copied-to object itself "back", therefore corrupting the already-copied data.

Replace with copies of padding fields we insert when constructing the LLVM type.

Fixes https://github.com/dotnet/runtimelab/issues/3257.

### DIFF
--- a/src/coreclr/jit/llvmcodegen.cpp
+++ b/src/coreclr/jit/llvmcodegen.cpp
@@ -2477,26 +2477,39 @@ void Llvm::consumeInitValAndEmitInitBlk(GenTree* initVal, Value* addrValue, Clas
 void Llvm::storeObjAtAddress(Value* baseAddress, Value* data, StructDesc* structDesc)
 {
     size_t fieldCount = structDesc->getFieldCount();
-    unsigned bytesStored = 0;
+    unsigned lastLlvmFieldIndex = INT32_MAX;
+
+    const llvm::StructLayout* dataLayout = nullptr;
+    if (data->getType()->isStructTy())
+    {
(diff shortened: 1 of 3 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

src/coreclr/jit/llvmcodegen.cpp
```diff
@@ -2489,33 +2477,47 @@ void Llvm::consumeInitValAndEmitInitBlk(GenTree* initVal, Value* addrValue, Clas
...
+    if (data->getType()->isStructTy())
+    {
+        dataLayout = m_context->Module.getDataLayout().getStructLayout(static_cast<llvm::StructType*>(data->getType()));
+    }
+    auto copyPaddingFields = [=, &lastLlvmFieldIndex](unsigned endLlvmFieldIndex) {
+        for (unsigned llvmFieldIndex = lastLlvmFieldIndex + 1; llvmFieldIndex < endLlvmFieldIndex; llvmFieldIndex++)
+        {
+            if (dataLayout != nullptr)
+            {
+                unsigned offset = static_cast<unsigned>(dataLayout->getElementOffset(llvmFieldIndex).getFixedValue());
```
yowl, line 2492: This is not the same as `fieldOffset ` in the caller presumably?

SingleAccretion: Yes, these are less than `fieldOffset`, for the fields between the ones described by the structure descriptor.

yowl (approved):
Thanks for the extra tests, looks good.

PR #3261, opened 2026-04-30
NativeAOT-LLVM: WIP Merge apr26

yowl:
Remove `Microsoft.NET.Workload.Mono.Toolchain.*Manifest-*.nupkg` from publishing.
Revert SDL swtich as it causes problems for LLVM (https://stackoverflow.com/questions/34892732/error-when-call-createphi-in-llvm)


yowl:
Will have a rethink, this is just taking too much time.

PR #3270, opened 2026-06-16
NativeAOT-LLVM: Update sample NativeLibraryHost.html script path for net10.0

DandiHira:
Updates the script source path in `NativeLibraryHost.html` from `net9.0` to `net10.0`. 

The `TargetFramework` of `NativeLibrary.csproj` is currently set to `net10.0`, so the published artifact path needs to be updated accordingly to prevent path mismatches.


SingleAccretion (approved):

jkotas (approved):
Thanks

PR #3272, opened 2026-06-19
[NativeAOT-LLVM] Update example usage

kajacx:
The generated binaries will be in a net10.0 folder. I also switched to forward slashes, since they work for all users.


SingleAccretion (approved):

jkotas (approved):

jkotas:
Thanks!

Issue #3271, opened 2026-06-18
[NativeAOT-LLVM] Incorrect GC-aware struct copy when using Span<T> over a nested managed struct


thomasfn:
## Summary

When using a `Span<T>` where T is a value type containing nested value types with object references, only the top-level object references use write barriers when a value is copied to the span. The others are copied using load/store without write barriers, leading to potential GC heap corruption as cards are not being set correctly, e.g. if the span refers to an array that's in gen1/2 and the object that was copied without the write barrier is in a younger generation.

_Edited_
It seems span isn't needed and just a simple array will repro this also.

_Edited2_
It's looking more like this is todo with generic instantiation of nested value types than specifically use of a span or an array, but it's troublesome to reproduce exactly and I'm way out of my depth on this so I'll leave this description as-is for now and hopefully someone more knowledgeable than I can crack the case for real.

## Environment

- **Microsoft.DotNet.ILCompiler.LLVM**: `10.0.0-rc.1.26306.1`
- **runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM**: `10.0.0-rc.1.26306.1`
- **.NET SDK**: 10.0
- **RID**: `wasi-wasm`

## Reproduction

The following example reproduces the incorrect codegen:

```CSharp
public struct RefWrapper(object value)
{
    public object Value = value;
}

[MethodImpl(MethodImplOptions.NoInlining)]
private static void TestAssignIssue()
{
    var testArr = new (object, RefWrapper)[1];
    TestGenericAssign(testArr, 0, (new object(), new RefWrapper(new object())));
    GC.KeepAlive(testArr);
}

[MethodImpl(MethodImplOptions.NoInlining)]
private static void TestGenericAssign<T>(Span<T> span, int index, T value)
{
    span[index] = value;
}
```

As with most GC issues it kind of requires the perfect storm to cause a crash.  It's tripping heap verification with `DOTNET_HeapVerify=1` and `DOTNET_GCStress=1` in my codebase, but seems to depend on the `Span` pointing to an array in gen2 and assigning items to it from a younger generation. A good example of this is `IEnumerable<T>.ToArray` which uses the `SegmentedArrayBuilder` which is renting arrays from an array pool (which thus make it to gen2 easily), where `T` is something like `(int, (int, object))` (the nesting is needed to confuse the codegen for `Span` ).

### Analysis

The generated code for `TestGenericAssign` above is as follows:

```wat
[...]

Issue #3268, opened 2026-06-14
[NativeAOT-LLVM] Incorrect GC-aware struct copy when using [InlineArray]


thomasfn:
## Summary

When using a managed struct with `[InlineArray]` applied, only the first element is copied using GC-aware field copy, the other elements are copied using bulk copy. This can lead to GC heap corruption.

## Environment

- **Microsoft.DotNet.ILCompiler.LLVM**: `10.0.0-rc.1.26306.1`
- **runtime.win-x64.Microsoft.DotNet.ILCompiler.LLVM**: `10.0.0-rc.1.26306.1`
- **.NET SDK**: 10.0
- **RID**: `wasi-wasm`

## Reproduction

The following example reproduces the incorrect codegen:

```CSharp
private struct InlineArrayOf3Objects
{
    [InlineArray(3)]
    private struct Slot
    {
        private object first;
    }

    private Slot slot;

    public object this[int i]
    {
        readonly get => slot[i];
        set => slot[i] = value;
    }
}

private class TestInline
{
    public InlineArrayOf3Objects A;

    public override string ToString()
    {
        return $"{A[0]}, {A[1]}, {A[2]}";
    }
}

[MethodImpl(MethodImplOptions.NoInlining)]
private static void TestInlineArray()
{
    var test1 = new TestInline();
    for (int i = 0; i < 3; ++i)
    {
        test1.A[i] = new object();
    }
    Console.WriteLine($"{test1}");

    var test2 = new TestInline();
    TestInlineArrayCopy(test1, test2);
    Console.WriteLine($"{test2}");
}

[MethodImpl(MethodImplOptions.NoInlining)]
private static void TestInlineArrayCopy(TestInline src, TestInline dst)
{
    dst.A = src.A;
}
```

Unfortunately I've had a hard time getting the minimal reproduction to actually cause a crash. It's definitely tripping heap verification with `DOTNET_HeapVerify=1` and `DOTNET_GCStress=1` in my codebase but usually happens sometime later and not right after the bad copy.

### Analysis

The generated code for `TestInlineArrayCopy` above is as follows:

```wat
 (func $TestInlineArrayCopy (param $0 i32) (param $1 i32) (param $2 i32)
  (local $3 i64)
  (i32.store
   (local.get $0)
   (i32.const 1107158)
  )
  (block $block
   (if
    (i32.eqz
     (i32.or
      (i32.eqz
       (local.get $1)
      )
      (i32.eqz
       (local.get $2)
[...]

PR #3274, opened 2026-06-24
[NativeAOT-LLVM] Fix missing cpobj write barriers

SingleAccretion:
Fixes https://github.com/dotnet/runtimelab/issues/3271.
Fixes https://github.com/dotnet/runtimelab/issues/3268.

### DIFF
--- a/src/coreclr/jit/llvmtypes.h
+++ b/src/coreclr/jit/llvmtypes.h
@@ -13,9 +13,10 @@
 struct TypeDescriptor
 {
     unsigned              Size;
+    unsigned              ElementCount;
     unsigned              FieldCount;
     CORINFO_FIELD_HANDLE* Fields;
-    unsigned              HasSignificantPadding;
+    bool                  HasSignificantPadding;
 };
 
 struct FieldDesc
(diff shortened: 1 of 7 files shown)


SingleAccretion:
@dotnet/nativeaot-llvm

andreakarasho:
Thank you! I hope this will get merged soon

src/coreclr/jit/llvmtypes.h
```diff
@@ -57,21 +58,37 @@ struct StructDesc
...
+        dsc->m_size = size;
+        dsc->m_fieldCount = fieldCount;
+        dsc->m_fields = fields;
+        dsc->m_hasSignificantPadding = hasSignificantPadding;
+        dsc->m_isInlineArray = false;
+        return dsc;
+    }
+
+    // This factory function takes the ownership of the passed in array of field descriptors.
+    static StructDesc* ForInlineArray(unsigned size, unsigned elementCount, FieldDesc* field, bool hasSignificantPadding)
```
yowl, line 83: These are very similiar, but I guess you prefer this for clarity?

yowl (approved):
Thanks, looks good.  Just the one comment on the similar functions.

src/coreclr/jit/llvmtypes.h
SingleAccretion: Yeah. I needed something to differentiate between the array and non-array cases... Another option was to expose `isInlineArray` as a ctor parameters, but it makes for a more muddied contract. In the end, it is ok to have a bit more code.

jkotas (approved):
Thanks

